PR #2737, opened 2021-08-12
chore(manifest): prevent incompatible/unsupported services in manifest

huanjani:
Throw errors when:
- `exec` is enabled with Windows OS
- `efs` is enabled with Windows OS
- a Windows platform is manually entered in the manifest for a RDWS (unsupported)

Don't surface the platform field for RDWS.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2754, opened 2021-08-19
chore: allow new platforms for non-App Runner workloads

huanjani:
This autopopulates the `platform` field in workload manifests with `windows/amd64` where appropriate, and sets the `CPU` and `mem` fields to the (soon-to-be) required values. 

Informative messages are logged whenever the platform is set for something other than what is detected (e.g. arch ARM -> amd);
the `platform` field has a value/is not nil whenever it is anything other than the default, `linux/amd64`, or whenever the architecture has been redirected, as above.

If a user tries to create a Request-Driven Web Services with Windows, we throw an error at `copilot svc init`; if, for some reason, a user tries to specify a Windows platform manually in their RDWS manifest, we throw an error at `copilot svc deploy`. 


internal/pkg/cli/interfaces.go
```diff
@@ -583,7 +583,7 @@ type runningTaskSelector interface {
 
 type dockerEngine interface {
 	CheckDockerEngineRunning() error
-	RedirectPlatform(string) (*string, error)
+	RedirectPlatform(string) (string, *string, error)
```
efekarakus: 🧐 The method signature is a bit unusual, it returns both a `string` and a `*string`. 

What do you think of the following series of refactors:
1. The package `dockerengine` should provide a `Platform()` method that only returns `(string, error)`. The platform that the docker daemon is using.
   The `dockerengine` object shouldn't know about what manifest platforms are supported.

2. The `manifest` pkg should own the logic to redirect platform if it can:
   ```go
   // RedirectPlatform returns a platform that's supported for the given manifest type.
   // If the container image os and arch is already supported, it returns the input string without any modifications.
   // Otherwise, if a platform cannot be redirected then an error is returned.
   func RedirectPlatform(imgPlatform, mftType string) (string, error)
   ```

efekarakus: _A complete sidenote comment 📝_ 

We should write our method inputs and outputs irrespective of how a client might use it.
So for example, in the existing method we could do:
```go
RedirectPlatform(string) (imgPlatform, adjustedPlatform string, err error)
```
I understand that `*string` was returned because `manifest.PlatformString` is supposed to be a string pointer, but if there is more than one client for `RedirectPlatform` then the return type doesn't make sense.

Also returning a type that make sense for another pkg like `manifest.PlatformString` is a hint that this logic should live in the `manifest` pkg instead 😄

efekarakus:
Looks good! Just a refactor for where we place the platform redirection logic 🙏

internal/pkg/cli/svc_init.go
```diff
@@ -223,15 +226,18 @@ func (o *initSvcOpts) Execute() error {
...
 		return fmt.Errorf("get/redirect docker engine platform: %w", err)
 	}
-	o.platform = platform
+	o.platform = targetPlatform
 	if o.platform != nil {
-		log.Warningf(`Your architecture type is currently unsupported. Setting platform %s instead.\n`, dockerengine.PlatformString(dockerengine.OSLinux, dockerengine.ArchAMD64))
-		if o.wkldType != manifest.RequestDrivenWebServiceType {
-			log.Warning("See 'platform' field in your manifest.\n")
+		// Error out early if Windows && AppRunner/RDWS.
+		if strings.Split(aws.StringValue(targetPlatform), "/")[0] == manifest.OSWindows && o.wkldType == manifest.RequestDrivenWebServiceType {
```
iamhopaul123: can we define `targetPlatform` to it's own type
```go
type Platform string

func (p *Platform) Parse() (string, string)
```
something like this would increase both readability and reliability...

internal/pkg/cli/svc_init.go
```diff
@@ -223,15 +226,18 @@ func (o *initSvcOpts) Execute() error {
...
+			return errors.New("Windows is not supported for App Runner services")
+		}
+		if detectedPlatform != aws.StringValue(targetPlatform) {
+			log.Warningf("Your platform %s is currently unsupported. Setting %s instead.\nSee 'platform' field in your manifest.\n", detectedPlatform, aws.StringValue(o.platform))
```
iamhopaul123: Why do we not return an error if the platform is not supported 🤔

iamhopaul123 (commented):
> for some reason, a user tries to specify a Windows platform manually in their RDWS manifest, we throw an error at copilot svc deploy.

Do we implement this in this PR 🤔

huanjani:
> > for some reason, a user tries to specify a Windows platform manually in their RDWS manifest, we throw an error at copilot svc deploy.
> 
> Do we implement this in this PR 🤔

No, it was in the previous one, #2737. I just mentioned it here to explain that both situations are now accounted for.

internal/pkg/cli/svc_init.go
huanjani: We are just informing them that while their platform isn't supported, we are still building their image but with a different platform.

internal/pkg/cli/job_init.go
```diff
@@ -323,6 +323,25 @@ func (o *initJobOpts) askSchedule() error {
 	return nil
 }
 
+func (o initJobOpts) legitimizePlatform() error {
+	detectedOs, detectedArch, err := o.dockerEngine.GetPlatform()
+	if err != nil {
+		return fmt.Errorf("get docker engine platform: %w", err)
+	}
+	detectedPlatform := aws.String(manifest.PlatformString(detectedOs, detectedArch))
```
efekarakus: nit: What do you think of keeping`manifest.PlatformString` private, and instead here we can call `dockerengine.PlatformString` instead since the os and arch detected come from docker?

internal/pkg/cli/job_init.go
```diff
@@ -323,6 +323,25 @@ func (o *initJobOpts) askSchedule() error {
...
+	platform, err := manifest.RedirectPlatform(o.image, detectedOs, detectedArch, o.wkldType)
```
efekarakus: It's a bit odd that `RedirectPlatform` takes an image as input to only shortcircuit. Can we remove `image` as a param from `RedirectPlatform` and add it as the first if-check in `legitimizePlatform` to shortcircuit?

internal/pkg/manifest/rd_web_svc.go
```diff
@@ -134,6 +134,7 @@ func (s *RequestDrivenWebService) windowsCompatibility() error {
 	if s.InstanceConfig.Platform == nil {
 		return nil
 	}
+	// Error out if user added Windows as platform in manifest.
```
efekarakus, line 137: For a follow-up PR: we should also error out I think if the arch specific is ARM for RDWS

internal/pkg/manifest/workload.go
```diff
@@ -729,6 +729,34 @@ func (p *PlatformArgs) bothSpecified() bool {
 	return (p.OSFamily != nil) && (p.Arch != nil)
 }
 
+// PlatformString returns a specified of the format <os>/<arch>.
+func PlatformString(os, arch string) string {
+	return fmt.Sprintf("%s/%s", os, arch)
+}
+
+// RedirectPlatform returns a platform that's supported for the given manifest type.
+func RedirectPlatform(image, os, arch, wlType string) (platform *string, err error) {
```
efekarakus: Can we return a string instead of a pointer? the caller can then decide if they need a pointer
```go
p, _ := RedirectPlatform(os, arch, t)
o.platform = &p
```

internal/pkg/manifest/workload.go
```diff
@@ -729,6 +729,34 @@ func (p *PlatformArgs) bothSpecified() bool {
...
+		return nil, nil
+	}
+	// Return an error if a platform cannot be redirected.
+	if wlType == RequestDrivenWebServiceType && os == OSWindows {
+		return nil, errors.New("Windows is not supported for App Runner services")
+	}
+	// All architectures must be 'amd64' (the only one currently supported); leave OS as is.
+	if arch != ArchAMD64 {
+		arch = ArchAMD64
+	}
```
efekarakus: nit:
```suggestion
	return platformString(os, ArchAMD64), nil
```

internal/pkg/manifest/rd_web_svc.go
huanjani: Did it here.

internal/pkg/manifest/workload.go
huanjani: Yes, sorry-- missed this in the last pass! Done now.

efekarakus (approved):

internal/pkg/cli/svc_init_test.go
```diff
@@ -529,7 +529,7 @@ func TestSvcInitOpts_Execute(t *testing.T) {
 						Name:           "frontend",
 						Type:           "Load Balanced Web Service",
 						DockerfilePath: "./Dockerfile",
-						Platform:       nil,
+						Platform:       &manifest.PlatformArgsOrString{PlatformString: nil},
```
Lou1415926, line 534: nit:
```suggestion
						Platform:       &manifest.PlatformArgsOrString{},
```

internal/pkg/manifest/backend_svc.go
```diff
@@ -53,6 +54,13 @@ func NewBackendService(props BackendServiceProps) *BackendService {
 	svc.BackendServiceConfig.ImageConfig.Build.BuildArgs.Dockerfile = stringP(props.Dockerfile)
 	svc.BackendServiceConfig.ImageConfig.Port = uint16P(props.Port)
 	svc.BackendServiceConfig.ImageConfig.HealthCheck = props.HealthCheck
+	if props.Platform != nil {
+		svc.BackendServiceConfig.Platform = props.Platform
+		if isWindowsPlatform(props.Platform) {
+			svc.BackendServiceConfig.TaskConfig.CPU = aws.Int(1024)
```
Lou1415926: We could probably do
```
var (
    windowsTaskCPU = 1024
    windowsTaskMemory = 2048
)
```
and use these variables instead (as well as for other workloads)

internal/pkg/manifest/job.go
```diff
@@ -101,6 +102,13 @@ func NewScheduledJob(props *ScheduledJobProps) *ScheduledJob {
 	job.ImageConfig.Build.BuildArgs.Dockerfile = stringP(props.Dockerfile)
 	job.ImageConfig.Location = stringP(props.Image)
 	job.ImageConfig.HealthCheck = props.HealthCheck
+	if props.Platform != nil {
+		job.Platform = props.Platform
+		if isWindowsPlatform(props.Platform) {
+			job.TaskConfig.CPU = aws.Int(1024)
```
Lou1415926: What do you think of defining a function for `TaskConfig`:
```
func (t *TaskConfig) useWindowsConfiguration() {
     t.CPU = aws.Int(windowsTaskCPU)
     t.Memory = aws.Int(windowsTaskMemory)
}
```
and then 
```
if isWindowsPlatform(props.Platform) {
			job.TaskConfig.useWindowsConfiguration()
}
```
If there are more windows-specific task configuration in the future (do you think there'd be any?), we only need to change the code at one place in `useWindowsConfiguration()`

internal/pkg/manifest/workload.go
```diff
@@ -729,6 +729,26 @@ func (p *PlatformArgs) bothSpecified() bool {
...
+
+// RedirectPlatform returns a platform that's supported for the given manifest type.
+func RedirectPlatform(os, arch, wlType string) (platform string, err error) {
+	// Return nil if passed the default platform.
+	if platformString(os, arch) == platformString(OSLinux, ArchAMD64) {
+		return "", nil
+	}
+	// Return an error if a platform cannot be redirected.
+	if wlType == RequestDrivenWebServiceType && os == OSWindows {
+		return "", errors.New("Windows is not supported for App Runner services")
```
Lou1415926: what do you think of promoting the error to a variable:
```
var errAppRunnerInvalidPlatformWindows = errors.New("Windows is not supported for App Runner services")
```
and use if here as well as in the `windowsCompatibility()` function for app runner?

internal/pkg/manifest/workload.go
```diff
@@ -729,6 +729,26 @@ func (p *PlatformArgs) bothSpecified() bool {
...
 
+// platformString returns a specified of the format <os>/<arch>.
+func platformString(os, arch string) string {
+	return fmt.Sprintf("%s/%s", os, arch)
+}
+
+// RedirectPlatform returns a platform that's supported for the given manifest type.
+func RedirectPlatform(os, arch, wlType string) (platform string, err error) {
+	// Return nil if passed the default platform.
+	if platformString(os, arch) == platformString(OSLinux, ArchAMD64) {
```
Lou1415926: Do you think it'd be worthy to do
```
var defaultPlatform = platformString(OSLinux, ArchAMD64)
```
and then
```
if platformString(os, arch) == defaultPlatform
```
?

internal/pkg/cli/job_init.go
```diff
@@ -323,6 +323,32 @@ func (o *initJobOpts) askSchedule() error {
...
+	}
+	o.platform = &platform
+	if platform == "" {
+		o.platform = nil
+	}
+	if o.platform != nil {
+		if o.platform != detectedPlatform {
+			log.Warningf("Your platform %s is currently unsupported. Setting %s instead.\nSee 'platform' field in your manifest.\n", detectedPlatform, aws.StringValue(o.platform))
+		}
+	}
```
Lou1415926: nit: do you think these can be refactored as:
```suggestion
        if platform ==  "" {
              // Default platform was detected.  No redirection was performed.
              return nil
        }
	if platform != detectedPlatform {
		log.Warningf("Your platform %s is currently unsupported. Setting %s instead.\nSee 'platform' field in your manifest.\n", detectedPlatform, platform)
		o.platform = &platform
	}
```

internal/pkg/cli/job_init.go
```diff
@@ -323,6 +323,32 @@ func (o *initJobOpts) askSchedule() error {
 	return nil
 }
 
+func (o *initJobOpts) legitimizePlatform() error {
```
Lou1415926, line 325: We don't necessarily need to address this in this PR. If this logic is the same for both job and service, what do you think of this function signature
```
func legitimizePlatform(image string, dockerEngine dockerEngine) (platform *string, error)
```
and call it as
```
platform, err := legitimizePlatform(o.image, o.dockerEngine)
if err != nil {
   return err
}
o.platform = platform
```
?

internal/pkg/cli/svc_init_test.go
Lou1415926: Do you think it's worthy to also add a test that checks whether the `platform` is redirected with 
```
mockDockerEngine: func(m *mocks.MockdockerEngine) {
				m.EXPECT().GetPlatform().Return("windows", "x86", nil)
			},
```
?

internal/pkg/template/templates/workloads/services/worker/manifest.yml
```diff
@@ -28,7 +28,7 @@ image:
 cpu: {{.CPU}}       # Number of CPU units for the task.
 memory: {{.Memory}}    # Amount of memory in MiB used by the task.
 {{- if .Platform}}
-platform: {{.Platform}}
+platform: {{.Platform.PlatformString}}
```
Lou1415926: https://github.com/aws/copilot-cli/blob/0d09b8ab883d3800f870e17194cf670dfc860595/internal/pkg/cli/job_init.go#L202

Let me know if I'm missing something - seems like `.Platform` ever `nil`, is that correct?

Lou1415926 (commented):
LGTM! I have just a few nits - mostly on reusing variables, and some questions!

internal/pkg/cli/svc_init_test.go
huanjani: Sure! You mean here in `svc_init_test` or in `manifest/workload_test`?

internal/pkg/manifest/job.go
huanjani: That's a good idea. Let's follow the "rule of three" and refactor with that if there's another task config field in the future (though I don't think there will be).

internal/pkg/manifest/workload.go
huanjani: Yeah, maybe. It's kind of 'six of one, half dozen of the other'. 
I like that with a var it's cleaner, but I also like knowing at a glance the details of what's happening.
I'll change it-- seems usable in multiple places.

internal/pkg/cli/svc_init_test.go
Lou1415926: I think here in svc_init_test is good if we wanted to add it!

internal/pkg/cli/job_init.go
huanjani: I like that. I need to a do a larger refactor that eliminates a lot of the redundancy among the workloads, so I'll include this in that refactor!

internal/pkg/template/templates/workloads/services/worker/manifest.yml
huanjani: Fixed.

internal/pkg/cli/job_init.go
```diff
@@ -323,6 +323,30 @@ func (o *initJobOpts) askSchedule() error {
...
+	// If the user passes in an image, their docker engine isn't necessarily running, and we can't do anything with the platform because we're not building the Docker image.
+	if o.image != "" {
+		return nil
+	}
+	detectedOs, detectedArch, err := o.dockerEngine.GetPlatform()
+	if err != nil {
+		return fmt.Errorf("get docker engine platform: %w", err)
+	}
+	detectedPlatform := aws.String(dockerengine.PlatformString(detectedOs, detectedArch))
```
Lou1415926:
```suggestion
	detectedPlatform := dockerengine.PlatformString(detectedOs, detectedArch)
```

internal/pkg/cli/job_init.go
```diff
@@ -323,6 +323,30 @@ func (o *initJobOpts) askSchedule() error {
...
+	platform, err := manifest.RedirectPlatform(detectedOs, detectedArch, o.wkldType)
+	if err != nil {
+		return fmt.Errorf("redirect docker engine platform: %w", err)
+	}
+	if platform == "" {
+		return nil
+	}
+	if platform != aws.StringValue(detectedPlatform) {
```
Lou1415926:
```suggestion
	if platform != detectedPlatform {
```

internal/pkg/cli/svc_init.go
```diff
@@ -419,6 +413,30 @@ func (o *initSvcOpts) askSvcPort() (err error) {
...
+func (o *initSvcOpts) legitimizePlatform() error {
+	// If the user passes in an image, their docker engine isn't necessarily running, and we can't do anything with the platform because we're not building the Docker image.
+	if o.image != "" {
+		return nil
+	}
+	detectedOs, detectedArch, err := o.dockerEngine.GetPlatform()
+	if err != nil {
+		return fmt.Errorf("get docker engine platform: %w", err)
+	}
+	detectedPlatform := aws.String(dockerengine.PlatformString(detectedOs, detectedArch))
```
Lou1415926:
```suggestion
	detectedPlatform := dockerengine.PlatformString(detectedOs, detectedArch)
```

internal/pkg/cli/svc_init.go
```diff
@@ -419,6 +413,30 @@ func (o *initSvcOpts) askSvcPort() (err error) {
...
+	platform, err := manifest.RedirectPlatform(detectedOs, detectedArch, o.wkldType)
+	if err != nil {
+		return fmt.Errorf("redirect docker engine platform: %w", err)
+	}
+	if platform == "" {
+		return nil
+	}
+	if platform != aws.StringValue(detectedPlatform) {
```
Lou1415926:
```suggestion
	if platform != detectedPlatform {
```

Lou1415926 (approved):
shipitt :shipit:  Feel free to remove the label once the nits are addressed!

internal/pkg/cli/job_init.go
huanjani: 🤦🏼‍♀️

Issue #2588, opened 2021-07-09
proposal: add Task Definition Overrides


iamhopaul123:
## Background
As more customers are using Copilot, our current manifest might not meet all their needs. For example in #2101, customers want `http.credential` to be configurable so as to support private repositories. Also as stated in one of our top issues with the most thumb-ups #948, it would be very helpful if they could overwrite arbitrary fields in their task definitions.

## Example Related Issues
- #948
- #1919 
- #1783 
- #2101

## UX Design
For specific fields to override, please refer to https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_TaskDefinition.html
```yaml
taskdef_override:
  - path: ContainerDefinitions[0].Ulimits.HardLimit
    value: 1024
  - ContainerDefinitions[0].Ulimits.Name: some limit
  - ContainerDefinitions[0].Ulimits.SoftLimit: 512
  - RequiresCompatibilities: [FARGATE]
```

## Restriction
Users will not be allowed to modify the following fields in the task definition
- Family
- ContainerDefinition.Name (our main container name)

## Override Behaviors
- For *single-value* options like `bool`, `string`, or `int`, the new value replaces the old value.
- For *multi-value* options like `list` or `map`, Copilot concatenates both sets of values.

## Milestones
- [x] new `override` package (subject to future changes), which
  - [x] validate the override rule
  - [x] translates the override rule
  - [x] applies override rules to the unmarshalled CFN template
  - [x] unmarshals and marshals the CFN template correctly
- [x] read the override section in the `manifest` package
- [x] apply the template override in `stack` pkg
- [ ] add docs and integ tests

buddhike:
Hey @iamhopaul123, Thanks for the awesome summary of this feature 🙏🏾. Do you know if overrides are going to be available for sidecars as well 😊?

iamhopaul123:
Hello @buddyspike sure it will be!

efekarakus:
The `taskdef_overrides` feature is now released! https://github.com/aws/copilot-cli/releases/tag/v1.10.0 🥳

PR #2756, opened 2021-08-20
chore(override): unmarshal/marshal CFN template and parse node from rule

Lou1415926:
This PR:
1. unmarshal/marshal CFN template to/from YAML nodes.
2. parse node from rule.
3. add a unit test for the exported function `CloudformationTemplate`.

Part of #2588


internal/pkg/template/override/rule.go
```diff
@@ -60,7 +61,58 @@ func (r Rule) validate() error {
 }
 
 func (r Rule) parse() (nodeUpserter, error) {
-	return nil, nil
+	pathSegments := strings.SplitAfterN(r.Path, pathSegmentSeparator, 2)
```
iamhopaul123: Why 2 🤔

iamhopaul123 (commented):
<s>A `Rule` path example will be `ContainerDefinitions[0].Ulimits[-].HardLimit`. There could be any `pathSeparator` (which is `.`). The return value for `parse()` should be a linked list, consisting of either `mapUpsertNode` or`seqIdxUpsertNode`. We might not want to modify the input or create extra `Rule`...

I wonder if we should have a helper function for a single rule and either handle the parsing recursively or using a loop 🤔 </s>

After taking a more careful look, the overall implementation LGTM!

internal/pkg/template/override/rule.go
```diff
@@ -60,7 +62,71 @@ func (r Rule) validate() error {
 }
 
 func (r Rule) parse() (nodeUpserter, error) {
-	return nil, nil
+	return r.parseRule()
+}
+
+func (r Rule) parseRule() (nodeUpserter, error) {
```
iamhopaul123: Maybe we should start from the path. Though it is slightly more complex (same complexity), it is easier to understand?
```suggestion
func parsePath(path string) *pathSegment {
  pathSegments := strings.SplitN(r.Path, pathSegmentSeparator, 2)
  // guaranteed by `validate()`.
  current := pathSegments[0]
  if len(pathSegments) < 2 {
    return &pathSegment { current: current }
   }
  return &pathSegment {
    current: current,
    next: parsePath(pathSegments[1])
  }
}
type pathSegment {
    current string
    next *pathSegment
}
func (r Rule) parseRule() (nodeUpserter, error) {
  segment := parsePath(r.Path)
  // then construct nodes based on the parsed path.
  for segment {
    ...
    segment := segment.next
  }
}
```

internal/pkg/template/override/rule.go
```diff
@@ -60,7 +62,71 @@ func (r Rule) validate() error {
...
+	}
+	nextNode, err := subRule.parseRule()
+	if err != nil {
+		return nil, err
+	}
+	baseNode.next = nextNode
+	return currNode(indexMatch, baseNode)
+}
+
+func currNode(indexMatch string, baseNode upsertNode) (nodeUpserter, error) {
```
iamhopaul123:
```suggestion
func (u *upsertNode) newNodeUpserter(indexMatch string) (nodeUpserter, error) {
```

internal/pkg/template/override/rule.go
```diff
@@ -60,7 +62,71 @@ func (r Rule) validate() error {
...
+	// https://pkg.go.dev/regexp#Regexp.FindStringSubmatch
+	// Given that path segment is valid (after `validate()`), `subMatches` contains four elements.
+	// subMatches[0] is the whole path segment, e.g. ContainerDefinitions[0].
+	// subMatches[1:] are individual capture groups. There are 3 capture groups.
+	// subMatches[1], the first capture group is "([a-zA-Z0-9_-]+)", i.e. the `key`.
+	// subMatches[2], the second capture group is "[<index>]".
+	// subMatches[3], the third capture group us "<index>".
+	// subMatches[2] and subMatches[3] could also be empty string, depending on whether there is "[<i>]" in path segment.
+	key, indexMatch := subMatches[1], subMatches[3]
+	baseNode := upsertNode{
```
iamhopaul123, line 80: nice!

iamhopaul123:
Part of #2588

internal/pkg/template/override/rule.go
iamhopaul123: Discussed offline will keep as is.

iamhopaul123 (approved):
yay :shipit:

internal/pkg/template/override/override.go
```diff
@@ -46,12 +47,22 @@ func parseRules(rules []Rule) ([]nodeUpserter, error) {
 	return ruleNodes, nil
 }
 
-func unmarshalCFNYaml(temp []byte) (*yaml.Node, error) {
-	return nil, nil
+func unmarshalCFNYAML(temp []byte) (*yaml.Node, error) {
```
efekarakus: nit: should we rename this to `unmarshalYAML` ? and update the error (seems like it should work regardless of CFN)

internal/pkg/template/override/override.go
```diff
@@ -46,12 +47,22 @@ func parseRules(rules []Rule) ([]nodeUpserter, error) {
...
+	var node yaml.Node
+	if err := yaml.Unmarshal(temp, &node); err != nil {
+		return nil, fmt.Errorf("unmarshal CloudFormation YAML template: %w", err)
+	}
+	return &node, nil
 }
 
-func marshalCFNYaml(content *yaml.Node) ([]byte, error) {
-	return nil, nil
+func marshalCFNYAML(content *yaml.Node) ([]byte, error) {
```
efekarakus: nit: samesies here `marshalYAML`

internal/pkg/template/override/override_test.go
```diff
@@ -373,3 +447,37 @@ func Test_applyRules(t *testing.T) {
 		})
 	}
 }
+
+func Test_CloudFormationTemplate(t *testing.T) {
```
efekarakus: Now that we have this public fn test can we delete safely the private method tests?

efekarakus: I think this particular test could greatly benefit of leveraging `testdata` in Go
https://github.com/aws/copilot-cli/blob/dea26599f8ec029c3d205bc72f4ee4ab36bc1847/internal/pkg/addon/output_test.go#L129

```
  inTplFileName string
  inRules []Rule
  
  wantedTplFileName string
}
```
Where we can one by one test different rules like `ulimitsRule` and so on

internal/pkg/template/override/rule.go
```diff
@@ -60,7 +62,46 @@ func (r Rule) validate() error {
 }
 
 func (r Rule) parse() (nodeUpserter, error) {
-	return nil, nil
+	return r.parseRule()
```
efekarakus: Do we need a `parseRule` method if it's delegated immediately or can we move all of its functionality here?

internal/pkg/template/override/rule.go
```diff
@@ -60,7 +62,46 @@ func (r Rule) validate() error {
...
+}
+
+func (r Rule) parseRule() (nodeUpserter, error) {
+	pathSegments := strings.SplitAfterN(r.Path, pathSegmentSeparator, 2)
+	currPathSegment := strings.TrimSuffix(pathSegments[0], ".")
```
efekarakus:
```suggestion
	currPathSegment := strings.TrimSuffix(pathSegments[0], pathSegmentSeparator)
```

internal/pkg/template/override/rule.go
```diff
@@ -60,7 +62,46 @@ func (r Rule) validate() error {
...
+		// This is the last segment.
+		baseNode.valueToInsert = &r.Value
+		return baseNode.newNodeUpserter(indexMatch)
+	}
+
+	subRule := Rule{
+		Path:  pathSegments[1],
+		Value: r.Value,
+	}
+	nextNode, err := subRule.parseRule()
```
efekarakus: recursion! 🎉

internal/pkg/template/override/rule.go
```diff
@@ -71,11 +112,36 @@ type upsertNode struct {
...
+	if indexMatch == seqAppendToLastSymbol {
+		return &seqIdxUpsertNode{
+			appendToLast: true,
+			upsertNode:   *m,
+		}, nil
+	}
+	index, err := strconv.Atoi(indexMatch)
+	if err != nil {
+		// This error also shouldn't occur given that `validate()` has passed.
+		return nil, fmt.Errorf("convert string %s to integer: %w", indexMatch, err)
```
efekarakus: I think it would be helpful to see the entire key here (to understand what is passed down). 
I know it's not ideal but I think we should return "covert index to integer" error here and wrap with the key in the caller

internal/pkg/template/override/rule.go
```diff
@@ -71,11 +112,36 @@ type upsertNode struct {
 	next          nodeUpserter
 }
 
-// NextNode returns the next node.
+// Next returns the next node.
 func (m *upsertNode) Next() nodeUpserter {
 	return m.next
 }
 
+func (m *upsertNode) newNodeUpserter(indexMatch string) (nodeUpserter, error) {
```
efekarakus: nit: What do you think of making this a function? 
```go
func newNodeUpserter(node yaml.Node, indexMatch string) (nodeUpserter, error)
```

efekarakus (commented):
Looks great 🎉 

Mostly a question around cleaning up the unit tests and small refactors

internal/pkg/template/override/rule.go
Lou1415926: Used `SplitN` instead of `SplitAfterN` so that we don't need to `TrimSuffix`!

internal/pkg/template/override/override_test.go
Lou1415926: Done!

internal/pkg/template/override/override.go
```diff
@@ -46,12 +47,22 @@ func parseRules(rules []Rule) ([]nodeUpserter, error) {
 	return ruleNodes, nil
 }
 
-func unmarshalCFNYaml(temp []byte) (*yaml.Node, error) {
-	return nil, nil
+func unmarshalYAML(temp []byte) (*yaml.Node, error) {
+	var node yaml.Node
+	if err := yaml.Unmarshal(temp, &node); err != nil {
+		return nil, fmt.Errorf("unmarshal CloudFormation YAML template: %w", err)
```
efekarakus: nit: can we update theese error messages to not mention CloudFormation?

internal/pkg/template/override/override_test.go
```diff
@@ -5,370 +5,193 @@ package override
...
-			next: node2,
-		},
-		index: 0,
-	}
-	return newTaskDefPropertyNode(node1)
+	})
 }
 
-func Test_applyRules(t *testing.T) {
+func Test_CloudFormationTemplate(t *testing.T) {
```
efekarakus, line 114: Woohoo looks great! 🤩

internal/pkg/template/override/rule.go
```diff
@@ -60,7 +62,66 @@ func (r Rule) validate() error {
...
+	}
+	// https://pkg.go.dev/regexp#Regexp.FindStringSubmatch
+	// Given that path segment is valid (after `validate()`), `subMatches` contains four elements.
+	// subMatches[0] is the whole path segment, e.g. ContainerDefinitions[0].
+	// subMatches[1:] are individual capture groups. There are 3 capture groups.
+	// subMatches[1], the first capture group is "([a-zA-Z0-9_-]+)", i.e. the `key`.
+	// subMatches[2], the second capture group is "[<index>]".
+	// subMatches[3], the third capture group us "<index>".
+	// subMatches[2] and subMatches[3] could also be empty string, depending on whether there is "[<i>]" in path segment.
+	key := subMatches[1]
```
efekarakus, line 79: nit: in a follow-up PR i think it'd be cool to actually move this info to struct so that we don't need the comments for example:
```go
type pathSegment struct {
   raw string
   
   // details on the segment.
   key string
   ref string
   index string
}

func parsePathSegment(raw string) (pathSegment, error)
```

efekarakus (approved):
Looks great! Is it possible to remove "CloudFormation" from the error messages and then :shipit: ?

internal/pkg/template/override/rule.go
Lou1415926: will follow up!

PR #2762, opened 2021-08-24
chore(stack): add rule prefix and override integ test

iamhopaul123:
Part of #2588. Work on the basis of #2756.


internal/pkg/deploy/cloudformation/stack/stack_local_integration_test.go
```diff
@@ -26,7 +25,12 @@ const (
 )
 
 func Test_Autoscaling_Integration(t *testing.T) {
```
Lou1415926: I wonder if we need a separate `Test_TaskDefinitionOverride_Integration` instead of adding it to this autoscaling integ test, just for clarity?

Lou1415926 (commented):
LGTM! One question ⬇️ !

internal/pkg/deploy/cloudformation/stack/stack_local_integration_test.go
iamhopaul123: Oooh sorry for the confusion. I should rename this function because it is not only for auto scaling validation now.

Lou1415926 (approved):

internal/pkg/deploy/cloudformation/stack/stack_local_integration_test.go
```diff
@@ -44,7 +48,15 @@ func Test_Autoscaling_Integration(t *testing.T) {
 	require.NoError(t, err)
 
 	t.Run("CloudFormation template must contain autoscaling resources", func(t *testing.T) {
-		path := filepath.Join("testdata", "autoscaling", wantedCFNTemplatePath)
+		path := filepath.Join("testdata", "stacklocal", wantedAutoScalingCFNTemplatePath)
+		wantedCFNBytes, err := ioutil.ReadFile(path)
+		require.NoError(t, err)
+		fmt.Println(tpl)
```
huanjani: just checking that this line is intentional

internal/pkg/template/override/rule.go
```diff
@@ -13,9 +13,10 @@ import (
 )
 
 const (
+	// PathSegmentSeparator is the separateor for path segments.
```
huanjani: nit: 
```suggestion
	// PathSegmentSeparator is the separator for path segments.
```

internal/pkg/deploy/cloudformation/stack/stack_local_integration_test.go
iamhopaul123: 🤦

huanjani (approved):
LGTM!

PR #2763, opened 2021-08-24
docs: add task def override

iamhopaul123:
Part of #2588.


site/content/docs/developing/overrides.en.md
```diff
@@ -0,0 +1,50 @@
+# Overrides
+If there are fields that are not configurable in the [manifest](../manifest/overview.en.md), users can bypass some of them by applying override rules to the CloudFormation template Copilot generates out of the manifest.
```
Lou1415926: We can probably mention that right now only task definition override is supported. Otherwise the user would need to infer this knowledge from the fact that there is only `taskdef_overrides` field in manifest.

site/content/docs/developing/overrides.en.md
```diff
@@ -0,0 +1,50 @@
...
+
+## How to specify override rules?
+For each override rule, users need to construct the **path** and **value** of the CloudFormation resource field they want to override.
+
+``` yaml
+<override section>:
```
Lou1415926: If we only have `taskdef_overrides` and don't have specific plan in expanding this to other resources right now, I think we can just say `taskdef_overrides` instead of `<override section>`

site/content/docs/developing/overrides.en.md
```diff
@@ -0,0 +1,50 @@
...
+    value: <value>
+```
+
+## Override Behaviors
+
+- Copilot recursively creates `map` fields if they don't exist in the **path**. For example: if `B` and `C` don't exist, `A.B.C` will create `B` and `C`
+
+- Use `-` as index to append a new member to a `list` field. The field will be initiated with this new member if it doesn't exist
+
+- **value** must be scalar value like `bool`, `string`, or `int`, and the new value replaces the old value if exists
```
huanjani: nit: 
```suggestion
- **value** must be a scalar value like `bool`, `string`, or `int`, and the new value replaces an existing value
```

site/content/docs/developing/overrides.en.md
```diff
@@ -0,0 +1,50 @@
+# Overrides
+If there are fields that are not configurable in the [manifest](../manifest/overview.en.md), users can bypass some of them by applying override rules to the CloudFormation template Copilot generates out of the manifest.
+
+## How to specify override rules?
+For each override rule, users need to construct the **path** and **value** of the CloudFormation resource field they want to override.
+
+``` yaml
+<override section>:
+  - path: <CFN field path>
```
Lou1415926: We can probably add some description of what exactly is a CFN field path, accompanied by some examples. Otherwise the user would still need to infer this from the "Task Definition Override" example below. 

For example, how deep a path need to be (e.g. can I just write ContainerDefinitions[0]: blahblah? -- which I know we can't), how to write a path that contains list parts? etc.

site/content/docs/developing/overrides.en.md
```diff
@@ -0,0 +1,50 @@
...
+taskdef_overrides:
+  - path: "ContainerDefinitions[0].Ulimits[-].Name"
+    value: "cpu"
+  - path: "ContainerDefinitions[0].Ulimits[0].SoftLimit"
+    value: 1024
+  - path: "ContainerDefinitions[0].Ulimits[0].HardLimit"
+    value: 2048
+```
+
+Below is exposing an extra UDP port for the main container.
```
huanjani:
```suggestion
Below is an example of exposing an extra UDP port for the main container.
```

site/content/docs/developing/overrides.en.md
```diff
@@ -0,0 +1,50 @@
...
+
+``` yaml
+<override section>:
+  - path: <CFN field path>
+    value: <value>
+```
+
+## Override Behaviors
+
+- Copilot recursively creates `map` fields if they don't exist in the **path**. For example: if `B` and `C` don't exist, `A.B.C` will create `B` and `C`
```
Lou1415926: Can it be "Copilot recursively creates fields....", removing the word `map`? I'm not sure if this is a universal terminology used across CloudFormation templates.

site/content/docs/developing/overrides.en.md
```diff
@@ -0,0 +1,50 @@
...
+
```
Lou1415926: Also, maybe we can include a general description first, something like "When interpreting the override rule, Copilot insert or update (upsert) the fields along the path."

And then discuss the "insert" case: Copilot recursively creates fields if they don't exist...

site/content/docs/developing/overrides.en.md
```diff
@@ -0,0 +1,50 @@
...
+- Use `-` as index to append a new member to a `list` field. The field will be initiated with this new member if it doesn't exist
```
Lou1415926:
```suggestion
- Use `-` as index to append a new member to the end of a `list` field.
```

I'm also not sure what does "The field will be initiated with this new member if it doesn't exist" mean 🤔

site/content/docs/developing/overrides.en.md
Lou1415926: If `bool`, `string` and `int` are the only acceptable scalar value, we should say so: "**value** must be scalar value: `bool`, `string`, or `int`"

site/content/docs/developing/overrides.en.md
```diff
@@ -0,0 +1,50 @@
...
+
+- **value** must be scalar value like `bool`, `string`, or `int`, and the new value replaces the old value if exists
+
+## Examples
+
+### Task Definition Override
+!!! Attention
+    Users are not allowed to modify the following fields in the task definition.
+
+    * Family
```
Lou1415926:
```suggestion
    * [Family](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-family)
```

site/content/docs/developing/overrides.en.md
```diff
@@ -0,0 +1,50 @@
...
+    * ContainerDefinitions[<index>].Name (name of any existing container)
```
Lou1415926:
```suggestion
    * [ContainerDefinitions[<index>].Name](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdefinitions.html#cfn-ecs-taskdefinition-containerdefinition-name)
```

I wonder if instead of "(name of any existing container)", we should just say they can't change the name of any container definitions created by Copilot.

site/content/docs/developing/overrides.en.md
```diff
@@ -0,0 +1,50 @@
...
+    value: 2048
+```
+
+Below is exposing an extra UDP port for the main container.
+
+``` yaml
+taskdef_overrides:
+  - path: "ContainerDefinitions[0].PortMappings[-].ContainerPort"
+    value: 2056
+  - path: "ContainerDefinitions[0].PortMappings[1].Protocol"
```
Lou1415926: We should explain how `PortMappins[1]` is the port mappings added by `PortMappins[-]` and how the order of the rules matters.

site/content/docs/developing/overrides.en.md
```diff
@@ -0,0 +1,50 @@
+# Overrides
```
efekarakus: Can we rename the file and this header to `Task Definition Overrides`? 🙏

site/content/docs/developing/overrides.en.md
efekarakus: Agreed I think we should be specific to only task def overrides in this doc. Eventually we can create more pages if we have other types of overrides.

iamhopaul123: Thought about that too. Problem is that once we add more sections I don't want to add them accordingly. Maybe it's ok for them to figure out what's available? I'm not sure doc writing is so hard

site/content/docs/developing/overrides.en.md
iamhopaul123: But then wouldn't they assume override = taskdef_override now and in future?

site/content/docs/developing/overrides.en.md
efekarakus: Is it strictly scalars, will this blow up? 
```
- path: ContainerDefinitions[0].Ulimits[-]
   value: 
      Name: "cpu"
      SoftLimit: 1024
      HardLimit: 2048
```

site/content/docs/developing/overrides.en.md
iamhopaul123: > Otherwise the user would still need to infer this from the "Task Definition Override" example below.

Hmm this is exactly why we have `example` section right 🤔  I was referring to CDK https://docs.aws.amazon.com/cdk/latest/guide/cfn_layer.html#cfn_layer_raw and for those who want to use this feature it might be safe for us to assume they have some CFN background.

site/content/docs/developing/overrides.en.md
iamhopaul123: that's a good idea!

site/content/docs/developing/overrides.en.md
iamhopaul123: > Agreed I think we should be specific to only task def overrides in this doc. Eventually we can create more pages if we have other types of overrides.

But then they'll share the same overriding pattern. should we concentrate them into one `override` page?

site/content/docs/developing/overrides.en.md
iamhopaul123: hmm actually this will not blow up

site/content/docs/developing/overrides.en.md
iamhopaul123: ok discussed offline we'll start from very specific page then in future we might move them together.

site/content/docs/developing/taskdef-overrides.en.md
```diff
@@ -0,0 +1,55 @@
+# Task Definition Overrides
+If there are fields that are not configurable in the [manifest](../manifest/overview.en.md), users can bypass some of the [ECS Task Definition setting](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html) by applying override rules to the CloudFormation template Copilot generates out of the manifest.
```
Lou1415926: nit:
```suggestion
Copilot generates CloudFormation templates using configuration specified in the [manifest](../manifest/overview.en.md). However, there are fields that are not configurable in the [manifest](../manifest/overview.en.md). For example, You might want to configure the [`Ulimits`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdefinitions.html#cfn-ecs-taskdefinition-containerdefinition-ulimits) of your workload container, but this is not supported in our manifest.

You can configure [ECS Task Definition settings](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html) like these by specifying override rules. These override rules will be applied to the CloudFormation template that Copilot generates out of the manifest.
```

The idea is that we might want to explain the use case of `taskdef_override` with more concrete details, so that people understand if this is what they are looking for just by reading the first few lines of this page.

I'm no English speaker either so what I'm suggesting here might not be ideal as well 😅

site/content/docs/developing/taskdef-overrides.en.md
```diff
@@ -0,0 +1,55 @@
...
+
+## How to specify override rules?
+For each override rule, users need to construct the **path** and **value** of the CloudFormation resource field they want to override.
```
Lou1415926: nit: 
```suggestion
For each override rule, users need to specify a **path** of the CloudFormation resource field they want to override, and a **value** of that field.
```

site/content/docs/developing/taskdef-overrides.en.md
```diff
@@ -0,0 +1,55 @@
...
+taskdef_overrides:
+  - path: <ECS Task Definition field path>
+    value: <value>
+```
+
+## Override Behaviors
+
+- Use `-` as index to append a new member to a `list` field
+
+- When applying the override rule, Copilot inserts or updates the fields along the path. More specifically, Copilot recursively inserts fields if they don't exist in the **path**. For example: if `B` and `C` don't exist, `A.B[-].C` will create `B` and `C`
```
Lou1415926:
```suggestion
- When applying an override rule, Copilot inserts or updates the fields along the path. More specifically, Copilot recursively inserts fields if they don't exist in the original CloudFormation template. For example: if a rule has the path `A.B[-].C`, Copilot will insert the field `B` and `C`. A concrete example can be found [below](<an anchor link to an example below>).
```

site/content/docs/developing/taskdef-overrides.en.md
```diff
@@ -0,0 +1,55 @@
...
+
+``` yaml
+taskdef_overrides:
+  - path: ContainerDefinitions[0].Ulimits[-]
+    value:
+      Name: "cpu"
+      SoftLimit: 1024
+      HardLimit: 2048
+```
+
```
Lou1415926, line 57: We might want to explain a bit how this works, for example
```
What Copilot does:
1. In the CloudFormation template that Copilot generates, the first `ContainerDefinition` is the main container for your workload. Using this override rule, Copilot looks for the first  `ContainerDefinition` (i.e. the main container definition). 
2. Since the `Ulimits` field doesn't exist in the original template, Copilot inserts a `Ulimits` field.
3. To add a first element to the `Ulimits` field, use `[-]` to signal an "append" action. You can think of this as appending to an empty list. 
```
and samesies for the other examples.

mkdocs.yml
```diff
@@ -47,6 +47,7 @@ nav:
       - Custom Environment Resources: docs/developing/custom-environment-resources.en.md
       - Secrets: docs/developing/secrets.en.md
       - Service Discovery: docs/developing/service-discovery.en.md
+      - Task Definition Overrides: docs/developing/taskdef-overrides.en.md
```
efekarakus: let's move this to last in the list as it's an advanced feature 🙏

site/content/docs/developing/taskdef-overrides.en.md
```diff
@@ -0,0 +1,57 @@
+# Task Definition Overrides
+Copilot generates CloudFormation templates using configuration specified in the [manifest](../manifest/overview.en.md). However, there are fields that are not configurable in the manifest. For example, You might want to configure the [`Ulimits`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdefinitions.html#cfn-ecs-taskdefinition-containerdefinition-ulimits) for your workload container, but it is not exposed in our manifest.
```
efekarakus, line 6: 👏 awesome description

site/content/docs/developing/taskdef-overrides.en.md
```diff
@@ -0,0 +1,57 @@
...
+
+You can configure [ECS Task Definition settings](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html) by specifying override rules, which will be applied to the CloudFormation template that Copilot generates out of the manifest.
```
efekarakus:
```suggestion
You can configure additional [ECS Task Definition settings](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html) by specifying `taskdef_overrides` rules, which will be applied to the CloudFormation template that Copilot generates out of the manifest.
```

site/content/docs/developing/taskdef-overrides.en.md
```diff
@@ -0,0 +1,57 @@
...
+
+``` yaml
+taskdef_overrides:
+  - path: <ECS Task Definition field path>
+    value: <value>
+```
+
+## Override Behaviors
+
+- Use `-` as index to append a new member to a `list` field
```
efekarakus: Can we make this the last item in the list?

site/content/docs/developing/taskdef-overrides.en.md
```diff
@@ -0,0 +1,57 @@
...
+You can configure [ECS Task Definition settings](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html) by specifying override rules, which will be applied to the CloudFormation template that Copilot generates out of the manifest.
+
+## How to specify override rules?
+For each override rule, you need to specify a **path** of the CloudFormation resource field you want to override, and a **value** of that field.
+
+``` yaml
+taskdef_overrides:
+  - path: <ECS Task Definition field path>
+    value: <value>
+```
```
efekarakus: What do you think of this? (inspired by https://datatracker.ietf.org/doc/html/rfc6902)

```
The following is an example valid `taskdef_overrides` field that can be applied to a manifest file:

``` yaml
taskdef_overrides:
- path: ContainerDefinitions[0].Cpu
  value: 512
- path: ContainerDefinitions[0].Memory
   value: 1024
```

Each rule is applied sequentially to the CloudFormation template. The resulting CloudFormation template becomes the target of the next rule. Evaluation continues until all rules are successfully applied or an error is encountered.
```

File: site/content/docs/developing/taskdef-overrides.en.md [hunk 22]
```diff
@@ -0,0 +1,57 @@
...
+
+## Override Behaviors
```
efekarakus:
```suggestion
### Path evaluation

The `path` field is a `'.'` character separated path to a target [Task Definition field under `Properties` in CloudFormation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html).
- Copilot recursively inserts fields if they don't ....
- To append a new member to a `list` field such as `Ulimits` you can use the special character `-`: `Ulimits[-]`.
```

site/content/docs/developing/taskdef-overrides.en.md
```diff
@@ -0,0 +1,57 @@
...
+
+- Use `-` as index to append a new member to a `list` field
+
+- When applying an override rule, Copilot inserts or updates the fields along the path. More specifically, Copilot recursively inserts fields if they don't exist in the CloudFormation template. For example: if a rule has the path `A.B[-].C` (`B` and `C` don't exist), Copilot will insert the field `B` and `C`. A concrete example can be found [below](#add-ulimits-to-the-main-container).
```
efekarakus: What happens if the field exists? Can we say that it's going to replace it. 
```go
- If the target path specifies an member that already exists, that member's value is replaced.
```

site/content/docs/developing/taskdef-overrides.en.md
```diff
@@ -0,0 +1,57 @@
...
+
+!!! Attention
+    The following fields in the task definition are not allowed to be modified.
+
+    * [Family](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-family)
+    * [ContainerDefinitions[<index>].Name](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdefinitions.html#cfn-ecs-taskdefinition-containerdefinition-name)
+
```
efekarakus, line 40: Should we add this?
```
### Testing

In order to ensure that your override rules will behave as expected, we recommend running `copilot svc package` or `copilot job package` to preview the generated CloudFormation template. 
```

site/content/docs/developing/taskdef-overrides.en.md
```diff
@@ -0,0 +1,57 @@
...
+## Examples
```
efekarakus, line 45: Nice examples!

efekarakus (commented):
I really like this new version! Some more small tweaks and we can merge it :)

site/content/docs/developing/taskdef-overrides.en.md
```diff
@@ -0,0 +1,75 @@
...
+
+Each rule is applied sequentially to the CloudFormation template. The resulting CloudFormation template becomes the target of the next rule. Evaluation continues until all rules are successfully applied or an error is encountered.
+
+## Path Evaluation
+
+- The `path` field is a `'.'` character separated path to a target [Task Definition field under `Properties` in CloudFormation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html).
+
+- Copilot recursively inserts fields if they don't exist in the CloudFormation template. For example: if a rule has the path `A.B[-].C` (`B` and `C` don't exist), Copilot will insert the field `B` and `C`. A concrete example can be found [below](#add-ulimits-to-the-main-container).
+
+- If the target path specifies an member that already exists, that member's value is replaced.
```
efekarakus: nit:
```suggestion
- If the target path specifies a member that already exists, that member's value is replaced.
```

site/content/docs/include/common-svc-fields.en.md
```diff
@@ -240,5 +240,16 @@ Optional. The full config file path in your custom Fluent Bit image.
 
 <div class="separator"></div>
 
+<a id="taskdef_overrides" href="#taskdef_overrides" class="field">`taskdef_overrides`</a> <span class="type">Array</span>
```
efekarakus: nit: maybe
```suggestion
<a id="taskdef_overrides" href="#taskdef_overrides" class="field">`taskdef_overrides`</a> <span class="type">Array of Rules</span>  
```

efekarakus (approved):
Looks good two nits!

Lou1415926 (approved):

PR #2705, opened 2021-08-05
chore: adds worker to deploy

xar-tol:
Adds worker to deploy.
Adds worker deploy tests.
Adds ListDeployedSNSTopics to mocks.

Addresses #2550

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2765, opened 2021-08-24
chore: add worker service to svc deploy

bvtujo:
Continued from #2705. 


internal/pkg/cli/svc_deploy.go
```diff
@@ -551,6 +560,21 @@ func (o *deploySvcOpts) stackConfiguration(addonsURL string) (cloudformation.Sta
 		conf, err = stack.NewRequestDrivenWebServiceWithAlias(t, o.targetEnvironment.Name, appInfo, *rc, urls)
 	case *manifest.BackendService:
 		conf, err = stack.NewBackendService(t, o.targetEnvironment.Name, o.targetEnvironment.App, *rc)
+	case *manifest.WorkerService:
+		topics, err := o.snsTopicGetter.ListSNSTopics(o.appName, o.envName)
+		if err != nil {
+			return nil, err
+		}
+		topicARNs := []string{}
```
```suggestion
		var topicARNs []string
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -551,6 +560,21 @@ func (o *deploySvcOpts) stackConfiguration(addonsURL string) (cloudformation.Sta
 		conf, err = stack.NewRequestDrivenWebServiceWithAlias(t, o.targetEnvironment.Name, appInfo, *rc, urls)
 	case *manifest.BackendService:
 		conf, err = stack.NewBackendService(t, o.targetEnvironment.Name, o.targetEnvironment.App, *rc)
+	case *manifest.WorkerService:
+		var topics []deploy.Topic
+		topics, err = o.snsTopicGetter.ListSNSTopics(o.appName, o.envName)
```
efekarakus, line 597: lol I was commenting in an older revision that this was the reason why the linter was failing 👍

internal/pkg/cli/svc_deploy.go
```diff
@@ -805,6 +842,29 @@ Please visit %s to check the validation status.
 	return nil
 }
 
+// StripNonAlphaNumFunc strips non-alphanumeric characters from an input string.
+func StripNonAlphaNumFunc(s string) string {
+	return nonAlphaNum.ReplaceAllString(s, "")
+}
+
+func (o *deploySvcOpts) buildWorkerQueueNames() (string, error) {
+	topics, err := o.snsTopicGetter.ListSNSTopics(o.appName, o.envName)
```
efekarakus: should we wrap this in a method so that we can cache the list of sns topics in a field in `deploySvcOpts` (we are making the call multiple times)

internal/pkg/cli/svc_deploy.go
```diff
@@ -805,6 +842,29 @@ Please visit %s to check the validation status.
 	return nil
 }
 
+// StripNonAlphaNumFunc strips non-alphanumeric characters from an input string.
+func StripNonAlphaNumFunc(s string) string {
```
efekarakus:
```suggestion
// stripNonAlphaNum strips non-alphanumeric characters from an input string.
func stripNonAlphaNumChars(s string) string {
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -774,6 +798,19 @@ func (o *deploySvcOpts) showSvcURI() error {
...
+			return err
+		}
+		retrieveEnvVarCode := fmt.Sprintf("const {eventsQueue%s} = JSON.parse(process.env.COPILOT_QUEUE_URIS)", queueNames)
+		actionRetrieveEnvVar := fmt.Sprintf(
+			`Update %s's code to leverage the injected environment variable "COPILOT_QUEUE_URIS".
+	For example, in JavaScript you can write %s.`,
+			o.name,
+			color.HighlightCode(retrieveEnvVarCode))
+		log.Successf("Deployed %s.\n\n%s\n", color.HighlightUserInput(o.name), actionRetrieveEnvVar)
+		return nil
```
efekarakus: It's a bit odd to have this in `showSvcURI` what do you think of moving this logic to `RecommendedActions`?

internal/pkg/cli/svc_deploy.go
```diff
@@ -805,6 +842,29 @@ Please visit %s to check the validation status.
...
+func StripNonAlphaNumFunc(s string) string {
+	return nonAlphaNum.ReplaceAllString(s, "")
+}
+
+func (o *deploySvcOpts) buildWorkerQueueNames() (string, error) {
+	topics, err := o.snsTopicGetter.ListSNSTopics(o.appName, o.envName)
+	if err != nil {
+		return "", err
+	}
+	var queueNames string
```
efekarakus: can we use the following?

```go
sb := new(strings.Builder)
_, _ = sb.WriteString("eventsQueue")
   ....
   _, _ = sb.WriteString(", %s", subName)
}
return sb.String(), nil
```

internal/pkg/cli/svc_package.go
```diff
@@ -179,6 +185,23 @@ func newPackageSvcOpts(vars packageSvcVars) (*packageSvcOpts, error) {
 			if err != nil {
 				return nil, fmt.Errorf("init backend service stack serializer: %w", err)
 			}
+		case *manifest.WorkerService:
+			var topics []deploy.Topic
+			topics, err = opts.snsTopicGetter.ListSNSTopics(opts.appName, opts.envName)
+			if err != nil {
+				return nil, err
+			}
+			topicARNs := []string{}
```
efekarakus: nit: samesies here

internal/pkg/cli/validate.go
```diff
@@ -699,6 +704,35 @@ func validatePubSubName(name string) error {
 	return nil
 }
 
+func validateTopicsExist(mft interface{}, topicARNs []string, app, env string) error {
+	type subscriptions interface {
+		Subscriptions() []manifest.TopicSubscription
+	}
```
efekarakus: is it possible to move the interface outside of the function and get rid of the error?
```suggestion
type subscriberManifest interface {
	Subscriptions() []manifest.TopicSubscription
}

func validateTopicsExist(mft subscriberManifest, topicARNs []string, app, env string) error {
	
```

efekarakus (commented):
Looks good! some small refactor suggestions

internal/pkg/cli/svc_deploy.go
```diff
@@ -82,6 +82,7 @@ type deploySvcOpts struct {
 	envUpgradeCmd       actionCommand
 	newAppVersionGetter func(string) (versionGetter, error)
 	endpointGetter      endpointGetter
+	snsTopicGetter      deployedEnvironmentLister
```
efekarakus, line 85: We should add this field in `init.go` too:
https://github.com/aws/copilot-cli/blob/807cee27b66685082e29b3193f1901423e64956b/internal/pkg/cli/init.go#L156-L174

internal/pkg/cli/svc_package.go
```diff
@@ -75,6 +75,7 @@ type packageSvcOpts struct {
 	identity          identityService
 	stackSerializer   func(mft interface{}, env *config.Environment, app *config.Application, rc stack.RuntimeConfig) (stackSerializer, error)
 	newEndpointGetter func(app, env string) (endpointGetter, error)
+	snsTopicGetter    deployedEnvironmentLister
```
efekarakus, line 78: It shouldn't error out but should we also add this here:
https://github.com/aws/copilot-cli/blob/807cee27b66685082e29b3193f1901423e64956b/internal/pkg/cli/job_package.go#L89-L90

internal/pkg/cli/validate.go
```diff
@@ -699,6 +704,35 @@ func validatePubSubName(name string) error {
 	return nil
 }
 
+func validateTopicsExist(mft interface{}, topicARNs []string, app, env string) error {
```
Lou1415926: 👋🏼 parameter drilling be away!

internal/pkg/cli/svc_deploy.go
Lou1415926: Maybe we can just use `template.StripNonAlphNumChars` that's already defined?

Lou1415926 (commented):
I don't have much to add! LGTM!

internal/pkg/cli/svc_deploy.go
```diff
@@ -560,6 +569,21 @@ func (o *deploySvcOpts) stackConfiguration(addonsURL string) (cloudformation.Sta
 		conf, err = stack.NewRequestDrivenWebServiceWithAlias(t, o.targetEnvironment.Name, appInfo, *rc, urls)
 	case *manifest.BackendService:
 		conf, err = stack.NewBackendService(t, o.targetEnvironment.Name, o.targetEnvironment.App, *rc)
+	case *manifest.WorkerService:
+		var topics []deploy.Topic
+		topics, err = o.snsTopicGetter.ListSNSTopics(o.appName, o.envName)
+		if err != nil {
+			return nil, err
```
iamhopaul123: Should we wrap this error?

internal/pkg/cli/svc_deploy.go
```diff
@@ -786,6 +844,10 @@ func (o *deploySvcOpts) showSvcURI() error {
 			Svc:         o.name,
 			ConfigStore: o.store,
 		})
+	case manifest.WorkerServiceType:
+		o.generateWorkerServiceRecommendedActions()
```
efekarakus, line 848: It's still a bit strange to have `generateWorkerServiceRecommendedActions` inside `showSvcURI`. Can we rename the method to `logSuccessfulDeployment`?

internal/pkg/cli/svc_deploy.go
```diff
@@ -817,6 +879,23 @@ Please visit %s to check the validation status.
 	return nil
 }
 
+func (o *deploySvcOpts) buildWorkerQueueNames() string {
+	var queueNames string
+	for _, subscription := range o.subscriptions {
+		if subscription.Queue != nil {
```
```suggestion
		if subscription.Queue == nil {
			continue;
		}
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -843,6 +922,11 @@ func buildSvcDeployCmd() *cobra.Command {
 			if err := opts.Execute(); err != nil {
 				return err
 			}
+
+			log.Infoln("Recommended follow-up actions:")
```
efekarakus: This will be printed for all other services now right? and the recommended actions will be empty.

Can we check if  `len(opts.RecommendedActions() > 0` and if so write the log statement?

efekarakus (commented):
Looks good! minor feedback

efekarakus (approved):

internal/pkg/cli/svc_deploy.go
Lou1415926: Is it possible to call `o.generateWorkerServiceRecommendedActions` and `log.Infoln("Recommended follow-up actions:")`  from within `RecommandedActions()`? I think it'd be ideal if we can put all recommended-actions related logic inside one function such that each function is decoupled as much as possible.

Lou1415926 (approved):
LGTM! I added the label just in case you'd like to address the feedback in this PR (if it's addressable). But feel free to remove it and we can address that later!

efekarakus:
> Is it possible to call o.generateWorkerServiceRecommendedActions and log.Infoln("Recommended follow-up actions:") from within RecommandedActions()?

We have one action item to add a recommendation if there is a `publish` field in the manifest. I can pick up this refactor in the upcoming PR!

PR #2774, opened 2021-08-26
chore: add progress tracker metadata to sns resources

efekarakus:


iamhopaul123 (approved):

internal/pkg/template/workload.go
```diff
@@ -424,6 +425,7 @@ func withSvcParsingFuncs() ParseOption {
 			"jsonQueueURIs":       generateQueueURIJSON,
 			"envControllerParams": envControllerParameters,
 			"logicalIDSafe":       StripNonAlphaNumFunc,
+			"join":                strings.Join,
```
Lou1415926: what do you think of using [`english.WordSeries`](https://pkg.go.dev/glitch.sh/go/humanize/english#WordSeries) 🤔

Lou1415926 (approved):
LGTM! Put a do-not-merge label just in case you'd like to use `english.WordSeries`!

internal/pkg/template/workload.go
efekarakus: Great idea!! Updated with both `WordSeries` and `PluralWord`

PR #2778, opened 2021-08-27
fix: correctly inject environment variables for subscriptions.

bvtujo:


internal/pkg/template/templates/workloads/partials/cf/envvars.yml
```diff
@@ -23,14 +23,19 @@ Environment:
 - Name: COPILOT_SNS_TOPIC_ARNS
   Value: '{{jsonSNSTopics .Publish.Topics}}'
 {{- end}}{{- end}}
+{{- if .Subscribe}}
+- Name: COPILOT_QUEUE_URI
+  Value: !Ref EventsQueue
+{{- end}}
 {{- if .Subscribe}}{{- if .Subscribe.Topics}}
-- Name: COPILOT_QUEUE_URIS
+- Name: COPILOT_TOPIC_QUEUE_URIS
```
efekarakus: If I'm following correctly, looks like we are going to render `COPILOT_TOPIC_QUEUE_URIS={}` if there are no topic queues.
I wonder if it will be confusing to users to see both env vars by default. It'd be neat to only  surface `COPILOT_QUEUE_URI` by default. Can we do something like this? 👇 

1. We add a method to `SubscribeOpts` like this:
```go
func (s *SubscribeOpts) HasTopicQueues() bool {
   for _, t := range s.topics {
      if t.Queue != nil { return true }
   }
   return false
}
```

and in the template we can do:
```go
{{- if .Subscribe.HasTopicQueues }}
- Name: COPILOT_TOPIC_QUEUE_URIS
...
```

efekarakus (commented):
Awesome, looks great! We have to fix the manifest to mention "COPILOT_QUEUE_URI". 
We can send a separate PR for it

efekarakus:
> We have to fix the manifest to mention "COPILOT_QUEUE_URI".
> We can send a separate PR for it

https://github.com/aws/copilot-cli/pull/2779

efekarakus (approved):

iamhopaul123 (approved):

Issue #2777, opened 2021-08-26
Error on env init command when importing a vpc as a cli parameter without subnets


adamjkeller:
Tested using copilot v1.9.0.

I receive the following error when I run `copilot env init --import-vpc-id <vpc-id-here> --name test --app migration-demo`:

```
✘ at least two private subnets must be imported
```

This behavior appears to have changed from 1.7.0. When I ran the same above command with 1.7.0, it walked me through the guided experience to choose what subnets I wanted to import.

<img width="1699" alt="Screen Shot 2021-08-26 at 2 55 59 PM" src="https://user-images.githubusercontent.com/7571023/131050478-23be8691-5b2a-4f96-adc3-744fd3186f80.png">

PR #2781, opened 2021-08-27
fix(cli): command with env/svc flag should work without app flag

### LINKED ISSUE #2776: Error when running copilot env show when app name contains a hyphen
Tested with copilot v1.9.0.

I created an app named `migration-demo` and an environment named `test`. When I run `copilot env show --name test`, I see the following error:

```
✘ get environment test in application : ValidationException: Parameter name: can't be prefixed with "ssm" (case-insensitive). If formed as a path, it can consist of sub-paths divided by slash symbol; each sub-path can be formed as a mix of letters, numbers and the following 3 symbols .-_
        status code: 400, request id: d6454b99-8806-40fb-a9cb-d61be1aa2fb0
```

When I run `copilot env show`, it works and returns the details of the environment.

```
(.env) adam:~ $ copilot env show
Only found one application, defaulting to: migration-demo
Only found one environment, defaulting to: test
About

  Name              test
  Production        false
  Region            us-west-2
  Account ID        123456789

Services

  Name              Type
  ----              ----

Tags

  Key                  Value
  ---                  -----
  copilot-application  migration-demo
  copilot-environment  test
```

iamhopaul123:
fixes #2776 

### DIFF
--- a/internal/pkg/cli/job_logs.go
+++ b/internal/pkg/cli/job_logs.go
@@ -78,8 +78,10 @@ func newJobLogOpts(vars jobLogsVars) (*jobLogsOpts, error) {
 // Validate returns an error if the values provided by flags are invalid.
 func (o *jobLogsOpts) Validate() error {
 	if o.appName != "" {
-		_, err := o.configStore.GetApplication(o.appName)
-		if err != nil {
+		if _, err := o.configStore.GetApplication(o.appName); err != nil {
+			return err
+		}
+		if _, err := o.configStore.GetJob(o.appName, o.name); err != nil {
 			return err
 		}
 	}
--- a/internal/pkg/cli/job_logs.go
+++ b/internal/pkg/cli/job_logs.go
@@ -81,6 +81,11 @@ func (o *jobLogsOpts) Validate() error {
 		if _, err := o.configStore.GetApplication(o.appName); err != nil {
 			return err
 		}
+		if o.envName != "" {
+			if _, err := o.configStore.GetEnvironment(o.appName, o.envName); err != nil {
+				return err
+			}
+		}
 		if o.name != "" {
 			if _, err := o.configStore.GetJob(o.appName, o.name); err != nil {
 				return err
--- a/internal/pkg/cli/job_logs.go
+++ b/internal/pkg/cli/job_logs.go
@@ -81,8 +81,10 @@ func (o *jobLogsOpts) Validate() error {
 		if _, err := o.configStore.GetApplication(o.appName); err != nil {
 			return err
 		}
-		if _, err := o.configStore.GetJob(o.appName, o.name); err != nil {
-			return err
+		if o.name != "" {
+			if _, err := o.configStore.GetJob(o.appName, o.name); err != nil {
+				return err
+			}
 		}
 	}
 
--- a/internal/pkg/cli/secret_init.go
+++ b/internal/pkg/cli/secret_init.go
@@ -148,19 +148,18 @@ func (o *secretInitOpts) Validate() error {
 		if err != nil {
 			return fmt.Errorf("get application %s: %w", o.appName, err)
 		}
+		if o.values != nil {
+			for env := range o.values {
+				if _, err := o.targetEnv(env); err != nil {
+					return err
+				}
+			}
+		}
 	}
 
 	if o.name != "" {
 		if err := validateSecretName(o.name); err != nil {
 			return err
-		}
-	}
-
-	if o.values != nil {
-		for env := range o.values {
-			if _, err := o.targetEnv(env); err != nil {
-				return err
-			}
 		}
 	}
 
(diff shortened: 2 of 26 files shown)


internal/pkg/cli/job_logs.go
```diff
@@ -78,8 +78,10 @@ func newJobLogOpts(vars jobLogsVars) (*jobLogsOpts, error) {
...
 func (o *jobLogsOpts) Validate() error {
 	if o.appName != "" {
-		_, err := o.configStore.GetApplication(o.appName)
-		if err != nil {
+		if _, err := o.configStore.GetApplication(o.appName); err != nil {
+			return err
+		}
+		if _, err := o.configStore.GetJob(o.appName, o.name); err != nil {
 			return err
 		}
```
efekarakus: Is `o.name` always set here?

internal/pkg/cli/svc_resume.go
```diff
@@ -53,15 +53,15 @@ func (o *resumeSvcOpts) Validate() error {
 		if err := o.validateAppName(); err != nil {
 			return err
 		}
-	}
-	if o.envName != "" {
-		if err := o.validateEnvName(); err != nil {
-			return err
+		if o.envName != "" {
```
efekarakus: I just wanted to make sure that all this changes will indeed result in an error right in case the user does something like this `copilot svc resume -e invalid-env`.

As long as the `Execute` or `Ask` logic ends up returning errors after these changes then I'm fine with it!

internal/pkg/cli/svc_status.go
```diff
@@ -92,15 +92,15 @@ func (o *svcStatusOpts) Validate() error {
 		if _, err := o.store.GetApplication(o.appName); err != nil {
 			return err
 		}
-	}
-	if o.svcName != "" {
-		if _, err := o.store.GetService(o.appName, o.svcName); err != nil {
-			return err
+		if o.envName != "" {
```
efekarakus: nit: we should consider flipping the if-statement in most of these methods to:
```go
if o.appName == "" { return nil }
```

efekarakus (commented):
Looks good! just some questions

internal/pkg/cli/job_logs.go
iamhopaul123: darn good call

internal/pkg/cli/svc_resume.go
iamhopaul123: Yeah they will error out in selector I think.
```
✘ select deployed services for application demo: list deployed service for environment ts: get environment config ts: couldn't find environment ts in the application demo
```

internal/pkg/cli/svc_status.go
iamhopaul123: ok done!

efekarakus (approved):

huanjani:
This fixes #2776, not #2777, right?

internal/pkg/cli/svc_resume.go
huanjani: Hm. Doesn't `AskEnvName()` return nil if the `o.name` is set? 
Can we add the validation to `AskEnvName()` for each command?  It seems a little odd to prompt with the selector if `-n` was set at the beginning....

internal/pkg/cli/job_logs.go
```diff
@@ -78,10 +78,14 @@ func newJobLogOpts(vars jobLogsVars) (*jobLogsOpts, error) {
 // Validate returns an error if the values provided by flags are invalid.
 func (o *jobLogsOpts) Validate() error {
 	if o.appName != "" {
-		_, err := o.configStore.GetApplication(o.appName)
-		if err != nil {
+		if _, err := o.configStore.GetApplication(o.appName); err != nil {
```
Lou1415926, line 81: Do we also need to validate `envName` flag then? `appName` and `envName` are used in `initLogsSvc` which are called by `svc logs` from within `Execute()`

internal/pkg/cli/secret_init.go
```diff
@@ -148,6 +148,13 @@ func (o *secretInitOpts) Validate() error {
 		if err != nil {
 			return fmt.Errorf("get application %s: %w", o.appName, err)
 		}
+		if o.values != nil {
```
Lou1415926, line 151: :woman-facepalming:

Lou1415926: What do you think we leave this chunk at its original place and modify `o.targetEnv` instead?
```go
func (o *secretInitOpts) targetEnv(envName string) (*config.Environment, error) {
        if o.appName == "" {
                return nil, nil
        }
 	env, err := o.store.GetEnvironment(o.appName, envName)
	if err != nil {
		return nil, fmt.Errorf("get environment %s in application %s: %w", envName, o.appName, err)
	}
	return env, nil
}
```

Lou1415926 (commented):
LGTM! One small suggestion / question ⬇️

iamhopaul123:
> This fixes #2776, not #2777, right?

true true. Thank you!

internal/pkg/cli/secret_init.go
iamhopaul123: Umm I think it still prefer handling it in the validate, since it would make more sense when reading `validate`

Lou1415926 (approved):

PR #2782, opened 2021-08-27
chore: always create the events queue for Worker Services

efekarakus:

### DIFF
--- a/internal/pkg/template/templates/workloads/partials/cf/subscribe.yml
+++ b/internal/pkg/template/templates/workloads/partials/cf/subscribe.yml
@@ -1,4 +1,3 @@
-{{- if .Subscribe}}
 EventsKMSKey:
   Metadata:
     'aws:copilot:description': 'A KMS key to encrypt messages in your queues'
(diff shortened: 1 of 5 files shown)


internal/pkg/template/templates/workloads/partials/cf/subscribe.yml
```diff
@@ -108,6 +108,8 @@ DeadLetterPolicy:
           Resource: !GetAtt DeadLetterQueue.Arn
 {{- end}}{{- end}}
 
+{{- end }}{{/* endif .Subscribe */}}
```
Lou1415926, line 111: `{{/* endif .Subscribe */}}`!

Lou1415926 (approved):

internal/pkg/template/templates/workloads/partials/cf/subscribe.yml
iamhopaul123: Is this the comment saying the `end` before is for `.Subscribe`? If so can we add for this https://github.com/aws/copilot-cli/blob/54269e26b714d4bf9d8aa67dc041152df35245a8/internal/pkg/template/templates/workloads/partials/cf/subscribe.yml#L229 as well?

iamhopaul123 (approved):
LGTM just one question! Feel free to remove the tag when it is ready.

PR #2783, opened 2021-08-27
chore(cli): display recommended actions on "copilot deploy"

efekarakus:

### DIFF
--- a/internal/pkg/cli/svc_deploy.go
+++ b/internal/pkg/cli/svc_deploy.go
@@ -219,7 +219,7 @@ func (o *deploySvcOpts) generateWorkerServiceRecommendedActions() {
 	retrieveEnvVarCode := "const eventsQueueURI = process.env.COPILOT_QUEUE_URI"
 	actionRetrieveEnvVar := fmt.Sprintf(
 		`Update %s's code to leverage the injected environment variable "COPILOT_QUEUE_URI".
-In JavaScript you can write %s.`,
+  In JavaScript you can write %s.`,
 		o.name,
 		color.HighlightCode(retrieveEnvVarCode),
 	)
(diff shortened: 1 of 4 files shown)


internal/pkg/cli/svc_deploy.go
```diff
@@ -219,7 +219,7 @@ func (o *deploySvcOpts) generateWorkerServiceRecommendedActions() {
 	retrieveEnvVarCode := "const eventsQueueURI = process.env.COPILOT_QUEUE_URI"
 	actionRetrieveEnvVar := fmt.Sprintf(
 		`Update %s's code to leverage the injected environment variable "COPILOT_QUEUE_URI".
-In JavaScript you can write %s.`,
+  In JavaScript you can write %s.`,
```
Lou1415926, line 222: Do you think we want to move the call to `generateWorkerServiceRecommendedActions` inside `RecommandedActions()`?

Lou1415926 (approved):

internal/pkg/cli/svc_deploy.go
efekarakus: Yup, I'll submit a separate refactor PR on monday. 

I think recommended actions should be something like this:
```go
type reachable interface {
   Port() int
}
type publisher interface {
    Topics() []*manifest.Topics
}
type subscriber interface {
    Subscriptions() []*manifest.Subscription
}

var actions []string
if _, ok := mft.(reachable); ok {
   actions = append(actions, o.generateEndpointRecs()...)
}
if _, ok := mft.(publisher); ok {
   actions = append(actions, o.generatePublishRecs()...)
}
if _, ok := mft.(subscriber); ok {
   actions = append(actions, o.generateSubscribeRecs()...)
}
```

This way the moment a new manifest allows the property of being reachable, or sns publishing then recommended actions will get the suggestion for free.

Hope that makes sense!

Lou1415926: This is cool!! Looking forward to it!

iamhopaul123 (approved):
:shipit:

internal/pkg/cli/svc_deploy.go
efekarakus: Here is the commit with the changes: https://github.com/efekarakus/copilot-cli/commit/689745434b06ccdc859ec8e78ce1127ee98d175b (PR#2787)

PR #2784, opened 2021-08-28
chore(cli): enable combination of flags and prompting for custom env resources

### LINKED ISSUE #2777: Error on env init command when importing a vpc as a cli parameter without subnets

huanjani:
Previously, it was assumed that users would 1. import neither VPC nor subnets and use the guided experience, or 2. import both VPC and subnets. This enables them to import one without the other.

Fixes #2777.

Use cases with these changes:

- import all custom resources: proceed as usual, including validation of two private subnets (okay to leave public subnets blank--log warning about load balancers needing two, but must have two private)
- import no custom resources: prompt to select VPC and subnets
- import just VPC: prompt to select subnets
- import two private: prompt to select VPC and public subnets
- import two public: prompt to select VPC and private subnets
- import just one private: error
- import just one public: warning, prompt to select VPC and private subnets


internal/pkg/cli/env_init.go
```diff
@@ -329,12 +329,11 @@ func (o *initEnvOpts) validateCustomizedResources() error {
 		return fmt.Errorf("cannot import or configure vpc if --%s is set", defaultConfigFlag)
 	}
 	if o.importVPC.isSet() {
-		// We allow 0 or 2+ public subnets.
+		// Allow passing in VPC without subnets, but error out early for too few subnets-- we won't prompt the user to select more of one type if they pass in any.
 		if len(o.importVPC.PublicSubnetIDs) == 1 {
-			return fmt.Errorf("at least two public subnets must be imported to enable Load Balancing")
+			log.Warningf("At least two public subnets must be imported to enable Load Balancing.\n")
```
Lou1415926:
```suggestion
			log.Warningln("At least two public subnets must be imported to enable Load Balancing.")
```

efekarakus: Should we still return an error here, but instead we can check if they set the flag and if they did then they must have specified at least two.

```go
if n := o.importVPC.PublicSubnets; n > 0 && n < 2 {
   return errors.New("at least two public subnets must be imported to enable Load Balancing")
}
```

internal/pkg/cli/env_init.go
```diff
@@ -329,12 +329,11 @@ func (o *initEnvOpts) validateCustomizedResources() error {
...
 		}
-		// We require 2+ private subnets.
-		if len(o.importVPC.PrivateSubnetIDs) < 2 {
+		if len(o.importVPC.PrivateSubnetIDs) == 1 {
```
efekarakus, line 336:
```suggestion
		if n := o.importVPC.PrivateSubnetIDs; n > 0 && n < 2 {
```

internal/pkg/cli/env_init.go
```diff
@@ -468,10 +467,16 @@ https://aws.amazon.com/premiumsupport/knowledge-center/ecs-pull-container-api-er
 		return fmt.Errorf("VPC %s has no DNS support enabled", o.importVPC.ID)
 	}
 	if o.importVPC.PublicSubnetIDs == nil {
-		publicSubnets, err := o.selVPC.PublicSubnets(envInitPublicSubnetsSelectPrompt, "", o.importVPC.ID)
+		input := selector.SubnetsInput{
```
efekarakus: nit: can we remove the variable `o.selVPC.Subnets({...})`

internal/pkg/cli/env_init.go
```diff
@@ -468,10 +467,16 @@ https://aws.amazon.com/premiumsupport/knowledge-center/ecs-pull-container-api-er
...
+			Msg:      envInitPublicSubnetsSelectPrompt,
+			Help:     "",
+			VPCID:    o.importVPC.ID,
+			IsPublic: true,
+		}
+		publicSubnets, err := o.selVPC.Subnets(input)
 		if err != nil {
 			if err == selector.ErrSubnetsNotFound {
```
```suggestion
			if errors.Is(err, selector.ErrSubnetsNotFound) {
```

internal/pkg/term/selector/ec2.go
```diff
@@ -68,30 +68,28 @@ func (s *EC2Select) VPC(msg, help string) (string, error) {
 	return extractedVPC.ID, nil
 }
 
-// PublicSubnets has the user multiselect public subnets given the VPC ID.
-func (s *EC2Select) PublicSubnets(msg, help, vpcID string) ([]string, error) {
-	return s.selectPublicSubnets(msg, help, vpcID)
-}
+type SubnetsInput struct {
```
efekarakus, line 72: Comments please 🙏

efekarakus (commented):
Looks great! Minor feedback 🥺

internal/pkg/cli/env_init.go
Lou1415926: I wonder if we should return an error here.

If we error out when we can only find one public subnet: https://github.com/aws/copilot-cli/blob/ab91cd60a0e526cb9198e941b2f3f42c3a9df923/internal/pkg/cli/env_init.go#L486 
```
if len(publicSubnets) == 1 {
   return errors.New("select public subnets: at least two public subnets must be selected to enable Load Balancing")
}
```
during `Ask()`, then we should probably also error out here as well.

Lou1415926 (commented):
LGTM!

internal/pkg/cli/env_init.go
huanjani: But shouldn't they be allowed to import 1 public and 2 private if they don't want load balancing?

huanjani: Hm. I wonder if instead we should not error out in `Ask()` (see my response to Efe, above)

internal/pkg/cli/env_init.go
huanjani: Discussed offline; leaving as is bc 1 is the only int between 0 and 2.

internal/pkg/cli/env_init.go
huanjani: Sorry-- I'm not really sure what you mean.... 😬

internal/pkg/cli/env_init.go
efekarakus: Hmm that's a good question!

My intuition was that we should be consistent with `Ask` (where we err if only 1 public subnet is selected)
I think for this scenario, we can err on the side of being restrictive at first and then relax it if customers ask for it.

internal/pkg/cli/env_init.go
efekarakus: 🤣 sorry that wasn't clear! 

```go
o.selVPC.Subnets(selector.SubnetsInput{
	Msg:      envInitPublicSubnetsSelectPrompt,
	VPCID:    o.importVPC.ID,
	IsPublic: true,
})
```

internal/pkg/cli/env_init.go
efekarakus: +1 sorry I just saw this chain of comments haha, I think it makes sense to err if they only choose 1 public subnet. It's most likely going to lead to problems rather than make them happy. If they want any public subnets (at least 2 most be selected), otherwise they can completely opt out.

efekarakus (approved):

Lou1415926 (approved):

PR #2787, opened 2021-08-28
chore: leverage interfaces to recommend deploy actions instead of switch statements

efekarakus:
There are also a _bajillion_ tiny refactors along the way.
The changes look big but it shouldn't be too bad mostly simplifying code or moving them. 

The main new logic can be found in this commit: https://github.com/efekarakus/copilot-cli/commit/689745434b06ccdc859ec8e78ce1127ee98d175b, the rest are refactors.

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._


internal/pkg/cli/cli.go
```diff
@@ -94,3 +96,31 @@ func relPath(fullPath string) (string, error) {
...
+	}
+	return nil
+}
+
+func logRecommendedActions(actions []string) {
+	if len(actions) == 0 {
+		return
+	}
+	log.Infoln("Recommended follow-up actions:")
+	for _, followup := range actions {
```
Lou1415926, line 129: What do you think of removing the "-" bullet point if `len(actions) == 1`?

bvtujo (approved):
DUDE, this is an awesome and super necessary refactor. Love the new run() logic.

internal/pkg/cli/svc_deploy.go
```diff
@@ -818,65 +811,88 @@ func (o *deploySvcOpts) retrieveAppResourcesForEnvRegion() error {
...
-		})
-	case manifest.WorkerServiceType:
-		o.generateWorkerServiceRecommendedActions()
-		log.Successf("Deployed %s.\n", color.HighlightUserInput(o.name))
-		return nil
-	default:
-		err = errors.New("unexpected service type")
+	describer, err := describe.NewReachableService(o.appName, o.name, o.store)
+	if err != nil {
+		return nil, err
```
Lou1415926, line 828: Would we want to wrap the error here `err`?

internal/pkg/describe/uri.go
```diff
@@ -0,0 +1,163 @@
...
+	fmtSvcDiscoveryEndpointWithPort = "%s.%s:%s" // Format string of the form {svc}.{endpoint}:{port}
+)
+
+// ReachableService represents a service describe that has an endpoint.
+type ReachableService interface {
+	URI(env string) (string, error)
+}
+
+// NewReachableService returns a ReachableService based on the type of the service.
+func NewReachableService(app, svc string, store ConfigStoreSvc) (ReachableService, error) {
```
Lou1415926, line 25: yasss

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -157,17 +121,23 @@ func (s *LoadBalancedWebService) MarshalBinary() ([]byte, error) {
 	return content.Bytes(), nil
 }
 
-// BuildRequired returns if the service requires building from the local Dockerfile.
-func (s *LoadBalancedWebService) BuildRequired() (bool, error) {
-	return requiresBuild(s.ImageConfig.Image)
+// Port returns the exposed port in the manifest.
+// A LoadBalancedWebService always has a port exposed therefore the boolean is always true.
+func (s *LoadBalancedWebService) Port() (port uint16, ok bool) {
+	return aws.Uint16Value(s.ImageConfig.Port), true
```
Lou1415926, line 127: Should we have errored out somewhere if the manifest doesn't have `port` 🤔 ?

internal/pkg/manifest/rd_web_svc.go
```diff
@@ -87,6 +71,20 @@ func (s *RequestDrivenWebService) MarshalBinary() ([]byte, error) {
 	return content.Bytes(), nil
 }
 
+// Port returns the exposed the exposed port in the manifest.
+// A RequestDrivenWebService always has a port exposed therefore the boolean is always true.
+func (s *RequestDrivenWebService) Port() (port uint16, ok bool) {
+	return aws.Uint16Value(s.ImageConfig.Port), true
```
Lou1415926, line 77: Same here - should we have errored out if `port` is not found in manifest?

Lou1415926 (approved):
This is awesome 😻  Putting the label just in case you want/need to address the feedback in this PR.

internal/pkg/cli/cli.go
efekarakus: Done! also did `english.PluralWord(len(actions, "action", "actions")`

internal/pkg/cli/svc_deploy.go
efekarakus: I didn't wrap on purpose because I think the underlying err msg is clear 💭

internal/pkg/manifest/lb_web_svc.go
efekarakus: mmm good question! I think it makes sense from the PoV of this method to assume that there is always a  `Port` for LBWS and RDWS.

I believe we always add 80 no matter what if the port is not set :)

iamhopaul123: Hmm that is not true...If port is not set in the manifest we'll set it to `0` instead of `80`. But I believe that validation should be part of the manifest validation im going to work on.

Lou1415926: I asked because I remember there being a segfault when `port` wasn't set for rd web service (the user must have removed the `port` line from manifest). 

The error happened during `deploy.cloudformation.stack`'s `Template()` function. That'd mean that we didn't check for the existence of the `port` field before, and didn't set a default port either.

But yeah I agree that the validation should happens during unmarshalling/manifest validation.

PR #2798, opened 2021-09-01
chore(manifest): refactor pointer struct to struct in manifest

Lou1415926:
This PR is a refactor that removes unnecessary pointers ™️ from manifest.

"Necessary pointers" ™️  to the following structs are not removed:
##### <s>`ContainerHealthCheck`</s> and `PlatfromArgsOrString` for all workloads (e.g. https://github.com/aws/copilot-cli/blob/ff0a32e3d0d6821408ba6f05fcad6a0450f75f24/internal/pkg/manifest/lb_web_svc.go#L64)
This is because of the way we marshal manifest object into YAML during `svc init` or `job init`.

<s>##### `EFSConfigOrBool`
A manifest input that has:
```
efs:
<the rest of the manifest>
```
is deemed invalid, and we error out during the conversion from a manifest object to a template object. We cannot remove the pointer right now because otherwise we cannot differentiate a said invalid input, and a valid input that didn't specify `efs` at all. 

This validation should instead happens at an earlier stage, potentially from within a dedicated `manifest.validation` package. We can remove the pointer after we can move the validation to the said earlier stage.</s>

##### `SQSQueue `
For this particular manifest field, a manifest input that has:
```
topics:
     - name: topic1
        queue:
<the rest of the manifest>
```
is deemed valid. If the `queue` field is specified without any value, we create the queue for that topic with default configurations. Hence, there is indeed a need to differentiate an empty queue from an absent queue. 


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -608,22 +601,18 @@ func convertNetworkConfig(network *manifest.NetworkConfig) *template.NetworkOpts
...
 }
 
-func convertEntryPoint(entrypoint *manifest.EntryPointOverride) ([]string, error) {
-	if entrypoint == nil {
-		return nil, nil
-	}
+func convertEntryPoint(entrypoint manifest.EntryPointOverride) ([]string, error) {
+	//if entrypoint == nil {
+	//	return nil, nil
+	//}
```
iamhopaul123: Should we remove them ❓

internal/pkg/manifest/storage.go
```diff
@@ -21,9 +21,14 @@ type Storage struct {
...
 
+// IsEmpty returns empty if the struct has all zero members.
+func (s *Storage) IsEmpty() bool {
+	return s.Ephemeral == nil && s.Volumes == nil
+}
+
 // Volume is an abstraction which merges the MountPoint and Volumes concepts from the ECS Task Definition
 type Volume struct {
-	EFS            *EFSConfigOrBool `yaml:"efs"`
+	EFS            *EFSConfigOrBool `yaml:"efs"` // TODO: this pointer can be removed after manifest validation is implemented. We need it to be a pointer right now to differentiate a non-specified EFS from an invalid empty EFS.
```
iamhopaul123: Maybe we don't need the comment and the pointer now.

internal/pkg/manifest/workload.go
```diff
@@ -186,8 +186,8 @@ func (i *Image) cacheFrom() []string {
 
 // ImageOverride holds fields that override Dockerfile image defaults.
 type ImageOverride struct {
-	EntryPoint *EntryPointOverride `yaml:"entrypoint"` // TODO: the type needs to be updated after we upgrade mergo
-	Command    *CommandOverride    `yaml:"command"`    // TODO: the type needs to be updated after we upgrade mergo
```
iamhopaul123, line 190: 🙏

iamhopaul123 (commented):
LGTM overall! Will approve after resolving these comments.

iamhopaul123 (approved):

internal/pkg/manifest/backend_svc.go
```diff
@@ -28,10 +28,10 @@ type BackendServiceConfig struct {
 	ImageConfig      ImageWithPortAndHealthcheck `yaml:"image,flow"`
 	ImageOverride    `yaml:",inline"`
 	TaskConfig       `yaml:",inline"`
-	*Logging         `yaml:"logging,flow"`
+	Logging          `yaml:"logging,flow"`
 	Sidecars         map[string]*SidecarConfig `yaml:"sidecars"`
```
efekarakus: can you remind me why we keep the pointer if it's a `map[string]<PStruct>`? are there other scenarios where the pointer is kept?

internal/pkg/manifest/workload.go
```diff
@@ -92,13 +92,13 @@ type Image struct {
 // ImageWithHealthcheck represents a container image with health check.
 type ImageWithHealthcheck struct {
 	Image       `yaml:",inline"`
-	HealthCheck *ContainerHealthCheck `yaml:"healthcheck"`
+	HealthCheck *ContainerHealthCheck `yaml:"healthcheck"` // This is used as a pointer so that we can use `{{- if .ContainerHealthCheck}}` for manifest rendering.
```
efekarakus: Can we do `{{- if not .ImageConfig.HealthCheck.IsEmpty}}` instead?

efekarakus (commented):
Looks good just a question 🤓

internal/pkg/manifest/backend_svc.go
Lou1415926: so that `mergo` will deep merge `SideCarConfig` as well. For example, say the original has a key of `sidecar1`, and the override also has a key `sidecar1`. Without the pointer, `mergo` replaces the `sidecar1` in the original with the one in  override, instead of deep merging them.

efekarakus: Nice can we add this as a comment if it's not mentioned somewhere 🥺

internal/pkg/manifest/workload.go
Lou1415926: Updated! :homer-disappering:

efekarakus (approved):

Issue #2796, opened 2021-09-01
proposal: Autoscaling for Worker Services


efekarakus:
## Proposal
This issue proposes introducing autoscaling for [Worker Services](https://aws.github.io/copilot-cli/docs/concepts/services/#worker-service) in the manifest based on the acceptable amount of queue delay.
```yaml
count:
  range: 1-10
  queue_delay:                    # <-- new field for Worker Services.               
    acceptable_latency: 10m        
    msg_processing_time: 250ms
```
With the example above, the goal is to ensure that messages arriving to the Worker Service are processed within 10 minutes while fluctuating between 1 to 10 tasks.
The `msg_processing_time` is the average processing time for a message and is required in order for Copilot to compute the acceptable backlog per task: `10*60/0.25 = 2400` messages per task. Copilot will then setup target tracking scaling and ensure that tasks scale to maintain the acceptable backlog per task.

See [Scaling based on Amazon SQS](https://docs.aws.amazon.com/autoscaling/ec2/userguide/as-using-sqs-queue.html) for additional details.

For details for how we're going to setup target tracking to maintain acceptable latency:
<details>

We will create an AWS Lambda function to emit a new Amazon CloudWatch metric: `BacklogPerTask`.

![worker-service-autoscaling](https://user-images.githubusercontent.com/879348/131725449-c62a452a-1745-4099-9673-576b447540f9.png)

When autoscaling is enabled,
1. Every minute a CloudWatch event will trigger a Lambda function: `BacklogPerTaskCalculator`
2. The lambda function will get the `ApproximateNumberOfMessagesVisible` for each queue in the Worker Service stack.
3. Then it will query for the number of running tasks in the ECS service. Dividing the two numbers yields the BacklogPerTask value.
4. It will write a log to CloudWatch logs in following the [CloudWatch embedded metric format](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Generation.html) in order to have the metric appear.

</details>

efekarakus:
This is now released in https://github.com/aws/copilot-cli/releases/tag/v1.11.0! 🎉

Docs: https://aws.github.io/copilot-cli/docs/manifest/worker-service/#count-queue-delay

PR #2804, opened 2021-09-02
chore: add Lambda fn to write BacklogPerTask CW metric

efekarakus:
We need to write this metric for Worker Service autoscaling.

Related #2796

### DIFF
--- a/cf-custom-resources/lib/backlog-per-task-calculator.js
+++ b/cf-custom-resources/lib/backlog-per-task-calculator.js
@@ -15,13 +15,19 @@ exports.handler = async (event, context) => {
   setupClients();
   try {
     const runningCount = await getRunningTaskCount(process.env.CLUSTER_NAME, process.env.SERVICE_NAME);
-    const timestamp = Date.now();
-    const promises = convertQueueNames(process.env.QUEUE_NAMES).map(async (queueName) => {
+    const backlogs = await Promise.all(
+      convertQueueNames(process.env.QUEUE_NAMES).map(async (queueName) => {
         const queueUrl = await getQueueURL(queueName);
-        const backlogPerTask = await getBacklogPerTask(queueUrl, runningCount);
-        logBacklogPerTaskMetric(process.env.NAMESPACE, timestamp, queueName, backlogPerTask);
-    });
-    await Promise.all(promises);
+        return {
+          queueName: queueName,
+          backlogPerTask: await getBacklogPerTask(queueUrl, runningCount),
+        };
+      })
+    );
+    const timestamp = Date.now();
+    for (const {queueName, backlogPerTask} of backlogs) {
+      emitBacklogPerTaskMetric(process.env.NAMESPACE, timestamp, queueName, backlogPerTask);
+    }
   } catch(err) {
     // If there is any issue we won't log a metric.
     // This is okay because autoscaling will maintain the current number of running tasks if a data point is missing.
@@ -52,7 +58,7 @@ const getBacklogPerTask = async (queueUrl, runningTaskCount) => {
  * @param queueName The name of the queue.
  * @param backlogPerTask The number of messages in the queue divided by the number of running tasks.
  */
-const logBacklogPerTaskMetric = (namespace, timestamp, queueName, backlogPerTask) => {
+const emitBacklogPerTaskMetric = (namespace, timestamp, queueName, backlogPerTask) => {
   console.log(JSON.stringify({
     "_aws": {
       "Timestamp": timestamp,
--- a/cf-custom-resources/lib/backlog-per-task-calculator.js
+++ b/cf-custom-resources/lib/backlog-per-task-calculator.js
@@ -1,0 +1,123 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+"use strict";
+
+const aws = require("aws-sdk");
+
+// AWS Clients that are overriden in tests.
+let ecs, sqs;
+
+/**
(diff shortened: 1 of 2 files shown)


cf-custom-resources/lib/backlog-per-task-calculator.js
```diff
@@ -0,0 +1,123 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+"use strict";
+
+const aws = require("aws-sdk");
+
+// AWS Clients that are overriden in tests.
+let ecs, sqs;
+
```
iamhopaul123, line 9: We don't even need a report function?

cf-custom-resources/lib/backlog-per-task-calculator.js
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * This lambda function calculates the backlog of SQS messages per running ECS tasks,
+ * and writes the metric to CloudWatch.
+ */
+exports.handler = async (event, context) => {
+  setupClients();
+  try {
+    const runningCount = await getRunningTaskCount(process.env.CLUSTER_NAME, process.env.SERVICE_NAME);
```
iamhopaul123, line 17: Where do we define `process`?

cf-custom-resources/lib/backlog-per-task-calculator.js
efekarakus: Yeah! Because this Lambda won't be invoked by a Custom Resource, it's instead going to be invoked every minute by a CloudWatch event so we don't need it.

cf-custom-resources/lib/backlog-per-task-calculator.js
efekarakus: It's a global module that's always available! https://nodejs.org/en/knowledge/getting-started/the-process-module/

cf-custom-resources/lib/backlog-per-task-calculator.js
iamhopaul123: gotcha!

cf-custom-resources/lib/backlog-per-task-calculator.js
```diff
@@ -0,0 +1,123 @@
...
+    const timestamp = Date.now();
```
iamhopaul123: Just a nit but maybe we should do it inside of `logBacklogPerTaskMetric` to make sure we are using the time right now.

iamhopaul123 (approved):
:shipit:

cf-custom-resources/lib/backlog-per-task-calculator.js
efekarakus: I wanted to share the same timestamp across all queues but it's a good point 💭 
Ideally it would happen after retrieving the backlogPerTask metric. I think we can refactor to do that

efekarakus: Done ! https://github.com/aws/copilot-cli/pull/2804/commits/9ab295833960dfb732e9ccd7e3bd9e7f88646d4d

cf-custom-resources/lib/backlog-per-task-calculator.js
```diff
@@ -0,0 +1,129 @@
...
+    );
+    const timestamp = Date.now();
+    for (const {queueName, backlogPerTask} of backlogs) {
+      emitBacklogPerTaskMetric(process.env.NAMESPACE, timestamp, queueName, backlogPerTask);
+    }
+  } catch(err) {
+    // If there is any issue we won't log a metric.
+    // This is okay because autoscaling will maintain the current number of running tasks if a data point is missing.
+    // See https://docs.aws.amazon.com/AmazonECS/latest/developerguide/service-autoscaling-targettracking.html#targettracking-considerations
+    console.error(`Unexpected error ${err}`);
```
Lou1415926, line 35: How do user be aware of the errors? Is it only when they realize the autoscaling isn't quite right, or do they get informed somehow if the lambda logs errors?

cf-custom-resources/lib/backlog-per-task-calculator.js
```diff
@@ -0,0 +1,129 @@
...
+  }
+};
+
+/**
+ * Returns the backlog per task. The backlog per task is calculated by dividing the number of messages in the queue with
+ * the number of running tasks.
+ * If there are no running task, we return the total number of messages in the queue so that we can start scaling up.
```
Lou1415926, line 42: Could it be possible that the calculated acceptable backlog > total number of messages, and in this case, do we want to scale up from 0, or do we want to stay 0?

Lou1415926 (commented):
LGTM!! One question on when #tasks = 0

cf-custom-resources/lib/backlog-per-task-calculator.js
efekarakus: Ohh this is a great question 🤔 

I didn't have this part of the original design but in `svc status` we can display an alarm for number of errors associated with this Lambda function. This way the user can be aware of the issue and then go dumpster-diving in the logs.

cf-custom-resources/lib/backlog-per-task-calculator.js
efekarakus: > Could it be possible that the calculated acceptable backlog > total number of messages

Hmm I don't think this is possible (unless I am missing something 🤔 ) We can only end up in this scenario if `runningCount` from ECS returns a float :P between [0, 1). But I might be missing something.

> do we want to scale up from 0,

Yeah, that's the intention. If there are messages in the queue then I think we should scale up otherwise if empty remain at 0.

Lou1415926 (approved):
:shipit: ship itt

PR #2806, opened 2021-09-03
chore: remove `applyEnv` unit tests and keep only those that are representative

Lou1415926:
Previously I've added a plethora of `ApplyEnv` unit tests to baseline our `ApplyEnv` bahaviours on each of the manifest, in attempt to upgrade mergo and fix manifest behaviors.

Now that the upgrade is completed, these extensive tests can be removed, replaced with a few `ApplyEnv` tests for each of the representative field. 

This PR is dependent on #2798.


internal/pkg/manifest/lb_web_svc.go
```diff
@@ -52,7 +52,7 @@ type LoadBalancedWebServiceConfig struct {
 	Logging          `yaml:"logging,flow"`
 	Sidecars         map[string]*SidecarConfig `yaml:"sidecars"` // NOTE: keep the pointers because `mergo` doesn't automatically deep merge map's value unless it's a pointer type.
 	Network          NetworkConfig             `yaml:"network"`
-	Publish          PublishConfig             `yaml:"publish"`
+	PublishConfig    PublishConfig             `yaml:"publish"`
```
iamhopaul123, line 55: Why do we rename this field? After renaming this seems to be not consistent with the `Network` above.

iamhopaul123 (approved):
LGTM just one question

internal/pkg/manifest/lb_web_svc.go
Lou1415926: Oh it was because this field has the same name as the struct function `Publish()`, which caused problem during testing (an error like `Publish` is not a function). I should actually do the same for the other workloads.

efekarakus (approved):

PR #2636, opened 2021-07-19
feat: build ARM-architecture-based images as AMD64

huanjani:
Previously, users could not build ARM-based images with Copilot (without employing workarounds) because Fargate does not currently support ARM-based tasks. This PR automatically builds images as `linux/amd64` when ARM (or other non-AMD) architectures are detected, using Docker's `--platform` flag. This is a best-effort redirect, as users' Docker servers must be multi-platform-capable in order for `--platform` to work. A warning is logged at `svc init` and info is logged at `svc deploy` IF non-amd64 architectures are detected.

An additional manifest field, `platform:` is surfaced iff OS/arch is being redirected, to provide visibility to the user and for extensibility (we can take changes in the future). Validation is performed when the manifest is unmarshaled. The field currently takes a string, but will accept a map in a future PR. 

_The workflow/experience does not change for amd64 users!_

Related: #2545; closes: #1949, #2369.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2814, opened 2021-09-08
fix: extend ARM platform changes through to deployment

huanjani:
These changes enable workload images on ARM machines to build as amd64 images.
If the underlying docker engine detected is ARM architecture-based, Copilot builds with `--platform linux/amd64` so workloads deploy rather than going through the whole workflow only to fail to spin up. 

Related: #2636.


internal/pkg/cli/job_init_test.go
```diff
@@ -525,7 +525,7 @@ func TestJobInitOpts_Execute(t *testing.T) {
 						Name:           "mailer",
 						Type:           "Scheduled Job",
 						DockerfilePath: "./Dockerfile",
-						Platform:       nil,
+						Platform:       &manifest.PlatformArgsOrString{PlatformString: nil},
```
efekarakus: nit:
```suggestion
						Platform:       &manifest.PlatformArgsOrString{},
```

internal/pkg/cli/svc_init_test.go
```diff
@@ -623,7 +623,7 @@ func TestSvcInitOpts_Execute(t *testing.T) {
 						Name:           "frontend",
 						Type:           "Load Balanced Web Service",
 						DockerfilePath: "./Dockerfile",
-						Platform:       nil,
+						Platform:       &manifest.PlatformArgsOrString{PlatformString: nil},
```
efekarakus: samesies here

internal/pkg/manifest/rd_web_svc.go
```diff
@@ -57,6 +58,9 @@ func NewRequestDrivenWebService(props *RequestDrivenWebServiceProps) *RequestDri
 	svc.RequestDrivenWebServiceConfig.ImageConfig.Image.Location = stringP(props.Image)
 	svc.RequestDrivenWebServiceConfig.ImageConfig.Build.BuildArgs.Dockerfile = stringP(props.Dockerfile)
 	svc.RequestDrivenWebServiceConfig.ImageConfig.Port = aws.Uint16(props.Port)
+	if props.Platform != nil {
+		svc.RequestDrivenWebServiceConfig.InstanceConfig.Platform = props.Platform
+	}
```
efekarakus: I'm trying to remember why we even allow instance config to have a Platform for RDWS? since it only supports linux arm64

internal/pkg/template/templates/workloads/jobs/scheduled-job/manifest.yml
```diff
@@ -34,8 +34,8 @@ image:
 
 cpu: {{.CPU}}       # Number of CPU units for the task.
 memory: {{.Memory}}    # Amount of memory in MiB used by the task.
-{{- if .Platform}}
-platform: {{.Platform}}   # See https://aws.github.io/copilot-cli/docs/manifest/scheduled-job/#platform
+{{- if .Platform.PlatformString}}
```
efekarakus, line 37: There is a hidden assumption here that the `.Platform` field is always non-nil. This is true because of the `cli` pkg but it's not obvious. These templates are rendered by the `manifest` pkg so it should be its responsibility to ensure that `Platform` always has a value.  


Can we maybe add this behavior in the default constructors?
```go
newDefaultBackendService() {
   ...
   BackendServiceConfig: BackendServiceConfig {
       ...,
       Platform: &PlatformArgsOrString{}, // Ensure .Platform is always set.
   }
}
```

internal/pkg/manifest/rd_web_svc.go
huanjani: Took it out of `RequestDrivenWebServiceProps` bc it won't go in manifests, but don't we need it in InstanceConfig for BuildArgs?

internal/pkg/manifest/backend_svc.go
```diff
@@ -52,6 +53,9 @@ func NewBackendService(props BackendServiceProps) *BackendService {
 	svc.BackendServiceConfig.ImageConfig.Build.BuildArgs.Dockerfile = stringP(props.Dockerfile)
 	svc.BackendServiceConfig.ImageConfig.Port = uint16P(props.Port)
 	svc.BackendServiceConfig.ImageConfig.HealthCheck = props.HealthCheck
+	if props.Platform != nil {
+		svc.BackendServiceConfig.Platform = props.Platform
+	}
```
iamhopaul123: Is it equivalent to
```suggestion
	svc.BackendServiceConfig.Platform = props.Platform
```
right now?

internal/pkg/template/templates/workloads/jobs/scheduled-job/manifest.yml
iamhopaul123: Why don't we set it to be struct instead of pointer of struct?

internal/pkg/manifest/rd_web_svc.go
huanjani: Won't we need this field if `ARM` is redirecting to `amd64` for `BuildArgs`?

internal/pkg/template/templates/workloads/jobs/scheduled-job/manifest.yml
huanjani: It's small, but for space considerations? We have other pointers to structs for other default values, too.

internal/pkg/manifest/backend_svc.go
huanjani: Without the if-statement, the default behavior (when there isn't a platform redirection) would make the whole Platform field nil, not just the member fields. So if we want the field to always be set as default, then we should leave the if-statement in.

internal/pkg/template/templates/workloads/jobs/scheduled-job/manifest.yml
iamhopaul123: It's not small now lol. we are now trying to make it consistent to always use non pointer unless it is necessary. For this one using non pointer we can be sure that .Platform is always set.

internal/pkg/manifest/backend_svc.go
iamhopaul123: but what's the difference between platform and [healthcheck](https://github.com/aws/copilot-cli/pull/2814/files#diff-2147c7bfea12eb3dff6a866c245ac1ed73e671d7c47bf06ff906e9dfc741d380R55)?

> the default behavior (when there isn't a platform redirection) would make the whole Platform field nil, not just the member fields.

But this is a constructor, `svc.BackendServiceConfig.Platform` should be initially nil right (or empty struct if you remove the pointer)?

internal/pkg/cli/job_init.go
```diff
@@ -194,6 +194,9 @@ func (o *initJobOpts) Execute() error {
 		return err
 	}
 	o.platform = platform
+	if o.platform != nil {
+		log.Warningf("Your architecture type is currently unsupported. Setting platform %s instead.\nSee 'platform' field in your manifest.\n", dockerengine.DockerBuildPlatform(dockerengine.LinuxOS, dockerengine.Amd64Arch))
```
efekarakus: nit: maybe
```suggestion
		log.Warningf("Your architecture type is currently unsupported. Setting 'platform' to %s instead in your manifest.\n", dockerengine.DockerBuildPlatform(dockerengine.LinuxOS, dockerengine.Amd64Arch))
```

internal/pkg/manifest/applyenv_test.go
```diff
@@ -1134,18 +1134,18 @@ func TestApplyEnv_Platform(t *testing.T) {
 	}{
 		"FIXED_BUG: composite fields: platform string is overridden if platform args is not nil": {
 			inSvc: func(svc *LoadBalancedWebService) {
-				svc.Platform = &PlatformArgsOrString{
+				svc.Platform = PlatformArgsOrString{
```
efekarakus: I think this test was deleted now in mainline 😅

internal/pkg/manifest/backend_svc.go
```diff
@@ -128,8 +130,9 @@ func newDefaultBackendService() *BackendService {
 		BackendServiceConfig: BackendServiceConfig{
 			ImageConfig: ImageWithPortAndHealthcheck{},
 			TaskConfig: TaskConfig{
-				CPU:    aws.Int(256),
-				Memory: aws.Int(512),
+				CPU:      aws.Int(256),
+				Memory:   aws.Int(512),
+				Platform: PlatformArgsOrString{},
```
efekarakus: no need to define it here now since it's no longer a pointer, the zero value `PlatformArgsOrString{}`

internal/pkg/manifest/job.go
```diff
@@ -144,8 +146,9 @@ func newDefaultScheduledJob() *ScheduledJob {
 		ScheduledJobConfig: ScheduledJobConfig{
 			ImageConfig: ImageWithHealthcheck{},
 			TaskConfig: TaskConfig{
-				CPU:    aws.Int(256),
-				Memory: aws.Int(512),
+				CPU:      aws.Int(256),
+				Memory:   aws.Int(512),
+				Platform: PlatformArgsOrString{},
```
efekarakus: same here, we can remove this field from all constructors :)

efekarakus (commented):
Looks great! just small refactor to remove the field from the constructor.

I'll approve :shipit: after mainline is merged!

internal/pkg/manifest/backend_svc_test.go
```diff
@@ -91,8 +92,9 @@ func TestNewBackendSvc(t *testing.T) {
 						},
 					},
 					TaskConfig: TaskConfig{
-						CPU:    aws.Int(256),
-						Memory: aws.Int(512),
+						CPU:      aws.Int(256),
+						Memory:   aws.Int(512),
+						Platform: PlatformArgsOrString{},
```
iamhopaul123: same here...

iamhopaul123 (approved):

internal/pkg/cli/job_init.go
huanjani: Okay, but I'll leave the wording as is in `svc_init` because of the platform field not being surfaced in RDWS.

efekarakus (approved):
pew pew :shipit:

PR #2815, opened 2021-09-08
chore(manifest): add manifest validation boilerplate

iamhopaul123:
Part of #2818. This PR adds boilerplate for new manifest validation for LB svc. Also added validation for mutually exclusive fields. Will add for the other workload types, advanced type validation, and consolidate existing validations in following PRs


internal/pkg/manifest/errors.go
```diff
@@ -4,9 +4,21 @@
...
 )
 
+var (
+	errEmptyContainerPort        = errors.New(`"port" must be specified`)
+	errInvalidBuildAndLocation   = errors.New(`must specify one, not both, not none, of "build" and "location"`)
+	errDuplicatedTargetContainer = errors.New(`must specify one, not both, of "target_container" and "targetContainer"`)
+	errInvalidRangeOpts          = errors.New(`must specify one, not both, of "range" and "min"/"max"`)
+	errInvalidAdvancedCount      = errors.New(`must specify one, not both, of "spot" and autoscaling fields`)
+	errInvalidAutoscaling        = errors.New(`must specify "range" if using autoscaling`)
+	errInvalidEFSConfiguration   = errors.New(`must specify one, not both, of "uid/gid" and "id/root_dir/auth"`)
```
efekarakus: What do you think of creating custom error types so that can guarantee consistency for new fields:

For example,
```go
type errFieldMustBeSpecified struct {
   field string
   parentField string // Optional. If empty we don't write "... if using '{parentField}'"
}

type errFieldMutualExclusive struct {
   firstField string
   secondField string
}
```

efekarakus: I think `errInvalidEFSAccessPoint` is unique enough that it doesn't warrant its own error type, maybe we can move it directly inside the appropriate `Validate` method? (so not even have it as a variable)

internal/pkg/manifest/errors.go
```diff
@@ -50,3 +62,20 @@ func (e *ErrUnknownProvider) Is(target error) bool {
...
 	return ok
 }
+
+type errInvalidIntRangeBand struct {
+	value string
+}
+
+func (e *errInvalidIntRangeBand) Error() string {
+	return fmt.Sprintf("invalid range value %s. Should be in format of ${min}-${max}", string(e.value))
+}
```
efekarakus: I think this is also unique enough that we can move it inside `(r IntRangeBand) Validate() error`

internal/pkg/manifest/errors.go
```diff
@@ -50,3 +62,20 @@ func (e *ErrUnknownProvider) Is(target error) bool {
...
+
+type errMinGreaterThanMax struct {
+	min int
+	max int
+}
+
+func (e *errMinGreaterThanMax) Error() string {
+	return fmt.Sprintf("min value %d cannot be greater than max value %d", e.min, e.max)
+}
```
efekarakus, line 90: samesies here i think this is more unique to how "min" and "max" fields work so it doesn't need a type. 

We should create types I think for only errors that can be re-used across fields. For value ranges for example, we can do:
```go
type errFieldInvalidEnumValue struct {
   fieldName string
   fieldValue interface{}
   
   validValues []interface{}
}

// example:
&errFieldInvalidEnumValue{fieldName: "cpu", fieldValue: 10, validValues: []interface{}{256, 512, ...}}
```

internal/pkg/manifest/validate.go
```diff
@@ -0,0 +1,388 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+)
+
+var (
+	intRangeBandRegexp = regexp.MustCompile(`^(\d+)-(\d+)$`)
+)
+
+// Validate returns if LoadBalancedWebServiceConfig is configured correctly.
+func (l LoadBalancedWebServiceConfig) Validate() error {
```
efekarakus: Can we make all receivers pointers so that the structs are not copied for each invocation of `Validate` 🙏

internal/pkg/manifest/validate.go
```diff
@@ -0,0 +1,388 @@
...
+	var err error
+	if err = l.ImageConfig.Validate(); err != nil {
```
efekarakus, line 21: hmmm 💭 I think this flow overall makes sense but I'm worried a little about unit testing.
Ideally we should be able to do this in the tests:
```go
lbws := &LoadBalancedWebServiceConfig{
   ImageConfig: mockValidate.EXPECT().Validate().Return(errors.New("some error"),
}

require.EqualError(t, lbws.Validate(), `validate "image": some error`) 
```
i.e. the unit tests for the parent field should only test if we are wrapping the errors with the field name. The leaf field unit tests should validate the actual error messages.

Currently, the tests are too coupled. If we change `ImageConfig.Validate` then we have to also update every single parent unit tests.

How can we get to decouple them? Maybe there is no way but we should at least in our unit tests only test `validate "image"` and not the sub error does that make sense?

internal/pkg/manifest/errors.go
iamhopaul123: im gonna still keep the min/max error since we are using this error in two places.

internal/pkg/manifest/validate.go
iamhopaul123: OK done!

internal/pkg/manifest/errors.go
```diff
@@ -50,3 +53,40 @@ func (e *ErrUnknownProvider) Is(target error) bool {
...
+	secondField string
+	mustExist   bool
+}
+
+func (e *errFieldMutualExclusive) Error() string {
+	var s string
+	if e.mustExist {
+		s = " not none,"
+	}
+	return fmt.Sprintf(`must specify one, not both,%s of "%s" and "%s"`, s, e.firstField, e.secondField)
```
efekarakus: My pea brain is having difficulty parsing this sentence with `mustExist` 😂 

What do you think of creating a new error type instead
```go
type errFieldsMissing struct {
  requiredFields []string
}
func (e *errFieldsMissing) Error() string { return "must specify one of ..." }
```

internal/pkg/manifest/storage.go
```diff
@@ -154,7 +153,10 @@ func (e *EFSVolumeConfiguration) unsetUIDConfig() {
 
 func (e *EFSVolumeConfiguration) isValid() error {
 	if !e.EmptyBYOConfig() && !e.EmptyUIDConfig() {
-		return errInvalidEFSConfiguration
+		return &errFieldMutualExclusive{
+			firstField:  "uid/gid",
+			secondField: "id/root_dir/auth",
```
efekarakus, line 158: Do think we should take a list for the inputs instead?
`"cannot specify uid, gid with any of id, root_dir, auth"`

internal/pkg/manifest/validate_test.go
```diff
@@ -0,0 +1,423 @@
...
+				ImageConfig: ImageWithPortAndHealthcheck{
+					ImageWithPort: ImageWithPort{
+						Image: Image{
+							Build:    BuildArgsOrString{BuildString: aws.String("mockBuild")},
+							Location: aws.String("mockLocation"),
+						},
+					},
+				},
+			},
+			wantedErrorPrefix: `validate "image": `,
```
efekarakus, line 39: nice!

efekarakus (commented):
Looks really good! just some error message suggestion and then :shipit:

internal/pkg/manifest/errors.go
iamhopaul123: `errFieldsMissing` it seems to be missing all the fields in it 🤔  How do we differentiate it from "must specify all of"? Maybe i'll make the code more readable instead!

internal/pkg/manifest/storage.go
iamhopaul123: Hmm I think I still prefer having
`must specify one, not both, of "uid/gid" and "id/root_dir/auth"`
which makes the field stand out and more explicit to users? However, this does remind me that we need to make sure they specify both `uid` and `gid`.

internal/pkg/manifest/errors.go
```diff
@@ -50,3 +53,39 @@ func (e *ErrUnknownProvider) Is(target error) bool {
...
+
+type errFieldMutualExclusive struct {
+	firstField  string
+	secondField string
+	mustExist   bool
+}
+
+func (e *errFieldMutualExclusive) Error() string {
+	if e.mustExist {
+		return fmt.Sprintf(`must specify one, not both, not none, of "%s" and "%s"`, e.firstField, e.secondField)
```
efekarakus:
```suggestion
		return fmt.Sprintf(`must specify one of "%s" and "%s"`, e.firstField, e.secondField)
```

efekarakus (approved):

internal/pkg/manifest/errors.go
```diff
@@ -50,3 +53,39 @@ func (e *ErrUnknownProvider) Is(target error) bool {
...
+	missingField      string
+	conditionalFields []string
+}
+
+func (e *errFieldMustBeSpecified) Error() string {
+	errMsg := fmt.Sprintf(`"%s" must be specified`, e.missingField)
+	if len(e.conditionalFields) == 0 {
+		return errMsg
+	}
+	return fmt.Sprintf(`%s if "%s" %s specified`, errMsg, strings.Join(e.conditionalFields, `", "`),
```
Lou1415926: We can maybe utilize `english.WordSeries` for the `conditionalFields` as well.

internal/pkg/manifest/storage.go
```diff
@@ -11,8 +11,7 @@ import (
 )
 
 var (
-	errUnmarshalEFSOpts        = errors.New(`cannot unmarshal efs field into bool or map`)
-	errInvalidEFSConfiguration = errors.New(`must specify one, not both, of "uid/gid" and "id/root_dir/auth"`)
+	errUnmarshalEFSOpts = errors.New(`cannot unmarshal efs field into bool or map`)
```
Lou1415926:
```suggestion
	errUnmarshalEFSOpts = errors.New(`cannot unmarshal "efs" field into bool or map`)
```

It seems from `manifes/errors.go` that we add quotations around fields names.

internal/pkg/manifest/svc.go
```diff
@@ -47,7 +47,10 @@ func (r *Range) IsEmpty() bool {
 // Parse extracts the min and max from RangeOpts
 func (r *Range) Parse() (min int, max int, err error) {
 	if r.Value != nil && !r.RangeConfig.IsEmpty() {
-		return 0, 0, errInvalidRangeOpts
+		return 0, 0, &errFieldMutualExclusive{
+			firstField:  "range",
+			secondField: "min/max",
```
Lou1415926: should this be "min/max/spot_from"?

Lou1415926: Also, I think `Range` is supposed to be a composite field that takes either a string, or a config. I'm not sure if this error is possible at all.

internal/pkg/manifest/validate.go
```diff
@@ -0,0 +1,410 @@
...
+	}
+	return &errMinGreaterThanMax{
+		min: min,
+		max: max,
+	}
+}
+
+// Validate returns if RangeConfig is configured correctly.
+func (r *RangeConfig) Validate() error {
+	min, max := aws.IntValue(r.Min), aws.IntValue(r.Max)
```
Lou1415926, line 277: Should we also make sure that if `min` is specified, then `max` must be specified too?

internal/pkg/manifest/svc.go
```diff
@@ -206,12 +209,18 @@ func (a *AdvancedCount) hasAutoscaling() bool {
 func (a *AdvancedCount) IsValid() error {
```
Lou1415926, line 201: i wonder if we still need this function given that we have validated everything?

internal/pkg/manifest/validate.go
```diff
@@ -0,0 +1,410 @@
...
+			firstField:  "spot",
+			secondField: "range/cpu_percentage/memory_percentage/requests/response_time",
+		}
+	}
+	if !a.Range.IsEmpty() {
+		if err := a.Range.Validate(); err != nil {
+			return fmt.Errorf(`validate "range": %w`, err)
+		}
+	}
+	if a.CPU != nil || a.Memory != nil || a.Requests != nil || a.ResponseTime != nil {
```
```suggestion
	if a.Range.IsEmpty() && (a.CPU != nil || a.Memory != nil || a.Requests != nil || a.ResponseTime != nil  ){
```

Otherwise, when `a.Range` is not empty and is valid, this error would still be returned while we shouldn't.

internal/pkg/manifest/validate.go
```diff
@@ -0,0 +1,410 @@
...
+func (a *AdvancedCount) Validate() error {
+	if a.Spot != nil && a.hasAutoscaling() {
+		return &errFieldMutualExclusive{
+			firstField:  "spot",
+			secondField: "range/cpu_percentage/memory_percentage/requests/response_time",
+		}
+	}
+	if !a.Range.IsEmpty() {
+		if err := a.Range.Validate(); err != nil {
+			return fmt.Errorf(`validate "range": %w`, err)
```
Lou1415926, line 225: I wonder if we need to wrap the error. Feels like the error message returned `Validate()` should be informative enough already?

internal/pkg/manifest/validate.go
```diff
@@ -0,0 +1,410 @@
...
+	}
+	if err = t.Storage.Validate(); err != nil {
+		return fmt.Errorf(`validate "storage": %w`, err)
+	}
+	return nil
+}
+
+// Validate returns if PlatformArgsOrString is configured correctly.
+func (p *PlatformArgsOrString) Validate() error {
+	if !p.PlatformArgs.isEmpty() {
```
Lou1415926, line 195: Should this be 
```
return p.PlatformArgs.Validate()
```
similar to `BuildArgsOrString`?

internal/pkg/manifest/validate.go
```diff
@@ -0,0 +1,410 @@
...
+
+// Validate returns if PlatformArgsOrString is configured correctly.
+// TODO: add validation once "feat/pencere" is merged.
+func (*PlatformArgs) Validate() error {
+	return nil
+}
+
+// Validate returns if Count is configured correctly.
+func (c *Count) Validate() error {
+	if !c.AdvancedCount.IsEmpty() {
```
Lou1415926, line 209: Same here. Since `count` is composite, can this be 
```
return c.AdvancedCount.Validate()
```
just like `BuildArgsOrString`?

internal/pkg/manifest/validate.go
```diff
@@ -0,0 +1,410 @@
...
+			missingField:      "range",
+			conditionalFields: []string{"cpu_percentage", "memory_percentage", "requests", "response_time"},
+		}
+	}
+	return nil
+}
+
+// Validate returns if Range is configured correctly.
+func (r *Range) Validate() error {
+	if !r.RangeConfig.IsEmpty() {
```
Lou1415926, line 238: Samesies here!

internal/pkg/manifest/validate.go
```diff
@@ -0,0 +1,410 @@
...
+}
+
+// Validate returns if MountPointOpts is configured correctly.
+func (*MountPointOpts) Validate() error {
+	return nil
+}
+
+// Validate returns if EFSConfigOrBool is configured correctly.
+func (e *EFSConfigOrBool) Validate() error {
+	if !e.Advanced.IsEmpty() {
```
Lou1415926, line 325: Samesies here!

internal/pkg/manifest/validate.go
```diff
@@ -0,0 +1,410 @@
...
+	if e.UID == nil && e.GID != nil {
+		return &errFieldMustBeSpecified{
+			missingField:      "uid",
+			conditionalFields: []string{"gid"},
+		}
+	}
+	if err := e.AuthConfig.Validate(); err != nil {
+		return fmt.Errorf(`validate "auth": %w`, err)
+	}
+	if aws.StringValue(e.AuthConfig.AccessPointID) != "" {
```
Lou1415926: I wonder if `aws.StringValue(e.AuthConfig.AccessPointID) != ""` could ever be true. What would the `manifest.yml` be like that `AccessPointID` is marshaled as `""`?

Same for `RootDirectory`.

internal/pkg/manifest/validate.go
```diff
@@ -0,0 +1,410 @@
...
+		if (aws.StringValue(e.RootDirectory) == "" || aws.StringValue(e.RootDirectory) == "/") &&
+			aws.BoolValue(e.AuthConfig.IAM) {
+			return nil
+		}
+		return fmt.Errorf("root_dir must be either empty or / and auth.iam must be true when access_point_id is in used")
```
```suggestion
		return fmt.Errorf(`"root_dir" must be either empty or "/" and "auth.iam" must be true when "access_point_id" is in used`)
```

internal/pkg/manifest/validate.go
```diff
@@ -0,0 +1,410 @@
...
+		if err := mp.Validate(); err != nil {
+			return fmt.Errorf(`validate "mount_points[%d]: %w`, ind, err)
+		}
+	}
+	return s.ImageOverride.Validate()
+}
+
+// Validate returns if NetworkConfig is configured correctly.
+func (n *NetworkConfig) Validate() error {
+	if !n.VPC.isEmpty() {
```
Lou1415926, line 386: I wonder if we need to check `isEmpty` before calling `Validate()`? Can the empty case be handled from within `Validate()`?

Lou1415926 (commented):
LGTM! Just a <s>few</s> plethora of questions!

internal/pkg/manifest/errors.go
iamhopaul123: That's a good idea!

internal/pkg/manifest/svc.go
iamhopaul123: Right right we should get rid of this.

internal/pkg/manifest/svc.go
iamhopaul123: That's the next step when i consolidate every validation else where!

internal/pkg/manifest/validate.go
iamhopaul123: Good call!

internal/pkg/manifest/validate.go
iamhopaul123: The wrapping rule is based on how it is formatted in our manifest:
```yaml
type AdvancedCount struct {
	Spot         *int           `yaml:"spot"` // mutually exclusive with other fields
	Range        Range          `yaml:"range"`
	CPU          *int           `yaml:"cpu_percentage"`
	Memory       *int           `yaml:"memory_percentage"`
	Requests     *int           `yaml:"requests"`
	ResponseTime *time.Duration `yaml:"response_time"`
}
```

internal/pkg/manifest/validate.go
iamhopaul123: The former is incorrect not this one. Will address.

internal/pkg/manifest/validate.go
iamhopaul123: Discussed offline

Lou1415926 (approved):
:shipit:

Issue #2686, opened 2021-08-03
copilot deploy hangs if no [default] credential is set


tiivik:
I had accidentally commented out the default region in my AWS credentials like so:
```
#[default]
aws_access_key_id=key-here
aws_secret_access_key=key-here
```
Then, running `copilot deploy --name api --env test` the command just hangs, does not error out or provide any context of why the deploy was hanging.

My mistake ofc, but could be useful if an error message was displayed.

Lou1415926:
Hi @tiivik ! Thanks for bringing this to our attention. It indeed hangs when credential is commented out while region is not. We will look into this and see how we can improve the error messaging!

tiivik:
> Hi @tiivik ! Thanks for bringing this to our attention. It indeed hangs when credential is commented out while region is not. We will look into this and see how we can improve the error messaging!

Thank you, appreciate it!

efekarakus:
Related https://github.com/aws/copilot-cli/issues/2731

efekarakus:
Related https://github.com/aws/copilot-cli/issues/2547

efekarakus:
This fix is now released in https://github.com/aws/copilot-cli/releases/tag/v1.11.0! 🎉

PR #2816, opened 2021-09-08
fix: error out if the credential is invalid

Lou1415926:
Previously, we don't error out if the credential is invalid or we fail to get credentials because of time out. 

A result of this is that we will keep using the session with the **invalid/empty credential**, which eventually causes the CLI to hang during a call to `GetCallerIdentity()` https://github.com/aws/copilot-cli/blob/77983421537d9f34575655ab8cba309f30610f33/internal/pkg/aws/identity/identity.go#L40

The hanging issue is reported in #2686 and #1419.

An  **invalid/empty credential** happens when:
1. The profile is not found, AND
2. The environment variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are NOT set.

The profile is not found when:
1. The file `~/.aws/credentials` is not found
2. The file is found, but the profile is not defined

After this fix, the output from `copilot app init` or `copilot init` in the case of  an **invalid/empty credential** would be:
```
Welcome to the Copilot CLI! We're going to walk you through some questions
to help you get set up with a containerized application on AWS. An application is a collection of
containerized services that operate together.

✘ ask app init: get credentials of session: RequestCanceled: request context canceled
caused by: context deadline exceeded
```

instead of hanging.

### DIFF
--- a/internal/pkg/cli/app_init.go
+++ b/internal/pkg/cli/app_init.go
@@ -104,7 +104,12 @@ func (o *initAppOpts) Validate() error {
 
 // Ask prompts the user for any required arguments that they didn't provide.
 func (o *initAppOpts) Ask() error {
-	if ok, _ := o.isSessionFromEnvVars(); ok { // Ignore the error, we do not want to crash for a warning.
+	ok, err := o.isSessionFromEnvVars()
+	if err != nil {
+		return err
+	}
+
+	if ok {
 		log.Warningln(`Looks like you're creating an application using credentials set by environment variables.
 Copilot will store your application metadata in this account.
 We recommend using credentials from named profiles. To learn more:


internal/pkg/cli/app_init.go
```diff
@@ -104,7 +104,12 @@ func (o *initAppOpts) Validate() error {
 
 // Ask prompts the user for any required arguments that they didn't provide.
 func (o *initAppOpts) Ask() error {
-	if ok, _ := o.isSessionFromEnvVars(); ok { // Ignore the error, we do not want to crash for a warning.
+	ok, err := o.isSessionFromEnvVars()
```
efekarakus, line 107: 🤣 awesome find! (and sorry for the bug :disappear:)

efekarakus (approved):

iamhopaul123 (approved):
Nice :shipit:  Also since we were ignoring the error, this seems more like a bug fixing to me. Can we edit the title/commit accordingly?

Issue #2731, opened 2021-08-10
Bug: copilot CLI fails if no default region found in profile


foragerr:
This is the same issue as #1115 , It seems to've reappeared in a recent version.

```
$ copilot --version
copilot version: v1.9.0

$ copilot env init
✘ ask for application: list applications: list applications: MissingRegion: could not find region configuration
```

Setting default region in profile, or setting `$AWS_REGION` will work
but copilot-cli no longer prompts for region as suggested in the the other issue thread

iamhopaul123:
Hello @foragerr. Glad you found the workaround! Yes this is inconsistent with the other issue thread and we should fix this behavior. Thank you for reporting the issue!

Issue #1115, opened 2020-07-10
Bug: copilot CLI fails if no default region found in profile


shrasool:
I am trying to set up copilot init demo app using the interactive CLI but I am unable as the default AWS profile does not have region specified (an optional field) so the init command fails.

```
hamidrasool@LAPTOP:~/copilot-demo-app$ copilot init
Note: It's best to run this command in the root of your Git repository.
Welcome to the Copilot CLI! We're going to walk you through some questions
to help you get set up with an application on ECS. An application is a collection of
containerized services that operate together.

Application name: copilot-demo
✘ get application copilot-demo: get application copilot-demo: MissingRegion: could not find region configuration
```

The missing feature in this case is a setting for --region parameter and optionally adding a prompt to the interactive init setup if it is not specified. Currently there is no such flag available, and surprised there isn't any planned in the roadmap.

```
hamidrasool@LAPTOP:~/copilot-demo-app$ copilot init --region eu-west-1
✘ unknown flag: --region
```

Current workaround is to add the region manually into `~/.aws/config` if needed to deploy for any particular region which should not be needed as it forces user to change their default settings for profile. Alternately setting a new profile for this purpose for each required region is equally not a pleasant user experience.

System tested on: Ubuntu 18.04 on WSL2

efekarakus:
Thanks @shrasool ! Glad you found the workarounds to get it working for now, we have a story (#36) to track permission related issues and we'll updated that issue as we improve the UX for our environment and application initialization!

efekarakus:
Heya @shrasool we released [v0.3.0](https://github.com/aws/copilot-cli/releases/tag/v0.3.0) which now prompts for a region if we don't detect one!

serverlesspolska:
Hi Guys,

I'm new to that tool however it fails for me due to `MissingRegion`.

I checked out the repo https://github.com/aws-samples/aws-copilot-sample-service.git and run `copilot init` which asked me for app name, and then it failed
```
copilot init
Welcome to the Copilot CLI! We're going to walk you through some questions
to help you get set up with an application on ECS. An application is a collection of
containerized services that operate together.

Application name: demo
✘ get application demo: get application demo: MissingRegion: could not find region configuration
```

I deliberately don't use the default region in my AWS config file and don't plan to start (as it is easy to execute to commend against the default region instead of intended one).

@efekarakus  how can I specify the region?

Also this fails as well
```
copilot env init
[...]

PR #2819, opened 2021-09-10
chore(cli): error out early and log friendly text if AWS region information is missing

Lou1415926:
This PR is addressing the experience described in #2731 and #1115. 


internal/pkg/aws/sessions/sessions.go
```diff
@@ -29,6 +30,8 @@ const (
 	clientTimeout                   = 30 * time.Second
 )
 
+var ErrMissingRegion = errors.New("missing region configuration")
+
 // Provider provides methods to create sessions.
 // Once a session is created, it's cached locally so that the same session is not re-created.
 type Provider struct {
```
efekarakus: I haven't read this completely ! so feel free to ignore the comment

I wonder if it would simplify anything if we do something like this:

```go
type Provider struct {
  ...
  logger func(format string, args ...interface{})
}


func NewProvider() *Provider {
  ...
  logger: log.Infof
}
```
and log the missing info stuff here. This way in the unit test we can have a logger that does actually nothing.

internal/pkg/aws/sessions/sessions.go
```diff
@@ -29,6 +30,8 @@ const (
 	clientTimeout                   = 30 * time.Second
 )
 
+var ErrMissingRegion = errors.New("missing region configuration")
```
iamhopaul123: nit: missing comment here

internal/pkg/aws/sessions/sessions.go
```diff
@@ -169,3 +179,7 @@ func userAgentHandler() request.NamedHandler {
 		},
 	}
 }
+
+func isRegionMissing(sess *session.Session) bool {
+	return sess.Config.Region == nil || *sess.Config.Region == ""
```
iamhopaul123:
```suggestion
	return aws.StringValue(sess.Config.Region) == ""
```
Maybe we don't need a function for this logic?

internal/pkg/cli/cli.go
```diff
@@ -130,3 +132,18 @@ func logRecommendedActions(actions []string) {
 		log.Infof("%s %s\n", prefix, followup)
 	}
 }
+
+func logFriendlyTextIfRegionIsMissing(err error) {
+	if !isMissingRegionErrorInChain(err) {
+		return
+	}
+	log.Infof(`It looks like your AWS region configuration is missing.
```
huanjani: So friendly! 😄 
Do we want to mention `aws configure` as an option, too, or too much?

iamhopaul123 (commented):
LGTM overall! Just two nit comments.

internal/pkg/aws/sessions/sessions.go
Lou1415926: I feel like the logging should still be happening from `cli` package. `NewProvider()` is called all over the place, for example `secretsmanager` and `config` packages. Whether we want the `logger` to be a `log.Infof` or a no-op only depends on `cli` though.  The result of this is that `secretsmanager` and `config`, being sometimes a dependency of `cli`, would then need to pass `cli`'s wills to `sessions` that whether the provider should have a `log.Infof` or a no-op as logger.

Lou1415926: We only log just before we return the error from `cli` - which means that we only log in the unit test if we are checking against the error. 

Expectedly, the log never appears in the unit test because we mock the provider such that the error wouldn't occur.

internal/pkg/cli/cli.go
Lou1415926: I sorta feel like a user at this point would have already been aware of `aws configure`? And using `aws configure` is the same as the first suggestion (modifying "~/.aws/config" file), I feel like it's probably more straightforward + transparent to tell them which specific file to modify directly than telling them to run a command that eventually modifies the file 🤔  Just my thoughts!

huanjani (approved):
:shipit:

internal/pkg/aws/sessions/sessions.go
```diff
@@ -59,6 +63,10 @@ func (p *Provider) Default() (*session.Session, error) {
 	if err != nil {
 		return nil, err
 	}
+	if aws.StringValue(sess.Config.Region) == "" {
+		return sess, ErrMissingRegion
```
efekarakus: hmmm, this is a bit unusual. The session itself is valid if the error is `ErrMissingRegion`. This breaks the convention of the first return argument being the zero value if there is an error, and we end up using the first arg in the `FromRole` method.

What do you think of instead:
1. This method should return `nil, ErrMissingRegion`.
2. In `FromRole` instead we don't call `p.Default` and instead write separately:
```
sess, err := session.NewSessionWithOptions(session.Options{
		Config:            *newConfig(),
		SharedConfigState: session.SharedConfigEnable,
})
```

internal/pkg/cli/cli.go
```diff
@@ -130,3 +132,18 @@ func logRecommendedActions(actions []string) {
...
+- We recommend including your region configuration in the "~/.aws/config" file.
+- Alternatively, you can run %s to set the environment variable.
+More information: https://aws.github.io/copilot-cli/docs/credentials/
+`, color.HighlightCode("export AWS_REGION=<application region>"))
+}
+
+func isMissingRegionErrorInChain(err error) bool {
+	return errors.As(err, &sessions.ErrMissingRegion)
```
efekarakus: oh interesting wouldn't `errors.Is(err, sessions.ErrMissingRegion)` also work?

internal/pkg/aws/sessions/sessions.go
efekarakus: I still think it's slightly problematic to have the logging in the `cli` pkg, because the moment we work on a new command we have to remember to add this log statement.

We have a few alternatives, but what do you think of this:
```go
package sessions

type ErrMissingRegion string
func (e *ErrMissingRegion) Error() string { return string(e) } // implements the error interface.

func (e *ErrMissingRegion) RecommendAction() string {  // implements new actionRecommender interface.
    return `It looks like your AWS region configuration is missing.
    ...`
}
```

Then in our main function for the binary: https://github.com/aws/copilot-cli/blob/28e91b9dc55fd0feed8bdf665679471843aa40d1/cmd/copilot/main.go#L25-L28

we do:

```go
package main
type actionRecommender interface {
   RecommendAction() string
}


if err := cmd.Execute(); err != nil {
  var ar actionRecommender
  if errors.As(err, &ar) {
    log.Infoln(ar.RecommendAction())
  }
  log.Errorln(err.Error())
  os.Exit(1)
}
```

efekarakus: I thinks this pattern would also scale well to other errors

efekarakus (approved):

PR #2821, opened 2021-09-10
fix(manifest): remove unnecessary `allowed_workers` field

efekarakus:
There is no way today to limit an SNS topic subscription to be based on the tag of SQS queues (https://aws.amazon.com/premiumsupport/knowledge-center/sqs-sns-subscribe-cloudformation/). Instead, by default SNS will limit subscriptions to be only within the same AWS account.  

Therefore, the `allowed_workers` field is not needed. This is not a breaking change, it makes a required field now obsolete. 


internal/pkg/template/templates/workloads/partials/cf/publish.yml
```diff
@@ -8,33 +8,23 @@
       TopicName: !Sub '${AWS::StackName}-{{logicalIDSafe $topic.Name}}'
       KmsMasterKeyId: 'alias/aws/sns'
 
-{{- if $topic.AllowedWorkers}}
 {{logicalIDSafe $topic.Name}}SNSTopicPolicy:
-    Metadata:
-      'aws:copilot:description': 'A SNS topic policy to allow {{wordSeries $topic.AllowedWorkers "and"}} {{pluralWord (len $topic.AllowedWorkers) "service" "services"}} to consume {{$topic.Name}} events'
```
iamhopaul123, line 14: Why do we want to remove this metadata completely?

internal/pkg/template/templates/workloads/partials/cf/publish.yml
```diff
@@ -8,33 +8,23 @@
...
               StringEquals:
-                'sqs:ResourceTag/copilot-application': !Sub '${AppName}'
-                'sqs:ResourceTag/copilot-environment': !Sub '${EnvName}'
-              ForAnyValue:StringEquals:
-                'sqs:ResourceTag/copilot-service': 
-                {{- range $worker := $topic.AllowedWorkers}}
-                  - '{{$worker}}'
-                {{- end}}
-  {{- end}}
+                "sns:Protocol": "sqs"
```
iamhopaul123, line 28: 🤦

iamhopaul123 (approved):

bvtujo (approved):

internal/pkg/template/templates/workloads/partials/cf/publish.yml
efekarakus: I opted to remove it because we're not doing anything interesting with it (like using tags), and the resource creation is v fast.

PR #2822, opened 2021-09-10
feat: add container healthcheck for sidecars

### LINKED ISSUE #2723: Dependency ordering being ignored
I'm using version 1.9.0 and am trying to have a dependency condition of "HEALTHY" for my main container on a sidecar container (for a scheduled job). The task definition being generated doesn't include any dependency ordering. Is there something I'm missing?

mainfiest for main container:
```yaml
depends_on:
  content: HEALTHY

sidecars:
  content:
    image: ...
```

iamhopaul123:
fixes #2723.

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -18,7 +18,6 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
-	"github.com/aws/aws-sdk-go/service/ecs"
 	"github.com/aws/copilot-cli/internal/pkg/manifest"
 	"github.com/aws/copilot-cli/internal/pkg/template"
 )
@@ -126,14 +125,14 @@ func convertSidecar(s convertSidecarOpts) ([]*template.SidecarOpts, error) {
 	return sidecars, nil
 }
 
-func convertContainerHealthCheck(hc manifest.ContainerHealthCheck) *ecs.HealthCheck {
+func convertContainerHealthCheck(hc manifest.ContainerHealthCheck) *template.ContainerHealthCheck {
 	if hc.IsEmpty() {
 		return nil
 	}
 	// Make sure that unset fields in the healthcheck gets a default value.
 	hc.ApplyIfNotSet(manifest.NewDefaultContainerHealthCheck())
-	return &ecs.HealthCheck{
-		Command:     aws.StringSlice(hc.Command),
+	return &template.ContainerHealthCheck{
+		Command:     hc.Command,
 		Interval:    aws.Int64(int64(hc.Interval.Seconds())),
 		Retries:     aws.Int64(int64(aws.IntValue(hc.Retries))),
 		StartPeriod: aws.Int64(int64(hc.StartPeriod.Seconds())),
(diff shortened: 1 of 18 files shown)


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -118,12 +119,28 @@ func convertSidecar(s convertSidecarOpts) ([]*template.SidecarOpts, error) {
...
 			DependsOn:    config.DependsOn,
 			EntryPoint:   entrypoint,
+			HealthCheck:  convertContainerHealthCheck(config.HealthCheck),
 			Command:      command,
 		})
 	}
 	return sidecars, nil
 }
 
+func convertContainerHealthCheck(hc manifest.ContainerHealthCheck) *ecs.HealthCheck {
```
efekarakus: I know that this is a legacy signature, but it's a bit odd that this `convert` function returns a type from the `aws-sdk` instead of `template`.

What do you think of redefining HealthCheck in `template` with the exact same signature as here?

internal/pkg/manifest/workload.go
```diff
@@ -408,15 +407,16 @@ func (lc *Logging) GetEnableMetadata() *string {
...
+	Port          *string              `yaml:"port"`
+	Image         *string              `yaml:"image"`
+	Essential     *bool                `yaml:"essential"`
+	CredsParam    *string              `yaml:"credentialsParameter"`
+	Variables     map[string]string    `yaml:"variables"`
+	Secrets       map[string]string    `yaml:"secrets"`
+	MountPoints   []SidecarMountPoint  `yaml:"mount_points"`
+	DockerLabels  map[string]string    `yaml:"labels"`
+	DependsOn     map[string]string    `yaml:"depends_on"`
+	HealthCheck   ContainerHealthCheck `yaml:"healthcheck"`
```
efekarakus, line 416: Should we add this now into:
https://github.com/aws/copilot-cli/blob/ecc3cf92643630cc1cf741cb5bf05b3e1e335472/internal/pkg/manifest/validate.go#L520-L527

efekarakus (commented):
Awesome, we also need to update the docs but we can do it in a separate PR if you'd prefer!

internal/pkg/deploy/cloudformation/stack/transformers.go
iamhopaul123: Agree that makes sense to me

internal/pkg/manifest/validate.go
```diff
@@ -523,6 +523,9 @@ func (s *SidecarConfig) Validate() error {
 			return fmt.Errorf(`validate "mount_points[%d]: %w`, ind, err)
 		}
 	}
+	if err := s.HealthCheck.Validate(); err != nil {
+		return fmt.Errorf(`validate "healthcheck": %w`, err)
+	}
```
efekarakus, line 528: can we add a prefix test for this 🥺

efekarakus (approved):

internal/pkg/manifest/validate.go
iamhopaul123: It's not possible yet boss, since the function doesn't return any error.

iamhopaul123: I'll add it once I add more tests to the boilerplate.

Lou1415926 (approved):
:shipit:

Issue #2818, opened 2021-09-09
[Design] Manifest Validation


iamhopaul123:
## Background

Right now even if the manifest is malformed, for example, some incorrect indentations (#2618) (https://github.com/aws/copilot-cli/issues/2618), we won't error out. Or they may remove required fields in the manifest without adding them back (#2689 (https://github.com/aws/copilot-cli/issues/2689)).

From users' perspective, this could be a pain because users might not be aware of the fact that their settings will not be rendered or rendered incorrectly. Whereas from Copilot developers' perspective, it is also paper cut for us because without having a comprehensive manifest validation means we need to constantly check the validity whenever we use their values.

## Implementation Plan

- [x] Add manifest validation
- [x] Consolidate existing validations elsewhere (e.g., `template` pkg) into the new manifest validation.
- [x] Add appropriate CI to make sure the comprehensiveness for our manifest validation.
- [x] Update the doc with the latest manifest validation for each field.

PR #2826, opened 2021-09-13
chore(manifest): add validate boilerplate for rest workload types

iamhopaul123:
Part of #2818.

After this PR, we should have boilerplate completed for every workload type. Then, we'll add more advanced basic types validation as well as start to consolidate validation elsewhere.

### DIFF
--- a/internal/pkg/manifest/validate.go
+++ b/internal/pkg/manifest/validate.go
@@ -15,7 +15,7 @@ var (
 	intRangeBandRegexp = regexp.MustCompile(`^(\d+)-(\d+)$`)
 )
 
-// Validate returns if LoadBalancedWebServiceConfig is configured correctly.
+// Validate returns nil if LoadBalancedWebServiceConfig is configured correctly.
 func (l *LoadBalancedWebServiceConfig) Validate() error {
 	var err error
 	if err = l.ImageConfig.Validate(); err != nil {
@@ -52,7 +52,7 @@ func (l *LoadBalancedWebServiceConfig) Validate() error {
 	return nil
 }
 
-// Validate returns if BackendServiceConfig is configured correctly.
+// Validate returns nil if BackendServiceConfig is configured correctly.
 func (b *BackendServiceConfig) Validate() error {
 	var err error
 	if err = b.ImageConfig.Validate(); err != nil {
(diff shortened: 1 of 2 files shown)


internal/pkg/manifest/validate.go
```diff
@@ -44,9 +44,132 @@ func (l *LoadBalancedWebServiceConfig) Validate() error {
...
+	return nil
+}
+
+// Validate returns if BackendServiceConfig is configured correctly.
+func (b *BackendServiceConfig) Validate() error {
+	var err error
+	if err = b.ImageConfig.Validate(); err != nil {
+		return fmt.Errorf(`validate "image": %w`, err)
+	}
+	if err = b.ImageOverride.Validate(); err != nil {
```
huanjani, line 61: Just curious why some errors are wrapped and others (here and below) aren't.

internal/pkg/manifest/validate.go
```diff
@@ -44,9 +44,132 @@ func (l *LoadBalancedWebServiceConfig) Validate() error {
...
-			return fmt.Errorf(`validate taskdef_overrides[%d]: %w`, i, err)
+	for ind, taskDefOverride := range l.TaskDefOverrides {
+		if err = taskDefOverride.Validate(); err != nil {
+			return fmt.Errorf(`validate taskdef_overrides[%d]: %w`, ind, err)
+		}
+	}
+	return nil
+}
+
+// Validate returns if BackendServiceConfig is configured correctly.
```
huanjani: nit: these comments (here and below) make it sound like these methods return bools. Maybe: 
```suggestion
// Validate returns nil if BackendServiceConfig is configured correctly.
```
or some other rewording?

internal/pkg/manifest/validate.go
iamhopaul123: It depends on if the struct is `inlined` or not (for better error msg helping them to debug)

internal/pkg/manifest/validate.go
iamhopaul123: good idea!

internal/pkg/manifest/validate.go
```diff
@@ -95,25 +224,28 @@ func (i *Image) Validate() error {
 	return nil
 }
 
-// Validate returns if BuildArgsOrString is configured correctly.
+// Validate returns nil if BuildArgsOrString is configured correctly.
 func (b *BuildArgsOrString) Validate() error {
+	if b.isEmpty() {
```
efekarakus, line 229: nice! i like this change of moving it inside, simplifies it for the callers.

efekarakus (approved):

huanjani (approved):
LGTM!

PR #2827, opened 2021-09-13
feat(manifest): add `queue_delay` for autoscaling worker services

### LINKED ISSUE #2796: proposal: Autoscaling for Worker Services

efekarakus:
Resolves #2796 

### DIFF
--- a/internal/pkg/manifest/svc.go
+++ b/internal/pkg/manifest/svc.go
@@ -6,6 +6,7 @@ package manifest
 
 import (
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
 	"time"
@@ -179,7 +180,7 @@ type AdvancedCount struct {
 	Memory       *int           `yaml:"memory_percentage"`
 	Requests     *int           `yaml:"requests"`
 	ResponseTime *time.Duration `yaml:"response_time"`
-	QueueScaling queueScaling   `yaml:"queue_delay"`
+	QueueScaling QueueScaling   `yaml:"queue_delay"`
 }
 
 // IsEmpty returns whether AdvancedCount is empty.
@@ -224,27 +225,28 @@ func (a *AdvancedCount) unsetAutoscaling() {
 	a.Memory = nil
 	a.Requests = nil
 	a.ResponseTime = nil
-	a.QueueScaling = queueScaling{}
+	a.QueueScaling = QueueScaling{}
 }
 
-// queueScaling represents the configuration to scale a service based on a SQS queue.
-type queueScaling struct {
+// QueueScaling represents the configuration to scale a service based on a SQS queue.
+type QueueScaling struct {
 	AcceptableLatency *time.Duration `yaml:"acceptable_latency"`
 	AvgProcessingTime *time.Duration `yaml:"msg_processing_time"`
 }
 
-// IsEmpty returns true if the queueScaling is set.
-func (qs *queueScaling) IsEmpty() bool {
+// IsEmpty returns true if the QueueScaling is set.
+func (qs *QueueScaling) IsEmpty() bool {
 	return qs.AcceptableLatency == nil && qs.AvgProcessingTime == nil
 }
 
 // AcceptableBacklogPerTask returns the total number of messages that each task can accumulate in the queue
 // while maintaining the AcceptableLatency given the AvgProcessingTime.
-func (qs *queueScaling) AcceptableBacklogPerTask() (int, error) {
+func (qs *QueueScaling) AcceptableBacklogPerTask() (int, error) {
 	if err := qs.Validate(); err != nil {
 		return 0, err
 	}
-	return 0, nil
+	v := math.Ceil(float64(*qs.AcceptableLatency) / float64(*qs.AvgProcessingTime))
+	return int(v), nil
 }
 
 // ServiceDockerfileBuildRequired returns if the service container image should be built from local Dockerfile.
--- a/internal/pkg/manifest/svc.go
+++ b/internal/pkg/manifest/svc.go
@@ -238,6 +238,15 @@ func (qs *queueScaling) IsEmpty() bool {
 	return qs.AcceptableLatency == nil && qs.AvgProcessingTime == nil
 }
 
+// AcceptableBacklogPerTask returns the total number of messages that each task can accumulate in the queue
+// while maintaining the AcceptableLatency given the AvgProcessingTime.
+func (qs *queueScaling) AcceptableBacklogPerTask() (int, error) {
+	if err := qs.Validate(); err != nil {
+		return 0, err
+	}
+	return 0, nil
+}
+
 // ServiceDockerfileBuildRequired returns if the service container image should be built from local Dockerfile.
 func ServiceDockerfileBuildRequired(svc interface{}) (bool, error) {
 	return dockerfileBuildRequired("service", svc)
--- a/internal/pkg/manifest/svc.go
+++ b/internal/pkg/manifest/svc.go
@@ -5,6 +5,7 @@
 package manifest
 
 import (
+	"errors"
 	"fmt"
 	"math"
 	"strconv"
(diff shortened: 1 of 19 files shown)


internal/pkg/manifest/validate.go
```diff
@@ -351,18 +352,51 @@ func (a *AdvancedCount) Validate() error {
...
 		}
 	}
+	if err := a.QueueScaling.Validate(); err != nil {
+		return fmt.Errorf(`validate "queue_delay": %w`, err)
+	}
+	return nil
+}
+
+// Validate returns nil if QueueScaling is configured correctly.
+func (qs *QueueScaling) Validate() error {
```
iamhopaul123, line 391: 🙏

internal/pkg/manifest/svc.go
```diff
@@ -224,6 +225,28 @@ func (a *AdvancedCount) unsetAutoscaling() {
...
+
+// IsEmpty returns true if the QueueScaling is set.
+func (qs *QueueScaling) IsEmpty() bool {
+	return qs.AcceptableLatency == nil && qs.AvgProcessingTime == nil
+}
+
+// AcceptableBacklogPerTask returns the total number of messages that each task can accumulate in the queue
+// while maintaining the AcceptableLatency given the AvgProcessingTime.
+func (qs *QueueScaling) AcceptableBacklogPerTask() (int, error) {
+	if err := qs.Validate(); err != nil {
```
iamhopaul123, line 259: Maybe add a `TODO` so that we won' forget to remove it once we finish the manifest validation

iamhopaul123 (approved):
:shipit:

internal/pkg/manifest/svc.go
efekarakus: I wasn't sure if I should keep it or not even after we have manifest validation 🤔 maybe I can add a comment saying that this method should assume that the input is already Validated.

Lou1415926: ooo we should be careful to only call `AcceptableBacklogPerTask` when `qs` is not empty. Otherwise we will panic from proceeding to dereference `qs.AcceptableLatency` and `qs.AvgProcessingTime`

internal/pkg/manifest/validate.go
```diff
@@ -351,18 +352,51 @@ func (a *AdvancedCount) Validate() error {
...
+	if qs.IsEmpty() {
```
Lou1415926: I wonder if it's better to call `Validate()` only when `!struct.IsEmpty()`, or consider `IsEmpty()` as valid and call `Validate()` whether the struct is empty or not.

internal/pkg/manifest/validate.go
```diff
@@ -351,18 +352,51 @@ func (a *AdvancedCount) Validate() error {
...
 	}
-	if a.Range.IsEmpty() && (a.CPU != nil || a.Memory != nil || a.Requests != nil || a.ResponseTime != nil) {
+	if a.Range.IsEmpty() && (a.CPU != nil || a.Memory != nil || a.Requests != nil || a.ResponseTime != nil || !a.QueueScaling.IsEmpty()) {
 		return &errFieldMustBeSpecified{
 			missingField:      "range",
-			conditionalFields: []string{"cpu_percentage", "memory_percentage", "requests", "response_time"},
+			conditionalFields: []string{"cpu_percentage", "memory_percentage", "requests", "response_time", "queue_delay"},
 		}
 	}
+	if err := a.QueueScaling.Validate(); err != nil {
```
Lou1415926: Do you think we should error out somewhere if `!a.QueueScalling.isEmpty()` while the workload is a not a worker service?

internal/pkg/manifest/svc.go
```diff
@@ -194,7 +196,7 @@ func (a *AdvancedCount) IgnoreRange() bool {
 
 func (a *AdvancedCount) hasAutoscaling() bool {
 	return !a.Range.IsEmpty() || a.CPU != nil || a.Memory != nil ||
-		a.Requests != nil || a.ResponseTime != nil
+		a.Requests != nil || a.ResponseTime != nil || !a.QueueScaling.IsEmpty()
```
Lou1415926: What if a non-worker service workload falsely specified `queue_delay`? Should we consider it as having autoscaling?

Lou1415926 (approved):
LGTM! Just a few questions and thoughts that don't beed to be addressed!

internal/pkg/manifest/validate.go
efekarakus: I opted this route because I didn't want the client to think about whether it's safe to call `Validate` or not.
This way client code becomes easy to reason about:
```
go through each field:
  call Validate()
```

internal/pkg/manifest/svc.go
efekarakus: yea 👍 that's why there is `Validate` here for now

internal/pkg/manifest/svc.go
efekarakus: ugh yea this is kindof an issue even today for the `requests` or `response_time` fields, what if a backend service only gives:
```yaml
count:
  range: 1-10
  requests: 1000
```
Then autoscaling resources will be created except the `ScalingPolicy` :| which might be a bit surprising to the users.

Let me see if I can think of a way of erroring out early 💭

internal/pkg/manifest/svc.go
Lou1415926: `Validate` returns `nil` error when `qs` is empty though! So it still seems possible to dereference a `qs. AcceptableLatency` that is `nil`, right? We should either make sure to call `AcceptableBacklogPerTask` only when appropriate or add an extra safeguard here, e.g. if it's empty, we return `0, errors.New("no queue autoscaling specified")`.

internal/pkg/manifest/svc.go
iamhopaul123: We could possibly do it in manifest validation to have this logic [here](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/manifest/validate.go#L19) 💭

iamhopaul123: Or update the Validate signature to accept an optional handler.

efekarakus: I'm adding a private field `workloadType` to `AdvancedCount` to help us narrow down which fields we should be checking

internal/pkg/manifest/validate.go
iamhopaul123: Actually I kind of get what wanxian said, if we return `nil` when the struct is empty, we cannot safely use the struct after validation because it could be an empty one. No matter we call `isEmpty` or not in `Validate()`, the caller needs to make sure the struct is not empty.

Overall I don't think we should use `Validate()` out of the `validate.go` file. Any thought?

efekarakus: I'll change it to validate only if the queue is not empty, but I think it should be okay for anybody to call the `Validate` function like in `AcceptableBacklog`.

internal/pkg/manifest/svc.go
efekarakus: OK ! @Lou1415926 and @iamhopaul123 can you both take a look at this commit for the changes? https://github.com/aws/copilot-cli/pull/2827/commits/87588b67ae6a0b7b71e23db2b213b7b51814325f 🥺

Issue #2799, opened 2021-09-01
Feature Request: Allow specifying of OutputArtifactFormat in pipeline.yml


jsutton:
Currently, copilot doesn't specify OutputArtifactFormat, so it defaults to CODE_ZIP, not CODEBUILD_CLONE_REF. This causes problems when trying to build a project that uses submodules. If this method is chosen, it would be nice to include an option to add a full clone including submodules to the buildspec.yml.

The other option would be to add a source to the codebuild project for the repository that has "Use Git submodules" selected. This has the benefit of reducing complexity on the user but increases complexity in that there are then two source checkouts.

jsutton:
For option1, it would be helpful to update the docs/templates for buildspec.yml to include the following:
```yaml
env:
  git-credential-helper: yes
```
and
```yaml
    commands:
      - echo "cd into $CODEBUILD_SRC_DIR"
      - cd $CODEBUILD_SRC_DIR
      # If your submodule does not use https, you may need to change it with a command like this
      - git config --file=.gitmodules submodule.<SubModulePath>.url https://git-codecommit.us-east-1.amazonaws.com/v1/repos/SubmoduleRepoName
      - git submodule update --init
```

Also, the pipeline policy and codebuild policy have to be updated:
```yaml
  BuildProjectPolicy:
    Type: AWS::IAM::Policy
    DependsOn: BuildProjectRole
    Properties:
      PolicyName: !Sub ${AWS::StackName}-CodeBuildPolicy
      PolicyDocument:
        Version: 2012-10-17
        Statement:
        #...omitted...
          - Effect: Allow
            Action:
              - codecommit:GitPull # Needs to be added for CODEBUILD_CLONE_REF
              - codecommit:GetBranch # Needs to be added for submodules?
              - codecommit:GetCommit # Needs to be added for submodules?
              - codecommit:GetRepository # Needs to be added for submodules?
            Resource:
              - "*"
  PipelineRolePolicy:
    Type: AWS::IAM::Policy
    Properties:
      PolicyName: !Sub ${AWS::StackName}-CodepipelinePolicy
      PolicyDocument:
        Version: 2012-10-17
        Statement:
          - Effect: Allow
            Action:
              - codepipeline:*
              - codecommit:GetBranch
              - codecommit:GetCommit
              - codecommit:UploadArchive
              - codecommit:GetUploadArchiveStatus
              - codecommit:CancelUploadArchive
              - codecommit:GetRepository # Needs to be added
[...]

PR #2830, opened 2021-09-14
feat: enable `OutputArtifactFormat` configuration for CodeCommit

Lou1415926:
Addressing #2799 .

This PR added the minimal required permissions as suggested by [CodeCommit](https://docs.aws.amazon.com/codecommit/latest/userguide/auth-and-access-control-permissions-reference.html), i.e. `codecommit:GitPull` and `codecommit:GetRepository`.

Additional permissions that are potentially needed, for example, to use submodules, can be added on user's end by modifying the policy directly (instead of through the CFN template). This is to ensure that Copilot is adding the minimally required permissions.


internal/pkg/template/templates/cicd/pipeline_cfn.yml
```diff
@@ -104,6 +104,14 @@ Resources:
               - ecr:CompleteLayerUpload
             Resource: '*'
             Condition: {StringEquals: {'ecr:ResourceTag/copilot-application': {{$.AppName}}}}
+          {{- if eq .Source.ProviderName "CodeCommit" }}
+          {{- if ne .Source.OutputArtifactFormat "" }}
```
huanjani: (here and two places below): I wonder if we should make this more strict, so there's less chance of failures after churning. Maybe we should look for only `CODEBUILD_CLONE_REF` since that's the only option besides the default.

"Values can be either CODEBUILD_CLONE_REF or CODE_ZIP. If unspecified, the default is CODE_ZIP."

internal/pkg/deploy/pipeline.go
```diff
@@ -160,10 +161,15 @@ func PipelineSourceFromManifest(mfSource *manifest.Source) (source interface{},
...
+			outputFormat = ""
+		}
 		return &CodeCommitSource{
-			ProviderName:  manifest.CodeCommitProviderName,
-			Branch:        (mfSource.Properties["branch"]).(string),
-			RepositoryURL: (mfSource.Properties["repository"]).(string),
+			ProviderName:         manifest.CodeCommitProviderName,
+			Branch:               (mfSource.Properties["branch"]).(string),
+			RepositoryURL:        (mfSource.Properties["repository"]).(string),
+			OutputArtifactFormat: (outputFormat).(string),
```
iamhopaul123: Maybe just
```suggestion
			OutputArtifactFormat: (mfSource.Properties["branch"]).(string),
```
since if the key doesn't exist, `mfSource.Properties["branch"]` would still be an empty string.

internal/pkg/template/templates/cicd/pipeline_cfn.yml
Lou1415926: Discussed offline.

If the value to `OutputArtifactFormat` is invalid, CodePipeline will still deploy successfully (with neither default nor full clone selected), and CloudFormation will NOT error out either. That is, the services will just swallow the invalid input. Us swallowing the invalid input is not really different from them swallowing the input.

We should either faithfully pass the value - either valid or invalid - to CFN, or error out early at manifest validation.

iamhopaul123: That is a good catch! I think we need a better validation for the pipeline manifest.

iamhopaul123 (approved):

internal/pkg/deploy/pipeline.go
Lou1415926: no `Properties` is a `map[string]interface{}`. It panics if we convert a `nil` `interface{}` to `string`

iamhopaul123: Oooh right i guess branch and repository are all required fields? But does that mean if users remove them from manifest we'll panic as well?

Lou1415926: I think so!! Let me take care of it.

internal/pkg/deploy/pipeline.go
```diff
@@ -123,16 +125,36 @@ type BitbucketSource struct {
 	ConnectionARN string
 }
 
+func convertProperty(properties map[string]interface{}, key string) (string, bool) {
+	v, ok := properties[key]
+	if !ok {
+		return "", false
+	}
+	return v.(string), true
```
efekarakus: nit: should we also check if the type can be converted?
```go
s, ok := v.(string)
if !ok {
  return "", false
}
return s, true
```

efekarakus (commented):
LGTM!

efekarakus:
Do we need to do something similar for other providers? https://docs.aws.amazon.com/codepipeline/latest/userguide/action-reference-CodestarConnectionSource.html

internal/pkg/deploy/pipeline.go
```diff
@@ -16,9 +16,14 @@ import (
 )
 
 const (
-	fmtInvalidRepo = "unable to locate the repository URL from the properties: %+v"
+	fmtInvalidRepo           = "unable to locate the repository URL from the properties: %+v"
```
huanjani: nit: maybe we can change the wording of this error msg slightly now that we have `fmtErrMissingProperty` because this err is only called when we can't parse the repo URL properly, right? 
```suggestion
	fmtInvalidRepo           = "unable to parse the repository from the URL %+v"
```

huanjani:
This looks fantastic! Just one nit/question.

internal/pkg/deploy/pipeline.go
```diff
@@ -16,9 +16,14 @@ import (
...
+	fmtErrMissingProperty    = "missing `%s` in properties"
+	fmtErrPropertyNotAString = "property `%s` is not a string"
 
 	defaultPipelineBuildImage = "aws/codebuild/amazonlinux2-x86_64-standard:3.0"
+
+	// DefaultPipelineBranch is the default repository branch to use for pipeline.
+	DefaultPipelineBranch = "main"
```
efekarakus: can we move the exported names above the private consts 🥺

internal/pkg/deploy/pipeline.go
```diff
@@ -110,29 +116,71 @@ type GitHubURL string
...
+		return "", fmt.Errorf(fmtErrMissingProperty, key)
+	}
+	vStr, ok := v.(string)
+	if !ok {
+		return "", fmt.Errorf(fmtErrPropertyNotAString, key)
+	}
+	return vStr, nil
+}
+
+func convertOptionalProperty(properties map[string]interface{}, key string, defaultValue string) (string, error) {
```
efekarakus, line 146: 🆒  fn signature!

efekarakus (approved):
:shipit: with a tiny req and Janice's feedback

PR #2836, opened 2021-09-15
feat: add `platform-os` and `platform-arch` flags to `task run`

huanjani:
Enable users to specify platform for the `task run` command via two new flags, `--platform-os` and `--platform-arch`. This information gets passed to the task's CFN template and also determines the Platform Version.

```
% copilot task run --help
Run a one-off task on Amazon ECS.

Usage
  copilot task run [flags]

Flags
      --app string                     Optional. Name of the application.
                                       Cannot be specified with 'default', 'subnets' or 'security-groups'.
      --cluster string                 Optional. The short name or full ARN of the cluster to run the task in. 
                                       Cannot be specified with 'app', 'env' or 'default'.
      --command string                 Optional. The command that is passed to "docker run" to override the default command.
      --count int                      Optional. The number of tasks to set up. (default 1)
      --cpu int                        Optional. The number of CPU units to reserve for each task.
      --default                        Optional. Run tasks in default cluster and default subnets. 
                                       Cannot be specified with 'app', 'env' or 'subnets'.
      --dockerfile string              Path to the Dockerfile.
                                       Mutually exclusive with -i, --image. (default "Dockerfile")
      --entrypoint string              Optional. The entrypoint that is passed to "docker run" to override the default entrypoint.
      --env string                     Optional. Name of the environment.
                                       Cannot be specified with 'default', 'subnets' or 'security-groups'.
      --env-vars stringToString        Optional. Environment variables specified by key=value separated by commas. (default [])
      --execution-role string          Optional. The ARN of the role that grants the container agent permission to make AWS API calls.
      --follow                         Optional. Specifies if the logs should be streamed.
      --generate-cmd string            Optional. Generate a command with a pre-filled value for each flag.
                                       To use it for an ECS service, specify --generate-cmd <cluster name>/<service name>.
                                       Alternatively, if the service or job is created with Copilot, specify --generate-cmd <application>/<environment>/<service or job name>.
                                       Cannot be specified with any other flags.
  -h, --help                           help for run
  -i, --image string                   The location of an existing Docker image.
                                       Mutually exclusive with -d, --dockerfile.
      --memory int                     Optional. The amount of memory to reserve in MiB for each task.
      --platform-arch string           Optional. Architecture of the task. Must be specified along with 'platform-os'.
      --platform-os string             Optional. Operating system of the task. Must be specified along with 'platform-arch'.
      --resource-tags stringToString   Optional. Labels with a key and value separated by commas.
                                       Allows you to categorize resources. (default [])
      --secrets stringToString         Optional. Secrets to inject into the container. Specified by key=value separated by commas. (default [])
      --security-groups strings        Optional. The security group IDs for the task to use. Can be specified multiple times.
                                       Cannot be specified with 'app' or 'env'.
      --subnets strings                Optional. The subnet IDs for the task to use. Can be specified multiple times.
                                       Cannot be specified with 'app', 'env' or 'default'.
      --tag string                     Optional. The container image tag in addition to "latest".
  -n, --task-group-name string         Optional. The group name of the task. 
                                       Tasks with the same group name share the same set of resources. 
                                       (default directory name)
      --task-role string               Optional. The ARN of the role for the task to use.

Examples
Run a task using your local Dockerfile and display log streams after the task is running. 
You will be prompted to specify an environment for the tasks to run in.
`$ copilot task run`
Run a task named "db-migrate" in the "test" environment under the current workspace.
`$ copilot task run -n db-migrate --env test`
Run 4 tasks with 2GB memory, an existing image, and a custom task role.
`$ copilot task run --count 4 --memory 2048 --image=rds-migrate --task-role migrate-role`
Run a task with environment variables.
`$ copilot task run --env-vars name=myName,user=myUser`
Run a task using the current workspace with specific subnets and security groups.
`$ copilot task run --subnets subnet-123,subnet-456 --security-groups sg-123,sg-456`
Run a task with a command.
`$ copilot task run --command "python migrate-script.py"`
```


internal/pkg/cli/task_run.go
```diff
@@ -302,6 +308,39 @@ func (o *runTaskOpts) Validate() error {
...
+		}
+	}
+	if o.memory == 0 {
+		o.memory = 512
+		if isWindowsOS(o.os) {
+			o.memory = manifest.WindowsTaskMemory
+		}
+	}
+
+	if o.cpu < 0 {
```
iamhopaul123, line 313: nit: we can just use `Uint16`

internal/pkg/cli/task_run.go
```diff
@@ -333,6 +376,32 @@ func (o *runTaskOpts) Validate() error {
 	return nil
 }
 
+func (o *runTaskOpts) validateOS() error {
```
iamhopaul123: Hmm iirc we have very similar logic somewhere for service right? Maybe we can reuse it?

internal/pkg/manifest/workload.go
```diff
@@ -69,10 +69,10 @@ var (
 
 	defaultPlatform = platformString(OSLinux, ArchAMD64)
 
-	windowsTaskCPU    = 1024
-	windowsTaskMemory = 2048
+	WindowsTaskCPU    = 1024
+	WindowsTaskMemory = 2048
 
-	windowsOSFamilies = []string{OSWindows, OSWindowsServer2019Core, OSWindowsServer2019Full}
+	WindowsOSFamilies = []string{OSWindows, OSWindowsServer2019Core, OSWindowsServer2019Full}
```
iamhopaul123: Maybe adding some comments since they are now public vars

internal/pkg/template/templates/task/cf.yml
```diff
@@ -33,6 +37,8 @@ Conditions:
     !Not [!Equals [ !Join ["", !Ref Command], ""]]
   HasEntryPoint:
     !Not [ !Equals [ !Join [ "", !Ref EntryPoint ], "" ] ]
+  HasPlatform:
```
iamhopaul123: `HasPlatform` sounds a little wield since it is more about if the platform is set by them otherwise we use default. Maybe `useCustomPlatform` or other similar things?

internal/pkg/cli/task_run.go
```diff
@@ -302,6 +308,39 @@ func (o *runTaskOpts) Validate() error {
...
+		return err
+	}
+
+	// Set default values for CPU and memory based on the operating system.
+	// The flag default value is 0. If a user sets `--cpu 0` or `--memory 0`, then the 0 value will be replaced by 256/1024 for cpu and 512/2048 for mem, depending on OS, instead of erroring out.
+	if o.cpu == 0 {
+		o.cpu = 256
+		if isWindowsOS(o.os) {
+			o.cpu = manifest.WindowsTaskCPU
+		}
```
iamhopaul123: This looks a bit off, since we are not supposed to assign values in `Validate()`...

internal/pkg/cli/task_run.go
huanjani: Oh, those validate the whole platform, but I can make helper methods that the platform validator can call and we can use here.

huanjani: Elsewhere we are validating the entire platform. (So far) this is the only place where users pass in the OS and Arch separately.

internal/pkg/template/templates/task/cf.yml
huanjani: Hmm. `HasTaskRole` also sets to a default if one isn't passed in. How about `HasNondefaultPlatform`? That also sounds weird. 🤔

internal/pkg/aws/ecs/ecs.go
```diff
@@ -331,7 +332,7 @@ func (e *ECS) RunTask(input RunTaskInput) ([]*Task, error) {
 			},
 		},
 		EnableExecuteCommand: aws.Bool(true),
-		PlatformVersion:      aws.String("1.4.0"),
+		PlatformVersion:      aws.String(input.PlatformVersion),
```
efekarakus, line 335: just to double check `aws.String("")` will result in LATEST right and not an error?

internal/pkg/cli/task_run.go
```diff
@@ -59,6 +61,11 @@ const (
 	fmtImageURI = "%s:%s"
 )
 
+var (
+	validOS   = []string{manifest.OSLinux, manifest.OSWindows, manifest.OSWindowsServer2019Full, manifest.OSWindowsServer2019Core}
+	validArch = []string{manifest.ArchAMD64, manifest.ArchX86}
```
efekarakus: Should we make the following variable public and validate the flags against these options?
https://github.com/aws/copilot-cli/blob/6cf6a8f0ba7589bff9ddd84a063448e3f560a1c2/internal/pkg/manifest/workload.go#L61-L68

This way in the future we'll only need to update this var and get other os/arch combos supported for free 🌈

internal/pkg/cli/task_run.go
```diff
@@ -302,6 +308,39 @@ func (o *runTaskOpts) Validate() error {
 		}
 	}
 
+	if noOS, noArch := o.os == "", o.arch == ""; noOS != noArch {
```
efekarakus, line 306: neat :O

internal/pkg/cli/task_run.go
efekarakus: Agreed! I think we should keep the default cpu and mem settings as is:
https://github.com/aws/copilot-cli/blob/d5c08711e55f9b346ecb49180c494b4e027f51da/internal/pkg/cli/task_run.go#L865-L866

It's a bit odd to read the default as 0 as user, they might think it's a legit accepted value. Instead, what do you think of returning an error here if `cpu` is not at least `MinWindowsTaskCPU`?

efekarakus: Also the user might be surprised to see a higher bill if we change the cpu and mem setting without asking them first.

internal/pkg/cli/task_run.go
huanjani: We could...but to convert it to a pointer to a string in the `stack` package we'd have to do:
```
{
			ParameterKey:   aws.String(taskCPUParamKey),
			ParameterValue: aws.String(strconv.FormatInt(int64(t.CPU), 10)),
		}
```
Is that worth avoiding the positive int validation?
Another option is to make sure that it's greater than 256. Is there any case in which it could be 0<cpu<256? Any idea why the validation was just for >0 before?

internal/pkg/manifest/workload.go
```diff
@@ -69,10 +69,10 @@ var (
 
 	defaultPlatform = platformString(OSLinux, ArchAMD64)
 
-	windowsTaskCPU    = 1024
-	windowsTaskMemory = 2048
+	WindowsTaskCPU    = 1024
+	WindowsTaskMemory = 2048
```
efekarakus: nit:
```suggestion
	MinWindowsTaskCPU    = 1024
	MinWindowsTaskMemory = 2048
```

efekarakus: Doc comments please for all exported names 🥺 
Also can we move the exported names above the private ones?

internal/pkg/task/config_runner.go
```diff
@@ -65,14 +71,21 @@ func (r *ConfigRunner) Run() ([]*Task, error) {
 		}
 		r.Subnets = subnets
 	}
+	platformVersion := "1.4.0"
```
efekarakus: I wonder if we can leave this empty and only set `1.0.0` if it's window and the command will work? 🤔 
This way if LATEST ever changes we don't need to update it here. (or we can set it to LATEST)

internal/pkg/task/env_runner.go
```diff
@@ -74,13 +80,21 @@ func (r *EnvRunner) Run() ([]*Task, error) {
 		return nil, fmt.Errorf(fmtErrSecurityGroupsFromEnv, r.Env, err)
 	}
 
+	platformVersion := "1.4.0"
+	for _, windowsOS := range manifest.WindowsOSFamilies {
+		if r.OS == windowsOS {
+			platformVersion = "1.0.0"
+		}
+	}
```
efekarakus: samesies here

internal/pkg/template/templates/task/cf.yml
```diff
@@ -62,6 +68,9 @@ Resources:
             ValueFrom: {{$valueFrom}}{{end}}
           {{- end}}
       Family: !Join ['-', ["copilot", !Ref TaskName]]
+      RuntimePlatform:
+        OperatingSystemFamily: !If [HasPlatform, !Ref OS, "LINUX"]
+        CpuArchitecture: !If [HasPlatform, !Ref Arch, "X86_64"]
```
efekarakus: Can we do the following?
```yaml
"RuntimePlatform" : {
      "Fn::If" : [
        "HasCustomPlatform",
        {"OperatingSystemFamily" : {"Ref" : "OS"}, "CpuArchitecture": {"Ref" : "Arch"}},
        {"Ref" : "AWS::NoValue"}
      ]
    }
```

internal/pkg/cli/task_run.go
iamhopaul123: i think in future it is likely they'll allow 0<cpu<256. Maybe just keep as is...

internal/pkg/template/templates/task/cf.yml
iamhopaul123: Maybe keep `HasPlatform` 😢

internal/pkg/task/config_runner.go
huanjani: I believe it has been hard-coded as `1.4.0` in the past out of concern that if there's a new LATEST that breaks things, we'll be in a bad way, so it's better to manually change it to a future platform version once we've confirmed that all is fine. But yeah, it is an optional field so if we decide we're comfortable with LATEST, then we should be able to leave it empty.

huanjani:
Good points re: cpu and mem. I'm already throwing the error here:
```
func (o *runTaskOpts) validateFlagsWithWindows() error {
	if !isWindowsOS(o.os) {
		return nil
	}
	if o.cpu < manifest.WindowsTaskCPU {
		return fmt.Errorf("CPU %d must be at least %d for a Windows-based task", o.cpu, manifest.WindowsTaskCPU)
	}
	if o.memory < manifest.WindowsTaskMemory {
		return fmt.Errorf("memory %d must be at least %d for a Windows-based task", o.memory, manifest.WindowsTaskMemory)
	}
	return nil
}
```
Should I also throw an error if non-Windows tasks are below their min values?

internal/pkg/template/templates/task/cf.yml
huanjani: changed to `HasCustomPlatform`

internal/pkg/cli/task_run.go
huanjani: Now validating as platform rather than separately!

internal/pkg/task/config_runner.go
efekarakus: Personally, I think it makes sense for us to stick to LATEST for the PlatformVersion, and then switch it up if we detect certain operating systems/architectures

huanjani: @iamhopaul123 How do you feel? (You were the one who questioned it in the design doc.)

internal/pkg/manifest/workload.go
```diff
@@ -49,15 +49,22 @@ var (
 	// WorkloadTypes holds all workload manifest types.
 	WorkloadTypes = append(ServiceTypes, JobTypes...)
 
-	// All placement options.
-	subnetPlacements = []string{PublicSubnetPlacement, PrivateSubnetPlacement}
+	// Minimum CPU and mem values required for Windows-based tasks.
+	WindowsTaskCPU    = 1024
+	WindowsTaskMemory = 2048
```
efekarakus:
```suggestion
	MinWindowsTaskCPU    = 1024
	MinWindowsTaskMemory = 2048
```
(we can also move these to `const` instead of `var`)

efekarakus (approved):
Looks good to me! Applied the label but feel free to remove it once the `platformVersion` discussion is resolved.
And one tiny request for the const names

internal/pkg/aws/ecs/ecs.go
huanjani: No, it'll throw this error: ` InvalidParameterException: Platform version can not be blank.`

internal/pkg/task/config_runner.go
huanjani: Discussed offline. Changed "1.4.0" for PlatformVersion to "LATEST" throughout codebase.

iamhopaul123: Sounds good to me!

iamhopaul123 (approved):

PR #2840, opened 2021-09-16
test(e2e): make sidecar suite more stable

iamhopaul123:

### DIFF
--- a/e2e/sidecars/sidecars_test.go
+++ b/e2e/sidecars/sidecars_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"net/http"
 	"os"
 	"strings"
@@ -233,7 +234,7 @@ var _ = Describe("sidecars flow", func() {
 			uri := route.URL + "/health-check"
 
 			// Service should be ready.
-			var resp *http.Response
+			resp := &http.Response{}
 			var fetchErr error
 			Eventually(func() (int, error) {
 				resp, fetchErr = http.Get(uri)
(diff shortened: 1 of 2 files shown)


e2e/sidecars/sidecars_test.go
```diff
@@ -248,29 +249,38 @@ var _ = Describe("sidecars flow", func() {
 		})
 
 		It("svc logs should display logs", func() {
+			var firelensCreated bool
 			var svcLogs []client.SvcLogsOutput
 			var svcLogsErr error
-			Eventually(func() ([]client.SvcLogsOutput, error) {
+			for i := 0; i < 10; i++ {
```
efekarakus, line 255: any reasons why you didn't go with the `Eventually` function? 

```go
Eventually(func() error {
  svcLogs, svcLogsErr = cli.SvcLogs(&client.SvcLogsRequest{
  if err != nil {
     return err
  }
  ...
  if noLogLine {
     return errors.New("no firelens log")
  }
  return nil
}, "300s", "20s").ShouldNot(HaveOccured()))

```

iamhopaul123: Reason is because this is not like the custom domain e2e test, and the waiting time really varies: most of the time it doesn't need to wait for a long time. The exponential backoff with jitter would be better in this situation I think.

efekarakus: Gotcha! ok honestly I'd rather do `Eventually` everywhere to have consistency but it really doesn't matter much for this 😄 so I'll leave it to your judgment!

efekarakus (approved):

e2e/sidecars/sidecars_test.go
iamhopaul123: I tested it many times before pushing it lol let's keep it for now since all we want is a stable suite. I'll def. change it if we ever fail again for this one!

huanjani (approved):
:shipit:

PR #2844, opened 2021-09-16
feat: enable `StartCommand` configuration for App Runner workloads

### LINKED ISSUE #2709: AppRunner : Please support StartCommand of CodeConfigurationValues
I want to use `command` override like compose file setting.
(example : https://github.com/sogaoh/laravel-app-runner-example/blob/main/appBase/run/docker-compose.yml#L17 )

So, please support StartCommand of CodeConfigurationValues in manifest.yml. 

refs:
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apprunner-service-codeconfigurationvalues.html
- https://docs.aws.amazon.com/apprunner/latest/api/API_CodeConfigurationValues.html
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/apprunner_service#start_command

Lou1415926:
This PR adds [`StartCommand`](https://docs.aws.amazon.com/apprunner/latest/api/API_ImageConfiguration.html#apprunner-Type-ImageConfiguration-StartCommand) configuration for App Runner workloads.

It is configurable as `start_command` in request-driven web service's manifest, accepting either a string or a slice of string, just like `command` in other workloads.

Resolves #2709 

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/rd_web_svc_integration_test.go
+++ b/internal/pkg/deploy/cloudformation/stack/rd_web_svc_integration_test.go
@@ -20,6 +20,7 @@ import (
 
 func TestRDWS_Template(t *testing.T) {
 	const (
+		envName               = "test"
 		manifestFileName      = "rdws-manifest.yml"
 		stackTemplateFileName = "rdws.stack.yml"
 	)
--- a/internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
@@ -42,5 +42,6 @@ publish:
 # You can override any of the values defined above by environment.
 environments:
   test:
+    start_command: "echo es by pink floyd"
     variables:
       LOG_LEVEL: debug        # Log level for the "test" environment.
--- a/internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
@@ -40,7 +40,7 @@ publish:
 #   project: project-name
 
 # You can override any of the values defined above by environment.
-# environments:
-#   test:
-#     variables:
-#       LOG_LEVEL: debug        # Log level for the "test" environment.
+environments:
+  test:
+    variables:
+      LOG_LEVEL: debug        # Log level for the "test" environment.
--- a/internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
@@ -26,6 +26,7 @@ image:
 cpu: 1024
 # Amount of memory in MiB used by the task.
 memory: 2048
+command: "shine on you"
 
 publish:
   topics:
@@ -42,6 +43,6 @@ publish:
 # You can override any of the values defined above by environment.
 environments:
   test:
-    start_command: "echo es by pink floyd"
+    command: crazy diamond
     variables:
       LOG_LEVEL: debug        # Log level for the "test" environment.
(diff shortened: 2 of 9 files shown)


internal/pkg/deploy/cloudformation/stack/rd_web_svc_integration_test.go
```diff
@@ -20,6 +20,7 @@ import (
 
 func TestRDWS_Template(t *testing.T) {
 	const (
+		envName               = "test"
```
iamhopaul123, line 23: Do we use this const?

internal/pkg/manifest/workload.go
```diff
@@ -272,6 +273,18 @@ func toStringSlice(s *stringSliceOrString) ([]string, error) {
 	return out, nil
 }
 
+func toString(s *stringSliceOrString) (string, error) {
```
iamhopaul123: Maybe a comment showing an example since we specifically use " " as the separator

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
```diff
@@ -103,6 +103,9 @@ Resources:
                   Fn::GetAtt: [ {{$stackName}}, Outputs.{{$var}}]
               {{- end}}
               {{- end}}
+            {{- if .StartCommand }}
+            StartCommand: {{.StartCommand}}
```
iamhopaul123, line 107: It seems like `start_command` exists in both code configuration and image configuration 🤔 if we just name it to be `StartCommand` how do users know which one is it? https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apprunner-service-codeconfigurationvalues.html#cfn-apprunner-service-codeconfigurationvalues-startcommand

Lou1415926: We dont use code configuration in our CFN tempalte.

Exactly one of the configuration and image configuration needs to be used, and we are using the latter one.

internal/pkg/manifest/rd_web_svc.go
```diff
@@ -28,6 +32,7 @@ type RequestDrivenWebServiceConfig struct {
 	InstanceConfig                    AppRunnerInstanceConfig `yaml:",inline"`
 	ImageConfig                       ImageWithPort           `yaml:"image"`
 	Variables                         map[string]string       `yaml:"variables"`
+	StartCommand                      StartCommand            `yaml:"start_command"`
```
efekarakus: should we do `yaml:"command"` instead? to be consistent with our other workload types

internal/pkg/deploy/cloudformation/stack/rd_web_svc_integration_test.go
Lou1415926: We do. So the original `envName` is defined in `stack_local_integration_test` but I think it's better to use a local variable, so we clearly see what's the envirnoment that's being used. Hence this variable shadows the other one.

internal/pkg/manifest/workload.go
Lou1415926: I removed this altogether, leaving `start_command` as `string` only (not `stringSliceOrString`). There is currently no good way to convert a slice to a string using shlex rules, and if we shouldn't do the conversion if we can't do it properly.

internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
```diff
@@ -26,6 +26,7 @@ image:
 cpu: 1024
 # Amount of memory in MiB used by the task.
 memory: 2048
+command: "shine on you"
```
efekarakus, line 29: 🎸

efekarakus (approved):

internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
Lou1415926: 💎

iamhopaul123 (approved):
:shipit:

PR #2849, opened 2021-09-17
chore: update set-value to >=4.0.1

iamhopaul123:
Fix dependabot alert

### DIFF
--- a/cf-custom-resources/package-lock.json
+++ b/cf-custom-resources/package-lock.json
@@ -2501,7 +2501,7 @@
         "get-value": "^2.0.6",
         "has-value": "^1.0.0",
         "isobject": "^3.0.1",
-        "set-value": "^2.0.0",
+        "set-value": ">=4.0.1",
         "to-object-path": "^0.3.0",
         "union-value": "^1.0.0",
         "unset-value": "^1.0.0"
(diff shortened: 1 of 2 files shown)


huanjani (approved):

cf-custom-resources/package-lock.json
```diff
@@ -2501,7 +2501,7 @@
         "get-value": "^2.0.6",
         "has-value": "^1.0.0",
         "isobject": "^3.0.1",
-        "set-value": "^2.0.0",
+        "set-value": ">=4.0.1",
```
efekarakus, line 2505: shouldn't we be updating the `dependencies` section instead of modifying `package-lock`?
https://github.com/aws/copilot-cli/blob/mainline/cf-custom-resources/package.json#L25

iamhopaul123: yeah that's why i put the label before 😃 now fixed

iamhopaul123: I still didn't add it to the package file because we don't directly depend on that package.

efekarakus: But won't it get overridden, I don't get how it's fixed 😅 
My understanding is that `package-lock` isnt' meant to be modified manually. If we ever upgrade a dependency in `package.json`, and run `npm install` then I think it will reset these changes, I'd expected the following to be our enforcing mechanism:
```js
"dependencies": {
   "set-value": ">=4.0.1",
```

iamhopaul123: Hmm i tried to do `npm install` and it didn't revert the change. But sure I'll just add it to package.json

efekarakus (approved):

PR #2851, opened 2021-09-17
chore: validate Dockerfile parsing against Docker's parser

efekarakus:
This PR contains some small refactors, but more importantly adds additional unit tests to ensure that our `EXPOSE` and `HEALTHCHECK` parsers match Docker's.

Once this change is merged, a follow-up will remove Docker as a dependency from the CLI.

### DIFF
--- a/internal/pkg/docker/dockerfile/dockerfile.go
+++ b/internal/pkg/docker/dockerfile/dockerfile.go
@@ -85,7 +85,6 @@ func New(fs afero.Fs, path string) *Dockerfile {
 		healthCheck:  nil,
 		fs:           fs,
 		path:         path,
-		parsed:       false,
 	}
 }
 
--- a/internal/pkg/docker/dockerfile/dockerfile.go
+++ b/internal/pkg/docker/dockerfile/dockerfile.go
@@ -62,8 +62,8 @@ type HealthCheck struct {
 
 // Dockerfile represents a parsed Dockerfile.
 type Dockerfile struct {
-	ExposedPorts []portConfig
-	HealthCheck  *HealthCheck
+	exposedPorts []portConfig
+	healthCheck  *HealthCheck
 	parsed       bool
 	path         string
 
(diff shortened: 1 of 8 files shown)


internal/pkg/docker/dockerfile/dockerfile_test.go
```diff
@@ -152,8 +81,11 @@ EXPOSE 80`),
 		},
 		"bad expose token single port": {
 			dockerfilePath: wantedPath,
-			dockerfile:     []byte(`EXPOSE $arg`),
-			wantedPorts: []portConfig{
+			dockerfile: []byte(`
+FROM nginx
+EXPOSE $arg
+`),
+			wantedPorts: []Port{
```
Lou1415926: nit: seems like we'd wanted `nil` for ports

Lou1415926 (approved):
lgtm! just one nit on a unit test!

internal/pkg/docker/dockerfile/dockerfile.go
```diff
@@ -62,50 +70,59 @@ type HealthCheck struct {
...
+// New returns an empty Dockerfile.
+func New(fs afero.Fs, path string) *Dockerfile {
 	return &Dockerfile{
-		ExposedPorts: []portConfig{},
-		HealthCheck:  nil,
+		exposedPorts: []Port{},
+		healthCheck:  nil,
 		fs:           fs,
 		path:         path,
 		parsed:       false,
```
iamhopaul123: Probably can be just
```suggestion
		fs:           fs,
		path:         path,
```

iamhopaul123 (approved):
🚢

internal/pkg/docker/dockerfile/dockerfile_test.go
efekarakus: I'm a dummy yeah! I changed this to `nil` and added a `require.Nil` (although it wouldn't have caught this)

internal/pkg/docker/dockerfile/dockerfile.go
efekarakus: Done!

PR #2852, opened 2021-09-17
chore(manifest): add advanced types validation I

iamhopaul123:
Part of #2818. This PR adds advanced types validation for
- percentage
- platform
- placement
- IP

Note that this PR doesn't move any existing validation in `UnmarshalYAML`.

### DIFF
--- a/internal/pkg/manifest/validate.go
+++ b/internal/pkg/manifest/validate.go
@@ -357,11 +357,7 @@ func (p *PlatformString) Validate() error {
 	if p == nil {
 		return nil
 	}
-	val := string(*p)
-	reg := regexp.MustCompile(`^.+\/.+$`)
-	if !reg.MatchString(val) {
-		return fmt.Errorf(`platform %s must be in format of <OS>/<Arch>`, val)
-	}
+	// TODO: Consolidate with "UnmarshalYAML".
 	return nil
 }
 
@@ -662,7 +658,7 @@ func (p *Placement) Validate() error {
 			return nil
 		}
 	}
-	return fmt.Errorf(`"placement" %s must be one of %#v"`, string(*p), subnetPlacements)
+	return fmt.Errorf(`"placement" %s must be one of %s`, string(*p), strings.Join(subnetPlacements, ", "))
 }
 
 // Validate returns nil if RequestDrivenWebServiceHttpConfig is configured correctly.
--- a/internal/pkg/manifest/validate.go
+++ b/internal/pkg/manifest/validate.go
@@ -6,6 +6,7 @@ package manifest
 import (
 	"errors"
 	"fmt"
+	"net"
 	"regexp"
 	"strconv"
 	"strings"
@@ -284,6 +285,12 @@ func (r *RoutingRule) Validate() error {
 			secondField: "targetContainer",
 		}
 	}
+	for ind, ip := range r.AllowedSourceIps {
+		if err = ip.Validate(); err != nil {
+			return fmt.Errorf(`validate "allowed_source_ips[%v]": %w`, ind, err)
+		}
+	}
+
 	return nil
 }
 
@@ -308,6 +315,17 @@ func (*Alias) Validate() error {
 	return nil
 }
 
+// Validate returns nil if IPNet is configured correctly.
+func (ip *IPNet) Validate() error {
+	if ip == nil {
+		return nil
+	}
+	if _, _, err := net.ParseCIDR(string(*ip)); err != nil {
+		return err
+	}
+	return nil
+}
+
 // Validate returns nil if TaskConfig is configured correctly.
 func (t *TaskConfig) Validate() error {
 	var err error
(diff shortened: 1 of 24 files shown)


internal/pkg/manifest/validate.go
```diff
@@ -284,6 +285,12 @@ func (r *RoutingRule) Validate() error {
 			secondField: "targetContainer",
 		}
 	}
+	for ind, ip := range r.AllowedSourceIps {
+		if err = ip.Validate(); err != nil {
+			return fmt.Errorf(`validate "allowed_source_ips[%v]": %w`, ind, err)
```
Lou1415926: would it be helpful if we show the exact string of the `IpNet` that has failed the validation?

internal/pkg/manifest/validate.go
```diff
@@ -328,9 +346,25 @@ func (t *TaskConfig) Validate() error {
...
 	return p.PlatformArgs.Validate()
 }
 
+// Validate returns nil if PlatformString is configured correctly.
+func (p *PlatformString) Validate() error {
+	if p == nil {
+		return nil
+	}
+	val := string(*p)
+	reg := regexp.MustCompile(`^.+\/.+$`)
```
Lou1415926: Do we want to validate the `OS` and `Arch` like this https://github.com/aws/copilot-cli/blob/7e3d2bfe1af5e7389806f55b9e0634d8d78ae3c3/internal/pkg/manifest/workload.go#L621 here, or during `Unmarshal`?

internal/pkg/manifest/validate.go
```diff
@@ -328,9 +346,25 @@ func (t *TaskConfig) Validate() error {
...
+	if !reg.MatchString(val) {
+		return fmt.Errorf(`cannot use %s for platform. Must match the regex "^.+\/.+$"`, val)
```
Lou1415926: Rather than giving the regex, perhaps it's more user friendly to say something like `"platform", when specified as a string, must be of form <OS>/<Arch>. For example, "linux/amd64"`

internal/pkg/manifest/validate.go
```diff
@@ -384,6 +418,24 @@ func (a *AdvancedCount) Validate() error {
...
+	return nil
+}
+
+// Validate returns nil if Percentage is configured correctly.
+func (p *Percentage) Validate() error {
+	if p == nil {
+		return nil
+	}
+	if val := int(*p); val < 0 || val > 100 {
+		return fmt.Errorf("cannot specify %v as Percentage. Must be an integer from 0 to 100", val)
```
```suggestion
		return fmt.Errorf("Must be an integer from 0 to 100")
```
I feel like this should suffice to tell the user what's wrong. The error produced would be `validate "cpu_percentage": must be an integer from 0 to 100`.

internal/pkg/manifest/validate.go
```diff
@@ -593,9 +645,25 @@ func (v *vpcConfig) Validate() error {
...
+func (p *Placement) Validate() error {
+	if p == nil {
+		return fmt.Errorf(`"placement" cannot be empty`)
+	}
+	for _, allowed := range subnetPlacements {
+		if string(*p) == allowed {
+			return nil
+		}
+	}
+	return fmt.Errorf(`"placement" %s is invalid. Must be one of %#v"`, string(*p), subnetPlacements)
```
```suggestion
	return fmt.Errorf(`"placement" must be one of %#v"`,  subnetPlacements)
```
Samesies here. The second sentence should probably suffice.

Lou1415926 (commented):
LGTM! Just some nits on error messages!

internal/pkg/manifest/validate.go
iamhopaul123: Maybe not as it is just wrapping an error. The error itself should include the value I think (i'll update)

internal/pkg/manifest/validate.go
iamhopaul123: Yes! I am planning to do it when moving all the logic in `Unmarshal` to `Validate()`

Lou1415926 (approved):
:shipit:

internal/pkg/manifest/validate.go
```diff
@@ -328,9 +346,25 @@ func (t *TaskConfig) Validate() error {
...
 }
 
+// Validate returns nil if PlatformString is configured correctly.
+func (p *PlatformString) Validate() error {
+	if p == nil {
+		return nil
+	}
+	val := string(*p)
+	reg := regexp.MustCompile(`^.+\/.+$`)
+	if !reg.MatchString(val) {
```
efekarakus: We should avoid using regular expressions so frequently (see this article by Rob Pike: https://commandcenter.blogspot.com/2011/08/regular-expressions-in-lexing-and.html) 

In this situation our expected string is fairly straightforward and we can get a better performance boost and less error prone code like this:
```go
for _, validPlatform := range validPlatforms {
  if val == validPlatform {
     return nil
  }
}
return fmt.Errorf(`platform %s must be one of: %s`, val, strings.Join(validPatforms, ", "))
```

I like his rule of thumb around regex:
> Another way to look at it is that lexers and parsing are matching statically-defined patterns, but regular expressions' strength is that they provide a way to express patterns dynamically.

internal/pkg/manifest/validate.go
```diff
@@ -593,9 +646,25 @@ func (v *vpcConfig) Validate() error {
...
+func (p *Placement) Validate() error {
+	if p == nil {
+		return fmt.Errorf(`"placement" cannot be empty`)
+	}
+	for _, allowed := range subnetPlacements {
+		if string(*p) == allowed {
+			return nil
+		}
+	}
+	return fmt.Errorf(`"placement" %s must be one of %#v"`, string(*p), subnetPlacements)
```
```suggestion
	return fmt.Errorf(`"placement" %s must be one of %s"`, string(*p), strings.Join(subnetPlacements, ", "))
```

internal/pkg/manifest/workload.go
```diff
@@ -594,11 +599,24 @@ func (hc *ContainerHealthCheck) ApplyIfNotSet(other *ContainerHealthCheck) {
 	}
 }
 
+// PlatformString represents the platform string consisting of OS family and architecture type.
+// For example: "windows/x86"
+type PlatformString string
+
+// PlatformStringP converts a string pointer to a PlatformString pointer.
+func PlatformStringP(s *string) *PlatformString {
```
efekarakus: This signature is so strange (accepting a string pointer).

I think @huanjani changed a bunch of stuff here in `feat/pencere` but I might be wrong. Just as a heads-up Janice that there will be again merge conflicts to deal with :| 

I wonder if we can move most of your changes to `mainline`.

iamhopaul123: no worries! i'll remove this func

efekarakus (approved):

PR #2864, opened 2021-09-22
chore: remove buildkit dependency 

efekarakus:
Removing the dependency saves us 3.2MB of space and brings us down to 42MB.
The bulk of the lexer is heavily inspired by https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/text/template/parse/lex.go


internal/pkg/docker/dockerfile/dockerfile.go
```diff
@@ -154,39 +151,29 @@ func (df *Dockerfile) parse() error {
 }
 
 // parse parses the contents of a Dockerfile into a Dockerfile struct.
-func parse(content string) (*Dockerfile, error) {
+func parse(name, content string) (*Dockerfile, error) {
```
iamhopaul123, line 154: Maybe
```suggestion
func parse(lexer *lexer) (*Dockerfile, error) {
```

internal/pkg/docker/dockerfile/lex.go
```diff
@@ -0,0 +1,241 @@
...
+const (
+	instrErr         instructionName = iota // an error occurred while scanning.
+	instrHealthCheck                        // a HEALTHCHECK instruction.
+	instrExpose                             // an EXPOSE instruction.
+	instrEOF                                // done scanning.
+)
+
+const (
+	exposeInstrMarker     = "expose "      // start of an EXPOSE instruction.
+	heathCheckInstrMarker = "healthcheck " // start of a HEALTHCHECK instruction.
```
iamhopaul123: nit:
```suggestion
	healthCheckInstrMarker = "healthcheck " // start of a HEALTHCHECK instruction.
```

internal/pkg/docker/dockerfile/lex.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+	}
+	return false
+}
+
+// trimInstruction trims the instrMarker prefix from line and returns it.
+func trimInstruction(line, instrMarker string) string {
+	normalized := strings.ToLower(line)
+	idx := strings.Index(normalized, instrMarker) + len(instrMarker)
+	return line[idx:]
```
iamhopaul123, line 225: maybe `trimPrefix`?

internal/pkg/docker/dockerfile/lex.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// trimContinuationLineMarker returns the line without any continuation line markers.
+// If the line doesn't have a continuation marker, then returns it as is.
+func trimContinuationLineMarker(line string) string {
+	for _, marker := range lineContinuationMarkers {
+		if strings.HasSuffix(line, marker) {
+			return line[:len(line)-1-len(marker)]
```
iamhopaul123: maybe `trimSuffix`?

internal/pkg/docker/dockerfile/lex.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+	}
+	return line
+}
+
+// trimLeadingWhitespaces removes any leading space characters.
+func trimLeadingWhitespaces(line string) string {
+	return strings.TrimLeftFunc(line, unicode.IsSpace)
```
iamhopaul123, line 241: maybe https://pkg.go.dev/strings#TrimSpace to trim both leading and trailing spaces?

iamhopaul123 (commented):
This is a really mind-blowing PR! Thank you for sharing this idea :tada:

internal/pkg/docker/dockerfile/lex.go
```diff
@@ -0,0 +1,241 @@
...
+
+const (
+	instrErr         instructionName = iota // an error occurred while scanning.
+	instrHealthCheck                        // a HEALTHCHECK instruction.
+	instrExpose                             // an EXPOSE instruction.
+	instrEOF                                // done scanning.
+)
+
+const (
+	exposeInstrMarker     = "expose "      // start of an EXPOSE instruction.
```
huanjani: or `instrMarkerExpose` and `instrMarkerHealthcheck`?

internal/pkg/docker/dockerfile/lex.go
```diff
@@ -0,0 +1,241 @@
...
+	return <-lex.instructions
+}
+
+// readLine loads the next line in the Dockerfile.
+// If we reached the end of the file, then isEOF is set to true.
+// If any unexpected error occurs during scanning, then err is not nil.
+func (lex *lexer) readLine() (isEOF bool, err error) {
+	if ok := lex.scanner.Scan(); !ok {
+		if err := lex.scanner.Err(); err != nil {
+			return false, fmt.Errorf("scan Dockerfile %s: %w", lex.name, err)
```
Lou1415926: I wonder if the lex's error cares about the file name information 🤔  If `lex` is just scanning and processing the content, maybe the file name information would be better suited in the error that wraps this error, that is in the caller error? Just a thought 💭

internal/pkg/docker/dockerfile/lex.go
```diff
@@ -0,0 +1,241 @@
...
+// run walks through the state machine for the lexer.
+func (lex *lexer) run() {
+	for state := lexContent; state != nil; {
+		state = state(lex)
+	}
+	close(lex.instructions)
+}
+
+// stateFn represents a state machine transition of the scanner going from one INSTRUCTION to the next.
+type stateFn func(*lexer) stateFn
```
huanjani, line 158: 🤯 👏🏼

huanjani (approved):

internal/pkg/docker/dockerfile/lex.go
Lou1415926: Is the extra `-1` for space `" "`?

internal/pkg/docker/dockerfile/lex.go
Lou1415926, line 158: computer science working 💻 🔮

internal/pkg/docker/dockerfile/dockerfile.go
efekarakus: Kept it as is because i removed `name` from the lexer following Wanxian's comment, so now I'm writing the name information in this function!

internal/pkg/docker/dockerfile/lex.go
efekarakus: renamed to `markerExposeInstr` and `markerHealthCheckInstr`!

internal/pkg/docker/dockerfile/lex.go
efekarakus: lol oops, changed the const names to `markerExposeInstr` and `markerHealthCheckInstr`

internal/pkg/docker/dockerfile/lex.go
efekarakus: Good point! removed

internal/pkg/docker/dockerfile/lex.go
efekarakus: `TrimPrefix` as is wouldn't work  because it's possible that the instruction has leading spaces, for example:
```
            EXPOSE 8080
```

But good point! to be safe with the `idx` I added the following guard-statement:
```go
if !strings.Contains(normalized, instrMarker) {
	return line
}
```

internal/pkg/docker/dockerfile/lex.go
efekarakus: Done!

efekarakus: omg that was a bug! The test cases passed because all my line-wraps had an extra space `abc \` instead of `abc\`.
I've added a new test case and removed the -1

internal/pkg/docker/dockerfile/lex.go
efekarakus: I mostly just followed the logic from buildkit, so I'm hesitant to change it in case it results in poor parsing

Lou1415926 (approved):
:shipit: !!!

PR #2869, opened 2021-09-23
chore(manifest): move platform validation and container dependencies 

iamhopaul123:
part of #2818


internal/pkg/manifest/validate.go
```diff
@@ -224,6 +265,45 @@ func (i *Image) Validate() error {
...
+}
+
+// Validate returns nil if DependsOn is configured correctly.
+func (d *DependsOn) Validate() error {
+	if d == nil {
+		return nil
+	}
+	for _, v := range *d {
+		status := strings.ToUpper(v)
+		var validStatus bool
```
```suggestion
		var isValid bool
```

internal/pkg/manifest/validate.go
```diff
@@ -623,9 +713,31 @@ func (s *SidecarConfig) Validate() error {
...
+	for name, status := range s.DependsOn {
+		if ok, err := isSidecarEssentialStatus(strings.ToUpper(status), name); s.isEssential() && !ok {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *SidecarConfig) isEssential() bool {
+	return s.Essential == nil || aws.BoolValue(s.Essential)
```
Lou1415926: ooo this reminds me. Should we document somewhere that the CloudFormation default for `essential` is true?

internal/pkg/manifest/validate.go
```diff
@@ -734,3 +846,62 @@ func (d *DeadLetterQueue) Validate() error {
...
+		return err
+	}
+	acyclic, cycle := dependencies.IsAcyclic()
+	if acyclic {
+		return nil
+	}
+	if len(cycle) == 1 {
+		return fmt.Errorf("container %s cannot depend on itself", cycle[0])
+	}
+	sort.SliceStable(cycle, func(i, j int) bool { return cycle[i] < cycle[j] })
```
Lou1415926, line 882: Is this so that the error message is consistent to ensure stable unit tests? If this is correct perhaps we can add a comment for it.

internal/pkg/manifest/validate.go
```diff
@@ -734,3 +846,62 @@ func (d *DeadLetterQueue) Validate() error {
...
+	return fmt.Errorf("circular container dependency chain includes the following containers: %s", cycle)
+}
+
+func buildDependencyGraph(opts validateNoCircularDependenciesOpts) (*graph.Graph, error) {
+	dependencyGraph := graph.NewGraph()
+	// Add any sidecar dependencies.
+	for name, sidecar := range opts.sidecarConfig {
+		for dep := range sidecar.DependsOn {
+			if _, ok := opts.sidecarConfig[dep]; !ok && dep != opts.mainContainerName {
+				return nil, errInvalidContainer
```
Lou1415926: Should we include `dep` in the error message somewhere?

internal/pkg/manifest/validate.go
```diff
@@ -8,18 +8,31 @@ import (
...
 var (
 	intRangeBandRegexp = regexp.MustCompile(`^(\d+)-(\d+)$`)
+
+	essentialContainerValidStatuses = []string{dependsOnStart, dependsOnHealthy}
+	dependsOnValidStatuses          = []string{dependsOnStart, dependsOnComplete, dependsOnSuccess, dependsOnHealthy}
 )
 
 // Validate returns nil if LoadBalancedWebServiceConfig is configured correctly.
-func (l *LoadBalancedWebServiceConfig) Validate() error {
+func (l *LoadBalancedWebServiceConfig) Validate(name string) error {
```
Lou1415926: What do you think of having an additional private field in the struct `LoadBalancedWebServiceConfig`?
```
type LoadBalancedWebServiceConfig struct {
     name string
     ...OtherCapitalizedExportedFields
}
```
Also, in this way `RequestDrivenWebService` wouldn't be affected by this PR. Currently it needs `name string` parameter in `Validate` even if it doesn't need to check for dependencies at all.

internal/pkg/manifest/validate.go
```diff
@@ -734,3 +846,62 @@ func (d *DeadLetterQueue) Validate() error {
...
+}
+
+func isSidecarEssentialStatus(status string, container string) (bool, error) {
+	if status == dependsOnStart {
+		return true, nil
+	}
+	return false, fmt.Errorf("essential sidecar container dependencies can only have status < %s >", dependsOnStart)
+}
+
+func isEssentialStatus(status string, container string) (bool, error) {
```
Lou1415926: 🤔  is `container` used in this function?

internal/pkg/manifest/validate.go
efekarakus: Hmm this unfortunately breaks the nice interface of every struct implementing a `Validate() error` interface. That was so nice 😭 It's also not clear that `name` in the method signature refers to a container name.

Do you know if `name` is ever different than: `l.WorkloadProps.Name`?

internal/pkg/manifest/validate.go
```diff
@@ -49,14 +62,21 @@ func (l *LoadBalancedWebServiceConfig) Validate() error {
...
-			return fmt.Errorf(`validate taskdef_overrides[%d]: %w`, ind, err)
+			return fmt.Errorf(`validate "taskdef_overrides[%d]": %w`, ind, err)
 		}
 	}
+	if err = validateDependencies(validateDependenciesOpts{
+		sidecarConfig:     l.Sidecars,
+		imageConfig:       &l.ImageConfig.Image,
+		mainContainerName: name,
+	}); err != nil {
+		return fmt.Errorf("validate dependencies: %w", err)
```
```suggestion
		return fmt.Errorf("validate container dependencies: %w", err)
```

internal/pkg/manifest/validate.go
```diff
@@ -49,14 +62,21 @@ func (l *LoadBalancedWebServiceConfig) Validate() error {
 	}
 	for ind, taskDefOverride := range l.TaskDefOverrides {
 		if err = taskDefOverride.Validate(); err != nil {
-			return fmt.Errorf(`validate taskdef_overrides[%d]: %w`, ind, err)
+			return fmt.Errorf(`validate "taskdef_overrides[%d]": %w`, ind, err)
 		}
 	}
+	if err = validateDependencies(validateDependenciesOpts{
```
efekarakus: nit: maybe instead of defining an additional type in the pkg
```suggestion
	if err = validateContainerDeps(containerName,  &l.ImageConfig.Image, l.Sidecars)
```

internal/pkg/manifest/validate.go
```diff
@@ -224,6 +265,30 @@ func (i *Image) Validate() error {
...
+		status := strings.ToUpper(v)
+		var validStatus bool
+		for _, allowed := range dependsOnValidStatuses {
+			if status == allowed {
+				validStatus = true
+				break
+			}
+		}
+		if !validStatus {
+			return fmt.Errorf("container dependency status must be one of < %s | %s | %s | %s >", dependsOnStart, dependsOnComplete, dependsOnSuccess, dependsOnHealthy)
```
```suggestion
			return fmt.Errorf("container dependency status must be one of %s", english.WordSeries([]string{dependsOnStart, dependsOnComplete, dependsOnSuccess, dependsOnHealthy}, "or"))
```

internal/pkg/manifest/validate.go
```diff
@@ -734,3 +812,108 @@ func (d *DeadLetterQueue) Validate() error {
...
+		aws.BoolValue(opts.sidecarConfig[name].Essential)
+}
+
+func isValidEssentialStatus(name, status string) error {
+	for _, allowed := range essentialContainerValidStatuses {
+		if status == allowed {
+			return nil
+		}
+	}
+	return fmt.Errorf("essential container %s can only have status < %s | %s >", name, dependsOnStart, dependsOnHealthy)
```
efekarakus: samesies here with `english.WordSeries` 🥺

internal/pkg/manifest/validate.go
```diff
@@ -734,3 +812,108 @@ func (d *DeadLetterQueue) Validate() error {
...
+}
+
+func validateNoCircularDependencies(opts validateDependenciesOpts) error {
+	dependencies, err := buildDependencyGraph(opts)
+	if err != nil {
+		return err
+	}
+	acyclic, cycle := dependencies.IsAcyclic()
```
efekarakus: What do you think of reverting the function signature? It's a bit more standard (like `v, ok := map[k]`)
```go
cycle, ok := dependencies.IsAcyclic()
```

internal/pkg/term/graph/graph.go
```diff
@@ -0,0 +1,96 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package graph provides functionality for graphes.
```
efekarakus: there is no 'e' 😆 
```suggestion
// Package graph provides functionality for directed graphs.
```

internal/pkg/term/graph/graph.go
```diff
@@ -0,0 +1,96 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
efekarakus: Can we move this out of `term` it's not terminal related functionality 🧐 

Maybe just directly under `internal/pkg`

internal/pkg/term/graph/graph.go
```diff
@@ -0,0 +1,96 @@
...
+
+const (
+	white nodeColor = iota + 1
+	gray
+	black
+)
+
+// Graph represents a directed graph.
+type Graph struct {
+	Nodes map[string][]string
```
efekarakus: We can refactor this to use beyond just `string` with generics once we get go1.18 :P

internal/pkg/term/graph/graph.go
```diff
@@ -0,0 +1,96 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package graph provides functionality for graphes.
+package graph
+
+type nodeColor int
+
+const (
```
efekarakus: can we add comments here explaining the purpose of the colors 🙏

internal/pkg/term/graph/graph.go
```diff
@@ -0,0 +1,96 @@
...
+	black
+)
+
+// Graph represents a directed graph.
+type Graph struct {
+	Nodes map[string][]string
+}
+
+// NewGraph initiates a new Graph.
+func NewGraph() *Graph {
```
efekarakus:
```suggestion
func New() *Graph {
```

internal/pkg/term/graph/graph.go
```diff
@@ -0,0 +1,96 @@
...
+	// Add origin node if doesn't exist.
+	if _, ok := g.Nodes[fromNode]; !ok {
+		g.Nodes[fromNode] = []string{}
+	}
+	// Check if edge exists between from and to Nodes.
+	for _, node := range g.Nodes[fromNode] {
+		if node == toNode {
+			hasNode = true
+		}
+	}
```
efekarakus: hmm i wonder if `Nodes map[string][]string` is the most efficient way of implementing this 🤔 

what if we do:
```go
type node struct {
   name string
   nodes map[string]node
}

type Graph struct {
   nodes map[string]node
}
```

internal/pkg/manifest/validate.go
iamhopaul123: I think they are the same one.

internal/pkg/manifest/validate.go
```diff
@@ -734,3 +812,108 @@ func (d *DeadLetterQueue) Validate() error {
 func (*OverrideRule) Validate() error {
 	return nil
 }
+
+type validateDependenciesOpts struct {
+	mainContainerName string
+	sidecarConfig     map[string]*SidecarConfig
+	imageConfig       *Image
```
Lou1415926: Question 🙋🏼‍♀️ ! Could this be kept as a non-pointer?

internal/pkg/manifest/validate.go
```diff
@@ -734,3 +812,108 @@ func (d *DeadLetterQueue) Validate() error {
...
+	if err := validateImageDependsOnStatus(opts); err != nil {
+		return err
+	}
+	if err := validateSidecarsDependsOnStatus(opts); err != nil {
+		return err
+	}
+	return validateNoCircularDependencies(opts)
+}
+
+func validateImageDependsOnStatus(opts validateDependenciesOpts) error {
```
```suggestion
func validateImageDependsOnEssentialContainerStatus(opts validateDependenciesOpts) error {
```

Another thought: since `validateImageDependsOnStatus` and `validateSidecarsDependsOnStatus` basically only validate whether an essential container dependency is specified correctly, what do you think of refactoring then into a function named
```go
validateEssentialContainerDependencies(dependsOn DependsOn)
```
and call by `validateEssentialContainerDependencies(i.imageConfig.DependsOn)` and `validateEssentialContainerDependencies(s.DependsOn)`

internal/pkg/manifest/validate.go
```diff
@@ -734,3 +812,108 @@ func (d *DeadLetterQueue) Validate() error {
...
+	if opts.imageConfig == nil || opts.imageConfig.DependsOn == nil {
+		return nil
+	}
+	for name, status := range opts.imageConfig.DependsOn {
+		if !isEssentialContainer(name, opts) {
+			continue
+		}
+		if err := isValidEssentialStatus(name, strings.ToUpper(status)); err != nil {
+			return fmt.Errorf("validate %s container dependencies status: %w", opts.mainContainerName, err)
```
Lou1415926: What do you think of this error message:
```suggestion
			return fmt.Errorf("validate %s container "depends_on": %w", opts.mainContainerName, err)
```
and samesies in `validateSidecarsDependsOnStatus`

internal/pkg/manifest/validate.go
iamhopaul123: good idea!

internal/pkg/manifest/validate.go
iamhopaul123: make sense

internal/pkg/manifest/validate.go
```diff
@@ -132,14 +159,21 @@ func (w *WorkerServiceConfig) Validate() error {
 	}
 	for ind, taskDefOverride := range w.TaskDefOverrides {
 		if err = taskDefOverride.Validate(); err != nil {
-			return fmt.Errorf(`validate taskdef_overrides[%d]: %w`, ind, err)
+			return fmt.Errorf(`validate "taskdef_overrides[%d]": %w`, ind, err)
 		}
 	}
+	if err = validateNoCircularDependencies(validateDependenciesOpts{
```
Lou1415926: Question 🙋🏼‍♀️  Why `validateNoCircularDependencies` instead of `validateDependencies`, like in lb and backend service?

internal/pkg/term/graph/graph.go
iamhopaul123: hmm I don't think this is a very good idea to represent a graph since it is not like a tree and has no "root node". To make it more efficient maybe
```golang
type Graph struct {
	Nodes map[string]Neighbors
}

type Neighbors map[string]bool
```
Do you think it is better?

internal/pkg/manifest/validate.go
```diff
@@ -734,3 +813,110 @@ func (d *DeadLetterQueue) Validate() error {
...
+	}
+	return fmt.Errorf("essential container %s can only have status %s", name, english.WordSeries([]string{dependsOnStart, dependsOnHealthy}, "or"))
+}
+
+func validateNoCircularDependencies(opts validateDependenciesOpts) error {
+	dependencies, err := buildDependencyGraph(opts)
+	if err != nil {
+		return err
+	}
+	cycle, acyclic := dependencies.IsAcyclic()
```
Lou1415926: tiny nit: maybe `acyclic` could be renamed as `ok` since given the function name `IsAcyclic`, it's clear that `ok` means it is indeed acyclic.

internal/pkg/manifest/validate.go
efekarakus: Wouldn't this: https://github.com/aws/copilot-cli/pull/2869/files#r716118843 solve this problem? 
Or am I missing something 🤔

Lou1415926: It doesn't look like `LoadBalancedWebServiceConfig` or any workload config has access to `WorkloadProps` though 😢

internal/pkg/manifest/validate.go
iamhopaul123: hmm the reason is because we have https://github.com/aws/copilot-cli/blob/807ae3f5fc370397c3220cd53464db087176f26e/internal/pkg/manifest/validate.go#L270 and also we are not validating only in either `Image` or `Sidecar` level. This validation serves more like a general dependency check as opposed to just one field.

internal/pkg/manifest/validate.go
iamhopaul123: 🤦

internal/pkg/manifest/validate.go
iamhopaul123: done!

site/content/docs/developing/sidecars.en.md
```diff
@@ -14,36 +14,6 @@ There are two ways of adding sidecars using the Copilot manifest: by specifying
 ### General sidecars
 You'll need to provide the URL for the sidecar image. Optionally, you can specify the port you'd like to expose and the credential parameter for [private registry](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/private-auth.html).
 
-``` yaml
```
Lou1415926, line 17: I feel like this manifest example still has its merit in showing all configurable fields at once 🤔

internal/pkg/graph/graph.go
```diff
@@ -0,0 +1,97 @@
...
+// Package graph provides functionality for directed graphs.
+package graph
+
+// nodeColor denotes the color of a node when running DFS in a graph.
+type nodeColor int
+
+const (
+	white nodeColor = iota + 1
+	gray
+	black
```
Lou1415926:
```suggestion
	white nodeColor = iota + 1 // A node is colored white if it is unvisited.
	gray      // A node is colored gray if it's visited but isn't finished (hence can potentially lead to a cycle).
	black     // A node is colored black if it's finished (hence not possible to lead to a cycle).
```
Or we could just rename them as `unvisited`, `visited`, and `skip`/`finished`, instead of adding comments 💭 .

internal/pkg/deploy/cloudformation/stack/validate.go
```diff
@@ -32,28 +32,25 @@ var (
...
+	errAccessPointWithRootDirectory = errors.New("`root_directory` must be empty or \"/\" when `access_point` is specified")
+	errAccessPointWithoutIAM        = errors.New("`iam` must be true when `access_point` is specified")
+	errUIDWithNonManagedFS          = errors.New("UID and GID cannot be specified with non-managed EFS")
+	errInvalidUIDGIDConfig          = errors.New("must specify both UID and GID, or neither")
+	errInvalidEFSConfig             = errors.New("bad EFS configuration: cannot specify both bool and config")
+	errReservedUID                  = errors.New("UID must not be 0")
+	errInvalidContainer             = errors.New("container dependency does not exist")
+	errInvalidDependsOnStatus       = fmt.Errorf("container dependency status must be one of < %s | %s | %s | %s >", dependsOnStart, dependsOnComplete, dependsOnSuccess, dependsOnHealthy)
+	errEssentialContainerStatus     = fmt.Errorf("essential container dependencies can only have status < %s | %s >", dependsOnStart, dependsOnHealthy)
```
efekarakus, line 43:
```suggestion
	errInvalidDependsOnStatus       = fmt.Errorf("container dependency status must be one of %s", english.WordSeries([]string{dependsOnStart, dependsOnComplete, dependsOnSuccess, dependsOnHealthy}, "or"))
	errInvalidEssentialContainerDependsOnStatus     = fmt.Errorf("essential container dependencies can only have status %s", english.WordSeries([]string{dependsOnStart, dependsOnHealthy}, "or"))
```

internal/pkg/deploy/cloudformation/stack/validate.go
```diff
@@ -200,7 +188,7 @@ func isEssentialStatus(status string, container string, c convertSidecarOpts) (b
 		if status == dependsOnStart {
 			return true, nil
 		}
-		return false, errEssentialSidecarStatus
+		return false, errEssentialContainerStatus
```
efekarakus, line 191: nit: Unrelated to your code changes, but this function name is misleading 😅   
I'd think a fn named `isEssentialStatus(status string, ...)` would do something like `return status == "essential"`

Can we rename/refactor this? Thank you so much, I honestly find it hard to understand what this function is doing

Lou1415926 (approved):
LGTM 👍🏼

internal/pkg/graph/graph.go
```diff
@@ -0,0 +1,97 @@
...
+
+// New initiates a new Graph.
+func New() *Graph {
+	return &Graph{
+		Nodes: make(map[string][]string),
+	}
+}
+
+// Add adds a connection between two Nodes.
+func (g *Graph) Add(fromNode, toNode string) {
```
efekarakus: nit: Just something to consider: https://dave.cheney.net/2019/09/24/be-wary-of-functions-which-take-several-parameters-of-the-same-type 

When reading the code:
```go
g.Add("efe", "penghao") // it's not clear which one is the from and to 
```
Maybe, an alternative:
```go
type Edge struct {
  From string
  To string
}
func (g *Graph) Add(edge Edge) { ... }
```

internal/pkg/graph/graph.go
```diff
@@ -0,0 +1,97 @@
...
+	hasNode := false
+	// Add origin node if doesn't exist.
+	if _, ok := g.Nodes[fromNode]; !ok {
+		g.Nodes[fromNode] = []string{}
+	}
+	// Check if edge exists between from and to Nodes.
+	for _, node := range g.Nodes[fromNode] {
+		if node == toNode {
+			hasNode = true
+		}
```
efekarakus: nit:
```suggestion
		if node == toNode {
			return
		}
```

site/content/docs/developing/sidecars.en.md
iamhopaul123: But the tricky thing is we don't have any precedence and if we add new fields we would need to update two places. Maybe the example below is good enough like what we have for in the manifest doc? https://github.com/aws/copilot-cli/pull/2869/files/807ae3f5fc370397c3220cd53464db087176f26e..3f53a0aadc342e33c580206b0534b65bfba71996#diff-379c5e41cad752c12c02bb73f22823f4cbe1d74b9168750d82ef3342cdde7ed5L73-L79

internal/pkg/manifest/validate.go
```diff
@@ -8,14 +8,28 @@ import (
...
+	dependsOnStart    = "START"
+	dependsOnComplete = "COMPLETE"
+	dependsOnSuccess  = "SUCCESS"
+	dependsOnHealthy  = "HEALTHY"
 )
 
 var (
 	intRangeBandRegexp = regexp.MustCompile(`^(\d+)-(\d+)$`)
+
+	essentialContainerValidStatuses = []string{dependsOnStart, dependsOnHealthy}
```
```suggestion
	essentialContainerDependsOnValidStatuses = []string{dependsOnStart, dependsOnHealthy}
```

internal/pkg/manifest/validate.go
```diff
@@ -734,3 +813,93 @@ func (d *DeadLetterQueue) Validate() error {
...
+			isEssential: config.Essential == nil || aws.BoolValue(config.Essential),
+		}
+	}
+	if err := validateDependsOnEssentialStatus(containerDependencies); err != nil {
+		return err
+	}
+	return validateNoCircularDependencies(opts)
+}
+
+func validateDependsOnEssentialStatus(deps map[string]containerDependency) error {
```
```suggestion
func validateDepsForEssentialContainers(deps map[string]containerDependency) error {
```

internal/pkg/manifest/job.go
```diff
@@ -36,6 +36,7 @@ type ScheduledJob struct {
 
 // ScheduledJobConfig holds the configuration for a scheduled job
 type ScheduledJobConfig struct {
+	name                    string
```
efekarakus, line 39: Are we setting these values anywhere?

internal/pkg/manifest/validate.go
```diff
@@ -734,3 +813,93 @@ func (d *DeadLetterQueue) Validate() error {
...
+			}
+			if err := isValidEssentialStatus(dep, strings.ToUpper(status)); err != nil {
+				return fmt.Errorf("validate %s container dependencies status: %w", name, err)
+			}
+		}
+	}
+	return nil
+}
+
+func isValidEssentialStatus(name, status string) error {
```
efekarakus: Can we make these function names a bit more specific, it's hard to tell what they are doing 🥺 
```go
func validateEssentialContainerDependency(containerName, condition string) error 
```
I know that these are already existing fn names, but it's a good opportunity to clean them up

internal/pkg/manifest/job.go
iamhopaul123: Not yet until we actually use this new manifest validation

efekarakus: Ok! I'm just worried that we'll forget setting it and everything will blow :P as long as it's written in your backlog I'm fine with it

internal/pkg/deploy/cloudformation/stack/validate.go
iamhopaul123: hmm i don't think we need to make any change to this pkg since we'll remove them very soon.

internal/pkg/deploy/cloudformation/stack/validate.go
iamhopaul123: same here

efekarakus (approved):

PR #2878, opened 2021-09-28
chore(manifest): move more validation from template pkg

iamhopaul123:
Part of #2818. This PR adds validation for
- task def override
- sidecar mount points
- storage config
- publish/subscribe

Also make `topic.queue` to able to accept either bool or map as input so as users can better specify a topic specific queue with default config.

### DIFF
--- a/internal/pkg/cli/validate.go
+++ b/internal/pkg/cli/validate.go
@@ -679,12 +679,10 @@ func validateSubscriptionKey(val interface{}) error {
 	if err != nil {
 		return errSubscribeBadFormat
 	}
-	err = validatePubSubName(aws.StringValue(sub.Name))
-	if err != nil {
+	if err := validatePubSubName(aws.StringValue(sub.Name)); err != nil {
 		return fmt.Errorf("invalid topic subscription topic name `%s`: %w", aws.StringValue(sub.Name), err)
 	}
-	err = basicNameValidation(aws.StringValue(sub.Service))
-	if err != nil {
+	if err = basicNameValidation(aws.StringValue(sub.Service)); err != nil {
 		return fmt.Errorf("invalid topic subscription service name `%s`: %w", aws.StringValue(sub.Service), err)
 	}
 	return nil
--- a/internal/pkg/cli/validate.go
+++ b/internal/pkg/cli/validate.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/spf13/afero"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/copilot-cli/internal/pkg/addon"
 	"github.com/aws/copilot-cli/internal/pkg/aws/apprunner"
@@ -678,13 +679,13 @@ func validateSubscriptionKey(val interface{}) error {
 	if err != nil {
 		return errSubscribeBadFormat
 	}
-	err = validatePubSubName(sub.Name)
+	err = validatePubSubName(aws.StringValue(sub.Name))
 	if err != nil {
-		return fmt.Errorf("invalid topic subscription topic name `%s`: %w", sub.Name, err)
+		return fmt.Errorf("invalid topic subscription topic name `%s`: %w", aws.StringValue(sub.Name), err)
 	}
-	err = basicNameValidation(sub.Service)
+	err = basicNameValidation(aws.StringValue(sub.Service))
 	if err != nil {
-		return fmt.Errorf("invalid topic subscription service name `%s`: %w", sub.Service, err)
+		return fmt.Errorf("invalid topic subscription service name `%s`: %w", aws.StringValue(sub.Service), err)
 	}
 	return nil
 }
@@ -715,7 +716,7 @@ func validateTopicsExist(subscriptions []manifest.TopicSubscription, topicARNs [
 	}
 
 	for _, ts := range subscriptions {
-		topicName := fmt.Sprintf(resourceNameFormat, app, env, ts.Service, ts.Name)
+		topicName := fmt.Sprintf(resourceNameFormat, app, env, aws.StringValue(ts.Service), aws.StringValue(ts.Name))
 		if !contains(topicName, validTopicResources) {
 			return fmt.Errorf(fmtErrTopicSubscriptionNotAllowed, topicName, env)
 		}
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -725,11 +725,13 @@ func convertTopicSubscription(t manifest.TopicSubscription, url, accountID, app,
 	if err != nil {
 		return nil, fmt.Errorf(`invalid topic subscription "%s": %w`, aws.StringValue(t.Name), err)
 	}
-	queue, err := convertQueue(t.Queue)
+	queue, err := convertQueue(t.Queue.Advanced)
 	if err != nil {
 		return nil, fmt.Errorf(`invalid topic subscription "%s": %w`, aws.StringValue(t.Name), err)
 	}
-
+	if aws.BoolValue(t.Queue.Enabled) {
+		queue = &template.SQSQueue{}
+	}
 	return &template.TopicSubscription{
 		Name:    t.Name,
 		Service: t.Service,
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -725,12 +725,16 @@ func convertTopicSubscription(t manifest.TopicSubscription, url, accountID, app,
 	if err != nil {
 		return nil, fmt.Errorf(`invalid topic subscription "%s": %w`, aws.StringValue(t.Name), err)
 	}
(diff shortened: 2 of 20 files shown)


internal/pkg/manifest/validate.go
```diff
@@ -810,7 +874,16 @@ func (d *DeadLetterQueue) Validate() error {
...
 // Validate returns nil if OverrideRule is configured correctly.
-func (*OverrideRule) Validate() error {
+func (r *OverrideRule) Validate() error {
+	if r == nil {
+		return nil
+	}
+	for _, s := range invalidTaskDefOverridePathRegexp {
+		re := regexp.MustCompile(fmt.Sprintf(`^%s$`, s))
+		if re.MatchString(r.Path) {
+			return fmt.Errorf("%s cannot be overidden with a custom value", s)
```
Lou1415926: What do you think about
```suggestion
			return fmt.Errorf(`"%s" cannot be overridden with a custom value`, s)
```

internal/pkg/manifest/validate.go
```diff
@@ -909,3 +982,44 @@ func buildDependencyGraph(opts validateDependenciesOpts) (*graph.Graph, error) {
...
+	}
+	// Name must contain letters, numbers, and can't use special characters besides underscores and hyphens.
+	if !awsSNSTopicRegexp.MatchString(name) {
+		return fmt.Errorf(`"name" can only contain letters, numbers, underscores, and hypthens`)
+	}
+	return nil
+}
+
+func isValidSubSvcName(name string) bool {
+	if !awsNameRegexp.MatchString(name) {
```
Lou1415926, line 1021: Do we need to check against `name == ""` here?

internal/pkg/cli/validate.go
```diff
@@ -678,13 +679,13 @@ func validateSubscriptionKey(val interface{}) error {
 	if err != nil {
 		return errSubscribeBadFormat
 	}
-	err = validatePubSubName(sub.Name)
+	err = validatePubSubName(aws.StringValue(sub.Name))
 	if err != nil {
```
efekarakus: nit: can we scope these errors to the if-block?
```go
if err := validatePubSubName(...); err != nil {
```

internal/pkg/manifest/validate.go
```diff
@@ -660,6 +682,9 @@ func (e *EFSVolumeConfiguration) Validate() error {
 			conditionalFields: []string{"gid"},
 		}
 	}
+	if e.UID != nil && *e.UID == 0 {
```
Lou1415926, line 685: nit: what do you think of 
```suggestion
	if aws.Uint32Value(e.UID) {
```
The validation for `TopicSubscription`'s `Name` field also seems to be taking advantage of this function and the fact that it returns a zero value if the pointer is `nil`.

internal/pkg/manifest/validate.go
```diff
@@ -810,7 +874,16 @@ func (d *DeadLetterQueue) Validate() error {
...
 
 // Validate returns nil if OverrideRule is configured correctly.
-func (*OverrideRule) Validate() error {
+func (r *OverrideRule) Validate() error {
+	if r == nil {
+		return nil
+	}
+	for _, s := range invalidTaskDefOverridePathRegexp {
+		re := regexp.MustCompile(fmt.Sprintf(`^%s$`, s))
+		if re.MatchString(r.Path) {
```
efekarakus, line 891: I'd love for us to go through this file at some point and remove basic regex's from the codebase

internal/pkg/manifest/validate.go
```diff
@@ -660,6 +682,9 @@ func (e *EFSVolumeConfiguration) Validate() error {
 			conditionalFields: []string{"gid"},
 		}
 	}
+	if e.UID != nil && *e.UID == 0 {
+		return fmt.Errorf(`"uid" must not be 0`)
```
Lou1415926, line 686:
```suggestion
		return fmt.Errorf(`"uid" must be specified and must not 0`)
```

internal/pkg/manifest/validate.go
```diff
@@ -909,3 +982,44 @@ func buildDependencyGraph(opts validateDependenciesOpts) (*graph.Graph, error) {
 	}
 	return dependencyGraph, nil
 }
+
+// Validate that paths contain only an approved set of characters to guard against command injection.
+// We can accept 0-9A-Za-z-_.
+func validatePath(input string) error {
```
efekarakus: can we make the regex and this function name more specific? is this the `http` path or `volumes` path it's not clear 🥺

efekarakus (commented):
Looks good! just a request to make sure we have our regex names and function names follow: https://a-nickels-worth.blogspot.com/2016/04/a-guide-to-naming-variables.html

internal/pkg/manifest/validate.go
iamhopaul123: Yeah agree. As for this PR it is safe I think since we are mostly strictly moving things around.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -725,11 +725,13 @@ func convertTopicSubscription(t manifest.TopicSubscription, url, accountID, app,
...
 	}
-	queue, err := convertQueue(t.Queue)
+	queue, err := convertQueue(t.Queue.Advanced)
 	if err != nil {
 		return nil, fmt.Errorf(`invalid topic subscription "%s": %w`, aws.StringValue(t.Name), err)
 	}
-
+	if aws.BoolValue(t.Queue.Enabled) {
+		queue = &template.SQSQueue{}
+	}
```
Lou1415926: What do you think of this?
```suggestion
	if aws.BoolValue(t.Queue.Enabled) {
           return &template.TopicSubscription{
		Name:    t.Name,
		Service: t.Service,
		Queue:   &template.SQSQueue{},
	}, nil
}
if queue, err := convertQueue(t.Queue.Advanced) ...
```

internal/pkg/manifest/validate.go
iamhopaul123: We check and make sure it is not empty here https://github.com/aws/copilot-cli/pull/2878/files/74860f1a6800f1e520a746a945026027e06fe9f5#diff-5fc13795122fe030799a09ef956a8b66cec3d40bf1a81266323edbf3626b5685R843

internal/pkg/manifest/validate.go
iamhopaul123: They are not quite the same...for the former if `e.UID` is nil then we won't do the validation and let it go, while the latter we'll throw an error which we shouldn't

internal/pkg/manifest/validate.go
iamhopaul123: Hmm but it is not a required field.

site/content/docs/manifest/worker-service.en.md
```diff
@@ -96,6 +96,9 @@ Required. The name of the SNS topic to subscribe to.
 <span class="parent-field">topic.</span><a id="topic-service" href="#topic-service" class="field">`service`</a> <span class="type">String</span>
 Required. The service this SNS topic is exposed by. Together with the topic name, this uniquely identifies an SNS topic in the copilot environment.
 
+<span class="parent-field">topic.</span><a id="topic-queue" href="#topic-queue" class="field">`queue`</a> <span class="type">Boolean or Map</span>
+Optional. Define SQS queue for the topic. If specified as a boolean, only default queue configuration will be used. Specify this field as a map for more customization of this topic-specific queue.
```
Lou1415926: nit:
```suggestion
Optional. Specify SQS queue configuration for the topic. If specified as `true`, the queue will be created  with default configuration. Specify this field as a map for customization of certain attributes for this topic-specific queue.
```

internal/pkg/manifest/worker_svc.go
```diff
@@ -51,9 +57,44 @@ func (s *SubscribeConfig) IsEmpty() bool {
 
 // TopicSubscription represents the configurable options for setting up a SNS Topic Subscription.
 type TopicSubscription struct {
-	Name    *string  `yaml:"name"`
-	Service *string  `yaml:"service"`
-	Queue   SQSQueue `yaml:"queue"`
+	Name    *string        `yaml:"name"`
+	Service *string        `yaml:"service"`
+	Queue   SQSQueueOrBool `yaml:"queue"`
```
Lou1415926, line 62: What do you think of using `SQSQueueOrBool` for `SubscribeConfig` as well? 

I know the queue is always created. But it may feels strange for user to see an unmarshal error if they specify `queue: true` for the service-level queue.

We can also error out or ignore it if we see `queue: false` for that queue.

What do you think?

Lou1415926 (commented):
Thanks for the `SQSQueueOrBool` update! I'm sorry for my pea-sized brain 😢

internal/pkg/manifest/worker_svc.go
iamhopaul123: I did that before and then I reverted it back because it doesn't make sense to allow `queue: true` if it is a field that cannot be turned off (which is the fact that we are going to ignore `queue: false`). It should be quick for them to figure it out as well once they read the doc.

efekarakus (approved):

Lou1415926 (approved):

PR #2879, opened 2021-09-28
test(e2e): add worker service test suite

efekarakus:
The test ensures:
1. The worker service receives and consumes sqs messages
2. Can contact other services via service discovery

### DIFF
--- a/e2e/worker/frontend/server.js
+++ b/e2e/worker/frontend/server.js
@@ -1,0 +1,38 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+'use strict';
+
+const express = require('express');
+const { SNSClient, PublishCommand } = require("@aws-sdk/client-sns");
+
+const PORT = 8080;
+const HOST = '0.0.0.0';
+const client = new SNSClient({ region: process.env.AWS_DEFAULT_REGION });
+const app = express();
+
+// Start the service waiting for an ack message.
+let status = 'waiting on acknowledgement';
+
+// Each health check request from the ALB will result in publishing an event.
+app.get('/', async (req, res) => {
+  const {events} = JSON.parse(process.env.COPILOT_SNS_TOPIC_ARNS);
+  const out = await client.send(new PublishCommand({
+    Message: "healthcheck",
+    TopicArn: events,
+  }));
+  console.log(JSON.stringify(out));
+  res.send('hello');
+});
+
+app.get('/status', (req, res) => {
+  res.send(status);
+});
+
+app.post('/ack', async (req, res) => {
+  status = 'consumed';
+  res.send('ok');
+});
+
+app.listen(PORT, HOST);
+console.log(`Running on http://${HOST}:${PORT}`);
--- a/e2e/worker/worker/index.js
+++ b/e2e/worker/worker/index.js
@@ -1,0 +1,41 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+const { SQSClient, ReceiveMessageCommand, DeleteMessageCommand } = require("@aws-sdk/client-sqs");
+const axios = require("axios");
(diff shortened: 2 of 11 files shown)


efekarakus:
Ran in my aws account:
```
Ran 11 of 11 Specs in 1756.561 seconds
SUCCESS! -- 11 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

Ginkgo ran 1 suite in 29m27.396748709s
Test Suite Passed
```

e2e/worker/worker_suite_test.go
```diff
@@ -0,0 +1,53 @@
...
+func TestWorker(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "App Runner Suite")
+}
+
+var _ = BeforeSuite(func() {
+	var err error
+	cli, err = client.NewCLI()
+	Expect(err).NotTo(HaveOccurred())
+	Expect(err).NotTo(HaveOccurred())
```
iamhopaul123:
```suggestion
```

e2e/worker/worker_suite_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	_, err = cli.SvcDelete(workerServiceName)
+	Expect(err).NotTo(HaveOccurred())
+
+	_, err = cli.EnvDelete(envName)
+	Expect(err).NotTo(HaveOccurred())
+
+	_, err = cli.AppDelete()
+	Expect(err).NotTo(HaveOccurred())
```
iamhopaul123: Why not just app delete? Is it because we have to remove them in a certain order?

e2e/worker/worker/index.js
```diff
@@ -0,0 +1,41 @@
...
+        QueueUrl: eventsQueue,
+        WaitTimeSeconds: 10,
+      }));
+      console.log(`results: ${JSON.stringify(out)}`);
+
+      if (out.Messages === undefined || out.Messages.length === 0) {
+        continue;
+      }
+
+      const resp = await axios.post(`http://frontend.${process.env.COPILOT_SERVICE_DISCOVERY_ENDPOINT}:8080/ack`);
```
iamhopaul123, line 30: Should we delete the message only after the post action is successful?

e2e/worker/frontend/server.js
```diff
@@ -0,0 +1,38 @@
...
+// Start the service waiting for an ack message.
+let status = 'waiting on acknowledgement';
+
+// Each health check request from the ALB will result in publishing an event.
+app.get('/', async (req, res) => {
+  const {events} = JSON.parse(process.env.COPILOT_SNS_TOPIC_ARNS);
+  const out = await client.send(new PublishCommand({
+    Message: "healthcheck",
+    TopicArn: events,
+  }));
```
iamhopaul123, line 23: Is it a problem if we accumulate too many messages in a queue before the worker service is able to consume them?

e2e/worker/worker_suite_test.go
efekarakus: Done!

e2e/worker/worker_suite_test.go
efekarakus: because i am stoopid and was just copy pasting without thinking 😂 . Done!

e2e/worker/worker/index.js
efekarakus: I don't think it particularly matters 🤷 because if we fail, and the messages never get deleted then `app delete` will delete the queue anyway with all of its messages.
The order shouldn't matter that much as long as we only make this request after receiving a real message

e2e/worker/frontend/server.js
efekarakus: Nope! SQS is cool 😎 

> A single Amazon SQS message queue can contain an unlimited number of messages. However, there is a 120,000 quota for the number of inflight messages for a standard queue and 20,000 for a FIFO queue. Messages are inflight after they have been received from the queue by a consuming component, but have not yet been deleted from the queue.

iamhopaul123 (approved):
:shipit:

Lou1415926 (approved):

PR #2895, opened 2021-10-04
test(manifest): add integ test making sure all structs implement Validate

iamhopaul123:
This PR
1. Add integ test to make sure every manifest struct or advanced type implements `Validate()`.
2. Fix a bug that `network.placement` should not be required even if `network` is not empty.

### DIFF
--- a/internal/pkg/manifest/validate.go
+++ b/internal/pkg/manifest/validate.go
@@ -44,7 +44,26 @@ var (
 )
 
 // Validate returns nil if LoadBalancedWebService is configured correctly.
-func (l *LoadBalancedWebService) Validate() error {
+func (l LoadBalancedWebService) Validate() error {
+	var err error
+	if err = l.LoadBalancedWebServiceConfig.Validate(); err != nil {
+		return err
+	}
+	if err = l.Workload.Validate(); err != nil {
+		return err
+	}
+	if err = validateContainerDeps(validateDependenciesOpts{
+		sidecarConfig:     l.Sidecars,
+		imageConfig:       l.ImageConfig.Image,
+		mainContainerName: aws.StringValue(l.Name),
+	}); err != nil {
+		return fmt.Errorf("validate container dependencies: %w", err)
+	}
+	return nil
+}
+
+// Validate returns nil if LoadBalancedWebServiceConfig is configured correctly.
+func (l LoadBalancedWebServiceConfig) Validate() error {
 	var err error
 	if err = l.ImageConfig.Validate(); err != nil {
 		return fmt.Errorf(`validate "image": %w`, err)
@@ -77,24 +96,30 @@ func (l *LoadBalancedWebService) Validate() error {
 			return fmt.Errorf(`validate "taskdef_overrides[%d]": %w`, ind, err)
 		}
 	}
-	name := aws.StringValue(l.Name)
-	if name == "" {
-		return &errFieldMustBeSpecified{
-			missingField: "name",
-		}
+	return nil
+}
+
+// Validate returns nil if BackendService is configured correctly.
+func (b BackendService) Validate() error {
+	var err error
+	if err = b.BackendServiceConfig.Validate(); err != nil {
+		return err
+	}
+	if err = b.Workload.Validate(); err != nil {
+		return err
 	}
 	if err = validateContainerDeps(validateDependenciesOpts{
-		sidecarConfig:     l.Sidecars,
-		imageConfig:       l.ImageConfig.Image,
-		mainContainerName: name,
+		sidecarConfig:     b.Sidecars,
+		imageConfig:       b.ImageConfig.Image,
+		mainContainerName: aws.StringValue(b.Name),
 	}); err != nil {
 		return fmt.Errorf("validate container dependencies: %w", err)
 	}
 	return nil
 }
 
-// Validate returns nil if BackendService is configured correctly.
-func (b *BackendService) Validate() error {
+// Validate returns nil if BackendServiceConfig is configured correctly.
+func (b BackendServiceConfig) Validate() error {
 	var err error
 	if err = b.ImageConfig.Validate(); err != nil {
 		return fmt.Errorf(`validate "image": %w`, err)
@@ -124,24 +149,19 @@ func (b *BackendService) Validate() error {
 			return fmt.Errorf(`validate "taskdef_overrides[%d]": %w`, ind, err)
 		}
 	}
-	name := aws.StringValue(b.Name)
-	if name == "" {
-		return &errFieldMustBeSpecified{
-			missingField: "name",
-		}
-	}
-	if err = validateContainerDeps(validateDependenciesOpts{
-		sidecarConfig:     b.Sidecars,
-		imageConfig:       b.ImageConfig.Image,
-		mainContainerName: name,
-	}); err != nil {
-		return fmt.Errorf("validate container dependencies: %w", err)
-	}
 	return nil
 }
 
 // Validate returns nil if RequestDrivenWebService is configured correctly.
-func (r *RequestDrivenWebService) Validate() error {
+func (r RequestDrivenWebService) Validate() error {
+	if err := r.RequestDrivenWebServiceConfig.Validate(); err != nil {
+		return err
+	}
+	return r.Workload.Validate()
+}
+
+// Validate returns nil if RequestDrivenWebServiceConfig is configured correctly.
+func (r RequestDrivenWebServiceConfig) Validate() error {
 	var err error
 	if err = r.ImageConfig.Validate(); err != nil {
 		return fmt.Errorf(`validate "image": %w`, err)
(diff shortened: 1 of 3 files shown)


internal/pkg/manifest/validate.go
```diff
@@ -44,7 +44,26 @@ var (
 )
 
 // Validate returns nil if LoadBalancedWebService is configured correctly.
-func (l *LoadBalancedWebService) Validate() error {
+func (l LoadBalancedWebService) Validate() error {
```
efekarakus, line 47: Is there a reason why we switched the receivers from pointers to values?

internal/pkg/manifest/validate.go
```diff
@@ -44,7 +44,26 @@ var (
...
+	var err error
```
efekarakus, line 48: nit: can we remove the variable with:
```go
if err := l...Validate();
```

internal/pkg/manifest/validate.go
```diff
@@ -548,29 +581,29 @@ func (a *AdvancedCount) Validate() error {
...
 	return nil
 }
 
 // Validate returns nil if Percentage is configured correctly.
-func (p *Percentage) Validate() error {
-	if p == nil {
-		return nil
-	}
-	if val := int(*p); val < 0 || val > 100 {
+func (p Percentage) Validate() error {
```
efekarakus, line 598: ok so by moving it to value receivers we're assuming the callers will do the nil-check from now on right?

internal/pkg/manifest/validate_integration_test.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+	}
+	var val validator
+	validatorType := reflect.TypeOf(&val).Elem()
+	// template.Parser is not a manifest struct.
+	var tpl template.Parser
+	templaterType := reflect.TypeOf(&tpl).Elem()
+	if !typ.Implements(templaterType) && !typ.Implements(validatorType) {
+		return fmt.Errorf(`%v does not implement "Validate()"`, typ)
+	}
```
efekarakus: I wonder if instead of ignoring `template.Parser` as a type we can check if the field has a `"yaml"` struct tag and if so check it? https://stackoverflow.com/a/23507821/1201381

What do you think?

internal/pkg/manifest/validate.go
iamhopaul123: Because it won't work otherwise since `Validate()` would be the method of `* LoadBalancedWebService`. Also I feel it doesn't make the code much cleaner with pointer receiver.

internal/pkg/manifest/validate.go
iamhopaul123: Hmm wouldn't this avoid defining `err` repetitively?

internal/pkg/manifest/validate.go
iamhopaul123: yeah that's right.

internal/pkg/manifest/validate_integration_test.go
iamhopaul123: good idea!

internal/pkg/manifest/validate_integration_test.go
```diff
@@ -69,16 +68,17 @@ func isValid(v reflect.Value) error {
...
+	if !typ.Implements(validatorType) {
 		return fmt.Errorf(`%v does not implement "Validate()"`, typ)
 	}
 	if typ.Kind() != reflect.Struct {
 		return nil
 	}
-	for i := 0; i < v.NumField(); i++ {
+	for i := 0; i < typ.NumField(); i++ {
+		// Skip if it is not a manifest yaml field.
+		if string(typ.Field(i).Tag) == "" {
```
efekarakus: small refactor 🥺 
```suggestion
		typ.Field(i).Tag.Get("yaml") == "" {
```

efekarakus (approved):

internal/pkg/manifest/validate_integration_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+func isValid(v reflect.Value) error {
+	typ := v.Type()
+	if typ.Kind() == reflect.Ptr {
+		typ = typ.Elem()
+	}
+	for _, k := range basicKinds {
+		if typ.Kind() == k {
```
Lou1415926: Do we need to care about whether aliases to basic kinds such as `PlatformString` implements `Validate()`?

internal/pkg/manifest/validate_integration_test.go
```diff
@@ -0,0 +1,87 @@
...
+	validatorType := reflect.TypeOf(&val).Elem()
+	if !typ.Implements(validatorType) {
+		return fmt.Errorf(`%v does not implement "Validate()"`, typ)
+	}
+	if typ.Kind() != reflect.Struct {
+		return nil
+	}
+	for i := 0; i < typ.NumField(); i++ {
+		// Skip if it is not a manifest yaml field.
+		if typ.Field(i).Tag.Get("yaml") == "" {
```
Lou1415926: What about fields such as `HealthCheckArgs` that does not have a tag?

Lou1415926 (commented):
Two questions 🤚🏼

internal/pkg/manifest/validate_integration_test.go
iamhopaul123: yeah we do. Only basic types don't need to

internal/pkg/manifest/validate_integration_test.go
iamhopaul123: That's true. I guess I'd rather revert this change @efekarakus

internal/pkg/manifest/validate_integration_test.go
Lou1415926: Yes but `PlatformString`'s kind is `string` right? Maybe we should check the `Type()` here instead of `Kind()`?

internal/pkg/manifest/validate_integration_test.go
efekarakus: 😭 I'm just not a fan because it feels so fragile otherwise. 
As a requirement, it makes sense to me to read "any field used during unmarshalling YAML should be validated", but now we lose that nice definition.

Isn't it good enough to have the `Validate` method on `HealthCheckArgsOrString` ? since that already will call internally `Validate` on any of its exclusive subfields. I'd rather have us keep `typ.Field(i).Tag.Get("yaml") == ""`

iamhopaul123: >  since that already will call internally Validate on any of its exclusive subfields.

Problem is this is not always guaranteed...other people could forget to do so, which is the whole point to have this test

internal/pkg/manifest/validate_integration_test.go
iamhopaul123: OK the previous implementation assumes we don't require advance type to have `Validated()`. Addressed in https://github.com/aws/copilot-cli/pull/2895/commits/a13333bf2d72e99cb014a4eb7d0cd51e5507808e. Good catch!

internal/pkg/manifest/validate_integration_test.go
```diff
@@ -62,19 +73,34 @@ func isValid(v reflect.Value) error {
...
-	for _, k := range basicKinds {
-		if typ.Kind() == k {
-			return nil
+	// Skip if it is a type that doesn't need to implement Validate().
+	if exceptionType(typ) {
+		return nil
+	}
+	// For slice and map, validate individual member.
+	if typ.Kind() == reflect.Array || typ.Kind() == reflect.Slice {
+		for i := 0; i < v.Len(); i++ {
```
Lou1415926: Just a thought 💭 .

Since `isValid` validates whether a type is implementing an interface as expected, should we use `t reflect.Type` as its parameter, instead of `v reflect.Vlaue`?

Then here we can use `elemTyp := typ.Elem()` to retrieve the element type of the array/slice, and then call `isValid(elemTyp)`.

internal/pkg/manifest/validate_integration_test.go
```diff
@@ -62,19 +73,34 @@ func isValid(v reflect.Value) error {
...
+			if err := isValid(v.Index(i)); err != nil {
+				return err
+			}
 		}
+		return nil
 	}
+	if typ.Kind() == reflect.Map {
+		for _, k := range v.MapKeys() {
```
Lou1415926: And also use `typ.Key()` and `typ.Elem()` here.

internal/pkg/manifest/validate_test.go
```diff
@@ -827,28 +827,29 @@ func TestTaskConfig_Validate(t *testing.T) {
...
 
-			if tc.wanted != nil {
-				require.EqualError(t, err, tc.wanted.Error())
-			} else {
-				require.NoError(t, err)
-			}
-		})
-	}
-}
+// func TestPlatformString_Validate(t *testing.T) {
```
Lou1415926: Is this intentionally commented out?

internal/pkg/manifest/validate_integration_test.go
```diff
@@ -17,11 +18,21 @@ import (
...
 	reflect.Complex64, reflect.Complex128,
-	reflect.Array, reflect.String, reflect.Slice, reflect.Map,
+}
+
+func basicTypesString() []string {
+	var types []string
+	for _, k := range basicKinds {
+		types = append(types, k.String())
+	}
+	types = append(types, reflect.TypeOf(map[string]string{}).String())
```
Lou1415926: What about this: 

In `isValid`, we detect the map's key and value types, and recursively call `isValid` if they are of non-basic types. 

Therefore, we don't need special treatment for `map[string]string` - since `string` and `string` are basic, they won't get called on `isValid`.

Same for slice.

iamhopaul123: this is what i was doing! i forgot to remove those

internal/pkg/manifest/validate_test.go
iamhopaul123: 🤦

internal/pkg/manifest/validate_integration_test.go
iamhopaul123: since we are now skipping private fields, we have to use value instead of type

iamhopaul123: but i guess you are right. we only need to check the first element since they are all the same

iamhopaul123: discussed offline now use `structfield.IsExported()` to check so that we can use type as the parameter.

Lou1415926 (approved):
hooray!

PR #2896, opened 2021-10-05
feat: enable colors on Windows

mattn:
On Windows, colors are not enabled. This change enable colors.
image
image

### DIFF
--- a/cmd/copilot/main.go
+++ b/cmd/copilot/main.go
@@ -13,6 +13,7 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/term/color"
 	"github.com/aws/copilot-cli/internal/pkg/term/log"
 	"github.com/aws/copilot-cli/internal/pkg/version"
+	"github.com/mattn/go-colorable"
 	"github.com/spf13/cobra"
 )
 
@@ -25,7 +26,9 @@ func init() {
 	cobra.EnableCommandSorting = false // Maintain the order in which we add commands.
 }
 
-func main() {
+func run() int {
+	defer colorable.EnableColorsStdout(nil)()
+
 	cmd := buildRootCmd()
 	if err := cmd.Execute(); err != nil {
 		var ac actionRecommender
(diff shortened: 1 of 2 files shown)


cmd/copilot/main.go
```diff
@@ -25,16 +26,23 @@ func init() {
 	cobra.EnableCommandSorting = false // Maintain the order in which we add commands.
 }
 
-func main() {
+func run() int {
+	defer colorable.EnableColorsStdout(nil)()
```
efekarakus: I wonder why it wasn't working before 🤔 
We're using the following writers when writing to stdout:
https://github.com/aws/copilot-cli/blob/459f729744c4b3d8e4adad2e832312a2dd1d8e94/internal/pkg/term/log/log.go#L23-L24

which uses your library 🤩 as far as I understand:
https://github.com/fatih/color/blob/a05da93ebe62ca9fc6791d3376ec4dad01196448/color.go#L25-L30

Shouldn't that have been working already or am I misunderstanding how `colorable.NewColorableStdout()` works?

efekarakus (commented):
Hi @mattn ! Thank you so much for the contribution, your library is awesome 😍

I just have a question below please 🙇

efekarakus:
Side note, is it possible to update the commit like this: `feat: enable colors on Windows`? so that it makes our CI checks happy 😆

cmd/copilot/main.go
mattn: Ah, copilot-cli already use go-colorable. Okay, I'll fix in another way soon.

mattn:
I'll squash commits if you want.

efekarakus (approved):

efekarakus:
> I'll squash commits if you want.

Not needed, thank you for the contribution!

Lou1415926 (approved):
Thank you for the contribution and your awesome library 😻

mattn:
Thank you!

Issue #2882, opened 2021-09-29
Failed to delete environment xx from application xxx.


afgallo:
Hi there!

I couldn't find an existing issue with the error below so I thought I'd create a new one.

> ✘ Failed to delete environment xx from application xxx.
✘ execute env delete: delete environment xx stack: wait until stack xx-xxx delete is complete: ResourceNotReady: failed waiting for successful resource state

I get this error when I run the following commands in order:

1. `copilot app init xxx --domain xx.com --resource-tags application=xx`
2. `copilot env init --name xx --app xxx --profile xxxx --region ap-southeast-2 --default-config`
3. `copilot app delete -n xxx`

My current setup is as follows:
1. copilot version v1.10.1
2. App version v.1.0.2
3. I use different AWS profiles for creating environments

Error details from CloudFormation logs:

<img width="848" alt="image" src="https://user-images.githubusercontent.com/9999415/135182187-ed67f13f-021b-464d-be4e-e75cccb78059.png">

I hope I provided enough details for you guys to diagnose the issue.

Cheers 🥂

Lou1415926:
Hi @afgallo !

Looks like this is the same issue as #2870. Basically, when you have created an application via `copilot app init --domain xx.com`, then an environment, and that the environment deployment fails - this problem will occur. Sorry for the inconvenience! We'll fix it as soon as we can.

For now, you can delete the environment stack by following the below steps:
1. Go to your Route53 console
2. Go to your environment hosted zone, which is named `<EnvironmentName>.<AppName>.<DomainName>`. 
3. There should be three records there: a "NS" record, an "SOA" record and a "CNAME" record. Delete the CNAME record. This is the "non-required resource record set" that prevents the hosted zone from being deleted, hence the stack.

After removing this record, you can try to run `copilot app delete` again. 

<img width="1446" alt="Screen Shot 2021-09-29 at 9 09 18 AM" src="https://user-images.githubusercontent.com/79273084/135308573-689aed16-af59-4345-8b47-53ae9cd14bb8.png">

afgallo:
Thanks @Lou1415926 ! The workaround worked!

efekarakus:
Hi all! The fix is now out: https://github.com/aws/copilot-cli/releases/tag/v1.12.0 🥳

PR #2897, opened 2021-10-05
fix(custom-resource): set physical resource ID as soon as possible to prevent failure in delete event

### LINKED ISSUE #2870: Delete validation records upon `copilot env init` failures
### Background
If an application is associated with a domain, (e.g. by running `copilot app init  --name my-app --domain my-domain.com`), when we provision am environment, we request a certificate for the environment domain, (e.g. `test.my-app.my-domain.com`), and upsert the validation record to the hosted zone.

### The Problem
When the certificate failed to validate, for example, because of a timeout, CloudFormation always fail to delete the environment stack.

### Why
When the certificate failed to validate, we do not fill the `PhysicalResourceID` with the certificate ARN. This results in this [conditional](https://github.com/aws/copilot-cli/blob/cbf43493e8e42d3d04c2394349199016229eaa3c/cf-custom-resources/lib/dns-cert-validator.js#L596) evaluating to `false` during a `DELETE` event (triggered because of the CloudFormation rollback), hence skipping the handler.

We should make sure to remove any upserted record during a `DELETE` event.

Lou1415926:
Fix #2870 and #2882

### DIFF
--- a/cf-custom-resources/lib/dns-cert-validator.js
+++ b/cf-custom-resources/lib/dns-cert-validator.js
@@ -1,5 +1,9 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+
+/* jshint node: true */
+/* jshint esversion: 8 */
+
 "use strict";
 
 const aws = require("aws-sdk");
@@ -41,7 +45,7 @@ let report = function (
     const https = require("https");
     const { URL } = require("url");
 
-    var responseBody = JSON.stringify({
+    let responseBody = JSON.stringify({
       Status: responseStatus,
       Reason: reason,
       PhysicalResourceId: physicalResourceId || context.logStreamName,
@@ -94,6 +98,7 @@ let report = function (
  * @param {string} aliases the custom domain aliases
  * @param {string} envHostedZoneId the environment Route53 Hosted Zone ID
  * @param {string} rootDnsRole the IAM role ARN that can manage domainName
+ * @param {region} region the environment region
  * @returns {string} Validated certificate ARN
  */
 const requestCertificate = async function (
@@ -108,12 +113,12 @@ const requestCertificate = async function (
 ) {
   const crypto = require("crypto");
   const [acm, envRoute53, appRoute53] = clients(region, rootDnsRole);
-  // For backward compatiblity, and make sure the SANs we use are unique.
-  const uniqueSansToUse = new Set([certDomain, `*.${certDomain}`]);
+  // For backward compatibility, and make sure the SANs we use are unique.
+  const uniqueSansToUse = new Set([certDomain, `*.${certDomain}`, ]);
   for (const alias of aliases) {
     uniqueSansToUse.add(alias);
   }
-  const sansToUse = [...uniqueSansToUse];
+  const sansToUse = [...uniqueSansToUse, ];
   const reqCertResponse = await acm
     .requestCertificate({
       DomainName: certDomain,
@@ -147,7 +152,7 @@ const requestCertificate = async function (
       })
       .promise();
     options = Certificate.DomainValidationOptions || [];
-    var readyRecordsNum = 0;
+    let readyRecordsNum = 0;
     for (const option of options) {
       if (option.ResourceRecord) {
         readyRecordsNum++;
@@ -296,9 +301,9 @@ const deleteHostedZoneRecords = async function (
   // Make sure DNS validation records are unique. For example: "example.com" and "*.example.com"
   // might have the same DNS validation record.
   const filteredRecordOption = [];
-  var uniqueValidateRecordNames = new Set();
+  let uniqueValidateRecordNames = new Set();
   for (const option of recordOptionsToDelete) {
-    var id = `${option.ResourceRecord.Name} ${option.ResourceRecord.Value}`;
+    let id = `${option.ResourceRecord.Name} ${option.ResourceRecord.Value}`;
     if (uniqueValidateRecordNames.has(id)) {
       continue;
     }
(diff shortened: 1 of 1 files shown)


cf-custom-resources/lib/dns-cert-validator.js
```diff
@@ -78,43 +82,49 @@ let report = function (
   });
 };
 
+const aliasesChanged = async function (
```
efekarakus: nit: I think this fn doesn't need to be `async`

cf-custom-resources/lib/dns-cert-validator.js
```diff
@@ -78,43 +82,49 @@ let report = function (
...
+  let prevAliases = await getAllAliases(
+      oldResourceProperties.Aliases
+  );
+  let aliasesToDelete = [...prevAliases,].filter(function (itm) {
+    return !aliases.has(itm);
+  });
+  let aliasesToAdd = [...aliases,].filter(function (itm) {
+    return !prevAliases.has(itm);
+  });
+  return !aliasesToAdd.length + aliasesToDelete.length === 0;
```
efekarakus: 😮  is this doing the right thing vs:
```suggestion
  return !((aliasesToAdd.length + aliasesToDelete.length) === 0);
```

cf-custom-resources/lib/dns-cert-validator.js
```diff
@@ -164,30 +188,52 @@ const requestCertificate = async function (
...
   );
 
   await acm
     .waitFor("certificateValidated", {
       // Wait up to 9 minutes and 30 seconds
       $waiter: {
-        delay: 30,
-        maxAttempts: 19,
+        delay: 1,
+        maxAttempts: 1,
```
efekarakus: Do we want these changes?

cf-custom-resources/lib/dns-cert-validator.js
```diff
@@ -545,50 +593,51 @@ exports.certificateRequestHandler = async function (event, context) {
...
-          props.EnvHostedZoneId,
-          props.RootDNSRole,
-          props.Region
+          sansToUse,
+          acm
         );
-        responseData.Arn = physicalResourceId = certificateArn;
+        responseData.Arn = physicalResourceId = response.CertificateArn;  // Set physicalResourceId as soon as we can.
+        options = await waitForValidationOptionsToBeReady(response.CertificateArn, sansToUse, acm);
+        await validateCertificate(options, envRoute53, appRoute53, props.EnvHostedZoneId, response.CertificateArn, acm);
```
efekarakus, line 621: oh i see, so now we separated the certificate request logic compared to before where we would merge all three operations into one place

cf-custom-resources/lib/dns-cert-validator.js
```diff
@@ -545,50 +593,51 @@ exports.certificateRequestHandler = async function (event, context) {
...
-            return !aliases.has(itm);
-          });
-          var aliasesToAdd = [...aliases].filter(function (itm) {
-            return !prevAliases.has(itm);
-          });
-          if (aliasesToAdd.length + aliasesToDelete.length === 0) {
-            break;
-          }
+        let shouldUpdate = await aliasesChanged(event.OldResourceProperties, aliases);
+        if (!shouldUpdate) {
```
efekarakus: nit:
```suggestion
        if (!aliasesChanged(event.OldResourceProperties, aliases)) {
```

cf-custom-resources/lib/dns-cert-validator.js
Lou1415926: oh no.. good catch and sorry :woman-facepalming:

cf-custom-resources/lib/dns-cert-validator.js
iamhopaul123: Maybe just
```
return aliasesToAdd.length + aliasesToDelete.length !== 0;
```

cf-custom-resources/lib/dns-cert-validator.js
```diff
@@ -78,43 +82,49 @@ let report = function (
...
+  let prevAliases = getAllAliases(
+      oldResourceProperties.Aliases
+  );
+  let aliasesToDelete = [...prevAliases,].filter(function (itm) {
+    return !aliases.has(itm);
+  });
+  let aliasesToAdd = [...aliases,].filter(function (itm) {
+    return !prevAliases.has(itm);
+  });
+  return aliasesToAdd.length + aliasesToDelete.length !== 0;
```
efekarakus: im dumb why not just 
```suggestion
  return aliasesToAdd.length + aliasesToDelete.length;
```

efekarakus (approved):

cf-custom-resources/lib/dns-cert-validator.js
```diff
@@ -478,13 +525,15 @@ const getAllAliases = function (aliases) {
   } catch (error) {
     throw new Error(`Cannot parse ${aliases} into JSON format.`);
   }
-  var aliasList = [];
-  for (var m in obj) {
-    aliasList.push(...obj[m]);
+  let aliasList = [];
+
+  for (const [_, aliases] of Object.entries(obj)) {
```
iamhopaul123, line 530: just curious why are we making this change 🤔

cf-custom-resources/lib/dns-cert-validator.js
Lou1415926: :homer-disappearing:

iamhopaul123 (approved):

cf-custom-resources/lib/dns-cert-validator.js
Lou1415926: Not much tbh. Just b/c it has slightly more readability +  JSLint always complains with the message "The body of a for in should be wrapped in an if statement to filter unwanted properties from the prototype."

PR #2901, opened 2021-10-06
feat: inject copilot environment variables to sidecars

### LINKED ISSUE #2862: Add copilot environment variables to sidecar environments
There are a few copilot environment variables that are added automatically to the environment of the main task container, such as:

- COPILOT_APPLICATION_NAME
- COPILOT_ENVIRONMENT_NAME
- COPILOT_SERVICE_DISCOVERY_ENDPOINT
- COPILOT_SERVICE_NAME

For my use case it is helpful to have this information also exposed in the environment of the sidecar containers.

As a workaround, it is possible to add the env variables via the manifest to the sidecar, as described at https://github.com/aws/copilot-cli/issues/1119

Lou1415926:
This PR injects copilot environment variables to sidecars, similar to what's been done for the main container.

Resolves #2862 


internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
```diff
@@ -36,9 +36,16 @@
     StartPeriod: {{$sidecar.HealthCheck.StartPeriod}}
     Timeout: {{$sidecar.HealthCheck.Timeout}}
 {{- end}}
-{{- if $sidecar.Variables}}
   Environment:
```
efekarakus, line 39: I wonder if we can just include `envvars.yml` instead here so that we don't have to update two different locations in case we add more env vars

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
```diff
@@ -1,6 +1,11 @@
 - Name: !Ref WorkloadName
   Image: !Ref ContainerImage
 {{include "secrets" . | indent 2}}
+  # We pipe certain environment variables directly into the task definition.
+  # This lets customers have access to, for example, their LB endpoint - which they'd
+  # have no way of otherwise determining.
+  Environment:
+{{include "envvars-copilot" $ | indent 2}}
```
iamhopaul123: I think we can keep using `.` for consistency as they are the same in this context.

internal/pkg/template/workload.go
```diff
@@ -90,15 +91,17 @@ type WorkloadNestedStackOpts struct {
...
+	Image      *string
+	Essential  *bool
+	Port       *string
+	Protocol   *string
+	CredsParam *string
+	Variables  map[string]string
+	Secrets    map[string]string
+	Storage    struct {
+		MountPoints []*MountPoint
+	}
```
iamhopaul123: I'm not a big fan of this...Can we make it something more specific like
```golang
type SidecarStorageOpts struct {
...
}
```

internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
```diff
@@ -36,17 +36,9 @@
...
-  - Name: {{$name}}
-    Value: {{$value | printf "%q"}}
-  {{- end}}
-  {{- if $sidecar.MountPoints}}
-  - Name: COPILOT_MOUNT_POINTS
-    Value: '{{jsonMountPoints $sidecar.MountPoints}}'
-  {{- end}}
-{{- end}}
+{{include "envvars-copilot" $ | indent 2}}
+{{include "envvars" . | indent 2}}
```
iamhopaul123: We probably need better names for both `envvars-copilot` and `envvars` 🤔 @efekarakus any thought on this?

iamhopaul123 (commented):
Looks good! Only a few (real) questions.

internal/pkg/template/workload.go
efekarakus: yeah I also like that!

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
```diff
@@ -1,6 +1,11 @@
 - Name: !Ref WorkloadName
   Image: !Ref ContainerImage
 {{include "secrets" . | indent 2}}
+  # We pipe certain environment variables directly into the task definition.
```
efekarakus: nit: let's just remove these comments. I think it's pretty well established at this point

internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
efekarakus: OK thanks to @Lou1415926 for explaining me the difference :D It took me a while to parse the PR

1. `"envvars-copilot"` takes as input `WorkloadOpts` (common across containers)
2. `"envvars"` takes as input container specific data

What do you think of the following names instead?
```
# "$" denotes the parent WorkloadOpts, whereas "." is the individual container.
{{include "envvars-common" $ | indent 2}}
{{include "envvars-container" . | indent 2}}
```

efekarakus (commented):
OK it looks good! It took me quite a while to grasp if the changes were safe but it looks great.

Let me know if the naming suggestions make sense.

internal/pkg/deploy/cloudformation/stack/stack_local_integration_test.go
```diff
@@ -59,6 +60,7 @@ func Test_Stack_Local_Integration(t *testing.T) {
 		wantedCFNBytes, err := ioutil.ReadFile(path)
 		require.NoError(t, err)
 
+		fmt.Println(tpl)
```
efekarakus: 😂 deleteee

efekarakus (approved):

iamhopaul123 (approved):

PR #2903, opened 2021-10-08
feat: allow worker service to publish

### LINKED ISSUE #2900: [Feature Request] Allow worker services to publish
Currently, a worker service can only subscribe to queues, not publish. We should support publishing for worker services as this is also considered a legit use case.

iamhopaul123:
Resolves #2900. Also fixes a bug that worker service deployment fails if the name of the topic it subscribes to contains "-" (e.g., "foo-bar").

Note that it doesn't allow a worker service to subscribe to itself.


e2e/worker/worker_test.go
```diff
@@ -156,5 +223,37 @@ publish:
...
+
+				body, err := ioutil.ReadAll(resp.Body)
+				if err != nil {
+					return err
+				}
+				fmt.Printf("response: %s\n", string(body))
+				// The counter service add to the counter when the worker service processes a message
+				if string(body) != "5" {
+					return fmt.Errorf("the message content is '%s', but expected '%s'", string(body), "5")
+				}
```
efekarakus: Instead of checking if it's 5 can we check at least 5?

```
count, _ := strconv.Atoi(s)
if count < 5 {
   return fmt.Errorf
}
```

internal/pkg/template/templates/workloads/partials/cf/instancerole.yml
```diff
@@ -43,7 +43,7 @@ InstanceRole:
               Action: 'sns:Publish'
               Resource:
               {{- range $topic := .Publish.Topics }}
-              - !Ref {{$topic.Name}}SNSTopic
+              - !Ref {{logicalIDSafe $topic.Name}}SNSTopic
```
efekarakus, line 46: classic 😂

efekarakus (commented):
Looks good just a small comment

internal/pkg/deploy/cloudformation/stack/testdata/workloads/worker-manifest.yml
```diff
@@ -23,6 +23,10 @@ count:
     msg_processing_time: 1s
 exec: true     # Enable running commands in your container.
 
+publish:
+  topics:
+    - name: givesOtherdogs
```
Lou1415926, line 28: doggo land giving out free doggos 😮

internal/pkg/template/templates/workloads/partials/cf/publish.yml
```diff
@@ -5,7 +5,7 @@
       'aws:copilot:description': 'A SNS topic to broadcast {{$topic.Name}} events'
     Type: AWS::SNS::Topic
     Properties:
-      TopicName: !Sub '${AWS::StackName}-{{logicalIDSafe $topic.Name}}'
+      TopicName: !Sub '${AWS::StackName}-{{$topic.Name}}'
```
Lou1415926, line 8: Would this be breaking for people that have deployed a worker service?

iamhopaul123: This is a bug...since it causes a fact that a user who deployed a workload with SNS topic to contain "-" cannot use the topic at all. When they are trying to deploy a worker service subscribing to the topic, Copilot would not be able to find the topic and shows some errors.

Lou1415926: gotcha!

Lou1415926 (approved):

efekarakus (approved):

PR #2904, opened 2021-10-08
chore: fix allow only CFN-ready values for `task run` flags; change logging logic

huanjani:


internal/pkg/cli/svc_init.go
```diff
@@ -458,12 +458,12 @@ func (o *initSvcOpts) legitimizePlatform() error {
...
 	if platform != detectedPlatform {
-		log.Warningf("Your architecture type is currently unsupported. Setting platform %s instead.\n", dockerengine.PlatformString(dockerengine.OSLinux, dockerengine.ArchAMD64))
-		if o.wkldType != manifest.RequestDrivenWebServiceType {
-			log.Warning("See 'platform' field in your manifest.\n")
-		}
-		o.platform = &platform
+		log.Warningf("Your architecture type is currently unsupported. Setting platform %s instead.\n", platform)
+	}
+	if platform != manifest.DefaultPlatform && o.wkldType != manifest.RequestDrivenWebServiceType {
+		log.Warning("See 'platform' field in your manifest.\n")
```
huanjani: I'm kind of thinking we don't need to log this at all. It doesn't seem more urgent to configure than other fields in the manifest. PH was right when he said that the current wording makes it seem like something is wrong; even if we soften it, it does seem a bit out of place. If someone is on Windows, this should just work. And if they're interested in changing from default CORE to FULL, they would know how to do it from the docs. What does everyone else think?

internal/pkg/cli/job_init.go
```diff
@@ -344,9 +344,12 @@ func (o *initJobOpts) legitimizePlatform() error {
 		return nil
 	}
 	if platform != detectedPlatform {
-		log.Warningf("Your platform %s is currently unsupported. Setting platform to %s in your manifest.\n", detectedPlatform, platform)
-		o.platform = &platform
+		log.Warningf("Your architecture type is currently unsupported. Setting platform %s instead.\n", platform)
 	}
+	if platform != manifest.DefaultPlatform {
```
iamhopaul123: Does this only happen when OS type is linux and arch type is non amd64?

internal/pkg/cli/job_init.go
```diff
@@ -344,9 +344,12 @@ func (o *initJobOpts) legitimizePlatform() error {
 		return nil
 	}
 	if platform != detectedPlatform {
```
iamhopaul123: Does this only happen when arch type is non amd64?

huanjani: correct-- if detectedPlatform is `linux/arm` we would redirect platform to become `linux/amd64`, so then they would not be equal.

internal/pkg/cli/job_init.go
huanjani: This happens anytime the platform we're building _isn't_ `linux/amd64`, so `windows/amd64` would not log the "unsupported" warning, but _would_ log this one. `windows/arm` would log both because it's both being redirected and the redirected platform (`windows/amd64`) is not the default. We only surface the `platform` field in the manifest when it's not `linux/amd64`.

iamhopaul123: Gotcha but don't we already return "" for `platform` if it is default platform? https://github.com/aws/copilot-cli/blob/fbd8c3fa98890621c84e58210c3ddf3430fc539f/internal/pkg/manifest/workload.go#L744-L745

which means this should always be true?

internal/pkg/cli/job_init.go
iamhopaul123: Maybe rename `platform` to be `redirectedPlatform` for readability?

internal/pkg/cli/job_init.go
```diff
@@ -336,17 +336,18 @@ func (o *initJobOpts) legitimizePlatform() error {
...
+	if redirectedPlatform == "" {
 		return nil
 	}
-	if platform != detectedPlatform {
-		log.Warningf("Your platform %s is currently unsupported. Setting platform to %s in your manifest.\n", detectedPlatform, platform)
-		o.platform = &platform
+	if redirectedPlatform != detectedPlatform {
+		log.Warningf("Your architecture type is currently unsupported. Setting platform %s instead.\n", redirectedPlatform)
 	}
+	log.Warningf("See 'platform' field in your manifest.\n")
```
efekarakus: nit: what do you think of?
```suggestion
		log.Warningf("Your architecture type %s is currently unsupported.\n", color.HighlightCode(detectedPlatform))
		log.Warningf("Setting %s instead in your manifest.\n", colorHighlightCode(fmt.Sprintf("platform: '%s'", redirectedPlatform))
	}
	
```

internal/pkg/cli/svc_init.go
efekarakus: Nice, I like your suggestion! 

Let's only log stuff if we "override" things i.e. `redirectedPlatform != detectedPlatform`.

internal/pkg/task/task.go
```diff
@@ -78,3 +86,13 @@ func convertECSTasks(ecsTasks []*ecs.Task) []*Task {
...
 	return tasks
 }
+
+// ValidOSs returns all valid CFN-friendly operating systems.
+func ValidOSs() []string {
+	return []string{osWindowsServer2019Core, osWindowsServer2019Full, osLinux}
+}
+
+// ValidArchs returns all valid CFN-friendly architectures.
+func ValidArchs() []string {
```
efekarakus: is there a reason why you prefered functions instead of exporting variables? I think this is neat it allows the values to be immutable 👍

internal/pkg/cli/task_run.go
```diff
@@ -306,7 +306,10 @@ func (o *runTaskOpts) Validate() error {
 	if noOS, noArch := o.os == "", o.arch == ""; noOS != noArch {
 		return fmt.Errorf("must specify either both `--%s` and `--%s` or neither", osFlag, archFlag)
 	}
-	if err := o.validatePlatform(); err != nil {
+	if err := o.validateOS(); err != nil {
+		return err
+	}
+	if err := o.validateArch(); err != nil {
```
efekarakus: This is fine i think for now, it'd be cool if we can write it in a way such that in the future specifying linux and arm would pass but not windows and arm 🤔

efekarakus (commented):
Looks great! just small suggestions

internal/pkg/cli/job_init.go
huanjani: I like it! Combining into one log line (but with \n) so "NOTE:" doesn't appear twice, and also using just `detectedArch`. Also not using the single quotes for the platform bc we don't in the manifest.

internal/pkg/cli/task_run.go
huanjani: Oh yeah. That's a quick fix. Will do it here.

internal/pkg/cli/svc_init.go
```diff
@@ -458,10 +458,7 @@ func (o *initSvcOpts) legitimizePlatform() error {
 		return nil
 	}
 	if redirectedPlatform != detectedPlatform {
-		log.Warningf("Your architecture type is currently unsupported. Setting platform %s instead.\n", redirectedPlatform)
-	}
-	if o.wkldType != manifest.RequestDrivenWebServiceType {
-		log.Warning("See 'platform' field in your manifest.\n")
+		log.Warningf("Your architecture type %s is currently unsupported. Setting platform %s instead.\n", color.HighlightCode(detectedArch), redirectedPlatform)
```
efekarakus, line 461: should this be consistent with job init?

efekarakus (approved):
Looks good! Just a tiny log request

iamhopaul123 (approved):

internal/pkg/cli/svc_init.go
huanjani: No, because `job init` doesn't have to weed out RDWS.

PR #2911, opened 2021-10-12
chore: friendly message when alias is used while domain is not associated with the app

Lou1415926:
We should display a friendly message when we detect that `alias` is specified in manifest while the app is not associated with a domain.

Previously, when this happens, an RDWS deployment fails when CloudFormation tries to create the custom resource. On the other hand, for LBWS, we swallow the alias input when domain is not associated. 

Therefore in this PR, we error out early for RDWS while displaying the message. For LBWS, we only display messages while not failing.

### DIFF
--- a/internal/pkg/cli/svc_deploy.go
+++ b/internal/pkg/cli/svc_deploy.go
@@ -51,7 +51,7 @@ const (
 	fmtForceUpdateSvcComplete = "Forced an update for service %s from environment %s.\n"
 )
 
-var aliasUsedWithoutDomainFriendlyTest = fmt.Sprintf("To use %s, your application must be associated with a domain: %s.\n",
+var aliasUsedWithoutDomainFriendlyText = fmt.Sprintf("To use %s, your application must be associated with a domain: %s.\n",
 	color.HighlightCode("http.alias"),
 	color.HighlightCode("copilot app init --domain example.com"))
 
@@ -539,7 +539,7 @@ func (o *deploySvcOpts) stackConfiguration(addonsURL string) (cloudformation.Sta
 	switch t := mft.(type) {
 	case *manifest.LoadBalancedWebService:
 		if o.targetApp.Domain == "" && !t.Alias.IsEmpty() {
-			log.Errorf(aliasUsedWithoutDomainFriendlyTest)
+			log.Errorf(aliasUsedWithoutDomainFriendlyText)
 			return nil, errors.New("alias specified when application is not associated with a domain")
 		}
 		if o.targetApp.RequiresDNSDelegation() {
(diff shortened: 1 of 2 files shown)


internal/pkg/cli/svc_deploy.go
```diff
@@ -51,6 +51,10 @@ const (
 	fmtForceUpdateSvcComplete = "Forced an update for service %s from environment %s.\n"
 )
 
+var aliasUsedWithoutDomainFriendlyTest = fmt.Sprintf("To use %s, first run %s to associate your application with your own domain.\n",
+	color.HighlightCode("alias"),
```
```suggestion
var aliasUsedWithoutDomainFriendlyTest = fmt.Sprintf("To use %s, your application must be associated with a domain: %s.\n",
	color.HighlightCode("http.alias"),
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -51,6 +51,10 @@ const (
...
+	color.HighlightCode("copilot app init --domain <your.domain>"))
```
```suggestion
	color.HighlightCode("copilot app init --domain example.com"))
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -534,6 +538,9 @@ func (o *deploySvcOpts) stackConfiguration(addonsURL string) (cloudformation.Sta
 	var conf cloudformation.StackConfiguration
 	switch t := mft.(type) {
 	case *manifest.LoadBalancedWebService:
+		if o.targetApp.Domain == "" && !t.Alias.IsEmpty() {
+			log.Errorf(aliasUsedWithoutDomainFriendlyTest)
```
efekarakus: why shouldn't we error here as well? seems like that would be a good thing to do

Lou1415926: This would fail operations that have been succeeding before. Previously people can still deploy the service, it's just that the aliases won't take effect.

efekarakus: I think that's okay to fail! I wouldn't consider it a breaking change, instead it's better validation 🤔 
Let's make both services consistent!

efekarakus (approved):

internal/pkg/cli/svc_deploy.go
```diff
@@ -51,6 +51,10 @@ const (
 	fmtForceUpdateSvcComplete = "Forced an update for service %s from environment %s.\n"
 )
 
+var aliasUsedWithoutDomainFriendlyTest = fmt.Sprintf("To use %s, your application must be associated with a domain: %s.\n",
```
huanjani: nit: did you mean
```suggestion
var aliasUsedWithoutDomainFriendlyText = fmt.Sprintf("To use %s, your application must be associated with a domain: %s.\n",
```
??

internal/pkg/cli/svc_deploy.go
huanjani: Do we need a unit test now that it's erroring out? Womp womp.

internal/pkg/cli/svc_deploy.go
Lou1415926: 😳

huanjani (approved):
:shipit:

Issue #2912, opened 2021-10-12
Substitute environment variables in manifests


iamhopaul123:
## Background

Currently Copilot manifest does not support any form of pseudo-variables. On one hand, users cannot refer to the shell environment variables. On the other hand, they have to specify very similar values for different environments repetitively using environment override. For example, in our [secret init doc](https://aws.github.io/copilot-cli/docs/commands/secret-init/#whats-next), users during next step after the secret creation, they have to specify different DB_PASSWORD for test and dev environment.

Given the above-mentioned reasons, it is worth for us to introduce appropriate pseudo-variables to our manifest to make it more flexible.

## Related Issues
- #1599 
- #1992
- #2483
- #2445
- #2590
- #2769

### Shell environment variables
```
name: my-service
type: Load Balanced Web Service
image:
  location: id.dkr.ecr.zone.amazonaws.com/project-name:${TAG}
  port: 3333
```

### Pre-defined variables
```
secrets:
   DB_PASSWORD: /copilot/my-app/${COPILOT_ENVIRONMENT_NAME}/secrets/db_password
environments:
    prod:
      secrets: 
          <specific environment variable for prod>
    dev:
      secrets:
          <specific environment variable for dev>
```

Available pre-defined variables:

- COPILOT_APPLICATION_NAME
- COPILOT_ENVIRONMENT_NAME

It’s worth noting that we won’t allow users to override any of our predefined variables. Once we detect they have shell variables with the same name, we’ll log warning messages saying we won’t override these variables for them. Otherwise, it is possible they have ${COPILOT_ENVIRONMENT_NAME} defined in the shell session where they run Copilot without realizing its existence.

Also Copilot should error out if users are trying to refer to an undefined env var in their manifest.

afgallo:
Oh, it's great that I've found this issue. I came across this limitation when I was trying to let copilot build my docker image but needed the ability to pass in an env variable to the build process. I guess the workaround now is to build the image manually and change my manifest to point to the image and not the dockerfile, @iamhopaul123 and @efekarakus ?

iamhopaul123:
> I guess the workaround now is to build the image manually and change my manifest to point to the image and not the dockerfile

Yes that is correct!

afgallo:
[...]

PR #2913, opened 2021-10-12
chore(manifest): add env var interpolator

iamhopaul123:
Part of #2912

### DIFF
--- a/internal/pkg/manifest/interpolate.go
+++ b/internal/pkg/manifest/interpolate.go
@@ -1,0 +1,67 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package manifest
+
+import (
+	"fmt"
+	"os"
+	"regexp"
+)
+
+const (
+	predefinedEnvVarKeyForAppName = "COPILOT_APPLICATION_NAME"
+	predefinedEnvVarKeyForEnvName = "COPILOT_ENVIRONMENT_NAME"
+)
+
+var (
+	// Modified from docker/compose
+	interpolatorRegexp = regexp.MustCompile(`\${([_a-zA-Z][_a-zA-Z0-9]*)}`)
+)
+
+type interpolator struct {
+	predefinedEnvVars map[string]string
+}
+
+type predefinedEnvVar struct {
+	appName string
+	envName string
+}
+
+func newInterpolator(envVars predefinedEnvVar) *interpolator {
+	return &interpolator{
+		predefinedEnvVars: map[string]string{
+			predefinedEnvVarKeyForAppName: envVars.appName,
+			predefinedEnvVarKeyForEnvName: envVars.envName,
+		},
+	}
+}
+
+func (i *interpolator) substitute(s string) (string, error) {
+	matches := interpolatorRegexp.FindAllStringSubmatch(s, -1)
+	if len(matches) == 0 {
+		return "", nil
+	}
+	replaced := s
+	for _, match := range matches {
+		// https://pkg.go.dev/regexp#Regexp.FindAllStringSubmatch
+		key := match[1]
+		subTextRegexp := regexp.MustCompile(fmt.Sprintf(`\${%s}`, key))
+		predefinedVal, isPredefined := i.predefinedEnvVars[key]
+		if isPredefined {
+			replaced = subTextRegexp.ReplaceAllString(replaced, predefinedVal)
+		}
+		osVal := os.Getenv(key)
+		if osVal == "" {
+			if isPredefined {
+				continue
+			}
+			return "", fmt.Errorf(`environment variable "%s" is not defined`, key)
+		}
+		if isPredefined {
+			return "", fmt.Errorf(`predefined environment variable "%s" cannot be overridden with "%s"`, key, osVal)
+		}
+		replaced = subTextRegexp.ReplaceAllString(replaced, osVal)
+	}
+	return replaced, nil
+}
--- a/internal/pkg/manifest/interpolate.go
+++ b/internal/pkg/manifest/interpolate.go
@@ -7,15 +7,16 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strings"
 )
 
 const (
-	predefinedEnvVarKeyForAppName = "COPILOT_APPLICATION_NAME"
-	predefinedEnvVarKeyForEnvName = "COPILOT_ENVIRONMENT_NAME"
+	reservedEnvVarKeyForAppName = "COPILOT_APPLICATION_NAME"
+	reservedEnvVarKeyForEnvName = "COPILOT_ENVIRONMENT_NAME"
 )
 
 var (
-	// Modified from docker/compose.
+	// Taken from docker/compose.
 	// Environment variable names consist solely of uppercase letters, digits, and underscore,
 	// and do not begin with a digit. （https://pubs.opengroup.org/onlinepubs/007904875/basedefs/xbd_chap08.html）
 	interpolatorEnvVarRegExp = regexp.MustCompile(`\${([_a-zA-Z][_a-zA-Z0-9]*)}`)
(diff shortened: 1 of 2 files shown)


internal/pkg/manifest/interpolate.go
```diff
@@ -0,0 +1,67 @@
...
+	"regexp"
+)
+
+const (
+	predefinedEnvVarKeyForAppName = "COPILOT_APPLICATION_NAME"
+	predefinedEnvVarKeyForEnvName = "COPILOT_ENVIRONMENT_NAME"
+)
+
+var (
+	// Modified from docker/compose
```
Lou1415926:
```suggestion
	// Modified from docker/compose.
	// Environment variable names consist solely of uppercase letters, digits, and underscore, and do not begin with a digit. （https://pubs.opengroup.org/onlinepubs/007904875/basedefs/xbd_chap08.html）
```

internal/pkg/manifest/interpolate.go
```diff
@@ -0,0 +1,67 @@
...
+	interpolatorRegexp = regexp.MustCompile(`\${([_a-zA-Z][_a-zA-Z0-9]*)}`)
```
Lou1415926: What do you think of something like
```suggestion
	interpolatorEnvVarRegExp = regexp.MustCompile(`\${([_a-zA-Z][_a-zA-Z0-9]*)}`)
```

internal/pkg/manifest/interpolate.go
```diff
@@ -0,0 +1,67 @@
...
+		return "", nil
+	}
+	replaced := s
+	for _, match := range matches {
+		// https://pkg.go.dev/regexp#Regexp.FindAllStringSubmatch
+		key := match[1]
+		subTextRegexp := regexp.MustCompile(fmt.Sprintf(`\${%s}`, key))
+		predefinedVal, isPredefined := i.predefinedEnvVars[key]
+		if isPredefined {
+			replaced = subTextRegexp.ReplaceAllString(replaced, predefinedVal)
```
Lou1415926: This is very unlikely to happen but we probably need to make sure `predefinedVal` do not have `$` in it. Only to call this out, no need to make code changes!

internal/pkg/manifest/interpolate.go
```diff
@@ -0,0 +1,67 @@
...
+		osVal := os.Getenv(key)
+		if osVal == "" {
+			if isPredefined {
+				continue
+			}
+			return "", fmt.Errorf(`environment variable "%s" is not defined`, key)
+		}
+		if isPredefined {
+			return "", fmt.Errorf(`predefined environment variable "%s" cannot be overridden with "%s"`, key, osVal)
+		}
```
Lou1415926: What do you think if we flip the logic a little:
```suggestion
		if osVal != "" && isPredefined {
				return "", fmt.Errorf(`predefined environment variable "%s" cannot be overridden with "%s"`, key, osVal)
		}
		if osVal == "" {
			return "", fmt.Errorf(`environment variable "%s" is not defined`, key)
		}
```

internal/pkg/manifest/interpolate.go
```diff
@@ -0,0 +1,67 @@
...
+func (i *interpolator) substitute(s string) (string, error) {
+	matches := interpolatorRegexp.FindAllStringSubmatch(s, -1)
+	if len(matches) == 0 {
+		return "", nil
+	}
+	replaced := s
+	for _, match := range matches {
+		// https://pkg.go.dev/regexp#Regexp.FindAllStringSubmatch
+		key := match[1]
+		subTextRegexp := regexp.MustCompile(fmt.Sprintf(`\${%s}`, key))
```
Lou1415926: What do you think of renaming it as:
```suggestion
		currSegmentRegex := regexp.MustCompile(fmt.Sprintf(`\${%s}`, key))
```

internal/pkg/manifest/interpolate.go
efekarakus: 👍 on Wanxian's suggestions, but also what about it was "modified"? or is this more "taken"

internal/pkg/manifest/interpolate.go
```diff
@@ -0,0 +1,67 @@
...
+type interpolator struct {
+	predefinedEnvVars map[string]string
+}
+
+type predefinedEnvVar struct {
+	appName string
+	envName string
+}
+
+func newInterpolator(envVars predefinedEnvVar) *interpolator {
```
efekarakus: Can we delete the `predefinedEnvVar` type?
```suggestion
func newInterpolator(appName, envName string) *interpolator {
```

internal/pkg/manifest/interpolate.go
efekarakus: Do we need this to be a regexp? 🥺 can't we do something along the lines:
```
strings.ReplaceAll(s, fmt.Sprintf("${%s}", key), os.Getenv(key)) 
```

internal/pkg/manifest/interpolate.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"fmt"
+	"os"
+	"regexp"
+)
+
+const (
+	predefinedEnvVarKeyForAppName = "COPILOT_APPLICATION_NAME"
+	predefinedEnvVarKeyForEnvName = "COPILOT_ENVIRONMENT_NAME"
```
efekarakus: nit:
```suggestion
	reservedEnvVarKeyForAppName = "COPILOT_APPLICATION_NAME"
	reservedEnvVarKeyForEnvName = "COPILOT_ENVIRONMENT_NAME"
```

internal/pkg/manifest/interpolate.go
```diff
@@ -0,0 +1,67 @@
...
+		predefinedVal, isPredefined := i.predefinedEnvVars[key]
+		if isPredefined {
+			replaced = subTextRegexp.ReplaceAllString(replaced, predefinedVal)
+		}
+		osVal := os.Getenv(key)
+		if osVal == "" {
+			if isPredefined {
+				continue
+			}
+			return "", fmt.Errorf(`environment variable "%s" is not defined`, key)
```
efekarakus: I don't think that `""` means it's actually undefined, we should be using [`os.LookupEnv`](https://pkg.go.dev/os#LookupEnv) instead 

i.e. we should allow "" if the bool is true.

internal/pkg/manifest/interpolate.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		if isPredefined {
+			return "", fmt.Errorf(`predefined environment variable "%s" cannot be overridden with "%s"`, key, osVal)
```
efekarakus: Let's not throw an error if the user's environment variable value is equal to our value.
Like if they set the appname to be the same as our app name, then np.

internal/pkg/manifest/interpolate.go
```diff
@@ -38,30 +40,29 @@ func newInterpolator(envVars predefinedEnvVar) *interpolator {
...
-			replaced = subTextRegexp.ReplaceAllString(replaced, predefinedVal)
-		}
 		osVal := os.Getenv(key)
-		if osVal == "" {
-			if isPredefined {
-				continue
-			}
-			return "", fmt.Errorf(`environment variable "%s" is not defined`, key)
+		if isPredefined && osVal != "" {
+			return "", fmt.Errorf(`predefined environment variable "%s" cannot be overridden with "%s"`, key, osVal)
```
Lou1415926:
```suggestion
			return "", fmt.Errorf(`predefined environment variable "%s" cannot be overridden by OS environment variable with the same name`)
```
We probably shouldn't include their os val in error message🤔 just to be safe.

I am also wondering if we need to error out in this case. Maybe we can just display a message telling them, well we detected that you have an env var with name colliding with our predefined var. We are using our predefined value instead of yours - so that people do not have to unset their OS env var in order to get this to work.

Lou1415926 (commented):
Small feedback that I overlooked last time!

internal/pkg/manifest/interpolate.go
efekarakus: Can we instead move the logic into a `validate` method?

```go
	for _, envVar := range envVarMatches {
		key := envVar[1] // https://pkg.go.dev/regexp#Regexp.FindAllStringSubmatch
		if err := i.validate(key); err != nil {
			return err
		}
		replaced = strings.ReplaceAll(replaced, fmt.Sprintf("${%s}", key), os.Getenv(key)) 
	}
```

internal/pkg/manifest/interpolate.go
iamhopaul123: make sense!

internal/pkg/manifest/interpolate.go
iamhopaul123: ok update see if you like this version

efekarakus (approved):
🎸 looks great!

Lou1415926 (approved):

PR #2914, opened 2021-10-13
chore: enable and add docs for PlatformArgs

huanjani:
Previously, we would lowercase platform fields to validate them, but not to transform them; consequently, unless values were already lowercased, they would not be recognized and therefore not deployed as non-default. This was not detected as a problem with PlatformStrings because we autogenerated those as lowercase, but this issue became apparent very quickly with PlatformArgs. 

This PR also adds an example of PlatformArgs in the docs. 


efekarakus (approved):
:shipit:

efekarakus:
Why are there 68 commits 😂 ?

huanjani:
> Why are there 68 commits 😂 ?

🙈 🤫

iamhopaul123 (commented):
Shall we update the validate function as well?

iamhopaul123 (approved):

PR #2916, opened 2021-10-13
chore: partial CFN template to support NLB

Lou1415926:


internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,111 @@
...
+NLBListener:
+  Type: AWS::ElasticLoadBalancingV2::Listener
+  Properties:
+    DefaultActions:
+      - TargetGroupArn: !Ref NLBTargetGroup
+        Type: forward
+    LoadBalancerArn: !Ref PublicNetworkLoadBalancer
+    Port: {{ .ListenerPort }}
+    Protocol: {{ .Protocol }}
+{{- if eq .Protocol "TLS }}
```
iamhopaul123:
```suggestion
{{- if eq .Protocol "TLS" }}
```

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,111 @@
...
+{{- if .AlpnPolicy}}
```
iamhopaul123: Maybe for simplicity we don't need to expose both `AlpnPolicy` and `SslPolicy` for now and add them ad hoc?

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,111 @@
...
+    Protocol: {{ .Protocol }}
+    TargetGroupAttributes:
+      - Key: deregistration_delay.timeout_seconds
+        Value: {{.DeregistrationDelay}}  # ECS Default is 300; Copilot default is 60.
+      - Key: stickiness.enabled
+        Value: !Ref Stickiness
+      - Key: deregistration_delay.connection_termination.enabled
+        Value: false # NOTE: Default is false  TODO: remove this comment and investigate if we should surface this or not.
+      - Key: preserve_client_ip.enabled
+        Value: false # NOTE: Default is false for TCP/TLS. Must be true for UDP / UCP_UDP # TODO: remove this comment & exclude this attribute from template.
```
iamhopaul123: What if the protocol is UDP? Will the value `false` be an issue?

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,111 @@
+PublicNetworkLoadBalancer:
+  Metadata:
+    'aws:copilot:description': 'A Network Load Balancer to distribute public traffic to your service'
+  Condition: CreateNLB # TODO: Add a condition in service template.
+  Type: AWS::ElasticLoadBalancingV2::LoadBalancer
+  Properties:
+    IpAddressType: ipv4
```
efekarakus: nit: do we need this?

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,111 @@
...
+    Scheme: internet-facing
+    Subnets:
+      Fn::importValue:
+        !Sub ${AppName}-${EnvName}-PublicSubnets
+    Type: String
```
efekarakus: Wut what is this 😂

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,111 @@
...
+
+NLBListener:
+  Type: AWS::ElasticLoadBalancingV2::Listener
+  Properties:
+    DefaultActions:
+      - TargetGroupArn: !Ref NLBTargetGroup
+        Type: forward
+    LoadBalancerArn: !Ref PublicNetworkLoadBalancer
+    Port: {{ .ListenerPort }}
+    Protocol: {{ .Protocol }}
```
efekarakus: I wonder if we should group these two under the same struct? 💭 
```suggestion
    Port: {{ .NLB.Listener.Port }}
    Protocol: {{ .NLB.Listener.Protocol }}
```

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,111 @@
...
+
+{{- if not .Aliases}}
+NLBDNSAlias:
+  Type: AWS::Route53::RecordSetGroup
+  Condition: HTTPSLoadBalancer
+  Properties:
+    HostedZoneId:
+      Fn::ImportValue:
+        !Sub "${AppName}-${EnvName}-HostedZone"
+    Comment: !Sub "NetworkLoadBalancer alias for service ${WorkloadName}"
```
iamhopaul123: Maybe
```suggestion
    Comment: !Sub "Default NetworkLoadBalancer alias for service ${WorkloadName}"
```

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
Lou1415926: This key-value pair will be removed as indicated by the `TODO`!

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,111 @@
...
+    DefaultActions:
+      - TargetGroupArn: !Ref NLBTargetGroup
+        Type: forward
+    LoadBalancerArn: !Ref PublicNetworkLoadBalancer
+    Port: {{ .ListenerPort }}
+    Protocol: {{ .Protocol }}
+{{- if eq .Protocol "TLS }}
+{{- if .AlpnPolicy}}
+    AlpnPolicy: {{ .AlpnPolicy }}
+{{- end}}
```
efekarakus: I might be wrong, but I don't think this is needed for TLS and we can ditch it for SslPolicy

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,111 @@
...
+    Certificates:
+      - CertificateArn:
+          Fn::importValue:
+            !Sub ${AppName}-${EnvName}-HTTPSCert # TODO: need to export this resource from env stack. Also, maybe rename HTTPS Cert to just Cert.
+{{- if .SSLPolicy }}
+    SslPolicy: {{ .SSLPolicy }}
```
efekarakus: should we hardcode this to: `ELBSecurityPolicy-TLS-1-2-2017-01` unless a user asks 1.1?

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,111 @@
...
+    LoadBalancerArn: !Ref PublicNetworkLoadBalancer
+    Port: {{ .ListenerPort }}
+    Protocol: {{ .Protocol }}
+{{- if eq .Protocol "TLS }}
+{{- if .AlpnPolicy}}
+    AlpnPolicy: {{ .AlpnPolicy }}
+{{- end}}
+    Certificates:
+      - CertificateArn:
+          Fn::importValue:
```
efekarakus:
```suggestion
          Fn::ImportValue:
```

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,111 @@
...
+            !Sub ${AppName}-${EnvName}-HTTPSCert # TODO: need to export this resource from env stack. Also, maybe rename HTTPS Cert to just Cert.
```
efekarakus: :homer-disappearing: this might need to be like this instead:
```
!GetAtt EnvControllerAction.HTTPSCert
```

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,111 @@
...
+
+NLBTargetGroup:
+  Metadata:
+    'aws:copilot:description': 'A target group to connect the network load balancer to your service'
+  Type: AWS::ElasticLoadBalancingV2::TargetGroup
+  Properties:
+    HealthCheckPath: {{.HTTPHealthCheck.HealthCheckPath}} # Default is '/'.
+    {{- if .HTTPHealthCheck.SuccessCodes }}
+    Matcher:
+      HttpCode: {{.HTTPHealthCheck.SuccessCodes}}
```
efekarakus: I don't think this will work with NLB since it's not aware of HTTP 🤔

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,111 @@
...
+    UnhealthyThresholdCount: {{.HTTPHealthCheck.UnhealthyThreshold}}
+    {{- end }}
+    {{- if .HTTPHealthCheck.Interval }}
+    HealthCheckIntervalSeconds: {{.HTTPHealthCheck.Interval}}
+    {{- end }}
+    {{- if .HTTPHealthCheck.Timeout }}
+    HealthCheckTimeoutSeconds: {{.HTTPHealthCheck.Timeout}}
+    {{- end }}
+    Port: !Ref ContainerPort
+    Protocol: {{ .Protocol }}
```
efekarakus: I'm not super sure if this should be `TLS` if the user has `<port>/tls`, instead it might be `TCP`.
>I can choose the communication protocol (TCP or TLS) that will be used between my NLB and my targets. If I choose TLS, communication is encrypted; this allows you to make use of complete end-to-end encryption in transit

https://aws.amazon.com/blogs/aws/new-tls-termination-for-network-load-balancers/

I think for us TLS termination will happen by the load balancer and then the request will be transmitted over TCP to to the container

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,111 @@
...
+    # IpAddressType: Omitted. Default is ipv4.
+    # Name: Omitted. Let CFN generates one.
+    # Tags: Omitted. Propagated from stack.
+    # ProtocolVersion: Only needed for HTTP/HTTPS protocol.
+    # Targets: Omitted.
+    # TargetGroupAttributes.stickiness.type: Omitted, since the only possible value for NLB is source_ip.
+
+{{- if not .Aliases}}
+NLBDNSAlias:
+  Type: AWS::Route53::RecordSetGroup
```
efekarakus: should we add a copilot metadata here and with a description for the default alias? like
``
An alias record: `{{.Name}}-nlb.{{.Env}}.{{.App}}.{{.Domain}} for the NLB`?
```

**Review by efekarakus — COMMENTED** [review]
Woot woot!

File: internal/pkg/template/templates/workloads/partials/cf/nlb.yml (hunk 11 above)
Comment (iamhopaul123) [reply internal/pkg/template/templates/workloads/partials/cf/nlb.yml]: +1 we should always depend on the env controller which outputs what env stack outputs

File: internal/pkg/template/templates/workloads/partials/cf/nlb.yml (hunk 5 above)
Comment (Lou1415926) [reply internal/pkg/template/templates/workloads/partials/cf/nlb.yml]: that was meant to be "network" 🤦🏼‍♀️

**Review by iamhopaul123 — APPROVED** [review]
:shipit:

File: internal/pkg/template/templates/workloads/partials/cf/nlb.yml [hunk 15]
```diff
@@ -0,0 +1,113 @@
...
+    VpcId:
+      Fn::ImportValue:
+        !Sub "${AppName}-${EnvName}-VpcId"
+
+{{- if not .Aliases}}
+NLBDNSAlias:
+  Metadata:
+    'aws:copilot:description': 'The default alias record for the network load balancer'
+  Type: AWS::Route53::RecordSetGroup
+  Condition: DNSDelegated
```
efekarakus, line 74: Is this a new condition that we'll add?

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -0,0 +1,113 @@
...
+        Value: false # NOTE: Default is false  TODO: remove this comment and investigate if we should surface this or not.
+    TargetType: ip
+    VpcId:
+      Fn::ImportValue:
+        !Sub "${AppName}-${EnvName}-VpcId"
+
+{{- if not .Aliases}}
+NLBDNSAlias:
+  Metadata:
+    'aws:copilot:description': 'The default alias record for the network load balancer'
```
efekarakus, line 72: nit: I wonder if we can populate this value for them. `{{.Name}}-nlb.{{.Env}.{{.App}.{{Domain}}:{{.Port}}`

efekarakus (approved):
Looks great!

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
Lou1415926: It is!

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
Lou1415926: We would need to
```
Fn::ImportValue:
                !Sub "${AppName}-${EnvName}-SubDomain"
```
Because the `Domain` isn't information directly known by the service, and is instead indirectly obtained via `ImportValue`. I wasn't super sure whether we wanted to call import function in Metadata?

PR #2925, opened 2021-10-15
docs: remove unexpected worker svc port

iamhopaul123:
Worker svc doesn't have `image.port` exposed. This PR removes unexpected `image.port` in worker svc doc.

### DIFF
--- a/site/content/docs/manifest/scheduled-job.en.md
+++ b/site/content/docs/manifest/scheduled-job.en.md
@@ -62,61 +62,7 @@ Alternatively, you can specify a cron schedule if you'd like to trigger the job 
 
 <div class="separator"></div>
 
-<a id="image" href="#image" class="field">`image`</a> <span class="type">Map</span>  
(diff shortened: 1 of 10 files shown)


site/content/docs/manifest/scheduled-job.en.md
```diff
@@ -62,61 +62,7 @@ Alternatively, you can specify a cron schedule if you'd like to trigger the job
...
-For example:
-```yaml
-image:
-  build: ./Dockerfile
-  depends_on:
-    nginx: start
-    startup: success
-```
-In the above example, the task's main container will only start after the `nginx` sidecar has started and the `startup` container has completed successfully.  
+{% include 'image-config.en.md' %}
```
Lou1415926, line 65: We were missing `image.credentials` for scheduled job as well O_0! Thanks for the fix!

Lou1415926 (approved):

huanjani (approved):
Thank you for fixing this!

PR #2926, opened 2021-10-15
docs: add manifest interpolation

iamhopaul123:
Part of #2912


mkdocs.yml
```diff
@@ -53,6 +53,7 @@ nav:
       - Sidecars: docs/developing/sidecars.en.md
       - Storage: docs/developing/storage.en.md
       - Task Definition Overrides: docs/developing/taskdef-overrides.en.md
+      - Environment Variables in Manifest: docs/developing/manifest-env-var.en.md
```
Lou1415926: I wonder if we should rename the existing `Environment Variables` section to `Environment Variables in Container` 🤔

Lou1415926: A thought for future improvement - we should probably also try to come up with a good ordering of our sidebar items. As more and more items are added in `Developing` section, it starts to look kind of messy. Maybe alphabetical order is what we should be looking for?

Lou1415926: In addition, what do you think of `Substituting Environment Variables in Manifests`?

This thought comes from the [docker page](https://docs.docker.com/compose/environment-variables/#substitute-environment-variables-in-compose-files) -  "Substitute environment variables in Compose files". I guess they do this in order to emphasize that this is to differentiate it from "setting environment variables in containers".

Given that we have two sidebar items containing the keyword "Environment Variables", I thought it might be confusing for first-time users to see both `Environment Variables` and `Environment Variables in Manifest`.

A related thought: I wonder if we should rename the existing `Environment Variables` section to `Environment Variables in Container` 🤔

Lou1415926: Alternatively, maybe we could do the same thing as the docker doc - putting `Substituting Environment Variables in Manifests` and `Environment Variables in Container` into the same doc page. 

An advantage of this is user wouldn't see two "environment variables" items in the sidebar and then get confused.

But a downside would be visibility to the new interpolation feature 🤔 .

site/content/docs/developing/manifest-env-var.en.md
```diff
@@ -0,0 +1,40 @@
+# Environment Variables in Manifest
+
+There are multiple scenarios that you might want to use environment variables in your [manifest](../manifest/overview.en.md). This page should help you find the information you need.
+
```
Lou1415926: I feel like we can ditch this line as it basically says nothing 🤣 
```suggestion
```

site/content/docs/developing/manifest-env-var.en.md
```diff
@@ -0,0 +1,40 @@
...
+## Shell environment variables
+It’s possible to use environment variables in your shell to populate values in your manifest files:
+
+``` yaml
+name: my-service
+type: Load Balanced Web Service
+image:
+  location: id.dkr.ecr.zone.amazonaws.com/project-name:${TAG}
+  port: 3333
```
Lou1415926: What do you think of just giving a partial manifest to reduce the noise in the example?
```suggestion
image:
  location: id.dkr.ecr.zone.amazonaws.com/project-name:${TAG}
```

Lou1415926: We can elaborate a bit more, for example:

Suppose the shell contains `TAG=version01`, the manifest example will be interpreted as (or "resolved to", not sure which wording is better) 

```
image:
  location: id.dkr.ecr.zone.amazonaws.com/project-name:version01
```
When Copilot defines the container, it will use the image located at "id.dkr.ecr.zone.amazonaws.com/project-name" and tagged with "version01".

site/content/docs/developing/manifest-env-var.en.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+!!! Info
+    At this stage, the value of an environment variable needs strictly to be a `string`. For example, Copilot won't automatically convert, let say "8080", to 8080 for your `port`.
```
Lou1415926:
```suggestion
At this moment, you can only substitute shell environment variables for places that accept strings, including fields that accept a `String` (e.g. `image.location`), an `Array of Strings` (e.g. `entrypoint`), or `Map` where the value should be interpreted as `String` (e.g. `secrets`).

As a counter example, you can't substitute environment variables for the `image.port` field, which accepts an `Integer`.
```

site/content/docs/developing/manifest-env-var.en.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Predefined variables
+Predefined variables are reserved environment variables resolved by Copilot before your manifest deployment:
```
Lou1415926:
```suggestion
Predefined variables are reserved variables that will be resolved by Copilot when interpreting the manifest.
```
I think we should avoid mentioning `environment` here because Copilot doesn't set these variables in their shell. We don't want user to think that we alter their local environment in any unexpected way.

site/content/docs/developing/manifest-env-var.en.md
```diff
@@ -0,0 +1,40 @@
...
+```yaml
+secrets:
+   DB_PASSWORD: /copilot/my-app/${COPILOT_ENVIRONMENT_NAME}/secrets/db_password
+environments:
+    prod:
+      secrets:
+          <specific environment variable for prod>
+    dev:
+      secrets:
+          <specific environment variable for dev>
```
Lou1415926: nit: I feel like only this segment is related to the topic - 
```suggestion
secrets:
   DB_PASSWORD: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/db_password
```

Lou1415926: We can elaborate a bit more:

Copilot will substitute `${COPILOT_APPLICATION_NAME}` and `${COPILOT_ENVIRONMENT_NAME}` with the names of the application and the environment where the workload is being deployed. For example, when you run `copilot svc deploy --app my-app --env test` to deploy the service to the `test` environment in your `my-app` application, Copilot will resolve `/copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/db_password` to `/copilot/my-app/test/secrets/db_password`. (For more information of how secret injection works, see [here](link to secrets page)).

site/content/docs/developing/manifest-env-var.en.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+Currently, available predefined environment variables include:
+
+- COPILOT_APPLICATION_NAME
+- COPILOT_ENVIRONMENT_NAME
```
Lou1415926: We should probably move this block above the example block, so that when people see `COPILOT_APPLICATION_NAME` and `COPILOT_ENVIRONMENT_NAME` in the example, they know instantly what they are.

Perhaps with a brief description of what these variable names resolve to as well.

site/content/docs/developing/manifest-env-var.en.md
```diff
@@ -0,0 +1,40 @@
...
+
+!!! Alert
+    :warning: Predefined variables are not allowed to be overridden.
```
Lou1415926:
```suggestion
    :warning: You cannot override a predefined variable with your shell environment variable. 
    
    For example, you cannot set `COPILOT_APPLICATION_NAME` to `my-other-app` in your shell environment, and run `copilot svc deploy --app my-app`.  `COPILOT_APPLICATION_NAME` will always resolve to the actual name of the application that the workload is deployed to (in this example, "my-app"), disregarding the shell environment.
```

mkdocs.yml
huanjani: What about "Manifest Environment Variables" and "Container Environment Variables"?

iamhopaul123: I like these ideas!

site/content/docs/developing/manifest-env-var.en.md
iamhopaul123: :homer-hiding-away:

site/content/docs/developing/manifest-env-var.en.md
iamhopaul123: +1 to above

> Perhaps with a brief description of what these variable names resolve to as well.

I think the name should be descriptive enough.

site/content/docs/developing/manifest-env-var.en.md
```diff
@@ -0,0 +1,39 @@
+# Environment Variables in Manifest
```
huanjani: nit:
(this one matches the subheadings on the env vars page with "my" but it seems a little odd to me)
```suggestion
# Environment Variables in My Manifest
```
or
```suggestion
# Environment Variables in the Manifest
```

huanjani (approved):

site/content/docs/developing/manifest-env-var.en.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Predefined variables
+Predefined variables are reserved variables that will be resolved by Copilot when interpreting the manifest. Currently, available predefined environment variables include:
+
+- COPILOT_APPLICATION_NAME
+- COPILOT_ENVIRONMENT_NAME
+
+```yaml
+secrets:
+   DB_PASSWORD: /copilot/my-app/${COPILOT_ENVIRONMENT_NAME}/secrets/db_password
```
efekarakus:
```suggestion
   DB_PASSWORD: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/db_password
```

site/content/docs/developing/manifest-env-var.en.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+Copilot will substitute `${COPILOT_APPLICATION_NAME}` and `${COPILOT_ENVIRONMENT_NAME}` with the names of the application and the environment where the workload is deployed. For example, when you run
+```
+$ copilot svc deploy --app my-app --env test
+```
+to deploy the service to the `test` environment in your `my-app` application, Copilot will resolve `/copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/db_password` to `/copilot/my-app/test/secrets/db_password`. (For more information of secret injection, see [here](../developing/secrets.en.md)).
+
+!!! Alert
```
efekarakus: What do you think of removing this alert? I think if they override they'll run into the error -- otherwise I feel like this might scare people for no reason 😅

efekarakus (approved):
Looks great, two minor requests then feel free to remove the label!

PR #2932, opened 2021-10-18
chore(manifest): move lb target validation to manifest pkg

iamhopaul123:
This PR moves the LB target validation from `stack` pkg  to `manifest` pkg.


internal/pkg/manifest/validate.go
```diff
@@ -1004,11 +1011,35 @@ type validateDependenciesOpts struct {
...
+	sidecarConfig     map[string]*SidecarConfig
+}
+
 type containerDependency struct {
 	dependsOn   DependsOn
 	isEssential bool
 }
 
+func validateLoadBalancerTarget(opts validateLoadBalancerTargetOpts) error {
+	targetContainer := aws.StringValue(opts.routingRule.TargetContainerCamelCase)
```
Lou1415926, line 1038: Super tiny nit/thought: since we always return early when we receive an empty input & we consider an empty input as valid, what do you think of
```
if opts.routingRule.TargetContainer == nil && opts.routingRule.TargetContainerCamelCase == nil {
   return nil
}

targetContainer == aws.StringValue(opts.routingRule.TargetContainerCamelCase)
...
```
In addition, a related thought: I wonder if we should to treat 
```
targetContainer: ""
```
and
```
// targetContainer not specified
```
differently, in that the first case where user explicitly ask for an empty `targetContainer` is invalid, and the second case is valid.

This is a super super tiny edgy case and probably no one has ever done `targetContainer: ""` in history. Just a random thought to mention!

Lou1415926 (approved):

internal/pkg/manifest/validate.go
iamhopaul123: good idea! addressed

huanjani (approved):
:shipit:

PR #2934, opened 2021-10-18
chore: merge feature branch into mainline

huanjani:


internal/pkg/manifest/validate.go
```diff
@@ -509,34 +509,23 @@ func (t TaskConfig) Validate() error {
...
-	return p.PlatformArgs.Validate()
-}
-
-// Validate returns nil if PlatformString is configured correctly.
-func (p PlatformString) Validate() error {
-	if err := validatePlatform(&p); err != nil {
-		return err
+	if p.PlatformString != nil {
+		if err := validateShortPlatform(p.PlatformString); err != nil {
+			return fmt.Errorf("validate platform: %w", err)
```
Lou1415926: We probably do not need to wrap this error, since `TaskConfig`'s `Validate` is wrapping this error with `validate "platform"` at line 495.

```suggestion
			return err
```

internal/pkg/manifest/validate.go
```diff
@@ -509,34 +509,23 @@ func (t TaskConfig) Validate() error {
 
 // Validate returns nil if PlatformArgsOrString is configured correctly.
 func (p PlatformArgsOrString) Validate() error {
-	if p.PlatformString != nil {
-		if err := p.PlatformString.Validate(); err != nil {
-			return err
+	if !p.PlatformArgs.isEmpty() {
```
Lou1415926, line 515: The `platform` field isn'r required right?

If it is not, what do you think we return early if it's empty:

```suggestion
        if p.isEmpty() {
            return nil
        }
        
        if !p.PlatformArgs.isEmpty() {
            // do the thing
        }

      if err := validateShortPlatform(p.PlatformString); err != nil {
           // do the thing
      }
      
      return nil
```

internal/pkg/manifest/validate.go
```diff
@@ -509,34 +509,23 @@ func (t TaskConfig) Validate() error {
...
-	if !p.bothSpecified() {
-		return errors.New(`fields "osfamily" and "architecture" must either both be specified or both be empty`)
-	}
-	if err := validateOS(p.OSFamily); err != nil {
-		return err
-	}
-	if err := validateArch(p.Arch); err != nil {
+// Validate returns nil if PlatformArgs is configured correctly.
+func (p PlatformArgs) Validate()error {
+	if err := validateAdvancedPlatform(p); err != nil {
```
Lou1415926: Could we keep the `p.bothSpecified` logic here instead of moving it to the caller?

internal/pkg/aws/ecs/ecs_test.go
```diff
@@ -492,7 +492,7 @@ func TestECS_Tasks(t *testing.T) {
 	}
 }
 
-func TestECS_StoppedServiceTaskss(t *testing.T) {
+func TestECS_StoppedServiceTasks(t *testing.T) {
```
iamhopaul123, line 495: 🤦

internal/pkg/cli/svc_init.go
```diff
@@ -454,6 +441,30 @@ func (o *initSvcOpts) askSvcPort() (err error) {
 	return nil
 }
 
+func (o *initSvcOpts) legitimizePlatform() error {
```
iamhopaul123: Due to the duplication I wonder if we should define it as a normal private function that doesn't belong to either `initSvcOpts` or `initJobOpts` so that it can be shared.

internal/pkg/manifest/validate.go
```diff
@@ -509,34 +509,23 @@ func (t TaskConfig) Validate() error {
 
 // Validate returns nil if PlatformArgsOrString is configured correctly.
 func (p PlatformArgsOrString) Validate() error {
-	if p.PlatformString != nil {
-		if err := p.PlatformString.Validate(); err != nil {
-			return err
+	if !p.PlatformArgs.isEmpty() {
+		if !p.PlatformArgs.bothSpecified() {
+			return errors.New(`fields "osfamily" and "architecture" must either both be specified or both be empty`)
 		}
```
iamhopaul123: It seems this should be within the `validate` function for `PlatformArgs`?

internal/pkg/manifest/validate.go
```diff
@@ -509,34 +509,23 @@ func (t TaskConfig) Validate() error {
...
 	}
-	return p.PlatformArgs.Validate()
-}
-
-// Validate returns nil if PlatformString is configured correctly.
-func (p PlatformString) Validate() error {
-	if err := validatePlatform(&p); err != nil {
-		return err
+	if p.PlatformString != nil {
+		if err := validateShortPlatform(p.PlatformString); err != nil {
```
iamhopaul123: Can we call this in the `Validate` function of `p.PlatformString`?

internal/pkg/manifest/validate.go
```diff
@@ -1134,3 +1123,32 @@ func isValidSubSvcName(name string) bool {
 	trailingMatch := trailingPunctRegExp.FindStringSubmatch(name)
 	return len(trailingMatch) == 0
 }
+
+func validateShortPlatform(platform *string) error {
```
iamhopaul123: We should probably move the logic into the `Validate` if it is only used in there

internal/pkg/manifest/validate.go
```diff
@@ -1134,3 +1123,32 @@ func isValidSubSvcName(name string) bool {
...
+	}
+	for _, validPlatform := range ValidShortPlatforms {
+		if strings.ToLower(aws.StringValue(platform)) == validPlatform {
+			return nil
+		}
+	}
+	return fmt.Errorf("platform %s is invalid; %s: %s", aws.StringValue(platform), english.PluralWord(len(ValidShortPlatforms), "the valid platform is", "valid platforms are"), english.WordSeries(ValidShortPlatforms, "and"))
+}
+
+func validateAdvancedPlatform(platform PlatformArgs) error {
```
iamhopaul123: same for this one

iamhopaul123 (commented):
LGTM overall just some suggestion for `Validate`. Also since we won't have any integ/e2e test for this feature (at least before it is firstly launched), we should probably manually test to make sure it all works as we expect (i'll do what I can on my side as well).

internal/pkg/task/env_runner.go
```diff
@@ -74,13 +78,24 @@ func (r *EnvRunner) Run() ([]*Task, error) {
 		return nil, fmt.Errorf(fmtErrSecurityGroupsFromEnv, r.Env, err)
 	}
 
+	platformVersion := "LATEST"
+	enableExec := true
+	for _, windowsOS := range validWindowsOSs {
+		if r.OS == windowsOS {
+			platformVersion = "1.0.0"
+			enableExec = false
```
Lou1415926: nit: 
```suggestion
			enableExec = false
			break
```

internal/pkg/task/config_runner.go
```diff
@@ -35,6 +35,10 @@ type ConfigRunner struct {
 
 	// Must not be nil if using default subnets.
 	VPCGetter VPCGetter
+
+	// Platform configuration
+	OS   string
+	Arch string
```
Lou1415926: Is `Arch` used anywhere?  (Same for env_runner.go)

internal/pkg/template/templates/task/cf.yml
```diff
@@ -33,6 +37,8 @@ Conditions:
     !Not [!Equals [ !Join ["", !Ref Command], ""]]
   HasEntryPoint:
     !Not [ !Equals [ !Join [ "", !Ref EntryPoint ], "" ] ]
+  HasCustomPlatform:
+    !Not [!Equals [!Ref OS, ""]]
```
Lou1415926, line 41: Can they have empty `OS` while having non-empty `Arch`?

internal/pkg/cli/task_run.go
```diff
@@ -317,6 +318,21 @@ func (o *runTaskOpts) Validate() error {
...
 	}
 
+	if noOS, noArch := o.os == "", o.arch == ""; noOS != noArch {
+		return fmt.Errorf("must specify either both `--%s` and `--%s` or neither", osFlag, archFlag)
+	}
+	if err := o.validatePlatform(); err != nil {
+		return err
+	}
+
+	if o.cpu < 0 {
```
Lou1415926: I'm wondering why we now allow `o.cpu` to be `0` 🤔 ? Same for `o.memory`.

internal/pkg/cli/task_run.go
```diff
@@ -348,6 +368,21 @@ func (o *runTaskOpts) Validate() error {
 	return nil
 }
 
+func (o *runTaskOpts) validatePlatform() error {
```
Lou1415926, line 369: This logic is repeating in `manifest` as well - we should probably come up with a way to refactor the code. Maybe we can change the function signature to something like `validatePlatform(os string, arch string)`... not necessarily a change that need to be done in this PR, but we might want to add this as a to-do.

internal/pkg/cli/task_run.go
```diff
@@ -423,6 +458,28 @@ func (o *runTaskOpts) validateFlagsWithSecurityGroups() error {
...
+		return fmt.Errorf("CPU %d must be at least %d for a Windows-based task", o.cpu, manifest.MinWindowsTaskCPU)
+	}
+	if o.memory < manifest.MinWindowsTaskMemory {
+		return fmt.Errorf("memory %d must be at least %d for a Windows-based task", o.memory, manifest.MinWindowsTaskMemory)
+	}
+	return nil
+}
+
+func isWindowsOS(os string) bool {
+	for _, windowsOS := range task.ValidOSs() {
```
Lou1415926: Isn't `osLinux` also returned from `ValidOSs()`?

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -143,6 +147,14 @@ func (s *LoadBalancedWebService) BuildRequired() (bool, error) {
 	return requiresBuild(s.ImageConfig.Image)
 }
 
+// IsWindows returns whether or not the service is building with a Windows OS.
+func (t TaskConfig) IsWindows() bool {
+	if t.Platform.PlatformString == nil {
+		return false
```
Lou1415926: Could it be that `t.Platform.PlatformString == nil` while `t.Platform.PlatformArgs` is not empty?

site/content/docs/include/common-svc-fields.en.md
```diff
@@ -33,8 +33,19 @@ Amount of memory in MiB used by the task. See the [Amazon ECS docs](https://docs
 
 <div class="separator"></div>
 
-<a id="platform" href="#platform" class="field">`platform`</a> <span class="type">String</span>  
-Operating system and architecture (formatted as `[os]/[arch]`) to pass with `docker build --platform`.
+<a id="platform" href="#platform" class="field">`platform`</a> <span class="type">String or Map</span>
```
Lou1415926, line 36: I'm not sure if we should include `platform` doc in rdws's manifest since it's technically not configurable at all, right?

internal/pkg/template/workload.go
```diff
@@ -40,6 +40,13 @@ const (
 	DisablePublicIP         = "DISABLED"
 	PublicSubnetsPlacement  = "PublicSubnets"
 	PrivateSubnetsPlacement = "PrivateSubnets"
+
+	// RuntimePlatform configuration.
```
Lou1415926, line 44: I'm wondering if we can define these variables at one place and reuse them across packages. Perhaps `deploy` would be a good place to do this? What do you think?

internal/pkg/task/task.go
```diff
@@ -49,10 +51,18 @@ type Task struct {
 
 const (
 	startedBy = "copilot-task"
+
+	// Platform options.
```
Lou1415926, line 56: Same concerns here!

Lou1415926 (commented):
I have a couple of questions - mainly because I probably don't have enough context of the code and am therefore missing something🤚🏼  I will also go ahead and read more as well, and mark questions as resolved if I figure out the answer before you respond!

internal/pkg/cli/svc_init.go
huanjani: I like being able to check for o.image and set o.platform within this method rather than in `Execute()`, and if it were a function rather than a method, we'd have to pass in so many arguments. I'd prefer to follow the Rule of Three in this situation. 😄

internal/pkg/cli/task_run.go
huanjani: 🤦🏼 the `=` got lost in all the many changes this went through! Restored.

internal/pkg/cli/task_run.go
huanjani: I agree, but it's slightly tricky because more platforms are allowed in the manifest than in `task run` because of the transformer. For example, `windows` is allowed in the manifest, but not as a `task run` flag.

internal/pkg/cli/task_run.go
huanjani: Doh! TY!

internal/pkg/cli/svc_init.go
iamhopaul123: If too many arguments is the only concern I still think we need to avoid duplication as possible since that means we only need to update in one place rather than both, when we want to adjust the logic. Can we put them into a struct?

internal/pkg/manifest/lb_web_svc.go
huanjani: Yes, good catch! That was a holdover from merging different stages. Fixed!

site/content/docs/include/common-svc-fields.en.md
huanjani: Correct. But this partial is not included in the rdws manifest, luckily! 😄

internal/pkg/template/templates/task/cf.yml
huanjani: Nope. Good question. We validate that both are passed in.

internal/pkg/template/workload.go
huanjani: Okay, the uppercase ones live in `template` and the lowercase ones live in `manifest`. Which actually makes a lot of sense!

internal/pkg/manifest/validate.go
huanjani: p.PlatformString doesn't have its own `Validate` function because it's just a *string now, remember? I integrated the logic of `validateShortPlatform` into `Validate`, though!

internal/pkg/cli/svc_init.go
huanjani: Yeah, you're right! Done!

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -149,9 +149,6 @@ func (s *LoadBalancedWebService) BuildRequired() (bool, error) {
 
 // IsWindows returns whether or not the service is building with a Windows OS.
 func (t TaskConfig) IsWindows() bool {
```
Lou1415926: Should we move this function to `manifest/workload.go`? Seems like all `TaskConfig` functions are defined there!

internal/pkg/task/config_runner.go
```diff
@@ -71,10 +70,11 @@ func (r *ConfigRunner) Run() ([]*Task, error) {
 	}
 	platformVersion := "LATEST"
 	enableExec := true
-	for _, windowsOS := range validWindowsOSs {
+	for _, windowsOS := range ValidWindowsOSs {
```
Lou1415926: I wonder if we can define a function in `task` package:
```go
func IsValidWindowOS() (os string){
   for _, validWindows := range validWindowsOSs {
      if os == validWindows {
           return true
       }
   }
   return false
}
```
and call it here
```go
if IsValidWindowsOS(r.os) {
   platformVersion = "1.0.0"
   enableExec = false
}
```

Then in `cli/task_run.go`, instead of defining a `isWindowsOS` and call it there, we can reuse this function 
```go
// In cli/task_run.go
if task.IsValidWindowsOS(o.os) {
// do the thing
}
```
This way, only the `task` package owns the logic that determines whether a task has a valid windows OS.

Lou1415926 (approved):

internal/pkg/task/task.go
```diff
@@ -67,6 +67,15 @@ var (
 	fmtTaskFamilyName = "copilot-%s"
 )
 
+func IsValidWindowsOS(os string) bool {
```
iamhopaul123, line 71: comment for public function please 🙏

iamhopaul123 (approved):

PR #2938, opened 2021-10-20
chore(manifest): ignore comment for interpolation

iamhopaul123:

### DIFF
--- a/internal/pkg/manifest/interpolate.go
+++ b/internal/pkg/manifest/interpolate.go
@@ -20,9 +20,10 @@ var (
 	// Environment variable names consist solely of uppercase letters, digits, and underscore,
 	// and do not begin with a digit. （https://pubs.opengroup.org/onlinepubs/007904875/basedefs/xbd_chap08.html）
 	interpolatorEnvVarRegExp = regexp.MustCompile(`\${([_a-zA-Z][_a-zA-Z0-9]*)}`)
-	// A yaml comment
+	// A YAML comment:
 	// 1. starts with "#"
-	// 2. must not proceed a non whitespace character
+	// 2. is preceded by at least one whitespace, except for when the line starts with a comment,
+	// then it can be proceeded by zero or more whitespace
 	// 3. ends with zero or more "\n"
 	yamlCommentRegExp = regexp.MustCompile(`(^\s*|\s+)#.*\n*`)
 )
@@ -49,11 +50,10 @@ func (i *Interpolator) Interpolate(s string) (string, error) {
 	rest := s
 	for {
 		// Only get the first match.
-		matches := yamlCommentRegExp.FindAllString(rest, 1)
-		if len(matches) == 0 {
+		comment := yamlCommentRegExp.FindString(rest)
+		if comment == "" {
 			break
 		}
-		comment := matches[0]
 		splitedRest := strings.SplitN(rest, comment, 2)
 		interpolated, err = i.interpolatePart(splitedRest[0])
 		if err != nil {
--- a/internal/pkg/manifest/interpolate.go
+++ b/internal/pkg/manifest/interpolate.go
@@ -57,19 +57,29 @@ func (i *Interpolator) Interpolate(s string) (string, error) {
 }
 
 func (i *Interpolator) applyInterpolation(node *yaml.Node) error {
-	for _, content := range node.Content {
-		if err := i.applyInterpolation(content); err != nil {
+	switch node.Tag {
+	case "!!map":
+		// The content of a map always come in pairs. If the node pair exists, return the map node.
+		// Note that the rest of code massively uses yaml node tree.
+		// Please refer to https://www.efekarakus.com/2020/05/30/deep-dive-go-yaml-cfn.html
+		for idx := 0; idx < len(node.Content); idx += 2 {
+			if err := i.applyInterpolation(node.Content[idx+1]); err != nil {
+				return err
+			}
+		}
+	case "!!str":
+		interpolated, err := i.interpolatePart(node.Value)
+		if err != nil {
 			return err
 		}
+		node.Value = interpolated
+	default:
+		for _, content := range node.Content {
+			if err := i.applyInterpolation(content); err != nil {
+				return err
+			}
+		}
 	}
-	if node.Tag != "!!str" {
-		return nil
-	}
-	interpolated, err := i.interpolatePart(node.Value)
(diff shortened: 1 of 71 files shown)


internal/pkg/manifest/interpolate.go
```diff
@@ -20,6 +20,11 @@ var (
 	// Environment variable names consist solely of uppercase letters, digits, and underscore,
 	// and do not begin with a digit. （https://pubs.opengroup.org/onlinepubs/007904875/basedefs/xbd_chap08.html）
 	interpolatorEnvVarRegExp = regexp.MustCompile(`\${([_a-zA-Z][_a-zA-Z0-9]*)}`)
+	// A yaml comment
+	// 1. starts with "#"
+	// 2. must not proceed a non whitespace character
+	// 3. ends with zero or more "\n"
```
Lou1415926: Does this effectively mean that a YAML comment is always at the rightmost part of a line? 

If yes, maybe we can substitute `\n*` with `($|\n)`.

internal/pkg/manifest/interpolate.go
```diff
@@ -20,6 +20,11 @@ var (
 	// Environment variable names consist solely of uppercase letters, digits, and underscore,
 	// and do not begin with a digit. （https://pubs.opengroup.org/onlinepubs/007904875/basedefs/xbd_chap08.html）
 	interpolatorEnvVarRegExp = regexp.MustCompile(`\${([_a-zA-Z][_a-zA-Z0-9]*)}`)
+	// A yaml comment
```
Lou1415926: nit:
```suggestion
	// A YAML comment:
```

internal/pkg/manifest/interpolate.go
```diff
@@ -20,6 +20,11 @@ var (
...
+	// 1. starts with "#"
+	// 2. must not proceed a non whitespace character
```
Lou1415926:
```suggestion
	// 2. is preceded by at least one whitespace, except for when the line starts with a comment, then it can be preceded by zero or more whitespace
```

internal/pkg/manifest/interpolate.go
```diff
@@ -20,6 +20,11 @@ var (
...
+	// 3. ends with zero or more "\n"
+	yamlCommentRegExp = regexp.MustCompile(`[^\s]*#.*\n*`)
```
Lou1415926:
```suggestion
	yamlCommentRegExp = regexp.MustCompile(`[(^\s*)(\s+)]#.*\n*`)
```

internal/pkg/manifest/interpolate.go
```diff
@@ -39,6 +44,33 @@ func NewInterpolator(appName, envName string) *Interpolator {
 
 // Interpolate substitutes environment variables in a string.
 func (i *Interpolator) Interpolate(s string) (string, error) {
+	var replaced, interpolated string
+	var err error
+	rest := s
+	for {
+		// Only get the first match.
+		matches := yamlCommentRegExp.FindAllString(rest, 1)
```
Lou1415926: What about using [`FindString`](https://pkg.go.dev/regexp#Regexp.FindString)?

internal/pkg/manifest/interpolate.go
```diff
@@ -39,6 +44,33 @@ func NewInterpolator(appName, envName string) *Interpolator {
...
+		if len(matches) == 0 {
+			break
+		}
+		comment := matches[0]
```
Lou1415926: `matches[0]` would also include the leading whitespace, which is technically not part of a comment, perhaps we should use a subgroup in the regex to extract only the comment part.

internal/pkg/manifest/interpolate.go
```diff
@@ -39,6 +44,33 @@ func NewInterpolator(appName, envName string) *Interpolator {
...
+		splitedRest := strings.SplitN(rest, comment, 2)
+		interpolated, err = i.interpolatePart(splitedRest[0])
+		if err != nil {
+			return "", err
+		}
+		replaced += fmt.Sprint(interpolated, comment)
+		rest = splitedRest[1]
```
Lou1415926: Just a thought 💭 : I wonder if we should process the string lines by lines... seems more intuitive..

iamhopaul123: I don't want to deal with the annoying additional "\n" at the end of the file and it would be slightly more efficient. That's my original thought. But yeah for readability and less error prone i'll make the change

iamhopaul123: OK after trying to do it line by line I found it even more complicated for the logic. Let me know your thought about it!

internal/pkg/manifest/interpolate.go
iamhopaul123: not doing it line by line so this doesn't apply...

internal/pkg/manifest/interpolate.go
iamhopaul123: That'll make the logic much more complicated since golang doesn't support negative lookahead...I think it is probably fine to let the comment include preceding whitespaces since it doesn't affect the final results?

Lou1415926: Would it work if you use `FindStringSubmatch`, and then get the subgroup by `matches[2]` ? `matches[0]` would be the whole matched string, `matches[1]` matches the first subgroup, which is `(^\s*|\s+)`, and `matches[2]` would be the comment part, which is `(#.*)`?

internal/pkg/manifest/interpolate.go
Lou1415926: Thanks for trying! I think what we have now also works.

internal/pkg/manifest/interpolate.go
efekarakus: Hiiiii I haven't read the PR at all, just saw this comment 😅 but throwing an alternative to consider:

https://pkg.go.dev/gopkg.in/yaml.v3#Node

What if we did the interpolation only on the `Value` field?
Decode the manifest into a `yaml.Node` and the loop through the contents replacing only the `Value` field if the Node is a `!!str` this way you won't touch the comments. Just an idea, I'll go back to disappearing 👻

iamhopaul123: As you wish boss 👮‍♀️

Lou1415926 (approved):

huanjani (approved):
👏🏼

Issue #2943, opened 2021-10-22
Getting Started example fails in codepipeline CodeBuild step with la -lah ./infrastructure exit status 2


mark-brooks-roostify:
Here's the error from the tail of my Build log:

```

[Container] 2021/10/22 16:10:23 Running command ls -lah ./infrastructure
--
879 | ls: cannot access ./infrastructure: No such file or directory
880 |  
881 | [Container] 2021/10/22 16:10:23 Command did not exit successfully ls -lah ./infrastructure exit status 2
882 | [Container] 2021/10/22 16:10:23 Phase complete: POST_BUILD State: FAILED
883 | [Container] 2021/10/22 16:10:23 Phase context status code: COMMAND_EXECUTION_ERROR Message: Error while executing command: ls -lah ./infrastructure. Reason: exit status 2
884 | [Container] 2021/10/22 16:10:23 Expanding base directory path: .
885 | [Container] 2021/10/22 16:10:23 Assembling file list
886 | [Container] 2021/10/22 16:10:23 Expanding .
887 | [Container] 2021/10/22 16:10:23 Expanding file paths for base directory .
888 | [Container] 2021/10/22 16:10:23 Assembling file list
889 | [Container] 2021/10/22 16:10:23 Expanding infrastructure/*
890 | [Container] 2021/10/22 16:10:23 Phase complete: UPLOAD_ARTIFACTS State: FAILED
891 | [Container] 2021/10/22 16:10:23 Phase context status code: CLIENT_ERROR Message: no matching artifact paths found

```

I'm following this [demo](https://aws.amazon.com/blogs/containers/enabling-continuous-workflows-for-aws-app-runner-service-with-persistency-using-aws-copilot-cli/)

mark-brooks-roostify:
Here's some additional, earlier log output:

```
[Container] 2021/10/22 16:10:20 Phase complete: BUILD State: SUCCEEDED
830 | [Container] 2021/10/22 16:10:20 Phase context status code:  Message:
831 | [Container] 2021/10/22 16:10:20 Entering phase POST_BUILD
832 | [Container] 2021/10/22 16:10:20 Running command ls -l
833 | total 38712
834 | -rwxrwxr-x 1 root root     3158 Oct 22 15:47 CONTRIBUTING.md
835 | -rwxrwxr-x 1 root root      814 Oct 22 15:47 Dockerfile
836 | -rwxrwxr-x 1 root root      926 Oct 22 15:47 LICENSE
[...]

PR #2946, opened 2021-10-22
chore(pipeline): codebuild should error earlier if no workload

iamhopaul123:
partially address #2943


internal/pkg/template/templates/cicd/buildspec.yml
```diff
@@ -25,12 +25,28 @@ phases:
...
+      - job_list=$(echo $job_ls_result | jq '.jobs')
+      - >
+        if [ ! "$job_list" = null ]; then
+          jobs=$(echo $job_ls_result | jq '.jobs[].name' | sed 's/"//g');
+        fi
+      # Raise error if no services or jobs are found.
+      - >
+        if [ "$svc_list" = null ] && [ "$job_list" = null ]; then
+          echo "No services or jobs found for the pipeline to deploy. Please create at least one service or job and push the manifest to the remote." 1>&2;
+          exit 1;
```
Lou1415926, line 48: <strike>Should we exit with some other number because https://tldp.org/LDP/abs/html/exitcodes.html "According to the above table, exit codes 1 - 2, 126 - 165, and 255 [1] have special meanings, and should therefore be avoided for user-specified exit parameters."?</strike>

On a second thought, I agree that `exit 1` should be fine as it is widely used and is general enough.

iamhopaul123: Hmm good question. Exit code 1 is for general use. Although the exit code itself doesn't carry a lot of meaning, the error log we present to them should make much more sense. What exit code do you think we should return instead? Maybe avoid using any reserved value?

Lou1415926: Updated my comment above ^

Lou1415926 (approved):

efekarakus (approved):
You tested this right boss? to ensure existing services won't be impacted

iamhopaul123:
> You tested this right boss? to ensure existing services won't be impacted

Yeah I manually tested it before.

PR #2953, opened 2021-10-25
chore: update recommended actions for secret init

Lou1415926:
Now that manifest interpolation is implemented, the `secret init` recommended actions should be updated to encourage user to take advantage of manifest interpolation.


internal/pkg/cli/secret_init.go
```diff
@@ -414,12 +415,12 @@ func (o *secretInitOpts) RecommendActions() error {
...
 				secretsPerEnv[envName] = make(map[string]string)
 			}
-			secretsPerEnv[envName][template.ToSnakeCaseFunc(secretName)] = fmt.Sprintf(fmtSecretParameterName, o.appName, envName, secretName)
+			secretsPerEnv[envName][template.ToSnakeCaseFunc(secretName)] = fmt.Sprintf(fmtSecretParameterNameManifestRecommendation, secretName)
 		}
 	}
 
 	templateRaw := `{{range $env, $secrets := .SecretsPerEnv -}}
-{{$env}}
+{{$env}}:
```
efekarakus: I might be misunderstanding but I think the cool thing now is that we don't need to `range $env, $secrets` so that our recommendation is even shorter.

instead we can just provide the generic version:
```
secrets:
  <name>: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/%s
```

Lou1415926: Good call! Missed this. Updated!

efekarakus (approved):

internal/pkg/cli/secret_init.go
```diff
@@ -24,15 +22,15 @@ import (
...
 	"github.com/aws/copilot-cli/internal/pkg/term/color"
 	"github.com/aws/copilot-cli/internal/pkg/term/log"
 	"github.com/aws/copilot-cli/internal/pkg/term/prompt"
 	"github.com/aws/copilot-cli/internal/pkg/term/selector"
 )
 
 const (
-	fmtSecretParameterName = "/copilot/%s/%s/secrets/%s"
+	fmtSecretParameterName           = "/copilot/%s/%s/secrets/%s"
+	fmtSecretParameterNameMftExample = "/copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_APPLICATION_NAME}/secrets/%s"
```
huanjani: ???
```suggestion
	fmtSecretParameterNameMftExample = "/copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/%s"
```

huanjani (approved):
🚢

PR #2954, opened 2021-10-25
chore(deploy): create LB web service stack with NLB options

Lou1415926:
This PR:
1. Conditionally include NLB partial template in the LBWS template
2. Add support to create create LB web service stack with NLB options


internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -79,14 +84,56 @@ func NewLoadBalancedWebService(mft *manifest.LoadBalancedWebService, env, app st
...
 	if err != nil {
 		return nil, err
 	}
 	webSvc.httpsEnabled = true
 	return webSvc, nil
 }
 
+// NewNetworkLoadBalancedWebService creates a new LoadBalancedWebService stack from a manifest file. It creates a
+// Network Load Balancer, with or without an Application Load Balancer.
+func NewNetworkLoadBalancedWebService(mft *manifest.LoadBalancedWebService, env, app string, rc RuntimeConfig, cidrBlocks []string) (*LoadBalancedWebService, error) {
```
iamhopaul123: There seems to be a lot of duplication as https://github.com/aws/copilot-cli/pull/2954/files#diff-7e36dd6cfea6499215e00fda7bd632bec4a52ffec8428ebe43981553b39c0cd2R55. I wonder if we could extract to a private function.

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -79,14 +84,56 @@ func NewLoadBalancedWebService(mft *manifest.LoadBalancedWebService, env, app st
...
+		publicSubnetCIDRBlocks: cidrBlocks,
+
+		parser: parser,
+	}, nil
+}
+
+// NewNetworkLoadBalancedWebServiceWithDNS creates a new LoadBalancedWebService stack from a manifest file. It creates a
+// Network Load Balancer, with or without an Application Load Balancer, and use the delegated domain name to create a
+// friendly alias for the Network Load Balancer.
+func NewNetworkLoadBalancedWebServiceWithDNS(mft *manifest.LoadBalancedWebService, env, app string, rc RuntimeConfig, cidrBlocks []string) (*LoadBalancedWebService, error) {
```
iamhopaul123: Question: do we use this if they config any TLS listener?

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
```diff
@@ -24,6 +24,9 @@ Parameters:
   HTTPSEnabled:
     Type: String
     AllowedValues: [true, false]
+  DNSDelegated:
```
iamhopaul123, line 27: Is this going to be specifically used for NLB, since we already have `HTTPSLoadBalancer` for ALB? If so, maybe we should rename it to be like `TLSEnabled`?

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
Lou1415926: We use this if their app is associated with a domain. It is a prerequisite of using TLS.

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
Lou1415926: DNSDelegated doesn't imply that user wants to use TLS termination though.

iamhopaul123 (approved):

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -43,50 +44,53 @@ type LoadBalancedWebService struct {
 	manifest     *manifest.LoadBalancedWebService
 	httpsEnabled bool
 
+	// Fields for LoadBalancedWebService that needs a Network Load Balancer.
+	dnsDelegationEnabled   bool
```
efekarakus: Maybe one thing to clarify as field comments is the difference between `httpsEnabled` and `dnsDelegationEnabled`? 🥺

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -43,50 +44,53 @@ type LoadBalancedWebService struct {
...
-			logRetention:        mft.Logging.Retention,
-			tc:                  mft.TaskConfig,
-			taskDefOverrideFunc: override.CloudFormationTemplate,
-		},
-		manifest:     mft,
-		httpsEnabled: false,
-
-		parser: parser,
-	}, nil
+// NewHTTPLoadBalancedWebService creates a new LoadBalancedWebService stack from a manifest file.
```
efekarakus:
```suggestion
// NewHTTPLoadBalancedWebService creates a new CFN stack with an ECS service and an ALB from a manifest file.
```

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -43,50 +44,53 @@ type LoadBalancedWebService struct {
...
+func NewHTTPLoadBalancedWebService(mft *manifest.LoadBalancedWebService, env, app string, rc RuntimeConfig) (*LoadBalancedWebService, error) {
+	return newLoadBalancedWebService(mft, env, app, rc)
 }
 
 // NewHTTPSLoadBalancedWebService  creates a new LoadBalancedWebService stack from its manifest that needs to be deployed to
```
efekarakus: nit: can we update this comment as well like above.

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -43,50 +44,53 @@ type LoadBalancedWebService struct {
...
+	webSvc, err := NewHTTPLoadBalancedWebService(mft, env, app, rc)
 	if err != nil {
 		return nil, err
 	}
 	webSvc.httpsEnabled = true
 	return webSvc, nil
 }
 
+// NewNetworkLoadBalancedWebService creates a new LoadBalancedWebService stack from a manifest file. It creates a
+// Network Load Balancer, with or without an Application Load Balancer.
```
efekarakus: What do you think of this alternative API:
1. We have a single constructor `NewLoadBalancedWebService`
2. We have many [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis)?
```go
func NewLoadBalancedWebService(mft *manifest.LoadBalancedWebService, env, app string, rc RuntimeConfig, opts  ...LoadBalancedWebServiceOption) (*LoadBalancedWebService, error)

type LoadBalancedWebServiceOption func(lbws *LoadBalancedWebService)

func WithNLB(cidrBlocks) LoadBalancedWebServiceOption { ... }
func WithNLBAndTLS(cidrBlocks) LoadBalancedWebServiceOption { ... }
```

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -66,12 +67,33 @@ NLBTargetGroup:
...
 
+NLBSecurityGroup:
+  Metadata:
+    'aws:copilot:description': 'A security group for your network load balancer to route traffic to service'
+  Type: AWS::EC2::SecurityGroup
+  Properties:
+    GroupDescription: Allow access from the network load balancer to service
+    SecurityGroupIngress:
+{{range $cidr := .NLB.PublicSubnetCIDRs}}
+      - CidrIp: {{$cidr}}
```
efekarakus, line 78: cool! one thing that we can do in our e2e test is to write a test that ensures that if I tried to access a task IP address from the public internet without going through NLB the request will be blocked.

This will help us feel good about security

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
```diff
@@ -45,6 +45,9 @@ NetworkConfiguration:
       {{- range $sg := .Network.SecurityGroups}}
       - {{$sg}}
       {{- end}}
+      {{- if .NLB}}
+      - !Ref NLBSecurityGroup
+      {{- end}}
```
efekarakus, line 50: one thing that we need to be careful of slowly is that there is a limit of 5 sg for the service.

I wonder if we should instead create a `ServiceSecurityGroup` that has ingress from `NLBSecurityGroup` so that we can always add more ingresses.
This should be a two-way door for us so it's okay if you don't update now.

efekarakus (commented):
Looks great!

internal/pkg/cli/svc_package.go
```diff
@@ -123,19 +123,16 @@ func newPackageSvcOpts(vars packageSvcVars) (*packageSvcOpts, error) {
...
-				serializer, err = stack.NewHTTPSLoadBalancedWebService(t, env.Name, app.Name, rc)
-				if err != nil {
-					return nil, fmt.Errorf("init https load balanced web service stack serializer: %w", err)
-				}
-			} else {
-				serializer, err = stack.NewLoadBalancedWebService(t, env.Name, app.Name, rc)
-				if err != nil {
-					return nil, fmt.Errorf("init load balanced web service stack serializer: %w", err)
-				}
+				opts = append(opts, stack.WithHTTPS())
```
efekarakus, line 131: yay awesome refactors! looks like it allowed us to get rid of else statements

efekarakus (approved):

PR #2965, opened 2021-10-28
chore(manifest) add comment to explain exclusive field validation that happens during unmarshal

Lou1415926:


efekarakus (approved):

iamhopaul123 (approved):
LGTM! Also we might need to do for every struct unmarshal with mutually exclusive fields?

Issue #2960, opened 2021-10-26
[Design] Copilot Better Workload Init/Deploy UX


iamhopaul123:
## Problems

Since our svc/job delete doesn’t remove local manifest when users delete their workloads, this can lead to the fact that when they are trying to recreate a workload that they deleted before, Copilot still asks them for unnecessary info (like Dockerfile path, port etc.) that will never get rendered in their manifests (we also don’t override the manifest file). This problem can happen more common than we previously thought, because svc/job delete is very likely to be the command users would use to “unstage” their workloads rather than setting the task number to 0. (See https://github.com/aws/copilot-cli/issues/2388, https://github.com/aws/copilot-cli/issues/1381, https://github.com/aws/copilot-cli/issues/1460)

Additionally, when Copilot deploys the workload, it honors the local directory name rather than the manifest name. It could be problematic if the "local directory name", "workload name in manifest", and "SSM record" don’t match with each other. (See https://github.com/aws/copilot-cli/issues/1994)

Given the two above-mentioned issues, we need to make improvement to our current workload init/deploy UX, so as to bring better user experience and clear obstacles especially for our new users as possible.

## Proposal

If local manifest already exists:
- [ ] `svc/job init` won't ask repetitive questions 
- [ ] `svc/job deploy` will error out if the directory name and the workload name in the manifest are not consistent
- [ ] `svc/job deploy` will be able to deploy the workload and create any missing SSM record automatically

PR #2966, opened 2021-10-29
chore(cli): skip asking questions if local manifest exists

iamhopaul123:
Part of #2960

Note that this PR changes the order of questions for `svc/job init`: we'll ask the name before the type so that if we find the local manifest already exists, we won't ask any repetitive questions.

However, `init` experience hasn't changed yet as we won't know if they want to init a service or a job, unless we ask them first.

### DIFF
--- a/internal/pkg/cli/job_init.go
+++ b/internal/pkg/cli/job_init.go
@@ -65,6 +65,7 @@ type initJobOpts struct {
 	prompt       prompter
 	sel          initJobSelector
 	dockerEngine dockerEngine
+	mftReader    manifestReader
 
 	// Outputs stored on successful actions.
 	manifestPath string
@@ -111,6 +112,7 @@ func newInitJobOpts(vars initJobVars) (*initJobOpts, error) {
 		prompt:       prompter,
 		sel:          sel,
 		dockerEngine: dockerengine.New(exec.NewCmd()),
+		mftReader:    ws,
 		initParser: func(path string) dockerfileParser {
 			return dockerfile.New(fs, path)
 		},
@@ -163,6 +165,15 @@ func (o *initJobOpts) Ask() error {
 	}
 	if err := o.askJobName(); err != nil {
 		return err
+	}
+	_, err := o.mftReader.ReadWorkloadManifest(o.name)
+	if err == nil {
+		// Return early if local manifest for the service already exists.
+		return nil
+	}
+	_, ok := err.(*workspace.ErrFileNotExists)
+	if !ok {
+		return fmt.Errorf("check if local manifest for job %s exists: %w", o.name, err)
 	}
 	dfSelected, err := o.askDockerfile()
 	if err != nil {
@@ -248,7 +259,6 @@ func (o *initJobOpts) askJobName() error {
 	if o.name != "" {
 		return nil
 	}
-
 	name, err := o.prompt.Get(
 		fmt.Sprintf(fmtWkldInitNamePrompt, color.Emphasize("name"), color.HighlightUserInput(o.wkldType)),
 		fmt.Sprintf(fmtWkldInitNameHelpPrompt, job, o.appName),
--- a/internal/pkg/cli/job_init.go
+++ b/internal/pkg/cli/job_init.go
@@ -160,20 +160,24 @@ func (o *initJobOpts) Validate() error {
 
 // Ask prompts for fields that are required but not passed in.
 func (o *initJobOpts) Ask() error {
-	if err := o.askJobType(); err != nil {
-		return err
-	}
 	if err := o.askJobName(); err != nil {
 		return err
 	}
-	_, err := o.mftReader.ReadWorkloadManifest(o.name)
+	localMft, err := o.mftReader.ReadWorkloadManifest(o.name)
 	if err == nil {
-		// Return early if local manifest for the service already exists.
+		jobType, err := localMft.WorkloadType()
+		if err != nil {
+			return fmt.Errorf("read type for service %s from local manifest: %w", o.name, err)
+		}
+		o.wkldType = jobType
 		return nil
 	}
 	_, ok := err.(*workspace.ErrFileNotExists)
 	if !ok {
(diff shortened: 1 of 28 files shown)


internal/pkg/cli/job_init.go
```diff
@@ -158,10 +160,23 @@ func (o *initJobOpts) Validate() error {
...
-	if err := o.askJobType(); err != nil {
+	if err := o.askJobName(); err != nil {
 		return err
 	}
-	if err := o.askJobName(); err != nil {
+	localMft, err := o.mftReader.ReadWorkloadManifest(o.name)
+	if err == nil {
+		jobType, err := localMft.WorkloadType()
+		if err != nil {
+			return fmt.Errorf("read type for service %s from local manifest: %w", o.name, err)
```
efekarakus:
```suggestion
			return fmt.Errorf("read 'type' field for job %s from local manifest: %w", o.name, err)
```

internal/pkg/cli/job_init.go
```diff
@@ -158,10 +160,23 @@ func (o *initJobOpts) Validate() error {
...
+		}
+		o.wkldType = jobType
+		return nil
+	}
+	_, ok := err.(*workspace.ErrFileNotExists)
+	if !ok {
```
efekarakus: in order to work with wrapped errors:
```suggestion
	var errNotFound *ErrFileNotExists
	if !errors.As(err, &errNotFound) {
```

internal/pkg/cli/job_init.go
```diff
@@ -158,10 +160,23 @@ func (o *initJobOpts) Validate() error {
...
+		return fmt.Errorf("check if local manifest for job %s exists: %w", o.name, err)
```
efekarakus:
```suggestion
		return fmt.Errorf("read manifest file for job %s: %w", o.name, err)
```

internal/pkg/cli/job_init.go
```diff
@@ -158,10 +160,23 @@ func (o *initJobOpts) Validate() error {
...
 	}
-	if err := o.askJobName(); err != nil {
+	localMft, err := o.mftReader.ReadWorkloadManifest(o.name)
+	if err == nil {
+		jobType, err := localMft.WorkloadType()
+		if err != nil {
+			return fmt.Errorf("read type for service %s from local manifest: %w", o.name, err)
+		}
+		o.wkldType = jobType
+		return nil
```
efekarakus: Should we log anything in this scenario saying that the file already exists so we are skipping prompts?

internal/pkg/cli/svc_init.go
```diff
@@ -217,31 +219,40 @@ func (o *initSvcOpts) Validate() error {
...
-	if err := o.askSvcType(); err != nil {
+	if err := o.askSvcName(); err != nil {
 		return err
 	}
-	if err := o.askSvcName(); err != nil {
+	localMft, err := o.mftReader.ReadWorkloadManifest(o.name)
+	if err == nil {
+		svcType, err := localMft.WorkloadType()
+		if err != nil {
+			return fmt.Errorf("read type for service %s from local manifest: %w", o.name, err)
```
efekarakus: same feedback here for `svc_init.go` as `job_init.go`

internal/pkg/workspace/workspace.go
```diff
@@ -185,26 +185,21 @@ func (ws *Workspace) workloadNames(match func(string) bool) ([]string, error) {
...
-		return nil, fmt.Errorf("read job %s manifest file: %w", name, err)
+		return nil, err
 	}
-	return mf, nil
-}
-
-func (ws *Workspace) readWorkloadManifest(name string) ([]byte, error) {
-	return ws.read(name, manifestFileName)
+	if mftName != mftDirName {
+		return nil, fmt.Errorf("name of the manifest and directory do not match")
```
efekarakus: should we add the values here:
```suggestion
		return nil, fmt.Errorf("name of the manifest '%s' and directory '%s' do not match", mftName, mftDirName)
```

internal/pkg/workspace/workspace.go
```diff
@@ -489,3 +482,27 @@ func RelPath(fullPath string) (string, error) {
...
+
+// WorkloadManifest represents raw local workload manifest.
+type WorkloadManifest []byte
+
+func (w WorkloadManifest) workloadName() (string, error) {
+	wl := struct {
+		Name string `yaml:"name"`
+	}{}
+	if err := yaml.Unmarshal(w, &wl); err != nil {
+		return "", err
```
efekarakus: nit: should we return a wrapped error here? 
```suggestion
		return "", fmt.Errorf("unmarshal manifest file to retrieve 'name': %w", err)
```

efekarakus (commented):
Awesome! small feedbacks

internal/pkg/cli/job_init.go
huanjani: Agree-- I think it's likely users won't know/remember/realize the manifest is (still) there.

internal/pkg/cli/job_init.go
```diff
@@ -158,10 +160,24 @@ func (o *initJobOpts) Validate() error {
...
 	}
-	if err := o.askJobName(); err != nil {
+	localMft, err := o.mftReader.ReadWorkloadManifest(o.name)
+	if err == nil {
+		jobType, err := localMft.WorkloadType()
+		if err != nil {
+			return fmt.Errorf(`read "type" field for job %s from local manifest: %w`, o.name, err)
+		}
+		o.wkldType = jobType
+		log.Infof("Local manifest file for job %s is found; Copilot will create from the existing manifest.", o.name)
```
efekarakus: Maybe this instead? 💭 
```suggestion
		log.Infof("Manifest file for job %s already exists. Skipping configuration.\n", o.name)
```
with "Copilot will create from the existing manifest" it's not exactly clear what Copilot will create

efekarakus (approved):
Small request to update the log messages and then :shipit:

huanjani (approved):
Thanks for doing this-- it makes a much smoother experience for users! Thanks, too, for improving names, etc.

PR #2970, opened 2021-10-31
fix: Build images for each environments

### LINKED ISSUE #2927: ECR image tags should have different value for each environments.
Docker image tags for each environments are currently the same value. (shrinked `$CODEBUILD_BUILD_ID`)
https://github.com/aws/copilot-cli/blob/fdf78af706272c2113ba593db7e7bd983dba2622/internal/pkg/template/templates/cicd/buildspec.yml#L38-L47
Even if you overwrite some variables in `environments` of `manifest.yml`, it will be updated with the value of the last `environment` processed.
So, `tag` value should be specified for each `environment` and `$CODEBUILD_BUILD_ID`. (While taking into account the character limit of image tag)

I can fix it for my `buildspec.yml`, but the template should be fixed.

mather:
Update `buildspec.yml` template for multi environment project.
- For each environments
    - Create tag 
    - Merge `.environments` definitions in `manifest.yml`
    - Build for merged `image` definition.
- Refinements for `jq`
    - Use `jq -r` to remove double quote from output. 
    - Alternative operator. (ref. https://stedolan.github.io/jq/manual/v1.6/#Alternativeoperator:// ) 
    - Merge object by `*`. (ref. https://stedolan.github.io/jq/manual/v1.6/#Multiplication,division,modulo:*,/,and% )

Fixes #2927 


internal/pkg/template/templates/cicd/buildspec.yml
```diff
@@ -22,7 +22,7 @@ phases:
       - export COLOR="false"
       # First, upgrade the cloudformation stack of every environment in the pipeline.
       - pipeline=$(cat $CODEBUILD_SRC_DIR/copilot/pipeline.yml | ruby -ryaml -rjson -e 'puts JSON.pretty_generate(YAML.load(ARGF))')
-      - pl_envs=$(echo $pipeline | jq '.stages[].name' | sed 's/"//g')
+      - pl_envs=$(echo $pipeline | jq -r '.stages[].name')
```
iamhopaul123, line 25: Oh this is slick!

internal/pkg/template/templates/cicd/buildspec.yml
```diff
@@ -85,49 +85,51 @@ phases:
...
-            repo=$(cat $CODEBUILD_SRC_DIR/infrastructure/$workload-$env.params.json | jq '.Parameters.ContainerImage' | sed 's/"//g');
+            tag=$(sed 's/:/-/g' <<<"${CODEBUILD_BUILD_ID}-${env}" | rev | cut -c 1-128 | rev)
+            image=$(echo $manifest | jq ".image * (.environments?.$env?.image? + {})")
+            image_location=$(echo $image | jq '.location')
+            if [ ! "$image_location" = null ]; then
+              echo "skipping image building because location is provided as $image_location";
+              continue
+            fi
+            base_dockerfile=$(echo $image | jq '.build')
+            build_dockerfile=$(echo $image | jq -r '.build?.dockerfile? // ""')
```
iamhopaul123, line 98: :+1:

internal/pkg/template/templates/cicd/buildspec.yml
```diff
@@ -85,49 +85,51 @@ phases:
...
-          echo "Relative Dockerfile path: $df_rel_path"
-          echo "Docker build context: $df_dir_path"
-          echo "Docker build args: $build_args"
-          echo "Running command: docker build -t $workload:$tag $build_args-f $df_path $df_dir_path";
-          docker build -t $workload:$tag $build_args-f $df_path $df_dir_path;
-          image_id=$(docker images -q $workload:$tag);
           for env in $pl_envs; do
-            repo=$(cat $CODEBUILD_SRC_DIR/infrastructure/$workload-$env.params.json | jq '.Parameters.ContainerImage' | sed 's/"//g');
+            tag=$(sed 's/:/-/g' <<<"${CODEBUILD_BUILD_ID}-${env}" | rev | cut -c 1-128 | rev)
+            image=$(echo $manifest | jq ".image * (.environments?.$env?.image? + {})")
```
iamhopaul123: This looks cool! Though I'm not very sure if it has the same behavior as our environment override (we are not doing a very good job in this buildspec as well 🙇 ). Could you help me to confirm if it works for the following scenarios?

- we use every field that is set in  `env.image` to update or insert the corresponding field in `image`
-  if `image.build` and `env.image.location` is set at the same time, we'll use `env.image.location`. Vise versa

Please let me know if anything doesn't make sense to you 🙏

iamhopaul123 (commented):
Thank you so much for contributing to this! Sorry I wasn't able to get to this issue last week. Also for semantic check could you edit the first commit message to make sure it follows the conventional commit as well (aka the same one as your title) 🙏

internal/pkg/template/templates/cicd/buildspec.yml
Lou1415926: This is awesome. 

The usage of `*` is super neat! However we might need to include an extra logic to address @iamhopaul123 's second concern. To add on top of iamhopaul123's comment -  consider this manifest:
```yaml
image:
   location: <an URL>

environments:
   test:
      image:
         // We want to build an image instead of using an existing one for the test environment
         build: 
             dockerfile: ./Dockerfile
```
After the `*` operator, which should be performing an upsert (right?), the output of `image=$(echo $manifest | jq ".image * (.environments?.$env?.image? + {})")` would be a struct that's equivalent to
```yaml
image:
   build: 
        dockerfile: ./Dockerfile
   image:
         location: <an URL>      
```
Next thing we do will be skip building the image because `image.location` is not null; however, the expected behavior should be using `image.build` to build and push a new image.

Therefore, we need to include an extra logic here, that unset `image.build` if `image.location` is specified in the environment override section, and vice versa.

Lou1415926 (commented):
Thank you very much for the awesome fix!

internal/pkg/template/templates/cicd/buildspec.yml
mather: @iamhopaul123 Thanks for reply.

> we use every field that is set in `env.image` to update or insert the corresponding field in `image`

Right. Here is a sample:
```
$ echo '{"sample": {"text": "bar", "aaa": "bbb"}, "another": {"sample": {"aaa": "BBB", "ccc": "ddd"}}}' | jq '.sample * .another.sample'
{
  "text": "bar",
  "aaa": "BBB",
  "ccc": "ddd"
}
```

> if `image.build` and `env.image.location` is set at the same time, we'll use `env.image.location`. Vise versa

Yes. In this case, docker image build will be skipped.

@Lou1415926 It will be merged as following, I think.

```yaml
image:
   build: 
        dockerfile: ./Dockerfile
   location: <an URL>
```

> Therefore, we need to include an extra logic here, that unset `image.build` if `image.location` is specified in the environment override section, and vice versa.

I see. Are there some documentation or implementation about this logic?

mather: I've thought about all the patterns, is this the behavior you are expecting?

- `image.location` * `env.image.location` -> Just merge (overwrite)
- `image.location` * `env.image.build` -> Unset `image.location` and use `image.build`
- `image.build` * `env.image.location` -> Just merge (`image.build` will be ignored)
- `image.build` * `env.image.build` -> Just merge ?

Or should I take other parameters (`credentials`, `labels`, `depends_on`) into consideration?

Lou1415926: Thanks for following up! Yes it is documented [here](https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#image-location) that says _"`image.location`... Mutually exclusive with image.build"_. We should add the same note for `image.build`'s doc as well - our bad! This logic is implemented in the internal code but unfortunately we can't easily reuse that code for buildspec's image building  : (

The other parameters are fine - only `image.location` * `image.build` needs the special treatment. Your summary sounds good to me!

mather: Thanks for reply!
I updated merge logic to ignore `.image.location` if `.environments.<env>.image` exists.

internal/pkg/template/templates/cicd/buildspec.yml
```diff
@@ -51,9 +51,9 @@ phases:
       # The tag is the build ID but we replaced the colon ':' with a dash '-'.
       # We truncate the tag (from the front) to 128 characters, the limit for Docker tags
       # (https://docs.docker.com/engine/reference/commandline/tag/)
-      - tag=$(sed 's/:/-/g' <<<"$CODEBUILD_BUILD_ID" | rev | cut -c 1-128 | rev)
       - >
         for env in $pl_envs; do
+          tag=$(sed 's/:/-/g' <<<"${CODEBUILD_BUILD_ID}-${env}" | rev | cut -c 1-128 | rev)
```
surajbarkale, line 56: I have been using `${CODEBUILD_BUILD_ID##*:}` to simply remove the leading part of the build id.

iamhopaul123 (approved):
This looks awesome! Thank you!

Lou1415926 (approved):
Thank you very much for your contribution!

The "do-not-merge" label is added in case you'd like to address [this last bit of feedback](https://github.com/aws/copilot-cli/pull/2970#discussion_r741598331) (otherwise the PR will just get automatically merged once two approvals are given). Let us know if you think the PR is done - so that we know when to remove the label. Thanks!

internal/pkg/template/templates/cicd/buildspec.yml
mather: I adopted this. thanks.

mather:
@Lou1415926 
> The "do-not-merge" label is added in case you'd like to address [this last bit of feedback](https://github.com/aws/copilot-cli/pull/2970#discussion_r741598331) (otherwise the PR will just get automatically merged once two approvals are given). Let us know if you think the PR is done - so that we know when to remove the label. Thanks!

I’m ready to merge. Thanks.

Lou1415926:
Hello there!! Just a quick update on `buildspec.yml` - 

We plan to update the `buildspec.yml` such that:
1. We use the same image tag if there is no `image` environment override;
2. Use environment-specific image tags otherwise (which is your changes in this PR!)

We will update as described above ⬆️  next week. 

Since it is the weekend tomorrow - which means we have at least a two-day gap until we make the said update - we are temporarily reverting the part of changes related to tags via #2994. This is so that the default `buildspec.yml` that we provide maintains existing behavior - until we send out the above mentioned update that incorporates your changes.

We are also considering eventually moving the image part of logic to go code, instead of in the lengthy shell script. So stay tuned if you are interested! 

Again thank you so much for your contribution!

PR #2977, opened 2021-11-02
chore(test): add a couple template integ tests for Windows

huanjani:


efekarakus (commented):
Nice! 

Should we also add a test similar to this: https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/deploy/cloudformation/stack/lb_web_service_integration_test.go

where we can have windows settings in the manifest -> and compare the output to a cfn template

efekarakus (approved):
:yay-fox:

Lou1415926 (approved):
LGTM!

Issue #2968, opened 2021-10-29
[Proposal] Customizable addons parameters


efekarakus:
## Problem

Today, [`addons/` templates](https://aws.github.io/copilot-cli/docs/developing/additional-aws-resources/) only accept the following 3 parameters:
```yaml
Parameters:
  App:
    Type: String
    Description: Your application's name.
  Env:
    Type: String
    Description: The environment name your service, job, or workflow is being deployed to.
  Name:
    Type: String
    Description: The name of the service, job, or workflow being deployed.
```
These parameters are sufficient to import resources from the environment stack ([example](https://github.com/aws/copilot-cli/blob/6326103e845e555d953e08c9a8a3a6d1fa8c651a/internal/pkg/template/templates/addons/aurora/cf.yml#L56-L57)).  
However, some addons resources require information from the service stack to be able to create them. Some sample usecases are:
*  Defining an `AWS::ApiGatewayV2::Api` addon template requires the [`DiscoveryService`](https://github.com/aws/copilot-cli/blob/6326103e845e555d953e08c9a8a3a6d1fa8c651a/internal/pkg/template/templates/workloads/partials/cf/servicediscovery.yml#L1) resource's ARN.
* Defining an addon template for dashboards, metrics and alarms require the [`Service`](https://github.com/aws/copilot-cli/blob/6326103e845e555d953e08c9a8a3a6d1fa8c651a/internal/pkg/template/templates/workloads/services/lb-web/cf.yml#L118) name.
* Connecting the [`EventsQueue`](https://github.com/aws/copilot-cli/blob/6326103e845e555d953e08c9a8a3a6d1fa8c651a/internal/pkg/template/templates/workloads/partials/cf/subscribe.yml#L61) of a Worker Service to an EventBus event rule requires the QueueURL.

This design issue proposes a new mechanism to add additional parameters to the addons template.

### Parameters file
We are proposing adding a reserved file: `addons.parameters.yml` to specify values for addons stack parameters:

```
.
└── addons/
    ├── template.yml
    └── addons.parameters.yml # 🆕 reserved file.
```
Where users can provide parameter values from the parent stack:

```yaml
Parameters:
  EventsQueue: !Ref EventsQueue
  DiscoveryServiceArn: !GetAtt DiscoveryService.Arn
```

When we detect an `addons.parameters.yml` file, we augment the parameters passed in the service stack with the values the user provides. 
The values for App, Env, and Name are still passed dynamically from the CLI.

[...]

PR #2978, opened 2021-11-02
chore(addons): add `Parameters` method to parse addons.parameters.yml files

efekarakus:
Related #2968


internal/pkg/addon/addons.go
```diff
@@ -86,17 +93,98 @@ func (a *Addons) Template() (string, error) {
...
+			WlName:    a.wlName,
+			ParentErr: err,
+		}
+	}
+	paramFiles := filterFiles(fnames, paramsMatcher)
+	if len(paramFiles) == 0 {
+		return "", nil
+	}
+	if len(paramFiles) > 1 {
+		return "", fmt.Errorf("defining %s is not allowed under %s addons", english.WordSeries(parameterFileNames, "and"), a.wlName)
```
huanjani: "defining multiple"?

huanjani:
Q: Do we need to worry about/validate against the same parameter being defined more than once and other issues like improper yaml syntax? 🤔

efekarakus:
> Q: Do we need to worry about/validate against the same parameter being defined more than once and other issues like improper yaml syntax? 🤔

Good question! For validation, I've limited only to ensuring our reserved parameters are not redefined again: `App`, `Env`, `Name`. Otherwise, I'll just assume what the user wrote is valid and have CloudFormation yell at them when trying to update the stack.

internal/pkg/addon/addons.go
efekarakus: Done!

internal/pkg/addon/addons.go
```diff
@@ -18,6 +20,11 @@ const (
 	StackName = "AddonsStack"
 )
 
+var (
+	yamlExtensions     = []string{".yaml", ".yml"}
+	parameterFileNames = []string{"addons.parameters.yml", "addons.parameters.yaml"}
```
iamhopaul123: Should the name just be "addons.parameters" and add each `yamlExtensions`?

internal/pkg/addon/addons.go
```diff
@@ -86,17 +93,98 @@ func (a *Addons) Template() (string, error) {
...
+			WlName:    a.wlName,
+			ParentErr: err,
+		}
+	}
+	paramFiles := filterFiles(fnames, paramsMatcher)
+	if len(paramFiles) == 0 {
+		return "", nil
+	}
+	if len(paramFiles) > 1 {
+		return "", fmt.Errorf("defining multiple %s is not allowed under %s addons", english.WordSeries(parameterFileNames, "and"), a.wlName)
```
iamhopaul123: Maybe
```suggestion
		return "", fmt.Errorf("defining multiple %s is not allowed under %s addons/", english.WordSeries(parameterFileNames, "and"), a.wlName)
```
Same for the following

internal/pkg/addon/addons.go
```diff
@@ -86,17 +93,98 @@ func (a *Addons) Template() (string, error) {
...
+	}
+	raw, err := a.ws.ReadAddon(a.wlName, paramFiles[0])
```
iamhopaul123: Maybe `paramFile := paramFiles[0]`?

internal/pkg/addon/addons.go
```diff
@@ -86,17 +93,98 @@ func (a *Addons) Template() (string, error) {
...
+		Env  yaml.Node `yaml:"Env"`
+		Name yaml.Node `yaml:"Name"`
+	}{}
+	if err := params.Decode(&content); err != nil {
+		return fmt.Errorf("decode content of parameters file %s under %s addons", fname, a.wlName)
+	}
+
+	for _, field := range []yaml.Node{content.App, content.Env, content.Name} {
+		if !field.IsZero() {
+			return fmt.Errorf("reserved parameters 'App', 'Env', and 'Name' cannot be declared in %s under %s addons", fname, a.wlName)
```
iamhopaul123: Just a callout but when in future we add more reserved parameters it could be a breaking change for some users.

internal/pkg/addon/addons.go
```diff
@@ -58,15 +65,15 @@ func (a *Addons) Template() (string, error) {
 		}
 	}
 
-	yamlFiles := filterYAMLfiles(fnames)
-	if len(yamlFiles) == 0 {
+	templateFiles := filterFiles(fnames, yamlMatcher, nonParamsMatcher)
```
iamhopaul123, line 75: nice!

internal/pkg/addon/addons.go
huanjani: Sorry sorry sorry 🙈  totally my fault. Can we change this to "defining both"?

huanjani: Oh, just seeing PH's fix above. That suggested change would resolve this!

internal/pkg/addon/addons.go
efekarakus: 👍 done!

internal/pkg/addon/addons.go
efekarakus: Done!

efekarakus: I just reverted to the original in case the length of `parameterFileNames` changes

internal/pkg/addon/addons.go
efekarakus: Done!

internal/pkg/addon/addons.go
efekarakus: We should be safe 💭 because we can't add more reserved parameters otherwise it will break existing users addon templates

huanjani (approved):
LGTM!

iamhopaul123 (approved):

PR #2979, opened 2021-11-02
chore(manifest): add network config for rdws

iamhopaul123:


internal/pkg/manifest/validate_test.go
```diff
@@ -940,15 +961,15 @@ func TestPlatformArgsOrString_Validate(t *testing.T) {
...
 		"error if only half of platform string is specified": {
-			in: PlatformArgsOrString{PlatformString: (*PlatformString)(aws.String("linux"))},
+			in:     PlatformArgsOrString{PlatformString: (*PlatformString)(aws.String("linux"))},
 			wanted: fmt.Errorf("platform 'linux' must be in the format [OS]/[Arch]"),
 		},
 		"error if only osfamily is specified": {
-		in: PlatformArgsOrString{
-			PlatformArgs: PlatformArgs{
-				OSFamily: aws.String("linux"),
+			in: PlatformArgsOrString{
```
huanjani, line 968: Thanks for fixing my indentation!! 😳

huanjani (approved):
LGTM!

internal/pkg/manifest/validate.go
```diff
@@ -924,6 +951,17 @@ func (v vpcConfig) Validate() error {
...
+// Validate returns nil if rdwsPlacement is configured correctly.
+func (p rdwsPlacement) Validate() error {
+	if err := (Placement)(p).Validate(); err != nil {
+		return err
+	}
+	if string(p) == string(PublicSubnetPlacement) {
+		return fmt.Errorf(`public is not supported for %s`, RequestDrivenWebServiceType)
+	}
+	return nil
+}
```
efekarakus: what do you think about flipping this conditional instead? 
```go
if string(p) == string(PrivateSubnetPlacement) {
   return nil
}
return fmt.Errorf("placement: '%s' is not supported for %s", string(p), RequestDrivenWebServiceType)
```
This way in case we add more placement options we won't have to update this method 🤔

efekarakus (approved):
Looks good, just a suggestion 👇

internal/pkg/manifest/validate_test.go
iamhopaul123: lol it's done automatically by my IDE

efekarakus (approved):

PR #2982, opened 2021-11-03
chore: fix e2e test by skipping `WorkspaceNotFound` error in `svc init` and `job init`

Lou1415926:
The e2e test failed because of #2966. When `copilot` commands is run for the first time in a directory, the workspace, indicated by the existence of `copilot/` folder,  may not exist. This error should be skipped, not surfaced, as well.

### DIFF
--- a/internal/pkg/cli/svc_init_test.go
+++ b/internal/pkg/cli/svc_init_test.go
@@ -474,7 +474,7 @@ func TestSvcInitOpts_Ask(t *testing.T) {
 			inDockerfilePath: "",
 
 			setupMocks: func(m initSvcMocks) {
-				m.mockMftReader.EXPECT().ReadWorkloadManifest(wantedSvcName).Return(nil, &workspace.ErrWorkspaceNotFound{})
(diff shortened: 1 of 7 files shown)


iamhopaul123 (approved):
🙇‍♂️

internal/pkg/cli/svc_init_test.go
```diff
@@ -474,7 +474,7 @@ func TestSvcInitOpts_Ask(t *testing.T) {
 			inDockerfilePath: "",
 
 			setupMocks: func(m initSvcMocks) {
-				m.mockMftReader.EXPECT().ReadWorkloadManifest(wantedSvcName).Return(nil, &workspace.ErrFileNotExists{FileName: wantedSvcName})
+				m.mockMftReader.EXPECT().ReadWorkloadManifest(wantedSvcName).Return(nil, &workspace.ErrWorkspaceNotFound{})
```
efekarakus: maybe we should keep one test case for ErrFileNotExists and update only one file for ErrWorkspaceNotFound?

Lou1415926: I updated the code, so that `svc_init_test` (here) return the error `ErrFileNotExists`, and `job_init_test` returns `ErrWorkspaceNotFound` - let me know if this is not what you meant!

efekarakus (approved):

PR #2983, opened 2021-11-03
chore(template): add `AddonsExtraParams` field for user defined params

efekarakus:
Related #2968

Follow-up PR will glue #2978 and this PR together.


internal/pkg/template/templates/workloads/partials/cf/addons.yml
```diff
@@ -11,5 +11,8 @@ AddonsStack:
       App: !Ref AppName
       Env: !Ref EnvName
       Name: !Ref WorkloadName
+      {{- if .AddonsExtraParams }}
+{{ .AddonsExtraParams | indent 6 }}
```
iamhopaul123, line 15: do we not set the indent here 🤔  https://github.com/aws/copilot-cli/pull/2978/files#diff-cc1f6de0ed3926d7713d70861fd635ec7ff5fccaa08d0d7d308411990761bf7cR144

internal/pkg/template/workload.go
```diff
@@ -356,6 +356,7 @@ type WorkloadOpts struct {
 	Aliases                  []string
 	Tags                     map[string]string        // Used by App Runner workloads to tag App Runner service resources
 	NestedStack              *WorkloadNestedStackOpts // Outputs from nested stacks such as the addons stack.
+	AddonsExtraParams        string                   // Additional user defined Parameters for the addons stack.
```
Lou1415926, line 359: Should this be `*string`?

internal/pkg/template/templates/workloads/partials/cf/addons.yml
efekarakus: They're different indentations 😄 

the one you linked allow us to marshal this way:
```yaml
Service:
   !Ref Service # has 2 spaces instead of yaml.Marshal default of 4
```

whereas this indent takes the whole block above and adds 6 spaces before it:
```yaml
      Service: # same as above but completely shifted by 6 spaces
        !Ref Service
```

internal/pkg/template/workload.go
efekarakus: I don't think it matters in this scenario 🤔 . It makes my life a bit easier if it's a string because the `Parameters` method's signature returns an empty string instead of nil if there is no extra params defined:
https://github.com/aws/copilot-cli/pull/2978/files#diff-cc1f6de0ed3926d7713d70861fd635ec7ff5fccaa08d0d7d308411990761bf7cR110

internal/pkg/template/templates/workloads/partials/cf/addons.yml
iamhopaul123: right right

iamhopaul123 (approved):

Lou1415926 (approved):

PR #2985, opened 2021-11-03
chore: remove custom resource that enables long ARN

Lou1415926:
This custom action is no longer needed now that long ARN is enabled by default starting April, 2021 (https://docs.aws.amazon.com/AmazonECS/latest/userguide/ecs-account-settings.html#ecs-resource-arn-timeline)


efekarakus (commented):
yay 🎉 all the red!

Can we also delete `enable-long-arn.js` and the test file?

iamhopaul123 (approved):

efekarakus (approved):

Issue #2660, opened 2021-07-26
Feature Request: Add PublicRouteTable to Outputs of App Cloudformation stack


TillaTheHun0:
Hey Copilot team. I have a feature request to add the `PublicRouteTable` resource, created as part of the root app Cloudformation stack, to the `Outputs` of that stack.

My use case is that I am wanting to add an s3 `VpcEndpoint` to my Copilot VPC as an `addon`. Though, in order to accomplish this, I must provide [a list of Route table ids](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpcendpoint.html#cfn-ec2-vpcendpoint-routetableids) to the `VpcEndpoint` resource in Cloudformation.

If the `PublicRouteTable` were added as an output of the app stack, I could use `ImportValue` to pull in the resource into an `addon` stack, similar to what we do for `{ 'Fn::ImportValue': !Sub '${App}-${Env}-EnvironmentSecurityGroup' }`

Perhaps there is another way to accomplish this? Another way to accomplish this would be for Copilot to automatically add a `VpcEndpoint` for common services like `s3`, when the VPC is created, but that is a separate feature request that I suspect would require more consideration. Outputting an already created resource would get me what I need. 😄 

**TL;DR**: Please add the `PublicRouteTable` resource to the `Outputs` of the app root Cloudformation stack.

Thanks for the consideration!

huanjani:
@TillaTheHun0, thanks for this suggestion! It is something we had already considered in the past, and now we have confirmation that it would indeed be useful to users!

TillaTheHun0:
Heya. Another use case for this:

In order to set up VPC peering, one has to add a routing rule to the source VPC routing table, pointing the Peer VPC CIDR block to the VPC peering connection. Adding the routing table id to the Copilot environment CF stack would make this possible without have to reference the static routing table id. We could instead import the value from the Copilot CF stack.

I hope that makes sense!

gautam-nutalapati:
Curious design related question for this feature,
Use case I am thinking of:
- SVC is deploying S3 VPCEndpoint. 
- It is deployed in 2 ENVs A and B.
- Env B is importing Env A VPC resources.
- If Env B is not exporting public RT (As that Env stack did not create VPC resources), SVC deployment in ENV B fails because VPCEndpoint add-on will not be able to reference public routing table output in Env B stack.

Question:
[...]

PR #2987, opened 2021-11-03
feat: export route table from environment stack

Lou1415926:
This PR addresses #2660. If an environment uses Copilot-default VPC, then the public route table ID is exported; as a result, user can import the route table ID in their addon stack by writing `'Fn::ImportValue': !Sub '${App}-${Env}-PublicRouteTableID`.

Note that the route table is only exported if Copilot-default VPC is used. If the user chooses to import their own VPC, Copilot doesn't export the route table.


huanjani (approved):
Thank you!

efekarakus:
Do we need to bump the version number of the environment template in order for these changes to be picked up?

gautam-nutalapati:
Probably we should also add check in [integration test](https://github.com/aws/copilot-cli/blob/09ae32389f000258666f7055d8a5afa5581874cc/internal/pkg/deploy/cloudformation/cloudformation_integration_test.go#L525) to validate this export.

internal/pkg/deploy/env.go
```diff
@@ -13,7 +13,7 @@ const (
 	// LegacyEnvTemplateVersion is the version associated with the environment template before we started versioning.
 	LegacyEnvTemplateVersion = "v0.0.0"
 	// LatestEnvTemplateVersion is the latest version number available for environment templates.
-	LatestEnvTemplateVersion = "v1.6.1"
+	LatestEnvTemplateVersion = "v1.6.2"
```
efekarakus: nit: Technically it's a feature and not a bug fix :P 
```suggestion
	LatestEnvTemplateVersion = "v1.7.0"
```

efekarakus (approved):

PR #2988, opened 2021-11-03
chore: create natgateway if rdws svc network is not empty

iamhopaul123:


internal/pkg/template/templates/workloads/services/rd-web/cf.yml
```diff
@@ -190,3 +190,5 @@ Resources:
                   - "*"
 {{- end }}
 {{include "publish" . | indent 2}}
+
+{{include "env-controller" . | indent 2}}
```
efekarakus: can we include this only if:
```
{{if eq .Network.SubnetsType "private"}}
```

internal/pkg/template/workload.go
```diff
@@ -531,8 +502,13 @@ func envControllerParameters(o WorkloadOpts) []string {
...
 		parameters = append(parameters, []string{"ALBWorkloads,", "Aliases,"}...) // YAML needs the comma separator; resolved in EnvContr.
 	}
+	if o.WorkloadType == "Request-Driven Web Service" {
+		if o.Network.SubnetsType == PrivateSubnetsPlacement {
+			parameters = append(parameters, "NATWorkloads,")
+		}
+	}
 	if o.Network.SubnetsType == PrivateSubnetsPlacement {
-		parameters = append(parameters, "NATWorkloads,") // YAML needs the comma separator; resolved in EnvContr.
+		parameters = append(parameters, "NATWorkloads,")
```
efekarakus, line 506: Would this result in "NATWorkloads,"  to be appended twice? maybe we don't need `if o.WorkloadType == "Request-Driven Web Service"`

efekarakus (approved):

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -494,6 +494,20 @@ func convertNetworkConfig(network manifest.NetworkConfig) *template.NetworkOpts
 	return opts
 }
 
+func convertRDWSNetworkConfig(network manifest.RequestDrivenWebServiceNetworkConfig) template.NetworkOpts {
+	if network.IsEmpty() {
+		return template.NetworkOpts{}
+	}
+	opts := template.NetworkOpts{}
```
```suggestion
      opts := template.NetworkOpts{}
	if network.IsEmpty() {
		return opts
	}
```

internal/pkg/manifest/rd_web_svc.go
```diff
@@ -44,10 +44,11 @@ func (c *RequestDrivenWebServiceNetworkConfig) IsEmpty() bool {
 	return c.VPC.isEmpty()
 }
 
-type rdwsPlacement Placement
+// RequestDrivenWebServicePlacement represents where to place tasks for request driven web service.
```
Lou1415926: nit:
```suggestion
// RequestDrivenWebServicePlacement represents where to place tasks for a Request-Driven Web Service.
```

internal/pkg/template/workload.go
```diff
@@ -392,67 +385,45 @@ type WorkloadOpts struct {
...
 	AppDNSName           *string
+
+	// Additional options for worker service templates.
+	Subscribe *SubscribeOpts
 }
 
 // ParseLoadBalancedWebService parses a load balanced web service's CloudFormation template
 // with the specified data object and returns its content.
 func (t *Template) ParseLoadBalancedWebService(data WorkloadOpts) (*Content, error) {
-	if data.Network == nil {
```
Lou1415926, line 423: Quick question b/c I'm likely missing something: what do we do with the defaults now?

iamhopaul123: This is previously a duplicated behavior as the default has already been handled in the converter.

internal/pkg/manifest/validate_test.go
```diff
@@ -1727,13 +1727,13 @@ func TestPlacement_Validate(t *testing.T) {
...
+func TestRequestDrivenWebServicePlacement_Validate(t *testing.T) {
 	testCases := map[string]struct {
-		in     *rdwsPlacement
+		in     *RequestDrivenWebServicePlacement
 		wanted error
 	}{
 		"should return an error if placement is public": {
-			in:     (*rdwsPlacement)(aws.String("public")),
+			in:     (*RequestDrivenWebServicePlacement)(aws.String("public")),
 			wanted: errors.New(`placement "public" is not supported for Request-Driven Web Service`),
```
huanjani, line 1737: Q: (I wonder if the error message was introduced in a different PR? Sorry if I missed it)...Should we explain a little bit more here so users know what the options are and what they mean? Or should we just let them refer to the docs? 🤔

Lou1415926 (approved):

internal/pkg/manifest/validate_test.go
iamhopaul123: Yeah it's in the last PR. The doc hasn't been out yet. I think they should refer to the doc otherwise they won't know how to set this field as we don't write it to the manifest by default.

Lou1415926: I think it also makes sense to say it in the initial manifest that Copilot produces

PR #2994, opened 2021-11-05
chore: quick follow-up to revert partial changes to buildspec

Lou1415926:
This PR is a quick follow up to #2970, reverting some changes. 

The previous PR assigns different tags to each environment image. However, if the images are the same across all environments, the tags shouldn't be different.

We will send out a follow-up PR to update the `buildspec` such that we assign the same tag / environment-specific tags to images depending on whether the images are the same across environments / environment-specific.

For now, we temporarily revert the change so that it maintains existing behavior until we send out an update.

### DIFF
--- a/internal/pkg/template/templates/cicd/buildspec.yml
+++ b/internal/pkg/template/templates/cicd/buildspec.yml
@@ -51,9 +51,9 @@ phases:
       # The tag is the build ID but we replaced the colon ':' with a dash '-'.
       # We truncate the tag (from the front) to 128 characters, the limit for Docker tags
       # (https://docs.docker.com/engine/reference/commandline/tag/)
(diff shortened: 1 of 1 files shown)


iamhopaul123 (commented):
Can we move it out of the env for loop now that it is not related to env anymore?

iamhopaul123 (approved):

efekarakus (approved):

mather:
> The previous PR assigns different tags to each environment image. However, if the images are the same across all environments, the tags shouldn't be different.

No, I don't think so.
You should use different tags to represent image id for each environments, even if image ids are same.
This PR removes my update about else-case of `if the images are the same across all environments`.
Too sad.

mather:
If docker images have same digest, single image will be pushed and has tags for each environments.
No need to concern if images are different or not in build script.
スクリーンショット 2021-11-06 22 04 36

PR #2996, opened 2021-11-07
fix: Implement reverted feature again in #2994

### LINKED ISSUE #2927: ECR image tags should have different value for each environments.

mather:
- Generate docker image tags for each environments, and workloads.

In #2994,

> For now, we temporarily revert the change so that it maintains existing behavior until we send out an update.

My PR (#2970) is not a feature update, but bugfix.
It does not affect the existing behavior as long as the digests match, just add tags for same digests and push image once (will be skipped for existing digest).
スクリーンショット 2021-11-06 22 04 36

The problem arises because using the same tag when the digest does not match changes the image that the tag represents.
スクリーンショット 2021-11-08 1 41 07

Fixes #2927 
Addresses #2970, #2994 

### DIFF
--- a/internal/pkg/template/templates/cicd/buildspec.yml
+++ b/internal/pkg/template/templates/cicd/buildspec.yml
@@ -51,9 +51,9 @@ phases:
       # The tag is the build ID but we replaced the colon ':' with a dash '-'.
       # We truncate the tag (from the front) to 128 characters, the limit for Docker tags
       # (https://docs.docker.com/engine/reference/commandline/tag/)
-      - tag=$(sed 's/:/-/g' <<<"${CODEBUILD_BUILD_ID##*:}" | rev | cut -c 1-128 | rev)
       - >
         for env in $pl_envs; do
+          tag=$(sed 's/:/-/g' <<<"${CODEBUILD_BUILD_ID##*:}-${env}" | rev | cut -c 1-128 | rev)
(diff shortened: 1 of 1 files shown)


iamhopaul123 (approved):
Hello @mather. Sorry for all the churn about this bug fix. What we previously disagreed with is that we thought the same images need to have both the same tag and the same digest. That's the reason why we reverted the change.

However, after a discussion within our team, we agreed to your justification and decided to move forward with your new PR. Sorry again for the premature decision we made before and thank you for your support and contribution to Copilot ❤️

efekarakus (approved):

mather:
@iamhopaul123 Thank you for your comment. Sorry for the inconvenience due to the lack of sharing of concepts.

PR #3001, opened 2021-11-08
chore(templates): add resource for rdws ⇔ private subnets

efekarakus:

### DIFF
--- a/internal/pkg/template/templates/workloads/partials/cf/vpc-connector.yml
+++ b/internal/pkg/template/templates/workloads/partials/cf/vpc-connector.yml
@@ -1,0 +1,39 @@
+ServiceSecurityGroup:
+  Type: AWS::EC2::SecurityGroup
+  Metadata:
+    'aws:copilot:description': 'A security group for your addons to allow ingress'
+  Properties:
+    GroupDescription: !Join [ '', [ !Ref AppName, '-', !Ref EnvironmentName, !Ref WorkloadName ] ]
+    VpcId:
+      Fn::ImportValue:
+        !Sub '${AppName}-${EnvName}-VpcId'
+    Tags:
+      - Key: Name
+        Value: !Sub 'copilot-${AppName}-${EnvironmentName}-${WorkloadName}'
+
+VpcConnector:
+  Type: AWS::AppRunner::VpcConnector
+  DependsOn: EnvControllerAction # Ensure NAT gateways are created before connecting to a VPC.
+  Metadata:
+    'aws:copilot:description': 'A VPC connector to link your service to private subnets in your environment'
+  Properties:
+    Subnets:
+      Fn::Split:
+        - ","
+        - Fn::ImportValue:
+            !Sub '${AppName}-${EnvName}-PrivateSubnets'
+    SecurityGroups:
+    - !Ref ServiceSecurityGroup
+    Tags:
+      - Key: copilot-application
+        Value: !Ref AppName
+      - Key: copilot-environment
+        Value: !Ref EnvName
+      - Key: copilot-service
+        Value: !Ref WorkloadName
+      {{- if .Tags }}
+      {{- range $name, $value := .Tags }}
+      - Key: {{$name}}
+        Value: {{$value}}
+      {{- end }}
+      {{- end }}
--- a/internal/pkg/template/templates/workloads/partials/cf/vpc-connector.yml
+++ b/internal/pkg/template/templates/workloads/partials/cf/vpc-connector.yml
@@ -1,7 +1,7 @@
 ServiceSecurityGroup:
   Type: AWS::EC2::SecurityGroup
   Metadata:
-    'aws:copilot:description': 'A security group for your addons to allow ingress'
+    'aws:copilot:description': 'A security group for your addons/ resources to allow ingress from your service'
   Properties:
     GroupDescription: !Join [ '', [ !Ref AppName, '-', !Ref EnvironmentName, !Ref WorkloadName ] ]
     VpcId:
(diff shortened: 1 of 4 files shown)


internal/pkg/template/templates/workloads/partials/cf/vpc-connector.yml
```diff
@@ -0,0 +1,39 @@
+ServiceSecurityGroup:
+  Type: AWS::EC2::SecurityGroup
+  Metadata:
+    'aws:copilot:description': 'A security group for your addons to allow ingress'
+  Properties:
+    GroupDescription: !Join [ '', [ !Ref AppName, '-', !Ref EnvironmentName, !Ref WorkloadName ] ]
```
Lou1415926, line 6: Should we write the description such as 'A security group for your RDS Aurora addons to allow ingress from VPC private subnets'

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
```diff
@@ -193,4 +199,8 @@ Resources:
 
 {{- if ne .EnvControllerLambda ""}}
 {{include "env-controller" . | indent 2}}
-{{- end}}
\ No newline at end of file
+{{- end}}
+
+{{- if (and (contains .FeatureFlags "VPCConnector") (eq .Network.SubnetsType "PrivateSubnets")) }}
```
Lou1415926, line 204: > `(eq .Network.SubnetsType "PrivateSubnets")` 

Is this meant to be an extra safeguard in addition to manifest validation against people that wants to put "public" for vpc?

internal/pkg/template/templates/workloads/partials/cf/vpc-connector.yml
```diff
@@ -0,0 +1,39 @@
...
+    VpcId:
+      Fn::ImportValue:
+        !Sub '${AppName}-${EnvName}-VpcId'
+    Tags:
+      - Key: Name
+        Value: !Sub 'copilot-${AppName}-${EnvironmentName}-${WorkloadName}'
```
iamhopaul123, line 12: Do they allow two SG with the same tag name? I'm thinking of a scenario where the previous stack failed to delete this SG and when you are trying to create the stack again, it could possibly fail if they require such kind of naming uniqueness.

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
efekarakus: The `FeatureFlags` expression is indeed meant to be an additional safeguard so that if anybody enters `placement: 'private'` today we don't create these resources 😅 

The `Network` expression is mandatory and will remain there even after we remove the `FeatureFlags` expression. We want to ensure that these resources are created only while connecting to private subnets

internal/pkg/template/templates/workloads/partials/cf/vpc-connector.yml
```diff
@@ -0,0 +1,39 @@
...
+      - Key: copilot-environment
+        Value: !Ref EnvName
+      - Key: copilot-service
+        Value: !Ref WorkloadName
+      {{- if .Tags }}
+      {{- range $name, $value := .Tags }}
+      - Key: {{$name}}
+        Value: {{$value}}
+      {{- end }}
+      {{- end }}
```
iamhopaul123, line 39: Why do we need this 🤔

internal/pkg/template/templates/workloads/partials/cf/vpc-connector.yml
efekarakus: I didn't want to say something specific to RDS because they could be using it also for elasticache clusters, opensearch clusters.

Maybe this?
```
'A security group for your addon resources to allow ingress from your service'
```

iamhopaul123 (commented):
LGTM overall! We probably need integ test for this feature as well

internal/pkg/template/templates/workloads/partials/cf/vpc-connector.yml
efekarakus: updated, let me know what you think!

internal/pkg/template/templates/workloads/partials/cf/vpc-connector.yml
efekarakus: For tags it's okay! I just tried this via the console and it's indeed allowed

internal/pkg/template/templates/workloads/partials/cf/vpc-connector.yml
efekarakus: By default app runner doesn't trickle down stack tags to the individual resources, so we need to redeclare them here.

We let them know and they're looking into fixing that AFAIK.

The `.Tags` is a special field that's available in rdws manifests: https://aws.github.io/copilot-cli/docs/manifest/rd-web-service/ although we are missing docs for it 🤦

efekarakus:
> LGTM overall! We probably need integ test for this feature as well

Yeah 100% ! It's part of the milestones, I'm waiting to remove the feature flag to add it

internal/pkg/template/templates/workloads/partials/cf/vpc-connector.yml
iamhopaul123: Yeah but
```yaml
    Tags:
      - Key: copilot-application
        Value: !Ref AppName
      - Key: copilot-environment
        Value: !Ref EnvName
      - Key: copilot-service
        Value: !Ref WorkloadName
```
should be handled by CFN no?

efekarakus: > By default app runner doesn't trickle down stack tags to the individual resources, so we need to redeclare them here.

nope unfortunately not today !

iamhopaul123: oooh gotcha

iamhopaul123 (approved):

internal/pkg/template/templates/workloads/partials/cf/vpc-connector.yml
iamhopaul123: Maybe add a comment describing the reason why this is an exception?

Lou1415926 (approved):

PR #3004, opened 2021-11-08
docs: update addons page for customizable parameters

### LINKED ISSUE #2968: [Proposal] Customizable addons parameters

efekarakus:
Resolves #2968 

### DIFF
--- a/site/content/docs/developing/additional-aws-resources.en.md
+++ b/site/content/docs/developing/additional-aws-resources.en.md
@@ -144,12 +144,14 @@ Parameters:
 Finally, update your template file to refer to the new parameter:
 ```yaml
 Parameters:
+  # Required parameters by AWS Copilot.
   App:
     Type: String
   Env:
     Type: String
   Name:
     Type: String
+  # Additional parameters defined in addons.parameters.yml
   ServiceName:
     Type: String
 ```
--- a/site/content/docs/developing/additional-aws-resources.en.md
+++ b/site/content/docs/developing/additional-aws-resources.en.md
@@ -38,14 +38,21 @@ When your service gets deployed, Copilot merges all these files into a single AW
 An addon template can be any valid CloudFormation template.   
 However, by default, Copilot will pass the `App`, `Env`, and `Name` [Parameters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html); you can customize your resource properties with [Conditions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/conditions-section-structure.html) or [Mappings](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/mappings-section-structure.html) if you wish to.
 
-Here are several possible ways to access [Resources](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html) from your ECS task:
+### Connecting addon resources to your workloads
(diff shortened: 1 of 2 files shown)


site/content/docs/developing/additional-aws-resources.en.md
```diff
@@ -115,3 +122,34 @@ On your next release, Copilot will include this template as a nested stack under
...
+```
+Finally, update your template file to refer to the new parameter:
+```yaml
+Parameters:
+  App:
+    Type: String
+  Env:
+    Type: String
+  Name:
+    Type: String
```
iamhopaul123, line 153: They don't necessarily need to define these params right? Maybe we can remove them from the example?

iamhopaul123 (approved):

site/content/docs/developing/additional-aws-resources.en.md
Lou1415926: Or we could add a comment that says `App` `Env` and `Name` are required to be there, and `ServiceName` is the additional one that got defined by `addons.parameters.yml`.

Lou1415926 (approved):

site/content/docs/developing/additional-aws-resources.en.md
efekarakus: They need to have these params! Otherwise, CFN deployment will fail 😿 

I'll update with Wanxian's recommendation tho

efekarakus: Done!

PR #3008, opened 2021-11-09
chore: pass on CIDR blocks if NLB is enabled

Lou1415926:
This PR is part of the effort to support NLB for LBWS. 

Preceding: #2954 

### DIFF
--- a/internal/pkg/cli/svc_deploy.go
+++ b/internal/pkg/cli/svc_deploy.go
@@ -384,7 +384,7 @@ func (o *deploySvcOpts) publicCIDRBlocks() ([]string, error) {
 	vpcID := envDescription.EnvironmentVPC.ID
 	subnets, err := o.subnetLister.ListVPCSubnets(vpcID)
 	if err != nil {
-		return nil, fmt.Errorf("list VPC subnets of the environmnent %s: %w", o.envName, err)
+		return nil, fmt.Errorf("list subnets of vpc %s in environment %s: %w", vpcID, o.envName, err)
 	}
 	var cidrBlocks []string
 	for _, subnet := range subnets.Public {
--- a/internal/pkg/cli/svc_deploy.go
+++ b/internal/pkg/cli/svc_deploy.go
@@ -587,7 +587,7 @@ func (o *deploySvcOpts) stackConfiguration(addonsURL string) (cloudformation.Sta
 
 		var opts []stack.LoadBalancedWebServiceOption
 
-		// TODO:
+		// TODO: https://github.com/aws/copilot-cli/issues/2918
 		// 1. Should error out if `nlb.alias` is specified with targetApp.Domain == ""
 		// 2. Should validate `nlb.alias` is specified
 		// 3. ALB block should not be executed if http is disabled
--- a/internal/pkg/cli/svc_deploy.go
+++ b/internal/pkg/cli/svc_deploy.go
@@ -75,7 +75,7 @@ type deploySvcOpts struct {
 	deployWkldVars
 
 	store               store
-	deployStore *deploy.Store
+	deployStore         *deploy.Store
 	ws                  wsSvcDirReader
 	imageBuilderPusher  imageBuilderPusher
 	unmarshal           func([]byte) (manifest.WorkloadManifest, error)
--- a/internal/pkg/cli/svc_deploy.go
+++ b/internal/pkg/cli/svc_deploy.go
@@ -11,6 +11,8 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/aws/copilot-cli/internal/pkg/aws/ec2"
+
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/copilot-cli/internal/pkg/apprunner"
 	"github.com/aws/copilot-cli/internal/pkg/docker/dockerengine"
@@ -89,6 +91,8 @@ type deploySvcOpts struct {
 	endpointGetter      endpointGetter
 	snsTopicGetter      deployedEnvironmentLister
 	identity            identityService
+	subnetLister        vpcSubnetLister
+	envDescriber        envDescriber
 
 	spinner progress
 	sel     wsSelector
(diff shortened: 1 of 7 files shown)


internal/pkg/cli/svc_deploy.go
```diff
@@ -305,6 +309,17 @@ func (o *deploySvcOpts) configureClients() error {
...
+	d, err := describe.NewEnvDescriber(describe.NewEnvDescriberConfig{
+		App:         o.appName,
+		Env:         o.envName,
+		ConfigStore: o.store,
+	})
+	if err != nil {
+		return fmt.Errorf("create describer for environment %s in application %s: %w", o.envName, o.appName, err)
+	}
+	o.envDescriber = d
+	o.subnetLister = ec2.New(defaultSessEnvRegion)
```
efekarakus: Should this be the session from the EnvManagerRole with the environment's region?

The dream state that we're trying to maintain I think:
* `init` commands use the default session (as it's administrative work to create resources)
* `deploy`, `show`, `status`, ... the rest use the env manager role. This way in the future admins can potentially restrict the EnvManagerRole even further and not allow devs on their teams to take certain operations.

I know that today we are not fully there yet, but we should take small steps towards it 🥾

internal/pkg/cli/svc_deploy.go
```diff
@@ -358,6 +373,23 @@ func (o *deploySvcOpts) configureClients() error {
...
 
+func (o *deploySvcOpts) publicCIDRBlocks() ([]string, error) {
+	envDescription, err := o.envDescriber.Describe()
+	if err != nil {
+		return nil, fmt.Errorf("describe environment %s: %w", o.envName, err)
+	}
+	vpcID := envDescription.EnvironmentVPC.ID
+	subnets, err := o.subnetLister.ListVPCSubnets(vpcID)
+	if err != nil {
+		return nil, fmt.Errorf("list VPC subnets of the environmnent %s: %w", o.envName, err)
```
efekarakus: Can we include the vpc id in the error as well? (and small typo)
```suggestion
		return nil, fmt.Errorf("list subnets of vpc %s in environment %s: %w", vpcID, o.envName, err)
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -550,6 +583,11 @@ func (o *deploySvcOpts) stackConfiguration(addonsURL string) (cloudformation.Sta
 		}
 
 		var opts []stack.LoadBalancedWebServiceOption
+
+		// TODO:
```
efekarakus: nit: can we link to the nlb issue here?

internal/pkg/cli/svc_deploy.go
```diff
@@ -550,6 +583,11 @@ func (o *deploySvcOpts) stackConfiguration(addonsURL string) (cloudformation.Sta
...
+		// 1. Should error out if `nlb.alias` is specified with targetApp.Domain == ""
+		// 2. Should validate `nlb.alias` is specified
+		// 3. ALB block should not be executed if http is disabled
```
efekarakus, line 593: oh i don't remember did we decide on NLB and ALB being mutually exclusive?

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -55,6 +55,7 @@ type LoadBalancedWebServiceConfig struct {
 	Network          NetworkConfig             `yaml:"network"`
 	PublishConfig    PublishConfig             `yaml:"publish"`
 	TaskDefOverrides []OverrideRule            `yaml:"taskdef_overrides"`
+	NLBConfig        bool                      // TODO: placeholder field that will be expanded to NLBConfig struct.
```
efekarakus, line 58: nit: let's not include dead code if the field isn't used yet, we can add NLBConfig once the struct is ready

internal/pkg/cli/svc_deploy.go
Lou1415926: No but they should be able to disable http by `http: fault`

efekarakus: oh gotcha gotcha, I misunderstood the comment 😅 thank you!

internal/pkg/cli/svc_deploy.go
Lou1415926: Makes sense! I've modified the `EnvManagerRole` adding a permission. Let me know what you think!

internal/pkg/cli/svc_deploy.go
```diff
@@ -73,6 +75,7 @@ type deploySvcOpts struct {
 	deployWkldVars
 
 	store               store
+	deployStore *deploy.Store
```
iamhopaul123: Just a nit: seems like the IDE doesn't help you to align the code automatically 🤔

efekarakus: the formatting seems to be off here 💭

internal/pkg/deploy/env.go
```diff
@@ -13,7 +13,7 @@ const (
 	// LegacyEnvTemplateVersion is the version associated with the environment template before we started versioning.
 	LegacyEnvTemplateVersion = "v0.0.0"
 	// LatestEnvTemplateVersion is the latest version number available for environment templates.
-	LatestEnvTemplateVersion = "v1.7.0"
+	LatestEnvTemplateVersion = "v1.7.1"
```
iamhopaul123: nice!

internal/pkg/template/templates/environment/partials/environment-manager-role.yml
```diff
@@ -212,7 +212,8 @@ EnvironmentManagerRole:
           Action: [
             "ec2:DescribeSubnets",
             "ec2:DescribeSecurityGroups",
-            "ec2:DescribeNetworkInterfaces"
+            "ec2:DescribeNetworkInterfaces",
+            "ec2:DescribeRouteTables"
```
iamhopaul123, line 216: I don't see why it is required in this PR 🤔

internal/pkg/manifest/lb_web_svc.go
Lou1415926: <strike>I will cherry-pick my manifest changes to this PR</strike>

internal/pkg/deploy/env.go
efekarakus: no need to update it 😎 from the point of view of the customers they are still till on v1.6.2 since v1.7.0 of the template hasn't been released yet

internal/pkg/template/templates/environment/partials/environment-manager-role.yml
efekarakus, line 216: These changes should be fine since it's strictly READ permissions, but we should call it out today while reviewing with AppSec

internal/pkg/deploy/env.go
iamhopaul123: oh right right

internal/pkg/template/templates/environment/partials/environment-manager-role.yml
Lou1415926: Because subnetLister who lists the VPC subnets calls DescribeRouteTables. EnvManagerRole doesn't have the permission to do this.

iamhopaul123: Oh ok gotcha

efekarakus (approved):

iamhopaul123 (approved):

PR #3010, opened 2021-11-09
chore: add `nlb` to manifest

Lou1415926:
This PR enables `nlb` in manifest.

This PR allows:
1. Using NLB and access the service through NLB default URL
2. Using NLB and access the service through NLB default alias -  formatted as <service-name>-nlb.env.app.domain.com - provided that the app is registered with a domain

Next:
- [ ] Enable TLS termination from manifest
- [ ] Enable `nlb.alias`
- [ ] Allow disabling `http` by `http:false`, also update env controller parameter
- [ ] Update `svc deploy` output to provide NLB URL

Previous PR: #3008 

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
+++ b/internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
@@ -185,7 +185,10 @@ func (s *LoadBalancedWebService) Template() (string, error) {
 		allowedSourceIPs = append(allowedSourceIPs, string(ipNet))
 	}
 
-	nlb := convertNetworkLoadBalancer()
+	nlb, err := s.convertNetworkLoadBalancer()
+	if err != nil {
+		return "", err
+	}
 	content, err := s.parser.ParseLoadBalancedWebService(template.WorkloadOpts{
 		Variables:                s.manifest.Variables,
 		Secrets:                  s.manifest.Secrets,
@@ -228,7 +231,7 @@ func (s *LoadBalancedWebService) Template() (string, error) {
 	return string(overridenTpl), nil
 }
 
-func (s *LoadBalancedWebService) loadBalancerTarget() (targetContainer *string, targetPort *string) {
+func (s *LoadBalancedWebService) httpLoadBalancerTarget() (targetContainer *string, targetPort *string) {
 	containerName := s.name
 	containerPort := strconv.FormatUint(uint64(aws.Uint16Value(s.manifest.ImageConfig.Port)), 10)
 	// Route load balancer traffic to main container by default.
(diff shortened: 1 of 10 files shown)


internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -185,10 +185,13 @@ func (s *LoadBalancedWebService) Template() (string, error) {
...
-	nlb := convertNetworkLoadBalancer()
+	nlb, err := s.convertNetworkLoadBalancer()
+	if err != nil {
+		return "", err
+	}
 	content, err := s.parser.ParseLoadBalancedWebService(template.WorkloadOpts{
-		Variables:                s.manifest.Variables,
-		Secrets:                  s.manifest.Secrets,
+		Variables:                s.manifest.TaskConfig.Variables,
+		Secrets:                  s.manifest.TaskConfig.Secrets,
```
iamhopaul123, line 194: Since `TaskConfig` is inline, maybe it's ok for us to abbreviate?

internal/pkg/manifest/validate.go
```diff
@@ -1124,14 +1144,18 @@ type validateWindowsOpts struct {
...
-func validateLoadBalancerTarget(opts validateLoadBalancerTargetOpts) error {
-	if opts.routingRule.TargetContainer == nil && opts.routingRule.TargetContainerCamelCase == nil {
+
+type validateTargetContainerOpts struct {
+	mainContainerName string
+	targetContainer   *string
+	sidecarConfig     map[string]*SidecarConfig
+}
+
+func validateTargetContainer(opts validateTargetContainerOpts) error {
```
iamhopaul123, line 1154: nice refactoring!

iamhopaul123 (approved):
Looks good!

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
Lou1415926: It gets complaints about `Variables` being ambiguous because there are `Variables` and `Secrets` instead of `Logging`, which is inline, as well

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -193,6 +193,29 @@ type RoutingRule struct {
 	AllowedSourceIps         []IPNet `yaml:"allowed_source_ips"`
 }
 
+func (r *RoutingRule) targetContainer() *string {
```
efekarakus, line 196: nice!

efekarakus (approved):
LGTM !

PR #3011, opened 2021-11-09
chore(cli): validate vpc connectivity before initializing aurora addon for rdws

efekarakus:

### DIFF
--- a/internal/pkg/cli/storage_init.go
+++ b/internal/pkg/cli/storage_init.go
@@ -205,7 +205,7 @@ func (o *initStorageOpts) Validate() error {
 		}
 	}
 	if o.storageType != "" {
-		if err := validateStorageType(o.storageType); err != nil {
+		if err := validateStorageType(o.storageType, validateStorageTypeOpts{}); err != nil {
 			return err
 		}
 	}
--- a/internal/pkg/cli/storage_init.go
+++ b/internal/pkg/cli/storage_init.go
@@ -205,7 +205,10 @@ func (o *initStorageOpts) Validate() error {
 		}
 	}
 	if o.storageType != "" {
-		if err := validateStorageType(o.storageType, validateStorageTypeOpts{}); err != nil {
+		if err := validateStorageType(o.storageType, validateStorageTypeOpts{
+			ws:           o.ws,
+			workloadName: o.workloadName,
+		}); err != nil {
 			return err
 		}
 	}
@@ -274,21 +277,6 @@ func (o *initStorageOpts) Ask() error {
 		return err
 	}
 
-	wkld, err := o.store.GetWorkload(o.appName, o.workloadName)
-	if err != nil {
-		return fmt.Errorf("get workload: %w", err)
-	}
-	if wkld.Type == manifest.RequestDrivenWebServiceType && o.storageType == rdsStorageType {
-		log.Warningf(`%s storage is launched in private subnets of your environment's VPC. 
-To reach the database from your %s please enable the HTTP Data API: 
-https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/data-api.html. 
-
-You will also need to an IAM policy that allows your %s to make Data API calls. 
-You can simply create the policy in the Aurora addon template and create an output. The policy will be automatically attached to your %s:
-https://aws.github.io/copilot-cli/docs/developing/additional-aws-resources/#what-does-an-addon-template-look-like
-`, rdsStorageTypeOption, manifest.RequestDrivenWebServiceType, manifest.RequestDrivenWebServiceType, manifest.RequestDrivenWebServiceType)
-	}
-
 	// Storage name needs to be asked after workload because for Aurora the default storage name uses the workload name.
 	if err := o.askStorageName(); err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


internal/pkg/cli/storage_init.go
```diff
@@ -317,23 +305,36 @@ https://aws.github.io/copilot-cli/docs/developing/additional-aws-resources/#what
...
+	}); err != nil {
+		if errors.Is(err, errRDWSNotConnectedToVPC) {
+			log.Errorf(`Your %s needs to be connected to a VPC in order to use a %s resource.
+You can enable VPC connectivity by updating your manifest with:
+%s
+`, manifest.RequestDrivenWebServiceType, o.storageType, color.HiglightCodeBlock(`network:
+  vpc:
+    placement: private`))
+		}
+		return err
```
iamhopaul123, line 339: Honestly I think we should reuse the code here for `Validate()` as well by defining the error message to a variable to make sure we error out early correctly if their flag set is not correct. The `Ask()` ideally gets user input and validates it, so we'll have 
```go
func (o *initStorageOpts) askStorageType() error {
  if o.storageType != "" {
    // Handled by Validate()
    return nil
  }
  ...
}
```

internal/pkg/cli/validate.go
```diff
@@ -280,17 +283,51 @@ func validatePath(fs afero.Fs, val interface{}) error {
...
-	return fmt.Errorf(fmtErrInvalidStorageType, storageType, prettify(storageTypes))
+	if mftType != manifest.RequestDrivenWebServiceType {
+		return nil
+	}
+	data := struct {
+		Network manifest.RequestDrivenWebServiceNetworkConfig `yaml:"network"`
+	}{}
+	if err := yaml.Unmarshal(mft, &data); err != nil {
+		return fmt.Errorf("invalid storage type %s: unmarshal manifest for %s to read network config: %w", rdsStorageType, workloadName, err)
+	}
```
iamhopaul123, line 326: I wonder if we should move it to the `workspace` pkg like `WorkloadType()`

internal/pkg/cli/storage_init.go
efekarakus: The problem is that it's possible the user doesn't provide a workload name when calling `Validate`, for example: `copilot storage init -t Aurora`.

This is why we have to validate twice, does that make sense?

internal/pkg/cli/validate.go
efekarakus: I wasn't sure either, I feel like this is very niche whereas `WorkloadType` will be used more commonly that's why I've kept it here instead

internal/pkg/cli/validate_test.go
```diff
@@ -311,35 +313,95 @@ func TestValidatePath(t *testing.T) {
...
+		"should allow S3 addons": {
 			input: "S3",
 			want:  nil,
 		},
-		"DDB okay": {
+		"should allow DynamoDB allows": {
 			input: "DynamoDB",
 			want:  nil,
 		},
-		//"RDS okay": {
```
Lou1415926, line 327: :homer-disappearing:

internal/pkg/cli/validate.go
```diff
@@ -280,17 +283,51 @@ func validatePath(fs afero.Fs, val interface{}) error {
...
+
+	if storageType == rdsStorageType {
+		return validateAuroraStorageType(opts.ws, opts.workloadName)
+	}
+	return nil
+}
+
+func validateAuroraStorageType(ws manifestReader, workloadName string) error {
+	if workloadName == "" {
```
Lou1415926, line 307: Should we validate again after getting the workload name?

internal/pkg/cli/storage_init.go
```diff
@@ -317,23 +305,36 @@ https://aws.github.io/copilot-cli/docs/developing/additional-aws-resources/#what
...
+		ws:           o.ws,
+		workloadName: o.workloadName,
+	}); err != nil {
+		if errors.Is(err, errRDWSNotConnectedToVPC) {
+			log.Errorf(`Your %s needs to be connected to a VPC in order to use a %s resource.
+You can enable VPC connectivity by updating your manifest with:
+%s
+`, manifest.RequestDrivenWebServiceType, o.storageType, color.HiglightCodeBlock(`network:
+  vpc:
+    placement: private`))
```
Lou1415926, line 337: Nice!

internal/pkg/cli/storage_init.go
iamhopaul123: Ohhh gotcha that makes sense. But then still I think we should add the logic to the `Validate()` because if without the log msg, the error might not be very clear to users.

efekarakus: That makes sense! will do

internal/pkg/cli/validate.go
efekarakus: We do! `validateStorageType` is called both in `Validate` and `Ask` :)

internal/pkg/term/color/color.go
```diff
@@ -82,11 +82,16 @@ func HighlightResource(s string) string {
...
 }
 
-// HighlightCode wraps the string with the ` character, colors it to denote it's a code block, and returns it.
+// HighlightCode wraps the string s with the ` character, colors it to denote it's code, and returns it.
 func HighlightCode(s string) string {
 	return HiCyan.Sprintf("`%s`", s)
 }
 
+// HiglightCodeBlock wraps the string s with ``` characters, colors it to denote it's a multi-line code block, and returns it.
+func HiglightCodeBlock(s string) string {
```
huanjani, line 91: Ooh-la-la!

efekarakus: We can now update our help menus and recommended actions with this fn, it'll print code much nicer

internal/pkg/cli/storage_init.go
efekarakus: Done!

internal/pkg/cli/storage_init.go
```diff
@@ -318,9 +303,8 @@ https://aws.github.io/copilot-cli/docs/developing/additional-aws-resources/#what
 
 func (o *initStorageOpts) askStorageType() error {
 	if o.storageType != "" {
-		return nil
+		return o.validateStorageType()
```
iamhopaul123, line 308: One last request: maybe we should add a comment here explaining why we don't `return nil` instead, which is our common pattern for `Ask()`.

iamhopaul123 (approved):

huanjani (approved):
:shipit:

PR #3017, opened 2021-11-10
chore(templates): add aurora addon template for rdws

efekarakus:
- View the git diff between the original `aurora/cf.yml` and `aurora/rdws/cf.yml` here: https://gist.github.com/efekarakus/17f6a13faa8ee4670239844b2b1cc047

### DIFF
--- a/internal/pkg/template/templates/addons/aurora/rdws/cf.yml
+++ b/internal/pkg/template/templates/addons/aurora/rdws/cf.yml
@@ -1,0 +1,173 @@
+Parameters:
+  App:
+    Type: String
+    Description: Your application's name.
+  Env:
+    Type: String
+    Description: The environment name your service, job, or workflow is being deployed to.
+  Name:
+    Type: String
+    Description: The name of the service, job, or workflow being deployed.
+  ServiceSecurityGroupId:
+    Type: String
+    Description: The security group associated with the VPC connector.
+  # Customize your Aurora Serverless cluster by setting the default value of the following parameters.
+  {{logicalIDSafe .ClusterName}}DBName:
+    Type: String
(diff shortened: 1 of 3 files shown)


internal/pkg/template/templates/workloads/services/rd-web/cf.yml
```diff
@@ -101,7 +101,12 @@ Resources:
               - Name: {{toSnakeCase $var}}
                 Value:
                   Fn::GetAtt: [ {{$stackName}}, Outputs.{{$var}}]
-              {{- end}}
+              {{- end }}
+              {{- range $var := .NestedStack.SecretOutputs }}
+              - Name: {{toSnakeCase $var}}_ARN
+                Value:
+                  Fn::GetAtt: [ {{$stackName}}, Outputs.{{$var}}]
+              {{- end }}
```
efekarakus, line 109: I added secrets arns as environment variables because users have to fetch the value of the secret from their application code. Therefore, they need to know the ARN.

In order to be **future compatible** for when app runner supports secrets, we're writing these env var names with the `_ARN` suffix. If we did not do this and the service starts supporting secrets then the environment variable names will collide.

internal/pkg/template/templates/addons/aurora/rdws/cf.yml
```diff
@@ -0,0 +1,173 @@
+Parameters:
```
efekarakus, line 1: View diff for the file here: https://gist.github.com/efekarakus/17f6a13faa8ee4670239844b2b1cc047

Lou1415926 (approved):
LGTM!

iamhopaul123 (approved):
:shipit:

PR #3018, opened 2021-11-10
chore: enable TLS termination for NLB

Lou1415926:
Previous PR: #3010 

### DIFF
--- a/internal/pkg/template/templates/environment/cf.yml
+++ b/internal/pkg/template/templates/environment/cf.yml
@@ -174,7 +174,7 @@ Resources:
       Protocol: HTTP
   HTTPSListener:
     Type: AWS::ElasticLoadBalancingV2::Listener
-    DependsOn: HTTPSCertArn
+    DependsOn: HTTPSCert
     Condition: ExportHTTPSListener
     Properties:
       Certificates:
--- a/internal/pkg/template/templates/environment/cf.yml
+++ b/internal/pkg/template/templates/environment/cf.yml
@@ -174,7 +174,7 @@ Resources:
       Protocol: HTTP
   HTTPSListener:
     Type: AWS::ElasticLoadBalancingV2::Listener
-    DependsOn: HTTPSCert
+    DependsOn: HTTPSCertArn
     Condition: ExportHTTPSListener
     Properties:
       Certificates:
@@ -356,6 +356,11 @@ Outputs:
     Value: !Ref HTTPSListener
     Export:
       Name: !Sub ${AWS::StackName}-HTTPSListenerArn
+  HTTPSCertArn:
+    Condition: DelegateDNS
+    Value: !Ref HTTPSCert
+    Export:
+      Name: !Sub ${AWS::StackName}-HTTPSCert
   DefaultHTTPTargetGroupArn:
     Condition: CreateALB
     Value: !Ref DefaultHTTPTargetGroup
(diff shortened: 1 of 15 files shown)


internal/pkg/template/templates/environment/cf.yml
```diff
@@ -174,7 +174,7 @@ Resources:
       Protocol: HTTP
   HTTPSListener:
     Type: AWS::ElasticLoadBalancingV2::Listener
-    DependsOn: HTTPSCert
+    DependsOn: HTTPSCertArn
```
iamhopaul123: Why does the listener depend on the output `HTTPSCertArn` 🤔

internal/pkg/template/workload.go
```diff
@@ -532,6 +533,9 @@ func envControllerParameters(o WorkloadOpts) []string {
 	if o.Storage != nil && o.Storage.requiresEFSCreation() {
 		parameters = append(parameters, "EFSWorkloads,")
 	}
+	if o.NLB != nil {
+		parameters = append(parameters, "NLB") // Invoke env controller so that the env controller action is invoked, retrieving the exported HTTPSCert ARN.
+	}
```
iamhopaul123, line 538: Just curious is this just to force an update to the env controller so that it can have the latest output?

internal/pkg/template/templates/environment/cf.yml
Lou1415926: it was a typo. Sorry.

internal/pkg/template/workload.go
Lou1415926: yes

iamhopaul123: Gotcha. For extensibility concern (like we don't need to add more param like this for the other features merely to trigger an update), I wonder if we should do similar thing as https://github.com/aws/copilot-cli/blob/1fbeceb3647e5711d787368ef2f0ec6ceca96a68/internal/pkg/template/templates/workloads/partials/cf/autoscaling.yml#L13

efekarakus (commented):
👍 to boss's comment, otherwise looks good!

internal/pkg/template/workload.go
```diff
@@ -91,6 +91,12 @@ var (
 	osFamiliesForPV100 = []string{
 		OSWindowsServerFull, OSWindowsServerCore,
 	}
+
+	// Changing the suffix version number will force trigger the env controller for once.
+	// The suffix version number should be the latest environment template version at the time when a force trigger needs to happen.
+	// This is useful when we add a new output in the environment stack, and we need to reference
+	// the new output from the workload stack. See example: https://github.com/aws/copilot-cli/pull/3018
+	forceTriggerEnvController = "ForceTriggerEnvController-v1.7.0"
```
efekarakus: Is the goal of this parameter to once and only once trigger the envcontroller to fetch outputs? Am I  misunderstanding something?


Is there a particular reason why we can't do, something like this:
```
{{- if eq .NLB.Listener.Protocol "TLS"}}
DependsOn: EnvControllerAction
{{- end }}
Properties:
 ...
{{- if eq .NLB.Listener.Protocol "TLS" }}
    Certificates:
    - CertificateArn:
         Fn::ImportValue: !Sub '${AppName}-${EnvName}-HTTPSCertArn'
```

Let me know if I'm missing something 🤔 I'd really love to be able to be able to remove this parameter if possible

Lou1415926:
Closing this for a big revise. The architecture assumed in this PR is incorrect. Instead of reusing the HTTPS certificate, the NLB needs its own certificate.

PR #3019, opened 2021-11-10
chore(cli): should be able to parse codecommit short url

### LINKED ISSUE #3012: Support for git-remote-codecommit on pipeline command
The AWS CodeCommit now provides an Git credential helper called `git-remote-codecommit` ([blog announcement](https://aws.amazon.com/about-aws/whats-new/2020/03/aws-codecommit-introduces-open-source-remote-helper/)) that simplifies interaction with repositories and IAM.

Looks like `copilot-cli` is not yet compatible with this helper. When I tried to create an Pipeline using the CLI version `1.12.0` i got the following error:

```bash
$ git remote -v
origin	codecommit://helloecs (fetch)
origin	codecommit://helloecs (push)

$ copilot pipeline init
1st stage: test
2nd stage: production
Repository URL: codecommit://helloecs
✘ unknown CodeCommit URL format: codecommit://helloecs
```

iamhopaul123:
fixes #3012 


internal/pkg/cli/pipeline_init.go
```diff
@@ -434,13 +437,17 @@ func (url ccRepoURL) parse() (ccRepoDetails, error) {
...
 	default:
 		return ccRepoDetails{}, fmt.Errorf("unknown CodeCommit URL format: %s", url)
 	}
-	// Double-check that parsed results is a valid region. Source: https://www.regextester.com/109163
-	match, _ := regexp.MatchString(`(us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d`, region)
-	if !match {
-		return ccRepoDetails{}, fmt.Errorf("unable to parse the AWS region from %s", url)
+	if region != "" {
+		// Double-check that parsed results is a valid region. Source: https://www.regextester.com/109163
+		match, _ := regexp.MatchString(`(us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d`, region)
```
Lou1415926, line 447: Do we need ability to parse these during `pipeline update` as well? https://github.com/aws/copilot-cli/blob/e67a24e90b02eb501eb2f09506bb7460686aa607/internal/pkg/deploy/pipeline.go#L34

iamhopaul123: Good callout! This seems to be another issue which is more about our manifest design. Posted a new design for tracking this https://github.com/aws/copilot-cli/issues/3021

internal/pkg/cli/pipeline_init.go
```diff
@@ -434,13 +437,17 @@ func (url ccRepoURL) parse() (ccRepoDetails, error) {
 	case strings.HasPrefix(urlString, "codecommit::"):
 		parsedURL := strings.Split(urlString, ":")
 		region = parsedURL[2]
+	case strings.HasPrefix(urlString, "codecommit://"):
+		// Use default profile region.
```
efekarakus, line 441: Would it be better to keep all region related logic in this method? i.e. move the above changes here, something like:
```go
appRegion := aws.StringValue(sess.Config.Region)
if appRegion == "" {
   return ccRepoDetails{}, errors.New("unable to find a region associated with your session")
}
region = appRegion
```

iamhopaul123: Hmm the reason behind this is because the `ccRepoURL.parse()` method is supposed to parse info from the URL instead of getting from elsewhere.

efekarakus (approved):

internal/pkg/cli/pipeline_init.go
efekarakus: that makes sense!

Lou1415926 (approved):

PR #3022, opened 2021-11-11
chore(cli): write tpl and params file for aurora and rdws

efekarakus:
Manual testing results in the resource creations.


internal/pkg/term/color/color.go
```diff
@@ -87,8 +87,8 @@ func HighlightCode(s string) string {
 	return HiCyan.Sprintf("`%s`", s)
 }
 
-// HiglightCodeBlock wraps the string s with ``` characters, colors it to denote it's a multi-line code block, and returns it.
-func HiglightCodeBlock(s string) string {
```
iamhopaul123, line 91: 🤣

internal/pkg/cli/storage_init.go
```diff
@@ -718,6 +763,22 @@ func (o *initStorageOpts) environmentNames() ([]string, error) {
...
+	mft, err := o.ws.ReadWorkloadManifest(o.workloadName)
+	if err != nil {
+		return "", fmt.Errorf("read manifest for %s: %w", o.workloadName, err)
+	}
+	t, err := mft.WorkloadType()
+	if err != nil {
+		return "", fmt.Errorf("read 'type' from manifest for %s: %w", o.workloadName, err)
+	}
+	o.workloadType = t
+	return o.workloadType, nil
```
iamhopaul123: We don't need to return the type but just use `o.workloadType` right? or we abandon `o.workloadType` this cached variable completely

iamhopaul123 (approved):
LGTM just a very nit comment

huanjani (approved):
Looks good to me. Added DNM label so you can address PH's Q.

internal/pkg/cli/storage_init.go
efekarakus: Refactored! from now we'll call `readWorkloadType` first in `Execute` and only refer to the cached variable afterwards

PR #3023, opened 2021-11-11
chore(manifest): add comment to connect vpc for rdws

efekarakus:


internal/pkg/template/templates/workloads/services/rd-web/manifest.yml
```diff
@@ -33,6 +33,11 @@ cpu: {{.InstanceConfig.CPU}}
 # Amount of memory in MiB used by the task.
 memory: {{.InstanceConfig.Memory}}
 
+# # Connect your App Runner service to your environment's VPC.
+# network:
+#   vpc:
+#     placement: 'private'
```
huanjani: We don't have single quotes in other commented out strings. Should we be consistent?

efekarakus: Good catch! updated

huanjani (approved):
LGTM!

iamhopaul123 (approved):

Issue #2812, opened 2021-09-08
Add support for GRPC Protocol in Target Groups


jeastham1993:
Hey there, thanks for putting together such a great CLI. Really helps to abstract away a lot of the complexity around VPC and network 'best practices'.

For a lot of internal services I deploy we use GRPC instead of HTTP, as far as I understand the cloud formation changes required I think it would be relatively straightforward to add support for conditional GRPC for load-balanced and backend services. I'm by no means an expert in the networking side of AWS, so please correct me if wrong. 

Something like

```yml

  TargetGroup:
    Metadata:
      'aws:copilot:description': 'A target group to connect the load balancer to your service'
    Type: AWS::ElasticLoadBalancingV2::TargetGroup
    Properties:
      HealthCheckPath: {{.HTTPHealthCheck.HealthCheckPath}} # Default is '/'.
{{- if .HTTPHealthCheck.SuccessCodes}}
      Matcher: 
        HttpCode: {{.HTTPHealthCheck.SuccessCodes}}
{{- end}}
{{- if .HTTPHealthCheck.HealthyThreshold}}
      HealthyThresholdCount: {{.HTTPHealthCheck.HealthyThreshold}}
{{- end}}
{{- if .HTTPHealthCheck.UnhealthyThreshold}}
      UnhealthyThresholdCount: {{.HTTPHealthCheck.UnhealthyThreshold}}
{{- end}}
{{- if .HTTPHealthCheck.Interval}}
      HealthCheckIntervalSeconds: {{.HTTPHealthCheck.Interval}}
{{- end}}
{{- if .HTTPHealthCheck.Timeout}}
      HealthCheckTimeoutSeconds: {{.HTTPHealthCheck.Timeout}}
{{- end}}
      Port: !Ref ContainerPort
      Protocol: HTTP
{{- if .HTTP.EnableGrpc}}
      ProtocolVersion: GRPC
{{- end}}

```

Again, networking is not my fortay so there may be more to think about. It would be a really helpful addition to an already great tool though.

seiichi1101:
I created PR, which enables to set protocol version.

efekarakus:
This is now released 🎉 in v1.13.0 https://github.com/aws/copilot-cli/releases/tag/v1.13.0 !

PR #3026, opened 2021-11-12
feat: Add protocol version setting

seiichi1101:
## Summary

Enable to set ALB Target Group Protocol Version.
The possible values are `GRPC`, `HTTP1`, and `HTTP2`. The default is `HTTP1`.

https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#target-group-protocol-version

## Example

example `manifest.yml` for gRPC

```
name: grpc

http:
  alias: test.copilot-e2e-tests.ecs.aws.dev
  protocolVersion: GRPC
  path: '/'
...
```

## Related issue

#2812

---


e2e/grpc-svc-app/grpc_svc_app_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+var (
+	initErr error
+)
+
+var _ = Describe("gRPC Service App", func() {
+	const domainName = "copilot-e2e-tests.ecs.aws.dev"
```
iamhopaul123: Since we run our e2e tests in parallel. I wonder if this would be a problem when this test suite is running with `app-with-domain` suite. Running them in order might not be a good idea as well since `app-with-domain` already takes decent amount of time. Maybe we should consider having an extra e2e test domain before adding this test suite. What do you think @efekarakus?

iamhopaul123 (commented):
The feature implementation looks good! Just have one concern regarding the new e2e test suite.

.release/buildspec_e2e.yml
```diff
@@ -91,6 +91,15 @@ batch:
...
           APP_REGION: us-west-2
+    - identifier: grpc_svc_app
+      env:
+        privileged-mode: true
+        type: LINUX_CONTAINER
+        compute-type: BUILD_GENERAL1_LARGE
+        image: aws/codebuild/standard:5.0
+        variables:
+          TEST_SUITE: grpc-svc-app
+          APP_REGION: us-west-2
```
efekarakus: Is it possible to update the region to `us-east-2`? 🙏  
Today we have 3 tests running in `us-west-2` and we're trying to balance the tests evenly across regions

e2e/grpc-svc-app/grpc_svc_app_test.go
efekarakus: Yeah, just to expand on @iamhopaul123's point: @seiichi1101 your code looks good!

It's just that we have limitations with our e2e tests and can run into problem if multiple tests use a domain name. Would you mind updating the manifest file to remove `alias: test.copilot-e2e-tests.ecs.aws.dev` ?

e2e/grpc-svc-app/grpc/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+# Install certs
+RUN apk --no-cache add ca-certificates
+# Copy the binary from the builder image
+COPY --from=builder /app ./
+# Make the binary executable
+RUN chmod +x ./grpc
+# Start the service
+ENTRYPOINT ["./grpc"]
+# The service runs on port 50051
+EXPOSE 50051
```
efekarakus: Is this a typical port for grpc?

I wonder if we can do something nice for users and automatically fill the `protocolVersion` field in the manifest field to be `grpc` if we detect certain port numbers?

e2e/grpc-svc-app/grpc_svc_app_test.go
```diff
@@ -0,0 +1,112 @@
...
+			wantedURLs := map[string]string{
+				"test": "https://test.copilot-e2e-tests.ecs.aws.dev",
+			}
+			for _, route := range svc.Routes {
+				// Validate route has the expected HTTPS endpoint.
+				Expect(route.URL).To(Equal(wantedURLs[route.Environment]))
+
+				// Make sure error response is nil.
+				tlsCredentials := credentials.NewTLS(&tls.Config{})
+				conn, err := grpc.Dial(route.URL, grpc.WithTransportCredentials(tlsCredentials))
```
efekarakus: is there a way of testing if the response message contains "Hello World" ?

go.mod
```diff
@@ -54,6 +54,8 @@ require (
 	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c // indirect
 	golang.org/x/term v0.0.0-20210503060354-a79de5458b56 // indirect
 	golang.org/x/text v0.3.6 // indirect
+	google.golang.org/genproto v0.0.0-20210602131652-f16073e35f0c // indirect
```
efekarakus: can we run `go mod tidy`? I don't think the main copilot binary needs a dependency on `grpc` 🙏

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -183,6 +183,7 @@ func (s LoadBalancedWebService) ApplyEnv(envName string) (WorkloadManifest, erro
 // RoutingRule holds the path to route requests to the service.
 type RoutingRule struct {
 	Path                *string                 `yaml:"path"`
+	ProtocolVersion     *string                 `yaml:"protocolVersion"`
```
efekarakus: What do you think of renaming the field like this? Since it's already under `http`.
```suggestion
	ProtocolVersion     *string                 `yaml:"version"`
```

_Side note: Copilot uses snake_case for field names, so protocol_version instead of protocolVersion_.

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
```diff
@@ -170,6 +170,9 @@ Resources:
 {{- end}}
       Port: !Ref ContainerPort
       Protocol: HTTP
+{{- if .ProtocolVersion}}
+      ProtocolVersion: {{.ProtocolVersion}}
```
efekarakus: is it possible that we accept any case for the value? Users should be able to write:
```
http:
  version: grpc
```

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -183,6 +183,7 @@ func (s LoadBalancedWebService) ApplyEnv(envName string) (WorkloadManifest, erro
 // RoutingRule holds the path to route requests to the service.
 type RoutingRule struct {
```
efekarakus, line 184: Is it possible to also update this method? 🥺  https://github.com/aws/copilot-cli/blob/479f892092b3979ef3cb961a0e1c1ceba89fc7de/internal/pkg/manifest/validate.go#L495

such that:
```go
var httpProtocolVersions = []string{"grpc", " http1", "http2"} // global variable.

// in Validate method:
if r.ProtocolVersion != nil {
   if !contains(strings.ToLower(*r.ProcolVersion), httpProtocolVersions) {
      return fmt.Errorf(`"protocol" field value '%s' must be one of %s`, *r.ProtocolVersion, english.WordSeries(httpProtocolVersions, "or"))
   }
}
```

efekarakus (commented):
Looks great, thank you so much for the contribution @seiichi1101 😍

Let me know what you think of the feedback below!

.release/buildspec_e2e.yml
seiichi1101: OK. I will fix it.

go.mod
seiichi1101: Sorry. I will fix it. 🙏

internal/pkg/manifest/lb_web_svc.go
seiichi1101: OK. I will rename to `version`.

e2e/grpc-svc-app/grpc_svc_app_test.go
seiichi1101: No, it's probably won't.
Because protocol version `grpc` of the target group is only working on listener protocol `HTTPS`.
Could you give some advice please? 

> To use the feature on your ALB, choose HTTPS as your listener protocol, gRPC as the protocol version for your target group and register instance or IP as targets for the configured target group. 

https://aws.amazon.com/about-aws/whats-new/2020/10/application-load-balancers-enable-grpc-workloads-end-to-end-http-2-support/

e2e/grpc-svc-app/grpc/Dockerfile
seiichi1101: I think so.
On the official tutorial, 50051 is usually assigned for server port in any language.
> https://grpc.io/docs/languages/

e2e/grpc-svc-app/grpc_svc_app_test.go
seiichi1101: Yes. It is possible if I append code like below.
But to do this, I need to import the sub-module package `github.com/aws/copilot-cli/e2e/grpc-svc-app/grpc/proto/gen`.

```
import (
  pb "github.com/aws/copilot-cli/e2e/grpc-svc-app/grpc/proto/gen"
...
        conn, err := grpc.Dial(route.URL, grpc.WithTransportCredentials(tlsCredentials))
        //code for request in grpc-src-app e2e
        Eventually(func() error {
          client := pb.NewGreeterClient(conn)
          _, err := client.SayHello(context.Background(), &pb.Empty{})
          return err
        }, "60s", "1s").Should(BeNil())
...
```
I thought it possible like [this code](https://github.com/aws/copilot-cli/blob/mainline/e2e/multi-svc-app/multi_svc_app_test.go#L14).
Could you give some advice, please?

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
seiichi1101: OK. I will try.

internal/pkg/manifest/lb_web_svc.go
seiichi1101: OK. I will try.

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
seiichi1101: I tried to try this, but I am thinking about the timing to convert the case.

1. convert to uppercase when loading the manifest file
2. convert to uppercase when generating the cloud formation template

Depending on how I implement it, I need to change the validation logic you asked me to add `internal/pkg/manifest/lb_web_svc.go`.

Since CloudFormation allows only uppercase GRPC/HTTP1/HTTP2 for the Protocol Version, I think it simply makes the user enter uppercase.

Could you give some advice, please?

e2e/grpc-svc-app/grpc_svc_app_test.go
efekarakus: Awesome thanks @seiichi1101 ! Then I think we can keep it as is, to be honest I think your integration tests are awesome, and we probably don't need an e2e test for this scenario.

e2e/grpc-svc-app/grpc/Dockerfile
efekarakus: Thanks for the info!

efekarakus (approved):

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
efekarakus: No worries the code looks great!

The conversion between "manifest" input to "cluodformation" input happens in the `cloudformation/stack` pkg. For example:
https://github.com/aws/copilot-cli/blob/47a6e7e4b90a1977233895476100f5b363641876/internal/pkg/deploy/cloudformation/stack/lb_web_svc.go#L168

I've added an extra commit to the PR to make the change, thank you! 🎉

iamhopaul123 (approved):

efekarakus:
This is now released 🎉 in v1.13.0 https://github.com/aws/copilot-cli/releases/tag/v1.13.0 !

PR #3039, opened 2021-11-16
chore: revert rdws vpc for next release

iamhopaul123:

### DIFF
--- a/internal/pkg/cli/storage_init.go
+++ b/internal/pkg/cli/storage_init.go
@@ -305,8 +305,6 @@ func (o *initStorageOpts) Ask() error {
 
 func (o *initStorageOpts) askStorageType() error {
 	if o.storageType != "" {
-		// Not all storage options, such as Aurora for RDWS, is available to workloads by default.
-		// We want to fail fast for the user once we know what workload type is the addon for.
 		return o.validateStorageType()
 	}
 	var options []prompt.Option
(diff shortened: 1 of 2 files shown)


internal/pkg/cli/validate.go
```diff
@@ -61,7 +59,6 @@ var (
 
 	// Aurora-Serverless-specific errors.
 	errInvalidRDSNameCharacters    = errors.New("value must start with a letter")
-	errRDWSNotConnectedToVPC       = fmt.Errorf("%s requires a VPC connection", manifest.RequestDrivenWebServiceType)
```
efekarakus: What do you think about retaining all the changes in `validate.go` this way it fails fast telling the users that RDWS cannot be associated with Aurora as it requires VPC connectivity?

efekarakus (approved):

internal/pkg/cli/storage_init.go
```diff
@@ -330,14 +328,6 @@ func (o *initStorageOpts) validateStorageType() error {
 		ws:           o.ws,
 		workloadName: o.workloadName,
 	}); err != nil {
```
Lou1415926, line 330: Do we want to mention "vpc connector" at all in the error message? Right now if the usre try to use RDS addon for an app runner service - and of course they wouldn't have `network` in the manifest - Copilot will say "you need a VPC connector" which doesn't make sense to them because, what is a VPC connector anyway 🤔

Lou1415926 (approved):
Putting the label just in case you need to make changes. Feel free to just remove it and merge though.

internal/pkg/cli/storage_init.go
iamhopaul123: It says `requires a VPC connection` which I think should be probably fine? As it is indeed the reason why they cannot use RDS addon 🤔

PR #3040, opened 2021-11-16
docs: add `http.version` to lbws manifest

efekarakus:
Also parse the Dockerfile for port `50051` that's commonly used for gRPC and set in the manifest file.

### DIFF
--- a/internal/pkg/initialize/workload.go
+++ b/internal/pkg/initialize/workload.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/copilot-cli/internal/pkg/config"
@@ -27,6 +28,10 @@ const (
 const (
 	jobWlType = "job"
 	svcWlType = "service"
+)
+
+const (
+	commonGRPCPort = uint16(50051)
 )
 
 var fmtErrUnrecognizedWlType = "unrecognized workload type %s"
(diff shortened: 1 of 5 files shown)


internal/pkg/manifest/lb_web_svc.go
```diff
@@ -23,6 +23,10 @@ const (
 	DefaultHealthCheckGracePeriod = 60
 )
 
+const (
+	GRPCProtocol = "gRPC" // GRPCProtocol is the HTTP protocol version for gRPC.
```
iamhopaul123, line 27: We should probably want to be consistent with
```go
```

iamhopaul123 (approved):
LGTM just a very tiny little nit

internal/pkg/manifest/lb_web_svc.go
efekarakus: Those values are the ones that are required by CFN, whereas the typical spelling is ["gRPC"](https://grpc.io/). We just ignore the case and move everything to upper case later when transforming the user input, does that make sense?

iamhopaul123: Yeah that makes sense and also why i said this is just very tiny little nit. it's just i think we'd better to be consistent. But yeah if they are all upper case sure that sounds good to me. Maybe the var name should be `httpProtocolVersionsUpperCase`? You decide boss.

internal/pkg/initialize/workload.go
```diff
@@ -298,16 +303,23 @@ func (w *WorkloadInitializer) newServiceManifest(i *ServiceProps) (encoding.Bina
 }
 
 func (w *WorkloadInitializer) newLoadBalancedWebServiceManifest(i *ServiceProps) (*manifest.LoadBalancedWebService, error) {
+	var httpVersion string
+	if i.Port == commonGRPCPort {
+		log.Infof("Detected port %s, setting HTTP protocol version to %s in the manifest.\n",
+			color.HighlightUserInput(strconv.Itoa(int(i.Port))), color.HighlightCode(manifest.GRPCProtocol))
+		httpVersion = manifest.GRPCProtocol
```
Lou1415926, line 310: Nice!

Lou1415926 (approved):

internal/pkg/initialize/workload.go
Lou1415926: (Adding the label just in case you want to make more changes boss)

Issue #3044, opened 2021-11-17
task run can't set docker context


nbarrera:
I find no way to define which is the docker context directory when using `copiilot task run`

I could use the `--dockerfile` argument and the docker context will be the `Dockerfile`'s directory which it is useful most of the times, but currently I was making use of the manifest's `build / context:` yaml property to use a different context.

That manifest is for a scheduled task, and I would like to also specify the build context when I 'm running that task manually as a one-off task.

The use case is, the scheduled task failed so I debug and fix it so I want to run it manually and not wait until next trigger time... this is not possible if your scheduled task is making use of the `build / context:` manifest property.

efekarakus:
Hi @nbarrera ! Thank you, the feature request makes a lot of sense. Looks like today we just assume that the context is the path of the dockerfile:
https://github.com/aws/copilot-cli/blob/4597bdc00df0898dd22472422b7fd61ba0faecba/internal/pkg/cli/task_run.go#L754-L756

In order to be backwards compatible, we can introduce a new flag `--build-context` to override the value.

In the mean time, would it be possible to copy the Dockerfile into the context path that might help you get unblocked

nbarrera:
Hi @efekarakus... yes my other option is to put Dockerfiles on the directory I was trying to set as the context. That's not so sexy as I have a monorepo structure, then I would end up with a directory for each module, but Dockerfiles for each module will be on the root, kind of awkward... still doable.

In the mean time I was checking the code that you now mention.

I came up with a fork and a [PR](https://github.com/nbarrera/copilot-cli/pull/1)... didn't went for the PR against your repo because I didn't follow all steps on the contribution guideline... I reached only up to `make tests`.

But anyway you guys may want to check it if you think it's useful.

thank you!

efekarakus:
Thank you! this is awesome 🎉 I can definitely pick it up from your PR and add the additional commits.

I should be able to work on it by end of day!

efekarakus:
If you want to, you can create the PR against this repo `aws/copilot` instead of your fork so you get the contribution :D

nbarrera:
will do, 👍 
thank you

for the record.. I kind of followed the uses of `--dockerfile` and mimic that for the new `--build-context` hope that gets you close enough.

efekarakus:
Thank you for the PR @nbarrera ! It's now merged 😄 it should go out next release

We have also updated the help menu of `task run` in the PR to make it easier to find flags:
[...]

PR #3046, opened 2021-11-17
feat: add `--build-context` flag to task run

nbarrera:
this optional flag allows the user to provide the docker context path

if not provided, it fallbacks to the previous behaveior so this change
is transparently retro compatible, it takes the context from the
Dockerfile directory.

Addresses #3044

### DIFF
--- a/internal/pkg/cli/task_run.go
+++ b/internal/pkg/cli/task_run.go
@@ -10,6 +10,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/spf13/pflag"
+
 	"github.com/aws/copilot-cli/internal/pkg/manifest"
 
 	"github.com/aws/copilot-cli/internal/pkg/docker/dockerengine"
@@ -84,10 +86,10 @@ type runTaskVars struct {
 
 	groupName string
 
-	image          string
-	dockerfilePath string
+	image                 string
+	dockerfilePath        string
 	dockerfileContextPath string
-	imageTag       string
+	imageTag              string
 
 	taskRole      string
 	executionRole string
@@ -115,7 +117,6 @@ type runTaskVars struct {
 type runTaskOpts struct {
 	runTaskVars
 	isDockerfileSet bool
-	isDockerfileContextSet bool
 	nFlag           int
 
 	// Interfaces to interact with dependencies.
@@ -238,7 +239,7 @@ func (o *runTaskOpts) configureRunner() (taskRunner, error) {
 			App: o.appName,
 			Env: o.env,
 
-			OS:   o.os,
+			OS: o.os,
 
 			VPCGetter:            vpcGetter,
 			ClusterGetter:        ecs.New(o.sess),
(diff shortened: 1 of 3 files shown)


internal/pkg/cli/task_run.go
```diff
@@ -909,6 +928,9 @@ Run a task with a command.
 			if cmd.Flags().Changed(dockerFileFlag) {
 				opts.isDockerfileSet = true
 			}
+			if cmd.Flags().Changed(dockerFileContextFlag) {
+				opts.isDockerfileContextSet = true
+			}
```
iamhopaul123: We probably don't need this as we dont' set the default value in https://github.com/aws/copilot-cli/pull/3046/files#diff-eec95b70c382971e4ec97c1db10feebe2b837deaba537fc9ec999e43ea62a9bbR946. We could simply check if it equals to "" to determine if this flag is set by users or not.

internal/pkg/cli/task_run.go
```diff
@@ -751,9 +763,16 @@ func (o *runTaskOpts) buildAndPushImage() error {
...
+	if o.dockerfileContextPath == "" {
+		dockerContext = filepath.Dir(o.dockerfilePath)
+	} else {
+		dockerContext = o.dockerfileContextPath
+	}
+
 	if _, err := o.repository.BuildAndPush(dockerengine.New(exec.NewCmd()), &dockerengine.BuildArguments{
 		Dockerfile: o.dockerfilePath,
-		Context:    filepath.Dir(o.dockerfilePath),
+		Context:    dockerContext,
```
iamhopaul123: I wonder if we can set the default value for `o.dockerfileContextPath` in `newTaskRunOpts()` to have the default value to be
```go
if o.dockerfileContextPath == "" {
  o.dockerfileContextPath = filepath.Dir(o.dockerfilePath)
}
```
Then we can have `Context: o.dockerfileContextPath` here.

internal/pkg/cli/task_run.go
```diff
@@ -310,12 +312,22 @@ func (o *runTaskOpts) Validate() error {
...
 		if _, err := o.fs.Stat(o.dockerfilePath); err != nil {
 			return err
 		}
 	}
 
+	if o.isDockerfileContextSet {
+		if _, err := o.fs.Stat(o.dockerfileContextPath); err != nil {
+			return err
+		}
+	}
```
iamhopaul123: After https://github.com/aws/copilot-cli/pull/3046/files#r751679516 we could simply have
```suggestion
        if o.image != "" {
        	if _, err := o.fs.Stat(filepath.Join(o.dockerfileContextPath, o.dockerfilePath)); err != nil {
		return err
		}
        }
```
to check if the dockerfile exists.

iamhopaul123 (approved):

internal/pkg/cli/task_run.go
efekarakus: 👍 Done!

internal/pkg/cli/task_run.go
efekarakus: I did this slightly differently but it's essentially the same if-statement

internal/pkg/cli/task_run.go
efekarakus: It's possible that the Dockerfile and Context and in completely separate directories, that's why the stat checks are separate

efekarakus (approved):

Lou1415926:
This is now released 🎉 in v1.14.0 https://github.com/aws/copilot-cli/releases/tag/v1.14.0 ! Thank you for the contribution!

PR #3056, opened 2021-11-18
chore(e2e): add sd to apprunner e2e suite

iamhopaul123:


e2e/apprunner/apprunner_suite_test.go
```diff
@@ -38,13 +39,7 @@ var _ = BeforeSuite(func() {
...
 
 var _ = AfterSuite(func() {
-	_, err := cli.SvcDelete(svcName)
-	Expect(err).NotTo(HaveOccurred())
-
-	_, err = cli.EnvDelete(envName)
-	Expect(err).NotTo(HaveOccurred())
-
-	_, err = cli.AppDelete()
+	_, err := cli.AppDelete()
```
efekarakus, line 42: 😂

e2e/apprunner/apprunner_test.go
```diff
@@ -180,12 +205,17 @@ var _ = Describe("App Runner", func() {
 			}
 		})
 
-		It("should return svc endpoints", func() {
+		It("should return svc endpoints and service discovery should work", func() {
```
efekarakus, line 222: niceee

Can we also add `copilot storage init -t Aurora` to the "front-end" service. I think both SD and RDS are common usecases that we'll want to make sure works for RDWS? 🙏

efekarakus: Maybe something like `http.Get(fmt.Sprintf("%s/db", route.URL))` which returns a timestamp from the database.

For example, in JavaScript I had this working:
```js
app.get('/db', async (req, res) => {
  const dbSecret = await client.getSecretValue({SecretId: process.env.WWWCLUSTER_SECRET_ARN}).promise();
  const {username, host, dbname, password, port} = JSON.parse(dbSecret.SecretString);
  try {
    const client = new Client({
      user: username,
      host: host,
      database: dbname,
      password: password,
      port: port,
    });
    client.connect();
    const response = await client.query('SELECT NOW() as now');
    
    res.send(`Hello results: ${JSON.stringify(response.rows[0])}`);
  } catch (err) {
    res.send(`db error: ${JSON.stringify(err)}`);
  };
});
```

efekarakus (commented):
Looks great for SD! 

One big request boss 🥺

e2e/apprunner/apprunner_test.go
iamhopaul123: Seems like even after the VPC support for App runner, there's still no native way for us to support aurora with our RDWS because app runner doesn't support secrets natively 🤯 they would need to use aws sdk in their code which require extra addons permissions to get the secret instead of directly using it, which I won't expect it to be a very popular pattern. Could we add it once app runner supports secrets from ssm and secret manager?

iamhopaul123: ok addressed

e2e/apprunner/back-end/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/aws/copilot-cli/e2e/appprunner/back-end
+
+go 1.14
```
efekarakus, line 3: 👴

e2e/apprunner/apprunner_test.go
```diff
@@ -180,45 +219,45 @@ var _ = Describe("App Runner", func() {
...
-				resp, fetchErr := http.Get(svc.Routes[0].URL)
+				resp, fetchErr := http.Get(route.URL)
+				return resp.StatusCode, fetchErr
+			}, "30s", "1s").Should(Equal(200))
+			Eventually(func() (int, error) {
+				resp, fetchErr := http.Get(fmt.Sprintf("%s/service-discovery-test/", route.URL))
+				return resp.StatusCode, fetchErr
+			}, "30s", "1s").Should(Equal(200))
+			Eventually(func() (int, error) {
+				resp, fetchErr := http.Get(fmt.Sprintf("%s/db/", route.URL))
```
efekarakus, line 236: nice! 🆒

efekarakus (approved):

Lou1415926 (approved):

PR #3057, opened 2021-11-18
chore(custom-resource): implement lambda function that validate certificate and add aliases for NLB 

Lou1415926:
This PR contains only logic for CREATE action for env-level aliases. UPDATE and DELETE will be implemented in a following pull request.


cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,301 @@
...
+ *
+ * @param {Set<String>} aliases the subject alternative names for the certificate.
+ * @param {String} physicalResourceID
+ * @return {String} The ARN of the requested certificate.
+ */
+async function requestCertificate(aliases, physicalResourceID) {
+    const { CertificateArn } =await acm.requestCertificate({
+        DomainName: certificateDomain,
+        IdempotencyToken: physicalResourceID,
+        SubjectAlternativeNames: aliases.size === 0? null: [...aliases],
```
iamhopaul123, line 178: Does it throw an error if we pass in an empty array?

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,301 @@
...
+        Tags: [
+            {
+                Key: "copilot-application",
+                Value: appName,
+            },
+            {
+                Key: "copilot-environment",
+                Value: envName,
+            },
```
iamhopaul123, line 187: We can probably tag with the service name as it is service specific.

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,301 @@
...
+/**
+ * Wait until the validation options are ready
+ *
+ * @param certificateARN
+ * @param {Set<String>} aliases for the service.
+ */
+async function waitForValidationOptionsToBeReady(certificateARN, aliases) {
+    let expectedCount = aliases.size + 1; // Expect one validation option for each alias and the cert domain.
+
+    let attempt; // TODO: This wait loops could be further abstracted.
```
iamhopaul123, line 207: I guess when we add update/delete?

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,301 @@
...
+        $waiter: {
+            delay: DELAY_CERTIFICATE_VALIDATED,
+            maxAttempts: ATTEMPTS_CERTIFICATE_VALIDATED,
+        },
+        CertificateArn: certificateARN,
+    }).promise();
+}
+
+/**
+ * Upsert the validation record for the alias, as well as adding the A record if the alias is not the default certificaite domain.
```
iamhopaul123: nit: typo

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,301 @@
...
+            HostedZoneId: envHostedZoneID,
+            MaxItems: "1",
+            StartRecordName: alias,
+        }).promise();
+
+        let recordSet = data["ResourceRecordSets"];
+        if (!recordSet || recordSet.length === 0) {
+            continue; // The alias record does not exist in the hosted zone, hence valid.
+        }
+        if (recordSet[0].AliasTarget && recordSet[0].AliasTarget.DNSName === `${loadBalancerDNS}.`) {
```
iamhopaul123:
```suggestion
        if (recordSet[0].AliasTarget？.DNSName === `${loadBalancerDNS}.`) {
```

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,301 @@
...
+            continue; // The record is an alias record and is in use by myself, hence valid.
+        }
+        throw new Error(`alias ${alias} is in use`);
```
iamhopaul123: Maybe alias is in used by xxx

iamhopaul123 (commented):
LGTM overall! Just a reminder even for create we still need to handle the case when alias is in app/root hosted zone.

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: `SubjectAlternativeNames` needs to either be specified with an array with length >= 1, or not specified at all (by not including the key, or having `null` as its value). So it'll throw an error if we pass in an empty array, but it won't if we pass in a `null`.

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Yes in a following PR - especially because DELETE also needs to have a wait loop like this. I think it may improve code's readability if this can be abstracted.

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: I'm not entirely sure if we should include the record value or alias target in the error message 🤔 If the custom domain is in use, it's possible that it's in use by another service (hence the alias target is a copilot LB's DNS); however, it's also possible that the record is a non-alias record, pointing to whatever address.

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: It's an ECMA 2020 syntax (by default we are using ECMA 5 I think). I tried enabling the ECMA 2020 parser but it still fails to acknowledge the optional chaining syntax. I will use this syntax after I figure out how to really enable ECMA 2020 :)

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,311 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+const AWS = require('aws-sdk');
+const ATTEMPTS_VALIDATION_OPTIONS_READY = 10;
+const ATTEMPTS_RECORD_SETS_CHANGE = 10;
+const DELAY_RECORD_SETS_CHANGE = 30;
+const ATTEMPTS_CERTIFICATE_VALIDATED = 19;
+const DELAY_CERTIFICATE_VALIDATED = 30;
```
efekarakus: Can we add the unit in the const name so that it's clear to someone that would change these values the ramifications for example:
```
const DELAY_RECORD_SETS_CHANGE_IN_S = 30;
```

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,311 @@
...
+
+exports.handler = async function (event, context) {
+    const props = event.ResourceProperties;
+
+    let {LoadBalancerDNS: loadBalancerDNS,
+        LoadBalancerHostedZoneID: loadBalancerHostedZoneID,
+        ServiceName: serviceName,
+        DomainName: domainName,
+    } = props;
+    let aliases = new Set(props.Aliases);
```
efekarakus: nit: can these be const?

cf-custom-resources/test/nlb-cert-validator-uptator-test.js
```diff
@@ -0,0 +1,311 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
huanjani: Can we fix the typo in the file name? 😄

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,311 @@
...
+
+    acm = new AWS.ACM();
+    envRoute53 = new AWS.Route53();
+    envHostedZoneID = props.EnvHostedZoneId;
+    envName = props.EnvName;
+    appName = props.AppName;
+    certificateDomain = `${serviceName}-nlb.${envName}.${appName}.${domainName}`;
+
+    let aliasesSorted = [...aliases].sort().join(",");
+    const physicalResourceID = `/${serviceName}/${aliasesSorted}`;
```
efekarakus, line 102: 1. Do you know if there is a limit on the length of the physical resource ID? Would you mind investigating that, I tried a quick search but couldn't find anything might be worthwhile experimenting.

2. Using the aliases in the resource ID makes total sense because we want to delete the old certificate right if the alias changes? if so can we add that as a short comment here?

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,311 @@
...
+            StartRecordName: alias,
+        }).promise().then((data) => {
+            let recordSet = data["ResourceRecordSets"];
+            if (!recordSet || recordSet.length === 0) {
+                return;
+            }
+            if (recordSet[0].AliasTarget && recordSet[0].AliasTarget.DNSName === `${loadBalancerDNS}.`) {
+                return; // The record is an alias record and is in use by myself, hence valid.
+            }
+            throw new Error(`alias ${alias} is in use`);
```
efekarakus: can we also add additional info for which LB uses it instead of the service's? something like:
```suggestion
            throw new Error(`Alias '${alias}' is already in use by another load balancer '${recordSet[0].AliasTarget.DNSName}'`);
```

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,311 @@
...
+/**
+ * Requests a public certificate from AWS Certificate Manager, using DNS validation.
+ *
+ * @param {Set<String>} aliases the subject alternative names for the certificate.
+ * @param {String} physicalResourceID
+ * @param {String} serviceName
+ * @return {String} The ARN of the requested certificate.
+ */
+async function requestCertificate(aliases, physicalResourceID, serviceName) {
+    const { CertificateArn } =await acm.requestCertificate({
```
```suggestion
    const { CertificateArn } = await acm.requestCertificate({
```

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,311 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+const AWS = require('aws-sdk');
+const ATTEMPTS_VALIDATION_OPTIONS_READY = 10;
+const ATTEMPTS_RECORD_SETS_CHANGE = 10;
+const DELAY_RECORD_SETS_CHANGE = 30;
+const ATTEMPTS_CERTIFICATE_VALIDATED = 19;
+const DELAY_CERTIFICATE_VALIDATED = 30;
+
+let acm, envRoute53, envHostedZoneID, appName, envName, certificateDomain;
```
efekarakus: should we add serviceName here too ? since we pass it down to other functions

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,311 @@
...
+    certificateDomain = `${serviceName}-nlb.${envName}.${appName}.${domainName}`;
+
+    let aliasesSorted = [...aliases].sort().join(",");
+    const physicalResourceID = `/${serviceName}/${aliasesSorted}`;
+
+    let handler = async function() {
+        switch (event.RequestType) {
+            case "Create":
+                await validateAliases(aliases, loadBalancerDNS);
+                const certificateARN = await requestCertificate(aliases, physicalResourceID, serviceName);
```
efekarakus: nit: I wonder if we should pass an object instead to make it more explicit, at first I wasn't clear why the physicalResourceId or svcName had to be passed down until I looked to the body of the function:
```js
await requestCertificate({aliases, idempotencyToken: physicalResourceID});
```

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,311 @@
...
+ * @param {String} loadBalancerDNS
+ * @param {String} loadBalancerHostedZone
+ */
+async function activate(validationOptions, certificateARN, loadBalancerDNS, loadBalancerHostedZone) {
+    let promises = [];
+    for (let option of validationOptions) {
+        promises.push(activateOption(option, loadBalancerDNS, loadBalancerHostedZone));
+    }
+    await Promise.all(promises);
+    console.log("finished upserting records");
```
efekarakus: nit: let's remove this

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,311 @@
...
+                break;
+            case "Update":
+            case "Delete":
+            default:
+                throw new Error(`Unsupported request type ${event.RequestType}`);
+        }
+    };
+
+    try {
+        await handler();
```
efekarakus: can we `Promise.race` the handler with a 13-14 min timeout so that we avoid the scenario where the custom resource is stuck for 3h:

> **Set reasonable timeout periods, and report when they're about to be exceeded**
> If an operation doesn't run within its defined timeout period, the function raises an exception and no response is sent to AWS CloudFormation.
>  https://aws.amazon.com/premiumsupport/knowledge-center/best-practices-custom-cf-lambda/

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,311 @@
...
+            TTL: 60,
+            ResourceRecords: [
+                {
+                    Value: option.ResourceRecord.Value,
+                },
+            ],
+        }
+    }];
+
+    if (option.DomainName !== certificateDomain) {
```
efekarakus, line 279: For me to understand the domain name is indeed updated when any `alias` is set right?

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,311 @@
...
+    envName = props.EnvName;
+    appName = props.AppName;
+    certificateDomain = `${serviceName}-nlb.${envName}.${appName}.${domainName}`;
+
+    let aliasesSorted = [...aliases].sort().join(",");
+    const physicalResourceID = `/${serviceName}/${aliasesSorted}`;
+
+    let handler = async function() {
+        switch (event.RequestType) {
+            case "Create":
```
efekarakus, line 106: I love the way you decomposed the body here! It's so much easier to understand the steps to associate a domain to a lb 🤩

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Yes. I was planning to add the lambda time out later since it introduces more testing logic into this PR, but I can do it now

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: To answer 2.:
Yes. For each unique combination of `[service name + alias names]`, the [certificate ('s subject alternative name) + validation records + A records] needed should be the same. If there is no change in `[service name + alias names]`, then the certificate shouldn't change, neither do the validation records and the A records; on the contrary, if a change is detected in `[service name + alias names]`, then a `replacement` surely needs to happen to at least one of the three resources.

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: I will update the error message to indicate that the alias "could" be another load balancer if the alias target isn't null.

It is still possible that the domain is in use because people manually added a, e.g. CNAME, record using the domain routing to an IP address or another domain (not necessarily an alias to another service).

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: As long as we don't allow changing the service name in the future!

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Yes the `option.DomainName` would be the alias domain name. 

The `option` is given by `acm` when it provides us with the validation options for each aliases. Each option should contain at least the following information:
1. The domain it's validated (i.e. the alias)
2. The validation record name
3. The validation record value

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Talked offline but I'm posting here so that this info is shared:
Answering 1.: There is virtually no limit on Physical Resource ID length - tried a 800-character long ID, and it was fine. However, if the ID is too long, the response object sent from the lambda will be too big, causing the custom resource to fail.

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Added!

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,333 @@
...
+                await validateAliases(aliases, loadBalancerDNS);
+                const certificateARN = await requestCertificate({
+                    aliases: aliases,
+                    idempotencyToken: CRYPTO
+                        .createHash("sha256")
+                        .update(physicalResourceID)
+                        .digest("hex")
+                        .substr(0, 32),});
```
efekarakus: It probably doesn't really matter since the `idempotencyToken` expires after 1h, but I don't know if it's safe enough to `subtr(0,32)` instead we can do `md5` which gives us exactly 32 digit long hashes and we're not doing anything security related so it should be fine:

```suggestion
                    idempotencyToken: CRYPTO
                        .createHash("md5")
                        .update(physicalResourceID)
                        .digest("hex")});
```

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,333 @@
...
+            }
+            let aliasTarget = recordSet[0].AliasTarget;
+            if (aliasTarget && aliasTarget.DNSName === `${loadBalancerDNS}.`) {
+                return; // The record is an alias record and is in use by myself, hence valid.
+            }
+
+            if (aliasTarget) {
+                throw new Error(`Alias ${alias} is in use by ${aliasTarget.DNSName}. This could be another load balancer of a different service.`);
+            }
+            throw new Error(`Alias ${alias} is in use`);
```
```suggestion
                throw new Error(`Alias ${alias} is already in use by ${aliasTarget.DNSName}. This could be another load balancer of a different service.`);
            }
            throw new Error(`Alias ${alias} is already in use`);
```

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -0,0 +1,333 @@
...
+}
+
+/**
+ * Requests a public certificate from AWS Certificate Manager, using DNS validation.
+ *
+ * @param {Object} requestCertificateInput is the input to requestCertificate, containing the alias and idempotencyToken.
+ * @return {String} The ARN of the requested certificate.
+ */
+async function requestCertificate(requestCertificateInput) {
+    let { aliases, idempotencyToken } = requestCertificateInput
```
```suggestion
async function requestCertificate( { aliases, idempotencyToken }) {
```

efekarakus (approved):
LGTM :shipit: after addressing feedback below

huanjani (approved):

PR #3070, opened 2021-11-23
chore(custom-resource): extend alias to app-level and domain-level for NLB

Lou1415926:
The previous PR #3057 takes into consideration only the case where the aliases are environment-level (e.g. `a.env.app.domain.com`); however, they could also be app-level (e.g. `a.app.domain.com`) or root-level (e.g. `a.domain.com`). This PR extends the cases considered as such.

In addition, the PR adds lazy-loading into the script.

Previous PR: #3057 

### DIFF
--- a/cf-custom-resources/lib/nlb-cert-validator-updater.js
+++ b/cf-custom-resources/lib/nlb-cert-validator-updater.js
@@ -85,7 +85,7 @@ const rootHostedZoneIDContext = () => {
     };
 }
 
-let hostecZoneID = {
+let hostedZoneID = {
     app: appHostedZoneIDContext(),
     root: rootHostedZoneIDContext(),
 }
@@ -239,11 +239,13 @@ async function validateAliases(aliases, loadBalancerDNS) {
             if (!recordSet || recordSet.length === 0) {
                 return;
             }
+            if (recordSet[0].Name !== alias) {
+                return;
+            }
             let aliasTarget = recordSet[0].AliasTarget;
             if (aliasTarget && aliasTarget.DNSName === `${loadBalancerDNS}.`) {
                 return; // The record is an alias record and is in use by myself, hence valid.
             }
-
             if (aliasTarget) {
                 throw new Error(`Alias ${alias} is already in use by ${aliasTarget.DNSName}. This could be another load balancer of a different service.`);
             }
@@ -432,14 +434,14 @@ async function domainResources (alias) {
         return {
             domain: domainTypes.AppDomainZone.domain,
             route53Client: clients.app.route53(),
-            hostedZoneID: await hostecZoneID.app(),
+            hostedZoneID: await hostedZoneID.app(),
         };
     }
     if (domainTypes.RootDomainZone.regex.test(alias)) {
         return {
             domain: domainTypes.RootDomainZone.domain,
             route53Client: clients.root.route53(),
-            hostedZoneID: await hostecZoneID.root(),
+            hostedZoneID: await hostedZoneID.root(),
         };
     }
     throw new Error(`unrecognized domain type for ${alias}`);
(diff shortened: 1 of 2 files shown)


cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -319,6 +324,107 @@ exports.deadlineExpired = function () {
...
+        },
+        AppDomainZone: {
+            regex: new RegExp(`^([^\.]+\.)?${appName}.${domainName}`),
+            domain: `${appName}.${domainName}`,
+        },
+        RootDomainZone: {
+            regex: new RegExp(`^([^\.]+\.)?${domainName}`),
+            domain: `${domainName}`,
+        },
+        OtherDomainZone: {},
```
huanjani: I'm just curious about this domainType. How is it used?

Lou1415926: Good catch! This is not used. I should just delete it!

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -9,7 +9,8 @@ const DELAY_RECORD_SETS_CHANGE_IN_S = 30;
 const ATTEMPTS_CERTIFICATE_VALIDATED = 19;
 const DELAY_CERTIFICATE_VALIDATED_IN_S = 30;
 
-let acm, envRoute53, envHostedZoneID, appName, envName, serviceName, certificateDomain;
+let acm, appRoute53, envRoute53, rootHostedZoneID, appHostedZoneID, envHostedZoneID;
```
efekarakus: what do you think of something like this? Does it make testing difficult?

```js
const clients = {
  app: {
     route53: ( () => { 
       let client;
       return () => {
          if (client) { return client }
          client = new AWS.Route53({ ...});
          return client;
        }
    })()
  },
  
  acm: (() => { let client; ... })(),
}
```
This way throughout the code we can do `client.app.route53()` or `client.acm()` everything is guaranteed to be a singleton and it removes all these global clients

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -141,8 +145,9 @@ async function validateAliases(aliases, loadBalancerDNS) {
     let promises = [];
 
     for (let alias of aliases) {
-        const promise = envRoute53.listResourceRecordSets({
-            HostedZoneId: envHostedZoneID,
+        let r = await domainResources(alias);
```
efekarakus:
```suggestion
        let {hostedZoneId, route53, domain} = await domainResources(alias);
```

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Good idea - updated!

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -9,13 +9,87 @@ const DELAY_RECORD_SETS_CHANGE_IN_S = 30;
...
+    let id;
+    return async () => {
+        if (!id) {
+            id = await hostedZoneIDByName(`${domainName}`);
+        }
+        return id
+    };
+}
+
+let hostecZoneID = {
```
efekarakus: 🤣 
```suggestion
let hostedZoneID = {
```

efekarakus (approved):
small typo and :shipit:

huanjani (approved):
Pretty nifty!

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -239,11 +239,13 @@ async function validateAliases(aliases, loadBalancerDNS) {
             if (!recordSet || recordSet.length === 0) {
                 return;
             }
+            if (recordSet[0].Name !== alias) {
```
efekarakus, line 242: 👍

PR #3073, opened 2021-11-25
docs: add more detail to publish partial

huanjani:


site/content/docs/include/publish.en.md
```diff
@@ -9,7 +9,7 @@ publish:
     - name: order-events
 ```
 
-In the example above, this manifest declares an SNS topic named `order-events` that other worker services which are deployed to the Copilot environment can subscribe to.
+In the example above, this manifest declares an SNS topic named `order-events` that other worker services which are deployed to the Copilot environment can subscribe to. An environment variable named `COPILOT_SNS_TOPIC_ARNS` is injected into your workload as a JSON string. For this example, you could write: `const {order-events} = JSON.parse(process.env.COPILOT_SNS_TOPIC_ARNS)` in JavaScript. For more detail, see the [pub/sub](../developing/publish-subscribe.en.md) page.
```
iamhopaul123: How about we make `const {order-events} = JSON.parse(process.env.COPILOT_SNS_TOPIC_ARNS)` a code block instead of inline? Just feel like it would be more readable...

efekarakus: Does this actually work `order-events` can you have a dash in the variable name? 😮 

Maybe we can also add a short line with a generic format too:
```json
{
  "firstTopicName": "arn:aws:sns:us-east-2:123456789012:firstTopic",
  "secondTopicName": "arn:aws:sns:us-east-2:123456789012:secondTopic",
}
```

huanjani: "Topic names must include only uppercase and lowercase ASCII letters, numbers, underscores, and hyphens, and must be between 1 and 256 characters long." from https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sns-topic.html and "Topic names are limited to 256 characters. Alphanumeric characters plus hyphens (-) and underscores (_) are allowed." from https://aws.amazon.com/sns/faqs/

efekarakus: sorry my comment wasn't clear! Hope the screenshot below makes it clearer:

Screen Shot 2021-12-01 at 8 44 13 AM

`order-events` as a SNS topic name is fine but users can't write the JS code above without an error since dashes in variable names aren't allowed. So my suggestion was to pick a sample topic name that doesn't have a dash or do:
```js
const topic = JSON.parse(process.env.COPILOT_SNS_TOPIC_ARNS)['order-events']
```

efekarakus (approved):

site/content/docs/include/publish.en.md
```diff
@@ -6,10 +6,16 @@ The `publish` section allows services to publish messages to one or more SNS top
...
   topics:
-    - name: order-events
+    - name: orderEvents
 ```
 
-In the example above, this manifest declares an SNS topic named `order-events` that other worker services which are deployed to the Copilot environment can subscribe to.
+In the example above, this manifest declares an SNS topic named `orderEvents` that other worker services deployed to the Copilot environment can subscribe to. An environment variable named `COPILOT_SNS_TOPIC_ARNS` is injected into your workload as a JSON string.
+
+In JavaScript, you could write:
+```yaml
```
iamhopaul123:
```suggestion
In JavaScript, you could write:
```js
```

iamhopaul123 (approved):

PR #3080, opened 2021-11-29
chore(pipeline): exit build stage if manifest validation failed

### LINKED ISSUE #3072: buildspec-- handle manifest validation failures
Currently, if a workload manifest fails the validation step, the cause for the build failure isn't immediately clear. We should try to fail faster or surface the error better.

huanjani:
Fixes #3072.

If a workload manifest fails, the pipeline build stage will fail, too, but only after many more steps, logs, and time. The validation error gets buried, and it isn't obvious that it's the cause of the failure.

This change checks if the `params.json` file that would be generated upon successful manifest validation is empty; if it is, the build exits with a suggestion to look for a manifest validation error message. There are, of course, still logs to wade through, but this should make the point of failure easier to pinpoint.

Manually tested with invalid service manifest (ARM + Spot), valid service manifest, invalid job manifest (bad platform), valid job manifest.

### DIFF
--- a/internal/pkg/template/templates/cicd/buildspec.yml
+++ b/internal/pkg/template/templates/cicd/buildspec.yml
@@ -51,23 +51,21 @@ phases:
       # The tag is the build ID but we replaced the colon ':' with a dash '-'.
       # We truncate the tag (from the front) to 128 characters, the limit for Docker tags
       # (https://docs.docker.com/engine/reference/commandline/tag/)
-      # Check if the params file is empty. If so, echo error msg and exit.
+      # Check if the `svc package` commanded exited with a non-zero status. If so, echo error msg and exit.
       - >
         for env in $pl_envs; do
           tag=$(sed 's/:/-/g' <<<"${CODEBUILD_BUILD_ID##*:}-${env}" | rev | cut -c 1-128 | rev)
           for svc in $svcs; do
           ./copilot-linux svc package -n $svc -e $env --output-dir './infrastructure' --tag $tag;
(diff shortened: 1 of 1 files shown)


internal/pkg/template/templates/cicd/buildspec.yml
```diff
@@ -51,14 +51,25 @@ phases:
       # The tag is the build ID but we replaced the colon ':' with a dash '-'.
       # We truncate the tag (from the front) to 128 characters, the limit for Docker tags
       # (https://docs.docker.com/engine/reference/commandline/tag/)
+      # Check if the params file is empty. If so, echo error msg and exit.
       - >
         for env in $pl_envs; do
           tag=$(sed 's/:/-/g' <<<"${CODEBUILD_BUILD_ID##*:}-${env}" | rev | cut -c 1-128 | rev)
           for svc in $svcs; do
           ./copilot-linux svc package -n $svc -e $env --output-dir './infrastructure' --tag $tag;
+          [ -s ./infrastructure/$svc-$env.params.json ]
```
iamhopaul123: What does this mean 🤔

Lou1415926 (approved):
Awesome!

internal/pkg/template/templates/cicd/buildspec.yml
huanjani: It checks if a file is empty or not.
I've removed it, as we can just check the exit code from the `svc package` command.

iamhopaul123 (approved):

PR #3081, opened 2021-11-30
chore: recommended action when domain registered with the app is not in Route53

Lou1415926:
Nowadays people can associate a domain with their application with `copilot app init --domain example.com`. At this moment,  it is required that the domain is registered with Route53 in order for future operation to work (e.g. enabling HTTPS). However, this information is not well surfaced right now.

This PR provides a better recommended action when it is detected that the domain isn't registered with Route53.


internal/pkg/aws/route53/route53domains.go
```diff
@@ -0,0 +1,52 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
iamhopaul123: Can we rename this file to be just `domain.go` or `domains.go`?

internal/pkg/aws/route53/route53domains.go
```diff
@@ -0,0 +1,52 @@
...
+	_, err := r.client.GetDomainDetail(&route53domains.GetDomainDetailInput{
+		DomainName: aws.String(domainName),
+	})
+	if err == nil {
+		return nil
+	}
+	var errInvalidInput *route53domains.InvalidInput
+	if !errors.As(err, &errInvalidInput) {
+		return fmt.Errorf("get domain detail: %w", err)
+	}
```
iamhopaul123: Maybe use strings.Contain for simplicity?

internal/pkg/cli/app_init.go
```diff
@@ -226,6 +233,19 @@ func (o *initAppOpts) validateAppName(name string) error {
...
 
+func (o *initAppOpts) isDomainOwned() error {
+	err := o.domainInfoGetter.IsDomainOwned(o.domainName)
+	if err == nil {
+		return nil
+	}
+	var errDomainNotFound *route53.ErrDomainNotFound
+	if errors.As(err, &errDomainNotFound) {
+		o.isDomainNotFoundInAccount = true
+		return nil
```
iamhopaul123: I wonder if we should error out and log out here instead of letting it go, since the user wants to create an app with domain but we in fact create an app without domain for them

internal/pkg/aws/route53/errors.go
```diff
@@ -3,7 +3,19 @@
...
 package route53
 
-import "errors"
+import (
+	"errors"
+	"fmt"
+)
 
 // ErrNoDefaultCluster occurs when the default cluster is not found.
 var ErrDomainNotExist = errors.New("domain does not exist")
```
efekarakus: What's the difference with this error? Can we delete this one now? (also its comment is outdated 🙈 )

internal/pkg/cli/app_init.go
efekarakus: Yeah, I agree w boss. We should err here and log the recommended action here 🙏

internal/pkg/aws/route53/errors.go
Lou1415926: This error is used when the domain hosted zone isn't found. Previously we used this to identify whether the domain lives in the account, because when a domain is registered, the domain hosted zone is automatically created.

Maybe I should change this to "domain hosted zone not found"?

efekarakus: ah okay that makes sense! yea I think your suggestion sounds good to me :)

internal/pkg/cli/app_init.go
```diff
@@ -52,8 +52,7 @@ type initAppOpts struct {
 	prog                 progress
 	isSessionFromEnvVars func() (bool, error)
 
-	cachedHostedZoneID        string
-	isDomainNotFoundInAccount bool
+	cachedHostedZoneID string
```
efekarakus: Looks like we can now delete this field 😎

internal/pkg/cli/app_init.go
```diff
@@ -240,8 +239,10 @@ func (o *initAppOpts) isDomainOwned() error {
 	}
 	var errDomainNotFound *route53.ErrDomainNotFound
 	if errors.As(err, &errDomainNotFound) {
-		o.isDomainNotFoundInAccount = true
-		return nil
+		log.Errorln("The account does not seem to own the domain that you entered. Please make sure that the " +
```
efekarakus: nit: can we ensure that the text stays within 80 characters as much as possible so that it renders nicely for the default terminal screen size? https://github.com/aws/copilot-cli/blob/mainline/STYLE_GUIDE.md#screen-sizes 
```go
log.Errorf(`The account does not seem to own the domain that you entered.
Please make sure that %s is registered with Route53 in your account.
To transfer domain registration to Route53, see:
https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/domain-transfer-to-route-53.html
`, o.domainName)

```

efekarakus (approved):
Looks great, just a small change and then feel free to remove the label!

iamhopaul123 (approved):

PR #3086, opened 2021-12-02
chore(describe): add platform info to `svc show` output

huanjani:
Sample outputs:
```
About

  Application       buildspec
  Name              worker
  Type              Worker Service
  Platform          LINUX/ARM64

Configurations
```
```
About

  Application       buildspec
  Name              empty-platform-field
  Type              Backend Service
  Platform          LINUX/X86_64

Configurations
```
```
{"service":"fe","type":"Load Balanced Web Service","platform":"LINUX/X86_64","application":"buildspec","configurations":[{"environment":"test","port":"80","cpu":"256","memory":"512","tasks":"1"},{"environment":"sing","port":"80","cpu":"256","memory":"512","tasks":"1"}],...}
```


efekarakus:
I haven't looked at the PR but looks cool!  
Is there a reason why you preferred to put the `"Platform"` in the `"About"` section instead of `"Configurations"`? Since it's possible with env overrides to specify a different platform per environment although it's unlikely anyone does that.

huanjani:
> 

Great question. I followed our original design doc because that's what was agreed upon, but I like the idea of moving it to Configurations. I can imagine circumstances in which folks might want different platforms for different envs! This should be an easy change bc it's already being called for each env.

efekarakus:
> I like the idea of moving it to Configurations.

Sounds good thank you!! Also feel free to change `minCellWidth` from 20 characters to a smaller number like 10 (whatever looks good) to make the table more compact

huanjani:
updated:
```
About

  Application  buildspec
  Name         with-platform-args
  Type         Backend Service

Configurations

  Environment  Tasks     CPU (vCPU)  Memory (MiB)  Platform      Port
  -----------  -----     ----------  ------------  --------      ----
  test         1         0.25        512           LINUX/X86_64  80
  sing           "         "           "           LINUX/ARM64     "
```

internal/pkg/describe/service.go
```diff
@@ -202,6 +204,22 @@ func NewServiceDescriber(opt NewServiceConfig) (*ServiceDescriber, error) {
 	}, nil
 }
 
+// Platform returns the platform of the task definition. If it is nil, fill in default values.
+func (d *ServiceDescriber) Platform() (*awsecs.ContainerPlatform, error) {
+	taskDefinition, err := d.ecsClient.TaskDefinition(d.app, d.env, d.service)
+	if err != nil {
+		return nil, fmt.Errorf("describe task definition for service %s: %w", d.service, err)
+	}
+	platform := taskDefinition.Platform()
```
Lou1415926: Would the sdk return an empty `platform` if it is `linux/x86_64`? Do you know at what conditions would `platform == nil` happen?

I am not entirely sure about setting a default value if it is `nil`; are we sure if it is `nil`, then the true running platform would be `linux/x86_64`?

internal/pkg/describe/backend_service.go
```diff
@@ -79,6 +80,7 @@ func (d *BackendServiceDescriber) Describe() (HumanJSONStringer, error) {
 
 	var configs []*ECSServiceConfig
 	var services []*ServiceDiscovery
+	var platform string
```
efekarakus, line 83: nit: can we move the declaration of the variable to inside the for-loop?

```go
platform := dockerengine....
```

internal/pkg/describe/rd_web_service.go
```diff
@@ -86,6 +86,7 @@ func (d *RDWebServiceDescriber) Describe() (HumanJSONStringer, error) {
 			Port:        service.Port,
 			CPU:         service.CPU,
 			Memory:      service.Memory,
+			Platform:    "LINUX/X86_64",
```
Lou1415926: An alternative to defaulting a platform string would be to not show the column at all if we do not have the information 🤔  

For RD web service, I feel like user wouldn't be interested in the platform information anyway since they should have known that the only valid platform is "linux/x86_64" if they have tried to use any other platform for the service.

internal/pkg/describe/backend_service.go
```diff
@@ -103,19 +105,25 @@ func (d *BackendServiceDescriber) Describe() (HumanJSONStringer, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("retrieve environment variables: %w", err)
+		}
 		configs = append(configs, &ECSServiceConfig{
 			ServiceConfig: &ServiceConfig{
 				Environment: env,
 				Port:        port,
 				CPU:         svcParams[cfnstack.WorkloadTaskCPUParamKey],
 				Memory:      svcParams[cfnstack.WorkloadTaskMemoryParamKey],
+				Platform:    platform,
```
efekarakus: nit: can we remove the variable completely and have the platform inlined like this:

```suggestion
				Platform:    dockerengine.PlatformString(containerPlatform.OperatingSystem, containerPlatform.Architecture),
```

internal/pkg/describe/pipeline_show_test.go
```diff
@@ -205,19 +205,19 @@ func TestPipelineDescriber_String(t *testing.T) {
...
 
 Stages
 
-  Name              Category            Provider            Details
-  ----              --------            --------            -------
-  Source            Source              GitHub              Repository: badgoose/repo
-  Build             Build               CodeBuild           BuildProject: pipeline-dinder-badgoose-repo-BuildProject
-  DeployTo-test     Deploy              CloudFormation      StackName: dinder-test-test
+  Name           Category  Provider        Details
+  ----           --------  --------        -------
```
efekarakus, line 217: :homer-disappearing: sorry for all the min width changes 😂

internal/pkg/describe/service.go
```diff
@@ -202,6 +204,22 @@ func NewServiceDescriber(opt NewServiceConfig) (*ServiceDescriber, error) {
 	}, nil
 }
 
+// Platform returns the platform of the task definition. If it is nil, fill in default values.
+func (d *ServiceDescriber) Platform() (*awsecs.ContainerPlatform, error) {
```
efekarakus: What is the difference between a `ServiceDescriber` and `ECSServiceDescriber` I find it so confusing 😖 

Do we need the method on both types?
Is there a way of renaming the types to bring clarity for the difference between the two?

efekarakus (commented):
Looks good! I'm just confused with what's going on with our structs :)

internal/pkg/describe/backend_service_test.go
```diff
@@ -355,37 +394,37 @@ func TestBackendSvcDesc_String(t *testing.T) {
...
 
 Configurations
 
-  Environment       Tasks               CPU (vCPU)          Memory (MiB)        Port
-  -----------       -----               ----------          ------------        ----
-  test              1                   0.25                512                 80
-  prod              3                   0.5                 1024                5000
+  Environment  Tasks     CPU (vCPU)  Memory (MiB)  Platform      Port
+  -----------  -----     ----------  ------------  --------      ----
+  test         1         0.25        512           LINUX/X86_64  80
```
iamhopaul123, line 404: Awesome!

internal/pkg/describe/service.go
huanjani: Done. Previously, `ServiceDescriber` was for App Runner workloads, and `ECSServiceDescriber` was for other types. I know have a "base" describer for shared elements and `ECSServiceDescriber` and `AppRunnerServiceDescriber`. Changed some other names because there were so many "service"s and "describer"s but I'm not wed to the specific names, so welcome suggested changes. 
Hopefully this refactor make the code clearer.

internal/pkg/describe/backend_service.go
huanjani: Yup. Done.

internal/pkg/describe/rd_web_service.go
huanjani: I have removed the `Platform` column for RDWS because you're right-- it's the only option, so it wouldn't be useful.

internal/pkg/describe/service.go
huanjani: Yes, the SDK doesn't return that field if it's `LINUX/X86_64`. 
I think it's okay to set the default value knowing that this is the case. 🤔  Maybe there's some way to monitor if this ever changes.
I can imagine scenarios in which users would want to know/confirm that their container was built on the default platform, like if they're sometimes using WSL or changing the platform in their manifest or working on a team with different platforms.

internal/pkg/describe/service.go
```diff
@@ -82,20 +81,21 @@ type apprunnerClient interface {
...
+	EnvVars() ([]*awsecs.ContainerEnvVar, error)
+	Secrets() ([]*awsecs.ContainerSecret, error)
 	ServiceStackResources() ([]*stack.Resource, error)
-	Service() (*apprunner.Service, error)
-	ServiceARN() (string, error)
-	ServiceURL() (string, error)
 }
 
-type ecsStackDescriber interface {
+type apprunnerStackDescriber interface {
```
efekarakus: Since the `Service`, `ServiceURL` methods use the `appRunnerClient` should we rename the interface?
```suggestion
type appRunnerDescriber interface 
```

internal/pkg/describe/service.go
```diff
@@ -82,20 +81,21 @@ type apprunnerClient interface {
 	DescribeService(svcArn string) (*apprunner.Service, error)
 }
 
-type apprunnerSvcDescriber interface {
+type ecsDescriber interface {
 	Params() (map[string]string, error)
+	Outputs() (map[string]string, error)
```
efekarakus, line 86: What do you think of creating this additional interface:
```go
// workloadStackDescriber is the interface to retrieve information from a deployed workload stack.
type workloadStackDescriber interface {
	Params() (map[string]string, error)
	Outputs() (map[string]string, error)
	ServiceStackResources() ([]*stack.Resource, error)
}
```
then we can do:
```go
// ecsDescriber is the interface to retrieve information from an ecs service.
// The ecsDescriber is composed of a workload stack describer and ecs client methods.
type ecsDescriber interface {
	workloadStackDescriber // Methods to retrieve generic configuration.
  
  	// Methods to retrieve ECS specific configuration.
	Platform() (*awsecs.ContainerPlatform, error)
	EnvVars() ([]*awsecs.ContainerEnvVar, error)
	Secrets() ([]*awsecs.ContainerSecret, error)
}

// similar comments here
type appRunnerDescriber {
	workloadStackDescriber
	
	Service() (*apprunner.Service, error)
	ServiceARN() (string, error)
	ServiceURL() (string, error)
}
```

internal/pkg/describe/service.go
```diff
@@ -191,59 +194,43 @@ func NewServiceDescriber(opt NewServiceConfig) (*ServiceDescriber, error) {
...
-	}
-	return taskDefinition.EnvironmentVariables(), nil
-}
-
-// Secrets returns the secrets of the task definition.
-func (d *ServiceDescriber) Secrets() ([]*awsecs.ContainerSecret, error) {
-	taskDefinition, err := d.ecsClient.TaskDefinition(d.app, d.env, d.service)
+// NewECSServiceDescriber instantiates a new non-App Runner service.
+func NewECSServiceDescriber(opt NewServiceConfig) (*ECSServiceDescriber, error) {
+	serviceDescriber, err := NewServiceStackDescriber(opt)
```
efekarakus: nit:
```suggestion
	stackDescriber, err := NewServiceStackDescriber(opt)
```

internal/pkg/describe/service.go
```diff
@@ -182,7 +184,8 @@ type NewServiceConfig struct {
 	DeployStore     DeployedEnvServicesLister
 }
 
-func NewServiceDescriber(opt NewServiceConfig) (*ServiceDescriber, error) {
+// NewServiceStackDescriber instantiates the core elements of a new service.
+func NewServiceStackDescriber(opt NewServiceConfig) (*serviceStackDescriber, error) {
```
efekarakus: nit: can this be private?
```suggestion
func newServiceStackDescriber(opt NewServiceConfig) (*serviceStackDescriber, error) {
```

efekarakus (approved):
omg looks so much better :coolcrying:

Some super tiny refactoring suggestions, feel free to remove the label after!

internal/pkg/cli/env_show_test.go
```diff
@@ -293,27 +293,27 @@ func TestEnvShow_Execute(t *testing.T) {
...
 Tags
 
   Key                  Value
   ---                  -----
   copilot-application  testApp
   copilot-environment  testEnv
-  key1              value1
-  key2              value2
+  key1    value1
+  key2    value2
```
Lou1415926: it's such a pain adjusting the column width :cool-crying:... I wonder if there is a way for tabwriter to dynamically adjust the width based on max width of a column...

Lou1415926 (approved):

Issue #1366, opened 2020-09-08
Support environment variable files in manifests


efekarakus:
### Ask
Enable passing multiple environment variables using external files.
```yaml
variables:
  LOG_LEVEL: info

env_file:
- <path from root of workspace>
```

See: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/taskdef-envfiles.html

From a customer ask:
> Env files are nice because they're pretty flexible/standardized -- whether you're writing a non-Docker service or a Dockerized service, you can (generally) take the same env files and re-use it across your architecture. So feels like it would really help to round-out the integration story there = )

iamhopaul123:
## Design Proposal
```yaml
variables:
  LOG_LEVEL: info

# env file within workload workspace (same level as the workload manifest), must be suffixed with ".env"
env_file: foo.env
```
## Implementation Plan

- [x] update svc/job deploy and CFN template
- [x] update pipeline buildspec
- [x] update manifest pkg and add manifest validation (this will enable the feature)
- [ ] add to e2e test and Copilot docs

efekarakus:
Hi folks! In the design proposal above from @iamhopaul123 he is reducing the API to a single file:
```yaml
env_file: <path>
```
The above manifest should work with environment overrides:
```
env_file: <test path>
environment:
  prod:
    env_file: <prod path>
```

Would that be sufficient for your use cases? Or do you need multiple file support? if so why?

Lou1415926:
This is now released 🎉 in v1.14.0 https://github.com/aws/copilot-cli/releases/tag/v1.14.0 !

PR #3087, opened 2021-12-02
chore: update cli pkg and CFN template for env file support

iamhopaul123:
Part of #1366

Most of the file changes are some boilerplates and unit/integ tests update.


internal/pkg/cli/interfaces.go
```diff
@@ -235,6 +235,10 @@ type manifestReader interface {
 	ReadWorkloadManifest(name string) (workspace.WorkloadManifest, error)
 }
 
+type envFileReader interface {
+	ReadSvcFile(svc, fname string) ([]byte, error)
```
efekarakus: I feel like we should allow them to place the env file anywhere they want as long as it's in the workspace.

For example, I could see this being a common setup:
```
.
├── copilot
│   └── backend
│       └── manifest.yml
└──backend
   ├── Dockerfile
   ├── docker-compose.yml
   ├── .env
   ├── Dockerfile
   └──server.js
```

In this situation, I'd expect the manifest to be write like this:
```yaml
env_file: backend/.env
```

Let me know what you think! 🤔

internal/pkg/cli/job_deploy.go
```diff
@@ -57,9 +59,14 @@ type deployJobOpts struct {
...
 	prompt  prompter
 
+	// cached variables
 	targetApp         *config.Application
 	targetEnvironment *config.Environment
 	targetJob         *config.Workload
+	appEnvResources   *stack.AppRegionalResources
+	appliedManifest   interface{}
+	addonsURL         string
+	EnvFileARN        string
```
efekarakus: nit: Can we make this field private?

internal/pkg/cli/job_deploy.go
```diff
@@ -57,9 +59,14 @@ type deployJobOpts struct {
 	sel     wsSelector
 	prompt  prompter
 
+	// cached variables
 	targetApp         *config.Application
 	targetEnvironment *config.Environment
 	targetJob         *config.Workload
+	appEnvResources   *stack.AppRegionalResources
```
efekarakus, line 69: nit: This is totally fine, I wonder if we should instead cache the individual resources so that it's a bit clearer what we are storing:
```
artifactBucketName string
ecrRepositoryURLs map[string]string // workload name to regional ECR repo URL.
```

internal/pkg/cli/job_deploy.go
```diff
@@ -130,63 +137,104 @@ func (o *deployJobOpts) Execute() error {
...
 	}
-
 	if err := o.envUpgradeCmd.Execute(); err != nil {
 		return fmt.Errorf(`execute "env upgrade --app %s --name %s": %v`, o.appName, o.targetEnvironment.Name, err)
 	}
-
 	if err := o.configureContainerImage(); err != nil {
 		return err
 	}
+	if err := o.pushToS3Bucket(); err != nil {
```
efekarakus: nit:
```suggestion
	if err := o.pushArtifactsToS3(); err != nil {
```

internal/pkg/cli/job_deploy.go
```diff
@@ -130,63 +137,104 @@ func (o *deployJobOpts) Execute() error {
...
+	}
+	content, err := o.ws.ReadSvcFile(o.name, fileName)
+	if err != nil {
+		return fmt.Errorf("read env file %s: %w", fileName, err)
+	}
+	if err := o.retrieveAppResourcesForEnvRegion(); err != nil {
+		return err
+	}
+	reader := bytes.NewReader(content)
+	url, err := o.s3.PutArtifact(o.appEnvResources.S3Bucket, fileName, reader)
```
efekarakus: I'm trying to think if replacing an existing env file in the S3 bucket could cause unexpected issues for the customer 💭 (I think we were safe with addons because changeset automatically detect changes)

Like for example: 
1. User updates the .env file in the repository but made no code changes
2. Since there is no change, a new container image is not pushed
3. The env file is uploaded and overrides the existing file in the S3 bucket
4. ⚠️  Deployment fails because change set is empty
5. In the mean time, a new task spins up due to autoscaling but pulls the wrong version of the env file.


Looking at the CDK, if I follow correctly they compute the hash of the file and use that as the key for the object:
https://github.com/aws/aws-cdk/blob/dd5e12d43231d991971de3121928377569b2dd63/packages/%40aws-cdk/aws-s3-assets/lib/asset.ts#L150-L151

I think we should do something similar: https://stackoverflow.com/a/45123990/1201381, compute the sha256 hash of the contents and upload that so that we don't end up in the situation of pulling the wrong file

internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-prod.stack.yml
```diff
@@ -209,6 +221,28 @@ Resources:
...
+                - Effect: 'Allow'
+                  Action:
+                    - 's3:GetBucketLocation'
+                  Resource:
+                    - !Join
+                      - ''
+                      - - 'arn:'
+                        - !Ref AWS::Partition
+                        - ':s3:::'
+                        - !Select [0, !Split ['/', !Select [5, !Split [':', !Ref EnvFileARN]]]]
```
efekarakus, line 244: rip lol what a mess

internal/pkg/deploy/cloudformation/stack/workload.go
```diff
@@ -63,6 +64,7 @@ const (
 type RuntimeConfig struct {
 	Image             *ECRImage         // Optional. Image location in an ECR repository.
 	AddonsTemplateURL string            // Optional. S3 object URL for the addons template.
+	EnvFileARN        string            // Optional. S3 object URLs for the env files.
```
efekarakus:
```suggestion
	EnvFileARN        string            // Optional. S3 object ARN for the env file.
```

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -162,6 +162,11 @@ func (s *LoadBalancedWebService) BuildArgs(wsRoot string) *DockerBuildArgs {
 	return s.ImageConfig.Image.BuildConfig(wsRoot)
 }
 
+// EnvFiles returns the env file name in a ws service directory.
+func (s *LoadBalancedWebService) EnvFiles() string {
```
efekarakus:
```suggestion
// EnvFile returns the path in the workspace of the environment file associated with the service.
func (s *LoadBalancedWebService) EnvFile() string {
```

efekarakus (commented):
Awesome feature boss 😎

internal/pkg/cli/job_deploy.go
iamhopaul123: Ooh that's a good catch!

internal/pkg/manifest/lb_web_svc.go
iamhopaul123: this func is not supposed to be included in the PR sorry...

internal/pkg/cli/job_deploy.go
iamhopaul123: ok done

internal/pkg/cli/interfaces.go
```diff
@@ -307,6 +307,7 @@ type wsAddonManager interface {
 
 type artifactUploader interface {
 	PutArtifact(bucket, fileName string, data io.Reader) (string, error)
+	Upload(bucket, key string, data io.Reader) (string, error)
```
efekarakus: 🤔 `PutArtifact` and `Upload` seem so similar. I feel like there is a way of getting rid of `PutArtifact`, what do you think of this:
```go
func (s3 *S3) Upload(r io.Reader,  bucket, key string) (string, error) { ... }

// MkdirTimestamp prefixes the key with the current timestamp "<timestamp>/key".
func MkdirTimestamp(key string) string

// MkdirSHA prefixes the key with the SHA256 hash of the contents of r "<hash>/key".
func MkdirSHA256(r io.Reader, key string) 
```

This way we can even chain keys: `Upload(buf, bucket, s3.MkdirTimestamp(s3.MkdirSHA256(buf, filename)),`

internal/pkg/initialize/workload.go
```diff
@@ -377,11 +377,10 @@ func newWorkerServiceManifest(i *ServiceProps) (*manifest.WorkerService, error)
 
 // relativeDockerfilePath returns the path from the workspace root to the Dockerfile.
 func relativeDockerfilePath(ws Workspace, path string) (string, error) {
-	copilotDirPath, err := ws.CopilotDirPath()
+	wsRoot, err := ws.Path()
```
efekarakus, line 380: nice!

internal/pkg/template/template.go
```diff
@@ -129,29 +124,6 @@ func (t *Template) UploadRequestDrivenWebServiceCustomResources(upload s3.Compre
 	return t.uploadCustomResources(upload, rdWkldCustomResourceFiles)
 }
 
-// UploadRequestDrivenWebServiceLayers uploads already-zipped layers for a request driven web service.
```
efekarakus, line 132: Why did we remove these boss?

efekarakus (commented):
Looks great! just a small refactor request and then :shipit:

internal/pkg/cli/interfaces.go
iamhopaul123: Agreed sounds like a good idea. Will do in the next PR!

iamhopaul123: Or nvm i'll put in this one

internal/pkg/template/template.go
iamhopaul123: double checked with @Lou1415926. they are not in used anymore

internal/pkg/cli/interfaces.go
iamhopaul123: done!

internal/pkg/aws/s3/s3.go
```diff
@@ -101,10 +82,19 @@ func (s *S3) ZipAndUpload(bucket, key string, files ...NamedBinary) (string, err
...
 // Upload uploads a file to an S3 bucket under the specified key.
-func (s *S3) Upload(bucket, key string, file NamedBinary) (string, error) {
-	buf := new(bytes.Buffer)
-	buf.Write(file.Content())
-	return s.upload(bucket, key, buf)
+func (s *S3) Upload(bucket, key string, data io.Reader) (string, error) {
+	return s.upload(bucket, key, data)
+}
+
+// MkdirTimestamp prefixes the key with the current timestamp "<timestamp>/key".
```
efekarakus:
```suggestion
// MkdirTimestamp prefixes the key with the current timestamp "manual/<timestamp>/key".
```

internal/pkg/cli/job_deploy.go
```diff
@@ -41,6 +45,7 @@ type deployJobOpts struct {
 
 	store              store
 	ws                 wsJobDirReader
+	fs                 *afero.Afero
```
efekarakus, line 48: just to make sure do we need to update anything in the `init.go` file since we touched the `deploy` commands?

efekarakus (approved):

internal/pkg/cli/job_deploy.go
iamhopaul123: good call! Added to both `init.go` and `deploy.go`

internal/pkg/aws/s3/s3.go
```diff
@@ -101,10 +82,19 @@ func (s *S3) ZipAndUpload(bucket, key string, files ...NamedBinary) (string, err
...
+	return s.upload(bucket, key, data)
+}
+
+// MkdirTimestamp prefixes the key with the current timestamp "manual/<timestamp>/key".
+func MkdirTimestamp(key string) string {
+	id := time.Now().Unix()
+	return path.Join(artifactDirName, strconv.FormatInt(id, 10), key)
+}
+
+// MkdirSHA prefixes the key with the SHA256 hash of the contents of r "<hash>/key".
```
huanjani: Q: what is "r"?

nit:
```suggestion
// MkdirSHA prefixes the key with the SHA256 hash of the contents of r "manual/<hash>/key".
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -424,62 +431,76 @@ func (o *deploySvcOpts) configureContainerImage() error {
...
+	if err := o.retrieveWorkspacePath(); err != nil {
+		return err
 	}
-	var tags []string
-	if imageTag != "" {
-		tags = append(tags, imageTag)
+	content, err := o.fs.ReadFile(filepath.Join(o.workspacePath, path))
+	if err != nil {
+		return fmt.Errorf("read env file %s: %w", path, err)
 	}
```
huanjani, line 461: What will the guidance for the env file look like? I'm worried that users aren't aware of buildArgs that are fetched under the hood (like `platform`), and won't know they have to put them in their env file.

internal/pkg/aws/s3/s3.go
iamhopaul123: I guess that means the value of <hash> will get replaced? I'll ditch for consistency

internal/pkg/cli/svc_deploy.go
iamhopaul123: It'll be the same as the dockerfile which we'll look in their workspace. I'll try to make that clear in the docs.

> won't know they have to put them in their env file.

What does this mean 🤔

huanjani: Do you have a link to the issue? I've been looking but can't find it. Clarifying the ask will help me! 🙏🏼 😸

iamhopaul123: https://github.com/aws/copilot-cli/issues/1366 ?

huanjani: Ah, thank you. NVM ☝🏼 . I get it now.

huanjani (approved):
🚢

PR #3088, opened 2021-12-02
chore: error out if env init is invoked on an existing environment

Lou1415926:
Currently `copilot env init` twice on the same environment goes through - the operation won't fail. However, only the first invocation counts, meaning that the second invocation does not update the stack, nor the SSM parameter store.

Since `copilot env init` can only create and initialize, not update, an environment, we should error out if an update attempt is made (by running `copilot env init` twice)


internal/pkg/cli/env_init.go
```diff
@@ -542,6 +548,24 @@ func (o *initEnvOpts) askAdjustResources() error {
 	return nil
 }
 
+func (o *initEnvOpts) validateDuplicateEnv() error {
+	envs, err := o.store.ListEnvironments(o.appName)
```
efekarakus: Should we use `o.store.GetEnvironment()` instead? as it should be a faster endpoint

internal/pkg/cli/env_init.go
```diff
@@ -542,6 +548,24 @@ func (o *initEnvOpts) askAdjustResources() error {
...
+	if err != nil {
+		return fmt.Errorf("validate if environment exists: %w", err)
+	}
+
+	for _, env := range envs {
+		if env.Name == o.name {
+			log.Errorf(`It seems like you are trying to init an environment that already exists.
+To re-create the environment, please run %s before running %s.
+`, color.HighlightCode(fmt.Sprintf("copilot env delete --name %s", o.name)), color.HighlightCode(fmt.Sprintf("copilot env init --name %s", o.name)))
```
```suggestion
			log.Errorf(`It looks like you are trying to init an environment that already exists.
To re-create the environment, please run:
1. %s 
2. And then %s
`, color.HighlightCode(fmt.Sprintf("copilot env delete --name %s", o.name)), color.HighlightCode(fmt.Sprintf("copilot env init --name %s", o.name)))
```

efekarakus (commented):
Looks good! just a small suggestion

internal/pkg/cli/env_init.go
```diff
@@ -363,6 +366,9 @@ func (o *initEnvOpts) askEnvName() error {
 		return fmt.Errorf("get environment name: %w", err)
 	}
 	o.name = envName
+	if err := o.validateDuplicateEnv(); err != nil {
+		return err
+	}
 	return nil
```
```suggestion
	return o.validateDuplicateEnv()
```

efekarakus (approved):

internal/pkg/cli/env_init.go
```diff
@@ -542,6 +545,26 @@ func (o *initEnvOpts) askAdjustResources() error {
...
 }
 
+func (o *initEnvOpts) validateDuplicateEnv() error {
+	_, err := o.store.GetEnvironment(o.appName, o.name)
+	if err == nil {
+		log.Errorf(`It seems like you are trying to init an environment that already exists.
+To re-create the environment, please run:
```
huanjani: tiny nit:
```suggestion
To recreate the environment, please run:
```

huanjani (approved):
🚢

PR #3091, opened 2021-12-02
chore: do not error out if TLD isn't supported by Route53

Lou1415926:
Since customer can bring their domain purchased outside of Route53, the TLD isn't necessarily supported by Route53. This skips the error.

### DIFF
--- a/internal/pkg/aws/route53/domain.go
+++ b/internal/pkg/aws/route53/domain.go
@@ -30,8 +30,8 @@ func NewRoute53Domains(s *session.Session) *Route53Domains {
 	}
 }
 
-// IsDomainRegisteredInRoute53 checks if the domain is owned by the account.
-func (r *Route53Domains) IsDomainRegisteredInRoute53(domainName string) error {
+// IsRegisteredDomain checks if the domain is owned by the account.
+func (r *Route53Domains) IsRegisteredDomain(domainName string) error {
 	_, err := r.client.GetDomainDetail(&route53domains.GetDomainDetailInput{
 		DomainName: aws.String(domainName),
 	})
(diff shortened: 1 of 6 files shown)


internal/pkg/aws/route53/domain.go
```diff
@@ -30,22 +30,26 @@ func NewRoute53Domains(s *session.Session) *Route53Domains {
 	}
 }
 
-// IsDomainOwned checks if the domain is owned by the account.
-func (r *Route53Domains) IsDomainOwned(domainName string) error {
+// IsDomainRegisteredInRoute53 checks if the domain is owned by the account.
+func (r *Route53Domains) IsDomainRegisteredInRoute53(domainName string) error {
```
efekarakus: nit: since this is the route53 client we probably don't need to repeat route53 in the method name:
```suggestion
func (r *Route53Domains) IsRegisteredDomain(domainName string) error {
```

internal/pkg/aws/route53/errors.go
```diff
@@ -16,11 +16,11 @@ func (e *ErrDomainHostedZoneNotFound) Error() string {
 	return fmt.Sprintf("hosted zone is not found for domain %s", e.domainName)
 }
 
-// ErrDomainNotFound occurs when the domain is not found in the account.
-type ErrDomainNotFound struct {
+// ErrDomainNotFoundInRoute53 occurs when the domain is not found in the account.
+type ErrDomainNotFoundInRoute53 struct {
```
efekarakus: I preferred the name before because it read `route53.ErrDomainNotFound` vs. `route53.ErrDomainNotFoundInRoute53` (see: https://github.com/golang/go/wiki/CodeReviewComments#package-names)

efekarakus (approved):

internal/pkg/aws/route53/domain.go
```diff
@@ -30,22 +30,26 @@ func NewRoute53Domains(s *session.Session) *Route53Domains {
...
 	}
-	var errInvalidInput *route53domains.InvalidInput
-	if !errors.As(err, &errInvalidInput) {
-		return fmt.Errorf("get domain detail: %w", err)
-	}
-	if !strings.Contains(err.Error(), fmt.Sprintf("Domain %s not found", domainName)) {
-		return fmt.Errorf("get domain detail: %w", err)
+	var errUnsupportedTLD *route53domains.UnsupportedTLD
+	if errors.As(err, &errUnsupportedTLD) {
+		// The TLD isn't supported by Route53, hence it can't have been registered witH Route53.
```
iamhopaul123:
```suggestion
		// The TLD isn't supported by Route53, hence it can't have been registered with Route53.
```

iamhopaul123 (approved):

PR #3101, opened 2021-12-06
build: update mergify config to use queue

efekarakus:
Used a similar config to the AWS CDK: https://github.com/aws/aws-cdk/blob/master/.mergify.yml

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,5 +1,17 @@
+queue_rules:
+  - name: default
+    conditions:
+      - status-success=build (compile-linux)
+      - status-success=build (compile-windows)
+      - status-success=build (compile-darwin)
+      - status-success=test
+      - status-success=integ-test
+      - status-success=staticcheck
+      - status-success=license
+      - status-success=Semantic Pull Request
+
 pull_request_rules:
-  - name: Merge on CI success and review approval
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -1,5 +1,17 @@
...
+    conditions:
+      - status-success=build (compile-linux)
+      - status-success=build (compile-windows)
+      - status-success=build (compile-darwin)
+      - status-success=test
+      - status-success=integ-test
+      - status-success=staticcheck
+      - status-success=license
+      - status-success=Semantic Pull Request
+
```
iamhopaul123, line 12: Why do we need this 🤔

Lou1415926 (approved):

.mergify.yml
efekarakus: I didn't dig deep into it, but AFAICT now PRs:
1. First get queued (`conditions` under `pull_request_rules`)
2. Then they get merged if the conditions in the queue are satisfied (this new section)

So for example, if there are 2 approvals and no "WIP" label and CI is passed, then the PR gets queued to be merged, then Mergify checks again if the CI statuses are all valid and if so merges it.

iamhopaul123 (approved):

Issue #2525, opened 2021-06-25
access task run command's exit code


nbarrera:
I got my CI's step executing a copilot task run, the task fails with `exit 1` and I would like the CI step to fail and bail without executing the following step on the CI pipeline.

Here is the tail of the `Dockerfile` I 'm feeding the `task run`

```
ENTRYPOINT yarn migrate-up
```

Here is an `task run` output excerpt (note command failed with exit 1)

```
% DOCKER_DEFAULT_PLATFORM=linux/amd64 copilot task run   --follow   
✔ Proposing infrastructure changes for stack task-x-titles-runtime 
...
...
...
copilot-task/x-title yarn run v1.22.5
...
...
copilot-task/x-title ERROR: connect ECONNREFUSED 127.0.0.1:27017 MongoServerSelectionError: connect ECONNREFUSED 127.0.0.1:27017
copilot-task/x-title     at Timeout._onTimeout (/usr/src/app/node_modules/migrate-mongo/node_modules/mongodb/lib/core/sdam/topology.js:438:30)
copilot-task/x-title     at listOnTimeout (internal/timers.js:557:17)
copilot-task/x-title     at processTimers (internal/timers.js:500:7)
copilot-task/x-title error Command failed with exit code 1.
copilot-task/x-title info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Task has stopped.
```

When issued manually from my computer, output is the same and right after that I get the `task run` exit code like this (note the `task run` exit code is 0 even if the command's exit code was 1)

```
% echo $?
0
```

Is there a way to get what's the exit code of the task that ran inside the container?

If I run that dockerfile locally I can get that:

```
docker run d295e8ac6951
...
...
...
copilot-task/x-title error Command failed with exit code 1.
copilot-task/x-title info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
% echo $?
1
```

So, if the `task run` command does not forward the command's exit code.., could I get it somehow?

thank you in advance guys

nbarrera:
I 've found some thought at the [command design issue](https://github.com/aws/copilot-cli/issues/702), but as far as I 've tested this is not implemented.

```
The CLI returns the task's container's exit code itself. So a failure ECS-side causes failure on CLI-side. Great for running within a pipeline.
```

efekarakus:
Ooh hi @nbarrera !

[...]

PR #3106, opened 2021-12-07
feat: Show tasks exit codes when all tasks are stopped

nbarrera:
When running `task run --follow` it will now print out the `taskArn` along with the container's `exitCode`.

This maybe useful to get the `exitCode` as requested on #2525. One could `tee` the `task run --follow` command then `grep`/`awk` it to get the exit codes and make decisions with that info.

If it's too verbose for the regular case maybe we could add an option flag for this.

Maybe is not the best implementation but I thought it would be clearer to propose a change instead of trying to explain the change with comments on the issue. I 'm not pushing for you to merge it at all... I 'm kind of checking wether you think is a good enough workaround to be able to get the exit codes from copilot or not.

Please feel free to discard the PR if it's not aligned with what you envision for the tool.

### DIFF
--- a/internal/pkg/logging/task.go
+++ b/internal/pkg/logging/task.go
@@ -117,6 +117,14 @@ func (t *TaskClient) allTasksStopped() (bool, error) {
 		}
 	}
 	t.tasks = runningTasks
+
+	if stopped {
+		// Show all stopped tasks exit codes when all tasks are stopped
+		for _, tresp := range tasksResp {
+			// not sure why, but I 'm assuming there's a single container
+			log.Infoln(fmt.Sprintf("Task %s exitCode: %d", *tresp.TaskArn, *tresp.Containers[0].ExitCode))
+		}
+	}
 	return stopped, nil
 }
 


efekarakus:
Awesome! Thanks for the PR @nbarrera, apologies for the delayed response I'll get back to you with review comments tomorrow morning PST 🙇

internal/pkg/logging/task.go
```diff
@@ -117,6 +117,14 @@ func (t *TaskClient) allTasksStopped() (bool, error) {
 		}
 	}
 	t.tasks = runningTasks
+
+	if stopped {
```
efekarakus, line 121: The `logging` package is used for writing stdout logs to an `io.Writer` from ECS tasks or an App Runner service.
I think this functionality should be moved to a different package to return the exit codes for the containers in a given task.

For example:
1. Creating the functionality in the `internal/pkg/task ` pkg:
```go
package task

// ContainerDescription represents a container part of an ECS task.
type ContainerDescription struct {
   Name string   // Friendly name for the container.
   TaskID string // The task ID the container belongs to. 
   
   ExitCode int64
   ExitReason string
}

// HumanString returns a friendly colorized summary of the container description.
func (cd *ContainerDescription) HumanString() string {
   buf := new(strings.Builder)
   code := fmt.Sprintf("%s", cd.ExitCode)
   if cd.ExitCode > 0 {
      code = color.Red.Sprint(cd.ExitCode)
   }
   buf.WriteString(fmt.Sprintf("Container %s in task %s exited with status code %s", cd.Name, cd.TaskID, code)
   if cd.ExitReason != "" {
      buf.WriteString(" with reason: %s", cd.ExitReason)
   }
   return buf.String()
}

// DescribeContainers returns the status of the containers part of the given tasks.
func (r *EnvRunner) DescribeContainers(tasks []*Task) []Container { ... }

// DescribeContainers returns the status of the containers part of the given tasks.
func (r *ConfigRunner) DescribeContainers(tasks []*Task) []Container { ... }
```

2. Moving the logging functionality to the `cli` pkg:
   https://github.com/aws/copilot-cli/blob/81b3dbbe9dc8cc8c65ebf5b7bc01b2f098ea609f/internal/pkg/cli/task_run.go#L587-L592
```go
if o.follow {
   ... 
   containers, err := o.runner.DescribeContainers(tasks)
   if err != nil {
       return fmt.Errorf("describe containers in running tasks: %v", err)
   }
   for _, container := range containers {
       log.Println(container.HumanString())
   }
}
```

efekarakus (commented):
Thanks for the contribution @nbarrera !

Apologies for the long comment below 🙇 , please let me know if you need additional guidance or help from someone on the team to take the PR through the finish line.

internal/pkg/logging/task.go
nbarrera: Hi @efekarakus, 👋  sorry for the delay and thanks for your comments.

I understand that you require the new changes need to fit the existing architecture, sorry I didn't noticed.

I 've been juggling my local code based on your notes, I didn't actually run the modified code.. but, I couldn't get two things:

* First, how is `task_run.go` going to execute `containers, err := o.runner.DescribeContainers(tasks)`? Because `o.runner` is under the `Runner` interface which only exposes `RunTask` method, I know it is in fact rather an `EnvRunner` or a `ConfigRunner` which they will have the `DescribeContainers` defined, but how do you access it through the `Runner` interface? Do you downcast, need an extra finer grain interface? Maybe I 'm missing something related with go which I 'm not familiar with.

* Then on `task/task.go` I was trying to implement

```
func (r *EnvRunner) DescribeContainers(tasks []*Task) []Container { ... }
func (r *ConfigRunner) DescribeContainers(tasks []*Task) []Container { ... }
```
So I was thinking on calling `aws/ecs/ecs.go`'s `DescribeTasks` there, but I 'm not sure how to get a handle to an instantiated ecs client object from that context.

efekarakus: Hi @nbarrera 👋 

Great questions, apologies for the delay 🙇  Please let me know if any of the clarifications below don't make sense:

> First, how is task_run.go going to execute containers, err := o.runner.DescribeContainers(tasks)? 

We can update the interface: https://github.com/aws/copilot-cli/blob/78d70410d345ac4c6d7b7fb808a235f526e3f3c9/internal/pkg/cli/interfaces.go#L389-L391 to something like:
```go
type taskDescribeRunner interface {
     Run() ([]*task.Task, error)
     DescribeContainers(tasks []*Task) []Container { ... }
}
```
Note that after making this change, you'll want to run `make gen-mocks` to regenerate the Go mock types.

> but I 'm not sure how to get a handle to an instantiated ecs client object from that context.

Both `EnvRunner` and `ConfigRunner` have a `Starter` field which is really an ECS service 😄: https://github.com/aws/copilot-cli/blob/78d70410d345ac4c6d7b7fb808a235f526e3f3c9/internal/pkg/task/env_runner.go#L45
My recommendations are the following:
1. Rename `Starter` to `ECS` since tasks are specific resources to ECS.
2. Rename the `Runner` interface in the `task` pkg to allow `DescribeTasks`:
```go
type ECSClient interface {
   RunTask(input ecs.RunTaskInput) ([]*ecs.Task, error)
   DescribeTasks(cluster string, taskARNs []string) ([]*ecs.Task, error) 
}
```
3. Run `make gen-mocks`

Hope this helps!

efekarakus:
Hi @nbarrera ! I'll close the PR for now as it's gotten stale, but when we make changes to showcase the exit codes we'll build on top of these existing commits.

Thanks again for the contribution!

PR #3107, opened 2021-12-07
fix(override): task def override should not append extra overrided node

iamhopaul123:


Lou1415926 (approved):

internal/pkg/template/override/rule.go
```diff
@@ -190,6 +190,7 @@ func (m *mapUpsertNode) upsertValue(content *yaml.Node) {
 	for i := 0; i < len(content.Content); i += 2 {
 		if m.key == content.Content[i].Value {
 			content.Content[i+1] = m.valueToInsert
+			return
```
efekarakus, line 193: 🤦

efekarakus (approved):

PR #3108, opened 2021-12-08
chore(custom-resources): retrieve validation options that need to be deleted

Lou1415926:
This PR:
1. Add skeleton for "Delete" action that deletes unused certificate and records 
2. Implement functions to determine which validation options need to be deleted

### What resources are being managed
The custom resource manages the certificate, the validation options, and the aliases. Upon "Delete" event, it needs to delete:
1. The certificate that is used by the service with the specified set of aliases;
2. The validation records that will be unused after deleting the certificate;
3. The A records.

Note that 1.[The certificate that is used by the service with the specified set of aliases] is determined by:
1. The canonical certificate domain name. Copilot only creates certificate using the name "svc-nlb.env.app.domain", so it also only deletes certificate with such name.
2. The tags. Copilot should only be able to delete certificate that's Copilot-tagged.
3. The SANs. If a "Delete" event is sent with a list of alias, that means the service doesn't need this specific combination of aliases anymore. The certificate used to validate this combination is no longer needed.

A certificate satisfying all three of these conditions should be safely deleted.

### Validation Options
After deleting the certificate, the validation options used to validate the certificate need to be cleaned up. However, we should make sure that the option being deleted is not used by other certificate, other service.

A validation option can be deleted only if:
1. The option's domain is one of the three Copilot recognized pattern: a.env.app.domain, a.app.domain, or a.domain. The hosted zones associated with these three patterns are Copilot-managed; by ensuring the domain follows one of these patterns, Copilot is only touching the records in the authorized hosted zones. 
2. Among all Copilot-tagged service certificate, it is used to validate only the certificate that is pending deletion. 
3. It is not currently in use by any other service.

Note that 1. and 2. are virtually the same requirement as indicated by the current ALB implementation.

### DIFF
--- a/cf-custom-resources/lib/nlb-cert-validator-updater.js
+++ b/cf-custom-resources/lib/nlb-cert-validator-updater.js
@@ -17,6 +17,8 @@ let defaultSleep = function (ms) {
 let sleep = defaultSleep;
 let random = Math.random;
 
+let cachedServiceCertificates;
+
 const appRoute53Context = () => {
     let client;
     return () => {
@@ -52,6 +54,16 @@ const acmContext = () => {
     };
 }
 
+const resourceGroupsTaggingAPIContext = () => {
+    let client;
+    return () => {
+        if (!client) {
+            client = new AWS.ResourceGroupsTaggingAPI();
+        }
+        return client;
+    };
+}
+
 const clients = {
     app: {
         route53: appRoute53Context(),
@@ -63,6 +75,7 @@ const clients = {
         route53:envRoute53Context(),
     },
     acm: acmContext(),
+    resourceGroupsTaggingAPI: resourceGroupsTaggingAPIContext(),
 }
 
 const appHostedZoneIDContext = () => {
@@ -158,6 +171,7 @@ exports.handler = async function (event, context) {
         LoadBalancerHostedZoneID: loadBalancerHostedZoneID,
     } = props;
     const aliases = new Set(props.Aliases);
+    cachedServiceCertificates = null; // Avoid using cached certificates from the last invocation.
 
     // Initialize global variables.
     envHostedZoneID = props.EnvHostedZoneId;
@@ -204,6 +218,8 @@ exports.handler = async function (event, context) {
                 break;
             case "Update":
             case "Delete":
+                await unusedValidationOptions(aliases, loadBalancerDNS);
+                break;
             default:
                 throw new Error(`Unsupported request type ${event.RequestType}`);
         }
@@ -234,12 +250,8 @@ async function validateAliases(aliases, loadBalancerDNS) {
             HostedZoneId: hostedZoneID,
             MaxItems: "1",
             StartRecordName: alias,
-        }).promise().then((data) => {
-            let recordSet = data["ResourceRecordSets"];
-            if (!recordSet || recordSet.length === 0) {
-                return;
-            }
-            if (recordSet[0].Name !== alias) {
+        }).promise().then(({ ResourceRecordSets: recordSet }) => {
+            if (!targetRecordExists(alias, recordSet)) {
                 return;
             }
             let aliasTarget = recordSet[0].AliasTarget;
@@ -279,7 +291,7 @@ async function requestCertificate({ aliases, idempotencyToken }) {
             {
                 Key: "copilot-service",
                 Value: serviceName,
-            },
+            }
         ],
         ValidationMethod: "DNS",
     }).promise();
@@ -400,6 +412,135 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
     }).promise();
 }
 
+/**
+ * Retrieve all certificates used for the service and cache the results.
+ * @returns {Array<Object>} An array of descriptions for the certificates used by the service.
+ */
+async function serviceCertificates() {
+    if (cachedServiceCertificates) {
+        return cachedServiceCertificates;
+    }
+    let { ResourceTagMappingList } = await clients.resourceGroupsTaggingAPI().getResources({
+        TagFilters: [
+            {
+                Key: "copilot-application",
+                Values: [appName],
+            },
+            {
+                Key: "copilot-environment",
+                Values: [envName],
+            },
+            {
+                Key: "copilot-service",
+                Values: [serviceName],
+            }
+        ],
+        ResourceTypeFilters: ["acm:certificate"]
+    }).promise();
+
+    cachedServiceCertificates = [];
+    let promises = [];
+    for (let {ResourceARN: arn} of ResourceTagMappingList) {
+        let promise = clients.acm().describeCertificate({
+            CertificateArn: arn
+        }).promise().then(( { Certificate} ) => {
+            cachedServiceCertificates.push(Certificate);
+        });
+        promises.push(promise);
+    }
+    await Promise.all(promises);
+    return cachedServiceCertificates;
+}
+
+/**
+ * Retrieve validation options that will be unused by the service after deleting the certificate.
+ * An option is deemed pending unused if it validates only the certificate(s) pending deletion.
+ *
+ * @param {Set<String>} aliases is the subject alternative names of the certificate(s) pending deletion.
+ * @returns {Set<Object>} The options that will be unused by the service.
+ */
+async function optionsUnusedByService(aliases) {
+    // NOTE: `Set` uses object reference to identify elements. That is, two options with the same value will not be
+    // identify as the same option. Therefore, we use map with object string as keys to identify object with actual
+    // values instead of object reference.
+    let optionsInUse = new Map();
+    let optionsPendingUnusedByService = new Map();
+
+    let certificates = await serviceCertificates();
+    for (const {DomainName: domainName, SubjectAlternativeNames: sans, DomainValidationOptions: validationOptions} of certificates) {
+        if (domainName === certificateDomain && setEqual(aliases.add(certificateDomain), new Set(sans))) {
+            // This is a certificate pending deletion. It is:
+            // 1. Copilot-tagged, 2. use the default certificate domain name, and 3. uses the SANs that need to be deleted.
+            for (let option of validationOptions) {
+                optionsPendingUnusedByService.set(JSON.stringify(option), option);
+            }
+            continue;
+        }
+        // Otherwise, this is a certificate that we do not want to touch; validation options of this certificate are deemed in use.
+        for (let option of validationOptions) {
+            optionsInUse.set(JSON.stringify(option), option);
+        }
+    }
+    for (const optionInUseKey of optionsInUse.keys()) {
+        optionsPendingUnusedByService.delete(optionInUseKey);
+    }
+    let options = new Set();
+    for (const optionPendingUnused of optionsPendingUnusedByService.values()) {
+        options.add(optionPendingUnused);
+    }
+    return options;
+}
+
+/**
+ * Retrieve validation options that will be unused by any service.
+ *
+ * @param {Set<String>} aliases
+ * @param {String} loadBalancerDNS
+ * @returns {Promise<Set<any>>}
+ */
+async function unusedValidationOptions(aliases, loadBalancerDNS) {
+    let optionsPendingUnused = await optionsUnusedByService(aliases);
+    let promises = [];
+    for (let option of optionsPendingUnused) {
+        let domainName = option["DomainName"];
+        let hostedZoneID, route53Client;
+        try {
+            ({hostedZoneID, route53Client} = await domainResources(domainName));
+        } catch (err) {
+            if (!err instanceof UnrecognizedDomainTypeError) {
+                throw err;
+            }
+            // This option has unrecognized pattern, we can't check if it is in use, so we skip deleting it.
+            optionsPendingUnused.delete(option);
+            continue;
+        }
+        const promise = route53Client.listResourceRecordSets({
+            HostedZoneId: hostedZoneID,
+            MaxItems: "1",
+            StartRecordName: domainName,
+        }).promise().then(({ ResourceRecordSets: recordSet }) => {
+            if (!targetRecordExists(domainName, recordSet)) {
+                return; // If there is no record using this domain, it is not in use.
+            }
+            let aliasTarget = recordSet[0].AliasTarget;
+            if (aliasTarget && aliasTarget.DNSName !== `${loadBalancerDNS}.`) {
+                return; // If it is an A record in use by myself, it is pending unused.
+            }
+            optionsPendingUnused.delete(option); // Otherwise, it is in use by other services.
+        });
+        promises.push(promise);
+    }
+    await Promise.all(promises);
+    return optionsPendingUnused;
+}
+
+function targetRecordExists(targetDomainName, recordSet) {
+    if (!recordSet || recordSet.length === 0) {
+        return false;
+    }
+    return recordSet[0].Name === targetDomainName;
+}
+
 exports.deadlineExpired = function () {
     return new Promise(function (resolve, reject) {
         setTimeout(
@@ -444,8 +585,32 @@ async function domainResources (alias) {
             hostedZoneID: await hostedZoneID.root(),
         };
     }
-    throw new Error(`unrecognized domain type for ${alias}`);
+    throw new UnrecognizedDomainTypeError(`unrecognized domain type for ${alias}`);
 }
+
+/**
+ * Returns true of setA is equal to setB.
+ * @param setA
+ * @param setB
+ * @returns {boolean}
+ */
+function setEqual(setA, setB) {
+    if (setA.size !== setB.size) {
+        return false;
+    }
+
+    for (let elem of setA) {
+        if (!setB.has(elem)) {
+            return false;
+        }
+    }
+    return true;
+}
+
+function UnrecognizedDomainTypeError(message = "") {
+    this.message = message;
+}
+UnrecognizedDomainTypeError.prototype = Error.prototype;
 
 exports.withSleep = function (s) {
     sleep = s;
--- a/cf-custom-resources/lib/nlb-cert-validator-updater.js
+++ b/cf-custom-resources/lib/nlb-cert-validator-updater.js
@@ -459,7 +459,7 @@ async function serviceCertificates() {
  * @param {Set<String>} aliases is the subject alternative names of the certificate(s) pending deletion.
  * @returns {Set<Object>} The options that will be unused by the service.
  */
-async function optionsUnusedByService(aliases) {
+async function unusedOptionsByService(aliases) {
     // NOTE: `Set` uses object reference to identify elements. That is, two options with the same value will not be
     // identify as the same option. Therefore, we use map with object string as keys to identify object with actual
     // values instead of object reference.
--- a/cf-custom-resources/lib/nlb-cert-validator-updater.js
+++ b/cf-custom-resources/lib/nlb-cert-validator-updater.js
@@ -9,8 +9,7 @@ const DELAY_RECORD_SETS_CHANGE_IN_S = 30;
 const ATTEMPTS_CERTIFICATE_VALIDATED = 19;
 const DELAY_CERTIFICATE_VALIDATED_IN_S = 30;
 
-let envHostedZoneID;
-let appName, envName, serviceName, certificateDomain, domainTypes, rootDNSRole, domainName;
+let envHostedZoneID, appName, envName, serviceName, certificateDomain, domainTypes, rootDNSRole, domainName;
 let defaultSleep = function (ms) {
     return new Promise((resolve) => setTimeout(resolve, ms));
 };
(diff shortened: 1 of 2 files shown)


cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -400,6 +412,135 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
...
 }
 
+/**
+ * Retrieve all certificates used for the service and cache the results.
+ * @returns {Array<Object>} An array of descriptions for the certificates used by the service.
+ */
+async function serviceCertificates() {
+    if (cachedServiceCertificates) {
+        return cachedServiceCertificates;
+    }
```
efekarakus: I wonder if it's worthwhile introducing caching state (probably of bugs due to caching vs. performance). I think given that we expect 1 certificate usually for a service, then I think it's safe to remove caching. Let me know if there is a particular reason why you opted into this approach!

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -400,6 +412,135 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
...
+}
+
+/**
+ * Retrieve validation options that will be unused by the service after deleting the certificate.
+ * An option is deemed pending unused if it validates only the certificate(s) pending deletion.
+ *
+ * @param {Set<String>} aliases is the subject alternative names of the certificate(s) pending deletion.
+ * @returns {Set<Object>} The options that will be unused by the service.
+ */
+async function optionsUnusedByService(aliases) {
```
```suggestion
async function unusedOptionsByService(aliases) {
```

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -400,6 +412,135 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
...
+        }
+        const promise = route53Client.listResourceRecordSets({
+            HostedZoneId: hostedZoneID,
+            MaxItems: "1",
+            StartRecordName: domainName,
+        }).promise().then(({ ResourceRecordSets: recordSet }) => {
+            if (!targetRecordExists(domainName, recordSet)) {
+                return; // If there is no record using this domain, it is not in use.
+            }
+            let aliasTarget = recordSet[0].AliasTarget;
```
efekarakus: Is there always only one record set?

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -400,6 +412,135 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
...
+            if (aliasTarget && aliasTarget.DNSName !== `${loadBalancerDNS}.`) {
+                return; // If it is an A record in use by myself, it is pending unused.
+            }
+            optionsPendingUnused.delete(option); // Otherwise, it is in use by other services.
```
efekarakus: doesn't this mean:
```suggestion
            if (aliasTarget && aliasTarget.DNSName !== `${loadBalancerDNS}.`) {
                return;
            }
            optionsPendingUnused.delete(option); // The record is for my NLB, delete it.
```

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -400,6 +412,135 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
...
+async function optionsUnusedByService(aliases) {
+    // NOTE: `Set` uses object reference to identify elements. That is, two options with the same value will not be
+    // identify as the same option. Therefore, we use map with object string as keys to identify object with actual
+    // values instead of object reference.
+    let optionsInUse = new Map();
+    let optionsPendingUnusedByService = new Map();
+
+    let certificates = await serviceCertificates();
+    for (const {DomainName: domainName, SubjectAlternativeNames: sans, DomainValidationOptions: validationOptions} of certificates) {
+        if (domainName === certificateDomain && setEqual(aliases.add(certificateDomain), new Set(sans))) {
```
efekarakus: I'm struggling to understand what `setEqual(aliases.add(certificateDomain), new Set(sans))`does 🤔 

1. Why do we need to add `certificateDomain` to `aliases`?
2. Do we want to constantly append to `aliases`? or should we create a copy of the set before appending

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -400,6 +412,135 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
...
+            continue;
+        }
+        // Otherwise, this is a certificate that we do not want to touch; validation options of this certificate are deemed in use.
+        for (let option of validationOptions) {
+            optionsInUse.set(JSON.stringify(option), option);
+        }
+    }
+    for (const optionInUseKey of optionsInUse.keys()) {
+        optionsPendingUnusedByService.delete(optionInUseKey);
+    }
```
efekarakus: Why do we need this?

efekarakus (commented):
Mostly clarification questions 🥺

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: If it is in use by my NLB, it should stay in `optionsPendingUnused` because it is still considered pending unused. Otherwise, it is in use by other service, therefore it is not "pending unused". I might need to look for a better way to structure the code here since it seems like it's introducing confusion 🤔

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: 1. Because the "[SubjectAlternativeName](https://docs.aws.amazon.com/acm/latest/APIReference/API_CertificateDetail.html#ACM-Type-CertificateDetail-SubjectAlternativeNames)" field returned from the API always include the canonical domain name of the certificate.
2. Good point! We should create a copy before appending.

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: We need to delete the options that are used by another certificate. An example of when this could happen would be a certificate replacement - after the new cert is created, CFN invokes a "Delete" event, when we need to delete the old certificate. This is to insure that we don't delete any validation options that are still in use by the new cert.

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Yes because in the request we specified `MaxItems:1`. Therefore, they will return <= 1 recordset starting from `domainName`.

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: To delete the certificate, we need to give the certificateARN, which can be obtained from `resourceGroupTaggingAPI`, or `ListCertificate`. I cache this to avoid an extra API call to `resourceGroupTaggingAPI` 🤔

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -52,6 +54,16 @@ const acmContext = () => {
...
 
+const resourceGroupsTaggingAPIContext = () => {
+    let client;
+    return () => {
+        if (!client) {
+            client = new AWS.ResourceGroupsTaggingAPI();
+        }
+        return client;
+    };
+}
```
iamhopaul123, line 62: ooh this is slick!

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -279,7 +291,7 @@ async function requestCertificate({ aliases, idempotencyToken }) {
             {
                 Key: "copilot-service",
                 Value: serviceName,
-            },
+            }
         ],
         ValidationMethod: "DNS",
```
iamhopaul123: nit for consistency
```suggestion
        ValidationMethod: "DNS"
```

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -204,6 +218,8 @@ exports.handler = async function (event, context) {
                 break;
             case "Update":
             case "Delete":
+                let unusedOptions = await unusedValidationOptions(aliases, loadBalancerDNS);
+                throw new Error(`The number of options to be deleted is: ${unusedOptions.size}`);
```
iamhopaul123, line 220: Why do we throw an error here?

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -400,6 +412,135 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
...
+        let domainName = option["DomainName"];
+        let hostedZoneID, route53Client;
+        try {
+            ({hostedZoneID, route53Client} = await domainResources(domainName));
+        } catch (err) {
+            if (!err instanceof UnrecognizedDomainTypeError) {
+                throw err;
+            }
+            // This option has unrecognized pattern, we can't check if it is in use, so we skip deleting it.
+            optionsPendingUnused.delete(option);
```
iamhopaul123: Modify `optionsPendingUnused` when we are looping through `optionsPendingUnused` seems a little strange to be 🤔 Maybe we should have a new var instead.

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -158,6 +171,7 @@ exports.handler = async function (event, context) {
         LoadBalancerHostedZoneID: loadBalancerHostedZoneID,
     } = props;
     const aliases = new Set(props.Aliases);
+    cachedServiceCertificates = null; // Avoid using cached certificates from the last invocation.
```
iamhopaul123: How come this could happen. Why could it be affected by the last invocation of the lambda?

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -400,6 +412,135 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
...
+async function unusedValidationOptions(aliases, loadBalancerDNS) {
+    let optionsPendingUnused = await optionsUnusedByService(aliases);
+    let promises = [];
+    for (let option of optionsPendingUnused) {
+        let domainName = option["DomainName"];
+        let hostedZoneID, route53Client;
+        try {
+            ({hostedZoneID, route53Client} = await domainResources(domainName));
+        } catch (err) {
+            if (!err instanceof UnrecognizedDomainTypeError) {
```
iamhopaul123: Just to double check. Theoretically this wouldn't happen as any unrecognized domain would be blocked by manifest validation right?

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -400,6 +412,135 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
...
+        const promise = route53Client.listResourceRecordSets({
+            HostedZoneId: hostedZoneID,
+            MaxItems: "1",
+            StartRecordName: domainName,
+        }).promise().then(({ ResourceRecordSets: recordSet }) => {
+            if (!targetRecordExists(domainName, recordSet)) {
+                return; // If there is no record using this domain, it is not in use.
+            }
+            let aliasTarget = recordSet[0].AliasTarget;
+            if (aliasTarget && aliasTarget.DNSName !== `${loadBalancerDNS}.`) {
```
iamhopaul123: nit maybe
```suggestion
            if (aliasTarget?.DNSName !== `${loadBalancerDNS}.`) {
```

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -444,8 +585,32 @@ async function domainResources (alias) {
...
+function setEqual(setA, setB) {
+    if (setA.size !== setB.size) {
+        return false;
+    }
+
+    for (let elem of setA) {
+        if (!setB.has(elem)) {
+            return false;
+        }
+    }
```
iamhopaul123, line 635: nit we could make it more efficient by sorting the elements and then join and compare the string, or compare them one by one.

iamhopaul123 (commented):
Also mostly some clarification questions 🥺

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Ohh this is only temporary and to show that the last unit test is correct.

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Global variable is sometimes cached between lambda for performance!

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: `domainName` is read from the certificate - if user uses Copilot in every step, then yes, this wouldn't happened. If they are importing their own certificate though, their certificate could have contained any domain name that we don't expect.

cf-custom-resources/lib/nlb-cert-validator-updater.js
iamhopaul123: nvm it's a set so the current implementation is faster

cf-custom-resources/lib/nlb-cert-validator-updater.js
iamhopaul123: wut 🤯

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -400,6 +412,135 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
...
 
+/**
+ * Retrieve all certificates used for the service and cache the results.
+ * @returns {Array<Object>} An array of descriptions for the certificates used by the service.
+ */
+async function serviceCertificates() {
+    if (cachedServiceCertificates) {
+        return cachedServiceCertificates;
+    }
+    let { ResourceTagMappingList } = await clients.resourceGroupsTaggingAPI().getResources({
```
iamhopaul123, line 459: Just a callout but we used to have certificates untagged for a long time until several months ago. I guess the worst scenario is they would need to remove the orphaned certificate by themselves?

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -400,6 +412,135 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
...
+ * @returns {Promise<Set<any>>}
+ */
+async function unusedValidationOptions(aliases, loadBalancerDNS) {
+    let optionsPendingUnused = await optionsUnusedByService(aliases);
+    let promises = [];
+    for (let option of optionsPendingUnused) {
+        let domainName = option["DomainName"];
+        let hostedZoneID, route53Client;
+        try {
+            ({hostedZoneID, route53Client} = await domainResources(domainName));
```
iamhopaul123:
```suggestion
            {hostedZoneID, route53Client} = await domainResources(domainName);
```

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: This script is NLB specific, so it will not touch any ALB certificate. All NLB shall be tagged as such so the historically orphaned certificate wouldnt' be a concern for this particular custom resource!

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: The round parenthesis is sort of required if `hostedZoneID` and `route53Client` are first declared, and then assigned values with deconstruction. Otherwise it doesn't know how to parse it : (

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Talked offline ! We can do this after enabling ES2020 syntax

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Talked offline! Long story short, after ES6 it's okay to do so. The deleted element will become a special empty value, hence the iterator's indexing will not be affetecd.

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -400,15 +413,166 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
...
-    });
-};
+/**
+ * Retrieve validation options that will be unused by any service.
+ *
+ * @param {Set<String>} aliases
+ * @param {String} loadBalancerDNS
+ * @returns {Promise<Set<Object>>}
+ */
+async function unusedValidationOptions(aliases, loadBalancerDNS) {
```
efekarakus, line 420: nice this function is a lot easier to understand in this revision! 👍

cf-custom-resources/lib/nlb-cert-validator-updater.js
efekarakus: IMO this is unlikely but my worry with caching is that it's possible someone creates a certificate manually with the tags and then on function invocation we have the wrong list. I feel like it's okay to write dumb code here as latency isn't of real concern.
I'd rest better if only clients are global across invocations, but not stateful data.

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -400,15 +413,166 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
...
+ * AWS-SDK mocks cannot mock the API call if the client is initialized in a callback.
+ * @returns {Promise<boolean>} True if it is considered in use; otherwise false.
+ */
+async function inUseByOtherServices(loadBalancerDNS, domainName, route53Client) {
+    let hostedZoneID;
+    try {
+        ({hostedZoneID} = await domainResources(domainName));
+    } catch (err) {
+        if (err instanceof UnrecognizedDomainTypeError) {
+            return true; // This option has unrecognized pattern, we can't check if it is in use, so we assume it is in use.
```
efekarakus, line 540: Should we log the error here just in case to troubleshoot? I think we log generic exceptions but it might be also information to log this one as it's suprising

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -400,15 +413,166 @@ async function activateOption(option, loadBalancerDNS, loadBalancerHostedZone) {
...
+ * validate a certificate that is pending deletion.
+ * @param {Array<Object>} certsPendingDeletion
+ * @param {Array<Object>} certsInUse
+ * @returns {Promise<Set<Object>>} options that are pending deletion.
+ */
+async function unusedOptionsByService(certsPendingDeletion, certsInUse) {
+    let optionsPendingDeletion = new Map();
+    for (const { DomainValidationOptions: validationOptions } of certsPendingDeletion) {
+        for (const option of validationOptions) {
+            optionsPendingDeletion.set(JSON.stringify(option), option);
```
efekarakus: Do you know if there are any fields that could be problematic here within validation options, like `ValidationStatus`?

I think what I'm trying to ask is instead of the entire JSON object is there a better way of telling that two validation options have the same intention? maybe just looking at `ResourceRecord`?

efekarakus (commented):
Nice, great job with refactoring between revisions!

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Gotcha gotcha. That makes sense to me!

To to clarify:
>  it's possible someone creates a certificate manually with the tags and then on function invocation we have the wrong list

This is to say the list would be wrong it the certificate is created manually in between API calls right? Like after it's cached, and before the next API call.

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Makes sense!

I think this could happen very rarely, given that the certificate is Copilot-managed. It wouldn't have passed cli level validation in the first place. 

I have added the log!

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Good call out! Updated the code to just look at `ResourceRecord`. 

We only manage `ResourceRecord` (not email); therefore, the `option` in this script is pretty much a synonymous to "the resource record of a validation option".

Therefore, it's correct to use `ResourceRecord` as the identifier, as well as only taking the option into consideration if it has a `ResourceRecord` field.

efekarakus (approved):

iamhopaul123 (approved):

Issue #780, opened 2020-03-25
Support for 3 AZ infrastructure


tachyonics:
To support use cases where there is a requirement for splitting services across 3 AZs, provide an option to create infrastructure with 3 AZs rather than 2.

flemzord:
Do you have any news about the possibility of adding this feature? Or a hint to make a PR?  cc @efekarakus

iamhopaul123:
Hello @flemzord. Currently it's not planned for the short term within the month. But we'd love to get a PR to add the feature! Would you mind to give a thumb-up for this issue https://github.com/aws/copilot-cli/issues/2260 so that we can better prioritize our feature requests? Thank you!

huanjani:
This is now released 🎉 in v1.14.0 https://github.com/aws/copilot-cli/releases/tag/v1.14.0 !

PR #3115, opened 2021-12-10
chore: Default config use 6 Subnets (3 Public & 3 Private)  (#780)

flemzord:
Add support for 3 Subnet for Public & Private (1 Subnet by AZ)

#780 

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/env.go
+++ b/internal/pkg/deploy/cloudformation/stack/env.go
@@ -47,8 +47,8 @@ const (
 
 	// Default parameter values
 	DefaultVPCCIDR            = "10.0.0.0/16"
-	DefaultPublicSubnetCIDRs  = "10.0.0.0/24,10.0.1.0/24"
-	DefaultPrivateSubnetCIDRs = "10.0.2.0/24,10.0.3.0/24"
+	DefaultPublicSubnetCIDRs  = "10.0.1.0/24,10.0.2.0/24,10.0.3.0/24"
+	DefaultPrivateSubnetCIDRs = "10.0.10.0/24,10.0.11.0/24,10.0.12.0/24"
 )
 
 var (


internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -47,8 +47,8 @@ const (
 
 	// Default parameter values
 	DefaultVPCCIDR            = "10.0.0.0/16"
-	DefaultPublicSubnetCIDRs  = "10.0.0.0/24,10.0.1.0/24"
-	DefaultPrivateSubnetCIDRs = "10.0.2.0/24,10.0.3.0/24"
+	DefaultPublicSubnetCIDRs  = "10.0.1.0/24,10.0.2.0/24,10.0.3.0/24"
+	DefaultPrivateSubnetCIDRs = "10.0.10.0/24,10.0.11.0/24,10.0.12.0/24"
```
efekarakus, line 51: Hi @flemzord 👋 

Thank you for the PR ! I believe the reason why we don't default to 3 AZs is because it's not guaranteed that a region has three AZs. For example: https://aws.amazon.com/about-aws/global-infrastructure/regions_az/
> New customers can access two Availability Zones in US West (Northern California).

In order to guarantee a working experience, we started with two AZs by default.

I believe a neat way of doing this would be to add a new flag to `copilot env init`:
```
--override-azs   Optional. Availability zone names (defaults to 2 AZs).
```

This way when initializing an environment you can choose to override defaults:
```
Would you like to use the default configuration?
> Yes, but I'd like configure the default resources (AZs, CIDR ranges)

What VPC CIDR would you like to use? [? for help] (10.0.0.0/16)
Which availability zones would you like to use? (us-east-1a,us-east-1b) us-east-1a,us-east-1b,us-east-1c
What CIDR would you like to use for your public subnets? [? for help] (10.0.0.0/24,10.0.1.0/24,10.0.2.0/24) 
```

Would that behavior address your feature request?

flemzord: Hello @efekarakus 

Thanks for the feedback, that would be a good idea for a workaround. 
I think it would be nice to change the default ip addressing anyway so that it allows adding subnets easily. 

I didn't know that there was still an AWS region with only 2 AZ 😨

efekarakus: We might not be able to touch the default ip addressing for backwards compatibility reasons (The stack update should fail while upgrading an environment template version if the cidr ranges change), but we can definitely change the defaults or allow overriding them when creating the environment

efekarakus: We should be able to address the feedback above early next week if that's okay with you? unless you would like to make the code changes 😄

efekarakus: Hi @flemzord !

Just for visibility, we started sending PRs for the individual sub-steps to support this feature request:
* https://github.com/aws/copilot-cli/pull/3123
* https://github.com/aws/copilot-cli/pull/3125

efekarakus:
OK! Once https://github.com/aws/copilot-cli/pull/3131 is merged this feature should be supported in the CLI!

flemzord:
I close this PR :) Thanks @efekarakus

PR #3116, opened 2021-12-10
fix: svc/job init error out if workload already exists

### LINKED ISSUE #3085: Upgrading from v.1.12 to v1.13 removes any public subnets defined in the env exports
After upgrading from v1.12 to v.1.13 I was left with a broken environment and could not deploy without destroying it and recreating it. 

Breakage was due to the Public Subnets export being removed when being upgraded.

Resolution for me was to tear it all down and recreate the environment with the proper imports. However, this solution isn't sustainable.

iamhopaul123:
fixes #3085

### DIFF
--- a/internal/pkg/cli/job_init.go
+++ b/internal/pkg/cli/job_init.go
@@ -258,7 +258,8 @@ func (o *initJobOpts) validateDuplicateJob() error {
 	if err == nil {
 		log.Errorf(`It seems like you are trying to init a job that already exists.
 To recreate the job, please run:
-1. %s. Note that the manifest will remain for reusing purpose.
+1. %s. Note: The manifest file will not be deleted and will be used in Step 2.
+If you'd prefer a new default manifest, please manually delete the existing one.
 2. And then %s
 `,
 			color.HighlightCode(fmt.Sprintf("copilot job delete --name %s", o.name)),
--- a/internal/pkg/cli/job_init.go
+++ b/internal/pkg/cli/job_init.go
@@ -28,10 +28,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const (
-	job = "job"
-)
-
 var (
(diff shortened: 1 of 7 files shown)


internal/pkg/cli/job_init.go
```diff
@@ -254,6 +253,26 @@ func (o *initJobOpts) RecommendActions() error {
 	return nil
 }
 
+func (o *initJobOpts) validateDuplicateJob() error {
+	_, err := o.store.GetJob(o.appName, o.name)
+	if err == nil {
+		log.Errorf(`It seems like you are trying to init a job that already exists.
+To recreate the job, please run:
```
huanjani, line 260: I wonder if we should mention that the manifest will remain and they can opt to reuse it or not.

iamhopaul123: That's a good idea!

internal/pkg/cli/svc_init.go
```diff
@@ -328,6 +332,26 @@ func (o *initSvcOpts) askSvcType() error {
 	return nil
 }
 
+func (o *initSvcOpts) validateDuplicateSvc() error {
+	_, err := o.store.GetService(o.appName, o.name)
+	if err == nil {
+		log.Errorf(`It seems like you are trying to init a service that already exists.
```
huanjani, line 338: same as with job, above-- mention manifest file?

Lou1415926 (approved):

internal/pkg/cli/job_init.go
```diff
@@ -254,6 +253,26 @@ func (o *initJobOpts) RecommendActions() error {
...
+1. %s. Note that the manifest will remain for reusing purpose.
```
huanjani: How about something like, "Note: The manifest file will not be deleted and will be used in Step 2. If you'd prefer a new default manifest, please manually delete the existing one."

huanjani (approved):
:shipit:

PR #3117, opened 2021-12-13
fix: asg should work with --force

### LINKED ISSUE #3092: Deploy --force is not working when auto scaling set.
Hi,

We had set the location in manifest to pull from the static image url (with the latest tag), eg. repo.example/staging/helloworld:latest
While we use the manifest without auto scaling, copilot svc deploy --force it work perfectly.
After that we try to add auto scaling eg.
 
environments:
  staging:
    http:
      api : api.example.com
    count:
      range:
        min: 2
        max: 3
        spot_from: 2
      cpu_percentage: 50

copilot svc deploy --force will be deploy successfully at the first time, but the second time it wont recreate the tasks.

This is the old manifest.yml
environments:
  staging:
    http:
      api : api.example.com

Thanks for you assistance.

iamhopaul123:
fixes #3092 


internal/pkg/aws/cloudformation/cloudformation.go
```diff
@@ -187,17 +187,22 @@ func (c *CloudFormation) Describe(name string) (*StackDescription, error) {
...
 		var notFound *ErrStackNotFound
 		if errors.As(err, &notFound) {
 			return false, nil
 		}
 		return false, err
 	}
-	return true, nil
+	status := aws.StringValue(resp.StackStatus)
+	return status == cloudformation.StackStatusUpdateRollbackComplete || status == cloudformation.StackStatusCreateComplete ||
+		status == cloudformation.StackStatusUpdateComplete, nil
```
efekarakus: Can you add a method in `stackstatus.go` like `func (ss StackStatus) IsComplete() bool` so we can do:
```
return (stackstatus(status)).IsComplete(), nil
```

internal/pkg/aws/cloudformation/cloudformation.go
```diff
@@ -187,17 +187,22 @@ func (c *CloudFormation) Describe(name string) (*StackDescription, error) {
 	return &descr, nil
 }
 
-// Exists returns true if the CloudFormation stack exists, false otherwise.
+// Exists returns true if an intact CloudFormation stack exists, false otherwise.
+// An intact stack can only be in one of the following status: UPDATE_ROLLBACK_COMPLETE,
+// CREATE_COMPLETE, or UPDATE_COMPLETE.
 // If an error occurs for another reason than ErrStackNotFound, then returns the error.
 func (c *CloudFormation) Exists(name string) (bool, error) {
```
efekarakus: I'm not sure if the method name is now obvious. A client might think that stacks statuses don't matter 💭 

What do you think of the following signature?
```go
func (c *CloudFormation) Exists(name string, wantedStatuses ...string) 
```

This way existing code doesn't need to be updated `Exist("myapp-myenv")` means all statuses are possible, whereas `Exists("myapp-myenv", cloudformation.CompleteStackStatuses...)` means only for "_COMPLETE" statuses.

internal/pkg/cli/svc_deploy.go
```diff
@@ -338,7 +339,8 @@ func (o *deploySvcOpts) configureClients() error {
 	}
 
 	// CF client against env account profile AND target environment region.
-	o.svcCFN = cloudformation.New(envSession)
+	o.deployer = cloudformation.New(envSession)
+	o.stackExistChecker = awscloudformation.New(envSession)
```
efekarakus, line 343: :cool-crying:

internal/pkg/deploy/cloudformation/stack/name.go
```diff
@@ -19,8 +19,8 @@ func (t TaskStackName) TaskName() string {
 	return strings.SplitN(string(t), "-", 2)[1]
 }
 
-// NameForService returns the stack name for a service.
-func NameForService(app, env, svc string) string {
+// NameForWorkload returns the stack name for a workload.
+func NameForWorkload(app, env, svc string) string {
```
efekarakus: nit:
```suggestion
func NameForWorkload(app, env, name string) string {
```

internal/pkg/aws/cloudformation/cloudformation.go
efekarakus: Also why do we want to limit to only these statuses?

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
```diff
@@ -3,11 +3,7 @@ Cluster:
   Fn::ImportValue:
     !Sub '${AppName}-${EnvName}-ClusterId'
 TaskDefinition: !Ref TaskDefinition
-{{- if .DesiredCountOnSpot}}
-DesiredCount: !Ref TaskCount
-{{- else if .Autoscaling}}
-DesiredCount: !GetAtt DynamicDesiredCountAction.DesiredCount
-{{- else }}
+{{- if not (and .Autoscaling .UpdateRequired) }}
```
efekarakus, line 6: It took me way too long to parse this conditional for some reason 😂

internal/pkg/template/workload.go
```diff
@@ -364,6 +364,9 @@ func (p RuntimePlatformOpts) isEmpty() bool {
 
 // WorkloadOpts holds optional data that can be provided to enable features in a workload stack template.
 type WorkloadOpts struct {
+	// Stack status variables.
+	UpdateRequired bool
```
```suggestion
	UpdateRequired bool // True if the stack already exists and will be updated.
```

efekarakus (commented):
Aww yeah all that deletion!

internal/pkg/aws/cloudformation/cloudformation.go
iamhopaul123: Because we only perform an update if the stack are in those statuses.

PR #3122, opened 2021-12-14
chore(custom-resources): "Delete" and "Update" NLB certificate and alias

Lou1415926:
This PR implements the "Delete" and "Update" action for the custom resource that manages the NLB certificate and aliases.

Following the #3108 PR.

### DIFF
--- a/cf-custom-resources/lib/nlb-cert-validator-updater.js
+++ b/cf-custom-resources/lib/nlb-cert-validator-updater.js
@@ -201,6 +201,13 @@ exports.handler = async function (event, context) {
 
     let handler = async function() {
         switch (event.RequestType) {
+            case "Update":
+                let oldAliases = new Set(event.OldResourceProperties.Aliases);
+                let oldAliasesSorted = [...oldAliases].sort().join(",");
+                if (oldAliasesSorted === aliasesSorted) {
+                    break;
+                }
+                // fallthrough
             case "Create":
                 await validateAliases(aliases, loadBalancerDNS);
                 const certificateARN = await requestCertificate({
@@ -212,7 +219,6 @@ exports.handler = async function (event, context) {
                 const options = await waitForValidationOptionsToBeReady(certificateARN, aliases);
                 await activate(options, certificateARN, loadBalancerDNS, loadBalancerHostedZoneID);
                 break;
-            case "Update":
             case "Delete":
                 let unusedOptions = await unusedValidationOptions(aliases, loadBalancerDNS);
                 await deleteCertificate(aliases);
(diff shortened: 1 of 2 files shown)


cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -451,6 +457,92 @@ async function unusedValidationOptions(aliases, loadBalancerDNS) {
...
+                }
+            }
+        });
+    }
+
+    let {hostedZoneID, route53Client} = await domainResources(option.DomainName);
+    let changeResourceRecordSetsInput = {
+        ChangeBatch: {
+            Comment: `Delete the validation record and A record for ${option.DomainName}`,
+            Changes: changes,
```
efekarakus, line 521: what's the max batch size, do we feel confident that we won't reach an error because we exceeded the max limit?

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -582,10 +674,9 @@ function targetRecordExists(targetDomainName, recordSet) {
     if (!recordSet || recordSet.length === 0) {
         return false;
     }
-    return recordSet[0].Name === targetDomainName;
+    return recordSet[0].Name === `${targetDomainName}.`;
```
efekarakus, line 677: 🤔

efekarakus (commented):
Looks great, just a question!

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: We do one `changeResourceRecordSets` call for each validation option; each call contains at most two changes in the batch: 1. one change is to remove the validation record; 2. one change is to remove the A-record of the alias.

Since there are at most two changes in each change batch, we won't max out! (Fairly sure that 2 can't be the cap)

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: Ya... the response from route53 has the trailing dot. I could upsert the record using a domain name with or without the trailing dot - they are treated the same. It's just the response does always have the trailing dot, no matter how I upserted it in the first place.

cf-custom-resources/lib/nlb-cert-validator-updater.js
efekarakus: derp 🤦

efekarakus (approved):

huanjani (approved):
Your descriptive const/var names in the code + tests makes this easier to understand! 👍🏼

PR #3125, opened 2021-12-15
chore(ec2): add ListAZs method

efekarakus:


internal/pkg/aws/ec2/ec2_test.go
```diff
@@ -158,6 +158,93 @@ func TestEC2_ListVPC(t *testing.T) {
...
+							Name:   aws.String("zone-type"),
+							Values: aws.StringSlice([]string{"availability-zone"}),
+						},
+						{
+							Name:   aws.String("state"),
+							Values: aws.StringSlice([]string{"available"}),
+						},
+					},
+				}).Return(&ec2.DescribeAvailabilityZonesOutput{
+					AvailabilityZones: []*ec2.AvailabilityZone{
```
huanjani, line 188: Should we add one or two that get filtered out?

efekarakus: Luckily the filtering doesn't happen client side ✌️ ! EC2 should return to us only AZs that are available so we don't have to do any filtering ourselves

huanjani: Ah, gotcha gotcha. Right, makes sense. 👍🏼

huanjani (approved):
LGTM!

Makefile
```diff
@@ -155,7 +155,7 @@ gen-mocks: tools
 	${GOBIN}/mockgen -package=mocks -destination=./internal/pkg/aws/ec2/mocks/mock_ec2.go -source=./internal/pkg/aws/ec2/ec2.go
 	${GOBIN}/mockgen -package=mocks -destination=./internal/pkg/aws/identity/mocks/mock_identity.go -source=./internal/pkg/aws/identity/identity.go
 	${GOBIN}/mockgen -package=mocks -destination=./internal/pkg/aws/route53/mocks/mock_route53.go -source=./internal/pkg/aws/route53/route53.go
-	${GOBIN}/mockgen -package=mocks -destination=./internal/pkg/aws/route53/mocks/mock_route53_domain.go -source=./internal/pkg/aws/route53/route53domains.go
+	${GOBIN}/mockgen -package=mocks -destination=./internal/pkg/aws/route53/mocks/mock_route53_domain.go -source=./internal/pkg/aws/route53/domain.go
```
Lou1415926, line 158: :homer-disappearing:

internal/pkg/aws/ec2/ec2.go
```diff
@@ -183,6 +187,33 @@ func (c *EC2) ListVPCs() ([]VPC, error) {
...
+	resp, err := c.client.DescribeAvailabilityZones(&ec2.DescribeAvailabilityZonesInput{
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("zone-type"),
+				Values: aws.StringSlice([]string{"availability-zone"}),
+			},
+			{
+				Name:   aws.String("state"),
+				Values: aws.StringSlice([]string{"available"}),
+			},
```
Lou1415926, line 201: Would we need the "opt-in-status" filter?

efekarakus: My understanding from the SDK description is that we don't need to: https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#DescribeAvailabilityZonesInput
```
    // Include all Availability Zones, Local Zones, and Wavelength Zones regardless
    // of your opt-in status.
    //
    // If you do not use this parameter, the results include only the zones for
    // the Regions where you have chosen the option to opt in.
    AllAvailabilityZones *bool `type:"boolean"`
```
(we don't use the param)

Lou1415926 (approved):

Issue #1579, opened 2020-10-27
Copilot doesn't provide ability to enable cloudwatch performance monitoring of cluster.


sanjeevtejyan:
To enable cloudwatch for copilot rolled-out cluster, one has to execute aws command manually to enable the cloudwatch to monitor the performance of individual task and service. But it can be made part of configuration when a ECS cluster is setup.

efekarakus:
Heya @sanjeevtejyan !

To make sure that I understand the request correctly, do you mean you're looking for a way of enabling [ECS CloudWatch Container Insights](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Container-Insights-metrics-ECS.html) with Copilot? Or are you asking about different metrics?

Thanks!

sanjeevtejyan:
yes that is correct. Can we hook any other metric to measure the performance of the cluster?

tam0ri:
Similar issue: #1614

Currently, `AWS::ECS::Cluster` resource does not have `ClusterSettings` property in a template.
https://github.com/aws/copilot-cli/blob/069bcf00d7edd8899db939963084ebd8b4767b01/templates/environment/versions/cf-v1.3.0.yml#L49-L57

So Container Insights setting depends on account level settings. I think it's useful if we can enable/disable when creates environment.

## Proposal

Adding an option for `env init`.

### Option 1:
```
--container-insights enabled | disabled
```

When the option is not specified, `ClusterSettings` property is removed from template and depends on account settings.


### Option 2:
```
--container-insights-enabled true | false
```

When the option is not specified, set true or false as default value.

dannysteenman:
+1 for this feature!

> Similar issue: #1614
> 
> Currently, `AWS::ECS::Cluster` resource does not have `ClusterSettings` property in a template.
> 
> https://github.com/aws/copilot-cli/blob/069bcf00d7edd8899db939963084ebd8b4767b01/templates/environment/versions/cf-v1.3.0.yml#L49-L57
> 
> So Container Insights setting depends on account level settings. I think it's useful if we can enable/disable when creates environment.
> 
> ## Proposal
> Adding an option for `env init`.
> 
> ### Option 1:
> ```
> --container-insights enabled | disabled
> ```
> 
> When the option is not specified, `ClusterSettings` property is removed from template and depends on account settings.
> 
> ### Option 2:
> ```
> --container-insights-enabled true | false
> ```
> 
> When the option is not specified, set true or false as default value.

[...]

PR #3126, opened 2021-12-15
feat(cli): add `--container-insights` flag to `env init`

flemzord:
I've added support for Container Insight for the next version of Copilot :) 
You can override the value via the CLI if you don't want to activate it, but it will be activated by default. 

#1579 


efekarakus:
Hi @flemzord ! Thanks again for the contribution 😄 

Appologies on not reviewing yesterday 🙇 , I'll get back to you with a real review once we finish #3115 (multi AZ support).

I just wanted to leave some high level feedback from glancing at the PR to hear your thoughts:
1. We didn't enable container insights by default for users as it results in additional costs, so we want it to be an opted-in feature instead of being on by default hope that makes sense.
   > Metrics collected by Container Insights are charged as custom metrics. For more information about CloudWatch pricing, see Amazon CloudWatch Pricing. 
   > https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/ContainerInsights.html

2. This might be a cool opportunity to think about how to leverage the `copilot env init --prod` flag.  
This is just a high level idea but I think it'd be 🤩 if users could configure "--prod" environment settings such as enabling container insights or access logs for ALB, so that anytime they use the `--prod` flag those features would be turned on 🤔

efekarakus:
For future code changes, is it possible to follow these steps in the contributing guide please 🙏 ? 

> * Let us know if you are interested in working on an issue by leaving a comment on the issue in GitHub. 
> * If you would like to propose a new feature, please open an issue on GitHub with a detailed description. This enables us to collaborate on the feature design more easily and increases the chances that your feature request will be accepted.
>  https://github.com/aws/copilot-cli/blob/mainline/CONTRIBUTING.md#contributing-code

This way the PRs can be merged faster once we align on the interface 🚀  
No worries for this PR, the changes are much appreciated but we have to think of the interface a bit more deeply I think 💭

flemzord:
@efekarakus I change my code for use `--prod` value. 
But, with or without the `--prod` flag, the value in the cloudformation template remains at 'false'. 
Do you have an idea?

huanjani:
Hi, @flemzord!
Just a note to say that we haven't forgotten about your PR and this is still on our radar. 🤗 
Our response time is a bit delayed due to the holiday season.
We'll be in touch! Thank you!

efekarakus:
Hi @flemzord ! Happy new year 🎈 

What do you think of enabling several observability fields like container insights, elb access logs, vpc flow logs all at once if `--prod` is specified (https://github.com/aws/copilot-cli/issues/1579#issuecomment-1004003827)?

flemzord:
Hey @efekarakus, Happy New Year too :) 
Why not, but I think we should ask for more information from the user as the bucket s3 target. 
For example, we use Control Tower, we have an AWS account dedicated to the Log, it would be nice to send the logs directly in it and not in the "application" account.

efekarakus:
> For example, we use Control Tower, we have an AWS account dedicated to the Log, it would be nice to send the logs directly in it and not in the "application" account.

Cool this makes total sense! Does this mean that you'd want to have a single `--prod` config that can be re-used while initializing environments? Or would like you to configure container insights, access logs individually each time you create a new environment?

efekarakus:
Hi @flemzord! Apologies on the delay, I've pretty much reverted the change to your old commit with some minor tweaks.

Now the help menu for `copilot env init` contains:
```
Telemetry Flags
      --container-insights   Optional. Enable CloudWatch Container Insights.
```

For future other telemetry options, we can add additional flags here.

internal/pkg/cli/env_init.go
```diff
@@ -798,7 +811,8 @@ func buildEnvInitCmd() *cobra.Command {
 	cmd.Flags().StringVar(&vars.tempCreds.SessionToken, sessionTokenFlag, "", sessionTokenFlagDescription)
 	cmd.Flags().StringVar(&vars.region, regionFlag, "", envRegionTokenFlagDescription)
 
-	cmd.Flags().BoolVar(&vars.isProduction, prodEnvFlag, false, prodEnvFlagDescription)
+	cmd.Flags().BoolVar(&vars.isProduction, prodEnvFlag, false, prodEnvFlagDescription) // Deprecated. Use telemetry flags instead.
```
iamhopaul123, line 826: Do we need to update the description showing we are deprecating this flag?

efekarakus: Not needed boss ✌️ because I've removed this flag at [L. 823](https://github.com/aws/copilot-cli/pull/3126/files/39fb9f9f27240e1b69130fb7a4070bd26d86c29b#diff-d61e7e024599d5b260984b80738cca7a2fc2a06ac37295b4c5900f9cea3fb8ebL823) so it doesn't show up in the help menu at all. However, it will still work for backwards compatible if a user is scripting the command

huanjani (approved):
This looks good to me; I like that it sets the stage for future related additions, too.

internal/pkg/cli/env_init.go
```diff
@@ -666,14 +678,15 @@ func (o *initEnvOpts) deployEnv(app *config.Application,
...
 			AccountPrincipalARN: caller.RootUserARN,
 		},
-		Prod:                 o.isProduction,
-		AdditionalTags:       app.Tags,
-		CustomResourcesURLs:  customResourcesURLs,
-		ArtifactBucketARN:    artifactBucketARN,
+		Prod:                o.isProduction,
+		AdditionalTags:      app.Tags,
+		CustomResourcesURLs: customResourcesURLs,
+    ArtifactBucketARN:    artifactBucketARN,
```
Lou1415926: nit: we can keep this line aligned with the rest 👀

internal/pkg/cli/env_init.go
```diff
@@ -782,7 +795,7 @@ func buildEnvInitCmd() *cobra.Command {
   /code $ copilot env init --name test --profile default --default-config
 
   Creates a prod-iad environment using your "prod-admin" AWS profile.
```
Lou1415926:
```suggestion
  Creates a prod-iad environment using your "prod-admin" AWS profile, as well as enabling container insights.
```

internal/pkg/cli/env_upgrade.go
```diff
@@ -284,13 +284,14 @@ func (o *envUpgradeOpts) upgradeEnvironment(upgrader envUpgrader, conf *config.E
 		App: deploy.AppInformation{
 			Name: conf.App,
 		},
-		Name:                 conf.Name,
-		ArtifactBucketKeyARN: artifactBucketKeyARN,
+		Name:                conf.Name,
+    ArtifactBucketKeyARN: artifactBucketKeyARN,
```
Lou1415926: nit: same here (aligning with other lines)

internal/pkg/cli/env_upgrade.go
```diff
@@ -314,11 +315,12 @@ func (o *envUpgradeOpts) upgradeLegacyEnvironment(upgrader legacyEnvUpgrader, co
 			App: deploy.AppInformation{
 				Name: conf.App,
 			},
-			Name:                 conf.Name,
-			ArtifactBucketKeyARN: artifactBucketKeyARN,
+			Name:                conf.Name,
+      ArtifactBucketKeyARN: artifactBucketKeyARN,
```
Lou1415926: nit: samesiess

internal/pkg/cli/env_upgrade_test.go
```diff
@@ -288,9 +291,12 @@ func TestEnvUpgradeOpts_Execute(t *testing.T) {
...
 					},
-					CFNServiceRoleARN:    "execARN",
-					CustomResourcesURLs:  map[string]string{"mockCustomResource": "mockURL"},
-					ArtifactBucketARN:    "arn:aws:s3:::mockBucket",
+					CFNServiceRoleARN:   "execARN",
+					CustomResourcesURLs: map[string]string{"mockCustomResource": "mockURL"},
+					Telemetry: &config.Telemetry{
+						EnableContainerInsights: true,
+					},
+          ArtifactBucketARN:    "arn:aws:s3:::mockBucket",
```
Lou1415926: nit: samesiess :homer-disappearing:

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -71,6 +71,15 @@ Resources:
       Configuration:
         ExecuteCommandConfiguration:
           Logging: DEFAULT
+{{- if .Telemetry }}
+      ClusterSettings:
+        - Name: containerInsights
```
Lou1415926, line 76: Will this accidentally disable an enabled container insights during `env upgrade`? What about 
```suggestion
    {{ - if .Telemetry.EnableContainerInsights}}
        - Name: containerInsights
           Value: enabled
     {{- end}}
```
?

internal/pkg/cli/env_init.go
efekarakus: ooh I was so confused how this happened but it's because of the commit that merged mainline to this branch.

Fixed!

internal/pkg/cli/env_init.go
efekarakus: Done

internal/pkg/cli/env_upgrade.go
efekarakus: Done

internal/pkg/template/templates/environment/cf.yml
efekarakus: 1. For existing client environments: `.Telemetry` is nil so we have to have a whole `{{- if .Telemetry}}` block in order to not panic while upgrading their environments.
2. For new environments, `.Telemetry` is always not nil and `EnableContainerInsights` is either `true` or `false` so it should be fine.

Does that make sense or am I missing a case?

Lou1415926: My concern was that it will disable the container insight if the block (i.e. the block of `ClusterSettings`) isn't rendered into the template. But this probably isn't a valid concern because for existing environments the template has always looked like this (without the block).

Lou1415926 (approved):

Lou1415926:
This is now released 🎉 in v1.14.0 https://github.com/aws/copilot-cli/releases/tag/v1.14.0 !

codekitchen:
🥳  We were excited to see this in the release notes! Sorry if this is the wrong place to ask, but I don't see it documented -- is there a way to turn this on for existing environments without re-creating them?

efekarakus:
Hi @codekitchen ! We're working on adding the flag to `copilot env upgrade` so hopefully next release. At the moment it can be turned on only for new environments, apologies 🙇

PR #3127, opened 2021-12-15
fix: asg should work with --force for svc deploy

### LINKED ISSUE #3092: Deploy --force is not working when auto scaling set.

iamhopaul123:
fixes #3092


internal/pkg/apprunner/apprunner.go
```diff
@@ -55,6 +57,19 @@ func (c Client) ForceUpdateService(app, env, svc string) error {
 	return c.appRunnerClient.WaitForOperation(id, svcARN)
 }
 
+// LastUpdatedAt returns the last updated time of the app runner service.
+func (c Client) LastUpdatedAt(app, env, svc string) (*time.Time, error) {
```
efekarakus: Can we return just `time.Time` (not a pointer)?

I tested this even if the service is not yet created, the `UpdatedAt` has a value that's equal to `CreatedAt`so it should be safe

internal/pkg/cli/svc_deploy.go
```diff
@@ -74,41 +75,44 @@ type uploadCustomResourcesOpts struct {
...
+	targetApp           *config.Application
+	targetEnvironment   *config.Environment
+	targetSvc           *config.Workload
+	appliedManifest     interface{}
+	imageDigest         string
+	buildRequired       bool
+	appEnvResources     *stack.AppRegionalResources
+	rdSvcAlias          string
+	svcUpdater          serviceUpdater
+	svcLastUpdateGetter serviceLastUpdateGetter
```
efekarakus: instead of creating a new field `svcLastUpdateGetter` what do you think of renaming the `serviceUpdater` interface to:
```go
type svcForceUpdater interface {
  ForceUpdateService(app, env, svc string) error
  LastUpdatedAt(app, env, svc string) (time.Time, error)
}
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -696,21 +709,33 @@ func (o *deploySvcOpts) deploySvc(addonsURL string) error {
 		return err
 	}
 
+	if o.cmdRunAt == nil {
+		o.cmdRunAt = aws.Time(time.Now())
+	}
```
efekarakus: nit: another way I think of achieving this would be

```go
type deploySvcOpts struct {
   ...
   now   func() time.Time
}

func newSvcDeployOpts(vars deployWkldVars) (*deploySvcOpts, error) {
   opts.now = time.Now
}
```
and in the tests we can put a `fakeNow` function

internal/pkg/cli/svc_deploy.go
```diff
@@ -696,21 +709,33 @@ func (o *deploySvcOpts) deploySvc(addonsURL string) error {
...
+		if !o.forceNewUpdate {
 			log.Warningf("Set --%s to force an update for the service.\n", forceFlag)
+			return fmt.Errorf("deploy service: %w", err)
+		}
+	}
+	// Force update the service if --force is set and the service is not updated by the CFN.
+	if o.forceNewUpdate {
+		lastUpdatedAt, err := o.svcLastUpdateGetter.LastUpdatedAt(o.appName, o.envName, o.name)
+		if err != nil {
+			return fmt.Errorf("get the last updated time for %s: %w", o.name, err)
```
efekarakus:
```suggestion
			return fmt.Errorf("get the last updated deployment time for service %s: %w", o.name, err)
```

efekarakus (commented):
Looks great, just one small suggestion

internal/pkg/apprunner/apprunner.go
iamhopaul123: I use the pointer because `time.Time{}` is a valid time so that we can't use `time.Time{}` to denote the field is empty (the same logic as `count`). In terms of safety as long as we use `aws.TimeValue()` I think we should be good? What do you think 🤔

efekarakus: Yeah, it makes sense, I think my point was that at least for App Runner this value is always set so it can't ever be `nil`. If that's the case also with ECS we can get rid of an unnecessary condition. Otherwise, we can keep it as is

iamhopaul123: nvm i think they are about the same since `time.Time{}` doesn't really happen

internal/pkg/cli/svc_deploy.go
```diff
@@ -108,7 +109,8 @@ type deploySvcOpts struct {
 	buildRequired     bool
 	appEnvResources   *stack.AppRegionalResources
 	rdSvcAlias        string
-	svcUpdater        serviceUpdater
+	svcUpdater        svcForceUpdater
+	now               func() time.Time
```
efekarakus, line 113: I think we need to update `init.go` and `deploy.go` as well with `now: time.Now`

internal/pkg/cli/svc_deploy.go
```diff
@@ -148,8 +150,11 @@ func newSvcDeployOpts(vars deployWkldVars) (*deploySvcOpts, error) {
 		},
 		newInterpolator: newManifestInterpolator,
 		cmd:             exec.NewCmd(),
-		sessProvider:    sessions.NewProvider(),
-		snsTopicGetter:  deployStore,
+		now: func() time.Time {
+			return time.Now()
+		},
```
efekarakus:
```suggestion
		now: time.Now,
```

efekarakus (approved):

internal/pkg/cli/svc_deploy.go
iamhopaul123: Oooh good call!

Lou1415926 (approved):
LGTM!

PR #3129, opened 2021-12-16
chore(custom-resources): update lambda script to use certificate ARN as physical resource ID

Lou1415926:
Previously, the script use `service/<sorted aliases>` as the physical resource ID. This is found not to be the most accurate representation of the resource. 

A certificate is not uniquely defined by canonical domain, SANs and tags. Although the Copilot workflow always maintain only one such certificate, it should also be robust against unforeseen situation where multiple such certificates exist. 


iamhopaul123 (approved):

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -214,14 +209,23 @@ exports.handler = async function (event, context) {
...
                         .digest("hex")});
+                physicalResourceID = certificateARN; // Update the physical resource ID if a new certificate is created.
                 const options = await waitForValidationOptionsToBeReady(certificateARN, aliases);
                 await activate(options, certificateARN, loadBalancerDNS, loadBalancerHostedZoneID);
                 break;
             case "Delete":
-                let unusedOptions = await unusedValidationOptions(aliases, loadBalancerDNS);
-                await deleteCertificate(aliases);
+                if (!physicalResourceID.startsWith("arn:")) {
+                    // This means no certificate has been created, nor any records. Exit without doing anything.
```
efekarakus: would this cause a weird exception if `physicalResourceID` is `undefined`?

cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -192,13 +192,8 @@ exports.handler = async function (event, context) {
...
     };
 
-    // NOTE: If the aliases have changed, then we need to replace the certificate being used, as well as deleting/adding
-    // validation records and A records. In general, any change in aliases indicate a "replacement" of the resources
-    // managed by the custom resource lambda; on the contrary, the same set of aliases indicate that there is no need to
-    // replace or update the certificate, nor the validation records or A records. Hence, we can use this as the physicalResourceID.
     let aliasesSorted = [...aliases].sort().join(",");
-    const physicalResourceID = `/${serviceName}/${aliasesSorted}`;
-
+    let physicalResourceID = event.PhysicalResourceId; // Keep old physical resource ID unchanged by default.
```
```suggestion
    let physicalResourceID = event.PhysicalResourceId; // The certificate ARN. By default, keep old physical resource ID unchanged.
```

efekarakus (approved):
Looks good, just a question. Feel free to remove label once addressed!

cf-custom-resources/lib/nlb-cert-validator-updater.js
Lou1415926: From what I have observed, the `physicalResourceID` would be provided as a weird looking random string (e.g. "a3dkdfanrdsa3") to the "Delete" event when the creation fails. However I can't find the documentation for this behavior, so I added an extra logic that says: `if (!physicalResourceID || !physicalResourceID.startsWith("arn:"))` to deal with the possibility when the physical resource ID is undefined.

PR #3131, opened 2021-12-17
feat: support more than 2 AZs on enviroment creation 

### LINKED ISSUE #780: Support for 3 AZ infrastructure

### LINKED ISSUE #1298: Feature Request: Specify Availability Zones when initializing environment
First off huge thanks for Copilot.

When I ran `copilot env init` with Load Balanced Web Service, copilot failed with the following error:
```
The following Availability Zones ap-northeast-1a cannot be associated with a load balancer
```
CloudFormation generated by copilot logged the same error.
CLI Version: 0.3.0

I figured out that `ap-northeast-1a` is mapped to `apne1-az3` and seems to have no support for Application Load Balancer.
It would be neccessary to have the options to specify Availability Zones to avoid this kind of error.

Workaround: Create a VPC and subnets in your own and select `No, I'd like to import existing resources (VPC, subnets)`.

### LINKED ISSUE #2260: Copilot should be able to deploy to multiple AZs
As a copilot user, I want to deploy Service to 3 Availability Zones with AWS Copilot, but it is pretty tough (see original request [here](https://dev.classmethod.jp/articles/aws-copilot-cannot-deploy-service-to-3az/):
- As a feature of Copilot, configuring ECS ​​tasks to deploy to 3AZ and 3 or more subnets is not supported.
- This is temporarily possible by directly updating the CloudFormation stack generated by Copilot.
- However, if you deploy the Service using Copilot after that, the stack will be overwritten and the subnet setting of the deployment destination will be returned.

efekarakus:
Resolves #1298, resolves #780, resolves #2260


internal/pkg/cli/env_init.go
```diff
@@ -545,6 +567,40 @@ func (o *initEnvOpts) askAdjustResources() error {
...
+func (o *initEnvOpts) askAZs() ([]string, error) {
+	if o.adjustVPC.AZs != nil {
+		return o.adjustVPC.AZs, nil
+	}
+	if o.ec2Client == nil {
+		o.ec2Client = ec2.New(o.sess)
+	}
+	azs, err := o.ec2Client.ListAZs()
+	if err != nil {
+		return nil, fmt.Errorf("list availability zones for region %s: %v", aws.StringValue(o.sess.Config.Region), err)
```
iamhopaul123, line 579: why not use `%w` to wrap the error?

internal/pkg/term/prompt/prompt.go
```diff
@@ -330,6 +344,11 @@ func RequireNonEmpty(v interface{}) error {
 	return survey.Required(v)
 }
 
+// RequireMinItems enforces at least min elements to be selected from MultiSelect.
+func RequireMinItems(min int) ValidatorFunc {
```
iamhopaul123, line 348: nice!

iamhopaul123 (approved):
LGTM!

internal/pkg/cli/env_init.go
efekarakus: It's really a tiny difference, when we use `%w` that means that the underlying error is exposed part of the `cli` pkg's interface. Since we don't expect any client to check the underlying ec2 error, I've just marked it as `%v`. It doesn't really matter for us :D

huanjani (approved):
The validator as an argument of MultiSelect is cool.

PR #3134, opened 2021-12-17
refactor(describe): re-order file components

huanjani:
Small housekeeping task of changing the order of the describe/service.go file's elements. 


internal/pkg/describe/service.go
```diff
@@ -358,3 +255,106 @@ func formatAppRunnerUrl(serviceURL string) string {
...
+// ServiceConfig contains serialized configuration parameters for a service.
+type ServiceConfig struct {
+	Environment string `json:"environment"`
+	Port        string `json:"port"`
+	CPU         string `json:"cpu"`
+	Memory      string `json:"memory"`
+	Platform    string `json:"platform,omitempty"`
+}
+
+type ECSServiceConfig struct {
```
efekarakus: Should we move these two and their methods to the top of the file since they are exported?

iamhopaul123 (approved):

internal/pkg/describe/service.go
huanjani: I moved them above env vars but below describers!

efekarakus (approved):

PR #3137, opened 2021-12-20
test(e2e): fix adjusted env by accepting default AZ options

efekarakus:

### DIFF
--- a/e2e/customized-env/customized_env_suite_test.go
+++ b/e2e/customized-env/customized_env_suite_test.go
@@ -41,6 +41,7 @@ var _ = BeforeSuite(func() {
 	appName = fmt.Sprintf("e2e-customizedenv-%d", time.Now().Unix())
 	vpcConfig = client.EnvInitRequestVPCConfig{
 		CIDR:               "10.1.0.0/16",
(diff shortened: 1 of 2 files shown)


e2e/customized-env/customized_env_suite_test.go
```diff
@@ -41,6 +41,7 @@ var _ = BeforeSuite(func() {
 	appName = fmt.Sprintf("e2e-customizedenv-%d", time.Now().Unix())
 	vpcConfig = client.EnvInitRequestVPCConfig{
 		CIDR:               "10.1.0.0/16",
+		AZs:                `""`, // Use default AZ options.
```
iamhopaul123, line 44: Why does it have to be `""` instead of an empty string 🤔

efekarakus: I wasn't sure how the `os/exec` pkg would behave if the argument is the empty string, I think it will do `--override-az-names --override-public-cidrs` instead of `--override-az-names "" --override-public-cidrs` that's why I explicitly marked it

iamhopaul123 (approved):

huanjani (approved):

PR #3138, opened 2021-12-20
feat: enable env file and add pipeline support

iamhopaul123:
part of #1366

### DIFF
--- a/internal/pkg/manifest/job.go
+++ b/internal/pkg/manifest/job.go
@@ -136,6 +136,11 @@ func (j *ScheduledJob) BuildRequired() (bool, error) {
 	return requiresBuild(j.ImageConfig.Image)
 }
 
+// EnvFile returns the location of the env file against the ws root directory.
+func (j *ScheduledJob) EnvFile() string {
+	return j.TaskConfig.EnvFile
+}
+
 // JobDockerfileBuildRequired returns if the job container image should be built from local Dockerfile.
 func JobDockerfileBuildRequired(job interface{}) (bool, error) {
 	return dockerfileBuildRequired("job", job)
--- a/internal/pkg/manifest/job.go
+++ b/internal/pkg/manifest/job.go
@@ -138,7 +138,7 @@ func (j *ScheduledJob) BuildRequired() (bool, error) {
 
 // EnvFile returns the location of the env file against the ws root directory.
 func (j *ScheduledJob) EnvFile() string {
-	return j.TaskConfig.EnvFile
+	return aws.StringValue(j.TaskConfig.EnvFile)
 }
 
 // JobDockerfileBuildRequired returns if the job container image should be built from local Dockerfile.
--- a/internal/pkg/manifest/validate.go
+++ b/internal/pkg/manifest/validate.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"path/filepath"
 	"regexp"
 	"sort"
 	"strconv"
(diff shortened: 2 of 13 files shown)


internal/pkg/manifest/workload.go
```diff
@@ -509,6 +509,7 @@ type TaskConfig struct {
 	Count          Count                `yaml:"count"`
 	ExecuteCommand ExecuteCommand       `yaml:"exec"`
 	Variables      map[string]string    `yaml:"variables"`
+	EnvFile        string               `yaml:"env_file"`
```
Lou1415926: A question! I think `string` totally works, but what do you think of `*string` for consistency?

internal/pkg/manifest/job.go
```diff
@@ -136,6 +136,11 @@ func (j *ScheduledJob) BuildRequired() (bool, error) {
 	return requiresBuild(j.ImageConfig.Image)
 }
 
+// EnvFile returns the location of the env file against the ws root directory.
+func (j *ScheduledJob) EnvFile() string {
```
Lou1415926, line 140: I wonder what do you think of  `func (t *TaskConfig) EnvFile() string`? Is it because RDWS doesn't support this that we don't want to use `TaskConfig` as the receiver?

internal/pkg/manifest/validate.go
```diff
@@ -27,6 +28,8 @@ const (
 	// Min and Max values for task ephemeral storage in GiB.
 	ephemeralMinValueGiB = 20
 	ephemeralMaxValueGiB = 200
+
+	validEnvFileExt = ".env"
```
```suggestion
	envFileExt = ".env"
```

internal/pkg/template/templates/cicd/buildspec.yml
```diff
@@ -84,6 +84,29 @@ phases:
...
+          type=$(echo $manifest | jq -r '.type')
+          if [ "$type" = "Request-Driven Web Service" ]; then
+            echo "skipping env file uploading for Request-Driven Web Service $workload";
+            continue
+          fi
+          for env in $pl_envs; do
+            env_files=$(echo $manifest | jq --arg env "$env" '{base_env_file: .env_file, env_env_file: (.environments? | .[$env]? | .env_file?) }')
+            env_file=$(echo $env_files | jq -r 'if (.env_env_file == null) then .base_env_file else .env_env_file end')
+            ENVFILE="./$env_file"
+            if [ -f "$ENVFILE" ]; then{{range $bucket := .ArtifactBuckets}}{{range $envName := $bucket.Environments}}
```
Lou1415926, line 100: Does this mean we only upload the files for environment that has been registered at the time of `pipeline init`?

huanjani (approved):
Nice work, esp on the dense, tricky buildspec logic!

internal/pkg/manifest/job.go
iamhopaul123: Good idea. I'll make the change

iamhopaul123: Oh actually it's for the ease of this
```
	type envFile interface {
		EnvFile() string
	}
	mf, ok := unmarshaledManifest.(envFile)
	if ok {
		return mf.EnvFile()
	}
```
And the same reason for `BuildArgs()`

internal/pkg/template/templates/cicd/buildspec.yml
iamhopaul123: Yeah that's correct.

Lou1415926 (approved):

PR #3142, opened 2021-12-21
chore: enable NLB, TLS termination, and NLB aliases

Lou1415926:
This PR:
1. Enable rendering NLB template if `nlb` is specified in manifest
2. Enable TLS termination, using a certificate managed by Copilot custom resource that is dedicated to the service
3. Enable alias for NLB

### DIFF
--- a/cf-custom-resources/lib/nlb-cert-validator-updater.js
+++ b/cf-custom-resources/lib/nlb-cert-validator-updater.js
@@ -652,9 +652,9 @@ async function inUseByOtherServices(loadBalancerDNS, domainName, route53Client) 
         ({hostedZoneID} = await domainResources(domainName));
     } catch (err) {
         if (err instanceof UnrecognizedDomainTypeError) {
-            console.log(`Found ${domainName} in subject alternative names. 
-It does not match any of these patterns: ".<env>.<app>.<domain>"， ".<app>.<domain>" or ".<domain>". 
-This is unexpected. We don't error out as it may not cause any issue.`);
+            console.log("Found ${domainName} in subject alternative names. " +
+"It does not match any of these patterns: '.<env>.<app>.<domain>'， '.<app>.<domain>' or '.<domain>'. " +
+"This is unexpected. We don't error out as it may not cause any issue.");
             return true; // This option has unrecognized pattern, we can't check if it is in use, so we assume it is in use.
         }
         throw err;
--- a/internal/pkg/cli/svc_deploy.go
+++ b/internal/pkg/cli/svc_deploy.go
@@ -626,7 +626,17 @@ func (o *deploySvcOpts) stackConfiguration() (cloudformation.StackConfiguration,
 				return nil, err
 			}
 			opts = append(opts, stack.WithHTTPS())
-			opts = append(opts, stack.WithDNSDelegation())
+
+			var caller identity.Caller
+			caller, err = o.identity.Get()
+			if err != nil {
+				return nil, fmt.Errorf("get identity: %w", err)
+			}
+			opts = append(opts, stack.WithDNSDelegation(deploy.AppInformation{
+				Name:                o.targetEnvironment.App,
+				DNSName:             o.targetApp.Domain,
+				AccountPrincipalARN: caller.RootUserARN,
+			}))
 		}
 		if !t.NLBConfig.IsEmpty() {
 			cidrBlocks, err := o.publicCIDRBlocks()
--- a/internal/pkg/cli/svc_deploy.go
+++ b/internal/pkg/cli/svc_deploy.go
@@ -606,7 +606,7 @@ func (o *deploySvcOpts) stackConfiguration() (cloudformation.StackConfiguration,
 	var conf cloudformation.StackConfiguration
 	switch t := mft.(type) {
 	case *manifest.LoadBalancedWebService:
-		if o.targetApp.Domain == "" && (!t.Alias.IsEmpty() || !t.NLBConfig.Aliases.IsEmpty()) {
+		if o.targetApp.Domain == "" && t.HasAliases() {
 			log.Errorf(aliasUsedWithoutDomainFriendlyText)
 			return nil, errors.New("alias specified when application is not associated with a domain")
 		}
--- a/internal/pkg/cli/svc_deploy.go
+++ b/internal/pkg/cli/svc_deploy.go
@@ -606,7 +606,7 @@ func (o *deploySvcOpts) stackConfiguration() (cloudformation.StackConfiguration,
 	var conf cloudformation.StackConfiguration
 	switch t := mft.(type) {
 	case *manifest.LoadBalancedWebService:
-		if o.targetApp.Domain == "" && !t.Alias.IsEmpty() {
+		if o.targetApp.Domain == "" && (!t.Alias.IsEmpty() || !t.NLBConfig.Aliases.IsEmpty()) {
 			log.Errorf(aliasUsedWithoutDomainFriendlyText)
 			return nil, errors.New("alias specified when application is not associated with a domain")
 		}
@@ -614,8 +614,6 @@ func (o *deploySvcOpts) stackConfiguration() (cloudformation.StackConfiguration,
 		var opts []stack.LoadBalancedWebServiceOption
 
 		// TODO: https://github.com/aws/copilot-cli/issues/2918
-		// 1. Should error out if `nlb.alias` is specified with targetApp.Domain == ""
-		// 2. Should validate `nlb.alias` is specified
 		// 3. ALB block should not be executed if http is disabled
 		if o.targetApp.RequiresDNSDelegation() {
 			var appVersionGetter versionGetter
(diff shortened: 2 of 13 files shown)


cf-custom-resources/lib/nlb-cert-validator-updater.js
```diff
@@ -652,9 +652,9 @@ async function inUseByOtherServices(loadBalancerDNS, domainName, route53Client)
         ({hostedZoneID} = await domainResources(domainName));
     } catch (err) {
         if (err instanceof UnrecognizedDomainTypeError) {
-            console.log(`Found ${domainName} in subject alternative names. 
-It does not match any of these patterns: ".<env>.<app>.<domain>"， ".<app>.<domain>" or ".<domain>". 
-This is unexpected. We don't error out as it may not cause any issue.`);
+            console.log("Found ${domainName} in subject alternative names. " +
```
efekarakus: I don't think expressions (like `${domainName}` get evaluated) if we use double quotes, it has to be a backtick afaik:
 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

internal/pkg/cli/svc_deploy.go
```diff
@@ -606,16 +606,14 @@ func (o *deploySvcOpts) stackConfiguration() (cloudformation.StackConfiguration,
 	var conf cloudformation.StackConfiguration
 	switch t := mft.(type) {
 	case *manifest.LoadBalancedWebService:
-		if o.targetApp.Domain == "" && !t.Alias.IsEmpty() {
+		if o.targetApp.Domain == "" && (!t.Alias.IsEmpty() || !t.NLBConfig.Aliases.IsEmpty()) {
```
efekarakus: nit: maybe we can have a method attached to `*LoadBalancedWebService` called `HasAliases` that encapsulates this condition

internal/pkg/cli/svc_deploy.go
```diff
@@ -625,8 +623,21 @@ func (o *deploySvcOpts) stackConfiguration() (cloudformation.StackConfiguration,
 			if err = validateLBSvcAliasAndAppVersion(aws.StringValue(t.Name), t.Alias, o.targetApp, o.envName, appVersionGetter); err != nil {
 				return nil, err
 			}
+			if err = validateLBSvcAliasAndAppVersion(aws.StringValue(t.Name), t.NLBConfig.Aliases, o.targetApp, o.envName, appVersionGetter); err != nil {
```
efekarakus, line 626: Hmm this is fine! but I wonder if there is a better package for us to put these runtime validations 💭 

Like the `manifest` pkg has static validations, It'd be neat if we can figure out a way of pruning the `svc deploy` cmd although I don't know how off the top of my head

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -20,6 +21,7 @@ const (
 	lbWebSvcRulePriorityGeneratorPath = "custom-resources/alb-rule-priority-generator.js"
 	desiredCountGeneratorPath         = "custom-resources/desired-count-delegation.js"
 	envControllerPath                 = "custom-resources/env-controller.js"
+	nlbCertManagerPath                = "custom-resources/nlb-cert-validator-updater.js"
```
efekarakus, line 24: nit: maybe in a separate revision or PR we can rename the custom resource to 
```suggestion
	nlbCertManagerPath                = "custom-resources/nlb-cert-manager.js"
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -256,6 +258,61 @@ func convertHTTPHealthCheck(hc *manifest.HealthCheckArgsOrString) template.HTTPH
 	return opts
 }
 
+func (s *LoadBalancedWebService) convertNetworkLoadBalancer() (*template.NetworkLoadBalancer, error) {
```
efekarakus: I wonder if we should return a struct here that has all the NLB related functionality to make the `Template` method more readable:
```go
type nlbConfig struct {
   settings *template.NetworkLoadBalancer
   
   // If a domain is associated these values are not empty.
   dnsRole           *string
   dnsName           *string
   certManagerLambda string
}

func (s *LoadBalancedWebService) convertNetworkLoadBalancer() (nlbConfig, error)

```

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -159,18 +158,14 @@ NLBCertManagerRole:
...
     Policies:
-      - PolicyName: "NLBCertManagerPolicyWrite"
+      - PolicyName: "NLBCertManagerPolicy"
         PolicyDocument:
           Version: '2012-10-17'
           Statement:
             - Sid: EnvHostedZoneUpdateAndWait
               Effect: Allow
               Action: route53:ChangeResourceRecordSets
-              Resource:
```
efekarakus: porque?

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
```diff
@@ -29,6 +29,16 @@
 {{- if eq .WorkloadType "Load Balanced Web Service"}}
   PortMappings:
     - ContainerPort: !Ref ContainerPort
+{{- if .NLB}}
+  {{if ne .NLB.Listener.TargetPort .NLB.MainContainerPort}} {{/*No need to add additional port if the target port is the same as image port*/}}
```
efekarakus, line 33: am I understanding this right `.NLB.MainContainerPort` is equal to `s.manifest.ImageConfig.Port`? If so do we need `.NLB.MainContainerPort`?

Lou1415926: It is equal, but `s.manifest.ImageConfig.Port` is only passed to the parameter `ContainerPort`, and not to `template.Workload`. So basically we need to pass it to the template data somehow. I put it under the `NLB` struct only because it is only needed when `nlb` is enabled.

internal/pkg/deploy/cloudformation/stack/transformers.go
Lou1415926: Updated boss!

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
Lou1415926: This was a mistake during rebasing. I thought this was the  "route53:GetChange" policy statement, which needs `Resource: "*"`. Fixed!

internal/pkg/cli/svc_deploy.go
```diff
@@ -606,16 +606,14 @@ func (o *deploySvcOpts) stackConfiguration() (cloudformation.StackConfiguration,
...
 			log.Errorf(aliasUsedWithoutDomainFriendlyText)
 			return nil, errors.New("alias specified when application is not associated with a domain")
 		}
 
 		var opts []stack.LoadBalancedWebServiceOption
 
 		// TODO: https://github.com/aws/copilot-cli/issues/2918
-		// 1. Should error out if `nlb.alias` is specified with targetApp.Domain == ""
-		// 2. Should validate `nlb.alias` is specified
 		// 3. ALB block should not be executed if http is disabled
```
iamhopaul123:
```suggestion
		// 1. ALB block should not be executed if http is disabled
```

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
iamhopaul123: We could potentially have an additional condition like:
```yaml
  AdditionalNLBPort:
    !Not [ !Equals [!Ref ContainerPort, {{ .NLB.Listener.TargetPort }}] ]
```

internal/pkg/cli/svc_deploy.go
Lou1415926: This comment is getting removed in the next PR but I committed the suggestion

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
Lou1415926: Wouldn't it look confusing to user, having this parameter?

iamhopaul123: hmm either works for me, but it is actually only a condition not a parameter.

Lou1415926: ya right typo!

I just have general concern in adding stuff to the template vs. changing how the template is rendered. This works, I'm just not sure if it's preferrable

iamhopaul123: If we add an additional param then we need to make sure they are always synced up with each other, which is kinda annoying. That's not a problem now but let's say in the future if we allow them to reuse the template and pass in whatever values to some params, then it would potentially be a problem (though we could fix later as well).

But I agree with you: adding an additional condition does not sound to be an elegant solution to me as well.

iamhopaul123 (approved):

efekarakus (approved):

Issue #3162, opened 2022-01-03
You cannot associate an existing service/application with a domain after it's already been created by Copilot CLI and deployed


glynjackson:
The first few hours playing with Copilot, so impressive, but I have run into one issue when trying to add a domain and alias. It would appear I cannot associate an existing service/application with a domain after it's already been created by Copilot CLI and deployed. 

_Apologies in advance if some of my terminology isn't correct._

- Running v1.13.0
- Using version 1.13.0 I created a new initial service and application and deployed it successfully (without initially using the --domain flag) 
- Later, I added `alias` in the existing manifest and got the following error message: 

```
✘ To use `http.alias`, your application must be associated with a domain: `copilot app init --domain example.com`.
✘ execute svc deploy: alias specified when application is not associated with a domain
```

- Following the CLI message above  I run  `copilot app init --domain api.mydomain.com` and redeploy, but it still gives the same error message and appeared to have no effect.

I read the documentation and noticed the following wording states...

```
As mentioned in the Application Guide, you can configure the domain name of your app when running copilot app init
``` 

[Ref](https://aws.github.io/copilot-cli/docs/developing/domain/)

My question:

- Can you create an alias for an existing application as the CLI output messages would lead me to believe, and if so, then is this a bug?
-  If you can't create an alias for an existing application after its creation then is the CLI message misleading?

efekarakus:
Hi @glynjackson ! Apologies for the late reply and thank you for the detailed issue description 😍 

I believe this is the same issue as https://github.com/aws/copilot-cli/issues/3045 and you're right unfortunately today you can't associate a domain once the application is created 🙇 
For now users are forced to create a second application with `--domain` with the exact same set of services and then run `copilot app delete` on the old application to delete the resources.

I agree the UX is confusing, we'll add a notice to the docs saying that the feature is supported only while creating the app the first time and can't be updated. We'll also track the feature request's status in #3045, if you wouldn't mind giving a 👍 there!

efekarakus:
Created a PR to add the notice here: https://github.com/aws/copilot-cli/pull/3164, once merged it should become public with our next release!

glynjackson:
Hi @efekarakus 

Thank you for the explanation, it now makes sense and resolves my initial confusion.

#3045 thumbs up given.

Also, thank you for PR #3164 clarifying the domain association, it really helps and I'm sure it will help others coming to Copilot for the first time too. 

[...]

PR #3164, opened 2022-01-04
docs: add notice clarifying domain association

efekarakus:
Related #3162


site/content/docs/developing/domain.en.md
```diff
@@ -1,5 +1,11 @@
 # Domain
 
+!!!attention
+    Today, a route53 domain name can be associated only while creating the app for the first time.
```
huanjani:
```suggestion
    Today, a Route 53 domain name can be associated only when running `app init`.  
```
or something like that? "for the first time" seems misleading because we then guide them into the solution, creating it a second time. 😆

site/content/docs/developing/domain.en.md
```diff
@@ -1,5 +1,11 @@
 # Domain
 
+!!!attention
+    Today, a route53 domain name can be associated only while creating the app for the first time.  
+    If you'd like to update your application with a domain ([#3045](https://github.com/aws/copilot-cli/issues/3045)),
```
huanjani, line 5: 🤔  Is it weird to link to a customer-opened GH issue in our official docs?

huanjani (approved):
Just a couple questions. Approving with the DNM label for efficiency given our asynchrony. 😺

site/content/docs/developing/domain.en.md
```diff
@@ -21,9 +27,6 @@ Currently, you can only use aliases under the domain you specified when creating
 
 We'll make this feature more powerful in the future by allowing you to import certificates and use any aliases!
 
-!!!info
```
iamhopaul123, line 24: Why do we get rid of this info block? I think the rule still applies right?

site/content/docs/developing/domain.en.md
```diff
@@ -34,9 +37,6 @@ http:
   alias: example.aws
 ```
 
-!!!info
```
iamhopaul123, line 37: same here

site/content/docs/developing/domain.en.md
efekarakus: Done!

site/content/docs/developing/domain.en.md
efekarakus: I think it's fine 🤷 I wanted to link it so that it can generate more 👍 from users

site/content/docs/developing/domain.en.md
efekarakus: The info blocks felt like implementation detail that might not be interesting for users, that's why I removed it. Let me know if you disagree!

iamhopaul123 (approved):

PR #3157, opened 2021-12-29
chore: split NLB certificate manager into two different custom resources

Lou1415926:
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the Apache 2.0 License.

PR #3166, opened 2022-01-04
chore: add NLB cert validator and custom domain to CFN template

Lou1415926:
This PR:
1. Upload NLB cert validator and custom domain lambda scripts to S3 bucket, if DNS delegation is enabled. This is because if we pass the scripts as code into the CFN template, the two scripts together would make the template too large -> so we need to upload them to s3 and refer to the s3 location instead.
2. Add cert validator custom resource and custom domain custom resource to NLB template, conditionally.
3. Add supporting resources to the custom resource, such as roles to invoke lambda.
4. Removed the outdated script from  #3157.

Preceding PR: #3157 


internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -55,6 +54,7 @@ type LoadBalancedWebService struct {
 	dnsDelegationEnabled   bool
 	publicSubnetCIDRBlocks []string
 	appInfo                deploy.AppInformation
+	nlbCustomResourceS3URL map[string]string
```
huanjani: nit:
```suggestion
	nlbCustomResourceS3URLs map[string]string
```

iamhopaul123 (approved):
LGTM 🎉

huanjani (approved):
🚢

PR #3171, opened 2022-01-06
chore: allow disabling ALB by specifying `false` for `http`

Lou1415926:
Preceding PR #3166


internal/pkg/cli/svc_package.go
```diff
@@ -123,14 +124,71 @@ func newPackageSvcOpts(vars packageSvcVars) (*packageSvcOpts, error) {
...
 					return nil, err
 				}
-				opts = append(opts, stack.WithHTTPS())
+				options = append(options, stack.WithNLB(cidrBlocks))
+				if app.RequiresDNSDelegation() {
+					resources, err := opts.appCFN.GetAppResourcesByRegion(app, env.Region)
+					if err != nil {
+						return nil, err
+					}
+					urls, err := uploadNLBWSCustomResources(&uploadCustomResourcesOpts{
```
iamhopaul123: Just a callout here: we don't upload anything in `svc package` such as for addons and build and upload their image. However, I think we should because otherwise the parameter file we generate isn't complete. Maybe we can leave as is and i'll do some refactoring later again anyway.

internal/pkg/cli/svc_deploy.go
```diff
@@ -386,15 +386,15 @@ func (o *deploySvcOpts) configureClients() error {
 	return nil
 }
 
-func (o *deploySvcOpts) publicCIDRBlocks() ([]string, error) {
-	envDescription, err := o.envDescriber.Describe()
+func publicCIDRBlocks(envDescriber envDescriber, subnetLister vpcSubnetLister, envName string) ([]string, error) {
```
iamhopaul123: The function name and parameters look a little bit strange as `publicCIDRBlocks` is just too bottom layer as a function in `cli` package, and `envDescriber` is already inited with the env name. Can we move [this whole code block](https://github.com/aws/copilot-cli/pull/3171/files#diff-887bb70e74d71d9ecbfe078d768a70bdd2f02043501ba27c8b254ac3a102e799R636-R668) to its own function instead of having `publicCIDRBlocks` in `svc_deploy`?

internal/pkg/cli/svc_deploy.go
```diff
@@ -622,41 +622,18 @@ func (o *deploySvcOpts) stackConfiguration() (cloudformation.StackConfiguration,
...
 	switch t := mft.(type) {
 	case *manifest.LoadBalancedWebService:
-		if o.targetApp.Domain == "" && t.HasAliases() {
-			log.Errorf(aliasUsedWithoutDomainFriendlyText)
-			return nil, errors.New("alias specified when application is not associated with a domain")
+		var appVersionGetter versionGetter
+		appVersionGetter, err = o.newAppVersionGetter(o.appName)
+		if err != nil {
+			return nil, fmt.Errorf("new app describer for application %s: %w", o.appName, err)
+		}
```
iamhopaul123, line 594: Can we not put this in `validateLBWSRuntime` as well? I think ideally the only diff between `deploy` and `package` should be `deploy` deploys directly through CFN while `package` generates a ready-to-deploy artifact (CFN template and parameter file).

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -304,15 +297,30 @@ func (s *LoadBalancedWebService) Parameters() ([]*cloudformation.Parameter, erro
...
+			{
+				ParameterKey:   aws.String(LBWebServiceHTTPSParamKey),
+				ParameterValue: aws.String(strconv.FormatBool(s.httpsEnabled)),
+			},
+			{
+				ParameterKey:   aws.String(LBWebServiceStickinessParamKey),
+				ParameterValue: aws.String(strconv.FormatBool(aws.BoolValue(s.manifest.RoutingRule.Stickiness))),
+			},
+		}
+		wkldParams = append(wkldParams, httpParams...)
```
iamhopaul123: It seems more straightforward to me to have
```suggestion
		wkldParams = append(wkldParams, []*cloudformation.Parameter{...}...)
```
🤔

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -104,11 +105,43 @@ func newDefaultLoadBalancedWebService() *LoadBalancedWebService {
...
+			Network: NetworkConfig{
+				VPC: vpcConfig{
+					Placement: &PublicSubnetPlacement,
 				},
 			},
+		},
+	}
+}
+
+// newDefaultEmptyLoadBalancedWebService returns an empty LoadBalancedWebService with only the default values set, without any load balancer configured.
```
iamhopaul123: nit
```suggestion
// newDefaultEmptyLoadBalancedWebService returns an empty LoadBalancedWebService with only the default values set, without any load balancer configuration.
```

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -104,11 +105,43 @@ func newDefaultLoadBalancedWebService() *LoadBalancedWebService {
...
+func newDefaultEmptyLoadBalancedWebService() *LoadBalancedWebService {
```
iamhopaul123: Can we try to reduce the duplication between `newDefaultEmptyLoadBalancedWebService` and `newDefaultLoadBalancedWebService`?

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -199,8 +232,42 @@ func (s LoadBalancedWebService) ApplyEnv(envName string) (WorkloadManifest, erro
...
 
-// RoutingRule holds the path to route requests to the service.
-type RoutingRule struct {
+// RoutingRuleConfigOrBool holds advanced configuration for routing rule or a boolean switch.
+type RoutingRuleConfigOrBool struct {
+	RoutingRuleConfiguration
+	Enabled *bool
+}
+
+func (r *RoutingRuleConfigOrBool) Disabled() bool {
```
iamhopaul123, line 218: comment pls 🙏

internal/pkg/manifest/transform.go
```diff
@@ -326,6 +327,31 @@ func (q sqsQueueOrBoolTransformer) Transformer(typ reflect.Type) func(dst, src r
 	}
 }
 
+type routingRuleConfigOrBoolTransformer struct{}
```
iamhopaul123, line 330: nice! it almost slips my mind 🤦

internal/pkg/manifest/validate.go
```diff
@@ -85,6 +85,12 @@ func (l LoadBalancedWebService) Validate() error {
 // Validate returns nil if LoadBalancedWebServiceConfig is configured correctly.
 func (l LoadBalancedWebServiceConfig) Validate() error {
 	var err error
+	if l.RoutingRule.Disabled() && l.NLBConfig.IsEmpty() {
+		return errors.New(`must have one of "http" or "nlb" enabled`)
```
iamhopaul123: maybe we can use `errFieldMutualExclusive` with `mustExist` on?

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
```diff
@@ -41,23 +36,32 @@ Parameters:
     Type: String
   TargetPort:
     Type: Number
+{{- if not .HTTPDisabled}}
```
iamhopaul123: I wonder if it is more like `ALBDisabled` since `HTTPDisabled` might include `HTTPS` can be used

internal/pkg/template/workload.go
```diff
@@ -531,7 +533,10 @@ func randomUUIDFunc() (string, error) {
 func envControllerParameters(o WorkloadOpts) []string {
 	parameters := []string{}
 	if o.WorkloadType == "Load Balanced Web Service" {
-		parameters = append(parameters, []string{"ALBWorkloads,", "Aliases,"}...) // YAML needs the comma separator; resolved in EnvContr.
+		if !o.HTTPDisabled {
```
iamhopaul123: Nice! Good catch.

iamhopaul123 (commented):
Phew seems like supporting disabling ALB is never an easy thing 😢 Good work!

internal/pkg/cli/svc_deploy.go
Lou1415926: Do you mean not passing `appVersionGetter` as a param to `validateLBWSRuntime`?

iamhopaul123: oh nvm sorry i think this looks good!

internal/pkg/cli/svc_deploy.go
```diff
@@ -959,6 +948,28 @@ func validateAppVersion(appName string, appVersionGetter versionGetter) error {
 	return nil
 }
 
+func validateLBWSRuntime(app *config.Application, envName string, t *manifest.LoadBalancedWebService, appVersionGetter versionGetter) error {
```
```suggestion
func validateLBWSRuntime(app *config.Application, envName string, mft *manifest.LoadBalancedWebService, appVersionGetter versionGetter) error {
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -959,6 +948,28 @@ func validateAppVersion(appName string, appVersionGetter versionGetter) error {
...
+		}
+	}
+
+	if err := validateLBSvcAlias(t.RoutingRule.Alias, app, envName); err != nil {
+		return err
+	}
+	if err := validateLBSvcAlias(t.NLBConfig.Aliases, app, envName); err != nil {
+		return err
+	}
+	return nil
```
```suggestion
	return validateLBSvcAlias(t.NLBConfig.Aliases, app, envName)
```

internal/pkg/cli/svc_package.go
efekarakus: I think to be consistent and merge this PR, we should not upload the custom resources here either and leave it blank.

Afterwards, we can do a the major change for `package`

efekarakus: Side note:
From perusing the CDK help menu, it looks like `cdk synthesize` gives deterministic endpoints for their assets. For example, although the command doesn't upload things to S3 it prepopulates the params with the hash (the same parameter value is then generated with `deploy`). 

I think we can do that for our S3 artifacts as well, but not for ECR's image digest :( For a digest it looks like the image needs to be in a repository, maybe we can force them to provide a `--tag/--digest` flag in this scenario or `--upload-assets`

internal/pkg/cli/svc_package.go
```diff
@@ -123,14 +124,71 @@ func newPackageSvcOpts(vars packageSvcVars) (*packageSvcOpts, error) {
 		var serializer stackSerializer
 		switch t := mft.(type) {
 		case *manifest.LoadBalancedWebService:
-			var opts []stack.LoadBalancedWebServiceOption
-			if app.RequiresDNSDelegation() {
-				if err := validateLBSvcAliasAndAppVersion(aws.StringValue(t.Name), t.Alias, app, env.Name, appVersionGetter); err != nil {
+			if err := validateLBWSRuntime(app, env.Name, t, appVersionGetter); err != nil {
```
efekarakus, line 127: @Lou1415926 @iamhopaul123 I'd love for us to propose a new package structure with the buildspec cleanup work so that these commands become lean.

It feels like most of the logic in these `case` blocks should move to the `deploy/cloudformation` pkg

internal/pkg/manifest/svc.go
```diff
@@ -292,7 +292,7 @@ type HTTPHealthCheckArgs struct {
 }
 
 func (h *HTTPHealthCheckArgs) isEmpty() bool {
-	return h.Path == nil && h.HealthyThreshold == nil && h.UnhealthyThreshold == nil &&
+	return h.Path == nil && h.SuccessCodes == nil && h.HealthyThreshold == nil && h.UnhealthyThreshold == nil &&
```
efekarakus, line 295: :mildpanic:

internal/pkg/manifest/validate.go
```diff
@@ -85,6 +85,12 @@ func (l LoadBalancedWebService) Validate() error {
...
+	}
+	if l.RoutingRule.Disabled() && (l.Count.AdvancedCount.Requests != nil || l.Count.AdvancedCount.ResponseTime != nil) {
+		return errors.New(`scaling based on "nlb" requests is not supported yet`)
```
```suggestion
		return errors.New(`scaling based on "nlb" requests or response time is not supported`)
```

internal/pkg/template/templates/workloads/partials/cf/envvars-common.yml
```diff
@@ -29,6 +29,8 @@
       {{- end}}
 {{- end}}{{- end}}
 {{- if eq .WorkloadType "Load Balanced Web Service"}}
+{{- if not .HTTPDisabled}}
```
efekarakus: Consider flipping the field name https://refactoring.com/catalog/removeDoubleNegative.html
```suggestion
{{- if .HTTPEnabled}}
```

internal/pkg/manifest/validate.go
Lou1415926: I'm going to use `errAtLeastOneFieldMustBeSpecified`!

internal/pkg/cli/svc_package.go
Lou1415926: That'd mean that until `package` is supported, people won't be able to use NLB through pipeline. Is that right?

I agree that we should keep it consistent. If we keep the "upload"s in `svc package`, and then later remove it after `pipeline package` is supported, it might be a breaking change for folks.

The downside is that between now and when `pipeline package` is done / when we use deterministic endpoints for assets, NLB isn't usable through pipeline.

Lou1415926: My proposal is to log an error message in `svc package` if it's run with an NLBWS that needs custom resources. In pipeline, it errors out and say "NLBWS using TLS or aliases isn't supported in pipeline yet". I will try to figure out a way to so that:
1. When people run `copilot svc package` manually, it doesn't error out and still produces a template; however, it spits out a friendly message that says the template doesn't work because the custom resource isn't uploaded.
2. In pipeline, it should fail at the build stage and says "NLBWS using TLS or aliases isn't supported in pipeline yet"

What do you think bosses?

iamhopaul123 (approved):

internal/pkg/cli/svc_package.go
efekarakus: Chatted w @iamhopaul123 earlier, but one thing that we can do to unblock this PR:
1. Update this PR to have NLB custom resources behave like other existing custom resources by minimizing and inlining the code in the CFN template
2. In order to avoid hitting the template body limit of 51KB (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html), we can create a separate PR that uploads the template to S3 instead (1MB limit):
   * Maybe under this path: `manual/templates/<service name>/<sha256 hash of template body>.yml`
   * We can merge the upload PR before this one in order to unblock this PR

internal/pkg/template/workload.go
```diff
@@ -214,6 +214,7 @@ type NetworkLoadBalancerListener struct {
 	TargetPort      string
 	SSLPolicy       *string
 	Aliases         []string
+	Stickness       *bool
```
iamhopaul123, line 217: Do we expose this in our manifest?

Lou1415926: This is a temporary relief for the sake of rendering the template without error. It was okay before because http couldn't be disabled and it was using `!Ref StickIness` which was actually the http stickIness. Now with http disabled, `!Ref StickIness` produces error because `Stickiness` as a parameter is taken out of the template. To avoid erroring out, I added this temporary relief.

The actual NLB health check is coming in the next PR.

internal/pkg/manifest/errors.go
```diff
@@ -100,7 +100,9 @@ func (e *errAtLeastOneFieldMustBeSpecified) Error() string {
 	for i, f := range e.missingFields {
 		quotedFields[i] = strconv.Quote(f)
 	}
-	return fmt.Sprintf(`must specify at least one of %s if "%s" is specified`,
-		english.WordSeries(quotedFields, "or"),
-		e.conditionalField)
+	errMsg := fmt.Sprintf("must specify at least one of %s", english.WordSeries(quotedFields, "or"))
+	if e.conditionalField != "" {
```
huanjani, line 104: 👏🏼

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -53,9 +53,11 @@ NLBTargetGroup:
       - Key: deregistration_delay.timeout_seconds
         Value: {{.DeregistrationDelay}}  # ECS Default is 300; Copilot default is 60.
 {{- if ne .NLB.Listener.Protocol "TLS"}}
+{{- if .NLB.Listener.Stickness }}
```
huanjani, line 56: nit: I think you are taking this out in the next PR, but in the future, maybe you could name a temporary placeholder something more obvious (like `TemporaryStickinessPlaceholder`-- I don't think I would have noticed it, or maybe would have thought it was a typo if PH hadn't asked about it! 😆

huanjani (approved):
Nice work!!!

PR #3173, opened 2022-01-07
chore: clarify UX for "location" field when Docker daemon isn't running

### LINKED ISSUE #3169: Fails deploy when no Docker is installed without helpful error message
When running `copilot init` without having Docker installed it seems to crash without properly notifying the user about the reason.

At the Dockerfile prompt it logs `Docker command is not found; Copilot won't build from a Dockerfile.` and then continues the prompt with `What's the location of the image to use? [? for help]`. No Dockerfiles are suggested but the prompt accepts text.
After entering the name of the Dockerfile to use (manually), it correctly sets the `image.location` in the manifest file and then  continues with the rest of the setup, asking to deploy a test envirionment. Upon approval, it then fails after some time at the `Linking account 085424649605 and region eu-central-1 to application pearup.` step with:
```
✘ Failed to link account <ACCOUNT_ID> and region <REGION> to application <APP>.
✘ deploy env test to application <APP>: adding test environment resources to application: operation 5 for stack set <APP>-infrastructure failed
```

Trying to deploy manually, it complains that no environment exists. Then running `copilot env init`, it fails with the same message as above.

Instead of failing with the above error, I would expect the CLI to recognise that Docker is not installed and then stop trying to create an environment.

I am running copilot version `v1.13.0` on `macOS 12.1 `.

efekarakus:
Resolves #3169


internal/pkg/cli/svc_init.go
```diff
@@ -77,6 +76,8 @@ You should set this to the port which your Dockerfile uses to communicate with t
 	svcInitPublisherPrompt     = "Which topics do you want to subscribe to?"
 	svcInitPublisherHelpPrompt = `A publisher is an existing SNS Topic to which a service publishes messages. 
 These messages can be consumed by the Worker Service.`
+
+	wkldInitImagePrompt = fmt.Sprintf("What's the %s ([registry/]repository[:tag|@digest]) of the image to use?", color.Emphasize("location"))
```
iamhopaul123, line 80: :+1:

internal/pkg/cli/svc_init.go
```diff
@@ -376,11 +377,11 @@ func (o *initSvcOpts) askImage() error {
 		return nil
 	}
 
-	var validator prompt.ValidatorFunc
+	validator := validateNotDockerfile
 	promptHelp := wkldInitImagePromptHelp
 	if o.wkldType == manifest.RequestDrivenWebServiceType {
 		promptHelp = wkldInitAppRunnerImagePromptHelp
-		validator = validateAppRunnerImage
+		validator = chainValidators(validator, validateAppRunnerImage)
```
iamhopaul123: Why not
```suggestion
		validator = chainValidators(validateNotDockerfile, validateAppRunnerImage)
```
🤔

internal/pkg/cli/validate.go
```diff
@@ -836,3 +850,14 @@ func validateSecretName(val interface{}) error {
 	}
 	return nil
 }
+
+func validateNotDockerfile(val interface{}) error {
+	s, ok := val.(string)
+	if !ok {
+		return errValueNotAString
+	}
+	if strings.HasSuffix(s, "Dockerfile") {
```
iamhopaul123: Hmm does "a location has `Dockerfile` as suffix" equal to "they provide a local Dockerfile location"? I feel we could just update the prompt message...

efekarakus: Sounds good, I've reverted the other changes and kept only the log. I agree it feels a bit unnecessary after the prompt msg update

internal/pkg/cli/svc_init.go
efekarakus: It doesn't matter anymore, but I think the change above is cleaner because it's like Go's `append` function where you can keep adding more validators on top of each other

iamhopaul123 (approved):

huanjani (approved):
:shipit:

PR #3174, opened 2022-01-10
docs: fix platform type

morugu:
Add Map to Platform type.
It is also available, as shown in the sample code.

| before | after |
|:---|:---:|
| image | image |


efekarakus (approved):
Thank you!

huanjani (approved):

huanjani:
Thanks so much for fixing this, @morugu!
If you wouldn't mind changing your commit to follow the Conventional Commits format (like `fix: add map to platform type`), then our checks will pass. Thank you!

Issue #3021, opened 2021-11-11
[Design] Pipeline manifest new source pattern


iamhopaul123:
## Problem

Right now our pipeline manifest repo field is as follow:

```yaml
source:
  provider: GitHub
  properties:
    branch: main
    repository: https://github.com/iamhopaul123/copilot-xray-example
```

Problem is there are all kinds of repo format given that we now support GitHub, Bitbucket, and CodeCommit (or more in future). If the user never changes the default repository we write to the manifest, it is probably fine. However, since the repository we accept for now is actually usually the repo console url, if they put a repo URL in different format (for example `ssh://` or `codecommit::`), `pipeline update` would fail because we can't parse the info we need from the URL and use it in the CFN template.

## Proposal

Instead of writing the URL to the manifest, what we can do is to provide different field set for different provider type, which would be more intuitive to users and less error prone to us. We can utilize our manifest validation to make sure their manifest is configured correctly.

### GitHub && BitBucket

```yaml
source:
  provider: GitHub
  properties:
    branch: main
    owner: <owner>
    repo: <repo name>
```

### CodeCommit

```yaml
source:
  provider: CodeCommit
  properties:
    branch: main
    repo: <repo name>
    region: <region>
```

For backward compatibility concern, we can still keep the same behavior for `repository` field, but stop writing to the new manifest to transit to the new manifest format. Also for visibility concern, we can move the repo console url to our `pipeline show`, instead of showing the URL in the manifest.

PR #3182, opened 2022-01-11
feat(pipelines): add branch detection to `pipeline init`

### LINKED ISSUE #2105: Pipeline config uses branch main instead of master
### Description

When creating pipelines with `copilot pipeline init` and selecting the repository, the `pipeline.yaml` configuration file is created using the `main` default branch instead of asking for confirmation or determining the current branch. The problem is that old repositories are still using `master` as default branch, causing problems when deploying the pipeline with `copilot pipeline update`.

### Version

```shell
$ copilot version
version: v1.4.0, built for linux
```

huanjani:
Fixes https://github.com/aws/copilot-cli/issues/2105.

Changes:

- When using flags, `--url` and `--branch` must both be passed in
- We validate both `--url` and `--branch` when they're passed in, confirming that the remote repo exists and that the branch is a branch of the indicated repo, rather than failing way down the line at deployment time
- Without flags, after the user selects a remote repo for the pipeline to follow, they are presented with the branches detected (using `git branch -a -l '[repoShortName]/*`) of that repo, and asked to select the branch the pipeline should follow
- There is no more default branch that gets automatically injected into the pipeline manifest
- We no longer trim the `.git` off the back of repo URLs, as they would be included if passed in with the `--url` flag, so now that we are validating those values against the results of `git remote -v`, they should match
- I didn't refactor the repo URL parsing unit tests further because they will likely change with https://github.com/aws/copilot-cli/issues/3021


internal/pkg/cli/pipeline_init.go
```diff
@@ -158,8 +157,23 @@ func (o *initPipelineOpts) Validate() error {
 		return err
 	}
 
+	// URL flag and branch flag must both be set or not.
+	if URL, branch := o.repoURL == "", o.repoBranch == ""; URL != branch {
```
iamhopaul123: Why do they have to specify branch flag when using the repo URL? Like why not set default branch for them?

internal/pkg/cli/pipeline_init.go
```diff
@@ -225,6 +242,51 @@ func (o *initPipelineOpts) validateURL(url string) error {
 	return nil
 }
 
+func (o *initPipelineOpts) validateURLExists(url string) error {
```
iamhopaul123: Maybe we should mention it is for local git. Same for `o.fetchAndParseURLs`

internal/pkg/cli/pipeline_init.go
```diff
@@ -225,6 +242,51 @@ func (o *initPipelineOpts) validateURL(url string) error {
...
+		return fmt.Errorf("fetch and parse URLs: %w", err)
+	}
+	if _, ok := repos[url]; !ok {
+		return fmt.Errorf("URL '%s' is not a local git remote; please check that you're in the correct directory", o.repoURL)
+	}
+	o.repoShortName = repos[url]
+	return nil
+}
+
+func (o *initPipelineOpts) validateBranch() error {
```
iamhopaul123, line 243: Also this is only for validating local branches (it could be not existed remotely)

internal/pkg/cli/pipeline_init.go
```diff
@@ -268,6 +330,16 @@ func (o *initPipelineOpts) askRepository() error {
 	return nil
 }
 
+func (o *initPipelineOpts) askBranch() error {
+	var err error
+	if o.repoBranch == "" {
+		if err = o.selectBranch(); err != nil {
+			return err
+		}
```
iamhopaul123:
```suggestion
	if o.repoBranch == "" {
		if err := o.selectBranch(); err != nil {
			return err
		}
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -326,38 +392,62 @@ func (o *initPipelineOpts) parseBitbucketRepoDetails() error {
...
-		return fmt.Errorf("get remote repository info: %w; make sure you have installed Git and are in a Git repository", err)
+		return fmt.Errorf("fetch and parse URLs: %w", err)
 	}
-	urls, err := o.parseGitRemoteResult(strings.TrimSpace(o.buffer.String()))
-	if err != nil {
-		return err
+
+	var formattedRepos []string
+	for url, name := range repos {
+		formattedRepos = append(formattedRepos, fmt.Sprintf("%s: %s", name, url))
```
iamhopaul123: How about we make `%s: %s` to be a string type called maybe `pipelineRepo` with two methods `url()` and `shortName()` and a constructor method `newPipelineRepo()`?

internal/pkg/cli/pipeline_init.go
```diff
@@ -225,6 +242,51 @@ func (o *initPipelineOpts) validateURL(url string) error {
...
 
+func (o *initPipelineOpts) validateURLExists(url string) error {
+	repos, err := o.fetchAndParseURLs()
+	if err != nil {
+		return fmt.Errorf("fetch and parse URLs: %w", err)
+	}
+	if _, ok := repos[url]; !ok {
+		return fmt.Errorf("URL '%s' is not a local git remote; please check that you're in the correct directory", o.repoURL)
+	}
+	o.repoShortName = repos[url]
```
iamhopaul123: Isn't `repos[url]` the remote name instead of `repo short name`?

internal/pkg/cli/pipeline_init.go
```diff
@@ -374,24 +464,37 @@ func (o *initPipelineOpts) selectURL() error {
...
 		cols := strings.Split(item, "\t")
-		url := strings.TrimSpace(strings.TrimSuffix(strings.Split(cols[1], " ")[0], ".git"))
-		urlSet[url] = true
+		URL := strings.TrimSpace(strings.Split(cols[1], " ")[0])
+		repos[URL] = cols[0]
+	}
+	return repos
+}
+
+func (o *initPipelineOpts) parseGitBranchResults(s string) ([]string, error) {
```
iamhopaul123, line 524: Could you add an example of the input string and the parsed result?

internal/pkg/cli/pipeline_init.go
```diff
@@ -374,24 +464,37 @@ func (o *initPipelineOpts) selectURL() error {
...
+	var branches []string
+	branchSet := make(map[string]bool)
+	items := strings.Split(s, "\n")
+	for _, item := range items {
+		elements := strings.Split(item, "/")
+		if len(elements) < 3 {
+			return nil, fmt.Errorf("unparsable format")
```
iamhopaul123:
```suggestion
			return nil, fmt.Errorf("cannot parse branch name from %s, item)
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -374,24 +464,37 @@ func (o *initPipelineOpts) selectURL() error {
...
+	}
+	return repos
+}
+
+func (o *initPipelineOpts) parseGitBranchResults(s string) ([]string, error) {
+	var branches []string
+	branchSet := make(map[string]bool)
+	items := strings.Split(s, "\n")
+	for _, item := range items {
+		elements := strings.Split(item, "/")
```
iamhopaul123, line 528: Does it work for Windows?

internal/pkg/cli/pipeline_init.go
```diff
@@ -374,24 +464,37 @@ func (o *initPipelineOpts) selectURL() error {
...
-		urlSet[url] = true
+		URL := strings.TrimSpace(strings.Split(cols[1], " ")[0])
+		repos[URL] = cols[0]
+	}
+	return repos
+}
+
+func (o *initPipelineOpts) parseGitBranchResults(s string) ([]string, error) {
+	var branches []string
+	branchSet := make(map[string]bool)
```
iamhopaul123: Why do we need this?

internal/pkg/cli/pipeline_init.go
Lou1415926: <strike> I guess this is to be consistent with the guided experience? Like with guided experience, you have to select a branch from the prompts </strike> nvm we didn't ask for branch before. Then same questions as PH!

iamhopaul123: hmm then can we not throw an error but ask them for the branch name if branch flag is not set?

internal/pkg/cli/pipeline_init.go
```diff
@@ -225,6 +242,51 @@ func (o *initPipelineOpts) validateURL(url string) error {
 	return nil
 }
 
+func (o *initPipelineOpts) validateURLExists(url string) error {
+	repos, err := o.fetchAndParseURLs()
+	if err != nil {
+		return fmt.Errorf("fetch and parse URLs: %w", err)
+	}
+	if _, ok := repos[url]; !ok {
+		return fmt.Errorf("URL '%s' is not a local git remote; please check that you're in the correct directory", o.repoURL)
```
Lou1415926: I feel like this is more like a friendly message that could be put into a `log.Errorf` block, for error, maybe we could return something like `fmt.Errorf("url %s not found in git remote", o.repoURL)`

internal/pkg/cli/pipeline_init.go
```diff
@@ -326,38 +392,62 @@ func (o *initPipelineOpts) parseBitbucketRepoDetails() error {
...
 	return nil
 }
 
 func (o *initPipelineOpts) selectURL() error {
-	// Fetches and parses all remote repositories.
-	err := o.runner.Run("git", []string{"remote", "-v"}, exec.Stdout(&o.buffer))
+	repos, err := o.fetchAndParseURLs()
 	if err != nil {
-		return fmt.Errorf("get remote repository info: %w; make sure you have installed Git and are in a Git repository", err)
+		return fmt.Errorf("fetch and parse URLs: %w", err)
```
Lou1415926: nit:
```suggestion
		return err
```

internal/pkg/cli/pipeline_init.go
Lou1415926: Instead of formatting strings for `SelectOne`, we could take advantage of this [`SelectOption`](https://github.com/aws/copilot-cli/blob/1ba3833f7cda22ac876902148db02881dd1a5a23/internal/pkg/term/prompt/select.go#L55):

```
var options []Option
for url, name := range repos {
   options = append(options, Option{Value: url, Hint: name})
}

url, err := o.prompt.SelectOption(
pipelineSelectURLPrompt, 
pipelineSelectURLHelpPrompt, 
options,
prompt.WithFinalMessage("Repository:"),
```

and the selected `url` would be just the url, without the short name.

internal/pkg/cli/pipeline_init.go
```diff
@@ -326,38 +392,62 @@ func (o *initPipelineOpts) parseBitbucketRepoDetails() error {
...
-		prompt.WithFinalMessage("Repository URL:"),
+		formattedRepos,
+		prompt.WithFinalMessage("Repository:"),
 	)
 	if err != nil {
 		return fmt.Errorf("select URL: %w", err)
 	}
-	if err := o.validateURL(url); err != nil {
+	repoParts := strings.Split(url, ": ")
+	if err := o.validateURLType(repoParts[1]); err != nil {
```
Lou1415926: Do we need to validate for the existence as well, just like in `Validate()`?

internal/pkg/cli/pipeline_init.go
```diff
@@ -326,38 +392,62 @@ func (o *initPipelineOpts) parseBitbucketRepoDetails() error {
...
+	// Fetches and parses all branches associated with the chosen repo.
+	err := o.runner.Run("git", []string{"branch", "-a", "-l", o.repoShortName + "/*"}, exec.Stdout(&o.branchBuffer))
+	if err != nil {
+		return fmt.Errorf("get repo branch info: %w", err)
+	}
+	branches, err := o.parseGitBranchResults(strings.TrimSpace(o.branchBuffer.String()))
+	if err != nil {
+		return fmt.Errorf("parse git branch results: %w", err)
+	}
+	o.branchBuffer.Reset()
```
Lou1415926: Does this call for a `fetchAndParseBranches` 🤔 ?

internal/pkg/cli/pipeline_init.go
huanjani: We don't have pipelines for Windows, unfortunately.
But good question. A quick search tells me that git on Windows will still return slashes in the direction we're expecting here.

iamhopaul123: Wait do you mean we don't have it for Windows container or running Copilot on Windows but for linux?

iamhopaul123: I know pipeline doesn't support Windows environment but if they are just running Copilot on Windows machine that should be fine right?

internal/pkg/cli/pipeline_init.go
huanjani: Yes. They're the same thing. Git uses the term "shortname" for the remote name (https://git-scm.com/book/en/v2/Git-Basics-Working-with-Remotes) but to decrease confusion between repo name, repo URL, and repo shortname, I'll change "shortname" to "remoteName."

iamhopaul123: I didn't know that (til). Ok just feel free to do whatever you think is right!

internal/pkg/cli/pipeline_init.go
huanjani: I think when we have multiple pipeline support, users have to be much more intentional about their branches. Right now the default branch name just gets stuck in the pipeline manifest and the users may not even notice until their pipeline build fails (like in #2105, and my own experience). 

I like the idea of asking for the flag if it's not set! Is that breaking the paradigm, though, doing an `Ask` within `Validate`? I'm not seeing any other "must be specified with" situations besides in `task run` and that seems to be a different story.

internal/pkg/cli/pipeline_init.go
huanjani: I'm not sure I understand. The `-a` flag has `git branch` return both local and remote branches. 
We're listing all branches that have `[remoteName]/*`.

iamhopaul123: oh nice i didn't realize that. Sounds good!

internal/pkg/cli/pipeline_init.go
Lou1415926: I think we usually `Ask` for properties that are not specified in flags. That is in `Ask()`, we check if `o.branch == nil`, and if it is, we `askForBranch`. So by the end of `Ask`, all required properties should have been specified - either through flags, or prompts, or a combination of both.
`task run` doesn't have such situations though, it only has "must not be specified with" situations 🤔

internal/pkg/cli/pipeline_init.go
huanjani: Ah, `SelectOne` is similar to what I was looking for, but the problem is that we need both the url (for parsing info out of) and the remoteName/shortname (to pass to `git branch`). That's why I added the remoteName in the first place-- we did not used to have that at all. What I need is a parseValueAndHintFromOptionFmt, but that would just do what I'm doing here. 🤔

internal/pkg/cli/pipeline_init.go
huanjani: I was trying to avoid the word "cannot" but I agree that this is better.

internal/pkg/cli/pipeline_init.go
huanjani: true, true. I'll add to docs as well.

huanjani: The same could be said of this https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/cli/pipeline_init.go#L339

internal/pkg/cli/pipeline_init.go
huanjani: Right...that is my concern. If someone passes in a branch/URL but not the other, we're performing an `Ask()` type of behavior by prompting them for the other, rather than erroring out. Which may be fine....

For `task run` I added the requirement of `platform-os` and `platform-arch` both being required if one is.

internal/pkg/cli/pipeline_init.go
huanjani: Oh, you mean confirm that the URL is a git remote? The options that this URL was selected from were generated by the `git remote -v` command, so this URL has to exist. 
Does that answer your question?

Lou1415926: right right! Sorry it slipped my mind!

internal/pkg/cli/pipeline_init.go
huanjani: Changed code to ask (via selector) for missing branch/URL if the other is passed in.
Only the branches of the passed in URL are listed in the selector; only the URLs that contain the passed in branch are listed.
If there is only one option for either branch or URL, that one automatically is set, and a message is logged. (Same code for `Ask()`.

internal/pkg/cli/pipeline_init.go
huanjani: I'll add to the docs! Thanks.

internal/pkg/cli/pipeline_init.go
```diff
@@ -158,19 +157,57 @@ func (o *initPipelineOpts) Validate() error {
...
+			return err
+		}
+		return nil
+	}
+
+	// If only the '--git-branch' is passed in, ask for URL.
+	if o.repoBranch != "" {
+		// Ask user to select URL if it is not set by flag.
+		// Remote repo choices include only those that have the passed in branch.
+		if err := o.filterThenSelectReposByBranch(); err != nil {
```
iamhopaul123: I don't think we should let users to select anything during `Validate` stage, which I think should be primarily used for flag validation 🤔 

Also since we have https://github.com/aws/copilot-cli/pull/3182/files#diff-73c79a293149f8bc34d482c1ba4034d1dea1f7c37b74a4d6cd23131393887f5aR410 why do we still need them to select here?

internal/pkg/cli/pipeline_init.go
```diff
@@ -158,19 +157,57 @@ func (o *initPipelineOpts) Validate() error {
...
+		if err := o.validateURLType(o.repoURL); err != nil {
+			return err
+		}
+
+		var err error
+		o.repoRemoteName, err = o.getRemoteNameFromURL()
+		if err != nil {
+			return err
+		}
+		if err := o.selectBranch(); err != nil {
```
iamhopaul123: same here. we should handle it in `Ask()` i think

internal/pkg/cli/pipeline_init.go
```diff
@@ -225,6 +265,121 @@ func (o *initPipelineOpts) validateURL(url string) error {
...
+	err := o.runner.Run("git", []string{"branch", "-a", "-l", remoteName + "/*"}, exec.Stdout(&o.branchBuffer))
+	if err != nil {
+		return nil, fmt.Errorf("get Git repo branch info: %w", err)
+	}
+	if o.branchBuffer.String() == "" {
+		return []string{}, nil
+	}
+	branches, err := o.parseGitBranchResults(strings.TrimSpace(o.branchBuffer.String()))
+	if err != nil {
+		return nil, fmt.Errorf("parse 'git branch' results: %w", err)
```
iamhopaul123: I vaguely remember we usually use double quotes 💭

internal/pkg/cli/pipeline_init.go
```diff
@@ -225,6 +265,121 @@ func (o *initPipelineOpts) validateURL(url string) error {
...
+			}
+		}
+	}
+
+	if len(reposWithBranch) == 1 {
+		log.Infof("Only one git repository detected. Your pipeline will follow '%s'. You may make changes in the generated pipeline manifest before deployment.\n", color.HighlightUserInput(reposWithBranch[0]))
+		if err := o.parseAndAssignRepoParts(reposWithBranch[0]); err != nil {
+			return nil
+		}
+
```
iamhopaul123: Hmm I'd prefer it to be an isolated function so that we can explicitly know when we assign values to `o.repoURL` and `o.repoRemoteName`.

```suggestion
	 repoURL, repoRemoteName, err := o.parseAndAssignRepoParts(reposWithBranch[0])
	 if err != nil {
			return nil
	}
	o.repoURL, o.repoRemoteName = repoURL, repoRemoteName
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -326,38 +484,71 @@ func (o *initPipelineOpts) parseBitbucketRepoDetails() error {
...
 }
 
 func (o *initPipelineOpts) selectURL() error {
-	// Fetches and parses all remote repositories.
-	err := o.runner.Run("git", []string{"remote", "-v"}, exec.Stdout(&o.buffer))
-	if err != nil {
-		return fmt.Errorf("get remote repository info: %w; make sure you have installed Git and are in a Git repository", err)
-	}
-	urls, err := o.parseGitRemoteResult(strings.TrimSpace(o.buffer.String()))
+	URLsAndRemoteNames, err := o.fetchAndParseURLs()
```
iamhopaul123: can we combine `selectURL` and `selectBranch` so that we won't call this method too many times?

internal/pkg/cli/pipeline_init.go
```diff
@@ -225,6 +265,121 @@ func (o *initPipelineOpts) validateURL(url string) error {
...
+		branches, err := o.fetchAndParseBranches(remoteName)
+		if err != nil {
+			return err
+		}
+		if len(branches) == 0 {
+			continue
+		}
+		for _, branch := range branches {
+			if branch == o.repoBranch {
+				reposWithBranch = append(reposWithBranch, fmt.Sprintf("%s: %s", remoteName, url))
```
iamhopaul123: I still think we should do this https://github.com/aws/copilot-cli/pull/3182#discussion_r784266297 so that we don't need to handle error from `parseAndAssignRepoParts` because it can't necessarily happen...

internal/pkg/cli/pipeline_init.go
huanjani: Okay. I misunderstood your comment https://github.com/aws/copilot-cli/pull/3182#discussion_r784286346. Will revert.

huanjani: The reason I had originally made it so that users had to pass in both flags is that in order to validate branch, we need URL. In the past, we didn't validate that the passed-in URL or branch exists, but I think we all agreed it would be better to validate them before the pipeline building proceeded.

internal/pkg/cli/pipeline_init.go
```diff
@@ -225,6 +265,121 @@ func (o *initPipelineOpts) validateURL(url string) error {
...
 
+func (o *initPipelineOpts) validateURLExists(url string) error {
+	repos, err := o.fetchAndParseURLs()
+	if err != nil {
+		return err
+	}
+	if _, ok := repos[url]; !ok {
+		return fmt.Errorf("URL '%s' is not a local git remote", o.repoURL)
+	}
+	o.repoRemoteName = repos[url]
```
Lou1415926: This feels like a side-effect of `validateURLExists`. It changes something globally but we don't necessarily expect this effect from its name. Since this function is similar to `getRemoteNameFromURL`, what do you think of keeping just the `getRemoteNameFromURL` function, and then in `Validate()`, modify this block:
```go
if o.repoURL != "" {
   if err := validateURLType(); err != nil {
        return err
   }
   remoteName, err := getRemoteNameFromURL()
   if err != nil {
       return err
   }
   o.repoRemoteName = remoteName
}
```
This function is only called in the `if o.repoURL != "" && o.repoBranch != ""` block, and I think it can be refactored out given the current feedbacks 👍🏼

internal/pkg/cli/pipeline_init.go
```diff
@@ -225,6 +265,121 @@ func (o *initPipelineOpts) validateURL(url string) error {
...
+	repoParts := strings.Split(repo, ": ")
+	if len(repoParts) != 2 {
+		return fmt.Errorf("cannot parse repo '%s'", repo)
+	}
+	o.repoURL = repoParts[1]
+	o.repoRemoteName = repoParts[0]
+	return nil
+}
+
+func (o *initPipelineOpts) filterThenSelectReposByBranch() error {
```
Lou1415926: Will this function be refactored out completely if we don't ask for anything during `Validate`?

internal/pkg/cli/pipeline_init.go
huanjani: Changed again to delay the asking until `Ask()`

internal/pkg/cli/pipeline_init.go
huanjani: If the URL is passed in with a flag, it's called once to validate that the URL exists. 
If the URL is not passed in with a flag, it's called once to populate the selector.

If we combine the two select methods, we can't use them flexibly for when users pass in only one flag.

internal/pkg/cli/pipeline_init.go
huanjani: Yup

internal/pkg/cli/pipeline_init.go
huanjani: Sorry-- can you clarify what you mean that it can't necessarily happen?

iamhopaul123: Like `parseAndAssignRepoParts` can't error out since we always have the input constructed as `%s: %s`.

internal/pkg/cli/pipeline_init.go
iamhopaul123: I think it might be fine if they only provide one flag and we don't validate that far. So if they only pass the URL flag then we check if the repo exists, whereas if they only pass the branch flag we can't really validate it until in `Ask()` when they pass in the URL. Eventually we'll error out if there's no available option for them.

For example, if they pass in `master` for the branch flag and then in `Ask()` we get to know their URL. However, when finding available options for them the length is 0 then we'll error out (if it is 1 we'll make it default instead of asking for them to choose)

internal/pkg/cli/pipeline_init.go
huanjani: I used a struct to hold remoteName and repoURL instead of a map, with a fmt const for `%s: %s`.

internal/pkg/cli/pipeline_init.go
```diff
@@ -553,6 +477,11 @@ func (o *initPipelineOpts) selectBranch() error {
 	return nil
 }
 
+type repoResult struct {
+	remoteName string
+	remoteURL string
+}
+
 // examples:
 // efekarakus	git@github.com:efekarakus/grit.git (fetch)
```
iamhopaul123, line 486: lol

internal/pkg/cli/pipeline_init.go
```diff
@@ -566,16 +495,19 @@ func (o *initPipelineOpts) selectBranch() error {
...
+	var repos []repoResult
 	items := strings.Split(s, "\n")
 	for _, item := range items {
 		if !strings.Contains(item, githubURL) && !strings.Contains(item, ccIdentifier) && !strings.Contains(item, bbURL) {
 			continue
 		}
 		cols := strings.Split(item, "\t")
 		URL := strings.TrimSpace(strings.Split(cols[1], " ")[0])
-		repos[URL] = cols[0]
+		repos = append (repos, repoResult{
```
iamhopaul123: nit: append(

internal/pkg/cli/pipeline_init.go
```diff
@@ -566,16 +495,19 @@ func (o *initPipelineOpts) selectBranch() error {
 // bbssh	ssh://git@bitbucket.org:teamsinspace/documentation-tests.git (fetch)
 
 // parseGitRemoteResults returns just the first (remote name) and second (url) columns of the `git remote -v` results as a map (url: name), and skips urls from unsupported sources.
-func (o *initPipelineOpts) parseGitRemoteResult(s string) map[string]string {
-	repos := make(map[string]string)
+func (o *initPipelineOpts) parseGitRemoteResult(s string) []repoResult {
```
iamhopaul123, line 502: hmm why do we use `[]repoResult` instead of a map?

iamhopaul123 (approved):

internal/pkg/cli/pipeline_init.go
```diff
@@ -225,6 +240,57 @@ func (o *initPipelineOpts) validateURL(url string) error {
 	return nil
 }
 
+func (o *initPipelineOpts) validateBranch() error {
+	// URL has already been checked to exist and be valid; repoRemoteName already set by validateURLExists.
```
Lou1415926: nit:
```suggestion
	// URL has already been checked to exist and be valid; repoRemoteName already set 
 by `getRemoteNameFromURL`.
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -225,6 +240,57 @@ func (o *initPipelineOpts) validateURL(url string) error {
...
+	o.repoBuffer.Reset()
+
+	return repos, nil
+}
+
+func (o *initPipelineOpts) fetchAndParseBranches(remoteName string) ([]string, error) {
+	// Fetches and parses all branches associated with the chosen repo.
+	err := o.runner.Run("git", []string{"branch", "-a", "-l", remoteName + "/*"}, exec.Stdout(&o.branchBuffer))
+	if err != nil {
+		return nil, fmt.Errorf("get Git repo branch info: %w", err)
```
Lou1415926: nit:
```suggestion
		return nil, fmt.Errorf("get Git branch info for remote repo %s: %w", remoteName, err)
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -326,42 +402,86 @@ func (o *initPipelineOpts) parseBitbucketRepoDetails() error {
...
 }
 
 func (o *initPipelineOpts) selectURL() error {
-	// Fetches and parses all remote repositories.
-	err := o.runner.Run("git", []string{"remote", "-v"}, exec.Stdout(&o.buffer))
-	if err != nil {
-		return fmt.Errorf("get remote repository info: %w; make sure you have installed Git and are in a Git repository", err)
-	}
-	urls, err := o.parseGitRemoteResult(strings.TrimSpace(o.buffer.String()))
+	URLsAndRemoteNames, err := o.fetchAndParseURLs()
```
Lou1415926: nit:
```suggestion
	urlsAndRemoteNames, err := o.fetchAndParseURLs()
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -326,42 +402,86 @@ func (o *initPipelineOpts) parseBitbucketRepoDetails() error {
...
 		return err
 	}
-	o.buffer.Reset()
+
+	var formattedRemoteNamesAndURLs []string
+	for _, repo := range URLsAndRemoteNames {
+		formattedRemoteNamesAndURLs = append(formattedRemoteNamesAndURLs, fmt.Sprintf(fmtRepoSelection, repo.remoteName, repo.remoteURL))
+	}
+	if len(formattedRemoteNamesAndURLs) == 1 {
+		log.Infof("Only one git repository detected. Your pipeline will follow '%s'. You may make changes in the generated pipeline manifest before deployment.\n", color.HighlightUserInput(formattedRemoteNamesAndURLs[0]))
```
Lou1415926: nit: since the URL that the pipeline will follow is really `repo.remoteURL` (without the prepending colon and remote name)- what do you think of 
```suggestion
		log.Infof("Only one git repository detected. Your pipeline will follow '%s'. You may make changes in the generated pipeline manifest before deployment.\n", color.HighlightUserInput(repo[0].remoteURL))
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -326,42 +402,86 @@ func (o *initPipelineOpts) parseBitbucketRepoDetails() error {
...
 	if err != nil {
 		return err
 	}
-	o.buffer.Reset()
+
+	var formattedRemoteNamesAndURLs []string
+	for _, repo := range URLsAndRemoteNames {
+		formattedRemoteNamesAndURLs = append(formattedRemoteNamesAndURLs, fmt.Sprintf(fmtRepoSelection, repo.remoteName, repo.remoteURL))
+	}
+	if len(formattedRemoteNamesAndURLs) == 1 {
```
Lou1415926, line 418: There are a few repetitive lines - the concern is that if we need to make changes to the implementation, we need to remember to change it at two places. What do you think of
```go
         var selectedRepo string
	if len(formattedRemoteNamesAndURLs) == 1 {
                log.Infoof()
                selectedRepo = formattedRemoteNamesAndURLs[0]
         } else {
                var err error
                selectedRepo, err = o.prompt.SelectOne()
                if err != nil {}
         }
         remote, url := o.parse(selectedRepo)
         // ... the rest
```

I feel like in this scenario reducing duplicate codes is preferred over returning early 🤔  but I could be wrong

internal/pkg/cli/pipeline_init.go
```diff
@@ -374,24 +494,54 @@ func (o *initPipelineOpts) selectURL() error {
...
+	var repos []repoResult
 	items := strings.Split(s, "\n")
 	for _, item := range items {
 		if !strings.Contains(item, githubURL) && !strings.Contains(item, ccIdentifier) && !strings.Contains(item, bbURL) {
 			continue
 		}
 		cols := strings.Split(item, "\t")
-		url := strings.TrimSpace(strings.TrimSuffix(strings.Split(cols[1], " ")[0], ".git"))
-		urlSet[url] = true
+		URL := strings.TrimSpace(strings.Split(cols[1], " ")[0])
```
Lou1415926: nit:
```suggestion
		url := strings.TrimSpace(strings.Split(cols[1], " ")[0])
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -374,24 +494,54 @@ func (o *initPipelineOpts) selectURL() error {
 // bbhttps	https://huanjani@bitbucket.org/huanjani/aws-copilot-sample-service.git (fetch)
 // bbssh	ssh://git@bitbucket.org:teamsinspace/documentation-tests.git (fetch)
 
-// parseGitRemoteResults returns just the trimmed middle column (url) of the `git remote -v` results,
-// and skips urls from unsupported sources.
-func (o *initPipelineOpts) parseGitRemoteResult(s string) ([]string, error) {
-	var urls []string
-	urlSet := make(map[string]bool)
+// parseGitRemoteResults returns just the first (remote name) and second (url) columns of the `git remote -v` results as a map (url: name), and skips urls from unsupported sources.
+func (o *initPipelineOpts) parseGitRemoteResult(s string) []repoResult {
```
Lou1415926, line 502: We could include an example input (just like the comment for `parseGitBranchResults` so that it's easy to see how the parsing is done 💭

internal/pkg/cli/pipeline_init.go
```diff
@@ -374,24 +494,54 @@ func (o *initPipelineOpts) selectURL() error {
...
 	}
-	for url := range urlSet {
-		urls = append(urls, url)
+	for _, repo := range repos {
+		if repo.remoteURL == o.repoURL {
+			return repo.remoteName, nil
+		}
 	}
-	return urls, nil
+	return "", fmt.Errorf("URL '%s' is not a local git remote", o.repoURL)
```
Lou1415926: I wonder if it's syntactically correct to say `fmt.Errorf("url '%s' is not a local git remote", o.repoURL)` instead?

Lou1415926 (approved):
LGTM! Just some nits, nothing major. Feel free to remove the label once addressed!

internal/pkg/cli/pipeline_init.go
huanjani: I changed it to a struct bc it felt more in line with the rest of our code, and easier to format then parse.

internal/pkg/cli/pipeline_init.go
huanjani: good catch!

internal/pkg/cli/pipeline_init.go
huanjani: great idea!

internal/pkg/cli/pipeline_init.go
huanjani: Yeah, good question. I feel like it may be a little easier to read as is, but I can see your point. I'll leave it as is for now, because we'll be making more changes to this file soon enough and we can reconsider then!

PR #3183, opened 2022-01-11
chore: push template to s3 and deploy workload stack

Lou1415926:
This PR is to address [this comment](https://github.com/aws/copilot-cli/pull/3171#discussion_r781820226) in a previous PR. When deploying workloads, instead of using the template body string, we first upload the template to the s3 bucket, and then refer to the URL in the API call. This is similar to the current approach to deploy a pipeline.

We need to do this because there is a 51KB limit of template body string, but 1MB if it's uploaded to s3 bucket. As the template is growing larger and larger, 51KB may become insufficient.

### DIFF
--- a/internal/pkg/cli/env_init.go
+++ b/internal/pkg/cli/env_init.go
@@ -292,7 +292,7 @@ func (o *initEnvOpts) Execute() error {
 
 	// 4. Start creating the CloudFormation stack for the environment.
 	if resources.S3Bucket == "" {
-		log.Errorf("Cannot find the S3 artifact bucket in %s region created by app. The S3 bucket is necessary for many future operations. For example, when you need addons to your services.", envRegion)
+		log.Errorf("Cannot find the S3 artifact bucket in %s region created by app %s. The S3 bucket is necessary for many future operations. For example, when you need addons to your services.", envRegion, app.Name)
 		return fmt.Errorf("cannot find the S3 artifact bucket in %s region", envRegion)
 	}
 	partition, err := partitions.Region(envRegion).Partition()
--- a/internal/pkg/cli/env_init.go
+++ b/internal/pkg/cli/env_init.go
@@ -10,14 +10,13 @@ import (
 	"os"
 	"strings"
 
-	"github.com/aws/aws-sdk-go/aws/endpoints"
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/copilot-cli/internal/pkg/aws/cloudformation"
 	"github.com/aws/copilot-cli/internal/pkg/aws/ec2"
 	"github.com/aws/copilot-cli/internal/pkg/aws/iam"
 	"github.com/aws/copilot-cli/internal/pkg/aws/identity"
+	"github.com/aws/copilot-cli/internal/pkg/aws/partitions"
 	"github.com/aws/copilot-cli/internal/pkg/aws/profile"
 	"github.com/aws/copilot-cli/internal/pkg/aws/s3"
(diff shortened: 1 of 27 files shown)


internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -106,12 +106,21 @@ type CloudFormation struct {
...
 	appStackSet    stackSetClient
-	s3Client       s3Client
+	appS3Client    s3Client
 	region         string
 }
 
+// CloudFormationOption represents an option to apply to a CloudFormation.
+type CloudFormationOption func(cf *CloudFormation)
+
+func WithAppS3Client(sess *session.Session) func(cf *CloudFormation) {
```
iamhopaul123: comment pls 🙏

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -123,9 +132,12 @@ func New(sess *session.Session) CloudFormation {
 			}))
 		},
 		appStackSet: stackset.New(sess),
-		s3Client:    s3.New(sess),
+		appS3Client: s3.New(sess),
```
iamhopaul123: Is there any reason why previously we can use the env manager role but now we have to use the profile role 🤔

iamhopaul123 (approved):

internal/pkg/deploy/cloudformation/cloudformation.go
Lou1415926: We didn't use `s3Client` when we deploy a workload, so the permission issue didn't surface. This field was previously used only by pipeline.

internal/pkg/deploy/cloudformation/cloudformation.go
Lou1415926: added!

internal/pkg/deploy/cloudformation/cloudformation.go
efekarakus: Check out the two commits that I pushed on top of your changes: https://github.com/efekarakus/copilot-cli/commits/upload-template-to-s3

I know that today for addons, we use the `defaultSessWithEnvRegion`, but I think that violates the idea that `copilot deploy` only needs to assume the `EnvManagerRole` to deploy a workload.

Instead, 
1. https://github.com/efekarakus/copilot-cli/commit/d22247fa1612d58adbe38504d1c782ea8c4ccfdc is a prototype that shows how we can augment the `EnvManagerRole` with permission to push to the artifact bucket
2. https://github.com/efekarakus/copilot-cli/commit/f8f10894bf60064221e31976ee7993424c25d911 easy peasy, swap the session

I tested the above and it works.

I think `env init` can get away with using `defaultSessWithEnvRegion` because you need to be an admin to create an environment, however to deploy a service users should only need ideally the EnvManagerRole.

Hope this makes sense!

internal/pkg/cli/env_init.go
```diff
@@ -290,7 +291,15 @@ func (o *initEnvOpts) Execute() error {
 	}
 
 	// 4. Start creating the CloudFormation stack for the environment.
-	if err := o.deployEnv(app, urls); err != nil {
+	if resources.S3Bucket == "" {
+		log.Errorf("Cannot find the S3 artifact bucket in %s region created by app. The S3 bucket is necessary for many future operations. For example, when you need addons to your services.", envRegion)
```
huanjani: nit: not sure if this was intended:
```suggestion
		log.Errorf("Cannot find the S3 artifact bucket in %s region created by app %s. The S3 bucket is necessary for many future operations. For example, when you need addons to your services.", envRegion, app)
```

huanjani (approved):
:shipit:

PR #3189, opened 2022-01-14
chore: UI update for NLB support

Lou1415926:
This PR:
1. Gets NLB URL in describer. This affects the URL that Copilot shows at the end of service deployment, as well as `copilot svc show`.
2. Validate during manifest validation that the protocol is valid

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -54,7 +54,7 @@ func convertSidecar(s map[string]*manifest.SidecarConfig) ([]*template.SidecarOp
 	}
 	var sidecars []*template.SidecarOpts
 	for name, config := range s {
-		port, protocol, err := parsePortMapping(config.Port)
+		port, protocol, err := manifest.ParsePortMapping(config.Port)
 		if err != nil {
 			return nil, err
 		}
@@ -114,22 +114,6 @@ func convertDependsOn(d manifest.DependsOn) map[string]string {
 		dependsOn[name] = strings.ToUpper(status)
 	}
 	return dependsOn
-}
-
-// Valid sidecar portMapping example: 2000/udp, or 2000 (default to be tcp).
-func parsePortMapping(s *string) (port *string, protocol *string, err error) {
-	if s == nil {
-		return nil, nil, nil
-	}
-	portProtocol := strings.Split(*s, "/")
-	switch len(portProtocol) {
-	case 1:
-		return aws.String(portProtocol[0]), nil, nil
-	case 2:
-		return aws.String(portProtocol[0]), aws.String(portProtocol[1]), nil
-	default:
-		return nil, nil, fmt.Errorf("cannot parse port mapping from %s", *s)
-	}
 }
 
 func convertAdvancedCount(a manifest.AdvancedCount) (*template.AdvancedCount, error) {
@@ -276,7 +260,7 @@ func (s *LoadBalancedWebService) convertNetworkLoadBalancer() (networkLoadBalanc
 	}
 
 	// Parse listener port and protocol.
-	port, protocol, err := parsePortMapping(nlbConfig.Port)
+	port, protocol, err := manifest.ParsePortMapping(nlbConfig.Port)
 	if err != nil {
 		return networkLoadBalancerConfig{}, err
 	}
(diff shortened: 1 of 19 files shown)


Lou1415926:
Screen Shot 2022-01-14 at 4 52 05 PM
Screen Shot 2022-01-14 at 4 52 23 PM

internal/pkg/describe/backend_service_test.go
```diff
@@ -529,6 +529,8 @@ Resources
 				environments:     []string{"test", "prod"},
 			}
 			human := backendSvc.HumanString()
+			fmt.Println(human)
+			fmt.Println(tc.wantedHumanString)
```
iamhopaul123:
```suggestion
```

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -316,3 +317,10 @@ func (e *Alias) ToStringSlice() ([]string, error) {
 	}
 	return out, nil
 }
+
+func (e *Alias) ToString() string {
```
iamhopaul123, line 322: Coment pls

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -116,22 +116,6 @@ func convertDependsOn(d manifest.DependsOn) map[string]string {
 	return dependsOn
 }
 
-// Valid sidecar portMapping example: 2000/udp, or 2000 (default to be tcp).
-func parsePortMapping(s *string) (port *string, protocol *string, err error) {
```
iamhopaul123, line 120: Nice!

internal/pkg/manifest/validate.go
```diff
@@ -616,6 +619,21 @@ func (c NetworkLoadBalancerConfiguration) Validate() error {
...
+func validateNLBPort(port *string) error {
+	_, protocol, err := ParsePortMapping(port)
+	if err != nil {
+		return err
+	}
+	if protocol == nil {
+		return nil
+	}
+	protocolVal := aws.StringValue(protocol)
+	if strings.ToLower(protocolVal) != "tcp" && strings.ToLower(protocolVal) != "udp" && strings.ToLower(protocolVal) != "tls" {
```
iamhopaul123: Could we make the available options for the port protocol to be enums? Also is `tcp_udp` a valid option?

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
```diff
@@ -36,6 +36,13 @@ Parameters:
     Type: String
   TargetPort:
     Type: Number
+{{- if .NLB }}
+  NLBAliases:
+    Type: String
+    Default: ""
+  NLBPort:
```
iamhopaul123, line 43: Why do we only expose aliases and port? How do we decide which fields of `template.NetworkLoadBalancerListener` we use parameter as opposed to go template?

internal/pkg/template/workload.go
```diff
@@ -208,7 +208,6 @@ type HTTPHealthCheckOpts struct {
 
 // NetworkLoadBalancerListener holds configuration that's need for a Network Load Balancer listener.
 type NetworkLoadBalancerListener struct {
-	Port            string
```
iamhopaul123, line 211: Should we remove the `Aliases` field as well?

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
Lou1415926: The initial motivation was simply to support grabbing NLB URL for the `URI()` function in `describe` package. To determine the url of the service, we need the port number for sure, and one of the aliases and the nlb's dns name. This is basically how the decision was made.

internal/pkg/template/workload.go
Lou1415926: Currently there is a go template version of `Aliases` used to determine whether the custom resource for custom domain should be rendered or not. 

It could use a `Condition` instead of go template in order to trigger the custom resource at appropriate time, but it adds a potentially confusing and weird-looking `Condition` block (i.e. something like  `UseAlias: !and [!eq [!Ref Aliases, ""], !Condition HasAssociatedDomain]`) so I refrained from doing so. 

What do you think?

huanjani (approved):
Lots of good stuff in here!

iamhopaul123 (approved):

Issue #3170, opened 2022-01-06
Support Podman either as alias to Docker or natively


DerGut:
It would be great if Podman could be used in addition to Docker for building images.

I had to install Docker to be able to use Copilot. But I don't think this has to be necessary as most of the Docker commands can be replaced by running `podman` in place of `docker` (as far as I know).

I realise, that it might be easier to accept a docker alias from the shell environment than it is to implement a new client like in [dockerengine.go](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/docker/dockerengine/dockerengine.go).
Under `e2e` there already seems to be an implementation that makes use of the bash shell context by running `bash -c docker ...`: https://github.com/aws/copilot-cli/blob/f1b8e46522fd46755152bba4ce1ecb70c11e11bc/e2e/internal/client/docker.go#L54-L56

Would this be a possible future addition or is there any technical limitation preventing Podman from being used here?

efekarakus:
Hi @DerGut !

> Would this be a possible future addition or is there any technical limitation preventing Podman from being used here?

It's definitely possible :) there are no blockers for Copilot to use Podman. I've added the feature request label to the issue, thank you!

matthewhembree:
I wonder if Podman is the best target or if something like Colima or Rancher Desktop would be a better direction.

My team also uses AWS SAM and there are [efforts](https://github.com/aws/aws-sam-cli/issues/3595) to enable Lima based container engines (Colima / Rancher Desktop). It seems unusual to use two AWS toolchains and need to have two separate container engines for each toolchain.

My organization looked at Podman and the lack of bind mount functionality on Mac and Windows is a blocker for us. That definitely blocks SAM local invocation as the SAM container bind mounts the source code. We don't predict that we'll be using volumes/EFS in Copilot/ECS; but for organizations that would, they would be blocked.

sombriks:
yes offer podman as option would be great, tried the command alias but didn't work.

nathanpeck:
Hey all, I have done a deep investigation into Podman + AWS Copilot. I was able to get Copilot to use Podman by adding a symlink: `ln -s /usr/local/bin/podman /usr/local/bin/docker`

Unfortunately there is a difference between how Podman does image digests and how Docker Desktop does image digests, which blocks the use of Podman at this time. I've created an issue on the Podman repo: https://github.com/containers/podman/issues/14779

[...]

PR #3201, opened 2022-01-21
feat: Support Podman to build container images

hown3d:
I added an experimental Environment Variable __COPILOT_PODMAN__, that lets users use Podman instead of docker to build their images.
An e2e Test using Podman was also added. It's run after the docker e2e test.

Addresses #3170


hown3d:
I merged upstream into the branch again. Is there something else to be done for review? @Lou1415926 @iamhopaul123

efekarakus:
Hi @hown3d ! 
I'm so sorry that it took this long to reply to you 🙇 thank you very much for the contribution. I've some high level questions before integrating with podman if you don't mind.

1. Do you know of any other tools that integrate with podman? Is there a reason why you preferred environment variable as a way of toggling?   
I'm looking to see if there is a precedent for using env variables to build with podman vs. an option in the manifest such as:
   ```yaml
   image:
     build:
        engine: 'podman' # default is 'docker'
   ```

2. What are the downsides today of just doing `alias docker=podman` like in the [homepage of podman](https://podman.io/)? I'm trying to understand if it's worthwhile adding and maintaining new code in the repo if the mitigation already works for users.

3. How about users that use the [copilot pipelines](https://aws.github.io/copilot-cli/docs/concepts/pipelines/)? `podman` won't be installed in the build project, do we deem it acceptable if the images are build with docker in the pipeline but with podman locally?

If possible a design github issue or commenting on https://github.com/aws/copilot-cli/issues/3170 with the various alternatives would make it easier for us to accept the PR:
> * If you would like to propose a new feature, please open an issue on GitHub with a detailed description. This enables us to collaborate on the feature design more easily and increases the chances that your feature request will be accepted.
> https://github.com/aws/copilot-cli/blob/mainline/CONTRIBUTING.md


Again I'm really sorry about the delay on responding to the PR.

nathanpeck:
It is possible to use Copilot with Podman instead of Docker by running the following on Mac:

```
ln -s /usr/local/bin/podman /usr/local/bin/docker
```

This puts a symlink in place of the Docker binary, which redirects the commands to Podman

nathanpeck:
Unfortunately after some deeper experimentation I found another blocker: Podman and Docker Desktop have differing image digest behavior which is currently breaking the AWS Copilot flow and causing the image references to be broken.

Will continue to look for workarounds to this issue, but in the meantime I have opened an issue on Podman itself: https://github.com/containers/podman/issues/14779

efekarakus:
Hi @hown3d, I'm closing the issue for now as Nathan pointed out we can reroute `docker` to `podman` but even then `digests` don't seem to match. I think if the `--tag` field is used `copilot svc deploy`, copilot won't use the digest sha so there is path to mitigation although not ideal.

PR #3212, opened 2022-01-24
chore: fix integration test

Lou1415926:
The failure was caused by #3183.

We add to the environment manager role additional permissions to put object into the S3 bucket. In normal workflow, the `Artifact...` data is passed from `cli` (e.g. `env init`) to the template. This is missing from the integ test.

An additional validation for `http.path` tags along. If `path` is missing, the deployment will fail anyway in a later stage. manifest validation could provide a more friendyl error mesage than the cfn error.

### DIFF
--- a/internal/pkg/manifest/validate.go
+++ b/internal/pkg/manifest/validate.go
@@ -550,7 +550,7 @@ func (CommandOverride) Validate() error {
 }
 
 func (r RoutingRuleConfigOrBool) Validate() error {
-	if aws.BoolValue(r.Enabled) == true {
+	if aws.BoolValue(r.Enabled) {
 		return &errFieldMustBeSpecified{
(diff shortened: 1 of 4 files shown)


internal/pkg/manifest/validate.go
```diff
@@ -549,6 +549,15 @@ func (CommandOverride) Validate() error {
 	return nil
 }
 
+func (r RoutingRuleConfigOrBool) Validate() error {
```
iamhopaul123, line 553: Comment pls

iamhopaul123 (approved):

internal/pkg/manifest/validate.go
Lou1415926: done!

huanjani (approved):

PR #3217, opened 2022-01-25
chore(cli): add cli/deploy pkg

iamhopaul123:


internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,875 @@
...
+// WorkloadDeployer holds all the metadata and clients needed to deploy a workload.
+type WorkloadDeployer struct {
+	Name     string
+	App      *config.Application
+	Env      *config.Environment
+	ImageTag string
+	S3Bucket string
+
+	// cached variables
+	mft           interface{}
```
efekarakus: Instead of having the empty interface and accepting any manifest types I wonder if we should have specific structs instead?

```go
type RDWSDeployer struct { ...}
type LBWSDeployer struct { ... }
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,875 @@
...
+type pathFinder interface {
+	Path() (string, error)
+}
+
+// WorkloadDeployer holds all the metadata and clients needed to deploy a workload.
+type WorkloadDeployer struct {
+	Name     string
+	App      *config.Application
+	Env      *config.Environment
+	ImageTag string
```
efekarakus, line 155: should the image tag be part of the `DeployWorkloadInput` instead?

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,875 @@
...
+	S3Bucket string
+
+	// cached variables
+	mft           interface{}
+	workspacePath string
+}
+
+// UploadArtifactsInputs is the input of UploadArtifacts.
+type UploadArtifactsInputs struct {
```
efekarakus: To be consistent with the aws sdk go, can we use singular? (example: https://docs.aws.amazon.com/sdk-for-go/api/service/ecs/#CreateCapacityProviderInput)

```suggestion
type UploadArtifactsInput struct {
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,875 @@
...
+	ShouldUploadArtifacts bool
+
+	Fs                 *afero.Afero
```
efekarakus: nit:
```suggestion
	FS                 *afero.Afero
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,875 @@
...
+	Uploader           Uploader
+	Templater          Templater
+	Ws                 WorkspaceReader
```
efekarakus:
```suggestion
	WW                 WorkspaceReader
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,875 @@
...
+	NewInterpolator    func(app, env string) Interpolator
+	Unmarshal          func([]byte) (manifest.WorkloadManifest, error)
+	ImageBuilderPusher ImageBuilderPusher
+}
+
+// UploadArtifactsOutputs is the output of UploadArtifacts.
+type UploadArtifactsOutputs struct {
+	BuildRequired bool
```
efekarakus: What does this mean? I guess my confusion is around why have both `ImageDigest` and `BuildRequired`

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,875 @@
...
+	ImageDigest   string
+	EnvFileARN    string
+	AddonsURL     string
+}
+
+// DeployWorkloadInputs is the input of DeployWorkload.
+type DeployWorkloadInputs struct {
```
efekarakus: Do all workloads need this much info? (It's okay if it does!)

Or is there an opportunity to do something like:
```go
type DeployRDWSInput struct { ...}
func (in *DeployRDWSInput) Deploy(deployer *WorkloadDeployer) error

type DeployLBWSInput struct { ...}
func (in *DeployLBWSInput) Deploy(deployer *WorkloadDeployer) error
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,875 @@
...
+		return nil, err
+	}
+	cmdRunAt := in.now()
+	if err := in.ServiceDeployer.DeployService(os.Stderr, stackConfigoutputs.conf, d.S3Bucket, awscloudformation.WithRoleARN(d.Env.ExecutionRoleARN)); err != nil {
+		var errEmptyCS *awscloudformation.ErrChangeSetEmpty
+		if !errors.As(err, &errEmptyCS) {
+			return nil, fmt.Errorf("deploy service: %w", err)
+		}
+		if !in.ForceNewUpdate {
+			log.Warningln("Set --force to force an update for the service.")
```
efekarakus, line 273: should we have a `logger` interface or something part of `WorkloadDeployer`?

efekarakus (commented):
I haven't reviewed completely, just leaving some initial feedback to hear your thoughts

internal/pkg/cli/deploy/deploy.go
iamhopaul123: `mft` is just a private cache variable not an input and the deployer reads from the workspace. If we have different deployer type then we'll have to handle the manifest reading in each of our subcommand. Could be something duplicated right 🤔

internal/pkg/cli/deploy/deploy.go
iamhopaul123: It is also used when building and uploading the image.

internal/pkg/cli/deploy/deploy.go
iamhopaul123: good idea!

internal/pkg/cli/deploy/deploy.go
iamhopaul123: but afero also uses `Fs` 🤔

internal/pkg/cli/deploy/deploy.go
iamhopaul123: Do you mean `WS`?

internal/pkg/cli/deploy/deploy.go
efekarakus: it's okay, just a nit but i bet afero probably regrets it, see stdlib: https://pkg.go.dev/io/fs#FS

iamhopaul123: ok gotcha i'll make the change

internal/pkg/cli/deploy/deploy.go
iamhopaul123: because of https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/cli/svc_deploy.go#L409-L410

but maybe we don't need the buildrequired at all as an output

internal/pkg/cli/deploy/deploy.go
iamhopaul123: I thought about it but it might be ok since it is within the `cli` package so that we don't output logs everywhere. Why do you want to add a `logger` interface? Like what functionality you think we should have here?

internal/pkg/manifest/svc.go
```diff
@@ -263,11 +263,7 @@ func (qs *QueueScaling) AcceptableBacklogPerTask() (int, error) {
 	return int(v), nil
 }
 
-// ServiceDockerfileBuildRequired returns if the service container image should be built from local Dockerfile.
-func ServiceDockerfileBuildRequired(svc interface{}) (bool, error) {
-	return dockerfileBuildRequired("service", svc)
-}
-
+// IsTypeAService returns is manifest type is service.
```
huanjani: tiny nit: 
```suggestion
// IsTypeAService returns if manifest type is service.
```

huanjani:
This is exciting! I remember you mentioning your plan to do this, but would you mind reminding me what the purpose/goal is? Is it to separate out the cli level-specific parts of the existing deploy package? Is this related to the `package` command work?

iamhopaul123:
> This is exciting! I remember you mentioning your plan to do this, but would you mind reminding me what the purpose/goal is? Is it to separate out the cli level-specific parts of the existing deploy package? Is this related to the `package` command work?

Yes. But this change only includes two public methods for the `workload deploy`. I'll add another one for `package` (see design doc for the new implementation plan). However, some private methods and functions will be shared for the last one.

huanjani (approved):
Nice work! Thanks for doing this.

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,873 @@
...
+	S3Bucket string
+
+	// cached variables
+	mft           interface{}
+	workspacePath string
+}
+
+// UploadArtifactsInput is the input of UploadArtifacts.
+type UploadArtifactsInput struct {
+	ShouldUploadArtifacts bool
```
efekarakus: This field is a bit odd, can we remove it? Why call the method if this field is false, the client already has that information so that should be able to do on their own:
```go
if shouldUploadArtifacts {
    d.UploadArtifacts(...)
}
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,873 @@
...
+
+	FS                 *afero.Afero
+	Uploader           Uploader
+	Templater          Templater
+	WS                 WorkspaceReader
+	NewInterpolator    func(app, env string) Interpolator
+	Unmarshal          func([]byte) (manifest.WorkloadManifest, error)
```
efekarakus: Do these need to be provided by the client?  They're fields used for testing so pushing the clients to provide the values for these are a bit awkward.

Can we instead have them be private fields in `DeployWorkload` that get set in the constructor, or alternatively:
```suggestion
	ManifestInterpolator     Interpolator
	Manifest                  manifest.WorkloadManifest
```
If you'd like to keep them as inputs to `UploadArtifact`

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,873 @@
...
+}
+
+// DeployWorkloadInput is the input of DeployWorkload.
+type DeployWorkloadInput struct {
+	ImageDigest    string
+	EnvFileARN     string
+	AddonsURL      string
+	RootUserARN    string
+	AppTags        map[string]string
+	ResourceTags   map[string]string
```
efekarakus, line 213: Dumb question: what was the difference between these two? 🤔

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,873 @@
...
+	// cached variables
+	mft           interface{}
+	workspacePath string
+}
+
+// UploadArtifactsInput is the input of UploadArtifacts.
+type UploadArtifactsInput struct {
+	ShouldUploadArtifacts bool
+
+	FS                 *afero.Afero
```
efekarakus: Should we accepting an interface instead for ease of testing? Maybe:
```go
type FileReader interface {
   ReadFile(string) ([]byte, error)
}
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,873 @@
...
+	AddonsURL      string
+	RootUserARN    string
+	AppTags        map[string]string
+	ResourceTags   map[string]string
+	ImageRepoURL   string
+	ForceNewUpdate bool
+
+	WS                     WorkspaceReader
+	NewInterpolator        func(app, env string) Interpolator
+	Unmarshal              func([]byte) (manifest.WorkloadManifest, error)
```
efekarakus: Same feedback as above

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,873 @@
...
+	if ok {
+		return mf.EnvFile()
+	}
+	// If the manifest type doesn't support envFiles, ignore and move forward.
+	return ""
+}
+
+type UploadCustomResourcesOpts struct {
+	uploader      CustomResourcesUploader
+	newS3Uploader func() (Uploader, error)
```
efekarakus: If these fields are private how will the client set them 🤔 ?  
My preference is for the `cli/deploy` package not to have any `newXXXX` fields to create clients. Instead, we should have dependency injection from the `cli` package.
1. The `cli` package is responsible for creating these objects
2. The `cli/deploy` package accepts only interfaces or data types:   
```suggestion
	CustomResources      CustomResourcesUploader
	S3                   Uploader
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,873 @@
...
+	}
+	mf, ok := unmarshaledManifest.(envFile)
+	if ok {
+		return mf.EnvFile()
+	}
+	// If the manifest type doesn't support envFiles, ignore and move forward.
+	return ""
+}
+
+type UploadCustomResourcesOpts struct {
```
efekarakus: nit: doc comments 🥺

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,873 @@
...
+	RootUserARN    string
+	AppTags        map[string]string
+	ResourceTags   map[string]string
+	ImageRepoURL   string
+	ForceNewUpdate bool
+
+	WS                     WorkspaceReader
+	NewInterpolator        func(app, env string) Interpolator
+	Unmarshal              func([]byte) (manifest.WorkloadManifest, error)
+	UploadOpts             *UploadCustomResourcesOpts
```
efekarakus: nit: should we move this below `ForceNewUpdate` since it's a data type and not an interface

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,873 @@
...
+	ServiceForceUpdater    ServiceForceUpdater
+	EndpointGetter         EndpointGetter
+	Spinner                Progress
+
+	now func() time.Time
+}
+
+// DeployWorkloadOutput is the output of DeployWorkload.
+type DeployWorkloadOutput struct {
+	RDWSAlias     string
```
efekarakus, line 233: Why do we have a specific field for `RDWS` for an output that should work for any workload types 🤔 ?   
I feel like this is an opportunity to use polymorphism (kinda like: https://refactoring.com/catalog/replaceConditionalWithPolymorphism.html)

```go
type workloadDeployOutput struct{ ... }
func (out workloadDeployOutput) Subscriptions() []manifest.TopicSubscription

type rdwsDeployOutput struct{ ... }
func (out rdwsDeployOutput) Subscriptions() []manifest.TopicSubscription
func (out rdwsDeployOutput) Endpoint() string

type lbwsDeployOutput struct{ ... }
func (out lbwsDeployOutput) Subscriptions() []manifest.TopicSubscription
func (out lbwsDeployOutput) Endpoint() string
```

This way all the output types implement the same interface `type output interface { Subscriptions() []manifest.TopicSubscription }` and some have even more methods:
```go
type addressableOutput interface {
   output
   Endpoint() string
}
if aout, ok := out.(addressableOutput) {
   log(aout.Endpoint())
}
log(out.Subscriptions())
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,873 @@
...
+	Subscriptions []manifest.TopicSubscription
+}
+
+// UploadArtifacts uploads the deployment artifacts (image, addons files, env files).
+func (d *WorkloadDeployer) UploadArtifacts(in *UploadArtifactsInput) (*UploadArtifactsOutput, error) {
```
efekarakus: Should custom resources be included in this method?

internal/pkg/cli/deploy/deploy.go
efekarakus: Sounds good! I didn't actually realize that this pkg was under the `cli` directory 😅

efekarakus (commented):
Sorry for the late review boss! Thanks for moving all this logic to its own package 👍 It's a great first step that leaves the codebase better than before. 

I think what I care the most about is: Having the **exported** names (types, methods) be the best they can be (dependency injection, good type names).   
The rest, body of the methods and private stuff, can remain as is.

After the next revision, I can approve!

internal/pkg/cli/deploy/deploy.go
iamhopaul123: agreed! i'll keep them as private vars

internal/pkg/cli/deploy/deploy.go
iamhopaul123: `AppTags` are the app level resource tags while `resourceTags` are tags assigned during `workload deploy`. We'll merge these two groups.

internal/pkg/cli/deploy/deploy.go
iamhopaul123: i'll put it into the constructor and make it to be a private field

internal/pkg/cli/deploy/deploy.go
iamhopaul123: Hmm I did this because if you look at aws sdk, the output of their APIs leave some fields empty if they are not applicable. For example, if the service is Fargate then some EC2 fields are not set in the output. Also, I thought it would be easier to use as well. What advantage do you think over the current implementation 🤔

internal/pkg/cli/deploy/deploy.go
iamhopaul123: Good callout but not for now. I'll add it once I add the feature allowing `workload package` to be able to generate CFN template/params that are ready to deploy. Before doing that, I'll do the refactoring keeping the current behavior.

internal/pkg/cli/deploy/deploy.go
iamhopaul123: But the problem is `manifest.UnmarshalWorkload` doesn't have any receiver 🤔

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,850 @@
...
+	app           *config.Application
+	env           *config.Environment
+	imageTag      string
+	s3Bucket      string
+	mft           interface{}
+	workspacePath string
+}
+
+// NewWorkloadDeployerInput is the input to for workloadDeployer constructor.
+type NewWorkloadDeployerInput struct {
```
efekarakus: nit: 
```suggestion
type WorkloadDeployerInput struct {
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,850 @@
...
+	ResourceTags   map[string]string
+	ImageRepoURL   string
+	ForceNewUpdate bool
+
+	CustomResourceUploader CustomResourcesUploader
+	S3Uploader             Uploader
+	SNSTopicsLister        SNSTopicsLister
+	ServiceDeployer        ServiceDeployer
+	NewSvcUpdater          func(func(*session.Session) ServiceForceUpdater)
+	NewAppVersionGetter    func(string) (VersionGetter, error)
```
efekarakus: Is there a reason why we cant do this instead:
```suggestion
	ForceUpdater          ServiceForceUpdater
	AppVersionGetter      VersionGetter
```

internal/pkg/cli/deploy/deploy.go
efekarakus: I'm not strongly opposed to it given that it's just one field at the moment, so if you want to feel free to proceed.  

But I don't agree that it's easier to use 🤔  How will this type progress in the future? My guess is like this:
```go
type DeployWorkloadOutput struct {
	RDWSAlias     string
	LBWSAlias     string
	BackendAlias string
	Subscriptions []manifest.TopicSubscription
}
```
Now each one of these fields need to be individually checked for their zero-value by the client, the `DeployWorkloadOutput` is a mixed bag of mutually exclusive fields. Instead compare it to using duck-typing: https://gist.github.com/efekarakus/6dd25c0dfc11badf5d506dd177bf881d

efekarakus (approved):
Looks good, approving and adding the label to unblock you!

Two comments below, let me know what you think 🧠

internal/pkg/cli/deploy/deploy.go
iamhopaul123: Ohh gotcha. I think it's because i haven't put it into the real production yet. I'll make this change when integrating this new package and removing the old codes.

internal/pkg/cli/deploy/deploy.go
iamhopaul123: For `NewSvcUpdater` because we won't know what updater we should initialize.

internal/pkg/cli/deploy/deploy.go
iamhopaul123: Seems like it is not easy to do what you proposed above because the order of recommendation becomes a problem :( let me know your thought on [this PR](https://github.com/aws/copilot-cli/pull/3239)!

internal/pkg/cli/deploy/deploy.go
```diff
@@ -0,0 +1,838 @@
...
+	S3Bucket string
+}
+
+// NewWorkloadDeployer is the constructor for workloadDeployer.
+func NewWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
+	ws, err := workspace.New()
+	if err != nil {
+		return nil, fmt.Errorf("new workspace: %w", err)
+	}
+	mft, err := workloadManifest(&workloadManifestInput{
```
Lou1415926, line 165: Sorry for being so late to the PR!!! 🙇🏼‍♀️ Just a concern to throw out here: I wonder if it should be considered a side effect that we are actually reading/applying env override/validating the manifest here, in a constructor 🤔 I feel like constructor, ideally should only init a few members, without doing the extra work mentioned just now.

iamhopaul123: The reason why I put the manifest reading in the constructor is because it is used in both `UploadArtifacts` and `DeployWorkload` (`GenerateCloudFormationTemplate` for `workload package` in future PR as well). If we don't handle the logic in the constructor, then we have to either 1) define it as a function within the `cli` package and call it for `workload deploy/package` or 2) call `workloadManifest` within each of the `workloadDeployer` method. What do you think?

Lou1415926: Ya I like the first option! I feel like it makes sense to not delegate the manifest responsibility to `deploy` package, since manifest operations are not unique to deployment - it is also needed for packaging (is this correct?)

iamhopaul123: > I feel like it makes sense to not delegate the manifest responsibility to deploy package

That makes sense to me. I'll make the change in the new PR.

> since manifest operations are not unique to deployment - it is also needed for packaging (is this correct?)

yes it is required but the thing is i put here also because it is used in `packaging` since as I mentioned `GenerateCloudFormationTemplate` also needs the applied manifest.

internal/pkg/cli/deploy/deploy.go
iamhopaul123: Nvm boss updated the new PR to address this fb

PR #3219, opened 2022-01-25
chore: minor fix on template

Lou1415926:


efekarakus (approved):

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -145,11 +145,7 @@ NLBCustomDomainAction:
     ServiceName: !Ref WorkloadName
     RootDNSRole: {{ .AppDNSDelegationRole }}
     DomainName:  {{ .AppDNSName }}
-    Aliases:
-      !If
-        - HasNLBAliases
-        - []
-        - !Split [",", !Ref NLBAliases]
+    Aliases: !Split [",", !Ref NLBAliases]
```
huanjani, line 148: Do we need to avoid potentially returning an array of one empty string?

huanjani: Oh, I see where you handled this below.

huanjani (approved):

PR #3221, opened 2022-01-25
chore: fix security group ingress and `http` validation

Lou1415926:

### DIFF
--- a/internal/pkg/manifest/validate.go
+++ b/internal/pkg/manifest/validate.go
@@ -556,6 +556,9 @@ func (r RoutingRuleConfigOrBool) Validate() error {
 			missingField: "path",
 		}
 	}
+	if r.Enabled != nil {
+		return nil
+	}
 	return r.RoutingRuleConfiguration.Validate()
 }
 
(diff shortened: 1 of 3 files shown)


huanjani (approved):

iamhopaul123 (approved):

internal/pkg/manifest/validate.go
```diff
@@ -556,6 +556,9 @@ func (r RoutingRuleConfigOrBool) Validate() error {
 			missingField: "path",
 		}
 	}
+	if r.Enabled != nil {
```
efekarakus, line 559: I find this method a little confusing to grasp because we set `Enabled = nil` when the `RoutingRuleConfiguration` is not empty.
I'd have expected `Enabled = nil` and `Enabled = aws.Bool(false)` to have the same effect.

What do you think of instead setting `Enabled = true` when we can decode the configuration? and this method becomes:
```go
if !aws.BoolValue(r.Enabled) {
    return nil
}
return r.RoutingRuleConfiguration.Validate()
```

Instead the `path` check can be moved inside of `func (r RoutingRuleConfiguration) Validate() error`

PR #3223, opened 2022-01-26
feat: allow `secrets` to refer to SecretsManager secrets by name

### LINKED ISSUE #2898: Inject SecretsManager secrets by providing only the name
Today, users can inject secrets from AWS Secrets Manager by entering the entire ARN of the secret:
```yaml
secrets:
   VARIABLE_NAME: "'arn:aws:secretsmanager:region:accountid:secret:secretname:keyname::'"
```
However, users (#2132, #1449) expressed an easier way of integrating with Secrets Manager similar to the SSM experience in the manifest file.

### Proposal
In order to make it easier to inject Secrets Manager secrets to your service, we're proposing modifying the manifest like this:
```yaml
secrets:
   VARIABLE_NAME:
       secretsmanager: 'secretname'
```

If the value of your secret variable (`VARIABLE_NAME`) is a string, then we'll assume it's an SSM parameter name.  
On the other hand, if it's a map with a `secretsmanager` field then you'll be able to provide just the name for the secret instead of the entire ARN.

efekarakus:
Resolves #2898

I'll update the secrets for sidecars and logging in a separate PR later. This is just for the main container.

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -778,3 +778,18 @@ func convertHTTPVersion(protocolVersion *string) *string {
 	pv := strings.ToUpper(*protocolVersion)
 	return &pv
 }
+
+func convertSecrets(secrets map[string]manifest.Secret) map[string]template.Secret {
+	if len(secrets) == 0 {
+		return nil
+	}
+	m := make(map[string]template.Secret)
+	for name, mftSecret := range secrets {
+		var tplSecret template.Secret = template.SecretFromSSMOrARN(mftSecret.Value())
+		if mftSecret.IsSecretsManagerName() {
+			tplSecret = template.SecretFromSecretsManager(mftSecret.Value())
+		}
+		m[name] = tplSecret
+	}
+	return m
+}
--- a/internal/pkg/template/templates/workloads/partials/cf/secrets.yml
+++ b/internal/pkg/template/templates/workloads/partials/cf/secrets.yml
@@ -1,8 +1,14 @@
 {{- if hasSecrets .}}
-Secrets:{{range $name, $valueFrom := .Secrets}}
+Secrets:
+{{- range $name, $secret := .Secrets}}
 - Name: {{$name}}
-  ValueFrom: {{$valueFrom}}{{end}}{{end}}{{if .NestedStack}}{{$stackName := .NestedStack.StackName}}{{range $secret := .NestedStack.SecretOutputs}}
+  ValueFrom: {{if not $secret.RequiresSub }}{{$secret.ValueFrom}}{{- else}} !Sub 'arn:${AWS::Partition}:{{$secret.Service}}:${AWS::Region}:${AWS::AccountId}:{{$secret.ValueFrom}}' {{- end }}
+{{- end}}
+{{- end}}
+{{- if .NestedStack}}{{$stackName := .NestedStack.StackName}}
+{{- range $secret := .NestedStack.SecretOutputs}}
 - Name: {{toSnakeCase $secret}}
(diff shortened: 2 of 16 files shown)


efekarakus:
Adding "do-not-merge" until v1.14 is shipped, but the PR is reviewable!

internal/pkg/template/templates/workloads/partials/cf/secrets.yml
```diff
@@ -1,8 +1,14 @@
 {{- if hasSecrets .}}
-Secrets:{{range $name, $valueFrom := .Secrets}}
+Secrets:
+{{- range $name, $secret := .Secrets}}
 - Name: {{$name}}
-  ValueFrom: {{$valueFrom}}{{end}}{{end}}{{if .NestedStack}}{{$stackName := .NestedStack.StackName}}{{range $secret := .NestedStack.SecretOutputs}}
+  ValueFrom: {{if not $secret.RequiresSub }}{{$secret.ValueFrom}}{{- else}} !Sub 'arn:${AWS::Partition}:{{$secret.Service}}:${AWS::Region}:${AWS::AccountId}:{{$secret.ValueFrom}}' {{- end }}
```
iamhopaul123, line 5: nice 👏

iamhopaul123 (approved):
Awesome! LGTM :tada

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -778,3 +778,18 @@ func convertHTTPVersion(protocolVersion *string) *string {
...
 	return &pv
 }
+
+func convertSecrets(secrets map[string]manifest.Secret) map[string]template.Secret {
+	if len(secrets) == 0 {
+		return nil
+	}
+	m := make(map[string]template.Secret)
+	for name, mftSecret := range secrets {
+		var tplSecret template.Secret = template.SecretFromSSMOrARN(mftSecret.Value())
```
huanjani, line 788: Just curious about your choice of syntax here. Why not use the shorthand `:=` here?

internal/pkg/template/workload_test.go
```diff
@@ -290,3 +290,23 @@ func TestRuntimePlatformOpts_IsDefault(t *testing.T) {
...
+func TestSsmOrSecretARN_RequiresSub(t *testing.T) {
+	require.False(t, ssmOrSecretARN{}.RequiresSub(), "SSM Parameter Store or secret ARNs do not require !Sub")
+}
+
+func TestSsmOrSecretARN_ValueFrom(t *testing.T) {
+	require.Equal(t, "/github/token", SecretFromSSMOrARN("/github/token").ValueFrom())
+}
+
+func TestSecretsManagerName_RequiresSub(t *testing.T) {
+	require.True(t, secretsManagerName{}.RequiresSub(), "secrets refering a SecretsManager name need to be expanded to a full ARN")
```
huanjani: tiniest of tiny nits: 
```suggestion
	require.True(t, secretsManagerName{}.RequiresSub(), "secrets referring to a SecretsManager name need to be expanded to a full ARN")
```

huanjani (approved):
This is super neat and will be a huge usability boost for our customers!
Adding the DNM label in case you want to address the tiny nit before merging.

internal/pkg/deploy/cloudformation/stack/transformers.go
efekarakus: ooh because when using the shorthand it assigns the variable the concrete type `ssmOrARNSecret`, so later while trying to assign a `secretsManagerNameSecret` the compiler yells at us saying that the types mismatch.

Instead, here I'm explicitly saying use the `template.Secret` interface instead of the concrete type so that a secrets manager secret can be assigned to the same variable

PR #3230, opened 2022-01-28
chore(pipeline): detect current local branch for pipeline to follow

### LINKED ISSUE #2105: Pipeline config uses branch main instead of master

huanjani:
Fixes #2105.

Rather than injecting default branch names into pipeline manifests, we now detect the user's current branch to best-guess the branch they would like their pipeline to follow. They can always change this value in the manifest, as before. Added log statement to further bring attention to this behavior. Docs updated as well. 


internal/pkg/cli/pipeline_init.go
```diff
@@ -283,12 +285,24 @@ func (o *initPipelineOpts) askGitHubRepoDetails() error {
...
+	err := o.runner.Run("git", []string{"rev-parse", "--abbrev-ref", "HEAD"}, exec.Stdout(&o.buffer))
+	o.repoBranch = strings.TrimSpace(o.buffer.String())
+	if err != nil {
+		o.repoBranch = defaultBranch
+	}
+	if strings.TrimSpace(o.buffer.String()) == "" {
+		o.repoBranch = defaultBranch
+	}
+	o.buffer.Reset()
+	log.Infof("Your pipeline will follow branch '%s'. You may make changes in the pipeline manifest before deployment.\n", color.HighlightUserInput(o.repoBranch))
```
efekarakus: nit: (just to be within 80 chars for the default terminal screen size)
```suggestion
	log.Infof(`Your pipeline will follow branch '%s'. 
You may make changes in the pipeline manifest before deployment.
`, color.HighlightUserInput(o.repoBranch))
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -344,6 +352,13 @@ func (o *initPipelineOpts) selectURL() error {
 	}
 	o.buffer.Reset()
 
+	// If there is only one returned URL, set it rather than prompt to select.
+	if len(urls) == 1 {
+		log.Infof("Only one git repository detected. Your pipeline will follow '%s'. You may make changes in the generated pipeline manifest before deployment.\n", color.HighlightUserInput(urls[0]))
```
efekarakus: nit: similarly here can we split up the lines to be within 80 chars

efekarakus (approved):
Looks great! Two small logging nits 🥺 and :shipit:

iamhopaul123 (approved):

PR #3232, opened 2022-01-30
feat: respect AWS_CONFIG_PATH environment variable.

### LINKED ISSUE #2494: AWS Config environment variables ignored by copilot env init
Using a script to run copilot I have environment variables 
AWS_CONFIG_FILE and 
AWS_SHARED_CREDENTIALS_FILE and
AWS_DEFAULT_REGION 
pointing to an area that is not my default .aws directory location on an AWS EC2 instance.
"copilot env init ..." fails with the error:
✘ read named profiles: read AWS config file /dir1/.aws/config, you might need to run aws configure first: load ini file /dir1/.aws/config: open /dir1/.aws/config: no such file or directory
The message is correct as there is no .aws directory in that location, hence the use of the environment variables which are being ignored.
I have confirmed the variables are set in the script.

acomagu:
aws-cli uses $AWS_CONFIG_PATH environment variable as the path of the config file(instead of `~/.aws/config`). But Copilot CLI doesn't. This PR fixes the inconsistency.

Fixes #2494.

### DIFF
--- a/internal/pkg/aws/profile/profile.go
+++ b/internal/pkg/aws/profile/profile.go
@@ -30,12 +30,11 @@ type Config struct {
 
 // NewConfig returns a new parsed Config object from $HOME/.aws/config.
 func NewConfig() (*Config, error) {
-	homeDir, err := os.UserHomeDir()
+	cfgPath, err := cfgPath()
 	if err != nil {
-		return nil, fmt.Errorf("get home directory: %w", err)
+		return nil, fmt.Errorf("get the path of configuration file: %w", err)
 	}
 
-	cfgPath := filepath.Join(homeDir, awsCredentialsDir, awsConfigFileName)
 	cfg, err := ini.New(cfgPath)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


iamhopaul123 (approved):
This is awesome! Thank you so much for contributing ❤️

internal/pkg/aws/profile/profile.go
```diff
@@ -55,3 +54,16 @@ func (c *Config) Names() []string {
 	}
 	return profiles
 }
+
+func cfgPath() (string, error) {
+	if os.Getenv("AWS_CONFIG_FILE") != "" {
+		return os.Getenv("AWS_CONFIG_FILE"), nil
+	}
```
Lou1415926, line 61: nit:
```suggestion
	if cfgPath := os.Getenv("AWS_CONFIG_FILE"); cfgPath != "" {
		return cfgPath, nil
	}
```

Lou1415926 (approved):
Thank you so much for contributing!

PR #3239, opened 2022-02-02
refactor(cli): apply deploy pkg to simplify workload deploy

iamhopaul123:


internal/pkg/cli/svc_deploy.go
```diff
@@ -367,195 +361,63 @@ func (o *deploySvcOpts) configureClients() error {
...
-			ImageTag: o.imageTag,
-			Digest:   o.imageDigest,
-		},
-		ServiceDiscoveryEndpoint: endpoint,
-		AccountID:                o.targetEnvironment.AccountID,
-		Region:                   o.targetEnvironment.Region,
-	}, nil
-}
-
 func uploadRDWSCustomResources(o *uploadCustomResourcesOpts, appEnvResources *stack.AppRegionalResources) (map[string]string, error) {
```
iamhopaul123, line 354: pending removal after refactoring workload package

internal/pkg/cli/svc_deploy.go
```diff
@@ -572,221 +434,6 @@ func uploadRDWSCustomResources(o *uploadCustomResourcesOpts, appEnvResources *st
...
-	}
-	mf, ok := unmarshaledManifest.(envFile)
-	if ok {
-		return mf.EnvFile()
-	}
-	// If the manifest type doesn't support envFiles, ignore and move forward.
-	return ""
-}
-
 func validateLBSvcAlias(aliases manifest.Alias, app *config.Application, envName string) error {
```
iamhopaul123, line 370: samesies

internal/pkg/cli/deploy/deploy.go
```diff
@@ -258,12 +226,12 @@ func (d *workloadDeployer) UploadArtifacts(in *UploadArtifactsInput) (*UploadArt
...
 // DeployWorkload deploys a workload using CloudFormation.
 func (d *workloadDeployer) DeployWorkload(in *DeployWorkloadInput) (*DeployWorkloadOutput, error) {
-	stackConfigoutput, err := d.stackConfiguration(in)
+	stackConfigOutput, err := d.stackConfiguration(in)
 	if err != nil {
 		return nil, err
 	}
-	cmdRunAt := in.now()
-	if err := in.ServiceDeployer.DeployService(os.Stderr, stackConfigoutput.conf, d.s3Bucket,
+	cmdRunAt := in.Now()
```
Lou1415926: I feel like it's fine to do `cmdRunAt := time.Now()` instead of passing the function in, since it doesn't care about any information about the service/env/app. What do you think?

Lou1415926 (commented):
Just a general discussion that doesn't need to be addressed ⬇️ 

Lots of clients are used only during deployment operations. They are currently configured in `cli` and then passed to `cli/deploy`.

I wonder if it's desirable to offload the configuration of these clients from `cli` package to `cli/deploy` pacakge. In this case, `cli` may just need to pass the service info (name, env, app, etc.) and the sessions to `cli/deploy`.

The motivation is that with the current implementation, if we need to add more client dependency for the deployment operations, we need to 
1.  extend the `DeployWorkloadInput` struct
2. configure the client in `svc_deploy.go`, store it as a member of `svcDeployOpts`
3. pass the client to the extended `DeployWorkloadInput`

It feels like a lot of linear growth of the codebase, and a lot of repetitive information being stored and passed around.

internal/pkg/cli/deploy/deploy.go
iamhopaul123: it's to facilitate unit test so that we can override the time.

internal/pkg/cli/svc_deploy.go
```diff
@@ -140,31 +98,49 @@ func newSvcDeployOpts(vars deployWkldVars) (*deploySvcOpts, error) {
...
+		Name:     o.name,
+		App:      o.targetApp,
+		Env:      o.targetEnv,
+		ImageTag: o.imageTag,
+		S3Bucket: o.appResources.S3Bucket,
+		Mft:      o.appliedManifest,
+	}
+	switch t := o.appliedManifest.(type) {
+	case *manifest.LoadBalancedWebService:
+		deployer, err = clideploy.NewLBDeployer(&in)
```
Lou1415926, line 122: 😍

internal/pkg/cli/svc_deploy.go
```diff
@@ -310,252 +284,92 @@ func (o *deploySvcOpts) askEnvName() error {
...
-	if err != nil {
-		return fmt.Errorf("create describer for environment %s in application %s: %w", o.envName, o.appName, err)
-	}
-	o.envDescriber = d
-
-	// ECR client against tools account profile AND target environment region.
-	repoName := fmt.Sprintf("%s/%s", o.appName, o.name)
-	registry := ecr.New(defaultSessEnvRegion)
-	o.imageBuilderPusher, err = repository.New(repoName, registry)
+	o.imageTag = imageTagFromGit(o.cmd, o.imageTag) // Best effort assign git tag.
```
Lou1415926, line 280: A question! How do we decide what clients should be configured here versus `cli/deploy`?

Lou1415926 (commented):
Love the new changes!

internal/pkg/cli/deploy.go
```diff
@@ -55,10 +51,6 @@ func newDeployOpts(vars deployWkldVars) (*deployOpts, error) {
 	if err != nil {
 		return nil, fmt.Errorf("new workspace: %w", err)
 	}
-	deployStore, err := deploy.NewStore(store)
```
efekarakus, line 58: yay all this red!

internal/pkg/cli/svc_deploy.go
iamhopaul123: It's just intuitively since we get the value from the tag and assign default value if they don't specify the flag, I feel it should be still in the `cli` package as it is input from users like `app`, `env`, and `svc`.

Lou1415926: ohhh wrong place to add the comment! It was more like a general questions, e.g. what about `appVersionGetter` and `appRegionalResources`? Are they configured in `cli` for the same reason?

iamhopaul123: ooh good point i think i just missed them

internal/pkg/cli/deploy/deploy.go
```diff
@@ -154,46 +150,221 @@ type WorkloadDeployerInput struct {
...
+		DeployStore: deployStore,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("create describer for environment %s in application %s: %w", in.Env.Name, in.App.Name, err)
+	}
+	return &lbSvcDeployer{
+		svcDeployer:            svcDeployer,
+		appVersionGetter:       versionGetter,
+		publicCIDRBlocksGetter: envDescriber,
+		lbMft:                  in.Mft.(*manifest.LoadBalancedWebService),
```
efekarakus: should we ensure that we don't panic with a statement like this:
```go
lbMft, ok := in.Mft.(*manifest.LoadBalancedWebService)
if !ok {
    return nil, fmt.Errorf("manifest is not of type %s", manifest.LoadBalancedWebService)
}
```

(same feedback for other constructors)

internal/pkg/cli/deploy/deploy.go
```diff
@@ -256,48 +407,168 @@ func (d *workloadDeployer) UploadArtifacts(in *UploadArtifactsInput) (*UploadArt
...
+		rdwsAlias: stackConfigOutput.rdSvcAlias,
+	}, nil
+}
+
+type workerSvcDeployOutput struct {
+	subs []manifest.TopicSubscription
+}
+
+// RecommendedActions returns the recommended actions after deployment.
+func (d *workerSvcDeployOutput) RecommendedActions() []string {
```
efekarakus, line 488: nice boss I really like this, do you feel like this made the codebase better too?

internal/pkg/cli/deploy/deploy.go
```diff
@@ -465,115 +736,182 @@ func (d *workloadDeployer) runtimeConfig(in *DeployWorkloadInput) (*stack.Runtim
...
 	rc, err := d.runtimeConfig(in)
 	if err != nil {
 		return nil, err
 	}
-	in.NewSvcUpdater(func(s *session.Session) ServiceForceUpdater {
-		return ecs.New(s)
-	})
-	var output stackConfigurationOutput
-	switch t := d.mft.(type) {
-	case *manifest.LoadBalancedWebService:
```
efekarakus, line 484: omg yes, getting rid of this switch statement 🔥 million times yes for this refactor

internal/pkg/cli/interfaces.go
```diff
@@ -21,6 +20,7 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/aws/codepipeline"
 	awsecs "github.com/aws/copilot-cli/internal/pkg/aws/ecs"
 	"github.com/aws/copilot-cli/internal/pkg/aws/s3"
+	clideploy "github.com/aws/copilot-cli/internal/pkg/cli/deploy"
 	"github.com/aws/copilot-cli/internal/pkg/config"
 	"github.com/aws/copilot-cli/internal/pkg/deploy"
```
efekarakus, line 25: man we have so many deploys 😂 not in this refactor obviously, but I think `pkg/deploy` is just noise. We can delete it and move almost all the logic to the existing packages like `manifest`, `cloudformation/stack` or `template`

efekarakus (commented):
so much better, definitely a step in the right direction 👍

efekarakus (approved):

internal/pkg/cli/deploy/deploy.go
iamhopaul123: yeah that's truee

internal/pkg/cli/interfaces.go
iamhopaul123: I'll add it to ongoing issues.

Lou1415926 (approved):

Issue #3231, opened 2022-01-29
[Design] Enable Multiple Pipelines


huanjani:
## Background
Though we have a [workaround](https://github.com/aws/copilot-cli/discussions/1925) for enabling multiple Copilot pipelines, we need a more built-in solution. Currently, pipeline names are auto-assigned using the "pipeline-[appName]-[repoName]" format. By changing the pipeline names in the pipeline manifest (`pipeline.yml`) before deployment (as suggested in the workaround), multiple pipelines can be created. For users following the branch-based development model, however, having different pipelines follow different branches becomes problematic when it comes time to merge the branches-- the generically-named pipeline files attempt to merge, but the user wants them to remain distinct for the separate pipelines.
Related: #1921

## Proposal
By allowing users to name their pipelines, we will avoid namespace collisions for pipelines within the same app and following the same repository. Pipeline files (`pipeline.yml` and `buildspec.yml`) will be housed within a `pipelines` subdirectory.
```
SRC_DIR
|---copilot
    |---svcA
        |---manifest.yml
    |---svcB
        |---manifest.yml
    |---pipeline.yml (pre-existing pipeline)
    |---buildspec.yml (pre-existing buildspec)
    |---pipelines
        |---dev
            |---manifest.yml
            |---buildspec.yml
        |---main
            |---manifest.yml
            |---buildspec.yml
        |---prod
            |---manifest.yml
            |---buildspec.yml
```
## Implementation Plan (Draft)
- [x] add pipeline selector to `pipeline deploy`, `pipeline status`, `pipeline show` https://github.com/aws/copilot-cli/pull/3382
- [x] add pipelines subdirectory and pathway to store pipeline manifests and buildspecs
- [x] change pipeline manifest and buildspec naming conventions https://github.com/aws/copilot-cli/pull/3381
- [x] prompt for pipeline name in `pipeline init`; handle existing pipeline name https://github.com/aws/copilot-cli/pull/3381
- [x] look for pipeline files in legacy location and new location #3267
- [x] e2e/integ tests
- [x] docs
- [x] opt: add `--local` flag to `pipeline ls`
- [ ] opt: enable workload selection (beyond selectively committing workload subdirectories)

wstewarttennes:
@huanjani do you have an ETA for when this will be live?

huanjani:
@wstewarttennes-- thank you for your patience! We are shooting to do another release the week of April 4. This feature should be included in that release!

[...]

PR #3247, opened 2022-02-03
chore(pipelines): add `--name` flag to `pipeline deploy`

huanjani:
Part of #3231.

### DIFF
--- a/internal/pkg/cli/pipeline_deploy.go
+++ b/internal/pkg/cli/pipeline_deploy.go
@@ -45,7 +45,7 @@ const connectionsURL = "https://console.aws.amazon.com/codesuite/settings/connec
 
 type deployPipelineVars struct {
 	appName          string
-	name			 string
+	name             string
 	skipConfirmation bool
 }
 
--- a/internal/pkg/cli/pipeline_deploy.go
+++ b/internal/pkg/cli/pipeline_deploy.go
@@ -45,6 +45,7 @@ const connectionsURL = "https://console.aws.amazon.com/codesuite/settings/connec
 
 type deployPipelineVars struct {
 	appName          string
+	name			 string
 	skipConfirmation bool
 }
 
@@ -60,7 +61,6 @@ type deployPipelineOpts struct {
 	ws               wsPipelineReader
 	codestar         codestar
 
-	pipelineName                 string
 	shouldPromptUpdateConnection bool
 }
 
@@ -100,6 +100,11 @@ func newDeployPipelineOpts(vars deployPipelineVars) (*deployPipelineOpts, error)
 
 // Validate returns an error if the flag values passed by the user are invalid.
 func (o *deployPipelineOpts) Validate() error {
+	if o.name != "" {
+		if err := o.validatePipelineName(); err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


internal/pkg/cli/pipeline_deploy.go
```diff
@@ -123,10 +128,13 @@ func (o *deployPipelineOpts) Execute() error {
 	if err != nil {
 		return fmt.Errorf("unmarshal pipeline manifest: %w", err)
 	}
-	if len(pipeline.Name) > 100 {
-		return fmt.Errorf(`pipeline name '%s' must be shorter than 100 characters`, pipeline.Name)
+	// Set pipeline name if not passed in with flag.
+	if o.name == "" {
+		if len(pipeline.Name) > 100 {
+			return fmt.Errorf(`pipeline name '%s' must be shorter than 100 characters`, pipeline.Name)
+		}
```
efekarakus: What do you think of moving this logic to `manifest/validate.go`?
```go
func (m *PipelineManifest) Validate() error 
```
That could be a nice way of having a similar interface between workload manifests and pipeline manifests.
And then we can remove `if o.name == ""` replacing it with:
```go
if err := pipeline.Validate(); err != nil { ... }
```

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -173,6 +181,22 @@ func (o *deployPipelineOpts) Execute() error {
 	return nil
 }
 
+func (o *deployPipelineOpts) validatePipelineName() error {
+	// TODO: After pipeline file structure changes, validate with o.ws.ListPipelines() instead.
```
efekarakus: nit: can we remove this TODO? you'll be addressing this soon anyway so I'm not super worried. 

side note: This is probably just me but I've never seen an effective TODO comment 🙈 I feel like they just pollute the codebase with things that will never happen and create confusion if they're not addressed quickly

efekarakus (commented):
Looks great! just a suggestion for validation.

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -114,7 +119,7 @@ func (o *deployPipelineOpts) Execute() error {
 	}
 	o.prog.Stop(log.Ssuccessf(fmtPipelineDeployResourcesComplete, color.HighlightUserInput(o.appName)))
 
-	// read pipeline manifest
+	// Read pipeline manifest.
 	data, err := o.ws.ReadPipelineManifest()
```
Lou1415926, line 123: Quick question: If there are multiple pipeline in the workspace, which pipeline manifest should we be reading?

internal/pkg/manifest/validate.go
```diff
@@ -401,6 +401,15 @@ func (s ScheduledJobConfig) Validate() error {
 	return nil
 }
 
+// Validate returns nil if the pipeline manifest is configured correctly.
+func (p PipelineManifest) Validate() error {
```
efekarakus, line 405: nit: @iamhopaul123 was there a reason why we don't want to have the receivers as pointers? since these are large structs, I wonder if we should make them pointers for performance reasons
```suggestion
func (p *PipelineManifest) Validate() error {
```

internal/pkg/cli/pipeline_deploy.go
huanjani: That will be added/addressed in a future PR. 😸

efekarakus (approved):
:shipit:

Lou1415926 (approved):
LGTM! Added the DNM label just in case you / @iamhopaul123  wants to followup with Efe's question.

huanjani:
> LGTM! Added the DNM label just in case you / @iamhopaul123 wants to followup with Efe's question.

Thanks! I think if we make a change we can do it as a separate PR and to the whole file.

PR #3249, opened 2022-02-04
refactor(cli): simplify workload package

iamhopaul123:

### DIFF
--- a/internal/pkg/cli/deploy/deploy.go
+++ b/internal/pkg/cli/deploy/deploy.go
@@ -402,8 +402,8 @@ type UploadArtifactsOutput struct {
 	AddonsURL   string
 }
 
-// StackConfiguration contains info to configure a workload CloudFormation stack.
-type StackConfiguration struct {
+// StackRuntimeConfiguration contains runtime configuration for a workload CloudFormation stack.
+type StackRuntimeConfiguration struct {
 	ImageDigest string
 	EnvFileARN  string
 	AddonsURL   string
@@ -413,7 +413,7 @@ type StackConfiguration struct {
 
 // DeployWorkloadInput is the input of DeployWorkload.
 type DeployWorkloadInput struct {
-	StackConfiguration
+	StackRuntimeConfiguration
 	ForceNewUpdate bool
 }
 
@@ -441,7 +441,7 @@ func (d *workloadDeployer) UploadArtifacts() (*UploadArtifactsOutput, error) {
 
 // GenerateCloudFormationTemplateInput is the input of GenerateCloudFormationTemplate.
 type GenerateCloudFormationTemplateInput struct {
-	StackConfiguration
+	StackRuntimeConfiguration
 }
 
 // GenerateCloudFormationTemplateOutput is the output of GenerateCloudFormationTemplate.
@@ -453,7 +453,7 @@ type GenerateCloudFormationTemplateOutput struct {
 // GenerateCloudFormationTemplate genrates a CloudFormation template and parameters for a workload.
 func (d *lbSvcDeployer) GenerateCloudFormationTemplate(in *GenerateCloudFormationTemplateInput) (
 	*GenerateCloudFormationTemplateOutput, error) {
-	output, err := d.stackConfiguration(&in.StackConfiguration)
+	output, err := d.stackConfiguration(&in.StackRuntimeConfiguration)
 	if err != nil {
 		return nil, err
 	}
@@ -462,7 +462,7 @@ func (d *lbSvcDeployer) GenerateCloudFormationTemplate(in *GenerateCloudFormatio
 
 // DeployWorkload deploys a load balanced web service using CloudFormation.
 func (d *lbSvcDeployer) DeployWorkload(in *DeployWorkloadInput) (ActionRecommender, error) {
-	stackConfigOutput, err := d.stackConfiguration(&in.StackConfiguration)
+	stackConfigOutput, err := d.stackConfiguration(&in.StackRuntimeConfiguration)
 	if err != nil {
 		return nil, err
 	}
@@ -475,7 +475,7 @@ func (d *lbSvcDeployer) DeployWorkload(in *DeployWorkloadInput) (ActionRecommend
 // GenerateCloudFormationTemplate genrates a CloudFormation template and parameters for a workload.
 func (d *backendSvcDeployer) GenerateCloudFormationTemplate(in *GenerateCloudFormationTemplateInput) (
 	*GenerateCloudFormationTemplateOutput, error) {
-	output, err := d.stackConfiguration(&in.StackConfiguration)
+	output, err := d.stackConfiguration(&in.StackRuntimeConfiguration)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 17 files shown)


internal/pkg/cli/deploy/deploy.go
```diff
@@ -397,13 +402,18 @@ type UploadArtifactsOutput struct {
 	AddonsURL   string
 }
 
+// StackConfiguration contains info to configure a workload CloudFormation stack.
+type StackConfiguration struct {
```
efekarakus: nit: Is there a way of making the comment or struct name more specific, what makes these configuration fields unique?
Technically manifest file values also allow you to configure a workload CFN stack, so I want to highlight which fields would be okay to add here in the future

internal/pkg/cli/deploy/deploy.go
```diff
@@ -429,25 +439,56 @@ func (d *workloadDeployer) UploadArtifacts() (*UploadArtifactsOutput, error) {
...
-	if err := d.deploy(in, *stackConfigOutput); err != nil {
+	if err := d.deploy(in.ForceNewUpdate, *stackConfigOutput); err != nil {
 		return nil, err
 	}
 	return nil, nil
 }
 
+// GenerateCloudFormationTemplate genrates a CloudFormation template and parameters for a workload.
+func (d *backendSvcDeployer) GenerateCloudFormationTemplate(in *GenerateCloudFormationTemplateInput) (
+	*GenerateCloudFormationTemplateOutput, error) {
```
efekarakus, line 477: wut is this new line expected?

internal/pkg/cli/deploy/deploy.go
```diff
@@ -557,21 +608,37 @@ func (d *jobDeployer) DeployWorkload(in *DeployWorkloadInput) (ActionRecommender
...
+	if err != nil {
+		return nil, fmt.Errorf("generate stack template configuration: %w", err)
+	}
+	return &GenerateCloudFormationTemplateOutput{
+		Template:   tpl,
+		Parameters: params,
+	}, nil
+}
+
+func (d *svcDeployer) deploy(forceNewUpdate bool, stackConfigOutput svcStackConfigurationOutput) error {
```
efekarakus, line 627: super nit: boolean parameters are usually considered a code smell if it's part of the _public_ API. Since it's not here, I don't care much 
https://martinfowler.com/bliki/FlagArgument.html

but if it's easy to refactor consider having two methods:
```go
func (d *svcDeployer) forceDeploy(out svcStackConfigOutput) error
func (d *svcDeployer) deploy(out svcStackConfigOutput) error
```
If it's a pain, leave it as is 👍

internal/pkg/cli/deploy/deploy.go
```diff
@@ -610,6 +677,26 @@ func (d *workloadDeployer) forceDeploy(in *forceDeployInput) error {
...
+		topicName := template.StripNonAlphaNumFunc(aws.StringValue(subscription.Name))
+		subName := fmt.Sprintf("%s%sEventsQueue", topicSvc, strings.Title(topicName))
+		if first {
+			sb.WriteString(subName)
+			first = false
+		} else {
+			sb.WriteString(fmt.Sprintf(", %s", subName))
+		}
+	}
+	return sb.String()
```
efekarakus: nit:
```suggestion
	var queueNames []string
	for _, subscription := range d.subs {
		if subscription.Queue.IsEmpty() {
			continue
		}
		svc := template.StripNonAlphaNumFunc(aws.StringValue(subscription.Service))
		topic := template.StripNonAlphaNumFunc(aws.StringValue(subscription.Name))
		queueNames = append(queueNames,  fmt.Sprintf("%s%sEventsQueue", svc, strings.Title(topic)))
	}
	return strings.Join(queueNames, ", ")
```

internal/pkg/cli/job_package.go
```diff
@@ -76,16 +66,6 @@ func newPackageJobOpts(vars packageJobVars) (*packageJobOpts, error) {
 		prompt:         prompter,
 	}
 
-	opts.stackSerializer = func(mft interface{}, env *config.Environment, app *config.Application, rc stack.RuntimeConfig) (stackSerializer, error) {
```
efekarakus, line 79: 😍

efekarakus (approved):

internal/pkg/deploy/cloudformation/stack/pipeline.go
```diff
@@ -19,17 +19,21 @@ type pipelineStackConfig struct {
 	parser template.Parser
 }
 
+// NewEnvStackConfig sets up a struct which can provide values to CloudFormation for
```
Lou1415926:
```suggestion
// NewPipelineStackConfig sets up a struct which can provide values to CloudFormation for
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -557,21 +608,37 @@ func (d *jobDeployer) DeployWorkload(in *DeployWorkloadInput) (ActionRecommender
...
-func (d *svcDeployer) deploy(in *DeployWorkloadInput, stackConfigOutput svcStackConfigurationOutput) error {
+func (d *workloadDeployer) generateCloudFormationTemplate(conf stackSerializer) (
+	*GenerateCloudFormationTemplateOutput, error) {
+	tpl, err := conf.Template()
+	if err != nil {
+		return nil, fmt.Errorf("generate stack template: %w", err)
+	}
+	params, err := conf.SerializedParameters()
+	if err != nil {
+		return nil, fmt.Errorf("generate stack template configuration: %w", err)
```
Lou1415926: was this meant to be ⬇️ ?
```suggestion
		return nil, fmt.Errorf("generate stack template parameters: %w", err)
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -429,25 +439,56 @@ func (d *workloadDeployer) UploadArtifacts() (*UploadArtifactsOutput, error) {
...
+
+// GenerateCloudFormationTemplateOutput is the output of GenerateCloudFormationTemplate.
+type GenerateCloudFormationTemplateOutput struct {
+	Template   string
+	Parameters string
+}
+
+// GenerateCloudFormationTemplate genrates a CloudFormation template and parameters for a workload.
+func (d *lbSvcDeployer) GenerateCloudFormationTemplate(in *GenerateCloudFormationTemplateInput) (
+	*GenerateCloudFormationTemplateOutput, error) {
```
Lou1415926, line 455: What do you think of ⬇️ ?
```
func (d *workloadDeployer) GenerateCloudFormationTemplate() ()
```
and 
```
func (d *serviceDeployer) DeployWorkload() ()
```
To extract the common logic shared by all workload/service ?

internal/pkg/cli/deploy/deploy.go
iamhopaul123: Otherwise it would be too long boss

internal/pkg/cli/deploy/deploy.go
iamhopaul123: That's a good point! But since it's not a public one and splitting it into two would add extra complexity. I'll just leave it as is.

internal/pkg/cli/deploy/deploy.go
iamhopaul123: Hmm there aren't many common logics other than two function calls which should be fine 🤔 (since I've already consolidated most common logic in `workloadDeployer.generateCloudFormationTemplate`)  And the main reason is because `stackConfiguration` is defined per workload type so that `workloadDeployer` doesn't have `stackConfiguration` as its method.

Lou1415926 (approved):

Issue #2641, opened 2021-07-21
`copilot task run` with `--secrets` option raise error AccessDeniedException


shgtkshruch:
Hi, there. I would like to run database migration command with [`copilot task run`](https://aws.github.io/copilot-cli/en/docs/commands/task-run/).

But `copilot task run` command with `--secrets` option raise `AccessDeniedException` error.
Secrets is created by [`copilot secret init`](https://aws.github.io/copilot-cli/en/docs/commands/secret-init/) command, and store in Parameter Store.

![スクリーンショット 2021-07-21 12 51 22](https://user-images.githubusercontent.com/5207601/126428047-7312dc44-b5e9-4a65-bd53-b0fb10bdafa1.png)

```sh
$ copilot --version
1.8.3

$ copilot task run \
  --app chronos \
  --env dev \
  --image 123456789.dkr.ecr.ap-northeast-1.amazonaws.com/chronos/rails:a40efe96b90bf506d834a57a5e476ecdbc935793 \
  --command 'bin/rails db:prepare' \
  --env-vars RAILS_LOG_TO_STDOUT=true,RAILS_ENV=dev \
  --secrets RAILS_MASTER_KEY=/copilot/chronos/dev/secrets/RAILS_MASTER_KEY \
  --follow

- No new infrastructure changes for stack task-chronosnos 
✘ Failed to run chronos.be running for chronos. 

✘ run task chronos: run task chronos: task 1c350295: ResourceInitializationError: unable to pull secrets or registry auth: execution resource retrieval failed: unable to retrieve secrets from ssm: service call has been retried 1 time(s): AccessDeniedException: User: arn:aws:sts::123456789:assumed-role...
```

ECS console is bellow.

![スクリーンショット 2021-07-21 12 30 51](https://user-images.githubusercontent.com/5207601/126427551-f5d48fe9-1c3b-48f0-aae6-4b5a37fe8922.png)

I think task execution role dose not have enough authority to pull secrets from Parameter Store. I know copilot task run has `--execution-role` option, but I don't know what value should specify for this option. How can I fix this error? Thank you!

Lou1415926:
Hi @shgtkshruch!

You are correct - the default execution role for the task doesn't have permission to pull secrets from ssm at this moment in order to limit the scope of a one-off task. 

The easiest way to achieve your goal is to modify the existing default execution role we've created and attached to the task for you.

[...]

PR #3256, opened 2022-02-08
fix(task): grant task-run execution role access to secrets based on tag

paragbhingre:
This PR fixes the issue #2641 

Fixed the issue by adding policies to the execution role when `--secrets` are provided. 

Tests - 
Following manual tests have been taken into consideration
1. When we send both `--app` and `--env` flags in `copilot task run` command then secrets are accessible only when `copilot-application` and `copilot-environment` tags are present with the correct values on the secret
2. When we send both `--app` and `--env` flags in `copilot task run` command then secrets are inaccessible only when `copilot-application` and `copilot-environment` tags are NOT present on the secret or they have incorrect values


internal/pkg/template/templates/task/cf.yml
```diff
@@ -89,6 +100,22 @@ Resources:
...
         - !Sub 'arn:${AWS::Partition}:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy'
+      {{- if .Secrets}}
+      Policies:
+        - PolicyName: 'PullSecrets'
+          PolicyDocument:
+            Version: '2012-10-17'
+            Statement:
+              - Effect: 'Allow'
+                Action:
+                - ssm:GetParameters
```
Lou1415926: Should we probably also take `secretsmanager` into account 🤔

internal/pkg/template/templates/task/cf.yml
```diff
@@ -89,6 +100,22 @@ Resources:
...
+            Version: '2012-10-17'
+            Statement:
+              - Effect: 'Allow'
+                Action:
+                - ssm:GetParameters  
+                Condition:
+                  StringEquals:
+                    "ssm:ResourceTag/copilot-application": !If [ HasAppAndEnv, !Ref App, !Ref "AWS::NoValue" ]
+                    "ssm:ResourceTag/copilot-environment": !If [ HasAppAndEnv, !Ref Env, !Ref "AWS::NoValue" ]
+                    "ssm:ResourceTag/copilot-task": !If [ HasAppAndEnv, !Ref "AWS::NoValue", !Ref TaskName ]
```
efekarakus: I really like the changes with allowing access to the `copilot-application` and `copilot-environment` tags  -- makes total sense to me. I am not sure though how obvious it's going to be when not using a copilot managed cluster, requiring the `copilot-task` tag for a one-time job seems a bit cumbersome.

What if we had an experience like this?

```
$ copilot task run --default --secrets hello=/hello,world=arn:aws:secretsmanager:us-west-2:111122223333:secret:aes128-1a2b3c
Looks like you're requesting ssm:GetParameters to the following SSM parameters:
* /hello 
and secretsmanager:GetSecretValue to the following Secrets Manager secrets:
* arn:aws:secretsmanager:us-west-2:111122223333:secret:aes128-1a2b3c

Do you grant permission to the ECS/Fargate agent for these secrets? (y/N)
```

Or if they'd like to skip the prompt they can pass a flag similar to what the CDK does with `--require-approval never`

```
$ copilot task run --default --secrets hello=/hello --require-approval never
```

Then we will augment the Task Execution role such that only the input resources are granted access.

Let me know what you think!

iamhopaul123: > Do you grant permission to the ECS/Fargate agent for these secrets? (y/N)

If they choose "N", do we just error out saying they can't use these secrets without granting access? And what are available options for `--require-approval`?

efekarakus: > If they choose "N", do we just error out saying they can't use these secrets without granting access? 

I assume we'll abort then, but we can look to what the CDK does as well.

> And what are available options for --require-approval?

hmm good point for us none, maybe we can flip the flag `--acknowledge-iam`

efekarakus (approved):
Looks good to me! Just to validate we tested it with both secrets (ssm and secretsmanager) and it works right? 😂 

Similarly, can we test as well with a secret that's not tagged to make sure it still cannot access it

paragbhingre:
> Looks good to me! Just to validate we tested it with both secrets (ssm and secretsmanager) and it works right? 😂
> 
> Similarly, can we test as well with a secret that's not tagged to make sure it still cannot access it

Both the secrets (ssm and secretsmanager) and scenarios you mentioned are tested ✅

internal/pkg/template/templates/task/cf.yml
```diff
@@ -39,6 +43,13 @@ Conditions:
     !Not [ !Equals [ !Join [ "", !Ref EntryPoint ], "" ] ]
   HasCustomPlatform:
     !Not [!Equals [!Ref OS, ""]]
+  HasApp:
+    !Not [!Equals [!Ref App, ""]]
+  HasEnv:
+    !Not [!Equals [!Ref Env, ""]]
```
iamhopaul123: Do we really need hasapp and has env?

internal/pkg/template/templates/task/cf.yml
```diff
@@ -89,6 +100,34 @@ Resources:
...
+                  Action:
+                    - secretsmanager:GetSecretValue
+                  Condition:
+                    StringEquals:
+                      "secretsmanager:ResourceTag/copilot-application": !If [ HasAppAndEnv, !Ref App, !Ref "AWS::NoValue" ]
+                      "secretsmanager:ResourceTag/copilot-environment": !If [ HasAppAndEnv, !Ref Env, !Ref "AWS::NoValue" ]
+                  Resource:
+                    - !Sub arn:${AWS::Partition}:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:*
+          - !Ref AWS::NoValue
+      {{- end}}
```
iamhopaul123: This is awesome!

internal/pkg/template/templates/task/cf.yml
paragbhingre: Actually we can remove this and put these conditions directly into `HasAppAndEnv` condition.

internal/pkg/template/templates/task/cf.yml
```diff
@@ -89,6 +96,34 @@ Resources:
...
+            PolicyDocument:
+              Version: '2012-10-17'
+              Statement:
+                - Effect: 'Allow'
+                  Action:
+                  - ssm:GetParameters
+                  Condition:
+                    StringEquals:
+                      "ssm:ResourceTag/copilot-application": !If [ HasAppAndEnv, !Ref App, !Ref "AWS::NoValue" ]
+                      "ssm:ResourceTag/copilot-environment": !If [ HasAppAndEnv, !Ref Env, !Ref "AWS::NoValue" ]
```
Lou1415926: Can we remove these  `!If` conditionals now that we have line 101-102?

internal/pkg/template/templates/task/cf.yml
```diff
@@ -89,6 +96,34 @@ Resources:
...
+                  Resource:
+                    - !Sub arn:${AWS::Partition}:ssm:${AWS::Region}:${AWS::AccountId}:parameter/*
+                - Effect: 'Allow'
+                  Action:
+                    - secretsmanager:GetSecretValue
+                  Condition:
+                    StringEquals:
+                      "secretsmanager:ResourceTag/copilot-application": !If [ HasAppAndEnv, !Ref App, !Ref "AWS::NoValue" ]
+                      "secretsmanager:ResourceTag/copilot-environment": !If [ HasAppAndEnv, !Ref Env, !Ref "AWS::NoValue" ]
```
Lou1415926: Samesies!

Lou1415926 (approved):

PR #3258, opened 2022-02-09
fix: stop truncating logs to 10 at a time when running `svc logs --follow`

### LINKED ISSUE #3220: `svc logs --follow` doesn't print all logs
Expected behavior
`copilot svc logs --follow` should print all lines printed by the application since the `copilot svc logs --follow` command was executed.

Actual behavior
`copilot svc logs --follow` only prints 10-line "chunks" of output. It doesn't print all lines. In practice, it seems to print the last 10 lines resulting from a test request. I'm assuming the fact that the "chunk" is 10 lines of output is related to the default value for `--limit` (10).

I seem to be able to get around this with `copilot svc logs --follow --limit 10000`, but this feels clumsy and I'm not confident that all logged lines are being printed.

Am I misunderstanding how `copilot svc logs --follow` is intended to be used?

efekarakus:
Fixes #3220 

### DIFF
--- a/internal/pkg/logging/service.go
+++ b/internal/pkg/logging/service.go
@@ -81,7 +81,7 @@ func (o WriteLogEventsOpts) startTime(now func() time.Time) *int64 {
 	}
 	if o.Follow {
 		// Start following log events from current timestamp.
-		return aws.Int64(now().Unix() * 1000) // Multiplied by 1000 to convert the Unix() timestamp in seconds to milliseconds
+		return aws.Int64(now().UnixMilli())
 	}
 	return nil
 }
--- a/internal/pkg/logging/service.go
+++ b/internal/pkg/logging/service.go
@@ -65,12 +65,16 @@ func (o WriteLogEventsOpts) limit() *int64 {
 	if o.Limit != nil {
 		return o.Limit
 	}
-	if o.StartTime != nil || o.EndTime != nil {
+	if o.hasTimeFilters() {
 		// If time filtering is set, then set limit to be maximum number.
 		// https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_GetLogEvents.html#CWL-GetLogEvents-request-limit
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


internal/pkg/logging/service.go
```diff
@@ -65,14 +67,29 @@ func (o WriteLogEventsOpts) limit() *int64 {
...
 	return aws.Int64(defaultServiceLogsLimit)
 }
 
+func (o WriteLogEventsOpts) startTime(now func() time.Time) *int64 {
+	if o.StartTime != nil {
+		return o.StartTime
+	}
+	if o.Follow {
+		// Start following log events from current timestamp.
+		return aws.Int64(now().Unix() * 1000) // Multiplied by 1000 to convert the Unix() timestamp in seconds to milliseconds
```
Lou1415926: nit: [or alternatively](https://pkg.go.dev/time#Time.UnixMilli)...
```suggestion
		return aws.Int64(now().UnixMilli())
```

Lou1415926 (approved):

internal/pkg/logging/service_test.go
```diff
@@ -54,9 +55,9 @@ firelens_log_router/fcfe4 10.0.0.00 - - [01/Jan/1970 01:01:01] "WARN some warnin
 		},
 	}
 	mockLimit := aws.Int64(100)
-	mockDefaultLimit := aws.Int64(10)
 	var mockNilLimit *int64
 	mockStartTime := aws.Int64(123456789)
+	mockCurrentTimestamp := time.Date(2020, 11, 23, 0, 0, 0, 0, time.UTC) // Copilot GA date :).
```
huanjani, line 60: 😻

internal/pkg/logging/service.go
efekarakus: wow I had no idea this method existed 😂 fixed!

I also changed it in other places in the codebase where we multiplied by 1000

iamhopaul123 (approved):

PR #3259, opened 2022-02-09
chore(templates): render observability configuration and necessary permissions

Lou1415926:

### DIFF
--- a/internal/pkg/template/templates/workloads/services/rd-web/cf.yml
+++ b/internal/pkg/template/templates/workloads/services/rd-web/cf.yml
@@ -134,7 +134,7 @@ Resources:
           EgressType: VPC
           VpcConnectorArn: !Ref VpcConnector
       {{- end }}
-      {{- if and (contains .FeatureFlags "Tracing") (eq .Tracing "AWSXRAY")}}
+      {{- if and (contains .FeatureFlags "Tracing") (eq .Observability.Tracing "AWSXRAY")}}
       ObservabilityConfiguration:
         ObservabilityEnabled: true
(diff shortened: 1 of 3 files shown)


internal/pkg/template/workload.go
```diff
@@ -488,6 +488,7 @@ type WorkloadOpts struct {
 	// Additional options for request driven web service templates.
 	StartCommand      *string
 	EnableHealthCheck bool
+	Tracing           string
```
efekarakus: nit: What do you think of moving this into something like this:
```suggestion
	Observability           ObservabilityOpts
```
```go
type ObservabilityOpts struct {
    Tracing string
}
```

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
```diff
@@ -134,6 +134,11 @@ Resources:
           EgressType: VPC
           VpcConnectorArn: !Ref VpcConnector
       {{- end }}
+      {{- if and (contains .FeatureFlags "Tracing") (eq .Tracing "AWSXRAY")}}
+      ObservabilityConfiguration:
+        ObservabilityEnabled: true
+        ObservabilityConfigurationArn: !Sub 'arn:aws:apprunner:${AWS::Region}:${AWS::AccountId}:observabilityconfiguration/DefaultConfiguration/1/00000000000000000000000000000001'
```
iamhopaul123, line 140: Are they going to create a default configuration for each region with ARN like this?

internal/pkg/template/workload.go
Lou1415926: Updated!

efekarakus (approved):

iamhopaul123 (approved):

PR #3257, opened 2022-02-08
chore: avoid init unnecessary clients for workload package

iamhopaul123:
https://github.com/aws/copilot-cli/pull/3249 introduces a breaking change for pipeline users because it initiates ECR client to call `DescribeRepositories` which is not needed for our workload package and is not permitted by our pipeline codebuild role.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the Apache 2.0 License.

PR #3261, opened 2022-02-09
chore(repository): do not call API within constructor

iamhopaul123:
https://github.com/aws/copilot-cli/pull/3249 introduces a breaking change for pipeline users because it initiates ECR client to call DescribeRepositories which is not needed for our workload package and is not permitted by our pipeline codebuild role.

Prefer this change over https://github.com/aws/copilot-cli/pull/3257 because we want to init client within the `cli/deploy` package so that we can make `cli` package cleaner. Otherwise, it doesn't make a lot of sense for us to init these unrelated clients. The worst case all the `svc/job deploy/package` will have to init all the dependencies for `deploy`, which is very confusing and cumbersome.

### DIFF
--- a/internal/pkg/repository/repository.go
+++ b/internal/pkg/repository/repository.go
@@ -7,8 +7,6 @@ package repository
 import (
 	"fmt"
 
-	"github.com/aws/aws-sdk-go/aws/session"
-	"github.com/aws/copilot-cli/internal/pkg/aws/ecr"
 	"github.com/aws/copilot-cli/internal/pkg/docker/dockerengine"
 )
 
@@ -28,24 +26,17 @@ type Registry interface {
 
 // Repository builds and pushes images to a repository.
 type Repository struct {
-	name     string
-	registry Registry
-}
-
-// New instantiates a new Repository.
-func New(name string, sess *session.Session) (*Repository, error) {
-	return &Repository{
-		name:     name,
-		registry: ecr.New(sess),
-	}, nil
+	Name     string
+	Registry Registry
+	Uri      string
 }
 
 // BuildAndPush builds the image from Dockerfile and pushes it to the repository with tags.
 func (r *Repository) BuildAndPush(docker ContainerLoginBuildPusher, args *dockerengine.BuildArguments) (digest string, err error) {
 	if args.URI == "" {
-		uri, err := r.registry.RepositoryURI(r.name)
+		uri, err := r.URI()
 		if err != nil {
-			return "", fmt.Errorf("get repository URI: %w", err)
+			return "", err
 		}
 		args.URI = uri
 	}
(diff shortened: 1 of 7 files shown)


internal/pkg/repository/repository.go
```diff
@@ -28,28 +30,24 @@ type Registry interface {
...
 }
 
 // New instantiates a new Repository.
-func New(name string, registry Registry) (*Repository, error) {
-	uri, err := registry.RepositoryURI(name)
-	if err != nil {
-		return nil, fmt.Errorf("get repository URI: %w", err)
-	}
-
+func New(name string, sess *session.Session) (*Repository, error) {
```
efekarakus: can we take a `Registry` as input? I don't think any other package besides `cli` should actually initialize clients so that we do dependency injection and reuse clients

```suggestion
func New(name string, registry Registry) (*Repository, error) {
```

efekarakus (commented):
Looks great, just a small feedback

internal/pkg/repository/repository.go
iamhopaul123: I actually was in two minds about this and changed at the last minute. It's because if I want the `cli` package to avoid directly using any lower level `aws/` packages, and control by passing sessions instead. What do you think?

efekarakus: That also makes sense, but I think we should establish a convention that encourages more taking interfaces as inputs if you're not the `cli` pkg so that engineers aren't confused how to create constructors.

It's okay if the `cli` package is a bit verbose, but the majority of the logic will be just about initializing clients

internal/pkg/repository/repository.go
```diff
@@ -28,28 +30,24 @@ type Registry interface {
 type Repository struct {
 	name     string
 	registry Registry
-
-	uri string
 }
 
 // New instantiates a new Repository.
-func New(name string, registry Registry) (*Repository, error) {
-	uri, err := registry.RepositoryURI(name)
```
Lou1415926, line 37: :homer-disappearing

Lou1415926 (commented):
nothing much to add on top of what efe suggested!

internal/pkg/repository/repository.go
iamhopaul123: oke done

internal/pkg/repository/repository.go
```diff
@@ -26,55 +26,51 @@ type Registry interface {
...
-	}
-
-	return &Repository{
-		name:     name,
-		uri:      uri,
-		registry: registry,
-	}, nil
+	Name     string
+	Registry Registry
+	Uri      string
```
efekarakus: nit: 
```suggestion
	URI      string
```

efekarakus: nit: I think constructors can still be useful to keep these fields private and provide different ways of initializing the object for example:
```go
type Repository struct {
  name string
  registry Registry
  uri string
}
func New(registry Registry, name string) *Repository

func NewWithURI(registry Registry, name, uri string) *Repository
```

efekarakus (approved):

Lou1415926 (approved):
lgtm!

PR #3267, opened 2022-02-10
chore(pipelines): add workspace pipeline lister

huanjani:
Related: #3231

This PR:
- adds the `--name` flag to the `pipeline delete` command and validates that the specified pipeline is among the deployed pipelines
- adds the ability to fetch the names of local workspace pipelines
- adds a pipeline name selector to `deploy`; only local workspace pipelines are included. If only one pipeline is detected (current state for all Copilot workspaces), default-picks that one.

The next PR will:
- add pipeline name selectors to `delete` (only deployed), `show` (only deployed)
- fix buggy `Ask()` functionality in `show` and `status`


internal/pkg/cli/pipeline_delete.go
```diff
@@ -130,26 +138,27 @@ func (o *deletePipelineOpts) Execute() error {
...
 		return fmt.Errorf("read pipeline manifest: %w", err)
 	}
-
 	pipeline, err := manifest.UnmarshalPipeline(data)
 	if err != nil {
 		return fmt.Errorf("unmarshal pipeline manifest: %w", err)
 	}
-
-	o.PipelineName = pipeline.Name
+	o.name = pipeline.Name
```
iamhopaul123: Does it mean we potentially override the pipeline name they give by unmarshaling the manifest?

internal/pkg/cli/pipeline_delete.go
```diff
@@ -88,21 +91,26 @@ func (o *deletePipelineOpts) Validate() error {
...
+			return err
+		}
 	}
-
 	return nil
 }
 
 // Ask prompts for fields that are required but not passed in.
 func (o *deletePipelineOpts) Ask() error {
+	if err := o.getNameAndSecret(); err != nil {
```
iamhopaul123, line 95: Do we ask any question here 🤔

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -180,17 +193,30 @@ func (o *deployPipelineOpts) Execute() error {
 }
 
 func (o *deployPipelineOpts) validatePipelineName() error {
```
iamhopaul123, line 228: Why is it different from how we validate pipeline name in `pipeline delete`?

internal/pkg/term/selector/selector.go
```diff
@@ -123,6 +124,16 @@ type WsWorkloadLister interface {
 	ListWorkloads() ([]string, error)
 }
 
+// WsPipelineLister wraps the method to get pipelines in the current workspace.
+type WsPipelineLister interface {
```
iamhopaul123: Can we make this private?

internal/pkg/term/selector/selector.go
```diff
@@ -884,6 +932,14 @@ func (s *WorkspaceSelect) retrieveWorkspaceWorkloads() ([]string, error) {
 	return localWlNames, nil
 }
 
+func (s *PipelineSelect) retrieveWorkspacePipelines() (map[string]string, error) {
```
iamhopaul123: Maybe we don't need this method as it doesn't add much value 🤔

internal/pkg/term/selector/selector.go
```diff
@@ -884,6 +932,14 @@ func (s *WorkspaceSelect) retrieveWorkspaceWorkloads() ([]string, error) {
...
+	localPipelineNames, err := s.ws.ListPipelines()
+	if err != nil {
+		return nil, err
```
iamhopaul123: We might want to wrap this error as it is a public API call

internal/pkg/cli/pipeline_delete.go
```diff
@@ -74,6 +76,7 @@ func newDeletePipelineOpts(vars deletePipelineVars) (*deletePipelineOpts, error)
 		deletePipelineVars: vars,
 		prog:               termprogress.NewSpinner(log.DiagnosticWriter),
 		prompt:             prompt.New(),
+		pipelineSvc:        codepipeline.New(defaultSess),
```
efekarakus: nit:
```suggestion
		codepipeline:        codepipeline.New(defaultSess),
```

internal/pkg/cli/pipeline_delete.go
```diff
@@ -36,22 +37,23 @@ var (
...
+	name               string
 	skipConfirmation   bool
 	shouldDeleteSecret bool
 }
 
 type deletePipelineOpts struct {
 	deletePipelineVars
 
-	PipelineName   string
 	PipelineSecret string
```
efekarakus: nit:
```suggestion
	ghAccessTokenSecretName string
```

internal/pkg/cli/pipeline_delete.go
```diff
@@ -216,6 +225,7 @@ func buildPipelineDeleteCmd() *cobra.Command {
 		}),
 	}
 	cmd.Flags().StringVarP(&vars.appName, appFlag, appFlagShort, tryReadingAppName(), appFlagDescription)
+	cmd.Flags().StringVarP(&vars.name, nameFlag, nameFlagShort, "", pipelineFlagDescription)
```
efekarakus, line 213: Is this command idempotent?
I dunno what's the behavior of doing `copilot svc delete -n frontend` twice, do we return an error the second time or succeed silently? As long as the two commands are consistent I'm okay with either implementation

internal/pkg/workspace/workspace.go
```diff
@@ -478,17 +528,16 @@ func (ws *Workspace) ListDockerfiles() ([]string, error) {
 	return dockerfiles, nil
 }
 
-// RelPath returns the path relative to the current working directory.
-func RelPath(fullPath string) (string, error) {
-	wkdir, err := os.Getwd()
+func (ws Workspace) PipelineNameFromManifest(path string) (string, error) {
```
efekarakus: Instead of having a separate method for the pipeline name, what do you think of changing the signature for `ReadPipelineManifest` to this?
```go
ReadPipelineManifest(path string) (*manifest.PipelineManifest, error)
```
this way the client can retrieve on their own whatever field is exposed in the `PipelineManifest`.

efekarakus: _sidenote nit_: In a separate PR, I think we should rename `manifest.PipelineManifest` to `manifest.Pipeline` so that we don't repeat ourselves

internal/pkg/term/selector/selector.go
```diff
@@ -163,13 +174,19 @@ type ConfigSelect struct {
...
+// WorkspaceSelect  is an application and environment selector, but can also choose a service or pipeline from the workspace.
 type WorkspaceSelect struct {
 	*Select
 	ws      WorkspaceRetriever
 	appName string
 }
 
+// PipelineSelect is a workspace pipeline selector.
+type PipelineSelect struct {
+	*Select
```
efekarakus: Do we need `*Select` or can we just get by with the following since `PipelineSelect` won't prompt for app or env names:
```suggestion
	prompt Prompter
```

internal/pkg/term/selector/selector.go
```diff
@@ -249,6 +266,13 @@ func NewWorkspaceSelect(prompt Prompter, store ConfigLister, ws WorkspaceRetriev
 	}
 }
 
+// NewWsPipelineSelect returns a new selector with pipelines from the local workspace.
+func NewWsPipelineSelect(ws WsPipelineSelector) *PipelineSelect {
+	return &PipelineSelect{
+		ws: ws,
```
efekarakus: do we need to initialize `*Select` or `prompter` here so that we don't panic, we most likely didn't run into an issue because there is always a single pipeline for now 😂

internal/pkg/term/selector/selector.go
```diff
@@ -673,6 +697,30 @@ func filterWlsByName(wls []*config.Workload, wantedNames []string) []string {
...
 }
 
+// Pipeline fetches all the pipelines in a workspace and prompts the user to select one.
+func (s *PipelineSelect) Pipeline(msg, help string) (name, path string, err error) {
+	pipelines, err := s.retrieveWorkspacePipelines()
+	if err != nil {
+		return "", "", err
+	}
+	if len(pipelines) == 0 {
+		return "", "", errors.New("no pipelines found")
```
efekarakus:
```suggestion
		return "", "", errors.New("no pipelines found in the workspace")
```

internal/pkg/term/selector/selector.go
```diff
@@ -163,13 +174,19 @@ type ConfigSelect struct {
 	workloadLister ConfigWorkloadLister
 }
 
-// WorkspaceSelect  is an application and environment selector, but can also choose a service from the workspace.
+// WorkspaceSelect  is an application and environment selector, but can also choose a service or pipeline from the workspace.
```
efekarakus: Is this accurate? to select a pipeline as well I would have expected `WorkspaceSelect` to have a `pipelineSelect` field of type `PipelineSelect`

internal/pkg/term/selector/selector.go
```diff
@@ -884,6 +932,14 @@ func (s *WorkspaceSelect) retrieveWorkspaceWorkloads() ([]string, error) {
 	return localWlNames, nil
 }
 
+func (s *PipelineSelect) retrieveWorkspacePipelines() (map[string]string, error) {
+	localPipelineNames, err := s.ws.ListPipelines()
+	if err != nil {
+		return nil, err
+	}
+	return localPipelineNames, nil
+}
```
efekarakus: Looks like this is the same as:
```go
func (s *PipelineSelect) retrieveWorkspacePipelines() (map[string]string, error) {
	return s.ws.ListPipelines()
}
```
Maybe we can delete this method altogether since it doesn't do any additional computation?

internal/pkg/workspace/workspace.go
```diff
@@ -335,6 +377,14 @@ func (ws *Workspace) pipelineManifestPath() (string, error) {
 	return pipelineManifestPath, nil
 }
 
+func (ws *Workspace) pipelineManifestPath() (string, error) {
```
efekarakus: nit:
```suggestion
func (ws *Workspace) pipelinesDirPath() (string, error) {
```

internal/pkg/workspace/workspace.go
```diff
@@ -151,6 +152,52 @@ func (ws *Workspace) ListWorkloads() ([]string, error) {
...
+	}
+	pipelines[name] = legacyPath
+	// Look for other pipelines.
+	pipelinesPath, err := ws.pipelineManifestPath()
+	if err != nil {
+		return nil, err
+	}
+	exists, err := ws.fsUtils.Exists(pipelinesPath)
+	if err != nil {
+		return nil, err
```
efekarakus: nit: should we wrap this err?

internal/pkg/workspace/workspace.go
```diff
@@ -151,6 +152,52 @@ func (ws *Workspace) ListWorkloads() ([]string, error) {
...
+			if strings.HasSuffix(file.Name(), "buildspec.yml") {
+				continue
+			}
+			// Read manifests of moved legacy pipeline and any other pipelines.
+			if strings.HasSuffix(file.Name(), ".yml") {
+				path := filepath.Join(pipelinesPath, file.Name())
+				name, err := ws.PipelineNameFromManifest(path)
+				if err != nil {
+					return nil, fmt.Errorf("read pipeline manifest: %w", err)
+				}
```
efekarakus: what do you think of not erroring but instead maybe `log.Info` this and skipping the file and `continue`ing? this way we can fetch as many pipeline paths as we can without stopping

efekarakus (commented):
Looks great! just a bunch of small feedback

internal/pkg/cli/pipeline_delete.go
huanjani: Good question. Yes, they behave similarly-- both fail validation:
```
✘ get pipeline pipelineName: PipelineNotFoundException: Account 'xxxxxxxxxxxxx' does not have a pipeline with name 'pipelineName'
```
and
```
✘ couldn't find service svcName in the application appName
```

internal/pkg/term/selector/selector.go
huanjani: Oops 🙈 ! That was a holdover from a previous version. Deleted.

internal/pkg/workspace/workspace.go
huanjani: Ah yes, much better and consistent.

internal/pkg/cli/pipeline_delete.go
huanjani: Good call. Made this conditional-- only sets name here if `o.name == ""`.

internal/pkg/cli/pipeline_delete.go
huanjani: Not yet, but in the next PR, we'll have a selector with all pipeline names (or default to one if only one found).

internal/pkg/cli/pipeline_deploy.go
huanjani: For `pipeline deploy` we look for the name in pipeline manifests because they may have only run `pipeline init` and the pipeline is not yet deployed.
For `pipeline delete`, we can only delete pipelines that have been deployed, so we look for deployed pipelines instead of local ones.

internal/pkg/cli/pipeline_delete.go
```diff
@@ -130,65 +136,61 @@ func (o *deletePipelineOpts) Execute() error {
...
 	if err != nil {
-		return fmt.Errorf("unmarshal pipeline manifest: %w", err)
+		return err
 	}
-
-	o.PipelineName = pipeline.Name
-
-	if secret, ok := (pipeline.Source.Properties["access_token_secret"]).(string); ok {
-		o.PipelineSecret = secret
+	if o.name == "" {
```
iamhopaul123, line 140: <strike>This doesn't look right..so if `o.name` was set, which means it was set by the flag, and we should validate in `Validate()` func. And in `Ask()` I think we should assume `o.name` is correct and use it as the pipeline name to get the pipeline manifest.</strike>

Actually I think the key is we should check if `o.name` is empty earlier rather than after getting `manifest.Name`

```go
if o.name == "" {
 // try to get the pipeline name
}
// read pipeline manifest
```

internal/pkg/cli/pipeline_delete.go
```diff
@@ -130,65 +136,61 @@ func (o *deletePipelineOpts) Execute() error {
...
-		}
-		return fmt.Errorf("read pipeline manifest: %w", err)
+		return err
 	}
-
-	pipeline, err := manifest.UnmarshalPipeline(data)
+	manifest, err := o.ws.ReadPipelineManifest(path)
 	if err != nil {
-		return fmt.Errorf("unmarshal pipeline manifest: %w", err)
+		return err
```
iamhopaul123: Maybe wrap the error returned by a public method?

internal/pkg/workspace/workspace.go
```diff
@@ -326,7 +378,7 @@ func (ws *Workspace) writeSummary(appName string) error {
 	return ws.fsUtils.WriteFile(summaryPath, serializedWorkspaceSummary, 0644)
 }
 
-func (ws *Workspace) pipelineManifestPath() (string, error) {
+func (ws *Workspace) PipelineManifestLegacyPath() (string, error) {
```
iamhopaul123: Comment pls

internal/pkg/cli/pipeline_delete.go
```diff
@@ -130,65 +136,61 @@ func (o *deletePipelineOpts) Execute() error {
...
-
-	if secret, ok := (pipeline.Source.Properties["access_token_secret"]).(string); ok {
-		o.PipelineSecret = secret
+	if o.name == "" {
+		o.name = manifest.Name
 	}
 
+	if secret, ok := (manifest.Source.Properties["access_token_secret"]).(string); ok {
+		o.ghAccessTokenSecretName = secret
+	}
```
iamhopaul123, line 146: Since it is not info from the customer input, I would suggest moving this part into `Execute()` before where we use it.

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -54,13 +57,15 @@ type deployPipelineOpts struct {
...
+	sel              wsPipelineSelector
 	prog             progress
 	prompt           prompter
 	region           string
 	envStore         environmentStore
 	ws               wsPipelineReader
 	codestar         codestar
 
+	path                         string
 	shouldPromptUpdateConnection bool
```
iamhopaul123: Maybe add a comment specifying they are cache variables, so that users would know they don't need to init these values.

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -54,13 +57,15 @@ type deployPipelineOpts struct {
...
 	app              *config.Application
+	sel              wsPipelineSelector
 	prog             progress
 	prompt           prompter
 	region           string
 	envStore         environmentStore
 	ws               wsPipelineReader
 	codestar         codestar
 
+	path                         string
```
iamhopaul123: Can we avoid using cache variables as possible (unless to facilitate unit test since the initialization is not typically conditional except for in the unit test we override it)?

The reason is because it can be very stateful and caller would be difficult to know what the current state is (if the initialization is conditional).

internal/pkg/cli/pipeline_delete.go
huanjani: 🤔  I think I know what you're getting at, but we have to read the manifest either way, so this way we don't read it twice.

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -180,17 +189,30 @@ func (o *deployPipelineOpts) Execute() error {
...
-	if pipeline.Name != o.name {
-		return fmt.Errorf(`pipeline %s not found in the workspace`, color.HighlightUserInput(o.name))
+	return fmt.Errorf(`pipeline %s not found in the workspace`, color.HighlightUserInput(o.name))
+}
+
+func (o *deployPipelineOpts) askPipelineName() error {
+	if o.name != "" {
+		return nil
 	}
+	name, path, err := o.sel.Pipeline(pipelineSelectPrompt, "")
```
iamhopaul123: Based on the reason above, i would suggest `sel.Pipeline()` to only return the name of the pipeline.

internal/pkg/cli/pipeline_delete.go
```diff
@@ -88,21 +89,26 @@ func (o *deletePipelineOpts) Validate() error {
...
+			return err
+		}
 	}
-
 	return nil
 }
 
 // Ask prompts for fields that are required but not passed in.
 func (o *deletePipelineOpts) Ask() error {
+	if err := o.getNameAndSecret(); err != nil {
```
efekarakus, line 95: It is pretty strange to parse and set these values in `Ask`, I'd have expected the parsing to happen in `Execute`.

Not in this PR! But I think it might be worthwhile moving all this logic into `Execute` instead 🤔 and having `Ask` here just call `PipelineSelector` for deployed pipelines. Do you agree?

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -108,6 +114,14 @@ func (o *deployPipelineOpts) Validate() error {
 	return nil
 }
 
+// Ask prompts the user for any required fields that are not provided.
+func (o *deployPipelineOpts) Ask() error {
+	if err := o.askPipelineName(); err != nil {
+		return err
+	}
+	return nil
```
efekarakus: nit:
```suggestion
	return o.askPipelineName()
```

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -180,17 +189,30 @@ func (o *deployPipelineOpts) Execute() error {
...
 	if err != nil {
-		return fmt.Errorf("read pipeline manifest: %w", err)
+		return fmt.Errorf("list pipelines: %w", err)
 	}
-	pipeline, err := manifest.UnmarshalPipeline(data)
-	if err != nil {
-		return fmt.Errorf("unmarshal pipeline manifest: %w", err)
+	for name, path := range pipelines {
+		if name == o.name {
+			o.path = path
```
efekarakus: It totally makes sense to put `o.path = path` here for performance reasons, but I feel like the most robust place where we need to guarantee that `path` has been set is in `Execute`.

What do you think of additionally creating a method for `Execute` that's like:
```go
func (o *deployPipelineOpts) Execute() error {
   if err := cachePipelinePath(); err != nil { return nil }
   ...
}

func (o *deployPipelineOpts) cachePipelinePath() error {
   if o.path != "" { return nil }
   pipelines, err := o.ws.ListPipelines()
   ...
}
```

internal/pkg/term/selector/selector.go
efekarakus: ping: don't we need to initialize the `prompt` field?

internal/pkg/term/selector/selector.go
efekarakus: ping :)

internal/pkg/workspace/workspace.go
```diff
@@ -151,6 +153,52 @@ func (ws *Workspace) ListWorkloads() ([]string, error) {
...
+	// Look for legacy pipeline.
+	legacyPath, err := ws.PipelineManifestLegacyPath()
+	if err != nil {
+		return nil, err
+	}
+	manifest, err := ws.ReadPipelineManifest(legacyPath)
+	if err != nil {
+		return nil, err
+	}
+	pipelines[manifest.Name] = legacyPath
```
efekarakus: Shouldn't we skip if we receive `ErrNoPipelineInWorkspace` instead of returning the error?

internal/pkg/workspace/workspace.go
```diff
@@ -202,21 +250,25 @@ func (ws *Workspace) ReadWorkloadManifest(mftDirName string) (WorkloadManifest,
...
 	}
 	if !manifestExists {
 		return nil, ErrNoPipelineInWorkspace
 	}
-	return ws.read(pipelineFileName)
+	data, err := ws.fsUtils.ReadFile(path)
+	if err != nil {
+		log.Infof("Unable to read pipeline manifest file at '%s'", path)
+		return nil, nil
+	}
```
efekarakus: hmm I think it makes sense for `ReadPipelineManifest` to return an error if the `path` is invalid and it should be up to the client code to skip the error or not.

efekarakus (commented):
👍 small feedback and then we can :shipit:

internal/pkg/workspace/workspace.go
```diff
@@ -151,6 +153,52 @@ func (ws *Workspace) ListWorkloads() ([]string, error) {
...
+	// Look for other pipelines.
+	pipelinesPath, err := ws.pipelinesDirPath()
+	if err != nil {
+		return nil, err
+	}
+	exists, err := ws.fsUtils.Exists(pipelinesPath)
+	if err != nil {
+		return nil, fmt.Errorf("check if pipeline manifest exists at %s: %w", pipelinesPath, err)
+	}
+	if exists {
```
iamhopaul123: Can we do if it doens't exist then exit?

internal/pkg/workspace/workspace.go
```diff
@@ -151,6 +153,52 @@ func (ws *Workspace) ListWorkloads() ([]string, error) {
 	})
 }
 
+// ListPipelines returns the names of all pipelines in the workspace.
+func (ws *Workspace) ListPipelines() (map[string]string, error) {
```
iamhopaul123: Maybe this could return to an array of object as opposed to a map? Because map is not helpful as an API output for the users to find out what the key-value means.

internal/pkg/cli/pipeline_delete.go
huanjani: Oh, looks like you're caching the manifest in #3283, so that changes things here. Will adjust.

internal/pkg/cli/pipeline_deploy.go
huanjani: I think we need to cache this variable bc it's info we need to use in multiple places and retrieve ourselves (as opposed to getting it from the user). And it looks like you have added cached variables to this file in #3283, so these can go together.

internal/pkg/cli/pipeline_deploy.go
huanjani: We need the path in order to locate the correct pipeline manifest. I have a method that returns the path based on the name, which we can use if we want this selector to just return the name, but then we'd have repetitive code in each file that needs to specify a pipeline.

internal/pkg/cli/pipeline_deploy.go
huanjani: I see where you're coming from. The problem is that we can't figure out the path until we have the name because we don't know which pipeline is the one to deploy. Because `o.name` and `o.path` are being set/confirmed together (either in `Ask()` or `Validate()`, depending on the name flag), is it weird to double-check that the path has been set, but assuming that the name has been set in order to check? Does that make sense?

internal/pkg/cli/pipeline_deploy.go
huanjani: Discussed offline.

internal/pkg/cli/pipeline_deploy.go
huanjani: Discussed offline.

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -68,6 +71,8 @@ type deployPipelineOpts struct {
 	newSvcListCmd    func(io.Writer) cmd
 	newJobListCmd    func(io.Writer) cmd
 
+	// cached variables
```
efekarakus, line 77: l. 78 has the same comment for `pipelineMft`, I think maybe this is a result of merging PRs. Can we consolidate both blocks?

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -227,23 +228,49 @@ func (o *deployPipelineOpts) Execute() error {
...
+	pipelineMft, err := o.ws.ReadPipelineManifest(path)
 	if err != nil {
-		return nil, fmt.Errorf("unmarshal pipeline manifest: %w", err)
+		return nil, fmt.Errorf("read pipeline manifest: %w", err)
 	}
-	if err := pipeline.Validate(); err != nil {
+	if err := pipelineMft.Validate(); err != nil {
 		return nil, fmt.Errorf("validate pipeline manifest: %w", err)
 	}
-	o.pipelineMft = pipeline
```
efekarakus, line 244: why did we remove this caching line?

internal/pkg/term/selector/selector.go
```diff
@@ -71,7 +71,8 @@ const (
 	taskFinalMsg        = "Task:"
 	workloadFinalMsg    = "Name:"
 	dockerfileFinalMsg  = "Dockerfile:"
-	topicFinalmessage   = "Topic subscriptions:"
+	topicFinalMessage   = "Topic subscriptions:"
```
efekarakus: nit: thanks for fixing this, for consistency:
```suggestion
	topicFinalMsg   = "Topic subscriptions:"
```

internal/pkg/term/selector/selector.go
```diff
@@ -123,6 +124,16 @@ type WsWorkloadLister interface {
 	ListWorkloads() ([]string, error)
 }
 
+// wsPipelineLister wraps the method to get pipelines in the current workspace.
+type wsPipelineLister interface {
+	ListPipelines() ([]workspace.Pipeline, error)
+}
+
+// WsPipelineSelector is a pipeline selector.
+type WsPipelineSelector interface {
```
efekarakus: Do we need this interface? it seems to wrap `wsPipelineLister`, maybe we can make that one public and rename like this:
```go
type WorkspacePipelinesLister interface {
    ListPipelines() ([]workspace.Pipeline, error)
}
```

internal/pkg/term/selector/selector.go
```diff
@@ -673,6 +698,36 @@ func filterWlsByName(wls []*config.Workload, wantedNames []string) []string {
...
+	}
+	var pipelineNames []string
+	for _, pipeline := range pipelines {
+		pipelineNames = append(pipelineNames, pipeline.Name)
+	}
+	if len(pipelineNames) == 1 {
+		log.Infof("Only found one pipeline; defaulting to: %s\n", color.HighlightUserInput(pipelineNames[0]))
+		return &workspace.Pipeline{
+			Name: pipelineNames[0],
+			Path: s.pipelinePath(pipelines, pipelineNames[0]),
```
efekarakus: nit:
```suggestion
			Name: pipelines[0].Name,
			Path: pipelines[0].Path,
```

internal/pkg/workspace/workspace.go
```diff
@@ -160,6 +162,66 @@ func (ws *Workspace) ListWorkloads() ([]string, error) {
 	})
 }
 
+// Pipeline holds identifying information about a pipeline.
+type Pipeline struct {
```
efekarakus: What do you think of renaming it to:
```suggestion
// PipelineManifest holds identifying information about a pipeline manifest file.
type PipelineManifest struct {
     Name string // Name of the pipeline inside the manifest file.
     Path string // Absolute path to the manifest file for the pipeline.
```
so that it reads `workspace.PipelineManifest` instead of `workspace.Pipeline`

internal/pkg/workspace/workspace.go
```diff
@@ -160,6 +162,66 @@ func (ws *Workspace) ListWorkloads() ([]string, error) {
...
+		return nil, err
+	}
+	manifest, err := ws.ReadPipelineManifest(legacyPath)
+	if err != nil {
+		log.Infof("Unable to read pipeline manifest at '%s'", legacyPath)
+	}
+	pipelines = append(pipelines, Pipeline{
+		Name: manifest.Name,
+		Path: legacyPath,
+	})
```
efekarakus: Right now if `ReadPipelineManifest` fails because the file doesn't exist, then we still append a `Pipeline` with `manifest.Name = ""`.

How about the following:
```go
var manifests []PipelineManifest

legacyMft, err := ws.legacyPipelineManifest()
if err != nil && !errors.Is(err, ErrNoPipelineInWorkspace) {
    return err
}
if err == nil {
   manifests = append(manifests, legacyMft)
}
```

internal/pkg/workspace/workspace.go
```diff
@@ -160,6 +162,66 @@ func (ws *Workspace) ListWorkloads() ([]string, error) {
...
+		if strings.HasSuffix(file.Name(), ".yml") {
+			path := filepath.Join(pipelinesPath, file.Name())
+			manifest, err := ws.ReadPipelineManifest(path)
+			if err != nil {
+				log.Infof("Unable to read pipeline manifest at '%s'", path)
+			}
+			pipelines = append(pipelines, Pipeline{
+				Name: manifest.Name,
+				Path: path,
+			})
```
efekarakus: Do we need an else statement here? Otherwise we will keep appending Pipeline objects with an empty name when there is an error:
```suggestion
			if err != nil {
				ws.logger.Infof("Unable to read pipeline manifest at '%s'", path)
			} else {
			        pipelines = append(pipelines, Pipeline{
				       Name: manifest.Name,
				       Path: path,
			        })
			 }
```

efekarakus: There is also a subtle change of moving the log to a `logger` field in the `Workspace` object that gets initialized in the constructor

internal/pkg/workspace/workspace.go
```diff
@@ -317,6 +382,16 @@ func IsInGitRepository(fs FileStat) bool {
...
 }
 
+// PipelineManifestLegacyPath returns the path to pipeline manifests before multiple pipelines (and the copilot/pipelines/ dir) were enabled.
+func (ws *Workspace) PipelineManifestLegacyPath() (string, error) {
+	copilotPath, err := ws.copilotDirPath()
+	if err != nil {
+		return "", err
+	}
+	pipelineManifestPath := filepath.Join(copilotPath, pipelineFileName)
+	return pipelineManifestPath, nil
```
efekarakus: nit: same change as below:
```suggestion
	return filepath.Join(copilotPath, pipelineFileName), nil
```

internal/pkg/workspace/workspace.go
huanjani: Okay! 
Then I'll also rename `manifest.PipelineManifest` to `manifest.Pipeline` now rather than later, to prevent confusion.

internal/pkg/workspace/workspace.go
```diff
@@ -162,25 +165,28 @@ func (ws *Workspace) ListWorkloads() ([]string, error) {
...
 	manifest, err := ws.ReadPipelineManifest(legacyPath)
+	if err != nil && !errors.Is(err, ErrNoPipelineInWorkspace) {
+		return nil, err
+	}
 	if err != nil {
-		log.Infof("Unable to read pipeline manifest at '%s'", legacyPath)
+		ws.logger("Unable to read pipeline manifest at '%s'", legacyPath)
 	}
-	pipelines = append(pipelines, Pipeline{
+	pipelineManifests = append(pipelineManifests, PipelineManifest{
```
efekarakus: if `err != nil` then we are still appending the following:
```go
PipelineManifest{Name: "", Path: legacyPath}
```

can we instead change it to:
```go
if err != nil {
   switch {
   case errors.Is(err, ErrNoPipelineInWorkspace):
       ws.logger("Unable to read pipeline manifest: ...")
   default:
       return nil, err
   }
} else {
   pipelineManifests = append(...)
}
```

internal/pkg/workspace/workspace.go
```diff
@@ -211,15 +217,16 @@ func (ws *Workspace) ListPipelines() ([]Pipeline, error) {
 			path := filepath.Join(pipelinesPath, file.Name())
 			manifest, err := ws.ReadPipelineManifest(path)
 			if err != nil {
-				log.Infof("Unable to read pipeline manifest at '%s'", path)
+				ws.logger("Unable to read pipeline manifest at '%s'", path)
```
efekarakus: nit: should we include the err here:
```suggestion
				ws.logger("Unable to read pipeline manifest at '%s': %v", path, err)
```

internal/pkg/workspace/workspace.go
huanjani: 🤦🏼

efekarakus (approved):

iamhopaul123 (approved):

internal/pkg/term/selector/selector.go
```diff
@@ -812,6 +862,128 @@ func (s *Select) Application(msg, help string, additionalOpts ...string) (string
...
 
+// Dockerfile asks the user to select from a list of Dockerfiles in the current
+// directory or one level down. If no dockerfiles are found, it asks for a custom path.
+func (s *WorkspaceSelect) Dockerfile(selPrompt, notFoundPrompt, selHelp, notFoundHelp string, pathValidator prompt.ValidatorFunc) (string, error) {
+	dockerfiles, err := s.ws.ListDockerfiles()
+	if err != nil {
+		return "", fmt.Errorf("list Dockerfiles: %w", err)
+	}
+	var sel string
+	dockerfiles = append(dockerfiles, []string{dockerfilePromptUseCustom, DockerfilePromptUseImage}...)
```
nodir-t, line 873:
```
dockerfiles = append(dockerfiles, dockerfilePromtUseCustom, DockerfilePromotUseImage)
```

no need to allocate a slice

internal/pkg/term/selector/selector.go
```diff
@@ -812,6 +862,128 @@ func (s *Select) Application(msg, help string, additionalOpts ...string) (string
...
+	sel, err = s.prompt.SelectOne(
```
nodir-t, line 874: nit: `sel, err := ...`
no need for `var sel string` above, right?

internal/pkg/term/selector/selector.go
```diff
@@ -673,6 +693,36 @@ func filterWlsByName(wls []*config.Workload, wantedNames []string) []string {
...
+// Pipeline fetches all the pipelines in a workspace and prompts the user to select one.
+func (s *PipelineSelect) Pipeline(msg, help string) (*workspace.PipelineManifest, error) {
+	pipelines, err := s.ws.ListPipelines()
+	if err != nil {
+		return nil, err
+	}
+	if len(pipelines) == 0 {
+		return nil, errors.New("no pipelines found")
+	}
+	var pipelineNames []string
```
nodir-t, line 705: noticed this pattern a few times in this CR. The `append` below might make multiple allocations despite we know the exact size from the start: it is exactly `len(pipelines)`, so in this and similar cases I suggest to preallocate the slice: `pipelineNames := make([]string, len(pipelines))`, and in the loop do `pipelineNames[i] = pipeline.Name`.

PR #3268, opened 2022-02-10
feat(cli): allow workload package to upload resources

iamhopaul123:
This change adds a flag to workload package command, allowing users to upload necessary resources so that the CFN template/param can be used for real deployment.


huanjani (approved):
lgtm!

internal/pkg/cli/flag.go
```diff
@@ -200,8 +201,9 @@ const (
 	imageTagFlagDescription     = `Optional. The container image tag.`
 	resourceTagsFlagDescription = `Optional. Labels with a key and value separated by commas.
 Allows you to categorize resources.`
-	stackOutputDirFlagDescription = "Optional. Writes the stack template and template configuration to a directory."
-	prodEnvFlagDescription        = "If the environment contains production services."
+	stackOutputDirFlagDescription  = "Optional. Writes the stack template and template configuration to a directory."
+	uploadResourcesFlagDescription = "Optional. Whether to upload resources that are required for deployment."
```
Lou1415926: If we make `--upload-resources` customer-facing, we might want to explain a little why&when they may want to use the flag, and maybe some example artifacts that are being uploaded. Otherwise, it may not be easy to understand what this flag is without looking at the source code.

For example,
```suggestion
	uploadResourcesFlagDescription = "Optional. Whether to upload resources (e.g. a lambda script that automates alias validation for your service) that are required for deployment. The output of `copilot svc package` cannot be directly used for deployment unless these resources are uploaded."
```
We can also consider adding these explanations in our doc, and just include a link to the doc here for conciseness.

internal/pkg/cli/svc_package.go
```diff
@@ -296,10 +298,23 @@ func (o *packageSvcOpts) getSvcTemplates(env *config.Environment) (*wkldCfnTempl
 	if err != nil {
 		return nil, err
 	}
+	uploadOut := clideploy.UploadArtifactsOutput{
+		ImageDigest: aws.String(""),
+	}
+	if o.uploadResources {
+		out, err := generator.UploadArtifacts()
+		if err != nil {
+			return nil, fmt.Errorf("upload deploy resources for %s: %w", o.name, err)
```
Lou1415926: nit:
```suggestion
			return nil, fmt.Errorf("upload resources required for deployment for %s: %w", o.name, err)
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -394,14 +394,14 @@ func NewWorkerSvcDeployer(in *WorkloadDeployerInput) (*workerSvcDeployer, error)
 
 // UploadArtifactsOutput is the output of UploadArtifacts.
 type UploadArtifactsOutput struct {
-	ImageDigest string
+	ImageDigest *string
```
Lou1415926, line 397: 🖐🏼  Question boss! Why pointer?

iamhopaul123: Because since we want to keep the behavior to avoid backward incompatible changes, we need `ImageDigest` to be "" so that it would know this image needs to be build and push (it's just the upload artifact flag is not set), whereas `ImageDigest` to be `nil` means it doesn't need to.

The change of `UploadArtifactsOutput` is for the ease of using (also aligns with our intention that "" is a valid zero value for `UploadArtifactsOutput`).

internal/pkg/cli/flag.go
```diff
@@ -200,8 +201,9 @@ const (
 	imageTagFlagDescription     = `Optional. The container image tag.`
 	resourceTagsFlagDescription = `Optional. Labels with a key and value separated by commas.
 Allows you to categorize resources.`
-	stackOutputDirFlagDescription = "Optional. Writes the stack template and template configuration to a directory."
-	prodEnvFlagDescription        = "If the environment contains production services."
+	stackOutputDirFlagDescription  = "Optional. Writes the stack template and template configuration to a directory."
+	uploadResourcesFlagDescription = "Optional. Whether to upload dependency resources (e.g., image that needs to build and push), so that the output of `copilot svc package` can be directly used for deployment."
```
Lou1415926:
```suggestion
	uploadResourcesFlagDescription = "Optional. Whether to upload dependency resources (e.g., an image that needs to be built and pushed), so that the output of `copilot svc package` can be directly used for deployment."
```

site/content/docs/commands/svc-package.en.md
```diff
@@ -10,11 +10,13 @@ $ copilot svc package
 ## What are the flags?
 
 ```bash
+  -a, --app string          Name of the application.
   -e, --env string          Name of the environment.
   -h, --help                help for package
   -n, --name string         Name of the service.
       --output-dir string   Optional. Writes the stack template and template configuration to a directory.
       --tag string          Optional. The service's image tag.
+      --upload-resources    Optional. Whether to upload dependency resources (e.g., image that needs to build and push), so that the output of `copilot svc package` can be directly used for deployment.
```
Lou1415926:
```suggestion
      --upload-resources    Optional. Whether to upload dependency resources (e.g., an image that needs to be built and pushed), so that the output of `copilot svc package` can be directly used for deployment.
```

site/content/docs/commands/job-package.en.md
```diff
@@ -16,20 +16,21 @@ $ copilot job package
   -n, --name string         Name of the job.
       --output-dir string   Optional. Writes the stack template and template configuration to a directory.
       --tag string          Optional. The container image tag.
+      --upload-resources    Optional. Whether to upload dependency resources (e.g., image that needs to build and push), so that the output of `copilot svc package` can be directly used for deployment.
```
Lou1415926:
```suggestion
      --upload-resources    Optional. Whether to upload dependency resources (e.g., an image that needs to be built and pushed), so that the output of `copilot svc package` can be directly used for deployment.
```

internal/pkg/cli/deploy/deploy.go
Lou1415926: oohh do you mean it's possible that:
1. image has been built and pushed
2. imageDigest is ""

to happen at the same time? I thought `imageDigest` is empty if and only if no image is built

iamhopaul123: if `imageDigest` is empty then we keep using the old URL without pushing the image (example: "1234567890.dkr.ecr.us-west-1.amazonaws.com/demo/frontend:latest") it was actually a bad behavior because it won't work and the url would be different if we are not using the latest tag.

if it is not empty we build and push the image and use the correct URL. 

if it is nil then we do nothing.

iamhopaul123: Comment added so that future maintainer won't remove it accidentally

Lou1415926 (approved):

PR #3269, opened 2022-02-11
fix: mark subnets with implicit igw association as public 

### LINKED ISSUE #3266: Subnets implicitly associated with an IGW are not detected as public
## Problem Summary

As of [copilot v1.15](https://github.com/aws/copilot-cli/releases/tag/v1.15.0),   
When running  `copilot env init` and choosing to import a VPC where there are subnets with an _implicit_ association to an internet gateway,  
Then Copilot fails to detect them as public subnets.

### Steps for reproduction

1. Run `copilot env init`
2. Choose the option "No, I'd like to import existing resources"
3. Choose the default VPC
You should get the following log statements:
```
Note: No existing subnets were found in VPC vpc-601c5218.
If you proceed without at least two public subnets, you will not be able to deploy Load Balanced Web Services in this environment.
```
However, instead we should have listed all the subnets in the VPC because they are by default public.

### Detailed description

When describing route tables of a VPC (`describe-route-tables`), we can tell if there is a route to the internet by checking if there is an internet gateway:
```json
"Routes": [ {
    "DestinationCidrBlock": "0.0.0.0/0",
    "GatewayId": "igw-0c0efaaeef43f8992",
    "Origin": "CreateRoute",
    "State": "active"
}]
```
If there is an `Association` with a `SubnetID` to the route table, we'll say that it's **explicitly** a public subnet.
```json
"Associations": [{
    "Main": false,
    "RouteTableAssociationId": "rtbassoc-0e8324d233a3d9825",
    "RouteTableId": "rtb-00ff9899d574dda1d",
    "SubnetId": "subnet-00046bd87aa3f459f",
    "AssociationState": {
       "State": "associated"
     }
}]
```

⚠️ However, if the subnet does not have any **explicit** associations with a route table, then we should be checking the ["main" route table](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Route_Tables.html#main-col-body) has a route to an internet gateway.
If it does, then that means that the subnet has a way of routing requests to the internet and hence it should be considered a public subnet. We'll say these subnets are **implicitly** public subnets.

```json
{
    "RouteTables": [
        {
            "Associations": [
                {
                    "Main": true,  // Main route table.
                    "RouteTableAssociationId": "rtbassoc-942d24e9",
                    "RouteTableId": "rtb-adb890d6",
                    "AssociationState": {
                        "State": "associated"
                    }
                }
            ],
            "RouteTableId": "rtb-adb890d6",
            "Routes": [
                {
                    "DestinationCidrBlock": "172.31.0.0/16",
                    "GatewayId": "local",
                    "Origin": "CreateRouteTable",
                    "State": "active"
                },
                {
                    "DestinationCidrBlock": "0.0.0.0/0",
                    "GatewayId": "igw-3542f24c",  // Has a route to an internet gateway
                    "Origin": "CreateRoute",
                    "State": "active"
                }
            ],
        }
    ]
}
```

efekarakus:
Fixes https://github.com/aws/copilot-cli/issues/3266


internal/pkg/aws/ec2/ec2.go
```diff
@@ -377,3 +364,75 @@ func toEC2Filter(filters []Filter) []*ec2.Filter {
...
+		}
+
+		if table.IsMain() {
+			index.mainTable = table
+		}
+	}
+	return index
+}
+
+// IsPublicSubnet returns true if the subnet has a route to an internet gateway regardless if the
```
huanjani: Nit: maybe we should explain in the comments what "explicit" and "implicit" mean in this context.

efekarakus: Done!

internal/pkg/aws/ec2/ec2.go
```diff
@@ -377,3 +364,76 @@ func toEC2Filter(filters []Filter) []*ec2.Filter {
...
+	// Explicit route table association for a subnet. A subnet can only be associated to one route table.
+	routeTableForSubnet map[string]*routeTable
+}
+
+func indexRouteTables(tables []*ec2.RouteTable) *routeTableIndex {
+	index := &routeTableIndex{
+		routeTableForSubnet: make(map[string]*routeTable),
+	}
+	for _, table := range tables { // Index all properties in a single pass.
+		table := (*routeTable)(table)
```
iamhopaul123, line 417: do we need to check the type in case the type can't be asserted?

internal/pkg/aws/ec2/ec2.go
```diff
@@ -377,3 +364,76 @@ func toEC2Filter(filters []Filter) []*ec2.Filter {
...
+	for _, route := range rt.Routes {
+		if strings.HasPrefix(aws.StringValue(route.GatewayId), internetGatewayIDPrefix) {
+			return true
+		}
+	}
+	return false
+}
+
+// AssociatedSubnets returns the list of subnet IDs associated with the route table.
+func (rt *routeTable) AssociatedSubnets() []string {
```
iamhopaul123, line 392: Does this function have to be public? same for the other new added public functions.

internal/pkg/aws/ec2/ec2.go
efekarakus: no need boss it's a compile type check

internal/pkg/aws/ec2/ec2.go
efekarakus: The type itself is private so I don't think it particularly matters since it doesn't get exposed by the package 🤷 do you know if there is a particular benefit? 
I marked it as public to indicate that they can be used by clients, and i would mark it as private if only `*routeTable` should use it

internal/pkg/aws/ec2/ec2.go
```diff
@@ -377,3 +364,76 @@ func toEC2Filter(filters []Filter) []*ec2.Filter {
...
 	return ec2Filter
 }
+
+type routeTable ec2.RouteTable
+
+// IsMain returns true if the route table is the default route table for the VPC.
+// If a subnet is not associated with a particular route table, then it will default to the main route table.
+func (rt *routeTable) IsMain() bool {
+	for _, association := range rt.Associations {
+		if aws.BoolValue(association.Main) {
```
Lou1415926, line 374: hmm I wonder why `Main` is an attribute of `association` but not `routeTable`. Sounds to me the latter makes much more sense 😅

Lou1415926 (approved):

internal/pkg/aws/ec2/ec2.go
efekarakus: That's a great point 😅 I have no idea why it's like that

iamhopaul123 (approved):

PR #3283, opened 2022-02-15
fix: simplify and fix buildspec and pipeline CFN template

iamhopaul123:
This PR simplifies pipeline buildspec and fixes two bugs:

1. Some parts of the buildspec file generated by `pipeline init` are always static (for example the S3 bucket mentioned above) which should be dynamic. The reason behind that is: if the user creates another environment in a new region, the pipeline won’t work for the new environment anymore because we won’t upload those prerequisite to the new S3 bucket.
2. Pipeline CFN template grabs all the local workloads by checking their local files, which is problematic if the workload has been deleted before.

### DIFF
--- a/internal/pkg/cli/flag.go
+++ b/internal/pkg/cli/flag.go
@@ -202,7 +202,7 @@ const (
 	resourceTagsFlagDescription = `Optional. Labels with a key and value separated by commas.
 Allows you to categorize resources.`
 	stackOutputDirFlagDescription  = "Optional. Writes the stack template and template configuration to a directory."
-	uploadResourcesFlagDescription = "Optional. Whether to upload dependency resources (e.g., an image that needs to be built and pushed), so that the output of `copilot svc package` can be directly used for deployment."
+	uploadResourcesFlagDescription = "Optional. Whether to upload dependency resources (e.g., an image that needs to be built and pushed), so that the output can be directly used for deployment."
 	prodEnvFlagDescription         = "If the environment contains production services."
 
 	limitFlagDescription = `Optional. The maximum number of log events returned. Default is 10
--- a/internal/pkg/cli/flag.go
+++ b/internal/pkg/cli/flag.go
@@ -32,7 +32,7 @@ const (
 	imageTagFlag          = "tag"
 	resourceTagsFlag      = "resource-tags"
 	stackOutputDirFlag    = "output-dir"
-	uploadResourcesFlag   = "upload-resources"
+	uploadAssetsFlag      = "upload-resources"
 	limitFlag             = "limit"
 	followFlag            = "follow"
 	sinceFlag             = "since"
@@ -201,9 +201,10 @@ const (
 	imageTagFlagDescription     = `Optional. The container image tag.`
 	resourceTagsFlagDescription = `Optional. Labels with a key and value separated by commas.
 Allows you to categorize resources.`
-	stackOutputDirFlagDescription  = "Optional. Writes the stack template and template configuration to a directory."
-	uploadResourcesFlagDescription = "Optional. Whether to upload dependency resources (e.g., an image that needs to be built and pushed), so that the output can be directly used for deployment."
-	prodEnvFlagDescription         = "If the environment contains production services."
+	stackOutputDirFlagDescription = "Optional. Writes the stack template and template configuration to a directory."
+	uploadAssetsFlagDescription   = `Optional. Whether to upload assets (container images, Lambda functions).
+Uploaded asset locations are filled in the template configuration.`
+	prodEnvFlagDescription = "If the environment contains production services."
 
 	limitFlagDescription = `Optional. The maximum number of log events returned. Default is 10
 unless any time filtering flags are set.`
--- a/internal/pkg/cli/flag.go
+++ b/internal/pkg/cli/flag.go
@@ -32,7 +32,7 @@ const (
 	imageTagFlag          = "tag"
 	resourceTagsFlag      = "resource-tags"
 	stackOutputDirFlag    = "output-dir"
-	uploadAssetsFlag      = "upload-resources"
+	uploadAssetsFlag      = "upload-assets"
 	limitFlag             = "limit"
 	followFlag            = "follow"
 	sinceFlag             = "since"
@@ -202,7 +202,7 @@ const (
 	resourceTagsFlagDescription = `Optional. Labels with a key and value separated by commas.
 Allows you to categorize resources.`
 	stackOutputDirFlagDescription = "Optional. Writes the stack template and template configuration to a directory."
-	uploadAssetsFlagDescription   = `Optional. Whether to upload assets (container images, Lambda functions).
+	uploadAssetsFlagDescription   = `Optional. Whether to upload assets (container images, Lambda functions, etc.).
 Uploaded asset locations are filled in the template configuration.`
 	prodEnvFlagDescription = "If the environment contains production services."
 
--- a/internal/pkg/cli/job_list.go
+++ b/internal/pkg/cli/job_list.go
@@ -80,11 +80,6 @@ func (o *listJobOpts) Execute() error {
 	return nil
 }
 
-// RecommendActions is a no-op for this command.
-func (o *listJobOpts) RecommendActions() error {
-	return nil
-}
-
 func buildJobListCmd() *cobra.Command {
 	vars := listWkldVars{}
 	cmd := &cobra.Command{
--- a/internal/pkg/cli/job_list.go
+++ b/internal/pkg/cli/job_list.go
@@ -41,7 +41,7 @@ func newListJobOpts(vars listWkldVars) (*listJobOpts, error) {
 		Store: store,
 		Out:   os.Stdout,
 
-		ShowLocalJobs: vars.shouldOutputJSON,
+		ShowLocalJobs: vars.shouldShowLocalWorkloads,
 		OutputJSON:    vars.shouldOutputJSON,
 	}
 
@@ -53,6 +53,12 @@ func newListJobOpts(vars listWkldVars) (*listJobOpts, error) {
 	}, nil
 }
 
+// Validate is a no-op for this command.
+func (o *listJobOpts) Validate() error {
+	return nil
+}
+
+// Ask asks for fields that are required but not passed in.
 func (o *listJobOpts) Ask() error {
 	if o.appName != "" {
 		return nil
@@ -71,6 +77,11 @@ func (o *listJobOpts) Execute() error {
 	if err := o.list.Write(o.appName); err != nil {
 		return err
 	}
+	return nil
+}
+
+// RecommendActions is a no-op for this command.
+func (o *listJobOpts) RecommendActions() error {
 	return nil
 }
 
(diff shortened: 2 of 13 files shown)


internal/pkg/template/templates/cicd/buildspec.yml
```diff
@@ -56,128 +56,21 @@ phases:
...
           if [ $? -ne 0 ]; then
             echo "Cloudformation stack and config files were not generated. Please check build logs to see if there was a manifest validation error." 1>&2;
             exit 1;
           fi
           done;
         done;
       - ls -lah ./infrastructure
-      # Concatenate jobs and services into one var for addons and env file
-      - WORKLOADS=$(echo $jobs $svcs)
-      # If addons exists, upload addons templates to each S3 bucket and write template URL to template config files.
```
Lou1415926, line 76: 👋🏼  Farewell!

internal/pkg/cli/job_list.go
```diff
@@ -41,7 +41,7 @@ func newListJobOpts(vars listWkldVars) (*listJobOpts, error) {
 		Store: store,
 		Out:   os.Stdout,
 
-		ShowLocalJobs: vars.shouldOutputJSON,
```
Lou1415926, line 44: 😰

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -179,28 +226,27 @@ func (o *deployPipelineOpts) Execute() error {
...
 	}
 	pipeline, err := manifest.UnmarshalPipeline(data)
 	if err != nil {
-		return fmt.Errorf("unmarshal pipeline manifest: %w", err)
+		return nil, fmt.Errorf("unmarshal pipeline manifest: %w", err)
 	}
-	if pipeline.Name != o.name {
-		return fmt.Errorf(`pipeline %s not found in the workspace`, color.HighlightUserInput(o.name))
+	if err := pipeline.Validate(); err != nil {
+		return nil, fmt.Errorf("validate pipeline: %w", err)
```
Lou1415926: nit:
```suggestion
		return nil, fmt.Errorf("validate pipeline manifest: %w", err)
```

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -179,28 +226,27 @@ func (o *deployPipelineOpts) Execute() error {
...
 	}
-	return nil
+	return pipeline, nil
```
Lou1415926, line 246: Should we cache the manifest?

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -179,28 +226,27 @@ func (o *deployPipelineOpts) Execute() error {
...
 }
 
 func (o *deployPipelineOpts) convertStages(manifestStages []manifest.PipelineStage) ([]deploy.PipelineStage, error) {
 	var stages []deploy.PipelineStage
-	workloads, err := o.ws.ListWorkloads()
```
Lou1415926, line 199: We probably need better function name for these various workload-listing functions...
Like from what I know there are ones that list ssm-registered workload, ones that that local-workspace workloads (that are not necessarily ssm-registered), and ones that list workloads that are ssm-registered and in the local workspace. It's easy to get these functions mixed up because they have the same or similar names.

internal/pkg/cli/pipeline_deploy_test.go
```diff
@@ -164,7 +167,7 @@ stages:
 					m.prog.EXPECT().Stop(log.Ssuccessf(fmtPipelineDeployResourcesComplete, appName)).Times(1),
 
 					m.ws.EXPECT().ReadPipelineManifest().Return([]byte(content), nil),
-					m.ws.EXPECT().ListWorkloads().Return([]string{"frontend", "backend"}, nil).Times(1),
+					m.actionCmd.EXPECT().Execute().Times(2),
```
Lou1415926, line 170: Once for service listing and once for job listing right?

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -95,19 +106,62 @@ func newDeployPipelineOpts(vars deployPipelineVars) (*deployPipelineOpts, error)
 		prog:               termprogress.NewSpinner(log.DiagnosticWriter),
 		prompt:             prompt.New(),
 		codestar:           cs.New(defaultSession),
+		newSvcListCmd: func(w io.Writer) actionCommand {
```
Lou1415926: Instead of `actionCommand`, what about [`cmd`](https://github.com/aws/copilot-cli/blob/d321374c655428b3665d46f5f562ea08caff5b0c/internal/pkg/cli/interfaces.go#L41)?

internal/pkg/cli/pipeline_deploy.go
Lou1415926: Also do you think this is rather a `fix` than a `chore`? Would this introduce breaking change?

iamhopaul123: i'll update it to `fix` as this is very obviously a bug fix, we don't consider it a breaking change

internal/pkg/cli/pipeline_deploy_test.go
iamhopaul123: yeah

internal/pkg/cli/flag.go
```diff
@@ -202,7 +202,7 @@ const (
 	resourceTagsFlagDescription = `Optional. Labels with a key and value separated by commas.
 Allows you to categorize resources.`
 	stackOutputDirFlagDescription  = "Optional. Writes the stack template and template configuration to a directory."
-	uploadResourcesFlagDescription = "Optional. Whether to upload dependency resources (e.g., an image that needs to be built and pushed), so that the output of `copilot svc package` can be directly used for deployment."
+	uploadResourcesFlagDescription = "Optional. Whether to upload dependency resources (e.g., an image that needs to be built and pushed), so that the output can be directly used for deployment."
```
efekarakus: Should we rename the flag to `uploadAssets`? to use the same terminology as CDK: https://docs.aws.amazon.com/cdk/v2/guide/assets.html

```suggestion
	uploadResourcesFlagDescription = `Optional. Whether to upload assets (container images, Lambda functions).
Uploaded asset locations are filled in the template configuration.`
```

internal/pkg/cli/pipeline_deploy.go
iamhopaul123: Added comment to `ListWorkloads` hopefully it could help with ambiguity for a little bit

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -60,8 +65,14 @@ type deployPipelineOpts struct {
...
 	ws               wsPipelineReader
 	codestar         codestar
+	newSvcListCmd    func(io.Writer) actionCommand
+	newJobListCmd    func(io.Writer) actionCommand
 
 	shouldPromptUpdateConnection bool
+
+	// cache variables
+	svcBuffer *bytes.Buffer
+	jobBuffer *bytes.Buffer
```
efekarakus: nit: can we make these `io.Writer` as well?

internal/pkg/template/templates/cicd/buildspec.yml
efekarakus: i love this boss, i think this is a good milestone too. Uploading assets is the most common thing that we need to change in the buildspec, from now on every time we have to touch the buildspec we can evaluate how to abstract it away

site/content/docs/commands/job-package.en.md
```diff
@@ -16,7 +16,7 @@ $ copilot job package
   -n, --name string         Name of the job.
       --output-dir string   Optional. Writes the stack template and template configuration to a directory.
       --tag string          Optional. The container image tag.
-      --upload-resources    Optional. Whether to upload dependency resources (e.g., an image that needs to be built and pushed), so that the output of `copilot svc package` can be directly used for deployment.
+      --upload-resources    Optional. Whether to upload dependency resources (e.g., an image that needs to be built and pushed), so that the output can be directly used for deployment.
```
efekarakus: (same feedback here for docs `--upload-assets`)

internal/pkg/cli/pipeline_deploy.go
iamhopaul123: Hmm i'm more inclined to keep as is because it is supposed to be `*bytes.Buffer` and we call `Byte()` when we unmarshal the json (we could use Encoder but we'll need to initiate and then call encode).

internal/pkg/cli/flag.go
```diff
@@ -32,7 +32,7 @@ const (
 	imageTagFlag          = "tag"
 	resourceTagsFlag      = "resource-tags"
 	stackOutputDirFlag    = "output-dir"
-	uploadResourcesFlag   = "upload-resources"
+	uploadAssetsFlag      = "upload-resources"
```
efekarakus:
```suggestion
	uploadAssetsFlag      = "upload-assets"
```

site/content/docs/commands/job-package.en.md
```diff
@@ -16,7 +16,8 @@ $ copilot job package
   -n, --name string         Name of the job.
       --output-dir string   Optional. Writes the stack template and template configuration to a directory.
       --tag string          Optional. The container image tag.
-      --upload-resources    Optional. Whether to upload dependency resources (e.g., an image that needs to be built and pushed), so that the output can be directly used for deployment.
+      --upload-resources    Optional. Whether to upload assets (container images, Lambda functions).
```
efekarakus:
```suggestion
      --upload-assets    Optional. Whether to upload assets (container images, Lambda functions).
```

site/content/docs/commands/svc-package.en.md
```diff
@@ -16,7 +16,8 @@ $ copilot svc package
   -n, --name string         Name of the service.
       --output-dir string   Optional. Writes the stack template and template configuration to a directory.
       --tag string          Optional. The service's image tag.
-      --upload-resources    Optional. Whether to upload dependency resources (e.g., an image that needs to be built and pushed), so that the output can be directly used for deployment.
+      --upload-resources    Optional. Whether to upload assets (container images, Lambda functions).
```
efekarakus:
```suggestion
      --upload-assets    Optional. Whether to upload assets (container images, Lambda functions).
```

internal/pkg/cli/flag.go
```diff
@@ -201,9 +201,10 @@ const (
 	imageTagFlagDescription     = `Optional. The container image tag.`
 	resourceTagsFlagDescription = `Optional. Labels with a key and value separated by commas.
 Allows you to categorize resources.`
-	stackOutputDirFlagDescription  = "Optional. Writes the stack template and template configuration to a directory."
-	uploadResourcesFlagDescription = "Optional. Whether to upload dependency resources (e.g., an image that needs to be built and pushed), so that the output of `copilot svc package` can be directly used for deployment."
-	prodEnvFlagDescription         = "If the environment contains production services."
+	stackOutputDirFlagDescription = "Optional. Writes the stack template and template configuration to a directory."
+	uploadAssetsFlagDescription   = `Optional. Whether to upload assets (container images, Lambda functions).
```
Lou1415926: nit:
```suggestion
	uploadAssetsFlagDescription   = `Optional. Whether to upload assets (container images and Lambda functions).
```
or
```suggestion
	uploadAssetsFlagDescription   = `Optional. Whether to upload assets (container images, Lambda functions, etc.).
```

Lou1415926 (approved):

efekarakus (approved):

PR #3284, opened 2022-02-15
chore(ws): exclude dockerignore files from Dockerfile list

### LINKED ISSUE #3274: Copilot includes .dockerignore file when suggesting docker files
[Docker 19.03 shipped the ability to have per-dockerfile `.dockerignore`s](https://stackoverflow.com/questions/40904409/how-to-specify-different-dockerignore-files-for-different-builds-in-the-same-pr) by searching for `<dockerfile-name>.dockerignore` before falling back to `.dockerignore`.

It seems that copilot is looking for files starting with `Dockerfile`, meaning that `Dockerfile.dockerignore` will be included in the suggestions when prompting for which dockerfile to use: 

```
Application name: ras-example-ecs
Workload type: Load Balanced Web Service
Service name: website

  Which Dockerfile would you like to use for website?  [Use arrows to move, type to filter, ? for more help]
    ./Dockerfile
  > ./Dockerfile.prod.dockerfile
    ./Dockerfile.prod.dockerfile.dockerignore
    Enter custom path for your Dockerfile
    Use an existing image instead
```

While there's a bunch of ways this could come up (since there isn't an official dockerfile extension), I think it would make sense to exclude files ending with `.dockerignore` under the assumption that they're an ignore file for a Dockerfile regardless of their name.

huanjani:
Resolves #3274.

### DIFF
--- a/internal/pkg/workspace/workspace.go
+++ b/internal/pkg/workspace/workspace.go
@@ -458,7 +458,7 @@ func (ws *Workspace) ListDockerfiles() ([]string, error) {
 		// Add current file if it is a Dockerfile and not a directory; otherwise continue.
 		if !wdFile.IsDir() {
 			fname := wdFile.Name()
-			if strings.Contains(strings.ToLower(fname), dockerfileName) && !strings.Contains(strings.ToLower(fname), dockerignoreName) {
+			if strings.Contains(strings.ToLower(fname), dockerfileName) && !strings.HasSuffix(strings.ToLower(fname), dockerignoreName) {
 				path := filepath.Dir(fname) + "/" + fname
 				dockerfiles = append(dockerfiles, path)
 			}
--- a/internal/pkg/workspace/workspace.go
+++ b/internal/pkg/workspace/workspace.go
@@ -477,7 +477,7 @@ func (ws *Workspace) ListDockerfiles() ([]string, error) {
 				continue
 			}
 			fname := f.Name()
-			if strings.Contains(strings.ToLower(fname), dockerfileName) && !strings.Contains(strings.ToLower(fname), dockerignoreName) {
+			if strings.Contains(strings.ToLower(fname), dockerfileName) && !strings.HasSuffix(strings.ToLower(fname), dockerignoreName) {
 				path := wdFile.Name() + "/" + f.Name()
 				dockerfiles = append(dockerfiles, path)
 			}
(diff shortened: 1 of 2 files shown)


internal/pkg/workspace/workspace.go
```diff
@@ -457,7 +458,7 @@ func (ws *Workspace) ListDockerfiles() ([]string, error) {
 		// Add current file if it is a Dockerfile and not a directory; otherwise continue.
 		if !wdFile.IsDir() {
 			fname := wdFile.Name()
-			if strings.Contains(strings.ToLower(fname), dockerfileName) {
+			if strings.Contains(strings.ToLower(fname), dockerfileName) && !strings.Contains(strings.ToLower(fname), dockerignoreName) {
```
efekarakus: nit: just to be a bit more robust:
```suggestion
			if strings.Contains(strings.ToLower(fname), dockerfileName) && !strings.HasSuffix(strings.ToLower(fname), dockerignoreName) {
```

internal/pkg/workspace/workspace.go
```diff
@@ -43,6 +43,7 @@ const (
 	ymlFileExtension = ".yml"
 
 	dockerfileName = "dockerfile"
+	dockerignoreName = "dockerignore"
```
efekarakus:
```suggestion
	dockerignoreName = ".dockerignore"
```

efekarakus (approved):
:shipit: after change, thanks for resolving this issue!

internal/pkg/workspace/workspace.go
```diff
@@ -476,7 +477,7 @@ func (ws *Workspace) ListDockerfiles() ([]string, error) {
 				continue
 			}
 			fname := f.Name()
-			if strings.Contains(strings.ToLower(fname), dockerfileName) {
+			if strings.Contains(strings.ToLower(fname), dockerfileName) && !strings.Contains(strings.ToLower(fname), dockerignoreName) {
```
Lou1415926: Do we need to change it here as well?
```suggestion
			if strings.Contains(strings.ToLower(fname), dockerfileName) && !strings.HasSuffix(strings.ToLower(fname), dockerignoreName) {
```

Lou1415926 (approved):
Thank you for fixing this! Added the label just in case you'd like to address the nit feedback!

PR #3285, opened 2022-02-15
chore(env): enable importing VPCs with only public (no private) subnets

### LINKED ISSUE #3282: feat(env): allow VPCs with no private subnets
Just as we allow users to use existing VPCs that don't have public subnets, we should allow the same with private subnets. We currently log a warning re: the limitations of no public subnets (load balancing), and should do the same for the limitations of no private subnets (future use of internal load balancers).

This change enables users to use default VPCs, as they come with default public subnets.

huanjani:
Currently, users have the option to create environments with imported VPCs that have only private subnets. They are warned that load balancing will not work, but are allowed to proceed. 

These changes enable a similar warn-but-allow workflow. The benefit is increased flexibility; for one, folks new to AWS can more easily try out Copilot with their default VPC. The downside is that in the future, if we add features that require private subnets (like internal load balancers), these environments won't have that option.

Fixes #3282.

### DIFF
--- a/internal/pkg/cli/env_init.go
+++ b/internal/pkg/cli/env_init.go
@@ -532,15 +532,16 @@ If you proceed without at least two public subnets, you will not be able to depl
 			IsPublic: false,
 		})
 		if err != nil {
-			if err == selector.ErrSubnetsNotFound {
-				log.Errorf(`No existing private subnets were found in VPC %s. You can either:
-- Create new private subnets and then import them.
-- Use the default Copilot environment configuration.
+			if errors.Is(err, selector.ErrSubnetsNotFound) {
+				log.Warningf(`No existing private subnets were found in VPC %s.
+If you proceed without at least two private subnets, you may not be able to access 
+all Copilot features with this environment in the future.
 `, o.importVPC.ID)
+			} else {
+				return fmt.Errorf("select private subnets: %w", err)
 			}
-			return fmt.Errorf("select private subnets: %w", err)
 		}
-		if len(privateSubnets) < 2 {
+		if len(privateSubnets) == 1 {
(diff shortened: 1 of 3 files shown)


internal/pkg/cli/env_init.go
```diff
@@ -532,15 +532,16 @@ If you proceed without at least two public subnets, you will not be able to depl
...
 		})
 		if err != nil {
-			if err == selector.ErrSubnetsNotFound {
-				log.Errorf(`No existing private subnets were found in VPC %s. You can either:
-- Create new private subnets and then import them.
-- Use the default Copilot environment configuration.
+			if errors.Is(err, selector.ErrSubnetsNotFound) {
+				log.Warningf(`No existing private subnets were found in VPC %s.
+If you proceed without at least two private subnets, you may not be able to access 
+all Copilot features with this environment in the future.
```
efekarakus: nit: maybe this to be more specific?
```suggestion
				log.Warningf(`No existing private subnets were found in VPC %s.
If you proceed without at least two private subnets, you won't be able 
to place services in private subnets.
```

efekarakus (commented):
Is it possible that we end up in a world where a VPC is imported with no public or private subnets 😂 is that even possible?

internal/pkg/cli/env_init.go
huanjani: 🤔  Seems like stating the obvious, maybe?
Maybe "If you proceed without private subnets, you will not be able to add them after the environment is created" or something like that?

huanjani:
> Is it possible that we end up in a world where a VPC is imported with no public or private subnets 😂 is that even possible?

Added a safety net to catch this just in case.

efekarakus (approved):

iamhopaul123 (approved):

PR #3288, opened 2022-02-16
chore(task): grant task-run execution role access to secrets without tags

paragbhingre:
This PR fixes the issue #2641 

Fixed the issue by adding policies to the execution role when --secrets are provided.

Tests -
Following manual tests have been taken into consideration
1. When we send both `--app` and `--env` flags in copilot task run command then secrets are accessible only when copilot-application and copilot-environment tags are present with the correct values on the secret
2. When we send both `--app` and `--env` flags in copilot task run command then secrets are inaccessible only when copilot-application and copilot-environment tags are NOT present on the secret or they have incorrect values
3. When SSM parameter without ARN has been provided in secrets without`--app` and `--env` flag then acknowledgement permissions will be prompted and on input `yes`, specific SSM parameter secret access permission will be added to the execution role.
4. with `--acknowledge-secrets-access` flag secrets access permission grant question will be skipped.


Lou1415926 (commented):
Just skimmed through and will take a second look later - but what do you think of adding some unit tests? For example, we want to ensure that in `cli` package, questions are prompted/not prompted as expected.

internal/pkg/cli/flag.go
```diff
@@ -156,6 +157,8 @@ Mutually exclusive with -%s, --%s.`, imageFlagShort, imageFlag)
 
 	clusterFlagDescription = fmt.Sprintf(`Optional. The short name or full ARN of the cluster to run the task in. 
 Cannot be specified with '%s', '%s' or '%s'.`, appFlag, envFlag, taskDefaultFlag)
+	acknowledgeSecretsAccessDescription = fmt.Sprintf(`Optional. If provided, does not ask grant permission for secrets. 
+Cannot be specified without '%s.`, acknowledgeSecretsAccessFlag)
```
Lou1415926: 🤔 Why can't it be specified with `acknowledgeSecretsAccessFlag`?

internal/pkg/cli/flag.go
```diff
@@ -156,6 +157,8 @@ Mutually exclusive with -%s, --%s.`, imageFlagShort, imageFlag)
 
 	clusterFlagDescription = fmt.Sprintf(`Optional. The short name or full ARN of the cluster to run the task in. 
 Cannot be specified with '%s', '%s' or '%s'.`, appFlag, envFlag, taskDefaultFlag)
+	acknowledgeSecretsAccessDescription = fmt.Sprintf(`Optional. If provided, does not ask grant permission for secrets.
```
Lou1415926: We have an example description to a similar flag that skips questions: https://github.com/aws/copilot-cli/blob/d321374c655428b3665d46f5f562ea08caff5b0c/internal/pkg/cli/flag.go#L196

I am trying to come up with a better description, perhaps:

```suggestion
	acknowledgeSecretsAccessDescription = fmt.Sprintf(`Optional. Skip the confirmation question and grant access to the secrets specified by `--secrets` flag)
```

I think we want to be clear about:
1. What it does (skips questions)
2. The end result (granting permissions)

internal/pkg/cli/task_run.go
```diff
@@ -79,13 +79,19 @@ Select %s to run the task in your default VPC instead of any existing applicatio
...
+)
+
 type runTaskVars struct {
 	count  int
 	cpu    int
 	memory int
 
 	groupName string
 
+	acknowledgeSecretsAccess bool
```
Lou1415926: super tiny nit: we can move this line just below `secrets map[string]string` for better grouping.

internal/pkg/cli/task_run.go
```diff
@@ -371,6 +379,42 @@ func (o *runTaskOpts) Validate() error {
 		}
 	}
 
+	if len(o.secrets) > 0 {
+		if err := o.validateSecrets(); err != nil {
+			return err
+		}
+	}
```
Lou1415926: I think this is perhaps better suited in `Ask` - we tend not to prompt any questions during `Validate`.

internal/pkg/cli/task_run.go
```diff
@@ -371,6 +379,42 @@ func (o *runTaskOpts) Validate() error {
...
+
+	return nil
+}
+
+func (o *runTaskOpts) validateSecrets() error {
```
Lou1415926: nit: I feel like this is less about "validating" but more about "asking for confirmation", reflecting on function name perhaps we can have:
```suggestion
func (o *runTaskOpts) confirmSecretsAccess() error {
```

internal/pkg/cli/task_run.go
```diff
@@ -371,6 +379,42 @@ func (o *runTaskOpts) Validate() error {
...
+	if o.appName == "" && o.env == "" && !o.acknowledgeSecretsAccess {
```
Lou1415926: We could flip the conditional and return early
```
if o.secrets == nil {
   return nil
}

if o.acknowledgeSecretsAccess {
   return nil
}
...
```
rather than writing the majority of the code inside of a huge if-block. This would improves code readability.

internal/pkg/cli/task_run.go
```diff
@@ -371,6 +379,42 @@ func (o *runTaskOpts) Validate() error {
...
+		fmt.Println("Looks like you're requesting ssm:GetParameters to the following SSM parameters:")
```
Lou1415926: We should always use the `log` package here. 

```suggestion
		log.Infoln("Looks like you're requesting ssm:GetParameters to the following SSM parameters:")
```

`log` wraps the printing with the desired writer to write to. In this particular case, we don't want to print to stdout, but instead stderr.

internal/pkg/cli/task_run.go
```diff
@@ -371,6 +379,42 @@ func (o *runTaskOpts) Validate() error {
...
+
+		for _, value := range o.secrets {
+			if !arn.IsARN(value) || strings.Contains(value, ":ssm:") {
```
Lou1415926: <strike>Why do we skip it if the ssm parameter is given as an ARN?</strike>
Edit: oops sorry misread the OR condition. I guess we can include one sentence here explaining this conditional, e.g. why do we assume a non-ARN value is an SSM parameter?

internal/pkg/cli/task_run.go
```diff
@@ -371,6 +379,42 @@ func (o *runTaskOpts) Validate() error {
...
+
+		fmt.Println("\nand secretsmanager:GetSecretValue to the following Secrets Manager secrets:")
+
+		for _, value := range o.secrets {
+			if strings.Contains(value, ":secretsmanager:") {
+				fmt.Println("* " + value)
+			}
+		}
+
+		v, err := o.prompt.Confirm(taskSecretsPermission, "")
```
Lou1415926: We can provide some help message here to explain 
1. Why we need the permissions
2. The result of confirming on this question.

internal/pkg/cli/task_run.go
```diff
@@ -371,6 +379,42 @@ func (o *runTaskOpts) Validate() error {
...
+		if err != nil {
+			return fmt.Errorf("failed to confirm deployment: %w", err)
```
Lou1415926: nit:
```suggestion
			return fmt.Errorf("prompt to confirm secrets access: %w", err)
```
To wrap an error, it might be better if we include contextual information such as what we are trying to do here (i.e. `"prompt to confirm secrets access"`), instead of `"failed to ..."`. In the majority of the codebase we tend to wrap errors this way. My thought is that `"failed to"` is redundant since error has occurred. What would really help here is some contextual information.

internal/pkg/cli/task_run.go
```diff
@@ -371,6 +379,42 @@ func (o *runTaskOpts) Validate() error {
...
+		}
+
+		if !v {
```
Lou1415926: Perhaps we can use a more descriptive variable name, e.g. `secretsAccessConfirmed`.

internal/pkg/cli/task_run.go
```diff
@@ -371,6 +379,42 @@ func (o *runTaskOpts) Validate() error {
...
+			return fmt.Errorf("can't use secret(s) without granting access to it")
```
Lou1415926: nit:
```suggestion
			return errors.New("access to secrets denied")
```

internal/pkg/cli/task_run.go
```diff
@@ -1007,6 +1052,7 @@ func BuildTaskRunCmd() *cobra.Command {
 	utilityFlags := pflag.NewFlagSet("Utility", pflag.ContinueOnError)
 	utilityFlags.AddFlag(cmd.Flags().Lookup(followFlag))
 	utilityFlags.AddFlag(cmd.Flags().Lookup(generateCommandFlag))
+	utilityFlags.AddFlag(cmd.Flags().Lookup(acknowledgeSecretsAccessFlag))
```
Lou1415926: Should this flag be placed under `"Utility"` or `"Task"`? I am not sure 🤔  Open for discussion!

internal/pkg/template/templates/task/cf.yml
```diff
@@ -122,8 +122,17 @@ Resources:
...
-      {{- end}}
+          - PolicyName: 'PullSecrets'
+            PolicyDocument:
+              Version: '2012-10-17'
+              Statement:
+                - Effect: 'Allow'
+                  Action:
+                    - ssm:GetParameters
+                    - secretsmanager:GetSecretValue
+                  Resource: {{range $name, $valueFrom := .Secrets}}
```
Lou1415926: What about the secretsmanager secrets?

internal/pkg/cli/task_run.go
paragbhingre: On line [382](https://github.com/aws/copilot-cli/blob/524658bf37f5a8b5fb4eeaecfa519b9105605111/internal/pkg/cli/task_run.go#L382) I am checking for `o.secrets > 0` so I guess we don't need to check for `o.secrets` again. 

Also  `o.acknowledgeSecretsAccess` is the only one extra condition that I have appended to `o.appName == "" && o.env == ""`. 
Is that okay? or you think I should separate `o.acknowledgeSecretsAccess` condition as well?

Lou1415926: It's no biggie whether `o.acknowledgeSecretsAccess` is in the same condition statement as app and env! I'm fine with both as long as it's obvious for any future readers what the conditional means.

The more important point is to flip the conditional and return early so that it improves code readability.

Lou1415926: Sorry for adding more on top of this 😛 I think we should also take `--execution-role` into consideration. Like we probably  shouldn't need to prompt the question if `--execution-role` is specified.

internal/pkg/template/templates/task/cf.yml
paragbhingre: .Secrets will have both secrets manager secrets and ssm params.  What is the confusion here?

internal/pkg/cli/flag.go
paragbhingre: oops, I meant to say Cann't use without secrets flag. Secrets flag has to be there right?

internal/pkg/cli/flag.go
paragbhingre: Yeah this sounds right

internal/pkg/template/templates/task/cf.yml
Lou1415926: oh yea my bad. I forgot that when `isARN` returns `false` it implies that it's an SSM parameter. 

Though I think ideally 💭 it'd be great if we can have separate policy statements like the `HasAppAndEnv`. Like one statement for `ssm` and one for `secretsmanager`. What do you think?

internal/pkg/cli/flag.go
Lou1415926: I am not sure if we should add this to the description since we don't really error out or anything if `--acknowledge-secrets-access` is specified while `--secrets` is not.

Instead, we could probably say something like `"This flag is useful only when --secrets is specified"`, or like `"Usage of this flag is only checked when --secrets is specified"`

iamhopaul123:
can we change the title to be `chore`? given that it is more like an enhancement 🤔

internal/pkg/template/templates/task/cf.yml
paragbhingre: I think current logic is pretty clean and less complex. If we decide to separate out the logic for ssm and secrets manager then we will need to add at least one more function to figure out if it is ssm or secrets manager and two for loops -- 1. fetch secrets manager secrets and 2 fetch ssm params in CFN template. We can very well implement that but then complexity of the code will be increased when we can do the same task in the current one liner logic.

internal/pkg/cli/task_run.go
```diff
@@ -79,6 +79,11 @@ Select %s to run the task in your default VPC instead of any existing applicatio
 Select %s to run the task in your default VPC instead of any existing environment.`, color.Emphasize(appEnvOptionNone))
 )
 
+var (
+	taskSecretsPermissionPrompt     = "\nDo you grant permission to the ECS/Fargate agent for these secrets? "
```
iamhopaul123: Do we need the "\n"?

internal/pkg/cli/task_run.go
iamhopaul123: maybe something like this?

```go
if o.executionRole != "" {
  return nil
}
if appName != "" || envName != "" {
  return nil
}
if !o.acknowledgeSecretsAccess {
  // log to prompt them to specify either acknowledge flag or execution role if they want to proceed
  return error
}
// log and ask questions
...
```
I think one benefit for this is it would make it crystal clear in what different situation we don't prompt the confirmation question and we'll have less indented code block

internal/pkg/cli/task_run.go
```diff
@@ -374,6 +382,42 @@ func (o *runTaskOpts) Validate() error {
...
+			//For Systems Manager Parameter Store parameter you can specify it as name or the ARN if it exists in the same Region as the task you are launching.
+			if !arn.IsARN(value) || strings.Contains(value, ":ssm:") {
+				log.Infoln("* " + value)
+			}
+		}
+
+		log.Infoln("\nand secretsmanager:GetSecretValue to the following Secrets Manager secrets:")
+
+		for _, value := range o.secrets {
+			if strings.Contains(value, ":secretsmanager:") {
```
iamhopaul123: Maybe add a comment specifying the reason why we don't need to check if it is an ARN for secret manager. BTW, do we have any validation mechanism for the secret value they pass in via flag?

internal/pkg/template/template_functions.go
```diff
@@ -24,6 +25,9 @@ const (
 func ReplaceDashesFunc(logicalID string) string {
 	return strings.ReplaceAll(logicalID, "-", dashReplacement)
 }
+func IsARN(value string) bool {
```
iamhopaul123: Awesome! Comment for public function pls 🙏

iamhopaul123 (commented):
Yayy this is super awesome! Just have a couple of suggestion but mostly +1 to @Lou1415926

internal/pkg/cli/task_run.go
paragbhingre: I think we should keep it under `"Task"` near secretsFlag

internal/pkg/cli/task_run.go
Lou1415926: @iamhopaul123  For secret validation mechanism, do you mean like validating if they exist and whether they are tagged?

iamhopaul123: not online check but offline to make sure they have to pass a valid ARN for either ssm or secretmanager

iamhopaul123: otherwise what if they provide a valid ARN but it is not either ssm or secret manager?

Lou1415926: I think it's just going to error out eventually after `RunTask`. But ya it'd be neat if we could catch that earlier.

iamhopaul123: yeah since it's just gonna be an offline check we can just do it in `Validate()`

paragbhingre: What is the meaning of online and offline check?

internal/pkg/cli/task_run.go
paragbhingre: After listing out the secrets to which we need permissions it's good to have one "\n" else everything looks too close to each other.

internal/pkg/cli/task_run.go
iamhopaul123: offline means we don't need to make API calls that require internet connect.

internal/pkg/cli/task_run.go
```diff
@@ -374,6 +382,48 @@ func (o *runTaskOpts) Validate() error {
 	return nil
 }
 
+func (o *runTaskOpts) confirmSecretsAccess() error {
+
```
Lou1415926, line 418: nit:
```suggestion
```

internal/pkg/cli/task_run.go
```diff
@@ -374,6 +382,48 @@ func (o *runTaskOpts) Validate() error {
...
+	log.Infoln("Looks like you're requesting ssm:GetParameters to the following SSM parameters:")
+
+	for _, value := range o.secrets {
+		//For Systems Manager Parameter Store parameter you can specify it as name or the ARN if it exists in the same Region as the task you are launching.
+		if !arn.IsARN(value) || strings.Contains(value, ":ssm:") {
+			log.Infoln("* " + value)
+		}
+	}
+
+	log.Infoln("\nand secretsmanager:GetSecretValue to the following Secrets Manager secrets:")
```
Lou1415926: I tried the command and thought it looked fine without the prepended newline! Just an input - it's okay to keep it.

internal/pkg/cli/task_run.go
Lou1415926: nit:
```suggestion
	taskSecretsPermissionPrompt     = "\nDo you grant permission to the ECS/Fargate agent for these secrets?"
``` 
I think `prompt.Confirm` seems to already creates a new line before the prompt. We probably don't need the extra new line. Adding together it's three new lines in a row, one at the end of the last secret (`"* secret"`), one imposed by `prompt.Confirm`, one prepended here.

internal/pkg/cli/task_run.go
```diff
@@ -79,6 +79,11 @@ Select %s to run the task in your default VPC instead of any existing applicatio
 Select %s to run the task in your default VPC instead of any existing environment.`, color.Emphasize(appEnvOptionNone))
 )
 
+var (
+	taskSecretsPermissionPrompt     = "\nDo you grant permission to the ECS/Fargate agent for these secrets? "
+	taskSecretsPermissionPromptHelp = "Copilot needs permission to access secrets. select yes to grant permission."
```
Lou1415926: nit:
```suggestion
	taskSecretsPermissionPromptHelp = "ECS/Fargate agent needs the permissions in order to fetch the secrets and inject them into your container."
```

internal/pkg/cli/task_run_test.go
```diff
@@ -636,6 +638,25 @@ func TestTaskRunOpts_Ask(t *testing.T) {
 
 			wantedError: errors.New("ask for environment: error selecting environment"),
 		},
+		"When secrets are provided without app and env leads to a secret access permission prompt": {
```
Lou1415926, line 644: We can also add one unit test against the `--execution-role` case!

internal/pkg/template/template_functions.go
```diff
@@ -25,6 +26,41 @@ func ReplaceDashesFunc(logicalID string) string {
...
 }
 
+// IsArnFunc takes a string value and determines if it's an ARN or not
+func IsARNFunc(value string) bool {
+	return arn.IsARN(value)
+}
+
+// IsSSMParamFunc takes a string value and determines if it's a SSM parameter
+func IsSSMParamFunc(value string) bool {
+	return strings.Contains(value, ":ssm:") || !strings.Contains(value, ":secretsmanager:")
```
Lou1415926: Since this is a public function, we should think of the implementation from a client perspective: does it make sense fir `IsSSMParamFunc("arn:somrandomservice")` to return `true`? `arn:somerandomservice` is not an SSM parameter and yet it's still returning true.

If we go with your current approach, I'd suggest keeping only the first part of the condition for this function, i.e.  `strings.Contains(value, ":ssm:")`. Then, in `task/cf.yml`, when you try to determine if it's an SSM param, instead of calling just `IsSSMParam $valueFrom`, do `or (IsSSMParam $valueFrom) (not IsARN $valueFrom)`

internal/pkg/template/template_functions.go
```diff
@@ -25,6 +26,41 @@ func ReplaceDashesFunc(logicalID string) string {
...
+
+// IsSecretsManagerSecretsFunc takes a string value and determines if it's a secretes manager secret
+func IsSecretsManagerSecretsFunc(value string) bool {
+	return strings.Contains(value, ":secretsmanager:")
+}
+
+// IsSSMParamFlagFunc takes secret values and determines if it consist a SSM parameter
+func IsSSMParamFlagFunc(secrets map[string]string) bool {
+	for _, value := range secrets {
+		if !arn.IsARN(value) || strings.Contains(value, ":ssm:") {
```
Lou1415926: We can perhaps reuse the other functions, e.g.
```
if !IsARNFunc(value) || IsSSMPramFunc(value)
```

internal/pkg/template/template_functions.go
```diff
@@ -25,6 +26,41 @@ func ReplaceDashesFunc(logicalID string) string {
...
+	return strings.Contains(value, ":ssm:") || !strings.Contains(value, ":secretsmanager:")
+}
+
+// IsSecretsManagerSecretsFunc takes a string value and determines if it's a secretes manager secret
+func IsSecretsManagerSecretsFunc(value string) bool {
+	return strings.Contains(value, ":secretsmanager:")
+}
+
+// IsSSMParamFlagFunc takes secret values and determines if it consist a SSM parameter
+func IsSSMParamFlagFunc(secrets map[string]string) bool {
```
Lou1415926: nit: can it be renamed to `hasSSMParamFunc`?

internal/pkg/template/templates/task/cf.yml
```diff
@@ -122,8 +122,33 @@ Resources:
                       "secretsmanager:ResourceTag/copilot-environment": !Ref Env
                   Resource:
                     - !Sub arn:${AWS::Partition}:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:*
-          - !Ref AWS::NoValue
-      {{- end}}
+          - PolicyName: 'PullSecrets'
+            PolicyDocument:
+              Version: '2012-10-17'
+              Statement:
+              {{- if isSSMParamFlag .Secrets}}
```
Lou1415926: An alternative would be to have `SSM` and `SecretsManager` as subfileds of `Secrets` (so instead of a `map`, `Secrets` is probably a `struct` now), then here it could be:
```
{{- if .Secrets.SSM}}
   - Effect: 'Allow'
     Action:
         - ssm:GetParameters
      Resource: {{range $name, $valueFrom := .Secrets}} 
       {{- if isARN $valueFrom }}
        - {{$valueFrom}}
        {{- else }}
         - !Sub arn:${AWS::Partition}:ssm:${AWS::Region}:${AWS::AccountId}:parameter{{$valueFrom}}
        {{- end}} {{- end}}
{{- end}}
```
In `cli/task_run.go`, pass ssm secrets to `Secrets.SSM`, and secretsmanager secrets to `Secrets.SecretsManager`.

paragbhingre: Instead of changing the the original `Secrets` I have create two new parameters `SSMParamSecrets` and `secretsManagerSecrets` which will be used everywhere in cfn template to find out if those secrets exist or not?

internal/pkg/template/template_functions.go
paragbhingre: This function is no more needed. I have removed all the public functions as we now have `SSMParamSecrets` and `secretsManagerSecrets` which helps us find out if those particular secrets are present or not.

internal/pkg/template/template_functions.go
paragbhingre: I have removed all the public functions as we now have SSMParamSecrets and secretsManagerSecrets which helps us find out if those particular secrets are present or not.

internal/pkg/template/template_functions.go
paragbhingre: Our new functionality with 2 new parameters i.e. `SSMParamSecrets` and `secretsManagerSecrets` takes care of the logic written in this function so we don't need this function.

internal/pkg/cli/task_run_test.go
paragbhingre: Done

internal/pkg/cli/task_run.go
paragbhingre: Changed.

internal/pkg/cli/task_run.go
```diff
@@ -101,11 +107,14 @@ type runTaskVars struct {
...
 
-	envVars      map[string]string
-	secrets      map[string]string
-	command      string
-	entrypoint   string
-	resourceTags map[string]string
+	envVars                  map[string]string
+	secrets                  map[string]string
+	SSMParamSecrets          map[string]string
+	secretsManagerSecrets    map[string]string
```
iamhopaul123: seems like this one is unused. can we fix this static check issue?

internal/pkg/cli/task_run.go
```diff
@@ -108,6 +109,8 @@ type runTaskVars struct {
 
 	envVars                  map[string]string
 	secrets                  map[string]string
+	SSMParamSecrets          map[string]string
```
Lou1415926: nit: this doesn't need to be exported it seems like
```suggestion
	ssmParamSecrets          map[string]string
```

internal/pkg/template/templates/task/cf.yml
```diff
@@ -96,7 +96,7 @@ Resources:
             Action: 'sts:AssumeRole'
       ManagedPolicyArns:
         - !Sub 'arn:${AWS::Partition}:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy'
-      {{- if .Secrets}}
```
Lou1415926, line 99: Just to double check, Is the template still correct with the combination of condition
```
- !If
  - HasAppAndEnv
  - Policy
  - !Ref AWS::NoValue
```
and the new nested ``{{ If }}``s ?

Lou1415926 (commented):
What do you think of refactoring `cli/task_run.go` to this sturcutre?
```go
func (o *runTaskOpts) Validate() {
	// Other validations...
	for _, value := range o.secrets {
			if !isSSM(value) && !isSecretsManager(value) {
				return errors.New("some error")
			}
		}
}

func (o *runtaskOpts) Ask() {
	if len(o.secrets) > 0 {
		o.categorizeSecrets()
		if err := o.confirmSecretsAccess(); err != nil {
			return err
		}
	}
}

func isSSM(value string) {
	// For SSM parameter you can specify it as ARN or name if it exists in the same Region as the task you are launching.
	return !template.IsARNFunc(value) || strings.Contains(value, ":ssm:")
}

func isSecretsManager(value string) {
	return strings.Contains(value, ":secretsmanager:")
}

func (o *runTaskOpts) categorizeSecrets() {
	o.secretsManagerSecrets = make(map[string]string)
	o.ssmParamSecrets = make(map[string]string)
	for name, value := range o.secrets {
		if isSSM(value){
			o.ssmParamSecrets[name] = value
		}
		if isSecretsManager(VALUE) {
			o.secretsManagerSecrets[name] = value
		}
	}
}

func (o *runTaskOpts) confirmSecretsAccess {
	// ...
	if o.acknowledgeSecretsAccess {
		return nil
	}

	for name, value := range o.ssmParamSecrets {
 		// log...
	}

	for name, value := range o.secretsmanager {
		// log....
	}
}
```
A few reasons that I proposed this:
1. Having `isSSM` and `isSecretsManager` functions centralizes the logic that checks the type of the secrets. In the future, if the criteria of `ssm` or `secretsmanager` changes (e.g. what if we allow secretsmanager to be passed as name of specifial form in the future?), we would only need to change one block of codes.
2. The `categorizeSecrets` function removes the repetative logic in the current `confirmSecretsAccess` (notice how the `if {} else {}` block contains similar logics). Moreover, categorizing secrets seems like a "side-effect" of `confirmSecretsAccess` (you can't tell from the name of `confirmSecretsAccess` that it also buckets the secrets into two categories). Therefore, I propose moving the logic out of `confirmSecretsAccess` and instead explicitly call `confirmSecretsAccess` from `Ask`.

Let me know if this proposal makes sense!

internal/pkg/template/templates/task/cf.yml
paragbhingre: So just to explain -- 
if we have App and Env then we will add a Policy else nothing will be added.
On the other hand if App and Env is not provided then we will prompt permission question and that flow will build SSMParamSecrets and secretsManagerSecrets maps and if those are present then nested if logics will work and we will add specific policy statements

Lou1415926: We are using both "Condition" and go-template rendering (i.e. `{{ If }}` ) to implement a `if-else` logic. Can we simplify it so that we use either "Condition" or go-template rendering? This may seem confusing for future readers.

Lou1415926: For example
```yaml
{{-If .Secrets}}
  {{- if and .App .Env}}
  {{- else if}}
  {{- end}}
{{- end}}
```

internal/pkg/cli/task_run.go
```diff
@@ -371,6 +382,74 @@ func (o *runTaskOpts) Validate() error {
 		}
 	}
 
+	if len(o.secrets) > 0 {
```
Lou1415926: This is fine but I feel like this `if` condition is implicitly covered by the for-loop. Like if `len(o.secrets) <= 0`, what's inside of the for-loop won't executed.

internal/pkg/template/template_functions.go
```diff
@@ -25,6 +26,11 @@ func ReplaceDashesFunc(logicalID string) string {
 	return strings.ReplaceAll(logicalID, "-", dashReplacement)
 }
 
+// IsArnFunc takes a string value and determines if it's an ARN or not
```
Lou1415926: nit:
```suggestion
// IsArnFunc takes a string value and determines if it's an ARN or not.
```

internal/pkg/template/templates/task/cf.yml
```diff
@@ -96,34 +89,65 @@ Resources:
             Action: 'sts:AssumeRole'
       ManagedPolicyArns:
         - !Sub 'arn:${AWS::Partition}:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy'
+
+      {{- if eq .ExecutionRole "" }}
```
Lou1415926: We probably don't need this condition because if `ExecutionRole` is given by user, we won't be using `DefaultExecutionRole` 🤔

internal/pkg/template/templates/task/cf.yml
```diff
@@ -96,34 +89,65 @@ Resources:
...
+          PolicyDocument:
+            Version: '2012-10-17'
+            Statement:
+              {{- if .SSMParamSecrets }}
+              - Effect: 'Allow'
+                Action:
+                  - ssm:GetParameters
+                Resource: {{range $name, $valueFrom := .SSMParamSecrets}}
+                  {{- if not (isARN $valueFrom) }}
+                  - !Sub arn:${AWS::Partition}:ssm:${AWS::Region}:${AWS::AccountId}:parameter{{$valueFrom}}
```
Lou1415926: What if `valueFrom` doesn't start with `"/"`?

internal/pkg/deploy/task.go
```diff
@@ -19,13 +19,15 @@ type CreateTaskResourcesInput struct {
...
-	EntryPoint    []string
-	EnvVars       map[string]string
-	Secrets       map[string]string
+	Image                 string
+	TaskRole              string
+	ExecutionRole         string
+	Command               []string
+	EntryPoint            []string
+	EnvVars               map[string]string
+	Secrets               map[string]string
```
Lou1415926: Is `Secrets` the union of `SSMParamSecrets` and `SecretsManagerSecrets`? Do you think we can ditch this and in `task/cf.yml` we replace
```
{{- if .Secrets }}
```
with 
```
{{- if or .SSMParamSecrets .SecretsManagerSecrets}}
```
and 
```yaml
for $name, $valueFrom := .Secrets
```
with
```
for $name, $valueFrom := .SSMParamSecrets
...
for $name, $valueFrom := .SecretsManagerSecrets
```
? 

Alternatively, we can make `SSMParamSecrets` and `SecretsManagerSecrets` child fields of `Secrets`.

internal/pkg/cli/task_run.go
```diff
@@ -377,6 +391,72 @@ func (o *runTaskOpts) Validate() error {
...
+	return nil
+}
+
+func isSSM(value string) bool {
+	// For SSM parameter you can specify it as ARN or name if it exists in the same Region as the task you are launching.
+	return !template.IsARNFunc(value) || strings.Contains(value, ":ssm:")
+}
+
+func isSecretsManager(value string) bool {
+	return strings.Contains(value, ":secretsmanager:")
```
iamhopaul123: Why don't we need to validate if it is a valid ARN?

internal/pkg/cli/task_run.go
```diff
@@ -377,6 +391,72 @@ func (o *runTaskOpts) Validate() error {
...
+}
+
+func (o *runTaskOpts) categorizeSecrets() {
+	o.secretsManagerSecrets = make(map[string]string)
+	o.ssmParamSecrets = make(map[string]string)
```
iamhopaul123: We might want to init these value in the constructor `newTaskRunOpts`

internal/pkg/template/templates/task/cf.yml
```diff
@@ -69,8 +62,11 @@ Resources:
           Environment:{{range $name, $value := .EnvVars}}
           - Name: {{$name}}
             Value: {{$value}}{{end}}{{end}}
-          {{- if .Secrets}}
-          Secrets:{{range $name, $valueFrom := .Secrets}}
+          {{- if .SSMParamSecrets}}
+          Secrets:{{range $name, $valueFrom := .SSMParamSecrets}}
```
iamhopaul123, line 62: What if `.SecretsManagerSecrets` is not empty and `.SSMParamSecrets` is empty. Do we still get the `Secrets:`? Or it will end up being part of `Environment`?

internal/pkg/cli/task_run.go
```diff
@@ -500,6 +580,12 @@ func (o *runTaskOpts) Ask() error {
 			return err
 		}
 	}
+	if len(o.secrets) > 0 {
+		o.categorizeSecrets()
```
iamhopaul123: We should do this after `confirmSecretsAccess` (would be best to place in `Execute()` so that we don't need to define extra member variables `ssmParamSecrets` and `secretsManagerSecrets`). Otherwise it might happen that both (`App`, `Env`) and (`ssmParamSecrets`, `secretsManagerSecrets`) are set.

Lou1415926: I think this is done during `Ask` because we can reuse the categorized variable during `Ask` (when we prompts, we list out all ssm secrets, and then all secretsmanager secrets). 

> Otherwise it might happen that both (App, Env) and (ssmParamSecrets, secretsManagerSecrets) are set.

I wonder why is this an issue though 🤔 ? If `App` `Env` can co-exist with `Secrets`, why can't they with `SSMParam` and `SecretsManagerSecrets`?

internal/pkg/cli/task_run.go
paragbhingre: The reason we are initializing them here is because they will be used only in case customers doesn't specify app and env and then we will initiate them and use it further, else we will not initiate them.

Lou1415926: No they are used with or without app and env right? Like as long as there is `--secrets` they are categorized into either ssm or secretsmanager.

internal/pkg/template/templates/task/cf.yml
```diff
@@ -96,34 +93,64 @@ Resources:
...
+                Action:
+                  - secretsmanager:GetSecretValue
+                Condition:
+                  StringEquals:
+                    "secretsmanager:ResourceTag/copilot-application": {{ .App }}
+                    "secretsmanager:ResourceTag/copilot-environment": {{ .Env }}
+                Resource:
+                  - !Sub arn:${AWS::Partition}:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:*
+
+      {{- else if or .SSMParamSecrets .SecretsManagerSecrets }}
```
Lou1415926: nit:
```yaml
{{- if or .SSMParamSecrets .SecretsManagerSecrets}}
    {{- if and .App .Env }}
    {{- else if or .SSMParamSecrets .SecretsManagerSecrets }}. # This if-condition is redundant
    {{- end}}
{{- end}}
```

internal/pkg/template/templates/task/cf.yml
```diff
@@ -81,8 +77,9 @@ Resources:
       NetworkMode: awsvpc
       Cpu: !Ref TaskCPU
       Memory: !Ref TaskMemory
-      ExecutionRoleArn: !If [HasExecutionRole, !Ref ExecutionRole, !GetAtt DefaultExecutionRole.Arn]
+      ExecutionRoleArn: {{- if eq .ExecutionRole "" }} !GetAtt DefaultExecutionRole.Arn {{- else }} {{.ExecutionRole}} {{- end }}
```
Lou1415926, line 76: nit: we can remove the `HasExecutionRole` condition and the `ExecutionRole` parameter if they are no longer in use.

Lou1415926 (commented):
Nice! Just two nits + discussiing/addressing PH's comment.

internal/pkg/template/templates/task/cf.yml
paragbhingre: Right make sense, I have modified if condition to - `{{- if or .SSMParamSecrets .SecretsManagerSecrets}}` So both the types of secrets will be added in `secrets:`

paragbhingre: ohh right, I have modified if condition with `{{- if or .SSMParamSecrets .SecretsManagerSecrets}}` so that `Secrets:` will have both SSMParamSecrets and SecretsManagerSecrets

internal/pkg/cli/task_run.go
iamhopaul123: Because https://github.com/aws/copilot-cli/pull/3288/files#diff-ee1122ecbf021b404a3085486571ea2beb4dff929bbe256e99b852c18f99dd16R100 and https://github.com/aws/copilot-cli/pull/3288/files#diff-ee1122ecbf021b404a3085486571ea2beb4dff929bbe256e99b852c18f99dd16R125 they have the same policy name making me feel they are contradict with each other and only one set exists at a time

paragbhingre: But we now have else condition [here](https://github.com/aws/copilot-cli/pull/3288/files#diff-ee1122ecbf021b404a3085486571ea2beb4dff929bbe256e99b852c18f99dd16R119) so either app and env block will be executed or `SSMParam` and `SecretsManagerSecrets` block will be executed.

iamhopaul123: Gotcha yeah i know it wouldn't be a real issue. It's just `o.categorizeSecrets()` looks very stateful to me.

iamhopaul123: Alternatively, we could consider caching it and have `categorizeSecrets` to return a struct of SSM and secret manager. So when the cache var is not `nil` we just return the cache var. Otherwise, we try to categorize and return the var. Purpose for this is to make it not that stateful: whenever we need these vars we call the function.

example: https://github.com/aws/copilot-cli/blob/e08a090addd294ca1d516bf32ce9c6fd5649463d/internal/pkg/cli/pipeline_deploy.go#L229

internal/pkg/cli/task_run.go
paragbhingre: Oops sorry, that was the case before not now. Yo are right Wan

internal/pkg/cli/task_run.go
Lou1415926:
```go
func (o *runTaskOpts) getCategorizedSecrets() {
	if o.ssmParamSecrets != nil || o.secretsManagerSecrets != nil {
                 return o.ssmParamSecrets, o.secretsManagerSecrets
          }
	for name, value := range o.secrets {
		if isSSM(value){
			o.ssmParamSecrets[name] = value
		}
		if isSecretsManager(VALUE) {
			o.secretsManagerSecrets[name] = value
		}
	}
        return o.ssmParamSecrets, o.secretsManagerSecrets 
}

func (o *runTaskOpts) confirmSecretsAccess() error {
   ssm, secrets := o.getCategorizedSecrets()
}

func (o *runTaskOpts) Execute() error {
   ssm, secrets := o.getCategorizedSecrets()
}
```

internal/pkg/template/templates/task/cf.yml
paragbhingre: This case is considered in the next commit

internal/pkg/cli/task_run.go
```diff
@@ -155,6 +163,10 @@ type runTaskOpts struct {
 	runTaskRequestFromECSService func(client ecs.ECSServiceDescriber, cluster, service string) (*ecs.RunTaskRequest, error)
 	runTaskRequestFromService    func(client ecs.ServiceDescriber, app, env, svc string) (*ecs.RunTaskRequest, error)
 	runTaskRequestFromJob        func(client ecs.JobDescriber, app, env, job string) (*ecs.RunTaskRequest, error)
+
+	// parameters to hold SSM Param and Secrets Manager Secrets
```
Lou1415926:
```suggestion
	// Cached variables to hold SSM Param and Secrets Manager Secrets.
```

internal/pkg/cli/task_run.go
```diff
@@ -377,6 +393,77 @@ func (o *runTaskOpts) Validate() error {
 		}
 	}
 
+	for _, value := range o.secrets {
+		//For Systems Manager Parameter Store parameter you can specify it as name or the ARN if it exists in the same Region as the task you are launching.
```
Lou1415926:
```suggestion
```

internal/pkg/template/templates/task/cf.yml
```diff
@@ -69,8 +58,11 @@ Resources:
           Environment:{{range $name, $value := .EnvVars}}
           - Name: {{$name}}
             Value: {{$value}}{{end}}{{end}}
-          {{- if .Secrets}}
-          Secrets:{{range $name, $valueFrom := .Secrets}}
+          {{- if or .SSMParamSecrets .SecretsManagerSecrets}}
+          Secrets:{{range $name, $valueFrom := .SSMParamSecrets}}
+          - Name: {{$name}}
+            ValueFrom: {{$valueFrom}}{{end}}
+          {{- end}}{{- if .SecretsManagerSecrets }}{{range $name, $valueFrom := .SecretsManagerSecrets}}
```
Lou1415926: Can we remove the first part? 
```suggestion
          {{range $name, $valueFrom := .SecretsManagerSecrets}}
```

internal/pkg/template/templates/task/cf.yml
```diff
@@ -96,34 +89,64 @@ Resources:
...
+              - Effect: 'Allow'
+                Action:
+                  - secretsmanager:GetSecretValue
+                Condition:
+                  StringEquals:
+                    "secretsmanager:ResourceTag/copilot-application": {{ .App }}
+                    "secretsmanager:ResourceTag/copilot-environment": {{ .Env }}
+                Resource:
+                  - !Sub arn:${AWS::Partition}:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:*
+
```
Lou1415926:
```suggestion
```

internal/pkg/template/templates/task/cf.yml
```diff
@@ -96,34 +89,64 @@ Resources:
...
+      {{- else }}
+
```
Lou1415926:
```suggestion
```

internal/pkg/template/templates/task/cf.yml
```diff
@@ -96,34 +89,64 @@ Resources:
...
-      {{- end}}
+                Resource: {{range $name, $valueFrom := .SSMParamSecrets}}
+                  {{- if not (isARN $valueFrom) }}
+                  - !Sub arn:${AWS::Partition}:ssm:${AWS::Region}:${AWS::AccountId}:parameter/{{ trimSlashPrefix $valueFrom}}
+                    {{- else }}
+                  - {{ $valueFrom }}
+                    {{- end }}
+                    {{ end }}
+              {{- end }}
+
```
Lou1415926:
```suggestion
```

Lou1415926 (commented):
Looks good! Mostly just some nits -

internal/pkg/cli/task_run.go
```diff
@@ -377,6 +393,77 @@ func (o *runTaskOpts) Validate() error {
...
+
+	for _, value := range ssmParamSecrets {
+		log.Infoln("* " + value)
+	}
+
+	log.Infoln("\nand secretsmanager:GetSecretValue to the following Secrets Manager secrets:")
+
+	for _, value := range secretsManagerSecrets {
+		log.Infoln("* " + value)
+	}
```
Lou1415926: We should consider the case when only either one of ssm or secretsmanager type is specified. You could get a prompt that look like:
```
Looks like you're requesting ssm:GetParameters to the following SSM parameters:
* some secrets

and secretsmanager:GetSecretValue to the following Secrets Manager secrets:

  Do you grant permission to the ECS/Fargate agent for these secrets? [? for help] (y/N)
```
with the current implementation.

paragbhingre: I actually thought of that use case but I thought it is okay to show prompts for both ssm and secrets manager together even when one of them is not present. I will handle this use case.

iamhopaul123 (approved):
LGTM 🎉

Lou1415926 (approved):
yay awesome!

PR #3292, opened 2022-02-17
chore(task): improve error message for AccessDenied to secrets

paragbhingre:
This PR improves error messaging when AccessDenied error occurs while accessing the secrets


internal/pkg/cli/task_run.go
```diff
@@ -578,6 +578,9 @@ func (o *runTaskOpts) Execute() error {
 
 	tasks, err := o.runTask()
 	if err != nil {
+		if strings.Contains(err.Error(), "ResourceInitializationError") && o.appName != "" && o.env != "" {
+			log.Error("It looks like your task is not able to pull the secrets. Did you tag secrets with \"copilot-application\" and \"copilot-environment\" tags \n")
```
efekarakus: nit: Go has a special syntax for "raw string literals" that makes it easier to write these long strings with escaped quotes: https://yourbasic.org/golang/multiline-string/
```suggestion
			log.Error(`It looks like your task is not able to pull the secrets.
Did you tag your secrets with the "copilot-application" and "copilot-environment" tags?
`)
```

internal/pkg/cli/task_run.go
```diff
@@ -578,6 +578,9 @@ func (o *runTaskOpts) Execute() error {
 
 	tasks, err := o.runTask()
 	if err != nil {
+		if strings.Contains(err.Error(), "ResourceInitializationError") && o.appName != "" && o.env != "" {
```
Lou1415926: Do you think we need to be more specific in checking the error, for example check if it contains keywords like "AccessDeniedException" and "unable to retrieve secrets"?

paragbhingre: make sense

efekarakus (approved):
😍

huanjani (approved):
:shipit:

PR #3295, opened 2022-02-17
chore(env): guide users who are importing only one type of subnet

huanjani:
Previously, we logged warnings about the absence of public/private subnets **only** if the VPC being imported (during `env init`) did not have that type of subnet. A more common use case is that the VPC has those subnets but the user chooses not to use them.

Users may not know, when selecting only private subnets to import, that they need to adapt their workload manifests with
```
network:
  vpc:
    placement: private
```
or that their environments will have limited capabilities (no load balancing!). 

These changes keep the warning about no private/public subnets present, but add a warning after the user has chosen not to add subnets of one of the types.

```
% copilot env init
Environment name: test
Credential source: [profile jan]
Default environment configuration? No, I'd like to import existing resources
VPC: vpc-07de0fad1234feff7

Public subnets: 
Note: If you proceed without public subnets, you will not be able to deploy 
Load Balanced Web Services in this environment, and will need to specify 'private' 
network placement in your workload manifest(s). See the manifest documentation 
specific to your workload type(s) (https://aws.github.io/copilot-cli/docs/manifest/overview/).

  Which private subnets would you like to use?  [Use arrows to move, space to select, type to filter]
  > [ ]  subnet-01b722c41f5e08618 (copilot-buildspec-test-priv0)
  [ ]  subnet-089834da347785694 (copilot-buildspec-test-priv1)
```
```
% copilot env init
Environment name: test
Credential source: [profile jan]
Default environment configuration? No, I'd like to import existing resources
VPC: vpc-07de0fad1234feff7
Public subnets: subnet-025f72e8f354d7497 (copilot-buildspec-test-pub0), subnet-0ad88c532a6a95cbb (copilot-buildspec-test-pub1)

Private subnets: 
Note: If you proceed without private subnets, you will not 
be able to add them after this environment is created.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.17
 
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.2
-	github.com/aws/aws-sdk-go v1.42.47
+	github.com/aws/aws-sdk-go v1.42.52
 	github.com/briandowns/spinner v1.18.1
 	github.com/dustin/go-humanize v1.0.0
 	github.com/fatih/color v1.13.0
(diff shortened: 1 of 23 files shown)


efekarakus (approved):
really nice PR description and warning message 😍

Lou1415926 (commented):
Awesome guidance!!

Do we need to show these logs for users they're importing VPCs via flags?

huanjani:
> Awesome guidance!!
> 
> Do we need to show these logs for users they're importing VPCs via flags?

If they import one type via flags, we prompt them to select the other type; if they proceed without selecting any, we log the warning. If they import both types via flags, they don't need the warning (we validate they pass in more than one). If they pass in none via flags, we prompt them to select both types.

Lou1415926 (approved):

Issue #3290, opened 2022-02-16
Issue with deletion of a service


lukonjun:
Hi, I am currently trying to get branch deployments with copilot going. For this I also need to delete services in a pipeline. 

I can delete the service via the command line anymore
```
$ copilot svc delete --name poc-branch-deployments-24 --env demo-env --yes
✘ ValidationException: Parameter name: can't be prefixed with "ssm" (case-insensitive). If formed as a path, it can consist of sub-paths divided by slash symbol; each sub-path can be formed as a mix of letters, numbers and the following 3 symbols .-_
	status code: 400, request id: f7d62910-1800-4459-ae2d-c7479322d1d7
```
Do you have any ideas what the reason for this could be, I specifically started a fresh Ubuntu VM where I installed the aws and copilot cli.

efekarakus:
Hi @lukonjun ! 

Apologies for the inconvenience 🙇 what happens if you provide the app name as as well `--app` while deleting the service? does that throw an error as well

lukonjun:
Hi @efekarakus, did not know passing the app flag there is valid, as it is not in the docs https://aws.github.io/copilot-cli/docs/commands/svc-delete/
However when I try this, copilot reports that it cant find the svc
```
ubuntu@primary:~$ copilot svc ls
Application: copilots-demos
Name                            Type
----                            ----
node-server                     Load Balanced Web Service
poc-branch-deployments-14       Load Balanced Web Service
poc-branch-deployments-15       Load Balanced Web Service
poc-branch-deployments-16       Load Balanced Web Service
poc-branch-deployments-pr-19    Load Balanced Web Service
poc-branch-deployments-pr-20    Load Balanced Web Service
poc-branch-deployments-pr-22    Load Balanced Web Service
poc-branch-deployments-pr-23    Load Balanced Web Service
poc-branch-deployments-pr-24    Load Balanced Web Service
poc-branch-deployments-staging  Load Balanced Web Service
poc-branch-deployments-17       Load Balanced Web Service
poc-branch-deployments-pr-21    Load Balanced Web Service
ubuntu@primary:~$ copilot svc delete --app copilots-demos --name poc-branch-deployments-24 --env demo-env --yes
✘ couldn't find service poc-branch-deployments-24 in the application copilots-demos
```

efekarakus:
Oh the issue might be because the `--name` value is `poc-branch-deployments-24` instead of `poc-branch-deployments-pr-24` the "-pr" substring is missing!

[...]

PR #3302, opened 2022-02-22
fix: validate app name before using it to validate svc name in `svc delete`

Lou1415926:
This fixes a bug that surfaces in #3290. We should make sure that `o.appName` is specified (either through flag or prompts) before we use it to validate service name.

This PR also reflects a new coding pattern that we'd like to adopt: validate only optional flags in `Validate` and ask or validate required flag in `Ask`.

In `Ask`, if the prompt is a selection prompt, then after the prompt we shouldn't do existence check. For example, if we need to prompt to select from existing services from application, then we shouldn't validate whether the selected service exists in the application, because this is redundant.

### DIFF
--- a/internal/pkg/cli/svc_delete.go
+++ b/internal/pkg/cli/svc_delete.go
@@ -99,11 +99,6 @@ func newDeleteSvcOpts(vars deleteSvcVars) (*deleteSvcOpts, error) {
 
 // Validate returns an error if the user inputs are invalid.
 func (o *deleteSvcOpts) Validate() error {
-	if o.name != "" {
-		if _, err := o.store.GetService(o.appName, o.name); err != nil {
-			return err
-		}
-	}
 	if o.envName != "" {
 		return o.validateEnvName()
 	}
@@ -112,11 +107,25 @@ func (o *deleteSvcOpts) Validate() error {
 
 // Ask prompts the user for any required flags.
 func (o *deleteSvcOpts) Ask() error {
-	if err := o.askAppName(); err != nil {
-		return err
+	if o.appName != "" {
+		if _, err := o.store.GetApplication(o.appName); err != nil {
+			return err
+		}
+	} else {
+		if err := o.askAppName(); err != nil {
+			return err
+		}
 	}
-	if err := o.askSvcName(); err != nil {
-		return err
+
+	if o.name != "" {
+		if _, err := o.store.GetService(o.appName, o.name); err != nil {
+			return err
+		}
+
+	} else {
+		if err := o.askSvcName(); err != nil {
+			return err
+		}
 	}
 
 	if o.skipConfirmation {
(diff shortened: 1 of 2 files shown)


internal/pkg/cli/svc_delete.go
```diff
@@ -99,11 +99,6 @@ func newDeleteSvcOpts(vars deleteSvcVars) (*deleteSvcOpts, error) {
 
 // Validate returns an error if the user inputs are invalid.
 func (o *deleteSvcOpts) Validate() error {
-	if o.name != "" {
-		if _, err := o.store.GetService(o.appName, o.name); err != nil {
-			return err
-		}
-	}
 	if o.envName != "" {
 		return o.validateEnvName()
```
efekarakus: should we just have return `nil` for `Validate`?
this could also blow up if the app name is empty

internal/pkg/cli/svc_delete.go
```diff
@@ -131,6 +137,9 @@ func (o *deleteSvcOpts) Ask() error {
 		// When a customer provides a particular environment,
 		// we'll just delete the service from that environment -
 		// but keep it in the app.
+		if err := o.validateEnvName(); err != nil {
```
huanjani: Should we validate it outside if `if o.skipConfirmation`? 
Because the user could pass in an env but not want to skip confirmation.

internal/pkg/cli/svc_delete_test.go
```diff
@@ -264,18 +273,26 @@ func TestDeleteSvcOpts_Ask(t *testing.T) {
 
 			mockPrompter := mocks.NewMockprompter(ctrl)
 			mockSel := mocks.NewMockconfigSelector(ctrl)
-			test.mockPrompt(mockPrompter)
-			test.mockSel(mockSel)
+			mockStore := mocks.NewMockstore(ctrl)
+			// test.mockPrompt(mockPrompter)
```
huanjani: nit: delete these two lines?

huanjani (approved):
:shipit:

internal/pkg/cli/svc_delete.go
```diff
@@ -99,26 +99,36 @@ func newDeleteSvcOpts(vars deleteSvcVars) (*deleteSvcOpts, error) {
 
 // Validate returns an error if the user inputs are invalid.
```
efekarakus: nit: can we update the `Validate` comment to mention that it's for optional inputs.

internal/pkg/cli/svc_delete.go
```diff
@@ -99,26 +99,36 @@ func newDeleteSvcOpts(vars deleteSvcVars) (*deleteSvcOpts, error) {
...
-			return err
-		}
-	}
-	if o.envName != "" {
-		return o.validateEnvName()
-	}
 	return nil
 }
 
 // Ask prompts the user for any required flags.
```
efekarakus: nit: 
```suggestion
// Ask prompts and validates for any required flags .
```

efekarakus (approved):

PR #3303, opened 2022-02-22
chore: add cmd name to user-agent

efekarakus:

### DIFF
--- a/internal/pkg/aws/sessions/sessions.go
+++ b/internal/pkg/aws/sessions/sessions.go
@@ -21,29 +21,46 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 )
 
+// Timeout settings.
 const (
-	userAgentHeader = "User-Agent"
-
 	maxRetriesOnRecoverableFailures = 8 // Default provided by SDK is 3 which means requests are retried up to only 2 seconds.
 	credsTimeout                    = 10 * time.Second
 	clientTimeout                   = 30 * time.Second
+)
+
+// User-Agent settings.
+const (
+	userAgentProductName = "aws-copilot"
 )
 
 // Provider provides methods to create sessions.
 // Once a session is created, it's cached locally so that the same session is not re-created.
 type Provider struct {
 	defaultSess *session.Session
+
+	// Metadata associated with the provider.
+	userAgentExtras []string
 }
 
 var instance *Provider
 var once sync.Once
 
 // NewProvider returns a session Provider singleton.
-func NewProvider() *Provider {
+func NewProvider(options ...func(*Provider)) *Provider {
 	once.Do(func() {
 		instance = &Provider{}
 	})
+	for _, option := range options {
+		option(instance)
+	}
 	return instance
+}
+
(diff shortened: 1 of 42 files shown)


internal/pkg/aws/sessions/sessions.go
```diff
@@ -107,7 +124,7 @@ func (p *Provider) FromRole(roleARN string, region string) (*session.Session, er
 	if err != nil {
 		return nil, fmt.Errorf("error creating default session: %w", err)
 	}
-	defaultSession.Handlers.Build.PushBackNamed(userAgentHandler())
+	defaultSession.Handlers.Build.PushBackNamed(p.userAgentHandler())
```
iamhopaul123, line 127: Do we need this 🤔

internal/pkg/cli/env_upgrade.go
```diff
@@ -70,7 +70,8 @@ func newEnvUpgradeOpts(vars envUpgradeVars) (*envUpgradeOpts, error) {
 	if err != nil {
 		return nil, fmt.Errorf("connect to config store: %v", err)
 	}
-	defaultSession, err := sessions.NewProvider().Default()
+	sessProvider := sessions.NewProvider(sessions.UserAgentExtras("env upgrade"))
```
iamhopaul123, line 73: Nice it is so easy to use!

internal/pkg/cli/pipeline_status.go
```diff
@@ -60,7 +60,7 @@ func newPipelineStatusOpts(vars pipelineStatusVars) (*pipelineStatusOpts, error)
 		return nil, fmt.Errorf("new workspace client: %w", err)
 	}
 
-	session, err := sessions.NewProvider().Default()
+	session, err := sessions.NewProvider(sessions.UserAgentExtras("pipeline status")).Default()
```
Lou1415926, line 62: I guess we'd need to make sure then all sessions are using this provider from now on.

Lou1415926 (approved):
neato!

internal/pkg/cli/pipeline_status.go
efekarakus: It's actually not so bad! because `NewProvider` is always a singleton 😄 

so if users call `sessions.NewProvider()` again afterwards it'll still fetch the same one that had "pipeline status" as a user agent 😎

iamhopaul123 (approved):
LGTM!

internal/pkg/aws/sessions/sessions.go
efekarakus: you're right that was weird that it's duplicated! I'll fix it in the follow-up PR, thanks for catching it!

efekarakus: Actually, I'm going to keep it ! 

Turns out it's used by the `defaultSession` to assume a role. So it allows us to have the `aws-copilot/` user-agent for the one `sts assume-role` call

iamhopaul123: ah gotcha

Issue #3237, opened 2022-02-01
`env show` should show deployed jobs


iamhopaul123:
Right now `env show` only shows deployed services. However, it would be great if it can show deployed jobs as well. Related https://github.com/aws/copilot-cli/discussions/3236.

PR #3316, opened 2022-03-01
chore(env): `env show` should show deployed jobs

paragbhingre:
This PR adds functionality to show jobs on `env show` command

This PR addresses #3237 

User interface - 
Screen Shot 2022-03-01 at 11 47 20 AM


internal/pkg/describe/env.go
```diff
@@ -246,6 +253,27 @@ func (d *EnvDescriber) filterDeployedSvcs() ([]*config.Workload, error) {
 	return deployedSvcs, nil
 }
 
+// filterDeployedJobs lists the jobs that are deployed on the given app and environment
```
Lou1415926, line 256: nit: it's nicee to have a description but function comment is not required if the function is not exported.

internal/pkg/describe/env.go
```diff
@@ -246,6 +253,27 @@ func (d *EnvDescriber) filterDeployedSvcs() ([]*config.Workload, error) {
...
+func (d *EnvDescriber) filterDeployedJobs() ([]*config.Workload, error) {
+	allJobs, err := d.configStore.ListJobs(d.app)
+	if err != nil {
+		return nil, fmt.Errorf("list services for app %s: %w", d.app, err)
```
Lou1415926:
```suggestion
		return nil, fmt.Errorf("list jobs for app %s: %w", d.app, err)
```

internal/pkg/describe/env.go
```diff
@@ -246,6 +253,27 @@ func (d *EnvDescriber) filterDeployedSvcs() ([]*config.Workload, error) {
...
+	}
+	jobs := make(map[string]*config.Workload)
+	for _, job := range allJobs {
+		jobs[job.Name] = job
+	}
+	deployedJobNames, err := d.deployStore.ListDeployedJobs(d.app, d.env.Name)
```
Lou1415926, line 266: hmm we don't necessarily address this in this particular PR but in the near future we should refactor this so that `listDeployedWorkloads` of `deployStore` returns `[]*config.Workload` instead of `[]string`. This will simplify the logic of `filterDeployedSvcs` and `filterDeployedJobs`.

internal/pkg/describe/env.go
```diff
@@ -274,6 +302,15 @@ func (e *EnvDescription) HumanString() string {
 		fmt.Fprintf(writer, "  %s\t%s\n", svc.Name, svc.Type)
 	}
 	writer.Flush()
+	fmt.Fprint(writer, color.Bold.Sprint("\nJobs\n\n"))
```
Lou1415926, line 305: Do you think we want to only show the `Jobs` section only when there is/are job(s) deployed (btw I think what you have now is fine too)

iamhopaul123 (approved):
Great job!

internal/pkg/describe/service.go
```diff
@@ -35,12 +35,14 @@ type ConfigStoreSvc interface {
 	ListEnvironments(appName string) ([]*config.Environment, error)
 	ListServices(appName string) ([]*config.Workload, error)
 	GetWorkload(appName string, name string) (*config.Workload, error)
+	ListJobs(appName string) ([]*config.Workload, error)
 }
 
 // DeployedEnvServicesLister wraps methods of deploy store.
 type DeployedEnvServicesLister interface {
```
Lou1415926, line 43: What do you think of renaming this to `DeployedEnvWorkloadsLister`?

internal/pkg/describe/env.go
paragbhingre: yes totally agree, I guess then we don't need to make 2 separate calls to bring `[]*config.Workload` and then `[]string` and add custom logic to filter them.

internal/pkg/describe/env.go
paragbhingre: I guess it is okay for the awareness of the customers that Jobs are also being showed as a part of `env show`. And also it aligns with the current implementation for services.

internal/pkg/describe/service.go
paragbhingre: I thought of doing so but there are a lot of places where we need to change if we change it to `DeployedEnvWorkloadsLister`. for example, impacted places are - 
1. env.go
2. env_test.go
3. lb_web_service.go
4. lb_web_service_test.go
5. mock_service.go
6. rd_web_service.go
7. rd_web_service_test.go
8. service.go
9. worker_service.go

If we really be touching so many files as a part of `env show` implementation?

Lou1415926 (approved):
🎊

PR #3317, opened 2022-03-01
chore(cli): refactor validation and ask for `svc init`

iamhopaul123:

### DIFF
--- a/internal/pkg/cli/svc_init.go
+++ b/internal/pkg/cli/svc_init.go
@@ -170,6 +170,8 @@ func newInitSvcOpts(vars initSvcVars) (*initSvcOpts, error) {
 		mftReader:    ws,
 		dockerEngine: dockerengine.New(exec.NewCmd()),
 	}
+	// Override the app flag since we don't only allow running this command within a workspace.
+	opts.appName = tryReadingAppName()
 	opts.dockerfile = func(path string) dockerfileParser {
 		if opts.df != nil {
 			return opts.df
@@ -180,23 +182,15 @@ func newInitSvcOpts(vars initSvcVars) (*initSvcOpts, error) {
 	return opts, nil
 }
 
-// Validate returns an error if the flag values passed by the user are invalid.
+// Validate returns an error for any invalid optional flags.
 func (o *initSvcOpts) Validate() error {
-	if o.appName == "" {
+	if o.appName != "" {
+		if _, err := o.store.GetApplication(o.appName); err != nil {
+			return fmt.Errorf("get application %s configuration: %w", o.appName, err)
+		}
+	} else {
+		// NOTE: This command is required to be executed under a workspace. We don't prompt for it.
 		return errNoAppInWorkspace
-	}
-	if o.wkldType != "" {
-		if err := validateSvcType(o.wkldType); err != nil {
-			return err
-		}
-	}
-	if o.name != "" {
-		if err := validateSvcName(o.name, o.wkldType); err != nil {
-			return err
-		}
-		if err := o.validateDuplicateSvc(); err != nil {
-			return err
-		}
 	}
 	if o.dockerfilePath != "" && o.image != "" {
 		return fmt.Errorf("--%s and --%s cannot be specified together", dockerFileFlag, imageFlag)
@@ -222,10 +216,28 @@ func (o *initSvcOpts) Validate() error {
 	return nil
 }
 
-// Ask prompts for fields that are required but not passed in.
+// Ask prompts for and validates any required flags.
 func (o *initSvcOpts) Ask() error {
-	if err := o.askSvcName(); err != nil {
-		return err
+	if o.wkldType != "" {
+		if err := validateSvcType(o.wkldType); err != nil {
+			return err
+		}
+	} else {
+		if err := o.askSvcType(); err != nil {
+			return err
+		}
+	}
+	if o.name != "" {
+		if err := validateSvcName(o.name, o.wkldType); err != nil {
+			return err
+		}
+		if err := o.validateDuplicateSvc(); err != nil {
+			return err
+		}
+	} else {
+		if err := o.askSvcName(); err != nil {
+			return err
+		}
 	}
 	localMft, err := o.mftReader.ReadWorkloadManifest(o.name)
 	if err == nil {
@@ -233,7 +245,9 @@ func (o *initSvcOpts) Ask() error {
 		if err != nil {
 			return fmt.Errorf(`read "type" field for service %s from local manifest: %w`, o.name, err)
 		}
-		o.wkldType = svcType
+		if o.wkldType != svcType {
+			return fmt.Errorf("manifest file for service %s exists with a different type %s", o.name, svcType)
+		}
 		log.Infof("Manifest file for service %s already exists. Skipping configuration.\n", o.name)
 		return nil
 	}
@@ -243,9 +257,6 @@ func (o *initSvcOpts) Ask() error {
 	)
 	if !errors.As(err, &errNotFound) && !errors.As(err, &errWorkspaceNotFound) {
 		return fmt.Errorf("read manifest file for service %s: %w", o.name, err)
-	}
-	if err := o.askSvcType(); err != nil {
-		return err
 	}
 	err = o.askDockerfile()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/pkg/cli/svc_init.go
```diff
@@ -170,6 +170,8 @@ func newInitSvcOpts(vars initSvcVars) (*initSvcOpts, error) {
 		mftReader:    ws,
 		dockerEngine: dockerengine.New(exec.NewCmd()),
 	}
+	// Override the app flag since we don't only allow running this command within a workspace.
```
Lou1415926: nit: 🤔 
```suggestion
	// Override the app flag since we only allow running this command within a workspace.
```

internal/pkg/cli/svc_init.go
```diff
@@ -180,11 +182,8 @@ func newInitSvcOpts(vars initSvcVars) (*initSvcOpts, error) {
 	return opts, nil
 }
 
-// Validate returns an error if the flag values passed by the user are invalid.
+// Validate returns an error for any invalid optional flags.
 func (o *initSvcOpts) Validate() error {
-	if o.appName == "" {
-		return errNoAppInWorkspace
-	}
 	if o.wkldType != "" {
```
Lou1415926: Should we skip validating `wkldType` and `name` in `Validate` as they are not optional flags?

Lou1415926: Do we consider `--subscribe-topics` required for worker service?

internal/pkg/cli/svc_init.go
```diff
@@ -222,8 +221,16 @@ func (o *initSvcOpts) Validate() error {
 	return nil
 }
 
-// Ask prompts for fields that are required but not passed in.
+// Ask prompts for and validates any required flags.
 func (o *initSvcOpts) Ask() error {
+	if o.appName != "" {
```
Lou1415926: Since `o.appName` is no longer a user input (it's either the workspace or empty), should we validate it here or in `Validate`?

internal/pkg/cli/svc_init.go
iamhopaul123: > Do we consider --subscribe-topics required for worker service?

It's an optional flag actually because a worker service could subscribe no service.

internal/pkg/cli/svc_init.go
iamhopaul123: "Override the app flag since we don't only allow running this command within a workspace." 🤔

internal/pkg/cli/svc_init.go
```diff
@@ -170,6 +170,8 @@ func newInitSvcOpts(vars initSvcVars) (*initSvcOpts, error) {
 		mftReader:    ws,
 		dockerEngine: dockerengine.New(exec.NewCmd()),
 	}
+	// Override the app flag since we don't only allow running this command within a workspace.
+	opts.appName = tryReadingAppName()
```
efekarakus: do you think we should return an error if the flag value is different than the workspace app name? I wonder if it would be surprising that I specified `copilot svc init -a wrongapp` and copilot just changes the app name for me

internal/pkg/cli/svc_init.go
huanjani: I agree with @Lou1415926's suggestion. Or "Override the app flag because this command must be run within the app's workspace."?

Should we log a msg that the passed-in app is being overridden? Otherwise they won't realize that that's happening.

internal/pkg/cli/svc_init.go
```diff
@@ -222,18 +216,38 @@ func (o *initSvcOpts) Validate() error {
...
 	}
 	localMft, err := o.mftReader.ReadWorkloadManifest(o.name)
 	if err == nil {
 		svcType, err := localMft.WorkloadType()
 		if err != nil {
 			return fmt.Errorf(`read "type" field for service %s from local manifest: %w`, o.name, err)
 		}
-		o.wkldType = svcType
+		if o.wkldType != svcType {
+			return fmt.Errorf("manifest file for service %s exists with a different type %s", o.name, svcType)
```
efekarakus, line 252: nice!

internal/pkg/cli/svc_init.go
iamhopaul123: good idea!

huanjani: Ha, jinx! I suggested logging a msg, above, but agree that erroring out would be good.

iamhopaul123: ok done!

internal/pkg/cli/svc_init.go
iamhopaul123: sorry i misread the msg...corrected

efekarakus (approved):

internal/pkg/cli/svc_init.go
```diff
@@ -180,23 +182,19 @@ func newInitSvcOpts(vars initSvcVars) (*initSvcOpts, error) {
...
 func (o *initSvcOpts) Validate() error {
-	if o.appName == "" {
-		return errNoAppInWorkspace
-	}
-	if o.wkldType != "" {
-		if err := validateSvcType(o.wkldType); err != nil {
-			return err
+	if o.appName != "" {
+		// This command must be run within the app's workspace.
+		if o.appName != o.wsAppName {
```
huanjani: If `tryReadingAppName()` can't find an app name in `.workspace`, this value will be `""` so if the user passes in an app flag outside of a workspace (which is when they would most likely pass an app flag), they will get this error about the workspace already being registered with "app [blank]" rather than `errNoAppInWorkspace`.

Maybe we should check for an empty `o.wsAppName` sooner?

iamhopaul123: ooh good call

internal/pkg/cli/svc_init.go
```diff
@@ -180,23 +182,19 @@ func newInitSvcOpts(vars initSvcVars) (*initSvcOpts, error) {
...
-// Validate returns an error if the flag values passed by the user are invalid.
+// Validate returns an error for any invalid optional flags.
 func (o *initSvcOpts) Validate() error {
-	if o.appName == "" {
-		return errNoAppInWorkspace
-	}
-	if o.wkldType != "" {
-		if err := validateSvcType(o.wkldType); err != nil {
-			return err
+	if o.appName != "" {
```
Lou1415926: This looks good to me but do we need `wsAppName` as a member if `initSvcOpts`? Could we just
```
wsApp := tryReadingAppName()
if o.appName != "" {
}
o.appName = wsApp
```

internal/pkg/cli/svc_init.go
```diff
@@ -126,7 +126,8 @@ type initSvcOpts struct {
 	topics       []manifest.TopicSubscription
 
 	// Cache variables
-	df dockerfileParser
+	wsAppName string
```
huanjani, line 129: Q: Are you making this a cached variable rather than calling `tryReadingAppName()` within `Ask()` for ease of testing? Or are there other benefits?

internal/pkg/cli/svc_init.go
```diff
@@ -222,18 +220,38 @@ func (o *initSvcOpts) Validate() error {
...
+	}
+	if o.name != "" {
+		if err := validateSvcName(o.name, o.wkldType); err != nil {
+			return err
+		}
+		if err := o.validateDuplicateSvc(); err != nil {
+			return err
+		}
+	} else {
+		if err := o.askSvcName(); err != nil {
```
Lou1415926, line 235: Looking at the code this function sounds more like a `askAndValidateSvcName` to me...

We decided on the `if o.name != "" { // validate } else { //ask}` pattern because the questions being asked are selection prompts, i.e. they don't need to be further validated.

But this is an input prompt, and we still need to validate the input after prompting. Do you think we should instead do
```
if o.name != {
  // ask
}
// validate
```
?

internal/pkg/cli/svc_init.go
iamhopaul123: it's because `appName` is now no longer a required flag so we put it within `Validate`. and cached variable is for ease of testing.

internal/pkg/cli/svc_init.go
iamhopaul123: it's for the ease of testing

internal/pkg/cli/svc_init.go
iamhopaul123: tbh this is less ideal as if they don't have flag specified we'll do `validateSvcName` twice.

iamhopaul123: looks better tho so i made the change

Lou1415926 (approved):

PR #3318, opened 2022-03-01
chore(cli): change Ask/Validate roles for `pipeline delete`

huanjani:


internal/pkg/cli/pipeline_delete.go
```diff
@@ -63,42 +72,50 @@ func newDeletePipelineOpts(vars deletePipelineVars) (*deletePipelineOpts, error)
...
+	} else {
+		if err := o.askAppName(); err != nil {
+			return err
+		}
+	}
 	if o.name != "" {
 		if _, err := o.codepipeline.GetPipeline(o.name); err != nil {
 			return err
 		}
 	}
```
iamhopaul123, line 112: If name is not empty should we ask users to select from a list? or that's the next step after this PR

iamhopaul123 (approved):

internal/pkg/cli/pipeline_delete.go
```diff
@@ -63,42 +72,50 @@ func newDeletePipelineOpts(vars deletePipelineVars) (*deletePipelineOpts, error)
...
 		secretsmanager:     secretsmanager.New(defaultSess),
 		pipelineDeployer:   cloudformation.New(defaultSess),
 		ws:                 ws,
+		store:              ssmStore,
 	}
 
 	return opts, nil
 }
 
 // Validate returns an error if the flag values passed by the user are invalid.
```
efekarakus: nit: can we update the comment to say that `Validate` is for any optional fields?

efekarakus (approved):

internal/pkg/cli/pipeline_delete.go
huanjani: Yup, that's the next step. 👍🏼

PR #3320, opened 2022-03-02
chore(cli): change Ask/Validate roles for `pipeline deploy`

huanjani:


internal/pkg/cli/pipeline_deploy.go
```diff
@@ -148,19 +150,34 @@ func newDeployPipelineOpts(vars deployPipelineVars) (*deployPipelineOpts, error)
...
+func (o *deployPipelineOpts) Ask() error {
+	if o.appName != "" {
+		if _, err := o.store.GetApplication(o.appName); err != nil {
+			return err
+		}
+	} else {
+		if err := o.askAppName(); err != nil {
+			return err
+		}
+	}
```
iamhopaul123: We shouldn't ask for the app name as the command has to be run within the workspace. Otherwise we won't be able to find the correct workspace. Please refer to https://github.com/aws/copilot-cli/pull/3317/files#diff-1e7f78366fc8d6eec3339ca563e5fcbe1226766d42a4d3a02b730d9fbd2cc5a7R174

huanjani: Yup, you're right! Fixing.

Lou1415926 (commented):
+1 to PH's otherwise LGTM!

internal/pkg/cli/interfaces.go
```diff
@@ -479,6 +484,7 @@ type pipelineEnvSelector interface {
 }
 
 type wsPipelineSelector interface {
+	appSelector
```
Lou1415926: Does `appSelector` still need to be part of this?

internal/pkg/term/selector/selector.go
```diff
@@ -178,6 +178,7 @@ type WorkspaceSelect struct {
 
 // PipelineSelect is a workspace pipeline selector.
 type PipelineSelect struct {
+	*Select
```
Lou1415926: Is this still needed in `PipelineSelect`?

Lou1415926 (approved):
LGTM :shipit:

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -143,24 +144,44 @@ func newDeployPipelineOpts(vars deployPipelineVars) (*deployPipelineOpts, error)
...
+	// The passed-in app name value must be the same as the workspace app name, as we need to be in the correct workspace to read the pipeline manifest.
+	if o.appName != "" {
+		if o.appName != o.wsAppName {
+			return fmt.Errorf("cannot specify app %s because the workspace is already registered with app %s", o.appName, o.wsAppName)
+		}
+		// Validate the app name.
+		if _, err := o.store.GetApplication(o.appName); err != nil {
+			return fmt.Errorf("get application %s configuration: %w", o.appName, err)
 		}
```
iamhopaul123: Maybe validate the app probably when the flag is not set as well? In case the app specified within ws file doesn't have a SSM record

huanjani: Yeah, I thought about that. Then I considered that we haven't validated the result of `TryReadingAppName()` anywhere previously. But yes, in the spirit of improving erroring, I'll change it here.

iamhopaul123 (approved):

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -143,24 +149,41 @@ func newDeployPipelineOpts(vars deployPipelineVars) (*deployPipelineOpts, error)
...
 
-// Validate returns an error if the flag values passed by the user are invalid.
+// Validate returns an error if the optional flag values passed by the user are invalid.
 func (o *deployPipelineOpts) Validate() error {
+	if o.wsAppName == "" {
+		return errNoAppInWorkspace
+	}
+	// The passed-in app name value must be the same as the workspace app name, as we need to be in the correct workspace to read the pipeline manifest.
+	if o.appName != "" && o.appName != o.wsAppName {
+			return fmt.Errorf("cannot specify app %s because the workspace is already registered with app %s", o.appName, o.wsAppName)
```
dannyrandall, line 165: nit: extra tab

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -143,24 +149,41 @@ func newDeployPipelineOpts(vars deployPipelineVars) (*deployPipelineOpts, error)
...
+	}
+	// Validate the app name.
+	if _, err := o.store.GetApplication(o.wsAppName); err != nil {
+		return fmt.Errorf("get application %s configuration: %w", o.wsAppName, err)
+	}
+	return nil
+}
+
+// Ask prompts the user for any unprovided required fields and validates them.
+func (o *deployPipelineOpts) Ask() error {
```
dannyrandall, line 175: thoughts on this instead? just to remove [the nesting](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow)
```
if o.name == "" {
	if err := o.askPipelineName(); err != nil {
		return err
	}

	return nil
}

return o.validatePipelineName()
```

huanjani: Yes! Normally that would be much lovelier, but in this case we want to avoid validating the pipeline name that came from the selector because the user was presented with only valid options.

dannyrandall: i think my suggestion does avoid validating the pipeline name that comes from the selector, although i just realized it could be simplified even more while still keeping the original logic:

```
if o.name == "" {
    return o.askPipelineName()
}

return o.validatePipelineName()
```

huanjani: Oh, sorry, right-- bc of the `return nil`. 

I'll also move the 'if' back to `validatePipelineName()`.

PR #3322, opened 2022-03-03
chore(cli): change Validate/Ask roles for `pipeline init`

huanjani:
to come for this command: adding name flag, refactoring unit tests


internal/pkg/cli/pipeline_init.go
```diff
@@ -137,43 +138,52 @@ func newInitPipelineOpts(vars initPipelineVars) (*initPipelineOpts, error) {
...
+	if o.appName != "" {
+		if o.appName != o.wsAppName {
+			return fmt.Errorf("cannot specify app %s because the workspace is already registered with app %s", o.appName, o.wsAppName)
+		}
+		// Validate the app name.
+		if _, err := o.store.GetApplication(o.appName); err != nil {
+			return fmt.Errorf("get application %s configuration: %w", o.appName, err)
+		}
 	}
+	o.appName = o.wsAppName
```
efekarakus: nit: it's surprising that an assignment happens in `Validate` can we do it in the constructor instead?
```go
wsAppName := tryReadingAppName()
if vars.appName == "" {
   vars.appName = wsAppName
}
```

iamhopaul123: Hmm but that means we would have to move the validation above there as well because we need the original value they pass in via the flag for validation. If we override like this then there's no way for us to tell them there's something wrong with their app flag value

huanjani: But we wouldn't override if they pass in a value....

efekarakus (approved):

internal/pkg/cli/pipeline_init.go
```diff
@@ -137,43 +144,49 @@ func newInitPipelineOpts(vars initPipelineVars) (*initPipelineOpts, error) {
...
-	if o.appName == "" {
+	// This command must be executed in the app's workspace because the pipeline manifest and buildspec will be created and stored.
+	if o.wsAppName == "" {
 		return errNoAppInWorkspace
 	}
-	if _, err := o.store.GetApplication(o.appName); err != nil {
-		return err
+	if o.appName != "" && o.appName != o.wsAppName {
+			return fmt.Errorf("cannot specify app %s because the workspace is already registered with app %s", o.appName, o.wsAppName)
+		}
```
dannyrandall: nit: extra tab

internal/pkg/cli/pipeline_init.go
```diff
@@ -137,43 +144,49 @@ func newInitPipelineOpts(vars initPipelineVars) (*initPipelineOpts, error) {
...
 			return err
 		}
+	} else {
+		if err := o.askRepository(); err != nil {
+			return err
+		}
 	}
 
-	if o.environments != nil {
+	if len(o.environments) > 0 {
```
dannyrandall, line 172: q: previously, `askEnvs()` was always called and therefore `o.envConfigs` would be set. Now, if `o.environments` has values in it, `askEnvs()` _isn't_ called, so `o.envConfigs` won't be set. Is that what we want to happen?

internal/pkg/cli/pipeline_init.go
```diff
@@ -137,43 +144,49 @@ func newInitPipelineOpts(vars initPipelineVars) (*initPipelineOpts, error) {
...
+		wsAppName:        wsAppName,
 	}, nil
 }
 
-// Validate returns an error if the flag values passed by the user are invalid.
+// Validate returns an error if the optional flag values passed by the user are invalid.
 func (o *initPipelineOpts) Validate() error {
-	if o.appName == "" {
+	// This command must be executed in the app's workspace because the pipeline manifest and buildspec will be created and stored.
+	if o.wsAppName == "" {
```
iamhopaul123: Can we reuse `validateInputApp` this function? Also for the previously merged PR for `pipeline` deploy sorry...

internal/pkg/cli/pipeline_init.go
huanjani: good catch! fixed.

internal/pkg/cli/pipeline_init.go
huanjani: Ooh, yes-- I hadn't seen that! Thanks!

internal/pkg/cli/pipeline_init.go
```diff
@@ -220,38 +227,49 @@ func (o *initPipelineOpts) validateURL(url string) error {
 	return nil
 }
 
-func (o *initPipelineOpts) askEnvs() error {
-	if len(o.environments) == 0 {
-		envs, err := o.sel.Environments(pipelineSelectEnvPrompt, pipelineSelectEnvHelpPrompt, o.appName, func(order int) prompt.PromptConfig {
-			return prompt.WithFinalMessage(fmt.Sprintf("%s stage:", humanize.Ordinal(order)))
-		})
+// To avoid duplicating calls to GetEnvironment, validate and get config in the same step.
+func (o *initPipelineOpts) validateEnvsAndGetConfigs() error {
```
dannyrandall, line 231: it looks like this function and `getEnvConfigs()` are the same, could we remove one?

dannyrandall: it looks like this function and `getEnvConfigs()` are the same, could we remove one?

huanjani: 🙈 I need to slow down! Changed it up a little; just made the storing of configs parts of the other methods.

dannyrandall (approved):
lgtm! :)

PR #3324, opened 2022-03-03
chore(cli): init workflow should be compatible with new ws app validation

iamhopaul123:

### DIFF
--- a/internal/pkg/cli/init.go
+++ b/internal/pkg/cli/init.go
@@ -187,6 +187,10 @@ func newInitOpts(vars initVars) (*initOpts, error) {
 	}
 	fs := &afero.Afero{Fs: afero.NewOsFs()}
 	cmd := exec.NewCmd()
+	wsAppName := aws.String(tryReadingAppName())
+	if aws.StringValue(wsAppName) == "" {
+		wsAppName = nil
+	}
 	return &initOpts{
 		initVars:     vars,
 		ShouldDeploy: vars.shouldDeploy,
@@ -228,6 +232,7 @@ func newInitOpts(vars initVars) (*initOpts, error) {
 					prompt:       prompt,
 					mftReader:    ws,
 					dockerEngine: dockerengine.New(cmd),
+					wsAppName:    wsAppName,
 					initParser: func(s string) dockerfileParser {
 						return dockerfile.New(fs, s)
 					},
@@ -251,7 +256,7 @@ func newInitOpts(vars initVars) (*initOpts, error) {
 					mftReader:    ws,
 					prompt:       prompt,
 					dockerEngine: dockerengine.New(cmd),
-					wsAppName:    tryReadingAppName(),
+					wsAppName:    wsAppName,
 				}
 				opts.dockerfile = func(path string) dockerfileParser {
 					if opts.df != nil {
--- a/internal/pkg/cli/init.go
+++ b/internal/pkg/cli/init.go
@@ -187,10 +187,6 @@ func newInitOpts(vars initVars) (*initOpts, error) {
 	}
 	fs := &afero.Afero{Fs: afero.NewOsFs()}
 	cmd := exec.NewCmd()
-	wsAppName := aws.String(tryReadingAppName())
-	if aws.StringValue(wsAppName) == "" {
-		wsAppName = nil
-	}
 	return &initOpts{
 		initVars:     vars,
 		ShouldDeploy: vars.shouldDeploy,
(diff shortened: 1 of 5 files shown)


internal/pkg/cli/svc_init.go
```diff
@@ -564,6 +559,23 @@ func (o *initSvcOpts) askSvcPublishers() (err error) {
...
+	if wsApp == "" {
+		// NOTE: This command is required to be executed under a workspace. We don't prompt for it.
+		return "", errNoAppInWorkspace
+	}
+	var appName string
+	// This command must be run within the app's workspace.
+	if inputApp != "" && inputApp != wsApp {
+		return "", fmt.Errorf("cannot specify app %s because the workspace is already registered with app %s", inputApp, wsApp)
+	}
+	appName = wsApp
```
dannyrandall: can we change this to `appName := wsApp` and remove the declaration above?

dannyrandall: or I guess maybe we can just get rid of `appName` entirely and just use `wsApp`?

internal/pkg/cli/init.go
```diff
@@ -187,6 +187,10 @@ func newInitOpts(vars initVars) (*initOpts, error) {
 	}
 	fs := &afero.Afero{Fs: afero.NewOsFs()}
 	cmd := exec.NewCmd()
+	wsAppName := aws.String(tryReadingAppName())
+	if aws.StringValue(wsAppName) == "" {
+		wsAppName = nil
```
efekarakus: Hmm I agree that this is the most convenient but it's not clear 🤔 

What if we did `initOpts.initEnvCmd.wsAppName = tryReadingAppName()` in the CLI before calling `Ask` or `Execute`?

internal/pkg/cli/svc_init.go
```diff
@@ -564,6 +559,23 @@ func (o *initSvcOpts) askSvcPublishers() (err error) {
 	return nil
 }
 
+func validateInputApp(wsApp, inputApp string, store store) (string, error) {
```
efekarakus: The string return is odd here, I'd expect `validateInputApp` to return only an error.
And then the client can set `o.appName = wsApp`

internal/pkg/cli/init.go
Lou1415926: This probably doesn't work because service init's `Ask` happens before app init's `Execute` 🤔 , at which time the workspace is pending creation but not really created yet.

internal/pkg/cli/job_init.go
```diff
@@ -71,8 +71,7 @@ type initJobOpts struct {
 	manifestPath string
 	platform     *manifest.PlatformString
 
-	// Cache variables
-	wsAppName string
+	wsAppName *string // Cache variable for local app name. Empty if there's no app. Nil if pending creation.
```
Lou1415926: I see we want to be able to communicate this message "the workspace will have been created by the time `Execute` is run; have faith in it and don't validate".

What do you think of introducing a var named "pendingWSCreation" instead of aggregating the state by introducing a pointer? I have no preference but the first option mayyy seem clearer.

internal/pkg/cli/init.go
iamhopaul123: I think I might need to put more comments there specifying `wsAppName` is a pointer string with three valid states (nil, empty and nonempty)

internal/pkg/cli/svc_init.go
iamhopaul123: You are right! I had some codes between before but will do thanks

internal/pkg/cli/init.go
Lou1415926: Or
```go
var wsAppName *string
if ws := tryReadingAppName(); ws != "" {
    wsAppName = aws.String(ws)
} // NOTE: otherwise, pass `nil` to signal that the ws is pending creation.
```

internal/pkg/cli/job_init.go
iamhopaul123: The reason why i use the same variable has another reason: if we make it two variables then we set either one of them for initiation whereas if we keep one then whenever we init the opts we always set `wsAppName`. But I agree this is not exactly the same as our manifest pointer field as the `nil` value doesn't really make sense without further explanation, which harms the readability

internal/pkg/cli/init.go
iamhopaul123: Ok i took wx's suggestion and split into two vars.

efekarakus: I still think both solutions (a `nil` pointer or a `bool`) is solving the problem in the reverse direction. We're making changes to `job init` and `svc init` so that it works well with `init`, ideally the commands shouldn't have to be adapted for a specific client.

Would the problem go away if `init` had it's own `Validate` and `Ask` methods without depending on any commands? So that we can load up all questions we care about and then create the workload commands objects.

efekarakus (approved):
:shipit: 

Writing here our discussion offline for spreading knowledge:
1. Given that the workflow for `copilot init` is different: 
   1. ask app name
   2. ask svc name // Problem.
   3. init app
   4. init service
   5. init env
   6. deploy service
  We agree that it makes sense to have a separate `Ask` method for the `init` cmd.
  
2. However, doing that change now would mean that we will have to duplicate code across both `svc init` and `init`. For example, if we want to change the validation logic for `svc init`, then we'll have to also update `init`.  
The long term solution to this problem is to have validation logic in _functions_ instead of methods, so that they can be shared across commands.  
But that's a larger change that is not worthwhile for now.

Lou1415926 (approved):

PR #3326, opened 2022-03-03
feat: check if service is available in the region

### LINKED ISSUE #3172: Check whether AppRunner is available in the region
I ran into a strange error message when trying to deploy a request driven web service in the Frankfurt (eu-central-1) region.

`Template format error: Unrecognized resource types: [AWS::AppRunner::Service]`

After some digging I found out that AppRunner isn't available in this region yet. It would be nice if copilot-cli could check whether a template is valid for the region and issue a warning.

paragbhingre:
This PR adds ability to check if service exist in the given target environment region for every `deploy` command.  

This PR resolves #3172


paragbhingre:
Oops some test errors, while I check that you can still take a look at implementation 😁

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -327,3 +329,7 @@ func (e *Alias) ToString() string {
 	}
 	return strings.Join(e.StringSlice, ",")
 }
+
+func (s *LoadBalancedWebService) IsServiceAvailableInRegion(region string) (bool, error) {
+	return regions.IsServiceAvailable(ecs.EndpointsID, region)
```
efekarakus: Can we move the constants to our `internal/aws` pkgs for example:
```go
// in internal/pkg/aws/ecs
package ecs

const ServiceID = ecs.ServiceID
```

We want to do this because it makes it easier to upgrade a dependency (aws-sdk-go) from v1 to v2. This way we have a single location to update `internal/pkg/aws` instead of updating multiple pkgs in the codebase. Hope that makes sense!

internal/pkg/aws/regions/regions.go
```diff
@@ -0,0 +1,22 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package regions
```
efekarakus: 🥺 https://github.com/golang/go/wiki/CodeReviewComments#package-comments

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -327,3 +329,7 @@ func (e *Alias) ToString() string {
 	}
 	return strings.Join(e.StringSlice, ",")
 }
+
+func (s *LoadBalancedWebService) IsServiceAvailableInRegion(region string) (bool, error) {
```
efekarakus: 🥺 https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

efekarakus (commented):
I haven't looked deeply to the PR, but at a glance it looks great!  
Just some minor feedback, feel free to not wait for my approval if others approve.

internal/pkg/manifest/lb_web_svc.go
efekarakus: (also I believe it needs to be `.ServiceID` instead of `.EndpointsID`)

Lou1415926: I believe `EndpointsID` is used here because of this description in the [aws-sdk-go doc](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#Endpoint.ServiceID):
> Deprecated: Use client package's EndpointsID value instead of these ServiceIDs. These IDs are not maintained, and are out of date.

Very confusing description since the parameter names are still `ServiceID` 🤔

Lou1415926: I went ahead and tested it - 
`apprunner` package's `ServiceID` is `"AppRunner"`, and its `EndpointsID` is `apprunner`. The `ServiceID` when passed into ` endpoints.RegionsForService` doesn't work, while the latter does.

internal/pkg/cli/svc_deploy.go
```diff
@@ -336,6 +340,16 @@ func workloadManifest(in *workloadManifestInput) (interface{}, error) {
 	if err != nil {
 		return nil, fmt.Errorf("apply environment %s override: %s", in.envName, err)
 	}
+	serviceInRegion, err := mft.IsServiceAvailableInRegion(in.targetEnv.Region)
+	if err != nil {
+		return nil, fmt.Errorf("%s available for region %s: %w ", in.svcType, in.targetEnv.Region, err)
```
Lou1415926: nit:
```suggestion
		return nil, fmt.Errorf("check if %s is available in region %s: %w ", in.svcType, in.targetEnv.Region, err)
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -336,6 +340,16 @@ func workloadManifest(in *workloadManifestInput) (interface{}, error) {
...
+	}
+
+	if !serviceInRegion {
+		log.Warningf(`%s might not be available in region %s, proceed with caution
```
Lou1415926: nit:
```suggestion
		log.Warningf(`%s might not be available in region %s, proceed with caution.
```

internal/pkg/manifest/backend_svc_test.go
```diff
@@ -905,3 +905,41 @@ func TestBackendSvc_ApplyEnv_CountOverrides(t *testing.T) {
...
+
+			// WHEN
+			got, gotErr := manifest.IsServiceAvailableInRegion(tc.region)
+
+			// THEN
+			if gotErr != nil {
+				require.EqualError(t, gotErr, tc.wantErr.Error())
+			} else {
+				require.Equal(t, tc.want, got)
+			}
```
Lou1415926:
```suggestion
			if tc.wantErr != nil {
				require.EqualError(t, gotErr, tc.wantErr.Error())
			} else { 
			        require.NoError(t, gotErr)
				require.Equal(t, tc.want, got)
			}
```

internal/pkg/manifest/backend_svc_test.go
```diff
@@ -905,3 +905,41 @@ func TestBackendSvc_ApplyEnv_CountOverrides(t *testing.T) {
...
+	testCases := map[string]struct {
+		in      *BackendService
+		region  string
+		want    bool
+		wantErr error
+	}{
+		"region exist": {
+			region:  "us-west-1",
+			want:    true,
+			wantErr: nil,
```
Lou1415926: nit: probably redundant since `wantErr` is initialized as `nil`.
```suggestion
```

internal/pkg/aws/regions/regions.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/copilot-cli/internal/pkg/aws/partitions"
+)
+
+// IsServiceAvailable returns true if the service ID is available in the given region.
+func IsServiceAvailable(sid string, region string) (bool, error) {
```
Lou1415926: Can we have unit tests for this function?

huanjani: nits:
```suggestion
func IsAvailableInRegion(sID string, region string) (bool, error) {
```
Here and below: can we make the function names more declarations than questions, for consistency, and because they're returning a booleans?

internal/pkg/aws/regions/regions.go
```diff
@@ -0,0 +1,22 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package regions
+
+import (
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/copilot-cli/internal/pkg/aws/partitions"
```
Lou1415926: This package is dependent on `internal/pkg/aws/partitions` - I think ideally the dependencies of `internal/pkg/aws` should be external package only.  Moreover, `partitions` package also rely on `aws/endpoints` package. So I think it probably makes sense to merge this `regions` package with `partitions` package

internal/pkg/manifest/backend_svc_test.go
Lou1415926: It's actually not clear what these unit tests in `manifest` package are testing 🤔 These tests seems to be testing `regions.IsServiceAvailable`, instead of `manifest`'s `IsServiceAvailable`

internal/pkg/cli/svc_deploy.go
huanjani: nit:
```suggestion
		return nil, fmt.Errorf("check %s availability for region %s: %w ", in.svcType, in.targetEnv.Region, err)
```

huanjani (commented):
This feature is going to be so helpful for our users! A couple small comments before I look again more deeply.

internal/pkg/cli/svc_deploy.go
huanjani: Haha, I was commenting on this at the same time as @Lou1415926. Her suggestion is better!

internal/pkg/cli/svc_deploy.go
huanjani: another tiny nit (for here and elsewhere):
```suggestion
		log.Warningf(`%s might not be available in region %s; proceed with caution
```

internal/pkg/manifest/lb_web_svc.go
paragbhingre: Yes even I did that testing and ServiceID did not work where as EndpintsID worked, hence used EndpointsID finally.

internal/pkg/manifest/backend_svc_test.go
paragbhingre: Yeah right, first I thought it is important to have a unit test for each kind of a service for example loadbalanced service or backend service and etc. But then I realized it is enough to write a test for `regions.IsServiceAvailable` with only two scenarios i.e. ecs service and apprunner service. So I have removed tests from other `service_test` files and have only one test for `regions.IsServiceAvailable`
Let me know if that makes sense?

internal/pkg/aws/regions/regions.go
paragbhingre: Done as per above comment

internal/pkg/aws/apprunner/apprunner.go
```diff
@@ -30,6 +30,7 @@ const (
 	// App Runner ImageRepositoryTypes
 	repositoryTypeECR       = "ECR"
 	repositoryTypeECRPublic = "ECR_PUBLIC"
+	EndpointsID             = apprunner.EndpointsID
```
Lou1415926: What do you think of moving these `EndpointsID` to `aws/partitions` so that all the endpoints are defined in a single package for easier maintenance?

```go
# in aws/partitions

const (
   AppRunnerEndpointsID = apprunner.EndpointsID
   ECSEndpointsID = ecs.EndpointsID
)

# in manifest/backend_svc.go
// IsServiceAvailableInRegion checks if service type exist in the given region.
func (s *BackendService) IsServiceAvailableInRegion(region string) (bool, error) {
	return partitions.IsAvailableInRegion(partitions.ECSEndpointsID, region)
}
```

iamhopaul123: Comment for public var please 🙏

internal/pkg/aws/partitions/partition.go
```diff
@@ -19,3 +19,16 @@ func (r Region) Partition() (endpoints.Partition, error) {
 	}
 	return partition, nil
 }
+
+// IsServiceAvailable returns true if the service ID is available in the given region.
```
iamhopaul123:
```suggestion
// IsAvailableInRegion returns true if the service ID is available in the given region.
```

internal/pkg/aws/partitions/partition_test.go
```diff
@@ -35,3 +35,45 @@ func TestRegion_Partition(t *testing.T) {
...
+func TestRegion_IsAvailableInRegion(t *testing.T) {
+	testCases := map[string]struct {
+		sID       string
+		region    string
+		want      bool
+		wantedErr error
+	}{
+		"ecs service exist in the given region": {
+			region: "us-west-2",
+			sID:    "ecs",
```
Lou1415926: nit: maybe -
```suggestion
			sID:   ECSEndpointsID,
```

internal/pkg/aws/partitions/partition_test.go
```diff
@@ -35,3 +35,45 @@ func TestRegion_Partition(t *testing.T) {
...
+			want:   true,
+		},
+		"ecs service does not exist in the given region": {
+			region: "us-west-3",
+			sID:    "ecs",
+			want:   false,
+		},
+		"apprunner service exist in the given region": {
+			region: "us-west-2",
+			sID:    "ecs",
```
Lou1415926: nit:
```suggestion
			sID:    AppRunnerEndpointsID,
```

internal/pkg/aws/partitions/partition.go
```diff
@@ -19,3 +19,16 @@ func (r Region) Partition() (endpoints.Partition, error) {
...
+func IsAvailableInRegion(sID string, region string) (bool, error) {
+	partition, err := Region(region).Partition()
+	if err != nil {
+		return false, err
+	}
+	regions, _ := endpoints.RegionsForService(endpoints.DefaultPartitions(), partition.ID(), sID)
+	if _, exist := regions[region]; exist {
+		return true, nil
+	}
+	return false, nil
```
iamhopaul123:
```suggestion
	regionServices, existInPartition := endpoints.RegionsForService(endpoints.DefaultPartitions(), partition.ID(), sID)
	if !existInPartition {
	  return false, nil
	}
	_, existInRegion := regionServices[region]
	return existInRegion, nil
```
In case if the service is not available in the partition.

internal/pkg/cli/svc_deploy.go
```diff
@@ -336,6 +340,16 @@ func workloadManifest(in *workloadManifestInput) (interface{}, error) {
 	if err != nil {
 		return nil, fmt.Errorf("apply environment %s override: %s", in.envName, err)
 	}
+	serviceInRegion, err := mft.IsServiceAvailableInRegion(in.targetEnv.Region)
```
iamhopaul123: I feel like `IsServiceAvailableInRegion` this should be a method of our [deployer](https://github.com/aws/copilot-cli/blob/183c7302ca142997d618e91bec9e7947b03aacd2/internal/pkg/cli/svc_deploy.go#L115) rather than the manifest struct 🤔

Lou1415926: +1 to this!

internal/pkg/cli/svc_package.go
```diff
@@ -296,6 +296,8 @@ func (o *packageSvcOpts) getSvcTemplates(env *config.Environment) (*wkldCfnTempl
 		interpolator: o.newInterpolator(o.appName, o.envName),
 		ws:           o.ws,
 		unmarshal:    o.unmarshal,
+		targetEnv:    o.targetEnv,
+		svcType:      "",
```
Lou1415926: If we move `IsServiceAvailableInRegion` to the `deployer` we probably wouldn't need to pass around `svcType` anymore

internal/pkg/aws/partitions/partition_test.go
```diff
@@ -45,22 +45,22 @@ func TestRegion_IsAvailableInRegion(t *testing.T) {
...
 			want:   false,
 		},
 		"apprunner service exist in the given region": {
 			region: "us-west-2",
-			sID:    "ecs",
+			sID:    AppRunnerEndpointsID,
 			want:   true,
 		},
 		"apprunner service does not exist in the given region": {
 			region: "us-west-3",
```
Lou1415926: Let's put `us-west-1` for this test because it exists & app runner is not available in this region 👀

iamhopaul123 (approved):

internal/pkg/aws/partitions/partition.go
```diff
@@ -19,3 +18,17 @@ func (r Region) Partition() (endpoints.Partition, error) {
...
+func IsAvailableInRegion(sID string, region string) (bool, error) {
+	partition, err := Region(region).Partition()
+	if err != nil {
+		return false, err
+	}
+	regionServices, existInPartition := endpoints.RegionsForService(endpoints.DefaultPartitions(), partition.ID(), sID)
+	if !existInPartition {
+		return false, nil
+	}
+	_, existInRegion := regionServices[region]
```
Lou1415926: tiny tiny nit: after checking the doc I feel like these var names may better descriptions ⬇️ 
```suggestion
	regions, partitionOrServiceExists := endpoints.RegionsForService(endpoints.DefaultPartitions(), partition.ID(), sID)
	if !partitionOrServiceExists {
		return false, nil
	}
	_, existInRegion := regions[region]
```

internal/pkg/cli/job_deploy.go
```diff
@@ -159,6 +159,15 @@ func (o *deployJobOpts) Execute() error {
...
 	}
+	serviceInRegion, err := deployer.IsServiceAvailableInRegion(o.targetEnv.Region)
+	if err != nil {
+		return fmt.Errorf("check if Scheduled Job(s) is available in region %s: %w", o.targetEnv.Region, err)
+	}
+
+	if !serviceInRegion {
+		log.Warningf(`Scheduled Job(s) might not be available in region %s; proceed with caution.
+`, o.targetEnv.Region)
+	}
```
Lou1415926: What do you think of checking service availability before `deployer.UploadArtifacts`? This way we error out much quicker.

internal/pkg/cli/job_deploy.go
```diff
@@ -159,6 +159,15 @@ func (o *deployJobOpts) Execute() error {
 	if err != nil {
 		return fmt.Errorf("upload deploy resources for job %s: %w", o.name, err)
 	}
+	serviceInRegion, err := deployer.IsServiceAvailableInRegion(o.targetEnv.Region)
+	if err != nil {
+		return fmt.Errorf("check if Scheduled Job(s) is available in region %s: %w", o.targetEnv.Region, err)
+	}
+
+	if !serviceInRegion {
+		log.Warningf(`Scheduled Job(s) might not be available in region %s; proceed with caution.
```
Lou1415926:
```suggestion
		log.Warningf(`Scheduled Job might not be available in region %s; proceed with caution.
```
I think it's okay without `(s)` because "Scheduled Job" here refers to a type, not a group of workloads.

internal/pkg/cli/svc_deploy.go
```diff
@@ -188,6 +188,15 @@ func (o *deploySvcOpts) Execute() error {
...
 	}
+	serviceInRegion, err := deployer.IsServiceAvailableInRegion(o.targetEnv.Region)
+	if err != nil {
+		return fmt.Errorf("check if %s is available in region %s: %w", o.svcType, o.targetEnv.Region, err)
+	}
+
+	if !serviceInRegion {
+		log.Warningf(`%s might not be available in region %s; proceed with caution.
+`, o.svcType, o.targetEnv.Region)
+	}
```
Lou1415926, line 200: Samesiess

Lou1415926 (approved):
Awesome feature! Feel free to remove the DNM label when you are ready.

PR #3328, opened 2022-03-04
chore(cli): change Validate/Ask roles for `pipeline show`

huanjani:
This command assumes the workspace appName but allows overriding in order to show non-local app pipeline info.

### DIFF
--- a/internal/pkg/cli/pipeline_show.go
+++ b/internal/pkg/cli/pipeline_show.go
@@ -99,11 +99,12 @@ func (o *showPipelineOpts) Validate() error {
 // Ask prompts for fields that are required but not passed in, and validates those that are.
 func (o *showPipelineOpts) Ask() error {
 	if o.name != "" {
+		// Validate the passed-in pipeline name.
 		if _, err := o.codepipeline.GetPipeline(o.name); err != nil {
-			return err
+			return fmt.Errorf("validate pipeline name %s: %w", o.name, err)
 		}
 	} else {
-		// Validate presence of appName in order to fetch pipelines.
+		// Check presence of appName in order to fetch pipelines.
 		if o.appName == "" {
 			if err := o.askAppName(); err != nil {
 				return err
@@ -113,6 +114,31 @@ func (o *showPipelineOpts) Ask() error {
 			return err
 		}
 	}
+	return nil
+}
+
+// Execute shows details about the pipeline.
+func (o *showPipelineOpts) Execute() error {
+	err := o.initDescriber(o.shouldOutputResources)
+	if err != nil {
+		return err
+	}
+
+	pipeline, err := o.describer.Describe()
+	if err != nil {
+		return fmt.Errorf("describe pipeline %s: %w", o.name, err)
+	}
+
+	if o.shouldOutputJSON {
+		data, err := pipeline.JSONString()
+		if err != nil {
+			return err
+		}
+		fmt.Fprint(o.w, data)
+	} else {
+		fmt.Fprint(o.w, pipeline.HumanString())
+	}
+
 	return nil
 }
 
@@ -133,7 +159,7 @@ func (o *showPipelineOpts) askPipelineName() error {
 	}
 
 	if len(pipelineNames) == 0 {
-		log.Infof("No deployed pipelines found for application %s.\n", color.HighlightUserInput(o.appName))
+		return fmt.Errorf("No deployed pipelines found for application %s.\n", color.HighlightUserInput(o.appName))
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


internal/pkg/cli/pipeline_show.go
```diff
@@ -87,34 +87,35 @@ func newShowPipelineOpts(vars showPipelineVars) (*showPipelineOpts, error) {
...
 	if o.appName != "" {
 		if _, err := o.store.GetApplication(o.appName); err != nil {
-			return err
+			return fmt.Errorf("validate application name: %w", err)
 		}
-	}
-	if o.pipelineName != "" {
-		if _, err := o.pipelineSvc.GetPipeline(o.pipelineName); err != nil {
+	} else {
+		if err := o.askAppName(); err != nil {
```
efekarakus: should we move this to `Ask` ? I feel like if we're going to prompt for it that means that the flag is required 💭

huanjani: Yeah, you're right. I was considering it optional bc we `tryReadingAppName()`, but it didn't sit well with me because yeah-- it's not optional in that we do need it to proceed. Moving.

huanjani: Actually, this is part of a larger bug with `pipeline show` and `pipeline status` that I've been wanting to fix for a long time. I'll add changes to this PR. If there is a pipeline manifest in the workspace from which the command is run, we show info/status for that workspace pipeline regardless of `o.appName`. That's because we first read the manifest for the pipeline name/other info, independently of the app name. We only use the app name if there isn't a detected pipeline manifest, at which point we retrieve pipeline names with the app name tag.

internal/pkg/cli/pipeline_show.go
```diff
@@ -199,7 +195,7 @@ func (o *showPipelineOpts) getPipelineNameFromManifest() (string, error) {
 
 // Execute shows details about the pipeline.
 func (o *showPipelineOpts) Execute() error {
-	if o.pipelineName == "" {
+	if o.name == "" {
```
Lou1415926: Is it true that `o.name` is either passed via the flag or prompted for? If so this seems like a dead if block.

internal/pkg/cli/pipeline_show.go
Lou1415926: Just a callout here: it seems like `appName` is only needed when we ask for pipeline name (we use `appName` to list out all the pipelines of the app). It however doesn't play a role in `Execute`. For this reason, I wouldn't consider `appName` "required". 

We should probably ask for `app` only when `o.name == ""` in `Ask`.

huanjani: Okay. I moved validation of passed-in appName back into `Validate()`, and only ask for `appName` if we're also asking for `name`. 

The code no longer reads the manifest for the pipeline name, because `show` and `status` are only concerned about deployed pipelines. Manually tested to confirm a much better experience.
Previously, if the user passed a flag for a different app, the pipeline information for the local pipeline would be returned instead because we were just reading the pipeline manifest. If there were no deployed pipelines, we would log a message saying no deployed pipelines were found, but then error out with cryptic error messaging when trying to run `Describe()`.

internal/pkg/cli/pipeline_show.go
```diff
@@ -87,34 +86,63 @@ func newShowPipelineOpts(vars showPipelineVars) (*showPipelineOpts, error) {
...
+}
+
+// Ask prompts for fields that are required but not passed in, and validates those that are.
+func (o *showPipelineOpts) Ask() error {
+	if o.name != "" {
+		// Validate the passed-in pipeline name.
+		if _, err := o.codepipeline.GetPipeline(o.name); err != nil {
+			return fmt.Errorf("validate pipeline name %s: %w", o.name, err)
+		}
+	} else {
```
Lou1415926: tiny nit: perhaps in order to remove the `else` block --
```go
if o.name != "" {
  _, err := o.codepipeline.GetPipeline(o.name)
  if err != nil {
	return fmt.Errorf("validate pipeline name %s: %w", o.name, err)
   }
   return nil
}

// Check presence of appName in order to fetch pipelines.
if o.appName == "" {
 // ...
}
return o.askPipelineName()
```

internal/pkg/cli/pipeline_show.go
```diff
@@ -87,34 +86,63 @@ func newShowPipelineOpts(vars showPipelineVars) (*showPipelineOpts, error) {
 	return opts, nil
 }
 
-// Validate returns an error if the flag values passed by the user are invalid.
+// Validate returns an error if the optional flag values passed by the user are invalid.
 func (o *showPipelineOpts) Validate() error {
 	if o.appName != "" {
 		if _, err := o.store.GetApplication(o.appName); err != nil {
-			return err
+			return fmt.Errorf("validate application name: %w", err)
```
Lou1415926, line 93: Interesting! Just to call this out - consider `copilot pipeline show --app non-existing-app --name my-pipeline`. We only need `appName` when we need to ask for pipeline name. So in this case,  the `--app` flag isn't really needed. Then the question is, should we still validate app?

The way it is here is to still validate the app. I think this is a nice way to sort of "remind" user that they may be thinking about a wrong app / wrong pipeline and I agree with the current approach. 

While for the other commands we do `validateOrAskAppName` in `Ask` instead of splitting the validation and ask, splitting them this way actually makes more sense for `pipeline show` because of the special nature of `--app` flag here.

Lou1415926 (approved):

efekarakus (approved):

Issue #2848, opened 2021-09-17
Suggestion: default network placement to be private


Fodoj:
Right now Copilot deploys tasks into a public subnet by default - which is a bit unexpected, at least for me. Given that there is already a publicly facing ALB, thus each task public ip and making it publicly-routable brings no benefits and is rather a best practice - unless I am missing something. What is the reason for this default and shouldn't copilot place tasks in private subnet instead?

iamhopaul123:
Hello @Fodoj. That is a great question! The main reason is because we think most of the users want their tasks to have public internet access by default. Based on that assumption we have cost concern, since if we place tasks in private subnets, customers would need to have NAT gateway in order to have public internet access for their tasks.

To control the traffic, we create security groups by default, allowing only certain inbound traffic to your service (aka from LB and same environment security group). so your service wont be publicly-routable.

Also right now it is possible to place tasks in private subnets:
- If you create an environment with Copilot generated VPC, then placement: 'private' will result in the creation of a NAT gateway.
- If you create an environment with an imported VPC, then placement: 'private' won't create anything. It will only place the tasks in private subnets. (you could create the VPC endpoints yourself for example, in order to have access to ECR to pull images)

We're in the future planning to introduce a new option: `placement: 'isolated'` to create the VPC endpoints if using a Copilot generated VPC. But no one had asked for them until now, so we didn't ship with it in our initial release.

Fodoj:
Thanks for clarification. Indeed, with current approach tasks are not accessible thanks to security group and private subnet will introduce more costs into the deployment - assuming backend service types, like workers, need to go to the Internet (and if they use AWS APIs.. they either do go there.. or people have to pay for VPC endpoints).

I always default to keeping as much as I can in private subnet in large environments, but for smaller setups benefits are not that obvious of course, given added complexity.

tekknolagi:
Is it possible to disable outbound network traffic from the container from either the Dockerfile or the manifest?

IgorEulalio:
> To control the traffic, we create security groups by default, allowing only certain inbound traffic to your service (aka from LB and same environment security group). so your service wont be publicly-routable.

Using NLB, the default behavior would be to add the public subnet CIDR where the NLB is placed in the task security group.

huanjani:
> Is it possible to disable outbound network traffic from the container from either the Dockerfile or the manifest?

@tekknolagi, apologies for the late reply! ~~No, it's not possible~~; may I ask what your use case is? Thanks!

[...]

PR #3330, opened 2022-03-04
docs: Updated Environments Concept Page

csantos:
- Changed verbiage in VPC and Networking section to explain workloads are placed in public subnet as a cost optimization.

Related to #2848 . Updated Copilot docs Concepts | Environments page explaining why Copilot places workloads in public subnet. While not changing Copilot's behavior, this lets users know why the default approach was chosen and how they can go about changing it if so desired.


site/content/docs/concepts/environments.en.md
```diff
@@ -29,9 +29,9 @@ When you first create a new environment, no services are deployed to it. To depl
 
 ### VPC and Networking
 
-Each environment gets its own multi-AZ VPC. Your VPC is the network boundary of your environment, allowing the traffic you expect in and out, and blocking the rest. The VPCs Copilot creates are spread across two availability zones to help balance availability and cost - with each AZ getting a public and private subnet.
+Each environment gets its own multi-AZ VPC. Your VPC is the network boundary of your environment, allowing the traffic you expect in and out, and blocking the rest. The VPCs Copilot creates are spread across two availability zones, with each AZ getting a public and private subnet, following [AWS best practices](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-security-best-practices.html). Your services are launched in the public subnets, to help manage costs. Services can be reached only through your load balancer.
```
efekarakus: I think we should mention something around security to ensure customers that while it's in public subnets it's still secure 😌.  
We forgot to update this blurb 🙈 , but it's also possible now to create services that are accessible via Service Discovery (`Backend Service`) or an SQS queue (`Worker Service`) so I think we can remove the last piece of the sentence.

```suggestion
Each environment gets its own multi-AZ VPC. Your VPC is the network boundary of your environment, allowing the traffic you expect in and out, and blocking the rest. The VPCs Copilot creates are spread across two availability zones, with each AZ getting a public and private subnet, following [AWS best practices](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-security-best-practices.html). Your services are launched by default in the public subnets while limiting ingress to only other services in your environment for security. Tasks are placed in public subnets to help manage costs by allowing egress to the internet without requiring a NAT gateway.
```

site/content/docs/concepts/environments.en.md
```diff
@@ -29,9 +29,9 @@ When you first create a new environment, no services are deployed to it. To depl
...
 
-Your services are launched in the public subnets but can be reached only through your load balancer.
+Workload subnet placement can be modified using the `network.vpc.placement` field in the AWS Copilot CLI manifest. For additional details, see this field in the [Backend Service](../manifest/backend-service.en.md), [Load Balanced Web Service](../manifest/lb-web-service.en.md), [Request-Driven Web Service](../manifest/rd-web-service.en.md), [Scheduled Job](../manifest/scheduled-job.en.md), or [Worker Service](../manifest/worker-service.en.md) manifest specifications.
```
efekarakus: nit: It's probably okay to just link to one of them so that we don't have to remember updating this section.
```suggestion
Workload subnet placement can be modified using the [`network.vpc.placement`](../manifest/lb-web-service.en.md#network-vpc-placement) field in the manifest.
```

efekarakus (commented):
yay thank you so much Carlos!

csantos (commented):
Thanks for the suggestions! I've gone ahead and committed them.

huanjani (approved):
Thanks so much for doing this!

efekarakus (approved):

Issue #3308, opened 2022-02-24
Support `FilterPolicy` in worker service


Lou1415926:
Copilot should provide a native way for user to specify subscription filter policy for their worker service. 

Related: discussion on [Gitter](https://gitter.im/aws/copilot-cli?at=6217a61ac435002500c104f2)

lsnape:
To dig into the use case, we have a CPU- and memory-intensive worker that has a min task count of 0. There can be several hours without events, so we'd like to not run the task when there's no work to do.

We have multiple workers subscribed to the same events topic. Not all workers are interested in every event on that topic. We could circumvent this by configuring the upstream producer write to multiple topics, but that doesn't take advantage of the decoupling between topics and subscriptions.

rfma23:
It would be really useful to have a way of specifying a FilterPolicy in a similar way as in AWS SAM, something along the lines of:
```
FilterPolicy:
  fare:
    - numeric:
        - '>='
        - 50
  distance:
    - numeric:
        - '>='
        - 20
```

iamhopaul123:
## Proposal

Thanks @rfma23 for giving this brilliant idea. To put everything together, we have a new field in our manifest for each topic as

``` yaml
filter_policy: json/yaml
```

For example:
```yaml
subscribe:
  topics:
    - name: events
      service: api
      filter_policy: {"store":["example_corp"],"event":[{"anything-but":"order_cancelled"}],"customer_interests":["rugby","football","baseball"],"price_usd":[{"numeric":[">=",100]}]}
      queue: # Define a topic-specific queue for the api-events topic.
        timeout: 20s
    - name: events
      service: fe
      filter_policy:
        store:
          - example_corp
        event:
          - order_cancelled
        encrypted:
          - false
        customer_interests:
          - basketball
          - baseball
  queue: # By default, messages from all topics will go to a shared queue.
    timeout: 45s
    retention: 96h
    delay: 30s
```

Referred to https://docs.aws.amazon.com/sns/latest/dg/example-filter-policies.html

rfma23:
> ## Proposal
> Thanks @rfma23 for giving this brilliant idea. To put everything together, we have a new field in our manifest for each topic as
> 
> ```yaml
> filter_policy: json/yaml
> ```
> 
> For example:
> 
> ```yaml
> subscribe:
>   topics:
>     - name: events
>       service: api
[...]

PR #3338, opened 2022-03-08
feat: support FilterPolicy in worker service

iamhopaul123:
Addresses #3308 


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -658,26 +658,48 @@ func convertSubscribe(s manifest.SubscribeConfig, accountID, region, app, env, s
...
+			return nil, err
+		}
 		subscriptions.Topics = append(subscriptions.Topics, ts)
 	}
 	subscriptions.Queue = convertQueue(s.Queue)
 	return &subscriptions, nil
 }
 
-func convertTopicSubscription(t manifest.TopicSubscription, url, accountID, app, env, svc string) *template.TopicSubscription {
+func convertTopicSubscription(t manifest.TopicSubscription, url, accountID, app, env, svc string) (
```
efekarakus: What do you think of passing a pointer instead since a `TopicSubscription` can be a fairly large object?
```suggestion
func convertTopicSubscription(sub *manifest.TopicSubscription) (
```
I can't tell if it's my lack of sleep 😅 but the remaining parameters here don't seem to be used in the body of the fn

internal/pkg/deploy/cloudformation/stack/transformers_test.go
```diff
@@ -1329,6 +1332,9 @@ func Test_convertSubscribe(t *testing.T) {
 					{
 						Name:    aws.String("name"),
 						Service: aws.String("svc"),
+						FilterPolicy: manifest.FilterPolicy{
+							String: aws.String(`{"store": ["example_corp"]}`),
```
efekarakus: Is `String`-form really a usecase that people want? Should we limit the release to only `map[string]interface{}` instead?
Like if users want to write 
```yaml
filter_policy: {"store":["example_corp"]}
```
this is equivalent to in YAML:
```yaml
filter_policy:
  store:
     - example_corp
```
But I don't know why anyone would want to write:
```yaml
filter_policy: '{"store":["example_corp"]}'
```

internal/pkg/manifest/workload.go
```diff
@@ -263,6 +264,41 @@ func (c *CommandOverride) ToStringSlice() ([]string, error) {
 	return out, nil
 }
 
+type stringOrInterface struct {
+	String    *string
+	Interface map[string]interface{}
```
efekarakus: If we have just:
```
type FilterPolicy map[string]interface{}
```
Will the marshaling to JSON work if the level of nesting in the filter policy is more than 1-level? https://stackoverflow.com/a/40737676/1201381

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -658,26 +658,48 @@ func convertSubscribe(s manifest.SubscribeConfig, accountID, region, app, env, s
...
+	*template.TopicSubscription, error) {
```
Lou1415926: super tiny nit: maybe  
```go
func convertTopicSubscription(t manifest.TopicSubscription, url, accountID, app, env, svc string) 
(*template.TopicSubscription, error) {
```

or 

```go
func convertTopicSubscription(t manifest.TopicSubscription, url, accountID, app, env, svc string (*template.TopicSubscription, error) {
```

Edit: ignore this this is outdated now

internal/pkg/template/workload.go
```diff
@@ -385,9 +385,10 @@ func (s *SubscribeOpts) HasTopicQueues() bool {
 
 // TopicSubscription holds information needed to render a SNS Topic Subscription in a container definition.
 type TopicSubscription struct {
-	Name    *string
-	Service *string
-	Queue   *SQSQueue
+	Name         *string
+	Service      *string
+	FilterPolicy *string
```
Lou1415926, line 390: Just to call it out here: I think we are probably fine with `string` instead of `*string` for a lot of these variables because `""` being a string of length of 0 will result in `false` when evaluated in  a `if` statement

internal/pkg/deploy/cloudformation/stack/transformers.go
iamhopaul123: lol sorry obviously i my lack is worse :(

internal/pkg/deploy/cloudformation/stack/transformers_test.go
iamhopaul123: I was debating myself on that too. The only reason why I add this is I think `filter_policy` is supposed to be a `string` in CFN and people might be surprised to see `'{"store":["example_corp"]}'` won't go through. Do you think we should remove for simplicity?

internal/pkg/manifest/workload.go
iamhopaul123: Yes it will i think https://github.com/aws/copilot-cli/pull/3338/files#diff-82d4b46cc8502909deb33c1730687e7b95ba23df615f068807c2889d833db702R35-R47

internal/pkg/deploy/cloudformation/stack/transformers_test.go
efekarakus: yeah I think we should remove it, we can always add it later on but we won't be able to remove it once shipped

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -648,36 +648,53 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
...
+		FilterPolicy: filterPolicy,
+	}, nil
+}
+
+func convertFilterPolicy(filterPolicy map[string]interface{}) (*string, error) {
+	bytes, err := json.Marshal(filterPolicy)
+	if err != nil {
+		return nil, fmt.Errorf(`convert "filter_policy" to a JSON string: %w`, err)
+	}
+	if string(bytes) == "null" {
```
efekarakus: wut lol 😂 

can we instead add a guard clause at the top of the function: 
```go
if filterPolicy == nil {
   return nil, nil 
}
```

efekarakus (approved):

internal/pkg/deploy/cloudformation/stack/transformers.go
iamhopaul123: it's in their codebase lol

iamhopaul123: hmm unfortunately we have to do that without pointer because of the mergo issue. maybe i can check the length of the map

Lou1415926 (approved):

PR #3339, opened 2022-03-08
chore(cli): change roles of Ask and Validate for `pipeline status`

huanjani:


internal/pkg/cli/pipeline_status.go
```diff
@@ -84,25 +83,30 @@ func newPipelineStatusOpts(vars pipelineStatusVars) (*pipelineStatusOpts, error)
...
 func (o *pipelineStatusOpts) Ask() error {
-	if err := o.askAppName(); err != nil {
-		return err
+	if o.name != "" {
+		_, err := o.codepipeline.GetPipeline(o.name)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
```
iamhopaul123:
```suggestion
	if o.name != "" {
		if _, err := o.codepipeline.GetPipeline(o.name); err != nil {
			return err
		}
		return nil
	} else {
	  if err := o.askPipelineName(); err != nil {
	    return err
	  }
	}
```

internal/pkg/cli/pipeline_status.go
```diff
@@ -144,22 +145,6 @@ func (o *pipelineStatusOpts) askAppName() error {
 }
 
 func (o *pipelineStatusOpts) askPipelineName() error {
```
Lou1415926, line 145: Not to address here but perhaps latter some day we could consider moving the parameter from the receiver to function arg if all of these `pipeline` commands have the same `askPipelineName`. E.g. [here](https://github.com/aws/copilot-cli/blob/76e100c2fb7955125cada30bfcd5d9bd8ed269e6/internal/pkg/cli/svc_deploy.go#L338)

internal/pkg/cli/pipeline_status.go
```diff
@@ -144,22 +145,6 @@ func (o *pipelineStatusOpts) askAppName() error {
...
-	pipelineName, err := o.getPipelineNameFromManifest()
-	if err == nil {
-		o.pipelineName = pipelineName
-		return nil
-	}
-
-	if errors.Is(err, workspace.ErrNoPipelineInWorkspace) {
-		log.Infof("No pipeline manifest in workspace for application %s, looking for deployed pipelines.\n", color.HighlightUserInput(o.appName))
-	}
-
```
iamhopaul123, line 162: Awesome!

Lou1415926 (approved):

internal/pkg/cli/pipeline_status.go
huanjani: Are there benefits to doing it this way? I'd had it like this for `pipeline show`, but agreed with feedback that it may be nicer to get rid of the `else`: https://github.com/aws/copilot-cli/pull/3328#discussion_r820907333

What do you think?

iamhopaul123: Oh i misread the code. I agree we can remove the `else` but it doesn't make sense to me to split the pipeline name flag check and the ask. Maybe we should check the app name first then the pipeline name.

huanjani: Oh, we did that because we only need the app name if we're asking for the pipeline name. If they pass in the pipeline name, we don't even need the app name. https://github.com/aws/copilot-cli/pull/3328#discussion_r819997029

huanjani: discussed offline. changed code for both `pipeline show` and `pipeline status` to require app name regardless of pipeline flag presence/absence.

internal/pkg/cli/pipeline_status_test.go
```diff
@@ -7,175 +7,100 @@ import (
...
+	store        *mocks.Mockstore
+	ws           *mocks.MockwsPipelineReader
+	prompt       *mocks.Mockprompter
+	codepipeline *mocks.MockpipelineGetter
+	describer    *mocks.Mockdescriber
+	sel          *mocks.MockappSelector
 }
 
-func TestPipelineStatus_Validate(t *testing.T) {
+func TestPipelineStatus_Ask(t *testing.T) {
```
iamhopaul123, line 28: Do we not need any unit test for validate 🤔

iamhopaul123 (approved):

internal/pkg/cli/pipeline_status_test.go
huanjani: Validate() is just returning nil, nothing to test! :)

PR #3332, opened 2022-03-04
feat: support disabling automatic rollback on `svc deploy`

Lou1415926:
Address:
#3156
#2811 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the Apache 2.0 License.

PR #3341, opened 2022-03-08
feat: support disabling automatic rollback on `job deploy` and `deploy`

Lou1415926:
This PR includes:
1. Change as titled
2. Refactor: removing passing of `forceDeploy` from job to deploy
3. Refactor: remove dead code
4. Refactor: avoid passing params twice (in receiver and in func param) to `newSvcDeployer` and `newJobDeployer`
5. Doc update

This is a follow-up from #3332 


internal/pkg/cli/job_deploy.go
```diff
@@ -168,9 +179,18 @@ func (o *deployJobOpts) Execute() error {
 			Tags:        tags.Merge(o.targetApp.Tags, o.resourceTags),
 		},
 		Options: deploy.Options{
-			ForceNewUpdate: o.forceNewUpdate,
```
efekarakus, line 171: lol was this dead code? where `foreceNewUpdate` is always `false` because the flag doesn't exist?

internal/pkg/cli/job_deploy.go
```diff
@@ -168,9 +179,18 @@ func (o *deployJobOpts) Execute() error {
...
+			DisableRollback: o.disableRollback,
 		},
 	}); err != nil {
+		if o.disableRollback {
+			stackName := stack.NameForService(o.targetApp.Name, o.targetEnv.Name, o.name)
+			rollbackCmd := fmt.Sprintf("aws cloudformation rollback-stack --stack-name %s --role-arn %s", stackName, o.targetEnv.ExecutionRoleARN)
+			log.Infof(`It seems like you have disabled automatic stack rollback for this deployment. To debug, you can visit the AWS console to inspect the errors.
```
efekarakus, line 179: are we printing this log statement twice, can we remove the one at l. 164?

efekarakus, line 179: are we printing this log statement twice, can we remove the one at l. 164?

internal/pkg/cli/job_deploy.go
Lou1415926: The true reason it is deemed "dead code" is because `forceNewUpdate` isn't used for `job` in `cli/deploy`.  Even when it's passed in as true, we don't force new update to the job.

In addition - yes the flag is not supported for `copilot job deploy`. But you could run `copilot deploy --forceNewUpdate` and the flag would be passed to job. Either way it took zero effect because it's not actually supported in `cli/deploy`.

internal/pkg/cli/job_deploy.go
Lou1415926: Thanks for catching this!! It was a big typo :homer-disappearing

efekarakus (approved):

dannyrandall (approved):
lgtm

PR #3352, opened 2022-03-11
fix: respect network replacement configured through manifest

### LINKED ISSUE #3346: Environment overrides for `network` are overwriting base definitions.
I'm am seeing this with `network:`. I am not sure if this is happening with other maps.

In comparison, this isn't merging in the way that I would expect helm would merge: Maps/keys are merged at all levels. Values/arrays are overwritten.

Is everything getting overridden at the map of `network:`?

This doesn't place the task in the private subnet:
```yaml
network:
  vpc:
    placement: private

environments:
  dev:
    network:
      vpc:
        security_groups:
          - sg-1234567890
```

This does place the task in the private subnet:
```yaml
environments:
  dev:
    network:
      vpc:
        placement: private
        security_groups:
          - sg-1234567890
```

edit: Thanks!

Lou1415926:
This PR fixes #3346 by:
1. Remove the custom unmarshal logic of `network` field. The logic preset the network's placement value, which causes the bug that's seen in #3346.
2. Use `const` instead of `var` for placement types.

### DIFF
--- a/internal/pkg/manifest/workload.go
+++ b/internal/pkg/manifest/workload.go
@@ -395,7 +395,7 @@ func (c *NetworkConfig) UnmarshalYAML(value *yaml.Node) error {
 	if err := value.Decode(&conf); err != nil {
 		return err
 	}
-	if conf.VPC.isEmpty() { // If after unmarshaling the user did not specify VPC configuration then reset it to public.
+	if conf.VPC.isEmpty() { // If after unmarshalling the user did not specify VPC configuration then reset it to public.
 		conf.VPC = defaultVPCConf
 	}
 	*c = NetworkConfig(conf)
--- a/internal/pkg/manifest/workload.go
+++ b/internal/pkg/manifest/workload.go
@@ -24,12 +24,12 @@ const (
 
 const (
 	// AWS VPC subnet placement options.
-	PublicSubnetPlacement  = "public"
-	PrivateSubnetPlacement = "private"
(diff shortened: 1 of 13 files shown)


internal/pkg/manifest/workload.go
```diff
@@ -565,3 +544,11 @@ func uint16P(n uint16) *uint16 {
 	}
 	return &n
 }
+
```
iamhopaul123, line 546: comment pls

internal/pkg/manifest/workload.go
```diff
@@ -23,14 +23,14 @@ const (
 )
 
 // AWS VPC subnet placement options.
```
iamhopaul123: maybe move this one below the const?

iamhopaul123 (approved):

internal/pkg/manifest/workload.go
```diff
@@ -552,6 +531,15 @@ func DockerfileBuildRequired(svc interface{}) (bool, error) {
 	return required, nil
 }
 
+// PlacementP converts a string to a `Placement` type and returns its pointer.
+func PlacementP(p string) *Placement {
```
efekarakus: What do you think of instead the following approach:
```go
package manifest

// Placement represents VPC subnet placement.
type Placement string

const (
   PublicSubnetPlacement Placement    = "public"
   PrivateSubnetPlacement Placement = "private"
)
```

efekarakus (approved):

PR #3360, opened 2022-03-14
chore(pipeline): ensure appName not empty for `pipeline deploy` commands

huanjani:
Previously, `tryReadingAppName()` was the default appName value at the cmd.Flags() level.
When we moved that functionality, we introduced a bug:
In the constructors for `SvcListCmd` and `JobListCmd`, we have `appName: vars.appName` and those were empty because we were calling `tryReadingAppName()` later.
This change 1. passes `o.appName` in to those two methods later and 2. gets the appConfig in the same call as validating appName.


dannyrandall (approved):
lgtm

iamhopaul123 (approved):

PR #3365, opened 2022-03-15
feat: task run subnets and envs

### LINKED ISSUE #3100: `task run` should allow additional security group configuration
Currently, with `--app` and `--env` specified, `copilot task run` automatically uses the environment security group for the task. However, user won't be able to configure additional security groups in addition to the environment security group, because  the `--security-groups` flag  is mutually exclusive with `--app` and `--env`.

### Example Use Case
User may want to run a task with the same configuration as one of their services. The environment security group, however, is not necessarily the only security group that their service is using. For example, if the service has an Aurora addon, it will also have an Aurora security group to access the DB.

The task should be able to use this security group as well, in addition to the environment security group.

### Proposal
When user specify `--app`, `--env` and `--security-groups` at the same time, we don't error out. Instead, we append the security groups specified by `--security-groups` to the environment security group.

### Related Issue
In #3094, user wants to run the task in the environment's account, with additional security groups and subnets configuration. Without being able to use `--security-groups` with `--app` and `--env`, they have to switch the default profile to the environment profile before running the `copilot task run` command with `--security-groups`, `--subnets` and `--cluster` flags.

With this proposal, (assuming they don't need additional subnets), they could have run it with `--app`, `--env` and `--security-groups` flags. This way, Copilot can automatically assume the environment manager role, and they wouldn't have to switch the profile.

### LINKED ISSUE #3103: `task run --generate-cmd` workflow
Currently, user can run `copilot task run --generate-cmd <app>/<env>/<svc>` to generate a `task run` command with an existing service's configuration. Then, they'd copy & paste and run the generated command to spin up the tasks.

The issue with this workflow is that the tasks will be spun up in the default profile's account, which is likely the application account, while user probably expect them to be in the environment's account (see #3094). Copilot should be able to provide a better workflow.

Here are two directions from which the improvement can be made:
1. Currently, the command is always generated with the `--subnets`, `--security-groups` and `--cluster` flags. Instead, Copilot could spit out a command that uses the `--app` and `--env` flag (potentially `--security-groups` flag as well - see #3100 )  if the argument is `<app>/<env>/<svc>`. This way, Copilot knows the task is meant to be placed in an environment, thus using the environment manager role to perform any operation instead of the default profile.
2. Copilot could still spit out the command with the `--subnets`, `--security-groups` and `--cluster` flags, just like what it currently does. However, it could prepend two commands `AWS_PROFILE=default; AWS_REGION=<env_region>;`. If Copilot detects that the environment account is different than the application account, it prints to `stderr` to ask user to change `AWS_PROFILE=default` to the environment profile. This way, Copilot makes sure that the user is aware that they may need to switch profile before spinning up the task. (Why `stderr`? Because otherwise user wouldn't receive any message if they had run `$(copilot task run --generate-cmd ...)`)

codekitchen:
feat: allow task run --subnets with --app and --env

Fixes #3100

feat: task --generate-cmd outputs --app and --env

Fixes #3103

### DIFF
--- a/internal/pkg/ecs/run_task_request_test.go
+++ b/internal/pkg/ecs/run_task_request_test.go
@@ -209,17 +209,18 @@ func Test_RunTaskRequestFromService(t *testing.T) {
 					Subnets:        []string{"sbn-1", "sbn-2"},
 					SecurityGroups: []string{"sg-1", "sg-2"},
 				}, nil)
-				m.EXPECT().ClusterARN(testApp, testEnv).Return("kamura-village", nil)
 			},
 			wantedRunTaskRequest: &RunTaskRequest{
 				networkConfiguration: ecs.NetworkConfiguration{
 					AssignPublicIp: "1.2.3.4",
-					Subnets:        []string{"sbn-1", "sbn-2"},
 					SecurityGroups: []string{"sg-1", "sg-2"},
 				},
 
 				executionRole: "execution-role",
 				taskRole:      "task-role",
+
+				appName: testApp,
+				envName: testEnv,
 
 				containerInfo: containerInfo{
 					image:      "beautiful-image",
@@ -233,8 +234,6 @@ func Test_RunTaskRequestFromService(t *testing.T) {
 						"truth": "go-ask-the-wise",
 					},
 				},
-
-				cluster: "kamura-village",
 			},
 		},
 		"unable to retrieve task definition": {
@@ -252,14 +251,6 @@ func Test_RunTaskRequestFromService(t *testing.T) {
 				m.EXPECT().ClusterARN(gomock.Any(), gomock.Any()).AnyTimes()
 			},
 			wantedError: errors.New("retrieve network configuration for service svc: some error"),
-		},
-		"unable to obtain cluster ARN": {
-			setUpMock: func(m *mocks.MockServiceDescriber) {
-				m.EXPECT().TaskDefinition(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
-				m.EXPECT().NetworkConfiguration(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
-				m.EXPECT().ClusterARN(testApp, testEnv).Return("", errors.New("some error"))
-			},
-			wantedError: errors.New("retrieve cluster ARN created for environment env in application app: some error"),
 		},
 	}
 
@@ -332,17 +323,18 @@ func Test_RunTaskRequestFromJob(t *testing.T) {
 					Subnets:        []string{"sbn-1", "sbn-2"},
 					SecurityGroups: []string{"sg-1", "sg-2"},
 				}, nil)
-				m.EXPECT().ClusterARN(testApp, testEnv).Return("kamura-village", nil)
 			},
 			wantedRunTaskRequest: &RunTaskRequest{
 				networkConfiguration: ecs.NetworkConfiguration{
 					AssignPublicIp: "1.2.3.4",
-					Subnets:        []string{"sbn-1", "sbn-2"},
 					SecurityGroups: []string{"sg-1", "sg-2"},
 				},
 
 				executionRole: "execution-role",
 				taskRole:      "task-role",
+
+				appName: testApp,
+				envName: testEnv,
 
 				containerInfo: containerInfo{
 					image:      "beautiful-image",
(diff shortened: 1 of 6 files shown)


codekitchen:
@Lou1415926 here are the changes we discussed on those two issues

I've run the unit and integration tests. I haven't gotten set up to run the e2e tests yet. I did build a new binary and used it against our app to verify that these changes allow us to run predeploy migrations using this new workflow as expected!

internal/pkg/ecs/run_task_request_test.go
```diff
@@ -209,18 +209,19 @@ func Test_RunTaskRequestFromService(t *testing.T) {
...
 				networkConfiguration: ecs.NetworkConfiguration{
 					AssignPublicIp: "1.2.3.4",
-					Subnets:        []string{"sbn-1", "sbn-2"},
 					SecurityGroups: []string{"sg-1", "sg-2"},
 				},
 
 				executionRole: "execution-role",
 				taskRole:      "task-role",
 
+				appName: testApp,
```
efekarakus, line 224: It just occurred to me that we don't have any unit tests for the `CLIString` method of `RunTaskRequest`.

Would it be possible to add one of to test the new scenario in the PR:
```go
func TestRunTaskRequest_CLIString(t *testing.T) {
		in     RunTaskRequest
		wanted string
	}{
		"generates cmd when --app and --env is provided with additional --security-groups": {},
	}
}
```

internal/pkg/task/env_runner.go
```diff
@@ -76,6 +79,7 @@ func (r *EnvRunner) Run() ([]*Task, error) {
 	if err != nil {
 		return nil, fmt.Errorf(fmtErrSecurityGroupsFromEnv, r.Env, err)
 	}
+	securityGroups = append(securityGroups, r.SecurityGroups...)
```
efekarakus: What do you think about checking the number of security groups given the limit of 5?
```go
if total := len(securityGroups) + len(r.SecurityGroups); total > 5 {
   return  nil, fmt.Errorf("unable to run task with more than 5 security groups: (%d) %s", total, strings.Join(append(securityGroups, r.SecurityGroups...), ","))
}
securityGroups = append(securityGroups, r.SecurityGroups...)
```

efekarakus (commented):
Looks great to me! Thank you so much ❤️ Apologies, for the late review.

Just a small feedback for an error case

internal/pkg/task/env_runner_test.go
```diff
@@ -194,6 +195,35 @@ func TestEnvRunner_Run(t *testing.T) {
 				},
 			},
 		},
+		"run in env with extra security groups success": {
```
Lou1415926, line 209: Nice! Thank you for writing this unit test!

internal/pkg/ecs/run_task_request_test.go
codekitchen: I was wondering about that gap myself, I'd be happy to add a test yeah

internal/pkg/task/env_runner.go
codekitchen: oh good call. I was also thinking that maybe we should dedup the list of SGs here, since I think `--generate-cmd` output is going to include the same core security group that `--app` and `--env` are going to include based on running inside a copilot environment.

codekitchen:
no worries, I'll make those tweaks soon

Lou1415926 (commented):
Thank you very much ❤️  I don't have anything to add on top of what Efe has said. This looks good to me!

codekitchen:
I've made those changes, thanks for the feedback. Let me know if you have a more preferred way to do the unique string append, I didn't see anything obvious in existing code or dependencies. I bet in another couple golang releases we'll have some generic functions in stdlib for this stuff, I look forward to it.

internal/pkg/task/env_runner.go
```diff
@@ -135,3 +140,21 @@ func (r *EnvRunner) validateDependencies() error {
 
 	return nil
 }
+
+func appendUniqueStrings(s1 []string, s2 ...string) []string {
```
efekarakus, line 144: this is totally fine, in case you're curious Go has also a little entry on how to achieve it: https://github.com/golang/go/wiki/SliceTricks#in-place-deduplicate-comparable

I'd lean to that implementation, but this is totally fine too!

efekarakus (approved):

internal/pkg/task/env_runner.go
codekitchen: that's a good page to bookmark, thanks!

Lou1415926 (approved):
LGTM - thank you! :shipit:

Issue #3366, opened 2022-03-15
Proposal: display deployed status of workloads in app show


paragbhingre:
Currently `app show` displays only name and type of the workload (i.e. services and jobs). 

```
copilot app show

Workloads

  Name             Type
  ----             ----
  apprunner        Request-Driven Web Service
  firstjob         Scheduled Job

```

After we add the functionality to display deployed status on workload it will look like

```
copilot app show

Workloads

  Name             Type                         Deployed Status
  ----             ----                         ----
  apprunner        Request-Driven Web Service   Deployed
  firstjob         Scheduled Job                Not Deployed

```

Lou1415926:
What do folks think about something like this
```
Workloads

  Name             Type                         Deployed to
  ----             ----                         ----
  apprunner        Request-Driven Web Service   prod, test
  firstjob         Scheduled Job                test
```
so that it's clear what env's the workloads' deployed to?

huanjani:
I was about to say the same thing. Maybe mention "environment(s)" in the heading to be even clearer.

PR #3379, opened 2022-03-17
feat: update app show to display if workload is deployed

paragbhingre:
This PR adds new environments column to the workloads table in `app show` command displaying the environments to which workloads are deployed to.  

This PR fixes the issue #3366


paragbhingre:
Please ignore test failures and observe the commented code as I have open this PR for your 1st glance so that we can decide which path to choose to display environments in workloads table. Please let me know your thoughts on the implementation. :)

internal/pkg/cli/app_show.go
```diff
@@ -121,6 +129,45 @@ func (o *showAppOpts) description() (*describe.App, error) {
...
+		for _, job := range deployedJobs {
+			workloadEnvs[job] = append(workloadEnvs[job], env.Name)
+		}
+		deployedSvcs, err := o.deployStoreClient.ListDeployedServices(o.name, env.Name)
+		if err != nil {
+			return nil, fmt.Errorf("list deployed services %s: %w", o.name, err)
+		}
+		for _, svc := range deployedSvcs {
+			workloadEnvs[svc] = append(workloadEnvs[svc], env.Name)
+		}
```
dannyrandall: This seems like a great use case for [`errgroup.Group`](https://pkg.go.dev/golang.org/x/sync/errgroup) to get all of the jobs/services in parallel. I'd be curious to see how much it would slow down listing a single environment and how much it would speed up with multiple environments.

internal/pkg/describe/app.go
```diff
@@ -64,14 +65,14 @@ func (a *App) HumanString() string {
 	}
 	fmt.Fprint(writer, color.Bold.Sprint("\nWorkloads\n\n"))
 	writer.Flush()
-	headers = []string{"Name", "Type"}
+	headers = []string{"Name", "Type", "Environments"}
 	fmt.Fprintf(writer, "  %s\n", strings.Join(headers, "\t"))
 	fmt.Fprintf(writer, "  %s\n", strings.Join(underline(headers), "\t"))
 	for _, svc := range a.Services {
-		fmt.Fprintf(writer, "  %s\t%s\n", svc.Name, svc.Type)
+		fmt.Fprintf(writer, "  %s\t%s\t%s\n", svc.Name, svc.Type, strings.Join(a.WorkloadEnvs[svc.Name], ","))
```
huanjani: nit: 
```suggestion
		fmt.Fprintf(writer, "  %s\t%s\t%s\n", svc.Name, svc.Type, strings.Join(a.WorkloadEnvs[svc.Name], ", "))
```

internal/pkg/describe/app.go
```diff
@@ -64,14 +65,14 @@ func (a *App) HumanString() string {
...
 	}
 	for _, job := range a.Jobs {
-		fmt.Fprintf(writer, "  %s\t%s\n", job.Name, job.Type)
+		fmt.Fprintf(writer, "  %s\t%s\t%s\n", job.Name, job.Type, strings.Join(a.WorkloadEnvs[job.Name], ","))
```
huanjani: same nit as above:
```suggestion
		fmt.Fprintf(writer, "  %s\t%s\t%s\n", job.Name, job.Type, strings.Join(a.WorkloadEnvs[job.Name], ", "))
```

huanjani:
> Please ignore test failures and observe the commented code as I have open this PR for your 1st glance so that we can decide which path to choose to display environments in workloads table. Please let me know your thoughts on the implementation. :)

I like the first option! 😄

internal/pkg/cli/app_show.go
```diff
@@ -121,6 +129,45 @@ func (o *showAppOpts) description() (*describe.App, error) {
...
+		for _, svc := range svcs {
+			flag, err := o.deployStoreSvc.IsServiceDeployed(o.name, env.Name, svc.Name)
+			if err != nil {
+				return nil, fmt.Errorf("services not available %s: %w", o.name, err)
+			}
+			if flag {
+				workloadEnvs[svc.Name] = append(workloadEnvs[svc.Name], env.Name)
+			}
+		}
+	}*/
```
Lou1415926: Let's ditch these for now 😄

internal/pkg/describe/app.go
```diff
@@ -23,13 +23,14 @@ import (
...
-	Jobs      []*config.Workload       `json:"jobs"`
-	Pipelines []*codepipeline.Pipeline `json:"pipelines"`
+	Name         string                   `json:"name"`
+	Version      string                   `json:"version"`
+	URI          string                   `json:"uri"`
+	Envs         []*config.Environment    `json:"environments"`
+	Services     []*config.Workload       `json:"services"`
+	Jobs         []*config.Workload       `json:"jobs"`
+	Pipelines    []*codepipeline.Pipeline `json:"pipelines"`
+	WorkloadEnvs map[string][]string      `json:"workloadenvs"`
```
Lou1415926: Do we want these information in JSON output?

My take is that we probably want to just show them in HumanString for now, and not JSON.

internal/pkg/cli/app_show_test.go
```diff
@@ -332,10 +346,10 @@ Environments
...
 Workloads
 
-  Name    Type
-  ----    ----
-  my-svc  lb-web-svc
-  my-job  Scheduled Job
+  Name    Type           Environments
+  ----    ----           ------------
+  my-svc  lb-web-svc     test,prod
+  my-job  Scheduled Job  test,prod
```
Lou1415926: Let's also include an example where the service is not deployed at all to see if the table still makes sense in that case.

internal/pkg/cli/app_show.go
```diff
@@ -36,25 +36,32 @@ type showAppVars struct {
...
-	defaultSession, err := sessions.ImmutableProvider(sessions.UserAgentExtras("app show")).Default()
+	sessProvider := sessions.ImmutableProvider(sessions.UserAgentExtras("app show"))
+	defaultSession, err := sessProvider.Default()
 	if err != nil {
 		return nil, fmt.Errorf("default session: %w", err)
 	}
 	store := config.NewSSMStore(identity.New(defaultSession), ssm.New(defaultSession), aws.StringValue(defaultSession.Config.Region))
+	deployStoreClient, err := deploy.NewStore(sessProvider, store)
+	if err != nil {
+		return nil, fmt.Errorf("deploy store client: %w", err)
```
Lou1415926: nit: to be consistent on error message wrapping:
```suggestion
		return nil, fmt.Errorf("connect to deploy store: %w", err)
```

internal/pkg/cli/app_show.go
```diff
@@ -36,25 +36,32 @@ type showAppVars struct {
...
-	store            store
-	w                io.Writer
-	sel              appSelector
-	pipelineSvc      pipelineGetter
-	newVersionGetter func(string) (versionGetter, error)
+	store             store
+	w                 io.Writer
+	sel               appSelector
+	pipelineSvc       pipelineGetter
+	deployStoreClient selector.DeployStoreClient
```
Lou1415926: ummm why we are using the `selector` package?

Lou1415926: ohh I see you are using the interface defined in `selector`. I think we should use the interface defined in `cli` package instead.

internal/pkg/describe/app.go
paragbhingre: Made sure that Marshal ignores this field for JSON output using `json"-"`

internal/pkg/cli/app_show_test.go
paragbhingre: added test case for the services which are not being deployed yet

internal/pkg/cli/app_show.go
paragbhingre: Using interface from `cli` package now

internal/pkg/cli/app_show.go
```diff
@@ -121,6 +129,23 @@ func (o *showAppOpts) description() (*describe.App, error) {
 	if err != nil {
 		return nil, fmt.Errorf("list jobs in application %s: %w", o.name, err)
 	}
+	//1st approach takes 2 seconds to display the table
```
Lou1415926: nit:
```suggestion
```

internal/pkg/cli/app_show.go
```diff
@@ -105,6 +112,7 @@ func (o *showAppOpts) Execute() error {
 }
 
 func (o *showAppOpts) description() (*describe.App, error) {
+	workloadEnvs := make(map[string][]string)
```
Lou1415926: nit: we can perhaps use a more descriptive name to this map, for example `envsDeployedTo`, `wkldDeployedToEnv`

internal/pkg/cli/app_show_test.go
```diff
@@ -332,10 +345,81 @@ Environments
...
+  Name    AccountID  Region
+  ----    ---------  ------
+  test    123456789  us-west-2
+  prod    123456789  us-west-1
+
+Workloads
+
+  Name    Type           Environments
+  ----    ----           ------------
+  my-svc  lb-web-svc
```
Lou1415926: What do folks think about having "Not Deployed" here instead of leaving them blank?

It might not be clear to user whether the blank means "we were't able to fetch the info" or "the service isn't deployed"

Lou1415926 (commented):
LGTM! Thank you for the unit tests!

I have mostly just nits and a suggestion/though on `Environments` column when the service isn't deployed at all. 

Also I think we should probably consider moving `description` from `app_show.go` to `describe/app.go`. Not the scope of this PR though, just a call out!

internal/pkg/cli/app_show.go
paragbhingre: wkldDeployedtoEnvs sounds good to me

internal/pkg/cli/app_show.go
```diff
@@ -121,6 +129,23 @@ func (o *showAppOpts) description() (*describe.App, error) {
 	if err != nil {
 		return nil, fmt.Errorf("list jobs in application %s: %w", o.name, err)
 	}
+	//1st approach takes 2 seconds to display the table
+	for _, env := range envs {
+		deployedJobs, err := o.deployStore.ListDeployedJobs(o.name, env.Name)
+		if err != nil {
+			return nil, fmt.Errorf("list deployed jobs %s: %w", o.name, err)
```
iamhopaul123: nit: i think it would be more helpful to show them what env are we checking, because the app name is very obvious...
```suggestion
			return nil, fmt.Errorf("list jobs deployed to %s: %w", env.Name, err)
```

internal/pkg/describe/app.go
```diff
@@ -23,13 +23,14 @@ import (
...
-	Jobs      []*config.Workload       `json:"jobs"`
-	Pipelines []*codepipeline.Pipeline `json:"pipelines"`
+	Name         string                   `json:"name"`
+	Version      string                   `json:"version"`
+	URI          string                   `json:"uri"`
+	Envs         []*config.Environment    `json:"environments"`
+	Services     []*config.Workload       `json:"services"`
+	Jobs         []*config.Workload       `json:"jobs"`
+	Pipelines    []*codepipeline.Pipeline `json:"pipelines"`
+	WorkloadEnvs map[string][]string      `json:"-"`
```
iamhopaul123: Hmm i think ideally the `json` format should show more info than the human output.

Lou1415926: I feel like we probably don't want to include the deployed env information now. It's a nice thing to show in human string, but I feel like we want to leave the door open to take it away.

iamhopaul123: Ok sounds good. Maybe we can add it in the future whenever either it is requested by customers or we need it from the other commands.

internal/pkg/cli/app_show_test.go
paragbhingre: Screen Shot 2022-03-22 at 4 43 16 PM

just want to show you how does it looks like after adding `Not Deployed`. It looks little crowded and makes it hard to read actual environments.

paragbhingre: As per discussion offline we are going with the following implementation - 
Screen Shot 2022-03-23 at 12 08 36 PM

internal/pkg/cli/app_show.go
paragbhingre: Great suggestion, implemented and reduced response time by 66% 🤩

internal/pkg/cli/app_show.go
```diff
@@ -112,7 +115,7 @@ func (o *showAppOpts) Execute() error {
 }
 
 func (o *showAppOpts) description() (*describe.App, error) {
-	workloadEnvs := make(map[string][]string)
+	wkldDeployedtoEnvs := make(map[string][]string)
```
dannyrandall: let's move this down by `var mux sync.Mutex` to make it clearer what `mux` is guarding

internal/pkg/cli/app_show.go
```diff
@@ -129,24 +132,45 @@ func (o *showAppOpts) description() (*describe.App, error) {
...
+		})
+	}
+	if err := g.Wait(); err != nil {
+		return nil, err
+	}
+	// Sort the map values so that `output` is consistent and the unit test won't be flaky.
+	for key, value := range wkldDeployedtoEnvs {
+		sort.Strings(value)
+		wkldDeployedtoEnvs[key] = value
+	}
```
dannyrandall: let's simplify this to:
```
for k := range wkldDeployedtoEnvs {
    sort.Strings(wkldDeployedtoEnvs[k])
}
```

internal/pkg/describe/app.go
```diff
@@ -69,10 +69,22 @@ func (a *App) HumanString() string {
 	fmt.Fprintf(writer, "  %s\n", strings.Join(headers, "\t"))
 	fmt.Fprintf(writer, "  %s\n", strings.Join(underline(headers), "\t"))
 	for _, svc := range a.Services {
-		fmt.Fprintf(writer, "  %s\t%s\t%s\n", svc.Name, svc.Type, strings.Join(a.WorkloadEnvs[svc.Name], ", "))
+		var envs string
+		if len(a.WkldDeployedtoEnvs[svc.Name]) > 0 {
+			envs = strings.Join(a.WkldDeployedtoEnvs[svc.Name], ", ")
+		} else {
+			envs = "-"
+		}
```
dannyrandall: nit:
```
envs := "-"
if len(a.WkldDeployedtoEnvs[svc.Name]) > 0 {
    envs = strings.Join(a.WkldDeployedtoEnvs[svc.Name], ", ")
}
```

dannyrandall (commented):
love the `errgroup`! awesome perf improvement😊

internal/pkg/cli/app_show.go
```diff
@@ -121,7 +132,45 @@ func (o *showAppOpts) description() (*describe.App, error) {
 	if err != nil {
 		return nil, fmt.Errorf("list jobs in application %s: %w", o.name, err)
 	}
+	var mux sync.Mutex
+	g, ctx := errgroup.WithContext(context.Background())
+	defer ctx.Done()
+	for i := range envs {
+		env := envs[i]
```
iamhopaul123, line 158:
```suggestion
	for _, env := range envs {
```

iamhopaul123 (commented):
Nice work for improving the performance! Thanks for diving deep into it. +1 to Danny's comments.

internal/pkg/cli/app_show.go
dannyrandall: this won't actually work (😢) - https://go.dev/doc/faq#closures_and_goroutines. I'm partial to the way Parag did it, but would be ok switching to 
```
for _, env := range envs {
    env := env
```

iamhopaul123: oooh that's right!

internal/pkg/cli/app_show.go
```diff
@@ -129,24 +132,45 @@ func (o *showAppOpts) description() (*describe.App, error) {
...
-		deployedSvcs, err := o.deployStore.ListDeployedServices(o.name, env.Name)
-		if err != nil {
-			return nil, fmt.Errorf("list deployed services %s: %w", o.name, err)
-		}
-		for _, svc := range deployedSvcs {
-			workloadEnvs[svc] = append(workloadEnvs[svc], env.Name)
-		}
-	}
+	var mux sync.Mutex
+	g, ctx := errgroup.WithContext(context.Background())
```
efekarakus: Should we use a context with a deadline to ensure we return a response back to users within 30s for example? https://pkg.go.dev/context#example-WithDeadline

internal/pkg/cli/app_show.go
paragbhingre: Actually, because we are now using errgroup it is making parallel calls to `listDeployedJobs` and `listDeployedServices`. In these parallel calls we are losing the context of `env` when we use it's value as `env.Name`. Hence we are using the `index` number to get the correct value of the environment and not the updated value in each context. Does it make sense?

internal/pkg/cli/app_show.go
```diff
@@ -129,24 +132,45 @@ func (o *showAppOpts) description() (*describe.App, error) {
...
+	defer ctx.Done()
+	for i := range envs {
+		env := envs[i]
+		g.Go(func() error {
```
efekarakus, line 159: What do y'all think of a format like this?
```go
g.Go(func() error { 
  return o.populateDeployedWorkloads(o.deployStore.ListDeployedJobs, deployedEnvsFor, env.Name) 
})
g.Go(func() error { 
   return o.populateDeployedWorkloads(o.deployStore.ListDeployedServices, deployedEnvsFor, env.Name) 
})

func (o *showAppOpts) populateDeployedWorkloads(
   listWorkloads func(app, env string) ([]string, error),  
   deployedEnvsFor map[string][]string, 
   env string) error {
  // ...
}

```

dannyrandall: I like that a lot! Just don't forget the mutex :)

internal/pkg/cli/app_show.go
```diff
@@ -5,18 +5,21 @@ package cli
...
-	"github.com/aws/copilot-cli/internal/pkg/aws/identity"
-
 	"github.com/aws/copilot-cli/internal/pkg/aws/codepipeline"
+	"github.com/aws/copilot-cli/internal/pkg/aws/identity"
 	"github.com/aws/copilot-cli/internal/pkg/aws/sessions"
 	"github.com/aws/copilot-cli/internal/pkg/config"
 	"github.com/aws/copilot-cli/internal/pkg/deploy"
 	"github.com/aws/copilot-cli/internal/pkg/term/prompt"
 	"github.com/aws/copilot-cli/internal/pkg/term/selector"
+	"golang.org/x/net/context"
```
efekarakus: This should be just "context", I believe this version was before it was promoted to the sdt lib
```suggestion
	"context"
```

internal/pkg/cli/app_show.go
```diff
@@ -103,6 +114,19 @@ func (o *showAppOpts) Execute() error {
 	fmt.Fprint(o.w, data)
 	return nil
 }
+func (o *showAppOpts) populateDeployedWorkloads(listWorkloads func(app, env string) ([]string, error), wkldDeployedtoEnvs map[string][]string, env string, mux *sync.Mutex) error {
```
efekarakus:
```suggestion
func (o *showAppOpts) populateDeployedWorkloads(listWorkloads func(app, env string) ([]string, error), deployedEnvsFor map[string][]string, env string, lock sync.Locker) error {
```

internal/pkg/cli/app_show.go
```diff
@@ -122,6 +146,28 @@ func (o *showAppOpts) description() (*describe.App, error) {
 		return nil, fmt.Errorf("list jobs in application %s: %w", o.name, err)
 	}
 
+	wkldDeployedtoEnvs := make(map[string][]string)
+	ctx, cancelWait := context.WithTimeout(context.Background(), waitForStackTimeout)
+	g, ctx := errgroup.WithContext(ctx)
+	defer cancelWait()
+	defer ctx.Done()
```
efekarakus:
```suggestion
	ctx, cancelWait := context.WithTimeout(context.Background(), waitForStackTimeout)
	defer cancelWait()
	g, ctx := errgroup.WithContext(ctx)
```

efekarakus (commented):
Looks good! few small changes

internal/pkg/cli/app_show.go
```diff
@@ -122,6 +146,28 @@ func (o *showAppOpts) description() (*describe.App, error) {
 		return nil, fmt.Errorf("list jobs in application %s: %w", o.name, err)
 	}
 
+	wkldDeployedtoEnvs := make(map[string][]string)
+	ctx, cancelWait := context.WithTimeout(context.Background(), waitForStackTimeout)
+	defer cancelWait()
+	g, ctx := errgroup.WithContext(ctx)
+	defer ctx.Done()
```
efekarakus: I think we can delete this 🤔, we don't use the derived context
```suggestion
	g, _ := errgroup.WithContext(ctx)
```

efekarakus (approved):

dannyrandall (approved):
lgtm! great job🎉

internal/pkg/cli/app_show.go
```diff
@@ -149,7 +149,7 @@ func (o *showAppOpts) description() (*describe.App, error) {
 	wkldDeployedtoEnvs := make(map[string][]string)
 	ctx, cancelWait := context.WithTimeout(context.Background(), waitForStackTimeout)
 	defer cancelWait()
-	g, ctx := errgroup.WithContext(ctx)
+	g, _ := errgroup.WithContext(ctx)
 	defer ctx.Done()
```
dannyrandall:
```suggestion
```

PR #3381, opened 2022-03-18
feat(pipelines): require pipeline name on `pipeline init`

dannyrandall:

### DIFF
--- a/internal/pkg/cli/interfaces.go
+++ b/internal/pkg/cli/interfaces.go
@@ -5,8 +5,9 @@ package cli
 
 import (
 	"encoding"
-	"github.com/aws/copilot-cli/internal/pkg/manifest"
 	"io"
+
+	"github.com/aws/copilot-cli/internal/pkg/manifest"
 
 	"github.com/aws/copilot-cli/internal/pkg/aws/ec2"
 
@@ -241,8 +242,8 @@ type wsPipelineManifestReader interface {
 }
 
 type wsPipelineWriter interface {
-	WritePipelineBuildspec(marshaler encoding.BinaryMarshaler) (string, error)
-	WritePipelineManifest(marshaler encoding.BinaryMarshaler) (string, error)
+	WritePipelineBuildspec(marshaler encoding.BinaryMarshaler, name string) (string, error)
+	WritePipelineManifest(marshaler encoding.BinaryMarshaler, name string) (string, error)
 }
 
 type serviceLister interface {
--- a/internal/pkg/cli/pipeline_init.go
+++ b/internal/pkg/cli/pipeline_init.go
@@ -43,7 +43,7 @@ import (
 
 const (
 	fmtPipelineInitNamePrompt  = "What would you like to %s this pipeline?"
-	pipelineInitNameHelpPrompt = "An identifier for your pipeline (e.g. beta, prod)."
+	pipelineInitNameHelpPrompt = "An identifier for your pipeline (e.g. release, test, prod)."
 
 	pipelineSelectEnvPrompt     = "Which environment would you like to add to your pipeline?"
 	pipelineSelectEnvHelpPrompt = "Adds an environment that corresponds to a deployment stage in your pipeline. Environments are added sequentially."
@@ -172,23 +172,11 @@ func (o *initPipelineOpts) Ask() error {
 	}
 	o.appName = o.wsAppName
 
-	if o.name == "" {
-		if err := o.askPipelineName(); err != nil {
-			return err
-		}
-	}
-	if err := validatePipelineName(o.name, o.appName); err != nil {
+	if err := o.askOrValidatePipelineName(); err != nil {
 		return err
 	}
 
-	// TODO: validate not a duplicate
-
-	if o.repoURL == "" {
-		if err := o.selectURL(); err != nil {
-			return err
-		}
-	}
-	if err := o.validateURL(o.repoURL); err != nil {
+	if err := o.askOrValidateURL(); err != nil {
 		return err
 	}
 
@@ -202,6 +190,22 @@ func (o *initPipelineOpts) Ask() error {
 	}
 
 	return nil
+}
+
+func (o *initPipelineOpts) askOrValidatePipelineName() error {
+	if o.name == "" {
+		return o.askPipelineName()
+	}
+
+	return validatePipelineName(o.name, o.appName)
+}
+
+func (o *initPipelineOpts) askOrValidateURL() error {
+	if o.repoURL == "" {
+		return o.selectURL()
+	}
+
+	return o.validateURL(o.repoURL)
 }
 
 // Execute writes the pipeline manifest file.
--- a/internal/pkg/cli/pipeline_init.go
+++ b/internal/pkg/cli/pipeline_init.go
@@ -42,6 +42,9 @@ import (
 )
 
 const (
+	fmtPipelineInitNamePrompt  = "What would you like to %s this pipeline?"
+	pipelineInitNameHelpPrompt = "An identifier for your pipeline (e.g. beta, prod)."
+
 	pipelineSelectEnvPrompt     = "Which environment would you like to add to your pipeline?"
 	pipelineSelectEnvHelpPrompt = "Adds an environment that corresponds to a deployment stage in your pipeline. Environments are added sequentially."
 
@@ -80,6 +83,7 @@ var (
 
 type initPipelineVars struct {
 	appName           string
+	name              string // Name of the pipeline
 	environments      []string
 	repoURL           string
 	repoBranch        string
@@ -166,26 +170,35 @@ func (o *initPipelineOpts) Validate() error {
 
 // Ask prompts for required fields that are not passed in and validates them.
 func (o *initPipelineOpts) Ask() error {
-	if o.repoURL != "" {
-		if err := o.validateURL(o.repoURL); err != nil {
+	if o.name == "" {
+		if err := o.askPipelineName(); err != nil {
 			return err
 		}
-	} else {
+	}
+	if err := validatePipelineName(o.name, o.appName); err != nil {
+		return err
+	}
+
+	// TODO: validate not a duplicate
+
+	if o.repoURL == "" {
 		if err := o.selectURL(); err != nil {
 			return err
 		}
 	}
+	if err := o.validateURL(o.repoURL); err != nil {
+		return err
+	}
 
-	if len(o.environments) > 0 {
-		if err := o.validateEnvs(); err != nil {
-			return err
-		}
-
-	} else {
+	if len(o.environments) == 0 {
 		if err := o.askEnvs(); err != nil {
 			return err
 		}
 	}
+	if err := o.validateEnvs(); err != nil {
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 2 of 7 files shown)


internal/pkg/cli/pipeline_init.go
```diff
@@ -166,26 +170,35 @@ func (o *initPipelineOpts) Validate() error {
```
huanjani, line 166: 🙏🏼  Would you mind moving `validateInputApp` from `Validate()` to `Ask()` please? 🙏🏼 
Seeing clearly, with some distance from #3322, that `appName` is not optional!

internal/pkg/cli/pipeline_init.go
```diff
@@ -166,26 +170,35 @@ func (o *initPipelineOpts) Validate() error {
...
-	if len(o.environments) > 0 {
-		if err := o.validateEnvs(); err != nil {
+	// TODO: validate not a duplicate
+
+	if o.repoURL == "" {
+		if err := o.selectURL(); err != nil {
 			return err
 		}
+	}
+	if err := o.validateURL(o.repoURL); err != nil {
```
huanjani: While this _is_ prettier than the `else`, in the other commands we were trying to avoid validating selected URLs because we already know they're valid. What do you think? 

(Same for pipelineName, above)

internal/pkg/cli/pipeline_init.go
```diff
@@ -166,26 +170,35 @@ func (o *initPipelineOpts) Validate() error {
...
+		return err
+	}
 
-	} else {
+	if len(o.environments) == 0 {
```
huanjani, line 183: Same here as above. Do we want to validate if we asked?

internal/pkg/cli/validate.go
```diff
@@ -217,6 +223,29 @@ func validateJobName(val interface{}) error {
 	return nil
 }
 
+func validatePipelineName(val interface{}, appName string) error {
+	// https://docs.aws.amazon.com/codepipeline/latest/userguide/limits.html
+
+	// compute the longest name a user can name their pipeline for this app
+	// since we prefix their name with 'pipeline-[app]-'
+	maxNameLen := maxPipelineNameLen - len(fmt.Sprintf(fmtPipelineName, appName, ""))
```
huanjani, line 231: ❤️

internal/pkg/cli/pipeline_init.go
```diff
@@ -545,13 +562,13 @@ func (o *initPipelineOpts) createPipelineManifest() error {
...
 
-	manifest, err := manifest.NewPipeline(pipelineName, provider, stages)
+	manifest, err := manifest.NewPipeline(o.name, provider, stages)
 	if err != nil {
 		return fmt.Errorf("generate a pipeline manifest: %w", err)
 	}
 
 	var manifestExists bool
-	manifestPath, err := o.workspace.WritePipelineManifest(manifest)
+	manifestPath, err := o.workspace.WritePipelineManifest(manifest, o.name)
```
huanjani, line 577: Question: do we need to pass in the name here? It should already be in the manifest after `manifest.NewPipeline`, right? Or is it for the `PipelineManifest` object?

internal/pkg/cli/pipeline_init.go
```diff
@@ -42,6 +42,9 @@ import (
 )
 
 const (
+	fmtPipelineInitNamePrompt  = "What would you like to %s this pipeline?"
```
huanjani, line 45: Nit: I'm seeing this enthusiastic "**name**" in `app init` but not in the other init commands. I think we should be consistent. I personally could do without it; what do you think?

internal/pkg/cli/pipeline_init.go
```diff
@@ -254,17 +282,8 @@ func (o *initPipelineOpts) askEnvs() error {
...
-	var envConfigs []*config.Environment
-	for _, environment := range o.environments {
-		envConfig, err := o.store.GetEnvironment(o.appName, environment)
-		if err != nil {
-			return fmt.Errorf("get config of environment %s: %w", environment, err)
-		}
-		envConfigs = append(envConfigs, envConfig)
-	}
-	o.envConfigs = envConfigs
+	o.environments = envs
```
huanjani, line 292: Thanks for taking out the duplicated envConfig code!

huanjani:
There is so much goodness in this PR! Just a few minor questions.

internal/pkg/cli/pipeline_init.go
dannyrandall: Yeah, we need it - [`WritePipelineManifest` takes a `encoding.BinaryMarshaler`](https://github.com/aws/copilot-cli/commit/f84e064aff6dd1fe0d1d82bb79c2dcfcb296bc39#diff-28e31e11ec2b0ddd8baf047d045914d541968c933e5bad45a2afed259be99551), so it doesn't know that `manifest` has a `Name` field. 

So we could change `wsPipelineWriter` to accept a `manifest.Pipeline` instead of the `BinaryMarshaler` to get rid of the name argument, which I would be fine doing. However, the [Service/Job manifest writing uses the same logic](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/workspace/workspace.go#L313-L328) as what's here currently. 

Thoughts?

internal/pkg/cli/pipeline_init.go
dannyrandall: This is a good call out....I've been looking around and it's looks like we're pretty inconsistent around this point. A few examples of highlighting that I found:
- [secret init](https://github.com/dannyrandall/copilot-cli/blob/baf44a7c3578efd5167a855a9a7a3faa39ece6bf/internal/pkg/cli/secret_init.go#L391)
- [job init](https://github.com/dannyrandall/copilot-cli/blob/0ca8e4d5638053867ae7c3cf7a2b1266a3e0ee75/internal/pkg/cli/job_init.go#L36)

So I'll leave it as is for now, but it's an inconsistency we should probably make a decision on and fix it to be the same everywhere.

internal/pkg/cli/pipeline_init.go
huanjani: Ah, gotcha gotcha. 
I think it's fine to leave as is. 
This is only tangentially related and more a note to myself than anything, but let's remember to consider, as we progress into next steps, the potential mismatch of path pipeline name and manifest pipeline name if a user changes it between `init` and `deploy`.

internal/pkg/cli/pipeline_init.go
```diff
@@ -545,13 +562,13 @@ func (o *initPipelineOpts) createPipelineManifest() error {
 		stages = append(stages, stage)
 	}
 
-	manifest, err := manifest.NewPipeline(pipelineName, provider, stages)
+	manifest, err := manifest.NewPipeline(o.name, provider, stages)
```
huanjani, line 571: This is where our name-spacing committee comes in! 😄  It would be nice to add in their prepending here, now, if it's ready, but we can do it later if need be.

internal/pkg/cli/interfaces.go
```diff
@@ -241,8 +242,8 @@ type wsPipelineManifestReader interface {
 }
 
 type wsPipelineWriter interface {
-	WritePipelineBuildspec(marshaler encoding.BinaryMarshaler) (string, error)
-	WritePipelineManifest(marshaler encoding.BinaryMarshaler) (string, error)
+	WritePipelineBuildspec(marshaler encoding.BinaryMarshaler, name string) (string, error)
+	WritePipelineManifest(marshaler encoding.BinaryMarshaler, name string) (string, error)
```
efekarakus, line 246: super nit: what do you think of flipping the parameter order just to be kindof consistent to [`os.WriteFile`](https://pkg.go.dev/os#WriteFile)

```suggestion
	WritePipelineBuildspec(name string, data encoding.BinaryMarshaler) (string, error)
	WritePipelineManifest(name string, data encoding.BinaryMarshaler) (string, error)
```

internal/pkg/cli/pipeline_init.go
efekarakus: I dig it 😄 the original idea was to have this bold term also becoming `prompt.WithFinalMessage`

internal/pkg/cli/pipeline_init.go
```diff
@@ -42,6 +42,9 @@ import (
 )
 
 const (
+	fmtPipelineInitNamePrompt  = "What would you like to %s this pipeline?"
+	pipelineInitNameHelpPrompt = "An identifier for your pipeline (e.g. beta, prod)."
```
efekarakus: nit:
```suggestion
	pipelineInitNameHelpPrompt = "An identifier for your pipeline (e.g. release, test, prod)."
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -157,35 +161,46 @@ func newInitPipelineOpts(vars initPipelineVars) (*initPipelineOpts, error) {
...
 			return err
 		}
 	}
+	if err := validatePipelineName(o.name, o.appName); err != nil {
+		return err
+	}
 
-	if len(o.environments) > 0 {
-		if err := o.validateEnvs(); err != nil {
+	// TODO: validate not a duplicate
```
efekarakus: What does this mean exactly? (This is my personal preference) but I find TODOs a bit intimidating. I can never tell if it's something that has to happen or it's just a nice to have and I'm safe to touch this code?

If it's just a nice-to-have, can we remove it please? 🥺

internal/pkg/cli/pipeline_init.go
efekarakus: 👍 I think we do something like this in some commands:

```go
if err := o.askOrValidatePipelineName(); err != nil { ...}

func (o *initPipelineOpts) askOrValidatePipelineName() error {
  if o.name != nil {
      return validatePipelineName(o.name, o.appName)
   }
   // prompt for stuff
}
```

efekarakus (commented):
Looks great 🤩

internal/pkg/cli/pipeline_init.go
dannyrandall: I think (maybe I'm wrong...) that we want the user provided name in the manifest file, not the prepended name - that will be added on in the `deploy` step.

internal/pkg/cli/pipeline_init.go
dannyrandall: good catch, I meant to remove that. It's something I need to do in my next PR :)

internal/pkg/cli/pipeline_init.go
dannyrandall: We do in this case, at least to keep the same behavior as before. I removed the validation code from `askEnvs()`, since it already was written in `validateEnvs()`

internal/pkg/cli/interfaces.go
dannyrandall: normally I'd that sounds like a great change, however since `WriteServiceManifest()` and `WriteJobManifest()` have this ordering, I think it makes more sense to stay consistent with them since they are right next to these functions. 

I would be open to changing all of them, but probably a little out of scope for this PR.
https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/workspace/workspace.go#L313-L328

internal/pkg/cli/pipeline_init.go
huanjani: Discussed offline. For the record: We're following the model of namespaced envs and workloads-- name in manifest will be the one chosen by user, not the prepended resource name.

internal/pkg/cli/validate_test.go
```diff
@@ -147,6 +147,63 @@ func TestValidateEnvironmentName(t *testing.T) {
...
+		"string with invalid characters": {
+			val:    "myPipe!",
+			wanted: errValueBadFormat,
+		},
+		"longer than 100 characters": {
+			val:               strings.Repeat("s", 101),
+			wantedErrorSuffix: fmt.Sprintf(fmtErrPipelineNameTooLong, 90),
+		},
+		"longer than 100 characters with pipeline-[app]": {
+			val:               strings.Repeat("x", 86),
```
huanjani, line 175: That is a much better way of making a long string than what I've done in the past! 😆

dannyrandall: i found it in someone else's test, i didn't think of it either 😂

huanjani (approved):
This looks fantastic! 🎉

efekarakus (approved):

PR #3382, opened 2022-03-18
chore(pipeline): add deployed pipeline selector

huanjani:
For each pipeline command that deals with deployed pipelines (`delete`, `show`, `status`), this adds a selector with deployed pipelines as options. Note: `deploy` can also deploy already-deployed pipelines, but must be in the workspace to do so, so fetching the workspace pipelines catches both undeployed-but-inited and already-deployed pipelines.

(The next PR will handle deleting github access token secrets in `pipeline_delete`.)


internal/pkg/cli/pipeline_delete.go
```diff
@@ -144,6 +160,22 @@ func (o *deletePipelineOpts) Execute() error {
...
+	sel     codePipelineSelector
+	command string
+}
+
+func askDeployedPipelineName(input *askDeployedPipelineNameInput) (string, error) {
+	pipeline, err := input.sel.DeployedPipeline(fmt.Sprintf(fmtPipelineCommandPrompt, color.HighlightUserInput(input.appName), input.command), "", map[string]string{
+		deploy.AppTagKey: input.appName,
+	})
+	if err != nil {
+		return "", fmt.Errorf("get deployed pipelines: %w", err)
```
dannyrandall, line 174: nit: get -> select (like `askAppName()`)

internal/pkg/cli/pipeline_delete.go
```diff
@@ -144,6 +160,22 @@ func (o *deletePipelineOpts) Execute() error {
...
 }
 
+type askDeployedPipelineNameInput struct {
+	appName string
+	sel     codePipelineSelector
+	command string
+}
+
+func askDeployedPipelineName(input *askDeployedPipelineNameInput) (string, error) {
+	pipeline, err := input.sel.DeployedPipeline(fmt.Sprintf(fmtPipelineCommandPrompt, color.HighlightUserInput(input.appName), input.command), "", map[string]string{
```
dannyrandall, line 170: i like this😌

internal/pkg/cli/pipeline_delete.go
dannyrandall: nit: get -> select (like `askAppName()`)

internal/pkg/cli/pipeline_delete.go
dannyrandall: i like this😌

dannyrandall (approved):
lgtm!

internal/pkg/cli/pipeline_delete.go
```diff
@@ -144,6 +160,22 @@ func (o *deletePipelineOpts) Execute() error {
 	return nil
 }
 
+type askDeployedPipelineNameInput struct {
```
efekarakus: I think it's okay if we have these 3 fields be passed as parameters instead of creating a new reserved name in the `cli` pkg:
```go
func askDeployedPipelineName(selector codePipelineSelector, appName, msg string)
```
Maybe instead of taking a command as input we can also take the prompt msg for extra flexibility.

internal/pkg/term/selector/selector.go
```diff
@@ -176,10 +181,17 @@ type WorkspaceSelect struct {
...
+// WsPipelineSelect is a workspace pipeline selector.
+type WsPipelineSelect struct {
 	prompt Prompter
-	ws     WorkspacePipelinesLister
+	ws     WsPipelinesLister
+}
+
+// CodePipelineSelect is a selector for deployed pipelines.
+type CodePipelineSelect struct {
+	*Select
```
efekarakus: Do we need to embed `*Select`? it doesn't look like we're using it

efekarakus (commented):
Looks great!

internal/pkg/cli/pipeline_delete.go
huanjani: Okay, I can make that change! 
But for the future: how do I know when it's better to make a struct and when it's better to use parameters? Is it that the struct is a more logical grouping of elements that belong together, but parameters are items that are needed but not really parts of a whole? And/or if the struct object will be used as such more than once?

internal/pkg/term/selector/selector.go
huanjani: We use it in `askAppName()` for `o.sel.Application(...)`: https://github.com/aws/copilot-cli/pull/3382/files#diff-f7c23022b6f1c5454bea7d03c7570bbddf15b0f27bb1aa09d41421d0176bc5ceR180

internal/pkg/cli/pipeline_delete.go
efekarakus: I should have written "nit" here I think - It's totally fine what you had before! I was just trying to reduce the number of names declared in the package since `cli` is so big 😅 

It's a good question, I declare a type when I can create a "data abstraction" and associate a set of behaviors (methods), or I want to create an "aggregate" type to bundle a bunch of related fields (like your situation or the AWS SDK).
Honestly, it was just a suggestion, it felt like the fn signature is small enough that we didn't need an extra type

internal/pkg/cli/pipeline_delete.go
```diff
@@ -144,6 +156,16 @@ func (o *deletePipelineOpts) Execute() error {
 	return nil
 }
 
+func askDeployedPipelineName(sel codePipelineSelector, appName, msg string) (string, error) {
+	pipeline, err := sel.DeployedPipeline(fmt.Sprintf(msg, color.HighlightUserInput(appName)), "", map[string]string{
```
efekarakus: Instead of taking the fmtString, can we let the client generate it:
```suggestion
func askDeployedPipelineName(sel codePipelineSelector, appName, msg string) (string, error) {
	pipeline, err := sel.DeployedPipeline(msg, "", map[string]string{
```

internal/pkg/term/selector/selector.go
efekarakus: Gotcha, I think it's just the names that's a bit odd - a "codepipeline selector" exposes also an "application selector" 🤔 

What do you think of something like this:
```go
type CodePipelineSelect struct {
    prompt  Prompter
    codepipeline  CodePipelineLister
}

type AppPipelineSelect struct {
    *Select
    *CodePipelineSelect
}
```

efekarakus (approved):
Looks good! gave it a :shipit: with two small feedback 🙏 feel free to remove the label afterwards

Issue #2694, opened 2021-08-04
Import and use an existing certificate for my domain.


makies:
I have a wildcard SSL certificate on AWS ACM.  When deploying, copilot does not use this certificate, creates another certificate and attempts DNS authentication, but it is not authenticated.

We want to import and use an existing certificate for my domain.

(Japanese)
AWSのACMでワイルドカードのSSL証明書を保持しています。デプロイする時に、copilotは、この証明書を使わずに、別な証明書を作成し、DNS認証を試みますが、それは認証されません。

独自ドメインの既存の証明書をインポートして使いたいです。

efekarakus:
Hi @makies ! Thank you for the feature request, it makes a lot of sense.

@iamhopaul123 on the team had started on this feature before 😬 https://github.com/aws/copilot-cli/pull/2413
We should be able to resurrect the changes 🧟  and continue from there!

heyheman11:
+1 on this issue, seems very hacky to get containers running over https if your domain isn't currently hosted on AWS

chriskuech:
@heyheman11 , were you ever able to get https containers running without hosting the domain on AWS?  I'm desperate for a solution.  It seems as though we cannot use a reverse proxy container as the healthchecks are http and only one port is supported.

heyheman11:
@chriskuech I followed the instructions from a previous reply from the team at AWS but here is a brief overview of what I did:

**1.** Ensure manifest.yml looks something like this (ensure container only exposes port 80):

```yml
http:
  path: "/"
  healthcheck:
    path: "/health_check/"
    success_codes: "200"

image:
  build:
    context: ./project
    dockerfile: ./project/Dockerfile
  port: 80
```

**2.** Create a certificate in ACM or import an exisiting one for the desired domain

**3.** Create a new listener rule on the existing load balancer that listens on port 443, use the certificate from step 2 and copy all rules from the existing rule.

**4.** Go to your DNS provider and create a new `CNAME` record to point to the ELB address

Essentially the https connection is terminated at the listener, and my server in the container only listens for http traffic on port 80.

chriskuech:
This seems to be a valid template snippet for what you are describing and it passes a linter, but fails in cloudformation with a "AWS::ElasticLoadBalancingV2::ListenerRule Validation exception", but doesn't seem to show anywhere what the actual validation issue is.

Any idea what I'm doing wrong?

```yaml
  HttpsListener:
[...]

PR #3386, opened 2022-03-21
chore: add CFN and update deploy for cert import

iamhopaul123:
Related #2694


internal/pkg/aws/acm/acm.go
```diff
@@ -0,0 +1,30 @@
...
+// New returns an ACM struct configured against the input session.
+func New(s *session.Session) *ACM {
+	return &ACM{
+		client: acm.New(s),
+	}
+}
+
+// ValidCertificatesAliases is a no-op and is supposed to validates if aliases are all valid against the provided ACM certificates.
+func (a *ACM) ValidCertificatesAliases(aliases []string, certs []string) error {
+	return nil
```
paragbhingre, line 29: Curious to know why does this return nil currently? Is it yet to implement this functionality?

iamhopaul123: yeah like i mentioned in the comment it is a no-op for now. Will add it in the next PR.

internal/pkg/aws/acm/acm.go
```diff
@@ -0,0 +1,30 @@
...
+
+// New returns an ACM struct configured against the input session.
+func New(s *session.Session) *ACM {
+	return &ACM{
+		client: acm.New(s),
+	}
+}
+
+// ValidCertificatesAliases is a no-op and is supposed to validates if aliases are all valid against the provided ACM certificates.
+func (a *ACM) ValidCertificatesAliases(aliases []string, certs []string) error {
```
efekarakus: nit:
```suggestion
// ValidateCertAliases is a no-op and is supposed to validates if aliases are all valid against the provided ACM certificates.
func (a *ACM) ValidateCertAliases(aliases []string, certs []string) error {
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -61,9 +62,13 @@ const (
 )
 
 var (
-	aliasUsedWithoutDomainFriendlyText = fmt.Sprintf("To use %s, your application must be associated with a domain: %s.\n",
+	rdwsAliasUsedWithoutDomainFriendlyText = fmt.Sprintf("To use %s, your application must be associated with a domain: %s.\n",
 		color.HighlightCode("http.alias"),
 		color.HighlightCode("copilot app init --domain example.com"))
+	ecsAliasUsedWithoutDomainFriendlyText = fmt.Sprintf("To use %s, your application must be associated with a domain: %s, or import any certificates to your environment: %s.\n",
```
efekarakus: Should we separate with new lines since the text is long?
```suggestion
	ecsAliasUsedWithoutDomainFriendlyText = fmt.Sprintf(`To use %s, your application must be:
* Associated with a domain: %s
* Or, using imported certificates to your environment: %s
`,
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -299,6 +309,7 @@ func NewLBDeployer(in *WorkloadDeployerInput) (*lbSvcDeployer, error) {
 		appVersionGetter:       versionGetter,
 		publicCIDRBlocksGetter: envDescriber,
 		lbMft:                  lbMft,
+		aliasCertValidator:     acm.New(svcDeployer.envSess),
```
efekarakus, line 320: just to validate, this is using the creds from the EnvManagerRole right?

internal/pkg/deploy/cloudformation/stack/lb_grpc_web_service_integration_test.go
```diff
@@ -89,6 +89,7 @@ func TestGrpcLoadBalancedWebService_Template(t *testing.T) {
 			actualString = strings.ReplaceAll(actualString, envControllerZipFile, "mockEnvControllerZipFile")
 			actualString = strings.ReplaceAll(actualString, dynamicDesiredCountZipFile, "mockDynamicDesiredCountZipFile")
 			actualString = strings.ReplaceAll(actualString, rulePriorityZipFile, "mockRulePriorityZipFile")
+			fmt.Println(actualString)
```
efekarakus: 😂 
```suggestion
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -916,6 +921,37 @@ func (d *lbSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (*svcS
 	}, nil
 }
 
+func (d *lbSvcDeployer) albOpts(rootUserARN string) ([]stack.LoadBalancedWebServiceOption, error) {
```
efekarakus: I'm not saying we should do this now, but I'm trying to think what would be the ideal code layout. This logic feels too low-level to be in the `cli/` pkgs 🤔 What do you think?

```go
stack, err := stack.NewLoadBalancedWebService(manifest, env, flags) // returns an error for things like errSvcWithNoAliasDeployingToEnvWithImportedCerts
```

I want to figure out what's the appropriate place to put `validateLBWSRuntime` maybe we need a `Validate` method like for manifests but for `stack` objects

internal/pkg/cli/deploy/errors.go
```diff
@@ -0,0 +1,15 @@
...
+
+import "fmt"
+
+type errSvcWithNoAliasDeployingToEnvWithImportedCerts struct {
+	name    string
+	envName string
+}
+
+func (e *errSvcWithNoAliasDeployingToEnvWithImportedCerts) Error() string {
+	return fmt.Sprintf("cannot deploy service %s without custom domain to environment %s with certificate imported", e.name, e.envName)
```
efekarakus: nit: should we say like this?
```suggestion
	return fmt.Sprintf("cannot deploy service %s without http.alias to environment %s with certificate imported", e.name, e.envName)
```

internal/pkg/config/env.go
```diff
@@ -29,18 +29,25 @@ type Environment struct {
...
+	ImportVPC      *ImportVPC `json:"importVPC,omitempty"`
+	VPCConfig      *AdjustVPC `json:"adjustVPC,omitempty"`
+	ImportCertARNs []string   `json:"importCertARNs,omitempty"`
 }
 
-// NewCustomizeEnv returns a new CustomizeEnv struct.
-func NewCustomizeEnv(importVPC *ImportVPC, adjustVPC *AdjustVPC) *CustomizeEnv {
-	if importVPC == nil && adjustVPC == nil {
+// HasImportedCerts return if the environment has imported certs.
+func (e *Environment) HasImportedCerts() bool {
```
efekarakus: nit: can we move the method under the type definition?

internal/pkg/config/env.go
```diff
@@ -29,18 +29,25 @@ type Environment struct {
...
+	return e.CustomConfig != nil && len(e.CustomConfig.ImportCertARNs) != 0
+}
+
+// CustomizeEnvP returns a pointer to the CustomizeEnv value passed in.
+func CustomizeEnvP(input CustomizeEnv) *CustomizeEnv {
```
efekarakus: out of curiousity, do we need the `CustomConfig` field's type to be `*CustomizeEnv` instead of `CustomizeEnv`? If we can make it a concrete type instead of a pointer then I think we can get rid of this func.

We can instead add 
```go
func (ce CustomizeEnv) IsEmpty() bool
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -37,14 +37,23 @@ Conditions:
...
+    - !Condition HasImportedCert
   CreateEFS:
     !Not [!Equals [ !Ref EFSWorkloads, ""]]
   CreateNATGateways:
     !Not [!Equals [ !Ref NATWorkloads, ""]]
   HasAliases:
     !Not [!Equals [ !Ref Aliases, "" ]]
+  HasImportedCert:
+{{- if .ImportCertARNs}}
+    !Equals [ 'true', 'true' ]
```
efekarakus: What do you think of having the cert ARNs as a parameter instead? and then doing `!Not [!Equals [ !Ref PublicALBCerts, "" ]]`

I think we have basically two options:
1. Either use CFN conditions properly with parameters and no Go templates
2. We delete the `ExportHTTPSListener` and `HasImportedCert` conditions and replace them with go template logic `{{if .NeedsToExportHTTPSListener }}`

internal/pkg/cli/deploy/deploy.go
Lou1415926: nit that I am not even sure myself:
```suggestion
	ecsAliasUsedWithoutDomainFriendlyText = fmt.Sprintf("To use %s, your application must be associated with a domain: %s, or import one or more certificates to your environment: %s.\n",
```

I was just worried about the usage of "any" would let people think that they could really import just **any** certificate and `alias` would work magically😆 I'm no English expert and I don't know if I'm fixing it right or if this even needs to be fixed.

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -183,17 +192,31 @@ Resources:
       Protocol: HTTP
   HTTPSListener:
     Type: AWS::ElasticLoadBalancingV2::Listener
-    DependsOn: HTTPSCert
     Condition: ExportHTTPSListener
     Properties:
       Certificates:
+{{- if .ImportCertARNs}}
+        - CertificateArn: {{index .ImportCertARNs 0}}
+{{- else }}
```
efekarakus, line 200: 🤦 Aaaaahhhhhhhhhhhhhhhhhhhhhhhhhhhh

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -397,8 +420,7 @@ Outputs:
     Export:
       Name: !Sub ${AWS::StackName}-SubDomain
   EnabledFeatures:
-    # We don't need to include Aliases because updating it always results in the CustomDomain action to update.
-    Value: !Sub '${ALBWorkloads},${EFSWorkloads},${NATWorkloads}'
+    Value: !Sub '${ALBWorkloads},${EFSWorkloads},${NATWorkloads},${Aliases}'
```
efekarakus: Why did we include `Aliases`? I'm a little worried that this will cause the environment updates to hit limits too frequently.

internal/pkg/cli/deploy/deploy.go
```diff
@@ -1176,23 +1212,32 @@ func validateAppVersion(appName string, appVersionGetter versionGetter) error {
 	return nil
 }
 
-func validateLBWSRuntime(app *config.Application, envName string, mft *manifest.LoadBalancedWebService, appVersionGetter versionGetter) error {
-	if app.Domain == "" && mft.HasAliases() {
-		log.Errorf(aliasUsedWithoutDomainFriendlyText)
-		return errors.New("alias specified when application is not associated with a domain")
+func validateLBWSRuntime(app *config.Application, env *config.Environment, mft *manifest.LoadBalancedWebService, appVersionGetter versionGetter, validator aliasCertValidator) error {
```
Lou1415926: Could this function be a method for `lbSvcDeployer`? Seems like it's only ever called by `lbSvcDeployer`.

internal/pkg/cli/deploy/deploy.go
```diff
@@ -916,6 +921,37 @@ func (d *lbSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (*svcS
...
 
+func (d *lbSvcDeployer) albOpts(rootUserARN string) ([]stack.LoadBalancedWebServiceOption, error) {
+	var opts []stack.LoadBalancedWebServiceOption
+	if d.app.RequiresDNSDelegation() {
+		opts = append(opts, stack.WithDNSDelegation(deploy.AppInformation{
+			Name:                d.app.Name,
+			DNSName:             d.app.Domain,
+			AccountPrincipalARN: rootUserARN,
+		}))
+		if !d.lbMft.RoutingRule.Disabled() {
```
Lou1415926: Should this be instead a part of the `validateLBWSRuntime` function?

internal/pkg/template/templates/environment/partials/lambdas.yml
```diff
@@ -12,28 +13,29 @@ CertificateValidationFunction:
...
     Timeout: 600
     MemorySize: 512
     Role: !GetAtt 'CustomResourceRole.Arn'
-    Runtime: nodejs12.x
+    Runtime: nodejs12.x 
+{{- end}}
 
-CustomDomainFunction:
-  Condition: HasAliases
+DNSDelegationFunction:
```
efekarakus, line 28: Why do we want the DNSDelegation Lambda still created, wouldn't the customer be in charge of all the hosted zones and the records when they import a cert?

internal/pkg/cli/deploy/deploy.go
```diff
@@ -916,6 +921,37 @@ func (d *lbSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (*svcS
...
+			if d.env.HasImportedCerts() && !d.lbMft.HasAliases() {
```
Lou1415926: `HasAliases` also return true if there is an alias for `nlb` field. I assume we only care about `http.alias` here right 🤔 ?

internal/pkg/cli/deploy/deploy.go
```diff
@@ -916,6 +921,37 @@ func (d *lbSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (*svcS
...
+				return nil, &errSvcWithNoAliasDeployingToEnvWithImportedCerts{
+					name:    d.name,
+					envName: d.env.Name,
+				}
+			}
+			opts = append(opts, stack.WithHTTPS())
+		}
+	} else {
```
Lou1415926: Do we error out if:
1. cert is imported
2. no `http.alias` is specified
Regardless of whether DNS is delegated? If so we could probably reduce a layer if `if-else` block.

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
```diff
@@ -293,16 +272,37 @@ Resources:
       ManagedPolicyArns:
         - !Sub arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
 
+{{- if .HTTPSListener}}
+{{- if not .Aliases}}
+  LoadBalancerDNSAlias:
```
efekarakus: should this be something like `if not aliases and not importing certs`?

internal/pkg/cli/deploy/deploy.go
```diff
@@ -1176,23 +1212,32 @@ func validateAppVersion(appName string, appVersionGetter versionGetter) error {
...
+	}
+	if env.HasImportedCerts() {
+		aliases, err := mft.RoutingRule.Alias.ToStringSlice()
+		if err != nil {
+			return fmt.Errorf("convert aliases to string slice: %w", err)
+		}
+		if err := validator.ValidCertificatesAliases(aliases, env.CustomConfig.ImportCertARNs); err != nil {
+			return fmt.Errorf("validate aliases against the imported certificate for env %s: %w", env.Name, err)
+		}
+		return validateLBSvcAlias(mft.NLBConfig.Aliases, app, env.Name)
```
Lou1415926: super nit: alternatively we can do `validateLBSvcAlias`right after the `if !mft.HasAliases()` block so that it doesn't have to be repeated at the end of two `if` blocks.

internal/pkg/cli/deploy/deploy.go
```diff
@@ -1176,23 +1212,32 @@ func validateAppVersion(appName string, appVersionGetter versionGetter) error {
...
 	}
-
 	if app.RequiresDNSDelegation() {
```
Lou1415926: Totally unrelated thought: we can maybe change this method name to `app.DNSDelegated` 🤔  Feels like a better description of its intention.

internal/pkg/cli/deploy/deploy.go
Lou1415926: If cert is imported but DNS isn't delegated, I think NLB alias doesn't work : (

Would it be clearer if we write it like this:

1. Validate the NLB aliases.
1.1 NLB aliases used only when domain delegation happens.
1.2 NLB aliases has the supported format
2. Validate the ALB aliases.
2.1 If import cert, validate ALB aliases against the cert
2.2 O/W
2.2.1 Validate that domain delegation happens
2.2.2  ALB alias has the supported format

internal/pkg/cli/deploy/errors.go
Lou1415926:
```suggestion
	return fmt.Sprintf("cannot deploy service %s without a custom domain to environment %s with certificate imported", e.name, e.envName)
```

internal/pkg/config/env.go
```diff
@@ -29,18 +29,25 @@ type Environment struct {
...
+	if input.ImportVPC == nil && input.VPCConfig == nil && input.ImportCertARNs == nil {
```
Lou1415926: Do we need to consider `len(input.ImportCertARNs) == 0` here?

internal/pkg/config/env.go
Lou1415926: Is this mainly refactoring or is there any other concern that led to this change?

internal/pkg/deploy/cloudformation/stack/lb_grpc_web_service_integration_test.go
Lou1415926: hehehe

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -202,6 +201,7 @@ func (s *LoadBalancedWebService) Template() (string, error) {
 		Variables:                      s.manifest.TaskConfig.Variables,
 		Secrets:                        convertSecrets(s.manifest.TaskConfig.Secrets),
 		Aliases:                        aliases,
+		HTTPSListener:                  s.httpsEnabled,
```
Lou1415926, line 213: How does this go-template data play with the CFN parameter `HTTPSEnabled`?

internal/pkg/template/templates/environment/partials/custom-resources.yml
```diff
@@ -36,7 +37,6 @@ CustomDomainAction:
   Metadata:
     'aws:copilot:description': 'Add an A-record to the hosted zone for the domain alias'
   Condition: HasAliases
-  DependsOn: HTTPSCert
```
Lou1415926, line 39: stoopid question: why does it not need to depend on `HTTPSCert` now?

internal/pkg/template/templates/environment/cf.yml
Lou1415926: dumb q: why do we now want to include Aliases?

internal/pkg/cli/deploy/deploy.go
iamhopaul123: :+1:

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
```diff
@@ -143,10 +138,15 @@ Resources:
     Type: AWS::ECS::Service
     DependsOn:
     {{- if .ALBEnabled}}
-     - WaitUntilListenerRuleIsCreated
```
Lou1415926, line 146: Stupid question again: why do we now not need the wait condition?

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
```diff
@@ -293,16 +272,37 @@ Resources:
       ManagedPolicyArns:
         - !Sub arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
 
+{{- if .HTTPSListener}}
```
Lou1415926, line 276: Having imported cert also create HTTPSListener right? Do we still want the DNS alias in this case?

internal/pkg/template/templates/environment/cf.yml
iamhopaul123: We can't do 1 because CFN doesn't support looping. 2 sounds ok to me. It's just I find too much go template make the template more difficult to read.

Alternatively, we could have both import cert as go template and CFN, and we pass the same value for them. That way we refer to the CFN param as much as we can, and for looping we use go template.

internal/pkg/template/templates/environment/partials/lambdas.yml
iamhopaul123: Discussed offline. Will keep it if they have domain specified.

internal/pkg/template/templates/environment/cf.yml
iamhopaul123: > I'm a little worried that this will cause the environment updates to hit limits too frequently.

it would be the same behavior as is 🤔 because atm `CustomDomainAction` get constantly updated whenever they specify new aliases. Alternatively, we could update the env controller so that it won't pass any alias to the env controller if env has cert imported.

iamhopaul123: Let's do it I think it would be a better option

internal/pkg/cli/deploy/deploy.go
iamhopaul123: Good call!

internal/pkg/template/templates/environment/partials/custom-resources.yml
iamhopaul123: Because it's redundant. It doesn't need to depend on the `HTTPSCert`.

internal/pkg/template/templates/environment/cf.yml
iamhopaul123: Explained here and i'll take another route https://github.com/aws/copilot-cli/pull/3386#discussion_r844515209

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
iamhopaul123: Because `WaitUntilListenerRuleIsCreated` is used for conditional dependency. In this PR I use go template to handle the logic so that we can ditch this.

internal/pkg/cli/deploy/deploy.go
huanjani: maybe "or import relevant certificate(s)..."?

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
```diff
@@ -393,11 +391,11 @@ Resources:
                   - !Sub "/${RulePath}/*"
       ListenerArn: !GetAtt EnvControllerAction.HTTPSListenerArn
       Priority: !GetAtt HTTPSRulePriorityAction.Priority
+{{- else}}  {{/* elseif .HTTPSListener */}}
```
efekarakus: What do you think of creating two separate partials `http-listener` and `https-listener`?

internal/pkg/template/templates/environment/cf.yml
efekarakus: > it would be the same behavior as is 🤔 

Oh I meant the Output value limit sorry (which I think is probably 4096 bytes).

> Let's do it I think it would be a better option

That sounds good to me too - but I didn't understand why you updated this value originally 💭 not that it matters if you're going to revert it 😝

internal/pkg/config/env.go
iamhopaul123: Unfortunately `omitempty` only works for nil struct (if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.) which means if we make it `CustomizeEnv` it will be shown in `env ls --json` as an empty struct if it is empty :(

internal/pkg/cli/deploy/deploy.go
iamhopaul123: Good call I don't know why I did this 🤦

internal/pkg/config/env.go
iamhopaul123: This is not a refactoring. This is a shorthand method similar to `stringP` or `aws.String()` that returns nil if it is empty, otherwise return a pointer that points to the struct.

internal/pkg/cli/deploy/deploy.go
iamhopaul123: Yes that is correct

internal/pkg/cli/deploy/deploy.go
iamhopaul123: > If cert is imported but DNS isn't delegated, I think NLB alias doesn't work : (

Good call!

> super nit: alternatively we can do validateLBSvcAliasright after the if !mft.HasAliases() block so that it doesn't have to be repeated at the end of two if blocks.

I still think the current validation ordering makes more sense to me. For example, if their app isn't associated with any domain and they don't import cert at all, it doesn't make sense for us to check if their aliases are valid against the hosted zones (since they don't exist at all)

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
iamhopaul123: :+1:

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
iamhopaul123: CFN parameter `HTTPSEnabled` we don't use it in our template anymore, the only reason why we keep it is because our operational commands depend on this parameter.

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
iamhopaul123: Actually, now that we require them to specify alias whenever they use imported certs for the env, we can skip this condition. Because when they don't have alias, they won't have imported certs.

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
iamhopaul123: Actually, now that we require them to specify alias whenever they use imported certs for the env, we can skip this condition. Otherwise we might put redundant validation here which is potentially confusing.

internal/pkg/template/templates/environment/cf.yml
iamhopaul123: > That sounds good to me too - but I didn't understand why you updated this value originally 💭

I think we might still want to add it. The reason is because: it is always guaranteed rn that whenever a new alias is added to the env stack, at least one resource (`CustomDomainAction`) will get updated so that our env controller won't get "no stack update" error. However, if an environment imports any certificates, `CustomDomainAction` won't be created and when service is switching the alias it is using, env controller will return an no-update error without the code change we are talking about. That's the initiative why I added this.

Alternatively, like I mentioned above, we could modify the input for env controller so that i won't add new aliases to the env stack if it imports any certificates. However, it would be a problem that the env stack `aliases` param won't have a full set of all the aliases which potentially could be a problem for our operational commands that grab alias info from there.

Or we could let env controller ignore the `no-update` error when it is updating the env stack, but that could also lead to unexpected consequences because we are ignoring a type of error.

> Oh I meant the Output value limit sorry (which I think is probably 4096 bytes).

That is true, but since the parameter length limit is also 4096 bytes, most likely even if we go with another route they'll face the same issue unless we use SSM or other places to store their aliases rather than using the CFN parameter.

Also I didn't see there's 4096 limit for output value (only saw 4096 byte limit for parameter value and 200 outputs limit) 🤔

internal/pkg/cli/deploy/deploy.go
```diff
@@ -61,9 +62,19 @@ const (
...
 
 var (
-	aliasUsedWithoutDomainFriendlyText = fmt.Sprintf("To use %s, your application must be associated with a domain: %s.\n",
+	rdwsAliasUsedWithoutDomainFriendlyText = fmt.Sprintf("To use %s, your application must be associated with a domain: %s.\n",
 		color.HighlightCode("http.alias"),
 		color.HighlightCode("copilot app init --domain example.com"))
+	ecsALBAliasUsedWithoutDomainFriendlyText = fmt.Sprintf(`To use %s, your application must be:
+* Associated with a domain: %s
+* Or, using imported certificates to your environment: %s
+`,
```
efekarakus: nit: to avoid super long lines
```suggestion
	ecsALBAliasUsedWithoutDomainFriendlyText = fmt.Sprintf(`To use %s, your application must be:
* Associated with a domain:
     %s
* Or, using imported certificates to your environment: 
     %s
`,
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -1176,23 +1192,53 @@ func validateAppVersion(appName string, appVersionGetter versionGetter) error {
 	return nil
 }
 
-func validateLBWSRuntime(app *config.Application, envName string, mft *manifest.LoadBalancedWebService, appVersionGetter versionGetter) error {
-	if app.Domain == "" && mft.HasAliases() {
-		log.Errorf(aliasUsedWithoutDomainFriendlyText)
-		return errors.New("alias specified when application is not associated with a domain")
+func (d *lbSvcDeployer) validateALBWSRuntime() error {
```
efekarakus, line 1197: nice, I appreciate the removal of all these parameters 🥰 and the validation logic is clear 👌

internal/pkg/config/env.go
efekarakus: Maybe we should rename like this, what do y'all think?
```go
type CustomizeEnv struct -> type CustomEnvConfig struct
func CustomizeEnvP -> func CustomizeEnv
```

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -62,64 +60,62 @@ type LoadBalancedWebService struct {
...
-}
-
-// WithDNSDelegation enables DNS delegation for a LoadBalancedWebService.
-func WithDNSDelegation(app deploy.AppInformation) func(s *LoadBalancedWebService) {
-	return func(s *LoadBalancedWebService) {
-		s.dnsDelegationEnabled = true
-		s.appInfo = app
-	}
+// NewLoadBalancedWebServiceOpts contains fields to configure LoadBalancedWebService.
+type NewLoadBalancedWebServiceOpts struct {
```
efekarakus: nit: nice, i like this too - way simpler, just a nit
```suggestion
type LoadBalancedWebServiceOpts struct {
```

internal/pkg/config/env.go
Lou1415926: Would having 
```
func (c *CustomizeEnv) IsEmpty() bool 
```
be a working alternative to this? So that we don't need `nil` to determine whether the struct is in fact empty?

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -62,64 +60,62 @@ type LoadBalancedWebService struct {
...
+	App                    *config.Application
+	Env                    *config.Environment
+	Manifest               *manifest.LoadBalancedWebService
+	RuntimeConfig          RuntimeConfig
+	RootUserARN            string
+	PublicSubnetCIDRBlocks []string
```
Lou1415926: nit: aggregating all params into an opts struct is necessary here since so many params are being passed. But what do you think of still having `PublicSubnetCIDRBlocks` passed as an option like before? After all only NLB services would care about `PublicSubnetCIDRBlocks`. Maybe we can include parameters that are applicable to all LBWS into the opts struct, and keep those conditional params in `LoadBalancedWebServiceOption`?

Another nice thing with `LoadBalancedWebServiceOption` was that we could clearly see what's being "switched on/off" under different scenarios. Now it's not obvious, for example, that for an ALBWS the `publicSubnetCIDRBlock` is going to be empty.

Just a thought - we don't need to take action right now. We can always refactor later.

internal/pkg/template/templates/environment/cf.yml
efekarakus: Discussed offline! we'll fetch the alias from the manifest and can remove this change 🥳

internal/pkg/template/workload.go
```diff
@@ -448,6 +450,7 @@ type WorkloadOpts struct {
 	Variables                map[string]string
 	Secrets                  map[string]Secret
 	Aliases                  []string
+	HTTPSListener            bool
```
efekarakus, line 461: :cool-crying: the ever growing list of options

internal/pkg/cli/env_init.go
```diff
@@ -262,7 +262,7 @@ func (o *initEnvOpts) Execute() error {
 		return fmt.Errorf("get identity: %w", err)
 	}
 
-	if app.RequiresDNSDelegation() {
+	if app.DNSDelegated() {
```
Lou1415926: Do you think we need / have a good way to separate the implementation of `RequiresDNSDelegation` and `DNSDelegated` ?

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -112,6 +114,10 @@ func (e *EnvStackConfig) Template() (string, error) {
 
 // Parameters returns the parameters to be passed into a environment CloudFormation template.
 func (e *EnvStackConfig) Parameters() ([]*cloudformation.Parameter, error) {
+	httpsListener := "false"
+	if len(e.in.ImportCertARNs) != 0 || e.in.App.DNSDelegationRole() != "" {
```
Lou1415926: question 🤚🏼  Why do we use `DNSDelegationRole` to determine whether it's delegated or not? Is there a more straightforward way for us to tell 🤔 ?

internal/pkg/template/templates/environment/partials/lambdas.yml
Lou1415926: > Discussed offline. Will keep it if they have domain specified.

Seems like it's not kept when import cert happens! This makes sense to me per our offline discussion but I just want to make it clear here in case others are reviewing!

internal/pkg/config/env.go
iamhopaul123: So we will have
 
```golang
CustomizeEnv(CustomEnvConfig) *CustomEnvConfig
```

It is a little strange to me 🤔

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -60,62 +60,75 @@ type LoadBalancedWebService struct {
...
+	addons, err := addon.New(aws.StringValue(conf.Manifest.Name))
 	if err != nil {
 		return nil, fmt.Errorf("new addons: %w", err)
 	}
 	var dnsDelegationEnabled, httpsEnabled bool
 	var appInfo deploy.AppInformation
 
-	if opts.App.DNSDelegated() {
+	if conf.App.Domain != "" {
 		dnsDelegationEnabled = true
```
Lou1415926, line 90: Should `dnsDelegationEnabled` be true only when `conf.App.Domain != "" && !conf.Env.HasImportedCerts()`

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -60,62 +60,75 @@ type LoadBalancedWebService struct {
...
 		appInfo = deploy.AppInformation{
-			Name:                opts.App.Name,
-			DNSName:             opts.App.Domain,
-			AccountPrincipalARN: opts.RootUserARN,
+			Name:                conf.App.Name,
+			Domain:              conf.App.Domain,
+			AccountPrincipalARN: conf.RootUserARN,
 		}
 		httpsEnabled = true
 	} else {
```
Lou1415926: Was it intentional to wrap it with an `else` block here, for semantic purpose?

iamhopaul123: It is intentional because if it has domain then `httpsEnabled` must be true and we don't have to check if it has certificates imported.

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
iamhopaul123: Good call! Addressed

internal/pkg/cli/deploy/deploy.go
```diff
@@ -894,17 +916,13 @@ func (d *lbSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (*svcS
...
-		}
-	}
-	conf, err := stack.NewLoadBalancedWebService(d.lbMft, d.env.Name, d.app.Name, *rc, opts...)
+	conf, err := stack.NewLoadBalancedWebService(stack.LoadBalancedWebServiceConfigs{
+		App:           d.app,
+		Env:           d.env,
+		Manifest:      d.lbMft,
+		RuntimeConfig: *rc,
+		RootUserARN:   in.RootUserARN,
+	}, opts...)
```
efekarakus, line 925: now that we have a `Config` type, can we delete `opts...`?

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -46,73 +47,77 @@ type loadBalancedWebSvcReadParser interface {
...
 }
 
-// WithDNSDelegation enables DNS delegation for a LoadBalancedWebService.
-func WithDNSDelegation(app deploy.AppInformation) func(s *LoadBalancedWebService) {
-	return func(s *LoadBalancedWebService) {
-		s.dnsDelegationEnabled = true
-		s.appInfo = app
-	}
+// LoadBalancedWebServiceConfigs contains fields to configure LoadBalancedWebService.
+type LoadBalancedWebServiceConfigs struct {
```
efekarakus: nit:
```suggestion
// LoadBalancedWebServiceConfig contains fields to configure LoadBalancedWebService.
type LoadBalancedWebServiceConfig struct {
```

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -98,10 +94,10 @@ func NewLoadBalancedWebService(conf LoadBalancedWebServiceConfigs,
 			AccountPrincipalARN: conf.RootUserARN,
 		}
 		httpsEnabled = true
-	} else {
-		if conf.Env.HasImportedCerts() {
-			httpsEnabled = true
-		}
+	}
+	if conf.Env.HasImportedCerts() {
+		httpsEnabled = true
```
Lou1415926, line 99: could it happen that env has imported cert but ALB is disabled by `http: false` 🤔

efekarakus (approved):
LGTM with small feedback

Lou1415926 (approved):

internal/pkg/cli/deploy/deploy.go
iamhopaul123: We had a discussion [here](https://github.com/aws/copilot-cli/pull/3386#discussion_r850888516) about this

rverpillot:
I'm currently testing this nice new feature with a private domain but we need to keep the dns delegation feature.

iamhopaul123:
> I'm currently testing this nice new feature with a private domain but we need to keep the dns delegation feature.

Hello @rverpillot. Just want to make sure I understand the question. So do you want us to create the env hosted zone as we do if you don't import any certificates? Or do you have an existing hosted zone in the same account and region as the environment, and Copilot should just insert the ALIAS record?

efekarakus:
Hi Regis @rverpillot 😄 !

As a follow-up to Penghao's question:
- would you have different hosted zones per alias? or a single hosted zone for all aliases?
- is there ever a situation where you'd want Copilot to create the hosted zone for you?

```yaml
# Option 1  - where the aliases get written all to the same hosted zone.
http:
  hosted_zone_id: 'Z0873220N255IR3MTNR4'
  alias:
  - example.com
  - v2.example.com

# Option 2 - separate hosted_zone_id per alias
http:
  alias:
  - name: example.com
    hosted_zone_id: 'Z0873220N255IR3MTNR4'
  - name: v2.example.com
    hosted_zone_id: 'Z04546471FOA5FFVM1JHZ'
```

rverpillot:
Bonjour @iamhopaul123 and @efekarakus !

When a domain is defined in `app init`, copilot creates automatically public hosted zones with DNS delegation and inserts DNS alias for each services. But because we would like to use copilot with private VPC and private hosted zones, copilot should be able to create DNS aliases in a private hosted zone.

To answer @efekarakus question 1, option 1 should be sufficient, in my mind.

For question 2: could copilot create automatically a private hosted zone for the environment (like `env.myapp.myprivate-domain.com`) and add generated aliases in it ? Does it make sense (I'm not confident about that now and we still need to work on our private DNS model) ? 

To experiment this use case, we use a forked version of copilot (with dirty patches) allowing internal ALB, and restoring the `dnsDelegationEnabled` flag when importing certificates. Because copilot creates public hosted zones, DNS doesn't work, but everything else seems ok (we successfully access the test application from our on-prem network). So this PR is a great step.

We're going to dig the private DNS part to give you clear answers.

iamhopaul123:
Bonjour @rverpillot.

> could copilot create automatically a private hosted zone for the environment (like env.myapp.myprivate-domain.com) and add generated aliases in it ?

Just wanted to double check with you: so you need DNS delegation like `app init --domain privatedomain.com` and we'll create the app hosted zone and env hosted zones. And if you don't specify `alias` we would just use generated alias `svc.env.app.privatedomain.com` by default.

How would you prioritize this feature request? Do you think it would be good enough for us to ship the hosted zone id thing in the manifest to unblock you first? In that scenarios you would need to create the private hosted zone and certificate on your own and we'll take care of the cert import and adding A record to the hosted zone for the internal ALB DNS.

rverpillot:
Hi @iamhopaul123.

Yes. Create automatically private hosted zones and generated alias would be fine, because the CD for private VPC or the CD for public VPC would be similar. But you are right, if copilot can use a hosted zone id and creates A records, it would be fine for us because we'll use Ansible playbooks to fill gaps.

Just 2 remarks:
- I'm not sure that creating app private hosted zone makes sense (moreover it's not possible to do before knowing the VPC id).
- Wouldn't it be better to import the hosted zone id while creating environment instead of using it explicitly in the manifest ?

efekarakus:
> I'm not sure that creating app private hosted zone makes sense (moreover it's not possible to do before knowing the VPC id).

That's a good point 😄 

> Wouldn't it be better to import the hosted zone id while creating environment instead of using it explicitly in the manifest ?

We're starting the implementation for "environment manifest" and I think creating the private hosted zone by defining it in the manifest would make sense, for example:

```yaml
type: 'Environment'
domain:
  private: 'amazon.com' # creates a private hosted zone for amazon.com
  
network:
  vpc:
     azs: 3 
```

For now as a middle ground, we can have an option for the `hosted_zone_id` in the service manifest and in a future release make the `hosted_zone_id` optional if the environment has a private domain name 💭

rverpillot:
Because VPC and project accounts will not be the same, we will have to create private hosted zone before deploying applications.
So creating a private hosted zone by copilot is definitely not a priority anymore for our use case (but could be relevant for other use cases).

PR #3390, opened 2022-03-23
test(e2e): include pipeline name on pipeline init

dannyrandall:
Fix failed e2e test where we are not providing a name on `pipeline init`. Also refactor pipeline commands to match other commands.

### DIFF
--- a/e2e/internal/client/cli.go
+++ b/e2e/internal/client/cli.go
@@ -731,7 +731,7 @@ func (cli *CLI) PipelineShow(opts PipelineShowInput) (PipelineShowOutput, error)
 	return out, nil
 }
 
-// PipelineShow runs:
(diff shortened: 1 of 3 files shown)


e2e/internal/client/cli.go
```diff
@@ -214,6 +214,33 @@ type PackageInput struct {
 	Tag     string
 }
 
+// PipelineInitInput contains the parameters for calling copilot pipeline init.
+type PipelineInitInput struct {
+	AppName      string
```
Lou1415926: nit: If we don't need `AppName` then we can remove them from the structs for now!

Lou1415926 (approved):

efekarakus (approved):

PR #3391, opened 2022-03-23
chore(cli): allow pipeline deletion outside of workspace

### LINKED ISSUE #3299: Delete a pipeline from outside of a workspace
For now, Copilot tries to read the `pipeline.yml` file when deleting a pipeline. This imposes an implicit requirement that people can only run `copilot pipeline delete` under an established workspace. Note that this behavior also impacts `copilot app delete` because `copilot pipeline delete` is part of the operation. We should remove this implicit requirement.

huanjani:
Fixes https://github.com/aws/copilot-cli/issues/3299.

Previously, we were reading the pipeline manifest 1. to fetch the pipeline name and 2. to look for `access-token-secret`. In order to allow pipeline deletion from outside of the workspace (because it is more flexible and also makes `app delete` more useful), we cannot rely on reading the pipeline manifest because it is inaccessible from outside of the workspace.

Changes in https://github.com/aws/copilot-cli/pull/3382 leveraged the new DeployedPipelines selector in `pipeline delete` so the user can choose the pipeline they would like to delete. This PR solves the remaining problem of how to delete the pipeline secret. Prior to v.1.4 (introduction of CodeStar Connections), Copilot stored GitHub access tokens as secrets in AWS Secrets Manager. Without reading the manifest, we can look for secrets named and tagged with the conventions `pipeline init` followed for GHv1 source connections. If folks changed their pipeline names or secrets tags, we may not catch the secrets. We do log an error saying that we didn't find any secrets, so they are at least aware that no secrets are being deleted. It's obviously better to err on the side of not deleting than deleting. If we do find a secret that we think is a Copilot secret, we prompt for confirmation before proceeding (unless the user passed in the `--delete-secret` flag.

With a pipeline matching the defaults, pipeline name `pipeline-appName-repoName` and secret `github-token-appName-repoName` with tag `Key: "copilot-application"` `Value: [timeStampAtTimeOfInit]`:
```
 % copilot pipeline delete
Found secret 'github-token-muckles-aws-copilot-sample-service'.
Sure? Yes // confirming that the user wants to delete the pipeline

  Are you sure you want to delete the source secret github-token-muckles-aws-copilot-sample-service associated with pipeline pipeline-muckles-aws-copilot-sample-service? [? for help] (y/N) 
```
Without a pipeline match:
```
% copilot pipeline delete
Found no Secrets Manager secrets to delete for pipeline secondPipeline.
Sure? No
✘ pipeline delete cancelled - no changes made
```

### DIFF
--- a/internal/pkg/aws/secretsmanager/secretsmanager.go
+++ b/internal/pkg/aws/secretsmanager/secretsmanager.go
@@ -83,8 +83,14 @@ func (s *SecretsManager) DeleteSecret(secretName string) error {
 	return nil
 }
 
+type DescribeSecretOutput struct {
+	Name *string
+	CreatedDate *time.Time
+	Tags []*secretsmanager.Tag
+}
+
 // DescribeSecret retrieves the details of a secret.
-func (s *SecretsManager) DescribeSecret(secretName string) (*secretsmanager.DescribeSecretOutput, error) {
+func (s *SecretsManager) DescribeSecret(secretName string) (*DescribeSecretOutput, error) {
 	resp, err := s.secretsManager.DescribeSecret(&secretsmanager.DescribeSecretInput{
 		SecretId: aws.String(secretName),
 	})
@@ -99,7 +105,12 @@ func (s *SecretsManager) DescribeSecret(secretName string) (*secretsmanager.Desc
 		}
 		return nil, fmt.Errorf("describe secret %s: %w", secretName, err)
 	}
-	return resp, nil
+
+	return &DescribeSecretOutput{
+		Name:        resp.Name,
+		CreatedDate: resp.CreatedDate,
+		Tags:        resp.Tags,
+	}, nil
 }
 
 // ErrSecretAlreadyExists occurs if a secret with the same name already exists.
--- a/internal/pkg/aws/secretsmanager/secretsmanager.go
+++ b/internal/pkg/aws/secretsmanager/secretsmanager.go
@@ -84,9 +84,9 @@ func (s *SecretsManager) DeleteSecret(secretName string) error {
 }
 
 type DescribeSecretOutput struct {
-	Name *string
+	Name        *string
 	CreatedDate *time.Time
-	Tags []*secretsmanager.Tag
+	Tags        []*secretsmanager.Tag
 }
 
 // DescribeSecret retrieves the details of a secret.
--- a/internal/pkg/aws/secretsmanager/secretsmanager.go
+++ b/internal/pkg/aws/secretsmanager/secretsmanager.go
@@ -130,5 +130,5 @@ type ErrSecretNotFound struct {
 }
 
 func (err *ErrSecretNotFound) Error() string {
-	return fmt.Sprintf("secret %s was not found", err.secretName)
+	return fmt.Sprintf("secret %s was not found: %s", err.secretName, err.parentErr)
 }
--- a/internal/pkg/aws/secretsmanager/secretsmanager.go
+++ b/internal/pkg/aws/secretsmanager/secretsmanager.go
@@ -18,6 +18,7 @@ import (
 type api interface {
 	CreateSecret(*secretsmanager.CreateSecretInput) (*secretsmanager.CreateSecretOutput, error)
 	DeleteSecret(*secretsmanager.DeleteSecretInput) (*secretsmanager.DeleteSecretOutput, error)
+	DescribeSecret(input *secretsmanager.DescribeSecretInput) (*secretsmanager.DescribeSecretOutput, error)
 }
 
 // SecretsManager wraps the AWS SecretManager client.
@@ -82,6 +83,25 @@ func (s *SecretsManager) DeleteSecret(secretName string) error {
 	return nil
 }
 
+// DescribeSecret retrieves the details of a secret.
+func (s *SecretsManager) DescribeSecret(secretName string) (*secretsmanager.DescribeSecretOutput, error) {
+	resp, err := s.secretsManager.DescribeSecret(&secretsmanager.DescribeSecretInput{
+		SecretId: aws.String(secretName),
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			if aerr.Code() == secretsmanager.ErrCodeResourceNotFoundException {
+				return nil, &ErrSecretNotFound{
+					secretName: secretName,
+					parentErr: err,
+				}
+			}
+		}
+		return nil, fmt.Errorf("describe secret %s: %w", secretName, err)
+	}
+	return resp, nil
+}
+
 // ErrSecretAlreadyExists occurs if a secret with the same name already exists.
 type ErrSecretAlreadyExists struct {
 	secretName string
@@ -90,4 +110,13 @@ type ErrSecretAlreadyExists struct {
 
 func (err *ErrSecretAlreadyExists) Error() string {
 	return fmt.Sprintf("secret %s already exists", err.secretName)
+}
+// ErrSecretNotFound occurs if a secret with the given name does not exist.
+type ErrSecretNotFound struct {
+	secretName string
+	parentErr  error
+}
+
+func (err *ErrSecretNotFound) Error() string {
+	return fmt.Sprintf("secret %s was not found", err.secretName)
 }
--- a/internal/pkg/aws/secretsmanager/secretsmanager.go
+++ b/internal/pkg/aws/secretsmanager/secretsmanager.go
@@ -78,7 +78,7 @@ func (s *SecretsManager) DeleteSecret(secretName string) error {
 	})
 
 	if err != nil {
-		return fmt.Errorf("delete secret %s from secrets manager: %+v", secretName, err)
+		return fmt.Errorf("delete secret %s from secrets manager: %w", secretName, err)
 	}
 	return nil
 }
@@ -93,7 +93,7 @@ func (s *SecretsManager) DescribeSecret(secretName string) (*secretsmanager.Desc
 			if aerr.Code() == secretsmanager.ErrCodeResourceNotFoundException {
 				return nil, &ErrSecretNotFound{
 					secretName: secretName,
-					parentErr: err,
+					parentErr:  err,
 				}
 			}
 		}
@@ -111,6 +111,7 @@ type ErrSecretAlreadyExists struct {
 func (err *ErrSecretAlreadyExists) Error() string {
 	return fmt.Sprintf("secret %s already exists", err.secretName)
 }
+
 // ErrSecretNotFound occurs if a secret with the given name does not exist.
 type ErrSecretNotFound struct {
 	secretName string
--- a/internal/pkg/aws/secretsmanager/secretsmanager_test.go
+++ b/internal/pkg/aws/secretsmanager/secretsmanager_test.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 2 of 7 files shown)


internal/pkg/aws/secretsmanager/secretsmanager.go
```diff
@@ -77,11 +78,30 @@ func (s *SecretsManager) DeleteSecret(secretName string) error {
 	})
 
 	if err != nil {
-		return fmt.Errorf("delete secret %s from secrets manager: %+v", secretName, err)
+		return fmt.Errorf("delete secret %s from secrets manager: %w", secretName, err)
```
dannyrandall, line 81: good catch😌

internal/pkg/aws/secretsmanager/secretsmanager.go
```diff
@@ -77,11 +78,30 @@ func (s *SecretsManager) DeleteSecret(secretName string) error {
...
 	return nil
 }
 
+// DescribeSecret retrieves the details of a secret.
+func (s *SecretsManager) DescribeSecret(secretName string) (*secretsmanager.DescribeSecretOutput, error) {
+	resp, err := s.secretsManager.DescribeSecret(&secretsmanager.DescribeSecretInput{
+		SecretId: aws.String(secretName),
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
```
dannyrandall, line 98: nit: `errors.As` instead (example in [errors package overview](https://pkg.go.dev/errors))

internal/pkg/cli/pipeline_delete.go
```diff
@@ -176,21 +179,34 @@ func (o *deletePipelineOpts) askAppName() error {
 }
 
 func (o *deletePipelineOpts) getSecret() error {
-	path, err := o.ws.PipelineManifestLegacyPath()
-	if err != nil {
-		return fmt.Errorf("get path to pipeline manifest: %w", err)
+	// Look for default secret name for GHv1 access token based on default pipeline name.
+	o.ghAccessTokenSecretName = o.pipelineSecretName()
+	output, err := o.secretsmanager.DescribeSecret(o.ghAccessTokenSecretName)
+	if err == nil {
```
dannyrandall: what do you think about swapping the error handling to handle `err != nil` first?

internal/pkg/cli/pipeline_delete.go
```diff
@@ -176,21 +179,34 @@ func (o *deletePipelineOpts) askAppName() error {
...
 	}
+	// To get here, either the secret was found but tags didn't match, or Secrets Manager returned a ResourceNotFoundException.
+	log.Infof("Found no Copilot-generated Secrets Manager secrets to delete for pipeline '%s'.\n", o.name)
+	o.ghAccessTokenSecretName = ""
 	return nil
 }
 
+// With GHv1 sources, we stored access tokens in SecretsManager. Pipelines generated prior to Copilot v1.4 have secrets named 'github-token-[appName]-[repoName]'. Pipelines prior to 1.16(?) were given default names of `pipeline-[appName]-[repoName]`. Users may have changed pipeline names, so this is our best-guess approach to deleting legacy pipeline secrets.
+func (o *deletePipelineOpts) pipelineSecretName() string {
+	appAndRepo := strings.TrimPrefix(o.name,"pipeline-")
```
dannyrandall: super super nit: gofmt should add a space after `,` :)

internal/pkg/aws/secretsmanager/secretsmanager_test.go
```diff
@@ -105,3 +106,145 @@ func TestSecretsManager_CreateSecret(t *testing.T) {
 		})
 	}
 }
+
+func TestSecretsManager_DeleteSecret(t *testing.T) {
```
dannyrandall, line 110: thanks for adding some tests here 🎉

internal/pkg/cli/pipeline_delete_test.go
```diff
@@ -238,19 +185,41 @@ func TestDeletePipelineOpts_Ask(t *testing.T) {
...
-				require.Equal(t, tc.wantedSecret, opts.ghAccessTokenSecretName, "expected secrets to match")
 			}
 		})
 	}
 }
 
 func TestDeletePipelineOpts_Execute(t *testing.T) {
+	mockResp := &sdkSecretsmanager.DescribeSecretOutput{
+		ARN:         aws.String("arn-goose"),
+		CreatedDate: aws.Time(time.Now()),
```
dannyrandall: i wonder if we should create a `now := time.Now()` at the top of the function to get reused. i don't think there's any guarantee that the struct will get created at exact the same time (although in some light testing, it usually is) so i'm worried the test may fail randomly

dannyrandall (commented):
you know when you look at a word for a long time and it starts to look like it's spelled wrong? i could not get over thinking secret was misspelled, but you def have it right :joy:

internal/pkg/aws/secretsmanager/secretsmanager.go
```diff
@@ -91,3 +111,13 @@ type ErrSecretAlreadyExists struct {
...
 }
+
+// ErrSecretNotFound occurs if a secret with the given name does not exist.
+type ErrSecretNotFound struct {
+	secretName string
+	parentErr  error
+}
+
+func (err *ErrSecretNotFound) Error() string {
+	return fmt.Sprintf("secret %s was not found", err.secretName)
```
Lou1415926: nit: perhaps we can wrap `parentErr` into this err message as well.

internal/pkg/aws/secretsmanager/secretsmanager.go
```diff
@@ -77,11 +78,30 @@ func (s *SecretsManager) DeleteSecret(secretName string) error {
...
 
 	if err != nil {
-		return fmt.Errorf("delete secret %s from secrets manager: %+v", secretName, err)
+		return fmt.Errorf("delete secret %s from secrets manager: %w", secretName, err)
 	}
 	return nil
 }
 
+// DescribeSecret retrieves the details of a secret.
+func (s *SecretsManager) DescribeSecret(secretName string) (*secretsmanager.DescribeSecretOutput, error) {
```
Lou1415926: nit: we could probably do something to `secretsmanager.DescribeSecretOutput` similar to what we do with [`ecs.Task`](https://github.com/aws/copilot-cli/blob/7962398baf9c6d2685e195873cbb35a5499690dc/internal/pkg/aws/ecs/task.go#L50).

Or alternatively, return only secret info that we think we need:
```go
type Secret struct{
   Name string
   Tag []*secretsmanager.Tag
}

func (s *SecretsManager) DescribeSecret(secretName string) (*Secret, error) {
...
  return &Secret{
     Name: resp.Name,
     Tags: resp.Tags,
  }, nil
}
```

Just so that `cli` isn't directly depending on `aws-sdk-go`!

internal/pkg/aws/secretsmanager/secretsmanager_test.go
```diff
@@ -105,3 +106,145 @@ func TestSecretsManager_CreateSecret(t *testing.T) {
...
+			resp, err := sm.DescribeSecret(tc.inSecretName)
+
+			// THEN
+			if tc.expectedError != nil {
+				require.EqualError(t, err, tc.expectedError.Error())
+			} else {
+				require.NoError(t, err)
+				require.Equal(t, tc.expectedResp, resp)
+			}
+			require.Equal(t, tc.expectedError, err)
```
Lou1415926: Do we still need this line?

internal/pkg/aws/secretsmanager/secretsmanager_test.go
```diff
@@ -105,3 +106,145 @@ func TestSecretsManager_CreateSecret(t *testing.T) {
...
+			mockSecretsManager := mocks.NewMockapi(ctrl)
+
+			sm := SecretsManager{
+				secretsManager: mockSecretsManager,
+			}
+
+			tc.callMock(mockSecretsManager)
+
+			// WHEN
+			oldSecretTags := secretTags
```
Lou1415926, line 236: This reminds me that we are now automatically tagging all secretsmanager secrets with timestamp for `copilot-application` tag.

I think we should ditch this behavior and instead let the client decide how to tag, maybe with a similar approach to [this example](https://github.com/aws/copilot-cli/blob/7962398baf9c6d2685e195873cbb35a5499690dc/internal/pkg/deploy/cloudformation/stack/lb_web_svc.go#L66). 

Not necessarily within the scope of this particular PR though - just to call it out.

internal/pkg/cli/pipeline_delete.go
huanjani: I had it that way originally, but then I had two instances of logging that long info line and resetting `o.ghAccessTokenSecretName` back to an empty string (once if secret not found, again if secret found but without the matching tags), so this seemed slightly more efficient.

internal/pkg/cli/pipeline_delete.go
huanjani: Good eyeballs!

internal/pkg/cli/pipeline_delete_test.go
huanjani: That's a great idea!

internal/pkg/aws/secretsmanager/secretsmanager.go
efekarakus: can we define our own `DescribeSecretOutput` even if it's just:
```go
type DescribeSecretOutput secretsmanager.DescribeSecretOutput
```
This way clients that define an interface won't have to import both our `secretsmanager` pkg and the AWS SDK's `secretsmanager`

internal/pkg/aws/secretsmanager/secretsmanager.go
```diff
@@ -91,3 +111,13 @@ type ErrSecretAlreadyExists struct {
 func (err *ErrSecretAlreadyExists) Error() string {
 	return fmt.Sprintf("secret %s already exists", err.secretName)
 }
+
+// ErrSecretNotFound occurs if a secret with the given name does not exist.
+type ErrSecretNotFound struct {
+	secretName string
+	parentErr  error
```
efekarakus, line 129: nit: looks like we don't use this field, should we remove it?

internal/pkg/cli/pipeline_delete.go
```diff
@@ -176,21 +179,34 @@ func (o *deletePipelineOpts) askAppName() error {
...
-	path, err := o.ws.PipelineManifestLegacyPath()
-	if err != nil {
-		return fmt.Errorf("get path to pipeline manifest: %w", err)
+	// Look for default secret name for GHv1 access token based on default pipeline name.
+	o.ghAccessTokenSecretName = o.pipelineSecretName()
+	output, err := o.secretsmanager.DescribeSecret(o.ghAccessTokenSecretName)
+	if err == nil {
+		// Check for Copilot-assigned tag for added assurance. If tags not found, fall through.
+		for _, tag := range output.Tags {
+			if aws.StringValue(tag.Key) == "copilot-application" && aws.StringValue(tag.Value) == output.CreatedDate.UTC().Format(time.UnixDate) {
```
efekarakus: 😞

internal/pkg/cli/pipeline_delete.go
```diff
@@ -176,21 +179,34 @@ func (o *deletePipelineOpts) askAppName() error {
...
-	}
-
-	if secret, ok := (manifest.Source.Properties["access_token_secret"]).(string); ok {
-		o.ghAccessTokenSecretName = secret
+	var notFoundErr *secretsmanager.ErrSecretNotFound
+	if err != nil && !errors.As(err, &notFoundErr) {
+		return fmt.Errorf("describe secret %s: %w", o.ghAccessTokenSecretName, err)
 	}
+	// To get here, either the secret was found but tags didn't match, or Secrets Manager returned a ResourceNotFoundException.
+	log.Infof("Found no Copilot-generated Secrets Manager secrets to delete for pipeline '%s'.\n", o.name)
```
efekarakus: Should we remove this log? most users will fall into this category and it seems like it might spook them off for no reason

internal/pkg/cli/pipeline_delete.go
```diff
@@ -176,21 +179,34 @@ func (o *deletePipelineOpts) askAppName() error {
...
+	o.ghAccessTokenSecretName = ""
 	return nil
 }
 
+// With GHv1 sources, we stored access tokens in SecretsManager. Pipelines generated prior to Copilot v1.4 have secrets named 'github-token-[appName]-[repoName]'. Pipelines prior to 1.16(?) were given default names of `pipeline-[appName]-[repoName]`. Users may have changed pipeline names, so this is our best-guess approach to deleting legacy pipeline secrets.
```
efekarakus: nit: can we split this comment into multiple lines 😂 Maybe instead of prior to v1.16 we can give the short git commit sha

efekarakus (commented):
Looks great, just some tiny feedback

internal/pkg/cli/pipeline_delete.go
```diff
@@ -176,21 +179,34 @@ func (o *deletePipelineOpts) askAppName() error {
...
-	if err != nil {
-		return fmt.Errorf("get path to pipeline manifest: %w", err)
+	// Look for default secret name for GHv1 access token based on default pipeline name.
+	o.ghAccessTokenSecretName = o.pipelineSecretName()
+	output, err := o.secretsmanager.DescribeSecret(o.ghAccessTokenSecretName)
+	if err == nil {
+		// Check for Copilot-assigned tag for added assurance. If tags not found, fall through.
+		for _, tag := range output.Tags {
+			if aws.StringValue(tag.Key) == "copilot-application" && aws.StringValue(tag.Value) == output.CreatedDate.UTC().Format(time.UnixDate) {
+				log.Infof("Found secret '%s'.\n", o.ghAccessTokenSecretName)
```
Lou1415926: We might need to add some explanation of why we are looking for a secret to delete in these log info. People that are using non-GHV1 might find it super confusing and unsettling 😭

internal/pkg/cli/pipeline_delete.go
Lou1415926: Samesies here!

internal/pkg/aws/secretsmanager/secretsmanager.go
huanjani: Good idea! Done.

huanjani: Haha, yup-- had already done just that based on Efe's comment above.

internal/pkg/aws/secretsmanager/secretsmanager_test.go
huanjani: This behavior is only for pipeline secrets, though, right? 
I'd be surprised if there are many, if any at all, users passing in their github tokens via flags. That is the only situation in which we're creating these secrets for pipelines these days.

internal/pkg/cli/pipeline_delete.go
huanjani: The only people who would see this log info are GHv1 people, unless they just _happened_ to name a secret `github-token-appName-repoName` and also tagged it `copilot-application: [timestamp]`.

But yes, they may not have known/noticed that the secret was created for them when they created their pipeline, so I'll add some explanation for the GHv1 people! 😄

internal/pkg/cli/pipeline_delete.go
huanjani: I may switch it again if I get rid of the log.info altogether....

internal/pkg/cli/pipeline_delete.go
iamhopaul123: Maybe we should use `deploy.AppTagKey`?

internal/pkg/aws/secretsmanager/secretsmanager_test.go
Lou1415926: I guess my point was mostly from the point of view of improving code structure, like how `secretsmanager` as a mere client should be agnostic of concepts such as `copilot-application` tag. 

For example, in the future, when we add `secretsmanager` to `copilot secret init`, we will most likely reuse `CreateSecret` function as well. It won't make sense for those secrets to be tagged with `copilot-application: <timestamp>`.

But again this is just a callout!

internal/pkg/aws/secretsmanager/secretsmanager_test.go
huanjani: good catch!

internal/pkg/cli/pipeline_delete.go
huanjani: 👻 🧟  I was mostly worried about GHv1 users who didn't follow use defaults having lingering undeleted secrets that they don't know about and are paying for. But you're both right-- tiny use case compared to the majority of users.

huanjani: Yes-- rather than adding more explanation to this one, I'll just delete it, as most current users will fall into this category.

internal/pkg/cli/pipeline_delete.go
huanjani: Oh cool, thanks. I didn't know that existed! 😆

internal/pkg/aws/secretsmanager/secretsmanager.go
huanjani: Cool. I am using `errors.As` in the client, `pipeline_delete`, but ran into some hiccups trying to use it here...will revisit it after this push! Thx!

internal/pkg/aws/secretsmanager/secretsmanager.go
huanjani: We use it in `pipeline_delete.go`.... Will look into streamlining this after this push! 😸

internal/pkg/cli/interfaces.go
```diff
@@ -5,10 +5,10 @@ package cli
 
 import (
 	"encoding"
+	"github.com/aws/copilot-cli/internal/pkg/aws/secretsmanager"
 	"io"
 
 	"github.com/aws/aws-sdk-go/aws/session"
-	"github.com/aws/aws-sdk-go/service/secretsmanager"
```
efekarakus: 🥳

internal/pkg/cli/pipeline_delete.go
```diff
@@ -182,28 +182,33 @@ func (o *deletePipelineOpts) getSecret() error {
...
-	var notFoundErr *secretsmanager.ErrSecretNotFound
-	if err != nil && !errors.As(err, &notFoundErr) {
 		return fmt.Errorf("describe secret %s: %w", o.ghAccessTokenSecretName, err)
 	}
-	// To get here, either the secret was found but tags didn't match, or Secrets Manager returned a ResourceNotFoundException.
-	log.Infof("Found no Copilot-generated Secrets Manager secrets to delete for pipeline '%s'.\n", o.name)
+
+	for _, tag := range output.Tags {
+		if aws.StringValue(tag.Key) == deploy.AppTagKey && aws.StringValue(tag.Value) == output.CreatedDate.UTC().Format(time.UnixDate) {
+			log.Infof("Found the Copilot-generated secret '%s' used to connect to your GitHub source repo.\n", o.ghAccessTokenSecretName)
```
efekarakus: I feel like since we are going to prompt them to confirm the removal, we can even get rid of this log

efekarakus (approved):

internal/pkg/aws/secretsmanager/secretsmanager_test.go
```diff
@@ -105,3 +106,148 @@ func TestSecretsManager_CreateSecret(t *testing.T) {
...
+			inSecretName: mockSecretName,
+			callMock: func(m *mocks.Mockapi) {
+				m.EXPECT().DescribeSecret(&secretsmanager.DescribeSecretInput{
+					SecretId: aws.String(mockSecretName),
+				}).Return(nil, mockError)
+			},
+			expectedError: fmt.Errorf("describe secret %s: %w", mockSecretName, mockError),
+		},
+
+		"should return no error if secret is not found": {
```
Lou1415926, line 196: nit: 
```suggestion
		"should return custom err if secret is not found": {
```

Lou1415926 (approved):

internal/pkg/aws/secretsmanager/secretsmanager_test.go
huanjani: 🤔  but we don't error out if the secret isn't found; we only error out with other kinds of errors.

PR #3401, opened 2022-03-28
chore(cli): check for duplicate pipelines on `pipeline init`

dannyrandall:
Example output on pipeline init with a local duplicate:
```
✘ It seems like you are trying to init a pipeline that exists,
but has not been deployed. To deploy this pipeline, please run:
`copilot pipeline deploy --name danny`
If you'd prefer a new default manifest, please manually delete the existing file
and then `copilot pipeline init --name danny`
✘ pipeline danny already exists
```

In this PR, I kept the tests in the same format (a separate struct field for each mocked type), but will be creating a PR to refactor that format to match [this](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/cli/svc_init_test.go#L48), which should clean up all of the empty functions floating around.

### DIFF
--- a/internal/pkg/aws/codepipeline/codepipeline.go
+++ b/internal/pkg/aws/codepipeline/codepipeline.go
@@ -19,6 +19,8 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/term/color"
 )
 
+var ErrPipelineNotFound = errors.New("pipeline not found")
+
 const (
 	pipelineResourceType = "codepipeline:pipeline"
 )
@@ -114,12 +116,15 @@ func New(s *session.Session) *CodePipeline {
 
 // GetPipeline retrieves information from a given pipeline.
 func (c *CodePipeline) GetPipeline(name string) (*Pipeline, error) {
-	input := &cp.GetPipelineInput{
+	resp, err := c.client.GetPipeline(&cp.GetPipelineInput{
 		Name: aws.String(name),
-	}
-	resp, err := c.client.GetPipeline(input)
-
+	})
 	if err != nil {
+		var notFoundErr *cp.PipelineNotFoundException
+		if errors.As(err, &notFoundErr) {
+			return nil, fmt.Errorf("get pipeline %s: %w", name, ErrPipelineNotFound)
+		}
+
 		return nil, fmt.Errorf("get pipeline %s: %w", name, err)
 	}
 
--- a/internal/pkg/aws/codepipeline/codepipeline.go
+++ b/internal/pkg/aws/codepipeline/codepipeline.go
@@ -123,7 +123,7 @@ func (c *CodePipeline) GetPipeline(name string) (*Pipeline, error) {
 	if err != nil {
 		var notFoundErr *cp.PipelineNotFoundException
 		if errors.As(err, &notFoundErr) {
-			return nil, fmt.Errorf("get pipeline %s: %w", name, ErrPipelineNotFound)
+			return nil, ErrPipelineNotFound
 		}
 
 		return nil, fmt.Errorf("get pipeline %s: %w", name, err)
--- a/internal/pkg/aws/codepipeline/codepipeline.go
+++ b/internal/pkg/aws/codepipeline/codepipeline.go
@@ -19,6 +19,7 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/term/color"
 )
 
+// ErrPipelineNotFound is returned if a given pipeline cannot be found.
 var ErrPipelineNotFound = errors.New("pipeline not found")
 
 const (
--- a/internal/pkg/aws/codepipeline/codepipeline.go
+++ b/internal/pkg/aws/codepipeline/codepipeline.go
@@ -118,6 +118,7 @@ func (c *CodePipeline) GetPipeline(name string) (*Pipeline, error) {
 		Name: aws.String(name),
 	}
 	resp, err := c.client.GetPipeline(input)
+
 	if err != nil {
 		return nil, fmt.Errorf("get pipeline %s: %w", name, err)
 	}
--- a/internal/pkg/aws/codepipeline/codepipeline.go
+++ b/internal/pkg/aws/codepipeline/codepipeline.go
@@ -19,9 +19,6 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/term/color"
 )
 
-// ErrPipelineNotFound is returned if a given pipeline cannot be found.
-var ErrPipelineNotFound = errors.New("pipeline not found")
-
 const (
 	pipelineResourceType = "codepipeline:pipeline"
 )
@@ -117,15 +114,11 @@ func New(s *session.Session) *CodePipeline {
 
 // GetPipeline retrieves information from a given pipeline.
 func (c *CodePipeline) GetPipeline(name string) (*Pipeline, error) {
-	resp, err := c.client.GetPipeline(&cp.GetPipelineInput{
+	input := &cp.GetPipelineInput{
 		Name: aws.String(name),
-	})
+	}
+	resp, err := c.client.GetPipeline(input)
 	if err != nil {
-		var notFoundErr *cp.PipelineNotFoundException
-		if errors.As(err, &notFoundErr) {
-			return nil, ErrPipelineNotFound
-		}
-
 		return nil, fmt.Errorf("get pipeline %s: %w", name, err)
 	}
 
--- a/internal/pkg/cli/interfaces.go
+++ b/internal/pkg/cli/interfaces.go
@@ -5,8 +5,9 @@ package cli
 
 import (
 	"encoding"
-	"github.com/aws/copilot-cli/internal/pkg/manifest"
 	"io"
+
+	"github.com/aws/copilot-cli/internal/pkg/manifest"
 
 	"github.com/aws/copilot-cli/internal/pkg/aws/ec2"
 
@@ -241,8 +242,8 @@ type wsPipelineManifestReader interface {
 }
 
 type wsPipelineWriter interface {
-	WritePipelineBuildspec(marshaler encoding.BinaryMarshaler) (string, error)
-	WritePipelineManifest(marshaler encoding.BinaryMarshaler) (string, error)
+	WritePipelineBuildspec(marshaler encoding.BinaryMarshaler, name string) (string, error)
+	WritePipelineManifest(marshaler encoding.BinaryMarshaler, name string) (string, error)
 }
 
 type serviceLister interface {
@@ -479,7 +480,12 @@ type pipelineEnvSelector interface {
 }
 
 type wsPipelineSelector interface {
-	Pipeline(prompt, help string) (*workspace.PipelineManifest, error)
+	WsPipeline(prompt, help string) (*workspace.PipelineManifest, error)
+}
+
+type codePipelineSelector interface {
+	appSelector
+	DeployedPipeline(prompt, help string, tags map[string]string) (string, error)
 }
 
 type wsSelector interface {
--- a/internal/pkg/cli/interfaces.go
+++ b/internal/pkg/cli/interfaces.go
@@ -5,8 +5,9 @@ package cli
 
 import (
 	"encoding"
-	"github.com/aws/copilot-cli/internal/pkg/aws/secretsmanager"
 	"io"
+
+	"github.com/aws/copilot-cli/internal/pkg/aws/secretsmanager"
 
 	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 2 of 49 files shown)


internal/pkg/aws/codepipeline/codepipeline.go
```diff
@@ -19,6 +19,8 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/term/color"
 )
 
+var ErrPipelineNotFound = errors.New("pipeline not found")
```
iamhopaul123: Comment for public vars pls 🙏

internal/pkg/cli/pipeline_init.go
```diff
@@ -192,6 +201,49 @@ func (o *initPipelineOpts) Ask() error {
...
+2. And then %s
+`,
+			color.HighlightCode(fmt.Sprintf("copilot pipeline delete --name %s", o.name)),
+			color.HighlightCode(fmt.Sprintf("copilot pipeline init --name %s", o.name)))
+		return fmt.Errorf("pipeline %s already exists", color.HighlightUserInput(o.name))
+	case !errors.Is(err, codepipeline.ErrPipelineNotFound):
+		return fmt.Errorf("validate if pipeline exists: %w", err)
+	}
+
+	// make sure pipeline doesn't exist locally
```
iamhopaul123, line 230: I wonder if we should strictly check their local files to see if it is already there. Reason is because `pipeline init` only writes local files. I wonder if it is expected if they have a pipeline deployed named `myPipeline` and then deleted the local manifest file. If they run `pipeline init -n myPipeline` we'll just create the manifest back for them. They could still tweak the manifest and then run `pipeline update` to update the stack 🤔

internal/pkg/cli/interfaces.go
```diff
@@ -243,6 +244,10 @@ type wsPipelineWriter interface {
 	WritePipelineManifest(marshaler encoding.BinaryMarshaler, name string) (string, error)
 }
 
+type wsPipelineLister interface {
```
paragbhingre: How about using the existing interface's (`wsPipelineGetter`) `ListPipelines()` method instead of creating new interface? Or there is specific use of this separate interface in the future?

internal/pkg/cli/pipeline_init.go
dannyrandall: That's an interesting thought 🤔 . Currently, if a service manifest file gets deleted, we block `svc init`, but you're right, `pipeline init` only writes local files, so we could handle it a little different.

I think the potential options in your situation are:
1. Error out (current outcome)
2. Pull the current pipeline configuration and recreate that manifest (no log, may not be possible without storing some info in SSM)
3. Generate a fresh pipeline configuration (log)

I think the best approach is what's currently written - to block `pipeline init` if there is a pipeline that's already deployed with that name, but I could be swayed to either of the other options if you think another option would be better. Thoughts?

internal/pkg/cli/pipeline_init.go
```diff
@@ -192,6 +201,49 @@ func (o *initPipelineOpts) Ask() error {
 	return nil
 }
 
+func (o *initPipelineOpts) validateDuplicatePipeline() error {
+	// make sure pipeline isn't already deployed
+	fullName := fmt.Sprintf(fmtPipelineName, o.appName, o.name)
+
+	_, err := o.pipeline.GetPipeline(fullName)
```
paragbhingre: Just curious to know if there is any priority order for calling `o.pipeline.GetPipeline(fullName)` vs `o.wsLister.ListPipelines()`? If we can do local check 1st and can eliminate the error case then only we can make an API call to check further is what my understanding is. Please correct me if I am missing something.

paragbhingre (commented):
LGTM, except couple questions.

internal/pkg/aws/codepipeline/codepipeline.go
```diff
@@ -114,12 +117,15 @@ func New(s *session.Session) *CodePipeline {
...
+	resp, err := c.client.GetPipeline(&cp.GetPipelineInput{
 		Name: aws.String(name),
-	}
-	resp, err := c.client.GetPipeline(input)
-
+	})
 	if err != nil {
+		var notFoundErr *cp.PipelineNotFoundException
+		if errors.As(err, &notFoundErr) {
+			return nil, fmt.Errorf("get pipeline %s: %w", name, ErrPipelineNotFound)
```
efekarakus: nit: the clients will most likely wrap w `get pipeline` then this second one becomes a bit redundant
```suggestion
			return nil, ErrPipelineNotFound
```

efekarakus: Can we add a test case to capture this error

internal/pkg/cli/pipeline_init.go
dannyrandall: Yeah - It would be nice to avoid the unnecessary API call. However, my thought was that if we did the local check (`ListPipelines()`) before the remote check (`GetPipeline()`) and they put in a name that has already been deployed, we'd see the local conflict but still need to do the remote check in order to print out the correct error message (whether they need to `delete` the pipeline or if they have yet to `deploy` it)

Does that thinking make sense? Or am I crazy lol

internal/pkg/cli/pipeline_init.go
```diff
@@ -94,6 +95,7 @@ type initPipelineOpts struct {
 	initPipelineVars
 	// Interfaces to interact with dependencies.
 	workspace      wsPipelineWriter
+	wsLister       wsPipelineLister
```
efekarakus: Instead of adding an additional field, should we update `workspace` to have a bigger interface?
```go
workspace wsPipelineIniter

type wsPipelineIniter interface {
  ... // all workspace pipeline methods to initialize a pipeline.
}
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -102,6 +104,7 @@ type initPipelineOpts struct {
 	store          store
 	prompt         prompter
 	sel            pipelineEnvSelector
+	pipeline       pipelineGetter
```
efekarakus: nit:
```suggestion
	codePipeline       pipelineGetter
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -192,6 +201,49 @@ func (o *initPipelineOpts) Ask() error {
...
+		if pipeline.Name == o.name {
+			log.Errorf(`It seems like you are trying to init a pipeline that exists,
+but has not been deployed. To deploy this pipeline, please run:
+%s
+If you'd prefer a new default manifest, please manually delete the existing file
+and then %s
+`,
+				color.HighlightCode(fmt.Sprintf("copilot pipeline deploy --name %s", o.name)),
+				color.HighlightCode(fmt.Sprintf("copilot pipeline init --name %s", o.name)))
+			return fmt.Errorf("pipeline %s already exists", color.HighlightUserInput(o.name))
```
efekarakus: nit:
```suggestion
			return fmt.Errorf("pipeline %s's manifest already exists", color.HighlightUserInput(o.name))
```

efekarakus (commented):
Looks great! a bunch of tiny things

internal/pkg/cli/pipeline_init.go
```diff
@@ -192,6 +201,49 @@ func (o *initPipelineOpts) Ask() error {
 	return nil
 }
 
+func (o *initPipelineOpts) validateDuplicatePipeline() error {
+	// make sure pipeline isn't already deployed
+	fullName := fmt.Sprintf(fmtPipelineName, o.appName, o.name)
```
Lou1415926: I wonder if we should check the existence of both
1. pipeline-app-pipelineName; and
2. pipelineName?

internal/pkg/cli/pipeline_init.go
paragbhingre: Either this works or we can use `ListPipelines()` from `wsPipelineGetter`

dannyrandall: i'll take this approach since `pipeline init` is the only place that currently uses `wsPipelineWriter`

internal/pkg/cli/pipeline_init.go
iamhopaul123: OK it makes sense to me!

internal/pkg/cli/interfaces.go
dannyrandall: Resolved in https://github.com/aws/copilot-cli/pull/3401#discussion_r836934744

internal/pkg/cli/pipeline_init.go
dannyrandall: This is a great idea! Should help avoid confusion between legacy and new pipelines.

internal/pkg/aws/codepipeline/codepipeline.go
dannyrandall: jk. i actually removed this error since i'm using `ListPipelineNamesByTags()` instead now

internal/pkg/cli/pipeline_init.go
paragbhingre: It does make sense to me, you are not crazy 🤩

internal/pkg/cli/pipeline_init.go
```diff
@@ -192,6 +199,57 @@ func (o *initPipelineOpts) Ask() error {
...
+	}
+
+	fullName := fmt.Sprintf(fmtPipelineName, o.appName, o.name)
+	for _, name := range names {
+		if strings.EqualFold(name, o.name) || strings.EqualFold(name, fullName) {
+			log.Errorf(`It seems like you are trying to init a pipeline that already exists.
+To recreate the pipeline, please run:
+1. %s. Note: The manifest file will not be deleted and will be used in Step 2.
+If you'd prefer a new default manifest, please manually delete the existing file.
+2. And then %s
```
huanjani: nit: can we add a verb after "then"? maybe "run"?

iamhopaul123 (approved):

internal/pkg/cli/pipeline_init.go
dannyrandall: see what you think about it now :)

internal/pkg/cli/pipeline_init.go
```diff
@@ -192,6 +199,57 @@ func (o *initPipelineOpts) Ask() error {
 	return nil
 }
 
+// validateDuplicatePipeline checks that the pipeline name isn't already used
+// by another pipeline, whether it's been deployed or just has a local manifest file.
+// We check for the existance of the name and the namespaced name to reduce
```
huanjani: tiny tiny nit: 
```suggestion
// We check for the existence of the name and the namespaced name to reduce
```

internal/pkg/cli/pipeline_init.go
huanjani: I agree, bc if they are naming a new pipeline "repoName", it would not be in a local manifest (because "pipeline-appName-repoName" would), but would end up with a namespace clash upon deployment.

internal/pkg/cli/pipeline_init.go
```diff
@@ -192,6 +199,57 @@ func (o *initPipelineOpts) Ask() error {
...
+	}
+
+	// make sure pipeline doesn't exist locally
+	pipelines, err := o.workspace.ListPipelines()
+	if err != nil {
+		return fmt.Errorf("get local pipelines: %w", err)
+	}
+
+	for _, pipeline := range pipelines {
+		if strings.EqualFold(pipeline.Name, o.name) || strings.EqualFold(pipeline.Name, fullName) {
```
huanjani, line 237: EqualFold! 😍

internal/pkg/cli/pipeline_init.go
huanjani: Oh, nvm. Finally read to the bottom. 😆

huanjani (approved):
Nice work! 😻

internal/pkg/cli/pipeline_init.go
```diff
@@ -192,6 +199,57 @@ func (o *initPipelineOpts) Ask() error {
...
+	names, err := o.codePipeline.ListPipelineNamesByTags(map[string]string{
+		deploy.AppTagKey: o.appName,
+	})
+	if err != nil {
+		return fmt.Errorf("validate if pipeline exists: %w", err)
+	}
+
+	fullName := fmt.Sprintf(fmtPipelineName, o.appName, o.name)
+	for _, name := range names {
+		if strings.EqualFold(name, o.name) || strings.EqualFold(name, fullName) {
```
Lou1415926, line 217: It'd be helpful for future readers if we add a comment here explaining that we are validating against both the new namespaced pipeline name (the `fullname`) and the old non-namespaced name.

dannyrandall: Explanation is in the docs for the function :blush:

PR #3403, opened 2022-03-29
chore(pipelines): make hardcoded paths dynamic in buildspec and pipeline CFN template

huanjani:
These changes point the buildspec to the pipeline manifest, whether it's at a legacy location or name-specific location, and also point the pipeline cfn template to the buildspec, whether it's at a legacy or location or a name-specific location. 

An added bonus is that users may indicate, in their pipeline manifests, paths to custom buildspecs.
```
build:
  buildspec: a/path/to/a/different/buildspec.yml
```   

### DIFF
--- a/internal/pkg/cli/pipeline_deploy.go
+++ b/internal/pkg/cli/pipeline_deploy.go
@@ -212,9 +212,9 @@ func (o *deployPipelineOpts) Execute() error {
 	o.shouldPromptUpdateConnection = bool
 
 	// Convert full manifest path to relative path.
-	relPath, err := relPath(o.pipeline.Path)
+	relPath, err := o.ws.Rel(o.pipeline.Path)
 	if err != nil {
-		return fmt.Errorf("convert pipeline manifest path: %w", err)
+		return err
 	}
 
 	// Convert environments to deployment stages.
--- a/internal/pkg/cli/pipeline_deploy.go
+++ b/internal/pkg/cli/pipeline_deploy.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"path/filepath"
 
 	"github.com/aws/aws-sdk-go/service/ssm"
 	"github.com/aws/copilot-cli/internal/pkg/aws/identity"
@@ -194,7 +195,7 @@ func (o *deployPipelineOpts) Execute() error {
 		return err
 	}
 
-	// If the source has an existing connection, get the correlating ConnectionARN .
+	// If the source has an existing connection, get the correlating ConnectionARN.
 	connection, ok := pipeline.Source.Properties["connection_name"]
 	if ok {
 		arn, err := o.codestar.GetConnectionARN((connection).(string))
@@ -210,13 +211,19 @@ func (o *deployPipelineOpts) Execute() error {
 	}
 	o.shouldPromptUpdateConnection = bool
 
-	// convert environments to deployment stages
+	// Convert full manifest path to relative path.
+	relPath, err := relPath(o.pipeline.Path)
+	if err != nil {
+		return fmt.Errorf("convert pipeline manifest path: %w", err)
+	}
+
+	// Convert environments to deployment stages.
 	stages, err := o.convertStages(pipeline.Stages)
 	if err != nil {
 		return fmt.Errorf("convert environments to deployment stage: %w", err)
 	}
 
-	// get cross-regional resources
+	// Get cross-regional resources.
 	artifactBuckets, err := o.getArtifactBuckets()
 	if err != nil {
 		return fmt.Errorf("get cross-regional resources: %w", err)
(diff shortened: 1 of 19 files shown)


internal/pkg/cli/pipeline_init.go
```diff
@@ -562,8 +563,7 @@ func (o *initPipelineOpts) createPipelineManifest() error {
 	for _, env := range o.envConfigs {
 
 		stage := manifest.PipelineStage{
-			Name:             env.Name,
-			RequiresApproval: env.Prod,
```
iamhopaul123, line 624: Do we mention this behavior in our doc now?

internal/pkg/deploy/pipeline.go
```diff
@@ -238,18 +240,23 @@ func PipelineSourceFromManifest(mfSource *manifest.Source) (source interface{},
 }
 
 // PipelineBuildFromManifest processes manifest info about the build project settings.
-func PipelineBuildFromManifest(mfBuild *manifest.Build) (build *Build) {
+func PipelineBuildFromManifest(mfBuild *manifest.Build, plName string) (build *Build) {
 	image := defaultPipelineBuildImage
 	environmentType := defaultPipelineEnvironmentType
+	path := fmt.Sprintf(fmtDefaultPipelineBuildspecPath, plName)
```
iamhopaul123: Does it work if they have a pipeline manifest in the legacy path and want to deploy the pipeline?

internal/pkg/cli/pipeline_init.go
```diff
@@ -589,6 +589,8 @@ func (o *initPipelineOpts) createPipelineManifest() error {
 		return err
 	}
 
+	o.manifestPath = manifestPath
```
efekarakus: nit: can we get rid of the temp variable `manifestPath` and now replace it with `o.manifestPath`?

internal/pkg/cli/pipeline_init.go
huanjani: Great question! Not in pipeline-specific docs, but we do still mention the now-deprecated `--prod` flag in a few places. Opened #3404. 

(Users can still require manual approval; we're just not automatically setting it based on the env's prod flag.)

internal/pkg/deploy/pipeline.go
efekarakus: 👍 to ph's comment, what if we took instead:
```go
func PipelineBuildFromManifest(mfBuild *manifest.Build, mfDirPath string) *Build {
   path := filepath.Join(mfDirPath, "buildspec.yml")
   ...
}
```
and in `pipeline_deploy.go`:
```go
deploy.PipelineBuildFromManifest(pipeline.Build, filepath.Dir(o.pipeline.Path))
```

huanjani: Fixed.

huanjani:
Tested manually by initing two pipelines and moving one to the legacy location. Both CFN templates generated the correct buildspec location.

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -210,13 +211,19 @@ func (o *deployPipelineOpts) Execute() error {
 	}
 	o.shouldPromptUpdateConnection = bool
 
-	// convert environments to deployment stages
+	// Convert full manifest path to relative path.
+	relPath, err := relPath(o.pipeline.Path)
```
efekarakus: do we want the path relative to the current working dir or from the root of the workspace?
I wonder if we should add a `workspace.Rel(o.pipeline.Path)` method that always returns `./copilot/path/to/file` where '.' is the root of the workspace.

Otherwise, I believe this might give the wrong the wrong path in the cfn template if the user runs `pipeline deploy` from a sub-directory

efekarakus (commented):
Looks great!

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -212,9 +212,9 @@ func (o *deployPipelineOpts) Execute() error {
 	o.shouldPromptUpdateConnection = bool
 
 	// Convert full manifest path to relative path.
```
iamhopaul123: nit: we might need to update the comment

internal/pkg/cli/pipeline_init.go
```diff
@@ -574,17 +574,17 @@ func (o *initPipelineOpts) createPipelineManifest() error {
...
 		if !ok {
 			return fmt.Errorf("write pipeline manifest to workspace: %w", err)
 		}
 		manifestExists = true
-		manifestPath = e.FileName
+		o.manifestPath = e.FileName
 	}
 
-	manifestPath, err = relPath(manifestPath)
+	o.manifestPath, err = relPath(o.manifestPath)
```
iamhopaul123: Do we need to change here as well?

huanjani: I took away `manifestPath` altogether.

iamhopaul123: Im still not clear why we don't need `ws.Rel` here. Is it because we require them to run `pipeline_init` within a workspace? But what if they are running the command within a sub dir?

huanjani: Oh, sorry! I thought you meant changing `manifestPath` to `o.manifestPath`! 😆

efekarakus (approved):

internal/pkg/cli/pipeline_init.go
huanjani: Fixed. Thanks!

iamhopaul123 (approved):

PR #3405, opened 2022-03-29
chore(deploy): add and apply ListDeployedPipelines

iamhopaul123:
Add and apply `ListDeployedPipelines`. Also refactored `codepipeline` pkg so that it won't depend on `resourcegroupapi` anymore.

### DIFF
--- a/internal/pkg/cli/app_show.go
+++ b/internal/pkg/cli/app_show.go
@@ -41,7 +41,7 @@ type showAppOpts struct {
 	w                  io.Writer
 	sel                appSelector
 	pipelineInfoGetter pipelineGetter
-	pipelineLister     pipelineLister
+	pipelineLister     deployedPipelineLister
 	newVersionGetter   func(string) (versionGetter, error)
 }
 
--- a/internal/pkg/cli/app_show.go
+++ b/internal/pkg/cli/app_show.go
@@ -37,12 +37,12 @@ type showAppVars struct {
 type showAppOpts struct {
 	showAppVars
 
-	store              store
-	w                  io.Writer
-	sel                appSelector
-	pipelineInfoGetter pipelineGetter
-	pipelineLister     deployedPipelineLister
-	newVersionGetter   func(string) (versionGetter, error)
+	store            store
+	w                io.Writer
+	sel              appSelector
+	codepipeline     pipelineGetter
+	pipelineLister   deployedPipelineLister
+	newVersionGetter func(string) (versionGetter, error)
 }
 
 func newShowAppOpts(vars showAppVars) (*showAppOpts, error) {
@@ -52,12 +52,12 @@ func newShowAppOpts(vars showAppVars) (*showAppOpts, error) {
 	}
 	store := config.NewSSMStore(identity.New(defaultSession), ssm.New(defaultSession), aws.StringValue(defaultSession.Config.Region))
 	return &showAppOpts{
-		showAppVars:        vars,
-		store:              store,
-		w:                  log.OutputWriter,
-		sel:                selector.NewSelect(prompt.New(), store),
-		pipelineInfoGetter: codepipeline.New(defaultSession),
-		pipelineLister:     deploy.NewPipelineStore(vars.name, rg.New(defaultSession)),
+		showAppVars:    vars,
+		store:          store,
+		w:              log.OutputWriter,
+		sel:            selector.NewSelect(prompt.New(), store),
+		codepipeline:   codepipeline.New(defaultSession),
+		pipelineLister: deploy.NewPipelineStore(vars.name, rg.New(defaultSession)),
 		newVersionGetter: func(s string) (versionGetter, error) {
 			d, err := describe.NewAppDescriber(s)
 			if err != nil {
@@ -131,9 +131,9 @@ func (o *showAppOpts) description() (*describe.App, error) {
 	}
 	var pipelineInfo []*codepipeline.Pipeline
 	for _, pipeline := range pipelines {
-		info, err := o.pipelineInfoGetter.GetPipeline(pipeline.ResourceName())
+		info, err := o.codepipeline.GetPipeline(pipeline.ResourceName)
 		if err != nil {
-			return nil, fmt.Errorf("get info for pipeline %s: %w", pipeline.HumanName(), err)
+			return nil, fmt.Errorf("get info for pipeline %s: %w", pipeline.Name(), err)
 		}
 		pipelineInfo = append(pipelineInfo, info)
 	}
--- a/internal/pkg/cli/app_show.go
+++ b/internal/pkg/cli/app_show.go
@@ -10,6 +10,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ssm"
 	"github.com/aws/copilot-cli/internal/pkg/aws/identity"
+	rg "github.com/aws/copilot-cli/internal/pkg/aws/resourcegroups"
 
 	"github.com/aws/copilot-cli/internal/pkg/aws/codepipeline"
 	"github.com/aws/copilot-cli/internal/pkg/aws/sessions"
@@ -36,11 +37,12 @@ type showAppVars struct {
 type showAppOpts struct {
 	showAppVars
 
-	store            store
-	w                io.Writer
-	sel              appSelector
-	pipelineSvc      pipelineGetter
-	newVersionGetter func(string) (versionGetter, error)
+	store              store
+	w                  io.Writer
+	sel                appSelector
+	pipelineInfoGetter pipelineGetter
+	pipelineLister     pipelineLister
+	newVersionGetter   func(string) (versionGetter, error)
 }
 
 func newShowAppOpts(vars showAppVars) (*showAppOpts, error) {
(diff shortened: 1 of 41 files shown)


internal/pkg/deploy/deploy.go
```diff
@@ -21,19 +23,25 @@ const (
...
-	snsResourceType   = "sns"
+	stackResourceType    = "cloudformation:stack"
+	pipelineResourceType = "codepipeline:pipeline"
+	snsResourceType      = "sns"
 
 	// fmtSNSTopicNamePrefix holds the App-Env-Workload- components of a topic name
 	fmtSNSTopicNamePrefix = "%s-%s-%s-"
 	snsServiceName        = "sns"
+
+	// pipeline name has a prefix in the format of "pipeline-${appName}-"
```
Lou1415926: nit: maybe
```suggestion
	// After v1.x, pipeline names are namespaced with a prefix in the format of "pipeline-${appName}-"
```

internal/pkg/deploy/deploy.go
```diff
@@ -87,6 +95,81 @@ func NewStore(sessProvider SessionProvider, store ConfigStoreClient) (*Store, er
...
+}
+
+// NewPipeline initiates a new pipeline.
+func NewPipeline(appName, resourceName string, isLegacy bool) Pipeline {
+	return Pipeline{
+		appName:      appName,
+		resourceName: resourceName,
+		isLegacy:     isLegacy,
+	}
+}
```
Lou1415926: nit: It feels to me that a "deployed" pipeline struct shouldn't be initiate-able from outside of the package, because of its nature of "having been deployed". Just a thought 💭

internal/pkg/cli/interfaces.go
```diff
@@ -438,8 +439,10 @@ type appUpgrader interface {
 
 type pipelineGetter interface {
 	GetPipeline(pipelineName string) (*codepipeline.Pipeline, error)
-	ListPipelineNamesByTags(tags map[string]string) ([]string, error)
-	GetPipelinesByTags(tags map[string]string) ([]*codepipeline.Pipeline, error)
+}
+
+type pipelineLister interface {
```
Lou1415926: tiny nit: what do you think about `deployedPipelineLister`?

internal/pkg/deploy/deploy.go
```diff
@@ -87,6 +95,81 @@ func NewStore(sessProvider SessionProvider, store ConfigStoreClient) (*Store, er
...
+
+// HumanName returns the human-friendly name of the deployed pipeline.
```
Lou1415926: Maybe we can give some example outputs here.

Lou1415926 (commented):
Yayy this is great!! Just a few nits and thoughts.

internal/pkg/deploy/deploy.go
iamhopaul123: It is never used :( it is because our unit tests in the other package need this and i don't want to make these fields public

iamhopaul123: <strike>I'll move this function to the test file</strike> (i can't do this) and add more comments

internal/pkg/cli/app_show.go
```diff
@@ -36,11 +37,12 @@ type showAppVars struct {
...
 
-	store            store
-	w                io.Writer
-	sel              appSelector
-	pipelineSvc      pipelineGetter
-	newVersionGetter func(string) (versionGetter, error)
+	store              store
+	w                  io.Writer
+	sel                appSelector
+	pipelineInfoGetter pipelineGetter
```
efekarakus: nit:
```suggestion
	codepipeline pipelineGetter
```

internal/pkg/cli/pipeline_list.go
```diff
@@ -35,11 +36,12 @@ type listPipelineVars struct {
 
 type listPipelineOpts struct {
 	listPipelineVars
-	pipelineSvc pipelineGetter
-	prompt      prompter
-	sel         configSelector
-	store       store
-	w           io.Writer
+	pipelineInfoGetter pipelineGetter
+	deployedPipelineLister
```
efekarakus: do you want to add a field here instead of embedding the interface?

```suggestion
	codepipeline pipelineGetter
	pipelineLister deployedPipelineLister
```

internal/pkg/deploy/deploy.go
```diff
@@ -87,6 +95,82 @@ func NewStore(sessProvider SessionProvider, store ConfigStoreClient) (*Store, er
 	return s, nil
 }
 
+// Pipeline is a deployed pipeline.
```
efekarakus, line 98: Was there a reason why you preferred adding this functionality to the `deploy` pkg vs. `describe`?

I find both of these packages today confusing with what responsibility they have 🤔

internal/pkg/deploy/deploy.go
```diff
@@ -87,6 +95,82 @@ func NewStore(sessProvider SessionProvider, store ConfigStoreClient) (*Store, er
...
+type Pipeline struct {
+	appName      string
+	resourceName string
+	isLegacy     bool
+}
+
+// NewPipeline initiates a new pipeline. Note that this function should only be used for unit tests.
+func NewPipeline(appName, resourceName string, isLegacy bool) Pipeline {
```
efekarakus: nit: Should we have the `Pipeline` fields public and avoid the constructor since `Pipeline` is just an aggregate data type?

I think a constructor is warranted if the type contains fields with an interface type

internal/pkg/deploy/deploy.go
```diff
@@ -87,6 +95,82 @@ func NewStore(sessProvider SessionProvider, store ConfigStoreClient) (*Store, er
...
+		resourceName: resourceName,
+		isLegacy:     isLegacy,
+	}
+}
+
+// HumanName returns the human-friendly name of the deployed pipeline.
+// For example: bad-goose (legacy)
+func (p Pipeline) HumanName() string {
+	if p.isLegacy {
+		return fmt.Sprintf("%s (legacy)", p.resourceName)
```
efekarakus: What do you think of removing "legacy" and just having `Name`? It might spook users and force them to migrate for no reason

internal/pkg/deploy/deploy.go
```diff
@@ -87,6 +95,82 @@ func NewStore(sessProvider SessionProvider, store ConfigStoreClient) (*Store, er
...
+}
+
+// PipelineStore fetches information on deployed pipelines.
+type PipelineStore struct {
+	appName string
+	getter  resourceGetter
+}
+
+// NewPipelineStore returns a new PipelineStore.
+func NewPipelineStore(appName string, getter *rg.ResourceGroups) *PipelineStore {
```
efekarakus: can we use an interface as input instead of a specific data type type `*rg.ResourceGroups`?

internal/pkg/deploy/deploy.go
iamhopaul123: that sounds good to me! i'll remove it from `deploy` package and have it implemented in the `describe` package in the following PR

internal/pkg/deploy/deploy.go
iamhopaul123: Done!

internal/pkg/deploy/deploy.go
```diff
@@ -21,19 +23,25 @@ const (
...
-	snsResourceType   = "sns"
+	stackResourceType    = "cloudformation:stack"
+	pipelineResourceType = "codepipeline:pipeline"
+	snsResourceType      = "sns"
 
 	// fmtSNSTopicNamePrefix holds the App-Env-Workload- components of a topic name
 	fmtSNSTopicNamePrefix = "%s-%s-%s-"
 	snsServiceName        = "sns"
+
+	// After v1.x, pipeline names are namespaced with a prefix in the format of "pipeline-${appName}-"
```
Lou1415926, line 43: nit:
```suggestion
	// After v1.16, pipeline names are namespaced with a prefix in the format of "pipeline-${appName}-"
```

internal/pkg/deploy/deploy.go
```diff
@@ -87,6 +95,59 @@ func NewStore(sessProvider SessionProvider, store ConfigStoreClient) (*Store, er
...
+}
+
+// PipelineStore fetches information on deployed pipelines.
+type PipelineStore struct {
+	appName string
+	getter  resourceGetter
+}
+
+// NewPipelineStore returns a new PipelineStore.
+func NewPipelineStore(appName string, getter resourceGetter) *PipelineStore {
```
efekarakus, line 120: it is so strange that we can have a public function take as input a private interface, I wonder what's a good usecase for it 🤔

efekarakus (approved):
We discussed a little yesterday with @Lou1415926 and @dannyrandall that the pipeline name should be retrieved from the `copilot-pipeline` tag if it's present, because we should let the resource name be autogenerated by CFN.

I think @Lou1415926 will handle in her PR so I'm approving it!

internal/pkg/deploy/deploy.go
```diff
@@ -87,6 +95,59 @@ func NewStore(sessProvider SessionProvider, store ConfigStoreClient) (*Store, er
...
+		return nil, fmt.Errorf("get pipeline resources by tags for app %s: %w", p.appName, err)
+	}
+	for _, pipelineRes := range pipelineResources {
+		name, err := getPipelineName(pipelineRes.ARN)
+		if err != nil {
+			return nil, err
+		}
+		pipeline := Pipeline{ResourceName: name, AppName: p.appName}
+		if _, ok := pipelineRes.Tags[PipelineTagKey]; !ok {
+			pipeline.IsLegacy = true
```
Lou1415926, line 144: Do you think we should use the value of `pipeineRes.Tags[PipelineTagKey]` as the pipeline name when we can? Like instead of parsing from resource name.

We will probably let CFN generates a name for the `CodePipeline` resource like we do for services, which means the new-world pipeline will have a resource name that contains random letters.

iamhopaul123: sounds good!

internal/pkg/deploy/deploy.go
iamhopaul123: that's exactly the reason why i used to take a concrete struct as input 😂

Lou1415926 (approved):

internal/pkg/deploy/deploy.go
iamhopaul123: But according to

```
func NewStore(sessProvider SessionProvider, store ConfigStoreClient) (*Store, error) {
```
maybe we should make the interface to be public

PR #3414, opened 2022-04-01
chore(pipeline): allow `pipeline init` to proceed with existing name

huanjani:
Instead of erroring out, `pipeline init` will now proceed after logging a warning.
This gives users the flexibility to update their pipeline manifest/buildspec files to the latest defaults by writing the manually deleted files as part of this workflow.

Also in this PR are some changes suggested in the bug bash: using the `--quiet` flag with `wget`, some wording and punctuation changes.

To come: skip prompting for config if reusing existing manifest. Asking for envs, URL, etc. is very misleading if we're not using that info.


internal/pkg/cli/pipeline_init.go
```diff
@@ -200,50 +200,40 @@ func (o *initPipelineOpts) Ask() error {
...
-but has not been deployed. To deploy this pipeline, please run:
-%s
-If you'd like a new default manifest, please manually delete the existing file, then run:
-%s
-`,
-				color.HighlightCode(fmt.Sprintf("copilot pipeline deploy --name %s", o.name)),
-				color.HighlightCode(fmt.Sprintf("copilot pipeline init --name %s", o.name)))
-			return fmt.Errorf("pipeline %s's manifest already exists", color.HighlightUserInput(o.name))
-		}
+		allPipelines = append(allPipelines, pipeline.Name)
```
efekarakus, line 220: Is it possible that there is a duplicate in the `allPipelines` slice (for example a legacy pipeline)  and we print twice?  
Should we ensure that this is a set by using `allPipelines := make(map[string]struct{})`

huanjani: Yes, it's possible, so we `return nil` after the first match.
Let me know if I'm missing something!

efekarakus: I missed that 😅

efekarakus (approved):

iamhopaul123 (approved):

PR #3416, opened 2022-04-01
chore: adapt to namespace/legacy in pipeline operational commands and pipeline delete

Lou1415926:

### DIFF
--- a/internal/pkg/cli/pipeline_delete.go
+++ b/internal/pkg/cli/pipeline_delete.go
@@ -59,7 +59,7 @@ type deletePipelineOpts struct {
 
 	ghAccessTokenSecretName string
 
-	// Interfaces to dependencies
+	// Interfaces to dependencies.
 	pipelineDeployer pipelineDeployer
 	codepipeline     pipelineGetter
 	prog             progress
@@ -68,6 +68,9 @@ type deletePipelineOpts struct {
 	secretsmanager   secretsManager
 	ws               wsPipelineGetter
 	store            store
+
+	// Cached variables.
+	targetPipeline deploy.Pipeline
 }
 
 func newDeletePipelineOpts(vars deletePipelineVars) (*deletePipelineOpts, error) {
@@ -116,16 +119,17 @@ func (o *deletePipelineOpts) Ask() error {
 			return err
 		}
 	}
+
 	if o.name != "" {
 		if _, err := o.codepipeline.GetPipeline(o.name); err != nil {
 			return err
 		}
 	} else {
-		pipelineName, err := askDeployedPipelineName(o.sel, o.appName, fmt.Sprintf(fmtPipelineDeletePrompt, color.HighlightUserInput(o.appName)))
+		pipeline, err := askDeployedPipelineName(o.sel, fmt.Sprintf(fmtPipelineDeletePrompt, color.HighlightUserInput(o.appName)), o.appName)
 		if err != nil {
 			return err
 		}
-		o.name = pipelineName
+		o.name = pipeline.Name
 	}
 
 	if o.skipConfirmation {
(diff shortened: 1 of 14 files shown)


internal/pkg/cli/pipeline_delete.go
```diff
@@ -116,23 +121,27 @@ func (o *deletePipelineOpts) Ask() error {
...
+		o.name = pipeline.Name
+		o.targetPipeline = pipeline
 	}
 
 	if o.skipConfirmation {
 		return nil
 	}
 	deleteConfirmed, err := o.prompt.Confirm(
-		fmt.Sprintf(pipelineDeleteConfirmPrompt, o.name, o.appName),
+		fmt.Sprintf(pipelineDeleteConfirmPrompt, o.targetPipeline.Name, o.appName),
```
iamhopaul123: I see in the other commands we use `o.name`. Ideally they should be equal but is there any reason why we use `o.targetPipeline.Name` here?

internal/pkg/deploy/cloudformation/stack/name.go
```diff
@@ -50,3 +50,10 @@ func NameForAppStack(app string) string {
 func NameForAppStackSet(app string) string {
 	return fmt.Sprintf("%s-infrastructure", app)
 }
+
+func NameForPipeline(app string, pipeline string, isLegacy bool) string {
```
iamhopaul123, line 66: Comment pls. Also i feel like `StackName()` could reuse this function.

iamhopaul123 (approved):
LGTM just two nits!

internal/pkg/cli/pipeline_delete.go
Lou1415926: no particular reason 🤣 I wanted it to be consistent so that we weren't using `o.name` here and `target.Name` there. But then I realized `target` is a cached variable that is not assumed to be always there, so I changed some back to `o.name`. The code is updated.

internal/pkg/deploy/cloudformation/stack/name.go
```diff
@@ -14,6 +14,9 @@ const taskStackPrefix = "task-"
 // TaskStackName holds the name of a Copilot one-off task stack.
 type TaskStackName string
 
+// After v1.16, pipeline names are namespaced with a prefix of "pipeline-${appName}-".
```
efekarakus:
```suggestion
// After v1.16, pipeline stack names are namespaced with a prefix of "pipeline-${appName}-".
```

internal/pkg/deploy/cloudformation/stack/name.go
```diff
@@ -50,3 +53,11 @@ func NameForAppStack(app string) string {
 func NameForAppStackSet(app string) string {
 	return fmt.Sprintf("%s-infrastructure", app)
 }
+
+// NameForPipeline returns the stack name for a pipeline, depending on whether it has been deployed using the legacy scheme.
+func NameForPipeline(app string, pipeline string, isLegacy bool) string {
+	if isLegacy {
+		return pipeline
+	}
+	return fmt.Sprintf(fmtPipelineNamespaced, app, pipeline)
```
efekarakus: do we validate anywhere that the length of this combo is not longer than 128 chars? or even better can we automatically balance things out automatically:
```go
// input: pipeline-{appName that's 100 chars}-{pipelineName that's 100 chars}

func shorten(s string, sampleSize int) string {
   if len(s) < sampleSize {
       return s
   }
   return s[:sampleSize]
}

return fmt.Sprintf(fmtPipelineNamespaced, shorten(app, 59 /* (128 - 10)/2 */), shorten(pipeline, 59))
```

huanjani (approved):
Looks good! Thanks for doing this.

internal/pkg/cli/pipeline_delete.go
```diff
@@ -162,14 +169,39 @@ func (o *deletePipelineOpts) Execute() error {
...
+	}
+	o.targetPipeline = &pipeline
+	return pipeline, nil
+}
+
+func askDeployedPipelineName(sel codePipelineSelector, msg, appName string) (deploy.Pipeline, error) {
 	pipeline, err := sel.DeployedPipeline(msg, "", appName)
 	if err != nil {
-		return "", fmt.Errorf("select deployed pipelines: %w", err)
+		return deploy.Pipeline{}, fmt.Errorf("select deployed pipelines: %w", err)
```
paragbhingre, line 187: 💯

PR #3417, opened 2022-04-01
feat(pipelines): `pipelines ls --local` flag support

dannyrandall:


internal/pkg/cli/pipeline_list.go
```diff
@@ -27,28 +32,45 @@ import (
...
+
 	defaultSession, err := sessions.ImmutableProvider(sessions.UserAgentExtras("pipeline ls")).Default()
 	if err != nil {
 		return nil, fmt.Errorf("default session: %w", err)
 	}
+
+	wsAppName := tryReadingAppName()
+	if vars.appName == "" {
+		vars.appName = wsAppName
+	}
```
iamhopaul123: I feel like we should get this only if they use the local flag.

internal/pkg/cli/pipeline_list.go
```diff
@@ -75,60 +107,157 @@ func (o *listPipelineOpts) Ask() error {
...
+	switch {
+	case o.shouldShowLocalPipelines && o.shouldOutputJSON:
+		return o.jsonOutputLocal(ctx)
+	case o.shouldShowLocalPipelines:
+		return o.humanOutputLocal()
+	case o.shouldOutputJSON:
+		return o.jsonOutputDeployed(ctx)
+	}
+
+	return o.humanOutputDeployed()
```
iamhopaul123, line 129: How about this
```suggestion
         if o.shouldOutputJSON {
                  ctx, cancel := context.WithTimeout(context.Background(), pipelineListTimeout)
                  defer cancel()
                  if o.shouldShowLocalPipelines {
                           return o.jsonOutputLocal(ctx)
                  }
                  return o.jsonOutputDeployed(ctx)
         }
         if o.shouldShowLocalPipelines {
                  return o.humanOutputLocal()
         }
         return o.humanOutputDeployed()
```

internal/pkg/cli/pipeline_list.go
```diff
@@ -75,60 +107,157 @@ func (o *listPipelineOpts) Ask() error {
...
+
+	var mux sync.Mutex
+	var res []*codepipeline.Pipeline
+
+	g, _ := errgroup.WithContext(ctx)
+
+	for i := range pipelines {
+		pipeline := pipelines[i]
+		g.Go(func() error {
+			info, err := getter.GetPipeline(pipeline.ResourceName)
```
iamhopaul123: Do you know how long it usually takes to make one API call like this 🤔

dannyrandall: Just ran a few tests - usually around 150-250 ms per call.

internal/pkg/cli/pipeline_list.go
dannyrandall: I'm don't think there's much overhead in creating a context (plus, ideally, it'd be nice if `humanOutputDeployed()` used one to call `ListDeployedPipelines()`), so I personally would rather have less nesting and keep it as is, but if you think the other option is better to read I'm fine switching.

internal/pkg/cli/pipeline_list.go
```diff
@@ -75,60 +110,157 @@ func (o *listPipelineOpts) Ask() error {
...
-		fmt.Fprintln(b, pipeline)
+// humanOutputDeployed prints the name of all pipelines in the given app that have been deployed.
+func (o *listPipelineOpts) humanOutputDeployed() error {
+	pipelines, err := o.pipelineLister.ListDeployedPipelines(o.appName)
+	if err != nil {
+		return fmt.Errorf("list deployed pipelines: %w", err)
 	}
-	return b.String()
+
+	sort.Slice(pipelines, func(i, j int) bool {
```
huanjani, line 221: nice

huanjani (approved):
LGTM!

internal/pkg/cli/flag.go
```diff
@@ -237,6 +237,7 @@ Defaults to all logs. Only one of end-time / follow may be used.`
 	pipelineResourcesFlagDescription = "Optional. Show the resources in your pipeline."
 	localSvcFlagDescription          = "Only show services in the workspace."
 	localJobFlagDescription          = "Only show jobs in the workspace."
+	localPipelineFlagDescription     = "Only show pipelines in the workspace."
```
huanjani, line 240: This is (correctly) consistent with the other `--local` flags, but just a note to ourselves that we should go through all of the flags to make sure we include "Optional" where should.

iamhopaul123 (approved):

internal/pkg/aws/codepipeline/codepipeline.go
```diff
@@ -38,6 +38,7 @@ type CodePipeline struct {
 
 // Pipeline represents an existing CodePipeline resource.
 type Pipeline struct {
+	// Name is the full name of the pipeline in CodePipeline.
```
Lou1415926: tiny nit:
```suggestion
	// Name is the resource name of the pipeline in CodePipeline, e.g. myapp-mypipeline-RANDOMSTRING.
```

internal/pkg/cli/pipeline_list.go
```diff
@@ -27,43 +32,78 @@ import (
...
+			return describe.NewPipelineDescriber(pipeline, false)
+		},
+		wsAppName: wsAppName,
 	}, nil
 }
 
 // Ask asks for and validates fields that are required but not passed in.
 func (o *listPipelineOpts) Ask() error {
+	if o.shouldShowLocalPipelines {
+		if err := validateInputApp(o.wsAppName, o.appName, o.store); err != nil {
```
Lou1415926: Do you think we can rename this function to something like `validateWorkspaceApp` so that it's clear that this function is used when the command needs to be run within a workspace?

internal/pkg/cli/pipeline_list.go
```diff
@@ -27,43 +32,78 @@ import (
...
 	if err != nil {
 		return nil, fmt.Errorf("default session: %w", err)
 	}
+
+	var wsAppName string
+	if vars.shouldShowLocalPipelines {
+		wsAppName = tryReadingAppName()
+		if vars.appName == "" {
+			vars.appName = wsAppName
+		}
```
Lou1415926: Line 155 has: `cmd.Flags().StringVarP(&vars.appName, appFlag, appFlagShort, tryReadingAppName(), appFlagDescription)`

Which gives `appName` a default value of the workspace app, which gets overridden if `--app` is used. This block might be doing the same thing?

internal/pkg/cli/pipeline_list.go
Lou1415926: Also what do you think of moving this block to `Validate` since all it does is validating that:
1. There is a workspace
2. The workspace doesn't conflict with `--app` flag
3. The app of the workspace exists

without needing to ask anything.

Lou1415926: nit: we could probably just:
```suggestion
		return validateInputApp(o.wsAppName, o.appName, o.store)
```

internal/pkg/cli/pipeline_list.go
```diff
@@ -75,60 +115,162 @@ func (o *listPipelineOpts) Ask() error {
...
+
+	cp := make(map[string]*describe.Pipeline)
+	for _, pipeline := range deployed {
+		cp[pipeline.Name] = pipeline
+	}
+
+	type combinedInfo struct {
+		Name         string `json:"name"`
+		ManifestPath string `json:"manifestPath"`
+		*describe.Pipeline
```
Lou1415926: `describe.Pipeline` also has a field called `Name` which marshals to `name` 🤔  Does this render two `name` fields in the JSON output?

internal/pkg/cli/pipeline_list.go
```diff
@@ -75,60 +115,162 @@ func (o *listPipelineOpts) Ask() error {
...
+			}
+
+			info, err := d.Describe()
+			if err != nil {
+				return fmt.Errorf("describe pipeline %q: %w", pipeline.ResourceName, err)
+			}
+
+			mux.Lock()
+			defer mux.Unlock()
+			res = append(res, info.(*describe.Pipeline))
```
Lou1415926: What do you think of adding this to prevent scary panic:
```gp
info, ok := err.(*descrube.Pipeline)
if !ok {
	continue
}
res = append(res, info)
```

internal/pkg/cli/pipeline_list.go
dannyrandall: no - it will only marshal this `Name` field. If we removed this `Name` field and instantiated a `&describe.Pipeline{}` for the local-only pipelines, they would get a bunch of empty fields in their json output (like `region`, `accountId`, etc)).

internal/pkg/cli/pipeline_list.go
dannyrandall: good call, i'll spit out an error here instead

internal/pkg/cli/pipeline_list.go
Lou1415926: What if a local pipeline is deployed, could we have a `combinedInfo` thats like
```go
combinedInfo {
  Name: "myPipeline",
  manifestPath: "copilot/pipelines/myPipeline/manifest.yml`,
  Pipeline: &describe.Pipeline{
      Name: "myPipeline",
      Pipeline: codepipeline.Pipeline {
           Name: "pipeline-myApp-myPipeline-RANDOMSTRING",
       }
  },
}
```

internal/pkg/cli/pipeline_list.go
dannyrandall: > Also what do you think of moving this block to Validate

I think since we'd have to have a similar block in `Ask()` _anyways_ (like this:
```
Validate() {
  if o.shouldShowLocal {
    return validateInputApp()
  }
}

Ask() {
  if o.shouldShowLocal {
    return nil
  }
  
  // other app name validation
}
```
) it makes sense to consider this block as part of the "app name" validation, and leave it in `Ask()`. Alternatively we could remove the check in ask, but then we'd call `store.GetApplication()` twice, since `validateInputApp()` already calls it.

internal/pkg/cli/pipeline_list.go
dannyrandall: good catch! 😊

internal/pkg/cli/pipeline_list.go
iamhopaul123: I wonder if we display too much info for `ls`? Ideally `ls` users only care about the list, otherwise they could use `show` 🤔

dannyrandall: If a local pipeline is deployed, that's probably exactly how it would look.
Here's what the json would look like (although pretend the `codepipeline.Pipeline` fields are filled in!)

input:
```
cli.combinedInfo{Name:"mypipeline", ManifestPath:"copilot/pipelines/mypipeline/manifest.yml", Pipeline:&describe.Pipeline{Name:"mypipeline", Pipeline:codepipeline.Pipeline{Name:"pipeline-myApp-mypipeline-RANDOMSTRING", Region:"", AccountID:"", Stages:[]*codepipeline.Stage(nil), CreatedAt:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), UpdatedAt:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC)}, Resources:[]*stack.Resource(nil)}}

```
json:
```
{"pipelines":[{"name":"mypipeline","manifestPath":"copilot/pipelines/mypipeline/manifest.yml","pipelineName":"pipeline-myApp-mypipeline-RANDOMSTRING","region":"","accountId":"","stages":null,"createdAt":"0001-01-01T00:00:00Z","updatedAt":"0001-01-01T00:00:00Z"}]}
```

Lou1415926: hmm interesting it seems like only one of `combinedInfo.Name` and `describe.Pipeline.Name` is getting marshaled into `name`. I didn't know this 😮

iamhopaul123: What do folks think of
```json
{
      "name":"mypipeline",
      "manifestPath":"copilot/pipelines/mypipeline/manifest.yml",
      "pipelineName":"pipeline-myApp-myPipeline-RANDOMSTRING",
}
```
Would this be confusing? `pipelineName` could be empty but the blast radius is reduced since the pipeline is not deployed.

dannyrandall: discussed offline. i'll go ahead with @iamhopaul123's suggestion since we can always add in additional fields to `--json --local` in the future.

Lou1415926 (approved):

iamhopaul123 (approved):
🚢

PR #3421, opened 2022-04-04
chore: do not trim pipeline stack name to 128 characters

Lou1415926:
This is reverting a part of the changes in #3416 that trim the pipeline stack name to 128 characters by smart-trimming from app name and pipeline name. CFN will error out if the stack name is too long, preventing the initial creation of the stack in the first place. This errorring out is early enough for user to change their pipeline name in order to meet the restriction. 

In addition, the PR updates `pipeline show` and `pipeline status` output.

### `pipeline show`
For `pipeline show`, the human string output will show the pipeline name as specified in manifest:
```
About

  Name        new-world
  Region      us-east-2
  AccountID   1111111111
  Created At  50 minutes ago
  Updated At  50 minutes ago
```
Its JSON output will show:
```
{"pipelineName":"new-world","name":"pipeline-e2e-try-new-world-Pipeline-11T2LFVLO1XN6","region"...
```
adding a new JSON field `pipelineName` whose value is the manifest pipeline name. This is to prevent breaking changes for folks that've been using the JSON output's `name` filed for automation.

### `pipeline status`
`pipeline status` doesn't show pipeline name as human string output, so no changes here.

It's original JSON output is: 
```
{"pipelineName":"pipeline-e2e-try-new-world-Pipeline-11T2LFVLO1XN6"
```
To prevent it from being breaking for users' automation, this PR adds a `name` field whose value is the manifest pipeline name.

Note that this may be confusing because the fields are named differently for `pipeline show` and `pipeline status` JSON.


huanjani (approved):

internal/pkg/describe/pipeline_show.go
```diff
@@ -48,16 +50,17 @@ func NewPipelineDescriber(pipelineName string, showResources bool) (*PipelineDes
 	pipelineSvc := codepipeline.New(sess)
 
 	return &PipelineDescriber{
+		resourceName:  resourceName,
 		pipelineName:  pipelineName,
 		pipelineSvc:   pipelineSvc,
 		showResources: showResources,
-		cfn:           stack.NewStackDescriber(pipelineName, sess),
+		cfn:           stack.NewStackDescriber(resourceName, sess),
```
efekarakus: does this work? isn't the stack name different than the pipeline resource name or am I misunderstanding something 😝 ?

internal/pkg/describe/pipeline_show.go
```diff
@@ -39,7 +41,7 @@ type PipelineDescriber struct {
 }
 
 // NewPipelineDescriber instantiates a new pipeline describer
-func NewPipelineDescriber(pipelineName string, showResources bool) (*PipelineDescriber, error) {
+func NewPipelineDescriber(resourceName string, pipelineName string, showResources bool) (*PipelineDescriber, error) {
```
efekarakus: It's a bit odd that both `resourceName` and `pipelineName` need to be passed 🤔 

The only other way that I can think of writing this atm is the  following, which I think it's worse so I think I'm okay keeping as is:
```go
describer := NewPipelineDescriber(resource, true)
pipeline, err := describe.Describe()
if err != nil { return err }
if p, ok := pipeline.(*describe.Pipeline); ok {
   p.Name = o.name
   return p
}
return pipeline
```

efekarakus (commented):
Looks good, just a question with the `stack.NewStackDescriber`

internal/pkg/describe/pipeline_show.go
Lou1415926: <strike>yes it's different! This is just a result of renaming the variable `pipelineName` to `resourceName` but the value has been the resource name all along. (Resource name being the CFN generated name for pipeline that has random string appended. It's not the stack name though)</strike>

Edit: fixed!

internal/pkg/describe/pipeline_show.go
Lou1415926: Just to facilitate future discussion here in case we want to revisit ⬇️ 

I also noticed that for workloads in `describe` package, only the Copilot-abstracted workload name is needed and not the actual resource name. The reason behind this I think was because we basically just need `TaskDefinition` information for workloads in `describe`, which have a deterministic name of `app-env-workload`. 

The equivalent of passing only workload name would be to pass only `pipelineName` :( Ideally we would pass only `pipelineName` and infer the `resourceName` but we can't do that since the latter is auto-generated by CFN.

> The only other way that I can think of writing this atm is the following, which I think it's worse so I think I'm okay keeping as is

Discussed offline - we will keep it as is for now because the alternative way requires us to remember setting `p.Name` whenever we want to produce human/JSON output. We should seek for an option that doesn't require us doing this though!

internal/pkg/describe/pipeline_show.go
```diff
@@ -11,7 +11,9 @@ import (
 	"text/tabwriter"
 
 	"github.com/aws/copilot-cli/internal/pkg/aws/sessions"
-	"github.com/aws/copilot-cli/internal/pkg/describe/stack"
+	"github.com/aws/copilot-cli/internal/pkg/deploy"
+	stack "github.com/aws/copilot-cli/internal/pkg/deploy/cloudformation/stack"
+	describeStack "github.com/aws/copilot-cli/internal/pkg/describe/stack"
```
efekarakus: https://rakyll.org/style-packages/#lowercase-only
```
describestack
```

internal/pkg/describe/pipeline_status.go
```diff
@@ -21,36 +21,42 @@ type pipelineStateGetter interface {
...
 
 // PipelineStatus contains the status for a pipeline.
 type PipelineStatus struct {
+	Name string `json:"name"`
 	codepipeline.PipelineState
 }
 
 // NewPipelineStatusDescriber instantiates a new PipelineStatus struct.
-func NewPipelineStatusDescriber(pipelineName string) (*PipelineStatusDescriber, error) {
+func NewPipelineStatusDescriber(resourceName string, pipelineName string) (*PipelineStatusDescriber, error) {
```
efekarakus: can we do the same and pass a `deploy.Pipeline` as input?

efekarakus (approved):

PR #3422, opened 2022-04-04
chore(pipelines): suggest pipeline name to be `[repo]-[branch]`

dannyrandall:
This changes `pipeline init` to suggest `[repo]-[branch]` as a name for the pipeline. Users can press enter to accept the suggested name, or type in a different name if they would like to. This is how the output looks on my machine:

(for branch = `prod`)
Screen Shot 2022-04-04 at 13 25 15

As part of this change, parsing the current branch/repo now occurs before asking for the name, and the output of that shows before asking for the pipeline name, as show in the screenshot above. I think it makes sense in this order, but lmk when you think.

Also refactored the `Execute()` tests to have a single `setupMocks(😇 )` function, and moved tests that dealt with parsing repo/branch to `Ask()`.


internal/pkg/cli/pipeline_init_test.go
```diff
@@ -1022,35 +857,18 @@ func TestInitPipelineOpts_Execute(t *testing.T) {
...
+				secretsmanager: mocks.NewMocksecretsManager(ctrl),
+				parser:         templatemocks.NewMockParser(ctrl),
+				sessProvider:   mocks.NewMocksessionProvider(ctrl),
+				cfnClient:      mocks.NewMockappResourcesGetter(ctrl),
+				store:          mocks.NewMockstore(ctrl),
 			}
-			if tc.mockSessProvider != nil {
-				tc.mockSessProvider(mockSessProvider)
+			if tc.setupMocks != nil {
+				tc.setupMocks(mocks)
```
iamhopaul123, line 786: 😇

iamhopaul123 (approved):

internal/pkg/cli/pipeline_init.go
```diff
@@ -301,12 +303,21 @@ func (o *initPipelineOpts) RequiredActions() []string {
 }
 
 func (o *initPipelineOpts) askPipelineName() error {
+	promptOpts := []prompt.PromptConfig{
+		prompt.WithFinalMessage("Pipeline name:"),
+	}
+
+	// only show suggestion if [repo]-[branch] is a valid pipeline name
```
Lou1415926: super tiny nit: I think we try to keep the comment as a full sentence, hence we should capitalize it and end it with a period.
```suggestion
	// Only show suggestion if [repo]-[branch] is a valid pipeline name.
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -301,12 +303,21 @@ func (o *initPipelineOpts) RequiredActions() []string {
...
+	suggestion := strings.ToLower(fmt.Sprintf("%s-%s", o.repoName, o.repoBranch))
+	if err := validatePipelineName(suggestion, o.appName); err == nil {
```
Lou1415926, line 302: noice!! Making sure we always give a valid suggestion!

Lou1415926 (approved):
LGTM :shipit: ! Just an irrelevant nit. Feel free to remove the label when you are ready!

Issue #2944, opened 2021-10-22
Better error message if there's no environment/workload in pipeline


iamhopaul123:
If there's no env/workload in pipeline, we need to error out earlier rather than simply throwing
```
[Container] 2021/10/22 17:22:10 Running command ls -lah ./infrastructure
--
869 | ls: cannot access ./infrastructure: No such file or directory
```

Issue #2533, opened 2021-06-28
Support for cross-account access with copilot pipeline 


dannysteenman:
I have a scenario in which I manage the test and production environments in separate AWS accounts. We have another account for shared services and that's where we also keep our ci/cd pipelines. 

It would be nice to follow best practices and have the ability to deploy copilot pipeline in a shared-services account and add a trust relationship to test and production so we're able to deploy containers cross-account.

I tried to see if it was possible to create a cross-account pipeline on the shared account but I hit the following error: 

```
arn:aws:iam::12345678912:role/pipeline-copilot-nginx-PipelineRole-Q146DG0LUU2A is not authorized to perform AssumeRole on role arn:aws:iam::12345678900:role/copilot-test-EnvManagerRole (Service: AWSCodePipeline; Status Code: 400; Error Code: InvalidStructureException; Request ID: 9f56a295-7bcd-4e16-a7de-5a4bd7e04ff9; Proxy: null)
```

The copilot app is deployed on test account while I try to deploy the pipeline in the shared account (accountnr is changed for sec reason).

Edit: I was able to get it to work when I manually added the shared account id to the trust entity of the `copilot-test-EnvManagerRole` role on the test account. But now that I think of it... there is another issue... because if you want to be able to deploy crossaccount then it's not handy to manage domains via the copilot app command. Then we actually need to have the ability to configure a domain per environment instead of managing it via the copilot app ```copilot app init --domain example.com``` Because if you use a specific domain e.g. "example.com" that is set up on R53 on production. Then you can't share the domain with a test account because it doesn't have access to that "HostedZoneID".

```
~ via v16.3.0 on test-account-cli (eu-central-1)
✦ ➜ copilot app init copilot --domain example.com
✘ get hosted zone ID for domain example.com: domain does not exist
```

So if we have more granular control e.g. manage domain via environment + setup a trust relationship for copilot pipeline, that would be awesome. For the time being, I'll go with setting up a pipeline per environment account, unless there is a better workaround, I'm all ears 👍

Lou1415926:
Hi @dannysteenman !

Glad it worked out for you, and thank you for diving deep and the suggestions!

[...]

PR #3423, opened 2022-04-04
docs: add multi pipeline and buildspec

huanjani:
Related: #2533, #3231, #2944.


internal/pkg/cli/pipeline_init.go
```diff
@@ -771,6 +771,7 @@ func buildPipelineInitCmd() *cobra.Command {
 		Example: `
   Create a pipeline for the services in your workspace.
   /code $ copilot pipeline init \
+  /code  --name my-pipeline
   /code  --url https://github.com/gitHubUserName/myFrontendApp.git \
```
dannyrandall: what do you think of this to match our suggestions?
```suggestion
  /code  --name my-frontend-app-main \
  /code  --url https://github.com/gitHubUserName/my-frontend-app.git \
```

site/content/docs/concepts/pipelines.en.md
```diff
@@ -35,31 +35,38 @@ Follow the three steps below, from your workspace root:
...
 
-This won't create your pipeline, but it will create some local files that will be used when creating your pipeline.
+This won't create your pipeline, but it will create some local files under `copilot/pipelines` that will be used when creating your pipeline.
 
-* __Release order__: You'll be prompted for environments you want to deploy to - select them based on the order you want them to be deployed in your pipeline (deployments happen one environment at a time). You may, for example, want to deploy to your `test` environment first, and then your `prod` environment.
+* __Pipeline name__: We suggest naming your pipeline `[repository name]-[branch name]` (press 'Enter' when asked, to accept the default name). This will distinguish it from your other pipelines, should you create multiple, and works well if you follow a pipeline-per-branch workflow.
+
+* __Release order__: You'll be prompted for environments you want to deploy to – select them based on the order you want them to be deployed in your pipeline (deployments happen one environment at a time). You may, for example, want to deploy to your `test` environment first, and then your `prod` environment.
+
+!!! Attention
```
huanjani: Maybe we should add a validation for this? Of course, envs can be added/deleted via the manifest after this step, but at least we could prevent some cases of it happening.
Opened an issue: #3424.

site/content/docs/commands/pipeline-init.en.md
```diff
@@ -11,14 +11,16 @@ $ copilot pipeline init [flags]
...
 -h, --help                         help for init
+-n, --name string                  Name of the pipeline.
+-u, --url string                   The repository URL to trigger your pipeline.
 ```
 
 ## Examples
 Create a pipeline for the services in your workspace.
 ```bash
 $ copilot pipeline init \
+--name my-pipeline
```
dannyrandall: same as earlier comment

site/content/docs/concepts/pipelines.en.md
iamhopaul123: > Do not create pipelines that follow the same branch and deploy to the same environment(s), as that will lead to a CloudFormation stack collision.

Why? from my understanding as long as they don't have the same pipeline name within the same app there won't be CFN stack naming collision 🤔

site/content/docs/concepts/pipelines.en.md
```diff
@@ -35,31 +35,38 @@ Follow the three steps below, from your workspace root:
...
 * __Tracking branch__: After you've selected the repository, Copilot will designate your current local branch as the branch your pipeline will follow. This can be changed in Step 2.
 
-### Step 2: Updating the Pipeline manifest (optional)
+### Step 2: Updating the Pipeline Manifest (optional)
 
-Just like your service has a simple manifest file, so does your pipeline. After you run `pipeline init`, two files are created: `pipeline.yml` and `buildspec.yml`, both in your `copilot/` directory. If you poke in, you'll see that the `pipeline.yml` looks something like this (for a service called "api-frontend" with two environments, "test" and "prod"):
+Just like your service has a simple manifest file, so does your pipeline. After you run `pipeline init`, two files are created: `manifest.yml` and `buildspec.yml`, both in a new `copilot/pipelines/[your pipeline name]` directory. If you poke in, you'll see that the `manifest.yml` looks something like this (for a service called "api-frontend" with two environments, "test" and "prod"):
 
 ```yaml
 # The manifest for the "pipeline-ecs-kudos-kohidave-demo-api-frontend" pipeline.
```
dannyrandall:
```suggestion
# The manifest for the "demo-api-frontend-main" pipeline.
```

dannyrandall (commented):
👏

site/content/docs/concepts/pipelines.en.md
huanjani: It's not a name collision, it's a deployment collision with stack update....

internal/pkg/cli/pipeline_init.go
huanjani: will also update examples for `status` and `show`

site/content/docs/concepts/pipelines.en.md
huanjani: Didn't you have this issue when you were trying it out during the bug bash? WX was reporting your experience, then I had a similar one...?

dannyrandall (approved):
lgtm!

internal/pkg/manifest/errors.go
```diff
@@ -26,7 +26,7 @@ type ErrInvalidPipelineManifestVersion struct {
 }
 
 func (e *ErrInvalidPipelineManifestVersion) Error() string {
-	return fmt.Sprintf("pipeline.yml contains invalid schema version: %d", e.invalidVersion)
+	return fmt.Sprintf("manifest.yml contains invalid schema version: %d", e.invalidVersion)
```
Lou1415926: Will we throw the same error message for users with the old file structure where their pipeline manifest is named `pipeline.yml` instead of `manifest.yml`?

internal/pkg/manifest/pipeline.go
```diff
@@ -239,7 +239,7 @@ func UnmarshalPipeline(in []byte) (*Pipeline, error) {
 		return &pm, nil
 	}
 	// we should never reach here, this is just to make the compiler happy
-	return nil, errors.New("unexpected error occurs while unmarshalling pipeline.yml")
+	return nil, errors.New("unexpected error occurs while unmarshalling manifest.yml")
```
Lou1415926, line 241: Same question here!

site/content/docs/concepts/pipelines.en.md
```diff
@@ -12,19 +12,19 @@ We won't get too philosophical about releasing software, but what's the point of
...
 Using an automated release tool like CodePipeline helps make your release manageable. Even if your release isn't particularly complicated, knowing that you can just `git push` to deploy your change always feels a little magical 🌈.
 
-## Pipeline structure
+## Pipeline Structure
 
 Copilot can set up a CodePipeline for you with a few commands - but before we jump into that, let's talk a little bit about the structure of the pipeline we'll be generating. Our pipeline will have the following basic structure:
 
 1. __Source Stage__ - when you push to a configured GitHub, Bitbucket, or CodeCommit repository branch, a new pipeline execution is triggered.
-2. __Build Stage__ - after your source code is pulled from your repository host, your service's container image is built and published to every environment's ECR repository.
+2. __Build Stage__ - after your source code is pulled from your repository host, your service's container image is built and published to every environment's ECR repository and any environment files are uploaded to S3.
```
Lou1415926: Does `"environment files"` here refers to addon templates and the [environment variable files](https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#env_file)? What do you think about saying something like `"...and any artifacts, such as your addon templates and your environment variable files..."`

site/content/docs/concepts/pipelines.en.md
iamhopaul123: Oh yeah I see what you mean.

Lou1415926: From my understanding, if two CodePipeline are trying to update the same stack, say `app-test-svc`, this is what will happen:
1. Pipeline 1 gets to update `app-test-svc`
2. Pipeline 2 tries to update `app-test-svc` and then receives an error that says the stack is being updated
3. Pipeline 2 keeps retrying and retrying and retr...
4. Pipeline finishes the update
5. Pipeline 2, having kept retrying, finally has its chance

So theoretically it shouldn't be an issue is my understanding 🤔 I feel like what happened to PH's was a very weird edge case because some manual operations were involved. I don't know about what you saw though! Maybe we can talk offline about your experience.

internal/pkg/manifest/errors.go
huanjani: I'll differentiate!

site/content/docs/concepts/pipelines.en.md
huanjani: I don't think Pipeline 2 will keep retrying; I believe it will fail and then the user has to manually 'Retry' via the console. We could make this a warning that that will happen rather than telling people not to do it if we think there is utility in being able to do this, even with the hiccup.

Lou1415926: I also tried the same thing and my observation was that an "UPDATE" by Pipeline 2 was triggered on `app-test-svc` immediately after the first update was finished, even though both pipelines are triggered at the same time 🤔

huanjani: Discussed offline. Removing warning for now.

internal/pkg/manifest/errors.go
```diff
@@ -19,14 +19,15 @@ func (e *ErrInvalidWorkloadType) Error() string {
...
+// ErrInvalidPipelineManifestVersion occurs when the pipeline.yml/manifest.yml file
 // contains invalid schema version during unmarshalling.
 type ErrInvalidPipelineManifestVersion struct {
 	invalidVersion PipelineSchemaMajorVersion
+	fileName       string
 }
 
 func (e *ErrInvalidPipelineManifestVersion) Error() string {
-	return fmt.Sprintf("manifest.yml contains invalid schema version: %d", e.invalidVersion)
+	return fmt.Sprintf("%s contains invalid schema version: %d", e.fileName, e.invalidVersion)
```
Lou1415926: Alternatively, what do you think of not including path name in errors thrown by `manifest` package?

```go
func (e *ErrInvalidPipelineManifestVersion) Error() string {
	return fmt.Sprintf("pipeline manifest contains invalid schema version: %d", e.invalidVersion)
}
```

and then in `workspace` package:
```go
pipelineManifest, err := manifest.UnmarshalPipeline(data)
	if err != nil {
		return nil, fmt.Errorf("unmarshal pipeline manifest %s: %w", path, err)
	}
```

I feel like `manifest` package should be agnostic of `path` information, while it makes sense for `workspace` to know `path`.

Let me know if this feels right to you!

huanjani: Sure. We can separate concerns. I guess it's already clear enough which file "pipeline manifest" refers to.

Lou1415926 (approved):

PR #3428, opened 2022-04-05
chore(cli): update log msgs for duplicate pipeline name, reorg file

huanjani:
Moved private methods below public ones.
Check local pipelines before deployed pipelines for duplicates.
Clarified log messages.


dannyrandall (approved):
🥰

internal/pkg/cli/pipeline_init.go
```diff
@@ -645,7 +646,9 @@ func (o *initPipelineOpts) createPipelineManifest() error {
 
 	manifestMsgFmt := "Wrote the pipeline manifest for %s at '%s'\n"
 	if manifestExists {
-		manifestMsgFmt = "Pipeline manifest file for %s already exists at %s, skipping writing it.\n"
+		manifestMsgFmt = `Pipeline manifest file for %s already exists at %s, skipping writing it.
```
efekarakus, line 649: If they already exist, can we make these a `log.Debug` instead of `log.Success` (samesies for buildspec)

huanjani: The next line is a `log.Debug` already. I feel like this information will get lost with the mass of grey. How do you feel about `log.Info`?

efekarakus: Info sounds good! I just want to increase our Signal/Noise ratio, maybe we can do `Info` and remove the second sentence from the logs ("Previously set repository URL, branch, and environment stages will remain." , "Previously set config will remain.")

huanjani: Oh, I added those because I want it to be clearer that even though we're (still) asking for envs, if they chose different ones than those in the existing manifest, the change won't be applied.

efekarakus: Oh I wanted to keep `Successf` when we write the file but `Infof` if it already exists :rip:

huanjani: 🤦🏼  done

internal/pkg/cli/pipeline_init.go
```diff
@@ -683,15 +689,18 @@ func (o *initPipelineOpts) createBuildspec() error {
...
-	}
 	buildspecPath, err = relPath(buildspecPath)
 	if err != nil {
 		return err
 	}
-	log.Successf(buildspecMsgFmt, color.HighlightResource(buildspecPath))
+	buildspecMsgFmt := "Wrote the buildspec for the pipeline's build stage at '%s'\n"
+	if buildspecExists {
+		buildspecMsgFmt = `Buildspec file for pipeline already exists at %s, skipping writing it.
+Previously set config will remain.`
```
efekarakus: I think we might need a new line at the end of this string
```suggestion
Previously set config will remain.
`
```

efekarakus (approved):

internal/pkg/cli/pipeline_init.go
huanjani: Good catch. Thanks!

PR #3433, opened 2022-04-06
feat(pipelines): add CSC connection selector and flags

huanjani:
Given our new multi-pipelines feature, more users will want to reuse their existing CodeStar Connections connections more frequently-- they will not want to generate a new resource for each pipeline, and they will not want to go through the added wait time and actions required for Copilot to generate new connections. Previously, we already allowed customers to pass in their existing connections in their[ pipeline manifests](https://aws.github.io/copilot-cli/docs/manifest/pipeline/#source-properties-connection-name) with
```
source:
  provider: GitHub
  properties:
    branch: main
    repository: https://github.com/<user>/frontend
    # Optional: specify the name of an existing CodeStar Connections connection.
    connection_name: a-connection
```
This PR's changes surface this option into the `pipeline init` workflow to give it more visibility and ease of use.

If a connection is indicated, whether by flag or selector, its name will be injected into the generated pipeline manifest:
```
source:
  # The name of the provider that is used to store the source artifacts.
  # (i.e. GitHub, Bitbucket, CodeCommit)
  provider: GitHub
  # Additional properties that further specify the location of the artifacts.
  properties:
    branch: main
    repository: https://github.com/huanjani/aws-copilot-sample-service
    connection_name: GHConnection
```
If no flag is passed in, no connections exist, or if users select "[No thanks]", the generated pipeline manifest will appear as before:
```
source:
  # The name of the provider that is used to store the source artifacts.
  # (i.e. GitHub, Bitbucket, CodeCommit)
  provider: Bitbucket
  # Additional properties that further specify the location of the artifacts.
  properties:
    branch: main
    repository: https://bitbucket.org/huanjani/newrepototesturl
    # Optional: specify the name of an existing CodeStar Connections connection.
    # connection_name: a-connection
```
Users may still add their connection at this stage; if no connection is indicated, Copilot will generate one upon deploy.

Note: the `deploy` package treats `connection_name` as a property of Source, but the `manifest` package handles it separately from the other properties in order to conditionally display the helpful comments and example (without connection) or the chosen connection (with connection).

Docs to come in a separate PR.


internal/pkg/aws/codestar/codestar.go
```diff
@@ -51,29 +51,50 @@ func (c *CodeStar) WaitUntilConnectionStatusAvailable(ctx context.Context, conne
 	}
 }
 
+// ListConnections retrieves all the CSC connections in the current account.
```
efekarakus, line 54: nit:
```suggestion
// ListConnections retrieves all the CSC connection names in the account.
```

internal/pkg/cli/flag.go
```diff
@@ -230,6 +231,7 @@ Defaults to all logs. Only one of end-time / follow may be used.`
 	githubURLFlagDescription         = "(Deprecated.) Use '--url' instead. Repository URL to trigger your pipeline."
 	githubAccessTokenFlagDescription = "GitHub personal access token for your repository."
 	gitBranchFlagDescription         = "Branch used to trigger your pipeline."
+	connectionFlagDescription        = "CodeStar Connections connection to your third-party repository."
```
efekarakus, line 234: I assume the input is a name right instead of an ARN? should we update the text?
```suggestion
	connectionFlagDescription        = "Optional. CodeStar connection name to connect to your repository."
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -195,6 +203,12 @@ func (o *initPipelineOpts) Ask() error {
 		return err
 	}
 
+	if o.githubAccessToken == "" && !strings.Contains(o.repoURL, ccIdentifier) {
```
efekarakus, line 206: I wonder if there is a way for us to push down this logic 🤔 
For example, what if we put a method that can be implemented by all manifest source providers:
```go
package manifest

func (p *githubV1Provider) RequiresConnection() bool
func (p *githubProvider) RequiresConnection() bool
func (p *codecommitProvider) RequiresConnection() bool
func (p *bitbucketProvider) RequiresConnection() bool
```
and then, over here in `pipeline_init.go`:
```go
provider, err := pipelineProvider()
if err != nil {
    return err
}
if provider.RequiresConnection() {
   return o.askConnection()
}
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -154,11 +161,12 @@ func newInitPipelineOpts(vars initPipelineVars) (*initPipelineOpts, error) {
 		cfnClient:        cloudformation.New(defaultSession),
 		store:            ssmStore,
 		prompt:           prompter,
-		sel:              selector.NewSelect(prompter, ssmStore),
 		runner:           exec.NewCmd(),
 		fs:               &afero.Afero{Fs: afero.NewOsFs()},
 		wsAppName:        wsAppName,
 		pipelineLister:   deploy.NewPipelineStore(rg.New(defaultSession)),
+		codestar:         cs,
```
efekarakus, line 168: do we need to store the field? it seems to be used only by the selector if I'm not missing something

internal/pkg/aws/codestar/codestar.go
```diff
@@ -51,29 +51,50 @@ func (c *CodeStar) WaitUntilConnectionStatusAvailable(ctx context.Context, conne
...
 // connection name passed in.
 func (c *CodeStar) GetConnectionARN(connectionName string) (connectionARN string, err error) {
+	connections, err := c.listConnections()
+	if err != nil {
+		return "", err
+	}
+
+	for _, connection := range connections {
+		if aws.StringValue(connection.ConnectionName) == connectionName {
+			// Duplicate connection names are supposed to result in replacement, so okay to return first match.
```
Lou1415926, line 77: It seems possible to have two connections with the same name but different ARN 🤔

internal/pkg/manifest/pipeline.go
```diff
@@ -173,6 +193,7 @@ type Pipeline struct {
 type Source struct {
 	ProviderName string                 `yaml:"provider"`
 	Properties   map[string]interface{} `yaml:"properties"`
+	Connection   string                 `yaml:"connection_name"`
```
efekarakus, line 196: Shouldn't `connection_name` be in Properties? https://aws.github.io/copilot-cli/docs/manifest/pipeline/ 
This seems to add the connection in two locations in the manifest:
```yaml
source:
  provider: GitHub
  properties:
     connection_name:
  connection_name:
```

internal/pkg/manifest/pipeline.go
```diff
@@ -124,7 +142,7 @@ type CodeCommitProperties struct {
 
 // NewProvider creates a source provider based on the type of
 // the provided provider-specific configurations
-func NewProvider(configs interface{}) (Provider, error) {
+func NewProvider(configs interface{}, connection string) (Provider, error) {
```
efekarakus, line 145: was there a reason why you preferred to take the connection as an extra parameter in the constructor instead of adding it to `GitHubProperties` and `BitbucketProperties`?

internal/pkg/cli/pipeline_init.go
Lou1415926, line 168: tiny nit:
```suggestion
		codestar:         cs.New(defaultSession),
```

The real question is where do we use the member `codestar`? Does it need to be a member?

internal/pkg/cli/pipeline_init.go
```diff
@@ -816,6 +846,7 @@ func buildPipelineInitCmd() *cobra.Command {
 	_ = cmd.Flags().MarkHidden(githubAccessTokenFlag)
 	cmd.Flags().StringVarP(&vars.repoBranch, gitBranchFlag, gitBranchFlagShort, "", gitBranchFlagDescription)
 	cmd.Flags().StringSliceVarP(&vars.environments, envsFlag, envsFlagShort, []string{}, pipelineEnvsFlagDescription)
+	cmd.Flags().StringVar(&vars.connection, connectionFlag, "", connectionFlagDescription)
```
Lou1415926, line 849: Do we also need a flag that people can use to skip the connection prompt? Or a related question - do we consider not prompting for a connection and only letting user use their own connection through flag?

internal/pkg/manifest/pipeline.go
```diff
@@ -241,16 +263,12 @@ func UnmarshalPipeline(in []byte) (*Pipeline, error) {
...
+// NeedsCodeStarConnection indicates to the manifest if this source requires a CSC connection.
+func (s Source) NeedsCodeStarConnection() bool {
+	if (s.ProviderName == GithubProviderName || s.ProviderName == BitbucketProviderName) && s.Connection == "" {
 		return true
-	case BitbucketProviderName:
-		return true
-	default:
-		return false
 	}
+	return false
```
Lou1415926, line 271: nit:
```suggestion
	return (s.ProviderName == GithubProviderName || s.ProviderName == BitbucketProviderName) && s.Connection == ""
```

internal/pkg/term/selector/selector.go
```diff
@@ -891,6 +913,23 @@ func (s *Select) Environments(prompt, help, app string, finalMsgFunc func(int) p
...
+// Connection fetches all the CodeStar in an account and prompts the user to select one.
+func (s *PipelineEnvsConnectionSelect) Connection(msg, help string) (string, error) {
+	connections, err := s.connectionLister.ListConnections()
+	if err != nil {
+		return "", fmt.Errorf("list connections: %w", err)
+	}
+	if len(connections) == 0 {
+		return "", nil
+	}
+	connections = append(connections, "[No thanks]")
```
Lou1415926, line 925: Perhaps we can make it a constant:
```go
selectConnectionNoneOption = "[No thanks]"
```

internal/pkg/manifest/pipeline.go
```diff
@@ -241,16 +263,12 @@ func UnmarshalPipeline(in []byte) (*Pipeline, error) {
 	return nil, errors.New("unexpected error occurs while unmarshalling manifest.yml")
 }
 
-// IsCodeStarConnection indicates to the manifest if this source requires a CSC connection.
-func (s Source) IsCodeStarConnection() bool {
-	switch s.ProviderName {
-	case GithubProviderName:
+// NeedsCodeStarConnection indicates to the manifest if this source requires a CSC connection.
+func (s Source) NeedsCodeStarConnection() bool {
+	if (s.ProviderName == GithubProviderName || s.ProviderName == BitbucketProviderName) && s.Connection == "" {
```
efekarakus, line 268: dang this is tricky, I'd love not to have to repeat logic - we're kind of duplicating when to ask for codestar in both `pipeline_init` and here 🤔

internal/pkg/term/selector/selector.go
```diff
@@ -891,6 +913,23 @@ func (s *Select) Environments(prompt, help, app string, finalMsgFunc func(int) p
...
+	connection, err := s.prompt.SelectOne(msg, help, connections, prompt.WithFinalMessage(connectionFinalMsg))
+	if err != nil {
+		return "", fmt.Errorf("select CodeStar connection: %w", err)
+	}
+	return connection, nil
```
efekarakus, line 930: instead of having the client handle the "[No thanks]" response, can we check for it here and return `"", nil` then?

internal/pkg/term/selector/selector.go
```diff
@@ -199,6 +205,13 @@ type AppPipelineSelect struct {
 	*CodePipelineSelect
 }
 
+// PipelineEnvsConnectionSelect is a selector for pipeline envs and CSC.
+type PipelineEnvsConnectionSelect struct {
```
efekarakus, line 209:
```suggestion
type ConnectionSelect struct {
```

efekarakus (commented):
Thank you for adding this! It'll make folks a lot happier I think

internal/pkg/cli/pipeline_init.go
```diff
@@ -352,6 +366,22 @@ func (o *initPipelineOpts) askEnvs() error {
 	return nil
 }
 
+func (o *initPipelineOpts) askConnection() error {
+	if o.connection != "" {
+		return nil
+	}
+	connection, err := o.sel.Connection(pipelineSelectConnectionPrompt, pipelineSelectConnectionHelpPrompt)
+	if err != nil {
+		return err
```
Lou1415926, line 375: Do we need to wrap the error?

internal/pkg/term/selector/selector.go
huanjani: Previously, the `PipelineEnvSelector`, used only in `pipeline_init`, was the selector. Now that it's also implementing `Connection`, I thought I would combine the two with a name that reflects that, like some of the other selectors. Is it better to separate them?

Issue #3402, opened 2022-03-28
[Design] Ordering deployments in a pipeline


efekarakus:
We propose adding support for  a `deployments` field under `stages` to the [pipeline manifest](https://aws.github.io/copilot-cli/docs/manifest/pipeline/) in order to control which workloads should be deployed and their deployment order in a stage.

```yaml
stage:
- name: string
  requires_approval: boolean      
  test_commands: [string, ...]              
  deployments:                    # Optional. New.
    <name>:                       # Name for the codepipeline Deploy action.
       stack_name: string         # Optional. Defaults to {app name}-{stage name}-{name}
       template_path: string      # Optional. Path to the generated template during Build. Defaults to infrastructure/{name}-{stage name}.yml
       template_config: string    # Optional. Path to the generated config file during Build. Defaults to infrastructure/{name}-{stage name}.params.json
       depends_on: [string, ...]  # Optional. Actions that need to happen before this one.
```

### An example
For example, let's say we have an application with two environments, "test" and "prod", and services organized in a monorepo:
```
copilot/
├── worker
│   └── manifest.yml
└── www
    └── manifest.yml
```

Now we can run `copilot pipeline init` and modify the pipeline manifest like this:
```yaml
name: pipeline-sample-app

stages:
-  name: test
   deployments:
     worker:
- name: prod
  deployments:
     worker:
```
The above manifest will allow us to restrict the pipeline to deploy only the "worker" service instead of both "worker" and "www" in parallel.

I could also modify my manifest file like this:
```yaml

stages:
-  name: test
   deployments:
     worker:
     www:
       depends_on:
       - worker
- name: prod
  deployments:
     worker:
     www:
       depends_on:
       - worker
```
To create an order between deployments, first the "worker" service will be deployed and then the "www" service.

Related #2935

efekarakus:
Code specification for the feature: https://gist.github.com/efekarakus/d02c31713b4f0d9adab24ba6c067a946

efekarakus:
This feature is now released with v1.18! https://github.com/aws/copilot-cli/releases/tag/v1.18.0

PR #3442, opened 2022-04-07
refactor: mv pipeline actions from template to go type

efekarakus:
Related #3402


internal/pkg/deploy/pipeline.go
```diff
@@ -442,41 +445,143 @@ func (s *GitHubV1Source) Owner() (string, error) {
...
+type associatedEnvironment struct {
+	// Name of the environment, must be unique within an application.
+	// This is also the name of the pipeline stage.
+	Name string
+
+	// The region this environment is created in.
+	Region string
+
+	// AppName represents the application name the environment is part of.
+	AppName string
```
Lou1415926, line 457: Is `AppName` always the app where the pipeline lives in?

internal/pkg/deploy/pipeline.go
```diff
@@ -442,41 +445,143 @@ func (s *GitHubV1Source) Owner() (string, error) {
...
+}
+
+// Name returns the name of the CodePipeline deploy action for a workload.
+func (a *WorkloadDeployAction) Name() string {
+	return fmt.Sprintf("CreateOrUpdate-%s-%s", a.name, a.envName)
+}
+
+// StackName returns the name of the workload stack to create or update.
+func (a *WorkloadDeployAction) StackName() string {
+	return fmt.Sprintf("%s-%s-%s", a.appName, a.envName, a.name)
```
Lou1415926, line 613: What do you think of returning `stack.NameForService(a.appName, a.envName, a.name)`?

internal/pkg/deploy/pipeline.go
```diff
@@ -442,41 +445,143 @@ func (s *GitHubV1Source) Owner() (string, error) {
...
+	return actions
+}
+
+// WorkloadDeployAction represents a CodePipeline action of category "Deploy" for a workload stack.
+type WorkloadDeployAction struct {
+	name    string // workload name.
+	envName string
+	appName string
+
+	requiresApproval bool
```
Lou1415926: Is `requiresApproval`a part of `WorkloadDeployAction` because we need it to increment the action's run order? If so, do you think this may be an indication that this logic 
```go
if requiresApproval {
   action.runOrder += 1
}
```
should happen in the stage struct?

internal/pkg/deploy/pipeline.go
```diff
@@ -442,41 +445,143 @@ func (s *GitHubV1Source) Owner() (string, error) {
...
 // of a workspace. A stage consists of the Config Environment the pipeline
 // is deploying to, the containerized services that will be deployed, and
 // test commands, if the user has opted to add any.
 type PipelineStage struct {
-	*AssociatedEnvironment
-	LocalWorkloads   []string
-	RequiresApproval bool
-	TestCommands     []string
+	*associatedEnvironment
+	localWorkloads   []string
```
Lou1415926, line 469: stupid question ✋🏼  why do we prefer making these fields private and use getter to get the value here?

internal/pkg/deploy/pipeline.go
efekarakus: Yeah! Did you have a concern or a refactor suggestion for it?

internal/pkg/deploy/pipeline.go
efekarakus: lol it's not stupid! It's the idea behind a data abstraction, we can change how things are represented internally but not change the interface. For example, I originally had:
```go
func (stg *PipelineStage) EnvManagerRoleARN() string {
	return fmt.Sprintf("arn:aws:iam::%s:role/%s-%s-EnvManagerRole", stg.accountID, stg.name, stg.envName)
}
```
but then I swapped it to return the `config.Env`'s env manager role ARN.

TBH it doesn't matter much for `PipelineStage`.  
However, I think this approach is nice for `WorkloadDeployAction`. Eventually, we'll have another type `EnvDeployAction` where I want to create a common interface between the two types:
```go
type DeployAction interface { // implemented by both WorkloadDeployAction and EnvDeployAction.
   Name() string
   StackName() string
   RunOrder() int
   TemplatePath() string
   TemplateConfigPath() string
}
```

internal/pkg/deploy/pipeline.go
efekarakus: Will fix good idea!

internal/pkg/deploy/pipeline.go
efekarakus: lol I tried this too but there is a cyclic dependency :rip:

internal/pkg/deploy/pipeline.go
Lou1415926: My concern was that it might be a piece of redundant information (there is `.AppName` as well at the top-level `CreatePipelineInput`). This means it may be possible that at some point in the future, by some accidents maybe, this `AppName` holds a different value  than `CreatePipelineInput.AppName` which will cause a bug.

I see why it's needed here, and how it plays a role in the `DeployAction` interface design. I don't have a good refactor suggestion though : (

internal/pkg/deploy/pipeline.go
efekarakus: Check out the latest version with the embedded `action` type, Hope you like it!

internal/pkg/deploy/pipeline.go
```diff
@@ -442,41 +445,197 @@ func (s *GitHubV1Source) Owner() (string, error) {
...
+		return nil
+	}
+	return &ManualApprovalAction{
+		name: stg.associatedEnvironment.Name,
+	}
+}
 
-	// The region this environment is stored in.
-	Region string
+func (stg *PipelineStage) Region() string {
```
iamhopaul123, line 510: Comment pls 🙏

internal/pkg/deploy/pipeline.go
```diff
@@ -442,41 +445,197 @@ func (s *GitHubV1Source) Owner() (string, error) {
...
+	return actions
+}
+
+type orderedRunner interface {
+	RunOrder() int
+}
+
+// action represents a generic CodePipeline action.
+type action struct {
+	prevActions []orderedRunner
```
iamhopaul123: Why do we need this 🤔 Can we record the max run order of the prev actions?

internal/pkg/deploy/pipeline.go
efekarakus: :homer-disappearing: fixed!

internal/pkg/deploy/pipeline.go
efekarakus: We could definitely do that! but then that means that the client of `PipelineStage` needs to call always the methods in a specific order:
1. `Approval`
2. `Deployments`
3. `Test`

Ideally, users should be able to call `stg.Test` or `stg.Deployments` and always get the same run order. This way of implementing is just one way! We can do some caching, but my thought was that we don't expect a huge slice anyway so it's most likely fine to just do a quick for-loop

iamhopaul123 (approved):

internal/pkg/deploy/pipeline.go
Lou1415926: the nitest nit: I'm not sure if this is necessary but maybe we can document this as `// The last actions to be executed immediately before this action` so that in the future we don't misunderstand this as "aggregated previous version" 🤔 ?

Lou1415926 (approved):
LGTM! Putting on a DNM label just in case you'd like to address the nit but feel free to just remove it

PR #3451, opened 2022-04-11
feat: load balanced web service observability

dannyrandall:


internal/pkg/template/templates/workloads/partials/cf/taskrole.yml
```diff
@@ -2,9 +2,18 @@ TaskRole:
   Metadata:
     'aws:copilot:description': 'An IAM role to control permissions for the containers in your tasks'
   Type: AWS::IAM::Role
-  Properties:{{if .NestedStack}}{{$stackName := .NestedStack.StackName}}{{if gt (len .NestedStack.PolicyOutputs) 0}}
-    ManagedPolicyArns:{{range $managedPolicy := .NestedStack.PolicyOutputs}}
-    - Fn::GetAtt: [{{$stackName}}, Outputs.{{$managedPolicy}}]{{end}}{{end}}{{end}}
+  Properties:
+    {{- if or (.NestedStack) (eq .Observability.Tracing "AWSXRAY")}}
```
efekarakus: We should create a `ManagedPolicyArns` field only if the `NestedStack` has `PolicyOutputs` right now it's possible that we render a template like this:
```
Properties:
  ManagedPolicyArns:
  AssumeRolePolicyDocument:
```
if there are addons and no managed policy and tracing afaict

efekarakus (commented):
Looks great, thanks Danny!

internal/pkg/template/templates/workloads/partials/cf/taskrole.yml
dannyrandall: good catch! fixed 😊

internal/pkg/template/templates/workloads/partials/cf/taskrole.yml
```diff
@@ -3,13 +3,14 @@ TaskRole:
     'aws:copilot:description': 'An IAM role to control permissions for the containers in your tasks'
   Type: AWS::IAM::Role
   Properties:
-    {{- if or (.NestedStack) (eq .Observability.Tracing "AWSXRAY")}}
+    {{- $hasAddonPolicies := and .NestedStack (gt (len .NestedStack.PolicyOutputs) 0)}}
```
efekarakus: will this result in a segfault if there are no addons since we're building w go1.17? 😢 I think w go1.18 it would shortcircuit and we'd be safe

dannyrandall: i'm crying rn😢

dannyrandall: just tested it, you're right. i'll fix :)

efekarakus (dismissed):
LGTM! 🚢 

Just added the label to validate with @Lou1415926 if that image is consistent with app runner, after that feel free to remove it

internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
```diff
@@ -26,6 +26,16 @@
...
 {{- end}}
+{{- if eq .Observability.Tracing "AWSXRAY"}}
+- Name: aws-otel-collector
+  Image: public.ecr.aws/aws-observability/aws-otel-collector:latest
+  LogConfiguration:
+    LogDriver: awslogs
+    Options:
+      awslogs-region: !Ref AWS::Region
+      awslogs-group: !Ref LogGroup
+      awslogs-stream-prefix: copilot
```
efekarakus, line 37: Thanks for adding this! do you have a sample log stream by any chance (how easy is it for a client to tell that the log stream is for the otel sidecar)?

internal/pkg/template/templates/workloads/partials/cf/taskrole.yml
```diff
@@ -102,5 +102,23 @@ TaskRole:
...
+                - 'logs:CreateLogStream'
+                - 'logs:DescribeLogStreams'
+                - 'logs:DescribeLogGroups'
+                - 'xray:PutTraceSegments'
+                - 'xray:PutTelemetryRecords'
+                - 'xray:GetSamplingRules'
+                - 'xray:GetSamplingTargets'
+                - 'xray:GetSamplingStatisticSummaries'
+                - 'ssm:GetParameters'
+              Resource: "*"
```
efekarakus: Out of curiosity, how did you know which actions to permit here?
I understand that for broader otel we might need additional permissions:
https://aws-otel.github.io/docs/setup/ecs/create-iam-policy but I was curious if we needed `ssm:GetParameters` if all this integration can be used for is "xray"?

internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
dannyrandall: It'll look like this!
```
app: copilot-playground
env: beta
svc: movies-lbws
```
Screen Shot 2022-04-15 at 09 11 11

internal/pkg/template/templates/workloads/partials/cf/taskrole.yml
dannyrandall: I got the policy from [here](https://aws-otel.github.io/docs/setup/permissions).

As far as I can tell, the only use for `ssm:GetParameters` is to get a custom configuration file, so you're right - we don't need it for this PR, where all we support is the default config. If we want to support custom config in the future, we'll have to add it then.

I think if we added support for custom configuration, and people wanted to start using that feature, they'd have to redeploy their service _anyways_, so changing the policy in the future should be ok - given that, do you think it makes sense to remove the `ssm:GetParameters` action for now?

efekarakus (approved):

iamhopaul123 (approved):
Maybe after this we could add it to `svc show` like what WX does for rdws 🤔

internal/pkg/template/templates/workloads/partials/cf/taskrole.yml
dannyrandall: Talked offline. We're removing it. 😊

PR #3460, opened 2022-04-12
feat: show observability configuration in `svc show` for RDWS

Lou1415926:


internal/pkg/aws/apprunner/service.go
```diff
@@ -20,9 +24,16 @@ type Service struct {
...
 	Port                 string
 	EnvironmentVariables []*EnvironmentVariable
+	Observability        ObservabilityConfiguration
 }
 
 type EnvironmentVariable struct {
 	Name  string
 	Value string
 }
+
```
paragbhingre, line 35: Comment please for this struct.

internal/pkg/describe/rd_web_service.go
```diff
@@ -112,11 +128,38 @@ func (d *RDWebServiceDescriber) Describe() (HumanJSONStringer, error) {
...
+func formatTracingConfiguration(configuration *apprunner.TraceConfiguration) *Tracing {
+	if configuration == nil {
+		return nil
+	}
+	return &Tracing{
+		Vendor: aws.StringValue(configuration.Vendor),
+	}
+}
+
+type Observability struct {
```
paragbhingre: Comment please for the structs

internal/pkg/aws/apprunner/apprunner.go
```diff
@@ -72,6 +73,18 @@ func (a *AppRunner) DescribeService(svcARN string) (*Service, error) {
 	}
 	sort.SliceStable(envVars, func(i int, j int) bool { return envVars[i].Name < envVars[j].Name })
 
+	var observabilityConfiguration ObservabilityConfiguration
+	if resp.Service.ObservabilityConfiguration != nil && aws.BoolValue(resp.Service.ObservabilityConfiguration.ObservabilityEnabled) {
+		if out, err := a.client.DescribeObservabilityConfiguration(&apprunner.DescribeObservabilityConfigurationInput{
+			ObservabilityConfigurationArn: resp.Service.ObservabilityConfiguration.ObservabilityConfigurationArn,
+		}); err == nil {
+			// NOTE: swallow the error otherwise, because observability is an optional description of the service.
+			// Example error: when "EnvManagerRole" doesn't have the "apprunner:ObservabilityConfiguration" permission.
```
iamhopaul123, line 82: If they dont have `apprunner:ObservabilityConfiguration`, can this condition be true? 🤔

```golang
aws.BoolValue(resp.Service.ObservabilityConfiguration.ObservabilityEnabled)
```

internal/pkg/aws/apprunner/service.go
```diff
@@ -20,9 +24,16 @@ type Service struct {
...
+type ObservabilityConfiguration struct {
+	TraceConfiguration *TraceConfiguration
+}
+
+type TraceConfiguration apprunner.TraceConfiguration
```
iamhopaul123, line 42: same here

internal/pkg/describe/rd_web_service.go
```diff
@@ -126,6 +169,7 @@ type rdWebSvcDesc struct {
 	Routes                  []*WebServiceRoute      `json:"routes"`
 	Variables               envVars                 `json:"variables"`
 	Resources               deployedSvcResources    `json:"resources,omitempty"`
+	Observability           []Observability         `json:"observability,omitempty"`
```
iamhopaul123: Should this be
```suggestion
	Observabilities           []Observability         `json:"observabilities,omitempty"`
```

internal/pkg/template/templates/environment/partials/environment-manager-role.yml
```diff
@@ -236,7 +236,8 @@ EnvironmentManagerRole:
             "apprunner:ListServices",
             "apprunner:PauseService",
             "apprunner:ResumeService",
-            "apprunner:StartDeployment"
+            "apprunner:StartDeployment",
+            "apprunner:DescribeObservabilityConfiguration"
```
iamhopaul123, line 240: Do we need to update the env version for this?

iamhopaul123 (commented):
LGTM overall!

internal/pkg/aws/apprunner/apprunner.go
Lou1415926: Typo!! I meant to say `...doesn't have the "apprunner:DescribeObservabilityConfiguration" permission.` i.e. the permission to make that API call.

internal/pkg/describe/rd_web_service.go
Lou1415926: It's just one `observability` configuration - it's only a list because there are several environments (sort of like a map but in the form of a list if I'm to say). I feel like singular is better suited in this particular case 🤔

internal/pkg/aws/apprunner/apprunner.go
iamhopaul123: Hmm but jf we keep doing env upgrade their env manager role should always be able to describe

Lou1415926: It's possible though that they haven't run `copilot svc deploy` before running `copilot svc show`

iamhopaul123: Oh that makes sense!

internal/pkg/describe/rd_web_service.go
```diff
@@ -151,6 +195,20 @@ func (w *rdWebSvcDesc) HumanString() string {
...
+		fmt.Fprintf(writer, "  %s\n", strings.Join(headers, "\t"))
+		fmt.Fprintf(writer, "  %s\n", strings.Join(underline(headers), "\t"))
+		for _, config := range w.Observability {
+			tracingVendor := "None"
+			if config.Tracing != nil && config.Tracing.Vendor != "" {
+				tracingVendor = config.Tracing.Vendor
+			}
+			fmt.Fprintf(writer, "  %s\t%s\n", config.Environment, tracingVendor)
+		}
+	}
```
efekarakus: I wonder if we could have all this logic as a method on the `Observalibility` type? 

```go
func (o *Observability) render(w io.Writer) 
```
so that with Danny's ECS changes we can have all workload types populate the `Observability` struct and then call this method to render which should reduce code duplication

internal/pkg/describe/rd_web_service.go
```diff
@@ -112,11 +129,40 @@ func (d *RDWebServiceDescriber) Describe() (HumanJSONStringer, error) {
...
+	return &tracing{
+		Vendor: aws.StringValue(configuration.Vendor),
+	}
+}
+
+type observabilityPerEnv []observabilityInEnv
+
+type observabilityInEnv struct {
+	Environment string   `json:"environment"`
+	Tracing     *tracing `json:"tracing,omitempty"`
```
iamhopaul123, line 174: qq: why do we make it `*tracing`?

iamhopaul123 (approved):
LGTM!

internal/pkg/describe/rd_web_service.go
Lou1415926: So that `omitempty` works:

> The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.


Having observability doesn't necessarily mean that it has tracing (it could be log, metrics, etc.) so I thought tracing should be rendered optionally in the JSON output.

internal/pkg/describe/rd_web_service.go
```diff
@@ -173,3 +225,25 @@ func (w *rdWebSvcDesc) HumanString() string {
...
+	for _, envObservabilityConfig := range w.Observability {
+		if !envObservabilityConfig.isEmpty() {
+			return true
+		}
+	}
+	return false
+}
+
+func (obs observabilityPerEnv) humanString(w io.Writer) {
+	headers := []string{"Environment", "tracing"}
```
huanjani: just wondering: is there a reason "tracing" isn't capitalized: "Tracing"?

internal/pkg/describe/rd_web_service.go
```diff
@@ -104,6 +111,16 @@ func (d *RDWebServiceDescriber) Describe() (HumanJSONStringer, error) {
 		}
 	}
 
+	isObservabilitiesEmpty := true
+	for _, oc := range observabilities {
+		if !oc.isEmpty() {
+			isObservabilitiesEmpty = false
```
efekarakus: nit:
```suggestion
			isObservabilitiesEmpty = false
			break
```

internal/pkg/describe/rd_web_service.go
```diff
@@ -112,11 +129,40 @@ func (d *RDWebServiceDescriber) Describe() (HumanJSONStringer, error) {
...
+type observabilityInEnv struct {
+	Environment string   `json:"environment"`
+	Tracing     *tracing `json:"tracing,omitempty"`
+}
+
+type tracing struct {
+	Vendor string `json:"vendor"`
+}
+
+func (o observabilityInEnv) isEmpty() bool {
```
efekarakus, line 177: nit: can we move this method under the type

internal/pkg/describe/rd_web_service.go
```diff
@@ -173,3 +225,25 @@ func (w *rdWebSvcDesc) HumanString() string {
 	writer.Flush()
 	return b.String()
 }
+
+func (w *rdWebSvcDesc) hasObservabilityConfiguration() bool {
```
efekarakus: can we make this a method of `observabilityPerEnv` instead of `rdWebSvcDesc` so that it can be leveraged by other service descriptions?

internal/pkg/describe/rd_web_service.go
Lou1415926: Fixed!

efekarakus (approved):

PR #3461, opened 2022-04-12
feat(templates): render min max config in cfn template

### LINKED ISSUE #3427: [Design] Support for configuring MinimumHealthyPercent and MaximumPercent
### Background ◀️

Today, copilot only support [rolling](https://docs.aws.amazon.com/whitepapers/latest/overview-deployment-options/rolling-deployments.html) deployment with `minimumHealthyPercent:100` and `maximumPercent: 200` where ECS spins up new tasks, wait until they are stable before stopping the old tasks.

However, some users want ECS to stop the existing ones before spinning up new ones, for example [issue/3153](https://github.com/aws/copilot-cli/issues/3153). Under the hood, this is setting `minimumHealthyPercent` to `0`, and `maximumPercent` to `100`.

### Proposal
We propose adding support for the above use case through a new field `deployment` in the manifest:

```
deployment:
    rolling: 'recreate' 
```

This will set `minimumHealthyPercent` and `maximumPercent` to `0/100` respectively.

Note that `deployment` is an optional fields in the manifest. If customers doesn’t set any value to it then they will see default behavior of copilot during deployment, which is `100/200` for `min/max`. This maintains backward compatibility.

### An Example
If we do not set `deployment` parameter with the value as rolling: 'recreate' then copilot default behavior will be applied. The default value of minimumHealthyPercent and maximumPercent will be set i.e. 100 and 200. 
If you have desired task count as 4, minimum healthy percent as 100% and maximum percent of 200%. Then ECS can spin up 4 new tasks before even stopping previous 4 tasks

If we do set `deployment` parameter with the value as rolling: 'recreate' then minimumHealthyPercent will be set to 0 and maximumPercent to 100. 
If you have desired task count as 1, minimum healthy percent as 0% and maximum percent of 100%. Then ECS can spin up 1 new task only after stopping the previous task. This is the exact use case we are trying solve through this proposal.

paragbhingre:
This PR resolves #3427 

With this PR we build the logic to configure the values of MinimumHealthyPercent and MaximumPercent based on the `deployment` parameter in the service manifest


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -609,6 +609,18 @@ func convertEntryPoint(entrypoint manifest.EntryPointOverride) ([]string, error)
 	return out, nil
 }
 
+func getDeployConfig(deployStrategyName string) template.DeploymentConfigurationOpts {
```
Lou1415926: tiny tiny nit: 
```suggestion
func convertDeploymentConfig(deployStrategyName string) template.DeploymentConfigurationOpts {
```

On a separate note, what do you think of this signature:
```convertDeploymentConfig(deploymentConfig manifest.DeploymentConfiguration) template.DeploymentConfigurationOpts```
?

File: internal/pkg/deploy/cloudformation/stack/transformers.go [hunk 2]
```diff
@@ -609,6 +609,18 @@ func convertEntryPoint(entrypoint manifest.EntryPointOverride) ([]string, error)
...
+	var deployConfigs template.DeploymentConfigurationOpts
+	if deployStrategyName == "recreate" {
```
dannyrandall: let's use `strings.EqualFold` instead to make the manifest value case insensitive

Lou1415926: nit: we can have a constant (maybe we already do from the last PR?) in `manifest` package `const deploymentStrategyRecreate = "recreate". And here we'd have:
```suggestion
	if deployStrategyName == manifest.deploymentStrategyRecreate {
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -609,6 +609,18 @@ func convertEntryPoint(entrypoint manifest.EntryPointOverride) ([]string, error)
...
+		deployConfigs.MinHealthyPercent = 0
+		deployConfigs.MaxHealthyPercent = 100
```
Lou1415926: We can constant-ize these values too
```
const (
    minHealthyPercentRecreate = 0
    maxPercentRecreate = 100
    minHealthyPercentDefault = 100
    maxPercentDefault = 200
)
```

internal/pkg/deploy/cloudformation/stack/transformers.go
dannyrandall: what do you think about changing the arg's type to `manifest.DeploymentConfiguration`?

internal/pkg/deploy/cloudformation/stack/worker_svc_test.go
```diff
@@ -189,6 +189,9 @@ Outputs:
 					StringSlice: []string{"here"},
 				}
 				svc.manifest.ExecuteCommand = manifest.ExecuteCommand{Enable: aws.Bool(true)}
+				svc.manifest.DeployConfig = manifest.DeploymentConfiguration{
+					Rolling: aws.String("incorrect value"),
```
Lou1415926: This test case looks a little weird to me where you can have "incorrect value" while still producing a default configuration 😆  Like conceptually we should error out instead.

In real life though I suppose this won't happen because of manifest validation. I think we should use either blank, or "recreate", or "default" in this test case.

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -133,6 +135,7 @@ func (s *BackendService) Template() (string, error) {
 		EnvControllerLambda:      envControllerLambda.String(),
 		Storage:                  convertStorageOpts(s.manifest.Name, s.manifest.Storage),
 		Network:                  convertNetworkConfig(s.manifest.Network),
+		DeploymentConfiguration:  deployConfig,
```
dannyrandall: if `getDeployConfig()` doesn't return an error, let's move the call down to here
```suggestion
		DeploymentConfiguration:  getDeployConfig(aws.StringValue(s.manifest.DeployConfig.Rolling)),
```

internal/pkg/manifest/testdata/backend-svc-customhealthcheck.yml
```diff
@@ -37,4 +37,6 @@ exec: true     # Enable running commands in your container.
 # You can override any of the values defined above by environment.
 #environments:
 #  test:
-#    count: 2               # Number of tasks to run for the "test" environment.
\ No newline at end of file
+#    count: 2               # Number of tasks to run for the "test" environment.
+#    deployment:            # deployment strategy for the "test" environment.
+#       rolling: 'recreate' # key is rolling deployment and value is deployment strategy name i.e. either default or recreate.
```
Lou1415926: I think the explanation may be confusing to user. We want to explain what "recreate" is, something like "Existing tasks are stopped before new ones are started"

dannyrandall: i'm wondering if we think `deployment` is common enough that it's worth putting into the manifest template? i don't want to overwhelm with _all_ of the options in the template. any thoughts on this?

internal/pkg/template/workload.go
```diff
@@ -341,6 +341,12 @@ type ObservabilityOpts struct {
 	Tracing string // The name of the vendor used for tracing.
 }
 
+// DeploymentConfiguraitonOpts holds values for MinHealthyPercent and MaxHealthyPercent.
+type DeploymentConfigurationOpts struct {
+	MinHealthyPercent int // MinHealthyPercent parameter represents an upper limit on the number of tasks in a service
+	MaxHealthyPercent int // MaxHealthyPercent represents a lower limit on the number of tasks in a service
```
Lou1415926: I feel like `MaxPercent` is a better name for this field for its real purpose. Also,
```suggestion
         MinHealthyPercent int // MinHealthyPercent represents the lower limit on the number of tasks that should be running for a service during a deployment or when a container instance is draining.
	MaxPercent int // MaxPercent represents the upper limit on the number of tasks that should be running for a service during a deployment or when a container instance is draining.
```

internal/pkg/manifest/testdata/backend-svc-nohealthcheck.yml
```diff
@@ -29,4 +29,6 @@ exec: true     # Enable running commands in your container.
 # You can override any of the values defined above by environment.
 #environments:
 #  test:
-#    count: 2               # Number of tasks to run for the "test" environment.
\ No newline at end of file
+#    count: 2               # Number of tasks to run for the "test" environment.
+#    deployment:            # deployment strategy for the "test" environment.
+#       rolling: 'recreate' # key is rolling deployment and value is deployment strategy name i.e. either default or recreate.
```
Lou1415926: Do we have to update these comments in test manifest data?

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: But `manifest.DeploymentConfiguration` type is different than  `template.DeploymentConfigurationOpts`

internal/pkg/manifest/testdata/backend-svc-customhealthcheck.yml
paragbhingre: So do you mean we shouldn't put it in all the templates and keep it few or you suggest not to keep it at all in the manifest templates?

internal/pkg/deploy/cloudformation/stack/transformers.go
dannyrandall: sorry, i wasn't very clear. i meant the input arg, the same idea @Lou1415926 mentioned [here](https://github.com/aws/copilot-cli/pull/3461#discussion_r849840841) :)

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: So we have `manifest.ecsRecreateRollingUpdateStrategy = "recreate"` from the last PR but as it is not exported we can not use it in the `transformers.go`. 
I can change it to `EcsRecreateRollingUpdateStrategy` in the manifest and use it in the `transformers.go` as `manifest.EcsRecreateRollingUpdateStrategy`

Lou1415926: Let's do it! Maybe instead of `Ecs` we can write `ECS`!

internal/pkg/manifest/testdata/backend-svc-customhealthcheck.yml
dannyrandall: basically i'm thinking about removing these lines from all of the manifest templates
```
#    deployment:            # deployment strategy for the "test" environment.
#       rolling: 'recreate' # key is rolling deployment and value is deployment strategy name i.e. either default or recreate.
```

the idea being we'd document the fields in our documentation, but not necessarily include them in the default template. i'm not really decided one way or the other tbh, so i'm fine leaving it as you have it for now.

Lou1415926: I feel like it'd be a nice little hint for users to be aware that they could use "recreate" for their test environment. The benefit would be a faster deployment in their test deployment. I'm fine with either though!

internal/pkg/manifest/testdata/backend-svc-nohealthcheck.yml
paragbhingre: Yes we need to as test manifest by default has those lines in it when they create new manifest.

internal/pkg/deploy/cloudformation/stack/worker_svc_test.go
paragbhingre: Right, make sense. Manifest validation will never let this kind of inputs go further. So this use case doesn't make sense. I will change it to have "default" value.

internal/pkg/manifest/testdata/backend-svc-nohealthcheck.yml
Lou1415926: <strike>These testdata doesn't test that unfortunately. 

They are data consumed by the integration test whose purpose would be to validate the template, not to validate the default manifest. 

I think in this case then we can remove all these lines in `testdata`</strike>

Sorry I misread the file path! Thought this was `deploy/cloudformation/stack/testdata`. This looks good to me!

internal/pkg/manifest/testdata/backend-svc-customhealthcheck.yml
dannyrandall: Makes sense, that does sound like a common want/pretty nice benefit. Let's leave it in the generated manifest files!

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -43,6 +43,14 @@ const (
 	capacityProviderFargate     = "FARGATE"
 )
 
+// MinimumHealthyPercent and MaximumPercent configurations as per deployment strategy
```
```suggestion
// MinimumHealthyPercent and MaximumPercent configurations as per deployment strategy.
```

internal/pkg/template/templates/workloads/services/backend/manifest.yml
```diff
@@ -55,4 +55,6 @@ exec: true     # Enable running commands in your container.
 # You can override any of the values defined above by environment.
 #environments:
 #  test:
-#    count: 2               # Number of tasks to run for the "test" environment.
\ No newline at end of file
+#    count: 2               # Number of tasks to run for the "test" environment.
+#    deployment:            # deployment strategy for the "test" environment.
```
Lou1415926: nit:
```suggestion
#    deployment:            # The deployment strategy for the "test" environment.
```

internal/pkg/template/templates/workloads/services/lb-web/manifest.yml
```diff
@@ -51,3 +51,5 @@ exec: true     # Enable running commands in your container.
 #environments:
 #  test:
 #    count: 2               # Number of tasks to run for the "test" environment.
+#    deployment:            # deployment strategy for the "test" environment.
```
Lou1415926:
```suggestion
#    deployment:            # The deployment strategy for the "test" environment.
```

internal/pkg/template/templates/workloads/services/worker/manifest.yml
```diff
@@ -63,3 +63,5 @@ subscribe:
 #environments:
 #  test:
 #    count: 2               # Number of tasks to run for the "test" environment.
+#    deployment:            # deployment strategy for the "test" environment.
```
Lou1415926:
```suggestion
#    deployment:            # The deployment strategy for the "test" environment.
```

Lou1415926 (approved):
LGTM just a few more nits!

internal/pkg/template/workload.go
```diff
@@ -341,6 +341,12 @@ type ObservabilityOpts struct {
 	Tracing string // The name of the vendor used for tracing.
 }
 
+// DeploymentConfiguraitonOpts holds values for MinHealthyPercent and MaxPercent.
+type DeploymentConfigurationOpts struct {
+	MinHealthyPercent int // MinHealthyPercent represents the lower limit on the number of tasks that should be running for a service during a deployment or when a container instance is draining.
```
dannyrandall: nit: move docs above struct field

dannyrandall (approved):
looks great!🥳

internal/pkg/template/workload.go
Lou1415926: addressed!

PR #3473, opened 2022-04-18
chore(elbv2): add ListenerRuleHostHeaders

iamhopaul123:
part of #2694

### DIFF
--- a/internal/pkg/aws/elbv2/elbv2.go
+++ b/internal/pkg/aws/elbv2/elbv2.go
@@ -44,7 +44,6 @@ func (e *ELBV2) ListenerRuleHostHeaders(ruleARN string) ([]string, error) {
 		return nil, fmt.Errorf("get listener rule for %s: %w", ruleARN, err)
 	}
 	if len(resp.Rules) == 0 {
-		// Ideally this should never happen.
 		return nil, fmt.Errorf("cannot find listener rule %s", ruleARN)
 	}
 	rule := resp.Rules[0]
@@ -56,7 +55,7 @@ func (e *ELBV2) ListenerRuleHostHeaders(ruleARN string) ([]string, error) {
 			}
 			if condition.HostHeaderConfig == nil {
 				// Ideally this should never happen.
-				continue
+				break
 			}
 			for _, value := range condition.HostHeaderConfig.Values {
 				hostHeaderSet[aws.StringValue(value)] = true
@@ -69,7 +68,7 @@ func (e *ELBV2) ListenerRuleHostHeaders(ruleARN string) ([]string, error) {
 	for hostHeader := range hostHeaderSet {
 		hostHeaders = append(hostHeaders, hostHeader)
 	}
-	sort.SliceStable(hostHeaders, func(i, j int) bool { return hostHeaders[i] < hostHeaders[j] })
+	sort.Slice(hostHeaders, func(i, j int) bool { return hostHeaders[i] < hostHeaders[j] })
 	return hostHeaders, nil
 }
 
--- a/internal/pkg/aws/elbv2/elbv2.go
+++ b/internal/pkg/aws/elbv2/elbv2.go
@@ -62,7 +62,6 @@ func (e *ELBV2) ListenerRuleHostHeaders(ruleARN string) ([]string, error) {
 			for _, value := range condition.HostHeaderConfig.Values {
 				hostHeaderSet[aws.StringValue(value)] = true
 			}
-			// Each rule can optionally include up to one of each of the conditions.
 			break
 		}
 	}
--- a/internal/pkg/aws/elbv2/elbv2.go
+++ b/internal/pkg/aws/elbv2/elbv2.go
@@ -6,6 +6,7 @@ package elbv2
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
@@ -19,6 +20,7 @@ const (
 
 type api interface {
 	DescribeTargetHealth(input *elbv2.DescribeTargetHealthInput) (*elbv2.DescribeTargetHealthOutput, error)
+	DescribeRules(input *elbv2.DescribeRulesInput) (*elbv2.DescribeRulesOutput, error)
 }
 
 // ELBV2 wraps an AWS ELBV2 client.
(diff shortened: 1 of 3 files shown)


internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -33,6 +35,44 @@ func New(sess *session.Session) *ELBV2 {
...
+// ListenerRuleHostHeaders returns all the host headers for a listener rule.
+func (e *ELBV2) ListenerRuleHostHeaders(ruleARN string) ([]string, error) {
+	resp, err := e.client.DescribeRules(&elbv2.DescribeRulesInput{
+		RuleArns: aws.StringSlice([]string{ruleARN}),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("get listener rule for %s: %w", ruleARN, err)
+	}
+	if len(resp.Rules) == 0 {
+		// Ideally this should never happen.
```
efekarakus: nit: no need for these comments :)

internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -33,6 +35,44 @@ func New(sess *session.Session) *ELBV2 {
 	}
 }
 
+// ListenerRuleHostHeaders returns all the host headers for a listener rule.
+func (e *ELBV2) ListenerRuleHostHeaders(ruleARN string) ([]string, error) {
```
efekarakus, line 39: Are you planning to have also a `path-pattern` method? so that we can concatenate the two and give an output to the clients?

internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -33,6 +35,44 @@ func New(sess *session.Session) *ELBV2 {
...
+		// Ideally this should never happen.
+		return nil, fmt.Errorf("cannot find listener rule %s", ruleARN)
+	}
+	rule := resp.Rules[0]
+	hostHeaderSet := make(map[string]bool)
+	for _, condition := range rule.Conditions {
+		if aws.StringValue(condition.Field) == "host-header" {
+			for _, value := range condition.Values {
+				hostHeaderSet[aws.StringValue(value)] = true
+			}
```
efekarakus, line 58: Wut, what's the difference between `condition.Values` and `condition.HostHeaderConfig.Values`?
If I look at our CFN templates, we only have the latter.

efekarakus (approved):

internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -33,6 +35,44 @@ func New(sess *session.Session) *ELBV2 {
...
+			// Each rule can optionally include up to one of each of the conditions.
+			break
+		}
+	}
+	var hostHeaders []string
+	for hostHeader := range hostHeaderSet {
+		hostHeaders = append(hostHeaders, hostHeader)
+	}
+	sort.SliceStable(hostHeaders, func(i, j int) bool { return hostHeaders[i] < hostHeaders[j] })
```
dannyrandall: any reason for stable sort over plain `sort.Slice()`?

internal/pkg/aws/elbv2/elbv2.go
iamhopaul123: Hmm I was planning to continue using `RulePath` 🤔

internal/pkg/aws/elbv2/elbv2.go
iamhopaul123: Yeah but the results always have both two. I check values just to be safe...

internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -33,6 +35,44 @@ func New(sess *session.Session) *ELBV2 {
...
+			}
+			if condition.HostHeaderConfig == nil {
+				// Ideally this should never happen.
+				continue
+			}
+			for _, value := range condition.HostHeaderConfig.Values {
+				hostHeaderSet[aws.StringValue(value)] = true
+			}
+			// Each rule can optionally include up to one of each of the conditions.
+			break
```
dannyrandall, line 65: can you help me understand why we `break` here, but `continue` on L59?

iamhopaul123: Good call! L59 we should break as well

internal/pkg/aws/elbv2/elbv2.go
iamhopaul123: It is just because we want to guarantee we always get the same output every time for the same input. However, here we could possibly use `Slice` because we only have one field for each element.

internal/pkg/aws/elbv2/elbv2.go
efekarakus: OK I just looked this up lol and these field names are so confusing 🤦 

`condition.Values` allows you to specify only a single host name although the type is an array.
`condition.HostHeaderConfig` allows you to specify multiple host names.

So yeah, we should check both 👍 , maybe drop a comment with an explanation?
```suggestion
			// Values is a legacy field that allowed specifying only a single host name.
			// The alternative is to use HostHeaderConfig for multiple values.
			// Only one of these fields should be set, but we collect from both to be safe.
			for _, value := range condition.Values {
				hostHeaderSet[aws.StringValue(value)] = true
			}
```

internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -33,6 +35,43 @@ func New(sess *session.Session) *ELBV2 {
...
+	}
+	rule := resp.Rules[0]
+	hostHeaderSet := make(map[string]bool)
+	for _, condition := range rule.Conditions {
+		if aws.StringValue(condition.Field) == "host-header" {
+			for _, value := range condition.Values {
+				hostHeaderSet[aws.StringValue(value)] = true
+			}
+			if condition.HostHeaderConfig == nil {
+				// Ideally this should never happen.
```
efekarakus: nit:
```suggestion
```

efekarakus (approved):

internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -33,6 +35,43 @@ func New(sess *session.Session) *ELBV2 {
...
+				break
+			}
+			for _, value := range condition.HostHeaderConfig.Values {
+				hostHeaderSet[aws.StringValue(value)] = true
+			}
+			// Each rule can optionally include up to one of each of the conditions.
```
dannyrandall: nit: this doesn't really make sense to me. maybe remove/rephrase it? it looks like we are only considering a single rule's first "host-header" condition.

dannyrandall (approved):

internal/pkg/aws/elbv2/elbv2.go
iamhopaul123: That's exactly what we are doing since there should be only one "host-header" condition 🥲

dannyrandall: right - just not sure that this comment reflects that. maybe just delete it?

PR #3476, opened 2022-04-19
build: upgrade to go1.18

efekarakus:
Unfortunately, the spec for go1.18.1 is not yet available here:
https://github.com/syndbg/goenv/tree/master/plugins/go-build/share/go-build


dannyrandall (approved):
😍😍😍

huanjani (approved):

PR #3478, opened 2022-04-19
refactor(describe): make LB/backend etc. service describers agnostic to env

iamhopaul123:


internal/pkg/describe/rd_web_service.go
```diff
@@ -55,6 +54,15 @@ func NewRDWebServiceDescriber(opt NewServiceConfig) (*RDWebServiceDescriber, err
 	return describer, nil
 }
 
+// ServiceARN retrieves the ARN of the app runner service.
+func (d *RDWebServiceDescriber) ServiceARN(env string) (string, error) {
+	err := d.initClients(env)
+	if err != nil {
+		return "", err
+	}
+	return d.envSvcDescribers[env].ServiceARN()
```
dannyrandall: probably should check and make sure their is a describer for `env` before calling the function so we don't panic
```
ard, ok := d.envSvcDescriber[env]
if !ok || ard == nil {
    // error
}

return ard.ServiceARN()
```

internal/pkg/describe/service.go
```diff
@@ -86,74 +86,111 @@ type serviceStackDescriber struct {
...
+	ignoredResources := map[string]bool{
+		rulePriorityFunction: true,
+		waitCondition:        true,
+		waitConditionHandle:  true,
+	}
+	for _, svcResource := range svcResources {
+		if ignoredResources[svcResource.Type] {
+			continue
+		}
+		resources = append(resources, svcResource)
```
dannyrandall: nit
```suggestion
		if !ignoredResources[svcResource.Type] {
			resources = append(resources, svcResource)
		}
```

internal/pkg/describe/rd_web_service.go
iamhopaul123: Good call! Addressed

dannyrandall (approved):
lgtm🥇

huanjani (approved):
🚢

PR #3479, opened 2022-04-19
chore(acm): implement ValidateCertAliases

iamhopaul123:
Part of #2694


internal/pkg/aws/acm/acm.go
```diff
@@ -24,7 +30,46 @@ func New(s *session.Session) *ACM {
 	}
 }
 
-// ValidateCertAliases is a no-op and is supposed to validates if aliases are all valid against the provided ACM certificates.
+// ValidateCertAliases validates if aliases are all valid against the provided ACM certificates.
 func (a *ACM) ValidateCertAliases(aliases []string, certs []string) error {
+	validAliases := make(map[string]bool)
+	// TODO: Parallelize `findValidAliasesAgainstCert` if needed.
```
efekarakus: 1. Can we remove the TODO? I'm okay with writing it if it _must_ be done
2. Can we do this now?

internal/pkg/aws/acm/acm.go
```diff
@@ -24,7 +30,46 @@ func New(s *session.Session) *ACM {
...
+	domainSet := make(map[string]bool)
+	domainSet[aws.StringValue(resp.Certificate.DomainName)] = true
+	for _, san := range resp.Certificate.SubjectAlternativeNames {
+		domainSet[aws.StringValue(san)] = true
+	}
+	var validAliases []string
+	for _, alias := range aliases {
+		// See https://docs.aws.amazon.com/acm/latest/userguide/acm-certificate.html
+		wildCardMatchedAlias := "*" + alias[strings.Index(alias, "."):]
+		if domainSet[alias] || domainSet[wildCardMatchedAlias] {
```
efekarakus, line 91: would it be safer to evaluate the `domainSet` entries against the alias? Something like

```
for each domain in domainSet:
  if regexp(domain).Matches(alias):
      add alias to valid list
```

iamhopaul123: I think it is strictly this logic because it would only match the closest domain. The wildcard here is not the same meaning as the regex one. For example, `*.example.com` shouldn't match `v1.v2.example.com`

efekarakus: It might be fine, but I just find `wildCardMatchedAlias` a bit strange.
Imagine if `alias` is just `example.com`, then `wildCardMatchedAlias` is `*.com` is this something that we want to check? 💭

iamhopaul123: They won't be able to have any validated cert for `*.com` 🤣  The checking itself is just O(1) since it is map looking up. Technically we can check if they have any SAN with wildcard, and if not we don't need `wildCardMatchedAlias`, but I doubt if it would be more efficient to check it in advance.

internal/pkg/aws/acm/acm.go
```diff
@@ -24,7 +38,58 @@ func New(s *session.Session) *ACM {
...
+	for _, alias := range aliases {
+		if !validAliases[alias] {
+			return fmt.Errorf("%s is not a valid domain against %s", alias, strings.Join(certs, ","))
+		}
+	}
 	return nil
 }
+
+func (a *ACM) findValidAliasesAgainstCert(aliases []string, cert string) ([]string, error) {
+	resp, err := a.client.DescribeCertificate(&acm.DescribeCertificateInput{
```
efekarakus: We can use `DescribeCertificateWithContext` and pass the group's context so that if the context is canceled we don't wait for the response

efekarakus (approved):

huanjani (approved):
:shipit:

PR #3482, opened 2022-04-20
test(e2e): update pipeline test for multiple pipelines

huanjani:
Related: https://github.com/aws/copilot-cli/issues/3231.

Warning: I was not able to trial run this locally, as I was attempting to clone a repo within a repo, so the embedded repo was inaccessible. (If you know of a way to do this locally, let me know!)

### DIFF
--- a/e2e/pipeline/pipeline_test.go
+++ b/e2e/pipeline/pipeline_test.go
@@ -74,7 +74,6 @@ var _ = Describe("pipeline flow", func() {
 				AppName: appName,
 				EnvName: "test",
 				Profile: "e2etestenv",
-				Region: "",
 				Prod:    false,
 			})
 			Expect(err).NotTo(HaveOccurred())
@@ -92,7 +91,7 @@ var _ = Describe("pipeline flow", func() {
 			_, err := copilot.EnvInit(&client.EnvInitRequest{
 				AppName: appName,
 				EnvName: "qa",
-				Profile: "e2eqaenv",
+				Profile: "e2etestenv",
 				Prod:    false,
 			})
 			Expect(err).NotTo(HaveOccurred())
@@ -111,6 +110,8 @@ var _ = Describe("pipeline flow", func() {
 			Expect(envs["test"]).NotTo(BeNil())
 			Expect(envs["prod"]).NotTo(BeNil())
 			Expect(envs["qa"]).NotTo(BeNil())
+			Expect(envs["test"].Region).NotTo(Equal(envs["prod"].Region))
+			Expect(envs["test"].Account).NotTo(Equal(envs["prod"].Account))
 		})
 	})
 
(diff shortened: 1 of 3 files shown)


dannyrandall (approved):

e2e/pipeline/pipeline_test.go
```diff
@@ -87,10 +87,19 @@ var _ = Describe("pipeline flow", func() {
 			})
 			Expect(err).NotTo(HaveOccurred())
 		})
-		It("env ls should list both envs", func() {
+		It("qa env init should succeed", func() {
+			_, err := copilot.EnvInit(&client.EnvInitRequest{
+				AppName: appName,
+				EnvName: "qa",
+				Profile: "e2etestenv",
+				Prod:    false,
```
iamhopaul123: I feel like we should remove `prod` from `EnvInitRequest` completely...

e2e/pipeline/pipeline_test.go
```diff
@@ -87,10 +87,19 @@ var _ = Describe("pipeline flow", func() {
 			})
 			Expect(err).NotTo(HaveOccurred())
 		})
-		It("env ls should list both envs", func() {
+		It("qa env init should succeed", func() {
```
iamhopaul123, line 88: I wonder if the `qa` env is necessary because the less env we have the less run time it will be. Also do you know how long does it usually take after this change?

efekarakus: PH's comment makes me think that it's probably a good idea to create two e2e tests for pipelines. What do you think?

If we go by the idea that each e2e test should test common scenarios, then we see two common scenarios for pipelines:
- Either users have a trunk pipeline `src -> test -> prod`
- Or users have separate pipelines for each env
   -  `src -> test`
   - `src -> prod`

Because of it maybe we should split the test?

e2e/pipeline/s3template.yml
```diff
@@ -0,0 +1,79 @@
...
+              SSEAlgorithm: AES256
+      PublicAccessBlockConfiguration:
+        BlockPublicAcls: true
+        BlockPublicPolicy: true
+
+  e2epipelineaddonBucketPolicy:
+    Metadata:
+      'aws:copilot:description': 'A bucket policy to deny unencrypted access to the bucket and its contents'
+    Type: AWS::S3::BucketPolicy
+    DeletionPolicy: Retain
```
efekarakus: can we delete this line so that the bucket gets deleted once the test is done?

e2e/pipeline/pipeline_test.go
huanjani: I am making a separate multi-pipeline e2e test, and restoring the former (single) pipeline test. But this will make the e2e tests take longer as a whole, because we'll be creating and deleting the `test` and `prod` envs twice.  Does this concern you, @iamhopaul123?

iamhopaul123: No worries! I think making it a separate one sounds good to me.

huanjani:
Closing in favor of #3496

PR #3483, opened 2022-04-20
refactor: augment "graph" pkg to use generics

efekarakus:
#thefutureisnow 

### DIFF
--- a/internal/pkg/graph/graph.go
+++ b/internal/pkg/graph/graph.go
@@ -4,8 +4,6 @@
 // Package graph provides functionality for directed graphs.
 package graph
 
-import "golang.org/x/exp/constraints"
-
 // nodeStatus denotes the visiting status of a node when running DFS in a graph.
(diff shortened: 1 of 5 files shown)


internal/pkg/graph/graph.go
```diff
@@ -14,53 +16,53 @@ const (
 )
 
 // Graph represents a directed graph.
-type Graph struct {
-	nodes map[string]neighbors
+type Graph[V constraints.Ordered] struct {
```
dannyrandall:
```suggestion
type Graph[T comparable] struct {
```

and then on `Edge`/`neighbors`

dannyrandall (commented):
i would ask _whyyyyyy_ since we only (?) use a strings implementation....but i'm excited to have it here so i won't ask :)

dannyrandall (approved):

huanjani (approved):

PR #3485, opened 2022-04-20
fix: grant the bucket owner control to addon template artifacts

### LINKED ISSUE #3453: Cloudformation addon templates fail with S3 error: Access Denied after updating to 1.16.0
We are using a Github v2 source step which requires ACLs be enabled on the artifact bucket and results in the owner of uploaded assets be the codepipeline user.  So, now when this step in the buildspec runs:
copilot-linux svc package -n $svc -e $env --output-dir './infrastructure' --tag $tag --upload-assets

The resulting addons files have ACLs set that make them inaccessible to the cloudformation tasks that run on code deployment in other accounts and cause "S3 error: Access Denied" and the CF task to fail.

I you set "bucket-owner-full-control" on the S3 PUTs I think everything would work.

Lou1415926:
Fix #3453 by:
1. Grant bucket owner (which is the app account owner) full control, in addition to the object owner (the account that uploads the artifact)
2. Use hash for addon template artifacts
3. Instead of uploading addons to the root dir "stackset/svc.addons.stack.yml", upload it to "stackset/manual/addons/svc/sha.yml"
4. Instead of uploading env files to "manual/sha/mock.env.file",  upload it to "manual/env-files/env.file/sha.env"


internal/pkg/aws/s3/s3.go
```diff
@@ -191,12 +197,24 @@ func (s *S3) isBucketExists(bucket string) (bool, error) {
 	return true, nil
 }
 
-func (s *S3) upload(bucket, key string, buf io.Reader) (string, error) {
-	resp, err := s.s3Manager.Upload(&s3manager.UploadInput{
+type uploadInputOption func(in *s3manager.UploadInput)
+
+func withACLBucketOwnerFullControl() func(in *s3manager.UploadInput)  {
```
efekarakus: Since S3's recommendation is to have the bucket owner own every object (https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html) should we remove `withACLBucketOwnerFullControl` as a function and just inline this setting in `upload`?

internal/pkg/aws/s3/s3.go
```diff
@@ -78,25 +79,30 @@ func (s *S3) ZipAndUpload(bucket, key string, files ...NamedBinary) (string, err
...
 
+// AddonsArtifactPath prefixes the key with the addons artifact path.
+func AddonsArtifactPath(key string) string {
+	return path.Join(artifactDirName, addonDirName, key)
+}
+
+// TemplateArtifactPath prefixes the key with the template artifact path.
+func TemplateArtifactPath(key string) string {
+	return path.Join(artifactDirName, templateDirName, key)
+}
```
efekarakus: These don't seem to fit the `s3` package 💭   
What do you think of moving them to `deploy` instead?

internal/pkg/deploy/cloudformation/workload.go
```diff
@@ -39,7 +39,8 @@ func (cf CloudFormation) pushWorkloadTemplateToS3Bucket(bucket string, config St
 		return "", fmt.Errorf("generate template: %w", err)
 	}
 	reader := strings.NewReader(template)
-	url, err := cf.s3Client.Upload(bucket, fmt.Sprintf(fmtWorkloadCFNTemplateName, config.StackName(), sha256.Sum256([]byte(template))), reader)
+	artifactName := path.Join(config.StackName(), fmt.Sprintf("%x.yml", sha256.Sum256([]byte(template))))
+	url, err := cf.s3Client.Upload(bucket, s3.TemplateArtifactPath(artifactName), reader)
```
efekarakus: What do you think of a path like:
```
templates/<stack name>/<sha>.yml
```

efekarakus (commented):
Thank you Wanxian!!

internal/pkg/deploy/cloudformation/workload.go
Lou1415926: Currently it's `manual/templates/stackname/sha.yml` what do ya think? The new code here doesn't change the path it's just some refactoring!

internal/pkg/aws/s3/s3.go
Lou1415926: I moved them to the `template/` package. My motivation was that it might be easier for us to manage/visualize from the codebase what our artifact folder structure would be, if these path definition is put together in a single package.

In the current codebase that are scattered in `cli/deploy`, `deploy` and `s3`. 

let me know what ya think!

internal/pkg/aws/s3/s3.go
```diff
@@ -192,11 +178,13 @@ func (s *S3) isBucketExists(bucket string) (bool, error) {
 }
 
 func (s *S3) upload(bucket, key string, buf io.Reader) (string, error) {
-	resp, err := s.s3Manager.Upload(&s3manager.UploadInput{
+	in := &s3manager.UploadInput{
 		Body:   buf,
 		Bucket: aws.String(bucket),
 		Key:    aws.String(key),
-	})
+		ACL: aws.String(s3.ObjectCannedACLBucketOwnerFullControl),
```
efekarakus, line 185: nit: is gofmt or a similar formatter running?

internal/pkg/cli/deploy/deploy.go
```diff
@@ -854,8 +858,9 @@ func (d *workloadDeployer) pushAddonsTemplateToS3Bucket(in *pushAddonsTemplateTo
 		}
 		return "", fmt.Errorf("retrieve addons template: %w", err)
 	}
-	reader := strings.NewReader(template)
-	url, err := in.uploader.Upload(d.resources.S3Bucket, fmt.Sprintf(deploy.AddonsCfnTemplateNameFormat, d.name), reader)
+	reader := strings.NewReader(tmpl)
+	artifactName := path.Join(d.name, fmt.Sprintf("%x.yml", sha256.Sum256([]byte(tmpl))))
```
iamhopaul123: I wonder if we should move `sha256.Sum256` to `AddonsArtifactPath`. Same for `TemplateArtifactPath` because it seems like whenever we use them we always call `sha256.Sum256`

internal/pkg/deploy/env.go
```diff
@@ -13,7 +13,7 @@ const (
 	// LegacyEnvTemplateVersion is the version associated with the environment template before we started versioning.
 	LegacyEnvTemplateVersion = "v0.0.0"
 	// LatestEnvTemplateVersion is the latest version number available for environment templates.
-	LatestEnvTemplateVersion = "v1.9.0"
```
iamhopaul123: we don't have to do this. `v1.9.0` hasn't been released.

internal/pkg/cli/deploy/deploy.go
Lou1415926: I would prefer the key being determined by the client but I'm fine either way

Lou1415926: addressed

internal/pkg/template/template.go
```diff
@@ -95,6 +100,21 @@ func New() *Template {
 	}
 }
 
+// AddonsArtifactPath prefixes the key with the addons artifact path.
+func AddonsArtifactPath(key string) string {
```
efekarakus: I wonder if we should have a new package for all of these? so that it reads a bit better `template.MkdirSHA256` is not obvious that it's for s3.


Let me know what you think of this layout:
```go
// Package artifactpath holds functions to generate the S3 object path for artifacts.
package template/artifactpath

// Addons returns the path to store addon artifact files.
func Addons(key string) string

// CFNTemplate returns the path to store cloudformation templates.
func CFNTemplate(key string) string

// SHA256Func returns a function that will namespace a key with the sha256 hash of its content.
func SHA256Func(content []byte) func(string) string  {
    return func(key string) string {
        return path.Join(fmt.Sprintf("%x", sha256.Sum256(content)), key)
    }
}

// Chain applies the path functions sequentially to nest the key.
func Chain(key string, fn ...func(string) string) string 
```

This would be the client's experience:
```go
artifactpath.Chain(key, artifactpath.SHA256Func(content), artifactpath.CFNTemplate)
```

internal/pkg/cli/deploy/deploy.go
iamhopaul123: But do we have an assumption that all the addons artifacts should do `sha256.Sum256`? Same for the env file and template uploads. If so maybe we would want
```go
// Env file upload
url, err := in.uploader.Upload(d.resources.S3Bucket, template.EnvFileArtifactPath(envFilePath, content), reader)

// Addons upload
url, err := in.uploader.Upload(d.resources.S3Bucket, template.AddonsArtifactPath(d.name, content), reader)

// Template upload
url, err := cf.s3Client.Upload(bucket, template.TemplateArtifactPath(config.StackName(), content), reader)
```

it's ok if you don't want to change. Not a big one tho.

iamhopaul123 (approved):

internal/pkg/cli/deploy/deploy.go
efekarakus: I agree that we should most likely always upload with sha256 :) this way we can eliminate this bug altogether for future code changes

internal/pkg/template/template.go
Lou1415926: Since we always want to add SHA256, I've inlined it into the function -> chaining was rendered unnecessary when SHA256 is inlined. Let me know what you think!

internal/pkg/template/artifactpath/artifactpath.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// MkdirSHA prefixes the key with the SHA256 hash of the contents of "manual/<hash>/key".
+func MkdirSHA256(key string, content []byte) string {
+	return path.Join(s3ArtifactDirName, fmt.Sprintf("%x", sha256.Sum256(content)), key)
+}
+
+// Addons returns the path to store addon artifact files.
+// Example: manual/addons/key/sha.yml.
+func AddonsWithSHA256(key string, content []byte) string {
```
efekarakus: nit: It should be fine to drop the WithSHA256 suffix now that we do it for everything

efekarakus (approved):

PR #3487, opened 2022-04-20
chore: env upgrade should add new params

iamhopaul123:


internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -144,9 +147,25 @@ func (e *EnvStackConfig) Parameters() ([]*cloudformation.Parameter, error) {
 			ParameterValue: aws.String(fmt.Sprintf(fmtServiceDiscoveryEndpoint, e.in.Name, e.in.App.Name)),
 		},
 		{
-			ParameterKey:   aws.String(createHTTPSListener),
+			ParameterKey:   aws.String(envParamCreateHTTPSListenerKey),
 			ParameterValue: aws.String(httpsListener),
 		},
+		{
```
efekarakus, line 153: Was there a reason why you added these boss? there is a default in the template that's `""`

efekarakus (approved):
LGTM!

internal/pkg/cli/env_upgrade.go
```diff
@@ -284,7 +284,8 @@ func (o *envUpgradeOpts) upgradeEnvironment(upgrader envUpgrader, conf *config.E
 	if err := upgrader.UpgradeEnvironment(&deploy.CreateEnvironmentInput{
 		Version: toVersion,
 		App: deploy.AppInformation{
-			Name: conf.App,
+			Name:   conf.App,
+			Domain: app.Domain,
```
dannyrandall: should we check that `conf.App == app.Name`?

internal/pkg/deploy/cloudformation/stack/env.go
iamhopaul123: discussed offline. It is because otherwise `env upgrade` would think we are trying to deprecate a parameter when we upgrade the env. I'll send a follow-up PR to include an integ test, making sure the `Parameter()` sets all the environment parameters we have in the template.

internal/pkg/cli/env_upgrade.go
iamhopaul123: I'll just use `app.Name`

dannyrandall (approved):

internal/pkg/cli/env_upgrade.go
```diff
@@ -315,7 +316,8 @@ func (o *envUpgradeOpts) upgradeLegacyEnvironment(upgrader legacyEnvUpgrader, co
 		if err := upgrader.UpgradeLegacyEnvironment(&deploy.CreateEnvironmentInput{
 			Version: toVersion,
 			App: deploy.AppInformation{
-				Name: conf.App,
+				Name:   conf.App,
```
dannyrandall: here too :)

PR #3489, opened 2022-04-21
chore: add skaffold to connect pipeline manifest deployments to template

efekarakus:


internal/pkg/deploy/pipeline.go
```diff
@@ -483,7 +484,16 @@ func (stg *PipelineStage) Init(env *config.Environment, mftStage *manifest.Pipel
 		Region:    env.Region,
 		AccountID: env.AccountID,
 	}
-	stg.localWorkloads = workloads
+	deployments := mftStage.Deployments
+	if len(mftStage.Deployments) == 0 {
```
iamhopaul123:
```suggestion
	deployments := mftStage.Deployments
	if len(deployments) == 0 {
```

iamhopaul123 (approved):

Makefile
```diff
@@ -86,7 +86,7 @@ package-custom-resources-clean:
 
 .PHONY: run-unit-test
 run-unit-test:
-	go test -race -count=1 -coverprofile=${COVERAGE} ${PACKAGES}
+	go test -coverprofile=${COVERAGE} ${PACKAGES}
```
dannyrandall, line 89: why remove race detector?

internal/pkg/manifest/validate_test.go
```diff
@@ -1002,23 +1002,81 @@ func TestPipelineManifest_Validate(t *testing.T) {
...
 			if tc.wantedError != nil {
 				require.EqualError(t, gotErr, tc.wantedError.Error())
+				return
+			}
+			if tc.wantedErrorMsgPrefix != "" {
+				require.Error(t, gotErr)
+				require.Contains(t, gotErr.Error(), tc.wantedErrorMsgPrefix)
+				return
+			}
+			require.NoError(t, gotErr)
```
dannyrandall: nit
```suggestion
			switch {
			case tc.wantedErr != nil:
				require.EqualError(t, gotErr, tc.wantedError.Error())
			case tc.wantedErrorMsgPrefix != "":
				require.Error(t, gotErr)
				require.Contains(t, gotErr.Error(), tc.wantedErrorMsgPrefix)
			default:
				require.NoError(t, gotErr)
			}
```

internal/pkg/deploy/pipeline.go
```diff
@@ -523,43 +533,74 @@ func (stg *PipelineStage) EnvManagerRoleARN() string {
...
 	}
-	return actions
+	return actions, nil
+}
+
+func (stg *PipelineStage) buildDeploymentsGraph() *graph.Graph[string] {
+	var names []string
+	for name, _ := range stg.deployments {
+		names = append(names, name)
+	}
```
dannyrandall, line 590: nit: https://pkg.go.dev/golang.org/x/exp/maps#Keys

internal/pkg/deploy/pipeline.go
```diff
@@ -523,43 +533,74 @@ func (stg *PipelineStage) EnvManagerRoleARN() string {
...
 
 	var prevActions []orderedRunner
-	for _, deployment := range stg.Deployments() {
-		prevActions = append(prevActions, &deployment)
+	deployActions, err := stg.Deployments()
+	if err != nil {
+		return nil, err
+	}
+	for _, action := range deployActions {
+		prevActions = append(prevActions, &action)
```
dannyrandall: this will fill `prevActions` with the same action (the last one in deployActions). gotta do something like
```
for i := range deployActions {
    prevActions = append(prevActions, &deployActions[i])
}
```

Makefile
efekarakus: It was slowing down unit tests! I thought that I can send a separate CR where in the `ci` workflow we run the test with and without race detector, what do you think?

internal/pkg/deploy/pipeline.go
efekarakus: 🤦 tyyy!

internal/pkg/deploy/pipeline.go
efekarakus: I keep debating if we want to import the `exp` module but I think I'd rather wait until they're promoted to the stdlib, I'll keep it as is for now

internal/pkg/manifest/validate_test.go
efekarakus: Done! I was keeping it consistent w the rest of tests but this is cleaner 😍

Makefile
dannyrandall: fair enough :) we can leave it off for now, but the `ci` approach sounds good to me.

internal/pkg/deploy/pipeline.go
dannyrandall: Sounds good 👍

dannyrandall (approved):
lgtm!

PR #3495, opened 2022-04-22
chore: update URI for cert import

iamhopaul123:
Related #2694. Specifically, instead of grabbing DNS info from env parameters, we now describe the listener rule for it.


internal/pkg/describe/service.go
```diff
@@ -108,24 +113,35 @@ func newServiceStackDescriber(opt NewServiceConfig, env string) (*serviceStackDe
 
 // Params returns the parameters of the service stack.
 func (d *serviceStackDescriber) Params() (map[string]string, error) {
+	if d.params != nil {
```
efekarakus, line 116: Can we think of any scenario where caching these values will be problematic? is it at all possible that a stack update happens in between calls and causes issues or do you think we're safe?

internal/pkg/template/templates/environment/partials/environment-manager-role.yml
```diff
@@ -19,6 +19,17 @@ EnvironmentManagerRole:
       PolicyDocument:
         Version: '2012-10-17'
         Statement:
+{{- if .ImportCertARNs}}
+        - Sid: ACM
```
efekarakus: nit:
```suggestion
        - Sid: ImportedCertificates
```

efekarakus (approved):
LGTM, just a question!

internal/pkg/describe/service.go
iamhopaul123: I was thinking about the same question but couldn't think of a scenario that's problematic for now. Unless we want to use the same struct to get params and then wait and then get it again. Maybe for `status/show` if we want to support "refreshing periodically to get the latest update" like what `--follow` does for `logs`. However, I think we could add a field to disable this caching when we need it. What do you think?

internal/pkg/describe/service.go
```diff
@@ -108,24 +113,35 @@ func newServiceStackDescriber(opt NewServiceConfig, env string) (*serviceStackDe
...
 	if err != nil {
 		return nil, err
 	}
+	d.params = descr.Parameters
 	return descr.Parameters, nil
 }
 
 // Params returns the outputs of the service stack.
 func (d *serviceStackDescriber) Outputs() (map[string]string, error) {
+	if d.outputs != nil {
```
dannyrandall, line 129: just to check, these functions with a cache are only called from a single goroutine, right?

internal/pkg/describe/uri.go
```diff
@@ -94,40 +93,46 @@ func (d *LBWebServiceDescriber) URI(envName string) (string, error) {
...
-		dnsName := fmt.Sprintf("%s.%s", d.svc, envOutputs[envOutputSubdomain])
-		uri.DNSNames = []string{dnsName}
-		uri.HTTPS = true
-	}
-	aliases := envParams[stack.EnvParamAliasesKey]
-	if aliases != "" {
-		value := make(map[string][]string)
-		if err := json.Unmarshal([]byte(aliases), &value); err != nil {
-			return albURI{}, err
+	if !ok || isHTTPS != "true" {
```
dannyrandall: nit
```suggestion
	if isHTTPS != "true" {
```

internal/pkg/template/templates/workloads/partials/cf/env-controller.yml
```diff
@@ -5,8 +5,10 @@ EnvControllerAction:
   Properties:
     ServiceToken: !GetAtt EnvControllerFunction.Arn
     Workload: !Ref WorkloadName
+{{- if not .UseImportedCerts}}
```
dannyrandall: nit
```suggestion
{{- if and (not .UseImportedCerts) (.Aliases)}}
```
plus remove extra `end`

dannyrandall (approved):

internal/pkg/describe/service.go
iamhopaul123: yeah

PR #3496, opened 2022-04-22
test(e2e): add separate multi-pipeline test

huanjani:
The existing (single) pipeline e2e test has a pipeline deploying a 'frontend' service to 'test' and 'prod' envs.
This adds a new e2e test that has two pipelines ('test' and 'prod') following two git branches ('test' and 'prod'), each deploying a 'frontend' service _with addons_ to a single env, 'test' and 'prod', respectively.


e2e/multi-pipeline/pipeline_suite_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Command-line tools.
+var (
+	copilot *client.CLI
+	aws     *client.AWS
+)
+
+// Application identifiers.
+var (
+	appName          = fmt.Sprintf("e2e-pipeline-%d", time.Now().Unix())
```
efekarakus: nit:
```suggestion
	appName          = fmt.Sprintf("e2e-multipipeline-%d", time.Now().Unix())
```

e2e/multi-pipeline/pipeline_test.go
```diff
@@ -0,0 +1,525 @@
...
+			cmd.Stdout = os.Stdout
+			cmd.Stderr = os.Stderr
+			cmd.Dir = repoName
+			Expect(cmd.Run()).NotTo(HaveOccurred())
+
+			cmd = exec.Command("git", "config", "user.name", "e2etest")
+			cmd.Stdout = os.Stdout
+			cmd.Stderr = os.Stderr
+			cmd.Dir = repoName
+			Expect(cmd.Run()).NotTo(HaveOccurred())
```
efekarakus: nit: can we move these to the `"setup CodeCommit repository"` section so that we need to repeat ourselves while pushing to the separate branches

e2e/multi-pipeline/pipeline_test.go
```diff
@@ -0,0 +1,525 @@
...
+	})
+
+	Context("when creating the test pipeline stack", func() {
+		It("checks out the test git branch", func() {
+			cmd := exec.Command("git", "checkout", "-b", "test")
+			cmd.Stdout = os.Stdout
+			cmd.Stderr = os.Stderr
+			cmd.Dir = repoName
+			Expect(cmd.Run()).NotTo(HaveOccurred())
+		})
```
efekarakus: Should we be creating the branch in the `"when creating the test pipeline manifest"` section so that it tracks the `test` branch instead of `master`?

And in these sections we can do `"git", "checkout", "test"`

efekarakus (commented):
Looks great!

We would also need to update the buildspec to enable the e2e test like here: https://github.com/aws/copilot-cli/blob/89cc9c4aac547f389dab4ccac3ecd40dfea7e4ec/.release/buildspec_e2e.yml#L112-L122

My recommendation is to pick one of the env regions to be `ap-northeast-3` (so that it's in Japan as this feature is popular there) and another `ap-southeast-1`

.release/buildspec_e2e.yml
```diff
@@ -120,6 +120,17 @@ batch:
...
           PRODENV_REGION: eu-central-1
+    - identifier: multi_pipeline
+      env:
+        privileged-mode: true
+        type: LINUX_CONTAINER
+        compute-type: BUILD_GENERAL1_LARGE
+        image: aws/codebuild/standard:5.0
+        variables:
+          TEST_SUITE: multi-pipeline
+          APP_REGION: eu-west-1
```
efekarakus: nit: what do you think of making the app_region also `ap-northeast-3`

efekarakus (approved):

dannyrandall (approved):
lgtm - though it's making me wonder if it'd be nice to have some helper functions like `CreateApp()`/`Environment()` or something. maybe that wouldn't work super well with ginkgo, idk.

PR #3497, opened 2022-04-22
feat(pipelines): support workload deployment orders in pipelines

### LINKED ISSUE #3402: [Design] Ordering deployments in a pipeline

efekarakus:
Resolves #3402


internal/pkg/graph/graph.go
```diff
@@ -100,21 +123,67 @@ func (g *Graph[V]) hasCycles(temp *findCycleTempVars[V], currVertex V) bool {
...
+}
+
+func (bfs *LevelOrder[V]) traverse(g *Graph[V], root V) {
+	queue := []V{root}
+	bfs.marked[root] = true
+	bfs.ranks[root] = 0
+	for {
+		if len(queue) == 0 {
+			return
+		}
```
dannyrandall: nit
```suggestion
	for len(queue) > 0 {
```

internal/pkg/graph/graph.go
```diff
@@ -100,21 +123,67 @@ func (g *Graph[V]) hasCycles(temp *findCycleTempVars[V], currVertex V) bool {
 	return false
 }
 
-// TopologicalSorter ranks vertices in a graph using topological sort.
-type TopologicalSorter[V comparable] struct {
-	ranks map[V]int
+// LevelOrder ranks vertices in a breadth-first search manner.
+type LevelOrder[V comparable] struct {
+	marked map[V]bool
```
dannyrandall: would it make sense to just create a new `marked` map inside `traverse()` each time it's called instead of having it on the struct? then it wouldn't have to get reset [here (L185)](https://github.com/aws/copilot-cli/pull/3497/files#diff-c28ac7314b17a9809b0254ccb9d73bb5ff6dad0081dd0329ac77e85a8f598cd4R185)

efekarakus: Thanks! much better, fixed it.

internal/pkg/graph/graph.go
```diff
@@ -100,21 +123,65 @@ func (g *Graph[V]) hasCycles(temp *findCycleTempVars[V], currVertex V) bool {
...
-	// TODO(efekarakus): Implement me.
-	return 0, nil
+// The second boolean return value is used to indicate whether the vertex exists in the graph.
+func (bfs *LevelOrder[V]) Rank(vtx V) (int, bool) {
+	r, ok := bfs.ranks[vtx]
+	return r, ok
+}
+
+func (bfs *LevelOrder[V]) traverse(g *Graph[V], root V) {
+	bfs.marked = make(map[V]bool) // Reset all markings before each run.
```
dannyrandall: nit :) 
```suggestion
	marked := make(map[V]bool) // Reset all markings before each run.
```

iamhopaul123: It seems like `marked` is only used within the `traverse()` func. Why do we make it a struct variable?

internal/pkg/graph/graph.go
```diff
@@ -100,21 +123,65 @@ func (g *Graph[V]) hasCycles(temp *findCycleTempVars[V], currVertex V) bool {
...
-	return nil, nil
+//
+// An example graph and their ranks is shown below to illustrate:
+// .
+//├── a          rank: 0
+//│   ├── c      rank: 1
+//│   │   └── f  rank: 2
+//│   └── d      rank: 1
+//└── b          rank: 0
+//    └── e      rank: 1
```
iamhopaul123, line 204: Nice explanation!

internal/pkg/graph/graph.go
```diff
@@ -82,6 +83,28 @@ func (g *Graph[V]) IsAcyclic() ([]V, bool) {
 	return nil, true
 }
 
+// Roots return a slice of vertices with no incoming edges.
```
dannyrandall: nit
```suggestion
// Roots returns a slice of vertices with no incoming edges.
```

internal/pkg/graph/graph_test.go
```diff
@@ -91,3 +92,157 @@ func TestGraph_IsAcyclic(t *testing.T) {
...
+			graph: &Graph[string]{
+				vertices: map[string]neighbors[string]{
+					"frontend": map[string]bool{
+						"backend": true,
+					},
+					"backend": map[string]bool{
+						"frontend": true,
+					},
+				},
+			},
```
dannyrandall: what do you think of creating the graph using the public API instead? like in `TestGraph_Roots()`

internal/pkg/graph/graph.go
```diff
@@ -100,21 +123,65 @@ func (g *Graph[V]) hasCycles(temp *findCycleTempVars[V], currVertex V) bool {
...
+	bfs.ranks[root] = 0
+	for len(queue) > 0 {
+		var vtx V
+		vtx, queue = queue[0], queue[1:]
+		for neighbor := range g.vertices[vtx] {
+			if bfs.marked[neighbor] {
+				continue
+			}
+			bfs.marked[neighbor] = true
+			if rank := bfs.ranks[vtx] + 1; rank > bfs.ranks[neighbor] { // Is the new rank higher than a previous traversal?
```
iamhopaul123: I might be stupid but I am a little confused here 🥲 Only when `bfs.marked[neighbor]` is false can the program proceed here, but then does that mean `bfs.ranks[neighbor]` will be empty too?

I feel like if a vertex is visited, we shouldn't skip too quickly but possibly think about if we should update the rank and update with the lowest rank rather than the highest. For example:
```
a -> b -> c
d -> c -> f
```
where both `a` and `d` has 0 rank. If we give `c` rank 2 rather than 1 then `f` will share the same rank as `c`.

internal/pkg/graph/graph.go
efekarakus: Done, addressed all feedback

efekarakus: Sorry that was a mistake! removed the field

internal/pkg/graph/graph.go
efekarakus: lol it's not stupid - 

> but then does that mean bfs.ranks[neighbor] will be empty too?

I think your example shows how it's possible that `ranks` is not empty. In your example, after `a -> b -> c` we will have:
```go
map[string]int {
    "a": 0,
    "b": 1,
    "c": 2
}
```
Then when we start from the second root, `d -> c -> f` we **do not** want the rank of `c` to become 1. It has a higher rank from a previous deployment, so we want to keep it at 2. The final ranks from the algorithm will be:
```go
map[string]int {
   "a": 0,
   "d": 0,
   "b": 1,
   "c": 2,
   "f": 3,
}
```

We can never set `c` to have rank 1 otherwise it will run in parallel with `b` which should happen prior to `c`. Hope this makes sense!

dannyrandall (approved):
😋😋😋😋😋😋😋😋😋😋😋😋😋😋😋😋😋😋😋😋😋😋😋😋

internal/pkg/graph/graph.go
iamhopaul123: Yeah now I agree we should set `c` to 2. However, if we start from `d` then since `c` is visited, it won't get updated to a higher rank `2`

efekarakus: it will boss because of this `if` statement. 
Since each time `traverse` is called `marked` is reset. After starting from `d` we will have: 
```
ranks: {"d": 0, "c": 1, f: "2"}
```
Then we reset all marks so `c` is no longer marked. 
When we start from `a`, in this if-statement we get:
```go
if rank /* 2 */ := bfs.ranks[vtx] /* rank of "b" = 1 */ + 1; rank > bfs.ranks[neighbor ] /*  2 > 1 so update ranks["c"] */
```

efekarakus:
boss has a good point where the wrong rank will be calculated for the following scenario:
```
a -> f  -> d
a -> b -> c -> d
```
in this scneario, `d` will have rank 2 instead of 3. So added the label until this test case and a modification is submitted

internal/pkg/graph/graph.go
```diff
@@ -15,7 +15,8 @@ const (
 
 // Graph represents a directed graph.
 type Graph[V comparable] struct {
-	vertices map[V]neighbors[V]
+	vertices  map[V]neighbors[V] // Adjacency list for each vertex.
+	indegrees map[V]int          // Number of incoming edges for each vertex.
```
iamhopaul123:
```suggestion
	inDegrees map[V]int          // Number of incoming edges for each vertex.
```

iamhopaul123 (approved):
LGTM just a nit 🚢

internal/pkg/graph/graph.go
```diff
@@ -28,15 +29,33 @@ type neighbors[V comparable] map[V]bool
...
+	neighbors, ok := g.vertices[vtx]
+	if !ok {
+		return nil
+	}
+	arr := make([]V, len(neighbors))
+	i := 0
+	for neighbor := range neighbors {
+		arr[i] = neighbor
+		i += 1
+	}
```
dannyrandall, line 55: nit
```suggestion
	var arr []V
	for neighbor := range neighbors {
		arr = append(arr, neighbor)
	}
```

dannyrandall (approved):
lgtm!

PR #3508, opened 2022-04-27
test(e2e): change env region

huanjani:
Service Discovery is not supported in the previous region (ap-northeast-3).


iamhopaul123 (approved):

efekarakus (approved):
So sad

PR #3513, opened 2022-04-28
refactor: make template package SSM agnostic

Lou1415926:
This PR removes the dependency of `config` package from `template` package.


internal/pkg/template/env.go
```diff
@@ -41,14 +39,38 @@ type EnvOpts struct {
...
 	ImportCertARNs []string
-	Telemetry      *config.Telemetry
+	Telemetry      *Telemetry
 
 	LatestVersion string
 }
 
+type VPCConfig struct {
+	Imported *ImportVPC
+	Managed  ManagedVPC
```
efekarakus: nit: can we add a comment here
```suggestion
	Imported *ImportVPC // If not-nil, use the imported VPC settings instead of the defaults from ManagedVPC.
	Managed  ManagedVPC
```

efekarakus (approved):

internal/pkg/template/env.go
Lou1415926: I tweaked the wording a little:
```
// If not-nil, use the imported VPC resources instead of the Managed VPC.
```
Because the managed VPC could be configured by user input (e.g. configuring cidr, az)

huanjani (approved):
LGTM!

PR #3514, opened 2022-04-29
fix: task run `--generate-cmd` should add quotation marks around map value

### LINKED ISSUE #3458: --generate-cmd secretsmanager YAML not well-formed
### Reproduction

1. clone https://github.com/hirokihokari/copilot-test
2. `copilot app init` # as "copilot-nginx"
3. `copilot init`
4. `copilot deploy`
5. `copilot task run --generate-cmd copilot-nginx/test/nginx`
6. run the generated command
```
provision resources for task copilot-test-nginx: check if changeset is empty: create change set copilot-<some id> for stack task-copilot-test-nginx: ValidationError: Template format error: YAML not well-formed. (line 72, column 120)
	status code: 400, request id: <some id>: describe change set copilot-<some id> for stack task-copilot-test-nginx: ChangeSetNotFound: ChangeSet [copilot-<some id>] does not exist
	status code: 404, request id: <some id>
```

7. Remove the secrets block from `manifest.yml`
8. `copilot deploy`
9. `copilot task run --generate-cmd copilot-nginx/test/nginx`
10. succeeds

### double colon?
If I remove the double colon at the end of secret arn, this error goes away and it switches to 
```
ResourceInitializationError: ... secrets manager: failed to retrieve secret from arn:aws:secretsmanag...
```

Is that the right format for secret manager arn?

I can't tell either syntax is wrong (trying to look for mal-formed arn=non-existent secret), or #3256 is not in effect, since the error message is truncated in the terminal as well as [on console](https://github.com/aws/containers-roadmap/issues/1133)

Lou1415926:
Fix: #3458

In addition, it fixes an application selector bug: when an additional option is present, where the selection would default to that additional option 


huanjani (approved):
Thank you for doing this!

dannyrandall (approved):

PR #3520, opened 2022-05-02
chore(manifest): add `network` to environment manifest

Lou1415926:
This PR:
1. Define environment's `network` field which contains VPC configurations
2. Implement validation to `network`
3. Translate `network` to CFN data


internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -127,18 +132,28 @@ func (e *EnvStackConfig) importedVPC() *template.ImportVPC {
...
+		CIDR:               DefaultVPCCIDR,
+		PublicSubnetCIDRs:  DefaultPublicSubnetCIDRs,
+		PrivateSubnetCIDRs: DefaultPrivateSubnetCIDRs,
+	}
+	// If a manifest is present, it is the only place we look at.
+	if e.in.Mft != nil {
+		if v := e.in.Mft.Network.VPC.ManagedVPC(); v != nil {
+			return *v
 		}
+		return defaultManagedVPC
```
efekarakus, line 145: out of curiosity, was there a reason why you prefered the `e.in.Network.VPC.ManagedVPC()` not to return the default VPC? 

What do you think of this method signature?
```
vpc, ok := Mft.Network.VPC.ManagedVPC() // ok is true means customer has custom settings, when false returns default copilot vpc configuration.

vpc := Mft.Network.VPC.ImportedVPC() // nil when user doesn't have custom import vpc settings
```

internal/pkg/manifest/env.go
```diff
@@ -0,0 +1,100 @@
...
+package manifest
+
+import (
+	"sort"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/copilot-cli/internal/pkg/template"
+)
+
+// Environment holds the configuration to build an environment.
```
```suggestion
// Environment is the manifest configuration for an environment.
```

internal/pkg/manifest/env_test.go
```diff
@@ -0,0 +1,243 @@
...
+		mockPublicSubnet2CIDR  = IPNet("10.0.1.0/24")
+		mockPrivateSubnet1CIDR = IPNet("10.0.3.0/24")
+		mockPrivateSubnet2CIDR = IPNet("10.0.4.0/24")
+	)
+	testCases := map[string]struct {
+		inContent    string
+		wantedStruct Environment
+	}{
+		"unmarshal with managed VPC": {
+			inContent: `name: test
```
efekarakus, line 29: awesome tests, and it is so cool that we're seeing the env manifest come to life!

internal/pkg/manifest/env_test.go
```diff
@@ -0,0 +1,243 @@
...
+		t.Run(name, func(t *testing.T) {
+			var env Environment
+			err := yaml.Unmarshal([]byte(tc.inContent), &env)
+			require.NoError(t, err)
+			require.Equal(t, tc.wantedStruct, env)
+		})
+	}
+}
+
+func TestEnv_ImportedVPC(t *testing.T) {
```
efekarakus: nit: should these be renamed to `TestEnvironmentVPCConfig_ImportedVPC`?

internal/pkg/manifest/validate_env.go
```diff
@@ -0,0 +1,145 @@
...
+	if v.managedVPCCustomized() {
+		return v.validateManagedVPC()
+	}
+	return nil
+}
+
+func (v environmentVPCConfig) validateImportedVPC() error {
+	for _, subnet := range v.Subnets.Public {
+		if aws.StringValue(subnet.SubnetID) == "" {
+			return errors.New(`validate "subnets": all subnets must be imported (with "id" field), not configured (with "cidr" or "az" fields), if vpc is imported`)
```
efekarakus: it might be helpful to include the index here like `validate "subnets": public[%d] must include "id" field if vpc is imported`

(same for all validations)

efekarakus (commented):
Amazing, thank you for all the effort!

internal/pkg/deploy/cloudformation/stack/env.go
Lou1415926: > out of curiosity, was there a reason why you preferred the e.in.Network.VPC.ManagedVPC() not to return the default VPC?

Not particularly "preferred". I thought of `Mft` here as "where user input comes in", and thought of `default` as a non-user-input - that was the mindset behind it!

I think one day we will want to let `manifest` package handles the default, eventually removing this middle layer. But I was hesitant to do it here because other contributors might find it confusing that sometimes the defaults is handled in `manifest` while other times in in `stack`.

> What do you think of this method signature?
```
if vpc := Mft.Network.VPC.ManagedVPC(); vpc != nil {
   return *vpc
}
```
I felt like if we are letting `manifest` handle the defaults here, then it's probably okay if we just return `vpc` without the `bool` because `stack` wouldn't need that info. What do ya think?

internal/pkg/manifest/env.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (v environmentVPCConfig) imported() bool {
+	return aws.StringValue(v.ID) != ""
+}
+
+func (v environmentVPCConfig) managedVPCCustomized() bool {
+	return aws.StringValue((*string)(v.CIDR)) != ""
+}
+
+func (v environmentVPCConfig) ImportedVPC() *template.ImportVPC {
```
efekarakus, line 45: nit: comment plz 🥺  (same for `ManagedVPC`)

internal/pkg/manifest/env_test.go
```diff
@@ -0,0 +1,243 @@
...
+		inVPCConfig environmentVPCConfig
+		wanted      *template.ManagedVPC
+	}{
+		"default vpc without custom configuration": {},
+		"with imported vpc": {
+			inVPCConfig: environmentVPCConfig{
+				ID: aws.String("vpc-1234"),
+			},
+		},
+		"with custom configuration to managed vpc": {
```
efekarakus: nit: maybe rename this test case to be a bit more verbose putting the emphasis on what it's catching:
```suggestion
		"ensure custom configuration is sorted by AZ": {
```

efekarakus (approved):

internal/pkg/manifest/validate_env.go
```diff
@@ -0,0 +1,145 @@
...
+		if aws.StringValue(subnet.SubnetID) == "" {
+			return fmt.Errorf(`validate "subnets": public[%d] must include "id" field if the vpc is imported`, idx)
+		}
+	}
+	for idx, subnet := range v.Subnets.Private {
+		if aws.StringValue(subnet.SubnetID) == "" {
+			return fmt.Errorf(`validate "subnets": private[%d] must include "id" field if the vpc is imported`, idx)
+		}
+	}
+	if len(v.Subnets.Private)+len(v.Subnets.Public) <= 0 {
```
dannyrandall: nit: i like `switch` :)

internal/pkg/manifest/env.go
```diff
@@ -0,0 +1,100 @@
...
+	if !v.imported() {
+		return nil
+	}
+	var publicSubnetIDs, privateSubnetIDs []string
+	for _, subnet := range v.Subnets.Public {
+		publicSubnetIDs = append(publicSubnetIDs, aws.StringValue(subnet.SubnetID))
+	}
```
dannyrandall, line 52:
```suggestion
	publicSubnetIDs := aws.StringValueSlice(v.Subnets.Public)
```

dannyrandall: same below

dannyrandall (approved):
lgtm, just a couple nit's

internal/pkg/manifest/env.go
Lou1415926: not super applicable because we are extracting the `subnet.SubnetID` field. `v.Subnets.Public` itself is a list of struct :(

dannyrandall: ahhh you're right😊

dannyrandall (approved):

PR #3526, opened 2022-05-04
chore: pin otel container version, change config to ecs-xray

dannyrandall:

### DIFF
--- a/internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
+++ b/internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
@@ -28,7 +28,9 @@
 {{- end}}
 {{- if eq .Observability.Tracing "AWSXRAY"}}
 - Name: aws-otel-collector
-  Image: public.ecr.aws/aws-observability/aws-otel-collector:latest
+  Image: public.ecr.aws/aws-observability/aws-otel-collector:v0.17.0
+  Command:
+    - --config=/etc/ecs/ecs-xray.yaml
   LogConfiguration:
     LogDriver: awslogs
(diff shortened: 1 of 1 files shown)


internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
```diff
@@ -28,7 +28,9 @@
 {{- end}}
 {{- if eq .Observability.Tracing "AWSXRAY"}}
 - Name: aws-otel-collector
-  Image: public.ecr.aws/aws-observability/aws-otel-collector:latest
+  Image: public.ecr.aws/aws-observability/aws-otel-collector:v0.17.0
+  Command:
+    - --config=/etc/ecs/ecs-xray.yaml
```
iamhopaul123, line 33: Could you remind me again why we change the config to `ecs-xray`?

dannyrandall: sure, a couple of things it does differently from the default one:
- doesn't enable metric export
- includes ecs/ec2 resource detection
- runs otlp on default ports (default runs the http receiver on a legacy port)

you can see ecs-xray [here](https://github.com/aws-observability/aws-otel-collector/blob/7d69912415c27c33cc796c4ca834c7c05332476e/config/ecs/ecs-xray.yaml) and the default [here](https://github.com/aws-observability/aws-otel-collector/blob/7d69912415c27c33cc796c4ca834c7c05332476e/config.yaml)

iamhopaul123 (approved):

Lou1415926 (approved):

PR #3528, opened 2022-05-04
chore(manifest): add `observability` to env manifest

Lou1415926:
This PR:
1. Define environment's observability field which cluster's container insights configuration
2. Implement validation
3. Translate it to CFN data


internal/pkg/manifest/validate_env.go
```diff
@@ -142,3 +142,7 @@ func (c subnetConfiguration) Validate() error {
 	}
 	return nil
 }
+
+func (o environmentObservability) Validate() error {
```
iamhopaul123, line 163: comment pls

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -69,7 +69,7 @@ Resources:
       Configuration:
         ExecuteCommandConfiguration:
           Logging: DEFAULT
-{{- if .Telemetry }}
+{{- if and .Telemetry .Telemetry.EnableContainerInsights}}
```
iamhopaul123, line 72: Does it work if `.Telemetry.EnableContainerInsights` is `&false`?

efekarakus (approved):
Nothing to add besides PH's comments, after addressing his feedback it LGTM

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -69,7 +69,7 @@ Resources:
...
       ClusterSettings:
         - Name: containerInsights
           {{- if .Telemetry.EnableContainerInsights}}
```
dannyrandall: i think (from https://github.com/golang/go/issues/12995 and some local testing) that this only tests if the `.Telemetry.EnableContainerInsights` is `nil` now that it's a `*bool` - which, because of L72, will always be true. i think you may need to register a "deref" function or something when creating the template to deref the value here

internal/pkg/template/templates/environment/cf.yml
Lou1415926: good catch!

internal/pkg/manifest/env.go
```diff
@@ -14,12 +14,12 @@ import (
...
 type Environment struct {
 	Workload          `yaml:",inline"`
-	environmentConfig `yaml:",inline"`
+	EnvironmentConfig `yaml:",inline"`
 
 	parser template.Parser
 }
 
-type environmentConfig struct {
+type EnvironmentConfig struct {
```
efekarakus, line 23: doc comment plz 🥺

internal/pkg/template/template_functions.go
```diff
@@ -86,6 +86,9 @@ func ToSnakeCaseFunc(s string) string {
 // IncFunc increments an integer value and returns the result.
 func IncFunc(i int) int { return i + 1 }
 
+// BoolValue dereference a boolean pointer.
+func BoolValue(p *bool) bool { return aws.BoolValue(p) }
```
efekarakus: Should we instead add a method to Telemetry?

```go
func (t *Telemetry) EnableContainerInsights() bool {
   if t == nil { return false }
   return aws.BoolValue(t.ContainerInsights)
}
```

Lou1415926: Updated! 

I want the `ClusterSettings` blob to be rendered only when `ContainerInsights` is explicitly set to true/false; that is it shouldn't be in the template if `ContainerInsights` is `nil`. That's why there is a bit of churn here.

See if you like what it is now!

internal/pkg/manifest/validate_integration_test.go
```diff
@@ -67,6 +68,12 @@ func Test_ValidateAudit(t *testing.T) {
 			require.NoError(t, err)
 		})
 	}
+
+	// Audit environment manifest.
+	t.Run("environment manifest", func(t *testing.T) {
```
efekarakus, line 73: 👏

efekarakus (approved):

dannyrandall (approved):

PR #3531, opened 2022-05-04
chore(cfn): add internal LB to env template

huanjani:
for this step, just http; https to come


internal/pkg/template/templates/environment/cf.yml
```diff
@@ -390,6 +470,31 @@ Outputs:
     Value: !Ref DefaultHTTPTargetGroup
     Export:
       Name: !Sub ${AWS::StackName}-DefaultHTTPTargetGroup
+  InternalLoadBalancerDNSName:
```
efekarakus: Would you mind checking which one of these outputs are we using in our codebase?

For example, I don't think service stacks need the `DefaultInternalALBHTTPTargetGroup` 🤔

internal/pkg/deploy/cloudformation/env_test.go
```diff
@@ -100,8 +101,10 @@ func TestCloudFormation_UpgradeEnvironment(t *testing.T) {
 					})
 				})
 				s3 := mocks.NewMocks3Client(ctrl)
-				s3.EXPECT().Upload("mockbucket", "manual/templates/phonetool-test/6806328b58e482e354cfd1879fecc310113fcf1b95c6bdf058c486f440087847.yml", gomock.Any()).Return("url", nil)
-
+				s3.EXPECT().Upload("mockbucket", gomock.Any(), gomock.Any()).DoAndReturn(func(bucket, key string, data io.Reader) (string, error) {
```
Lou1415926: The previous way of doing it (with the `6806328b58e482e354cfd1879fecc310113fcf1b95c6bdf058c486f440087847`) has its merit of testing that we are using the sha as the name. But this LGTM as well. Just adding my thoughts here!

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -104,7 +108,28 @@ Resources:
       Tags:
         - Key: Name
           Value: !Sub 'copilot-${AppName}-${EnvironmentName}-lb'
-  # Only accept requests coming from the public ALB or other containers in the same security group.
+  InternalLoadBalancerSecurityGroup:
+    Metadata:
+      'aws:copilot:description': 'A private security group for your internal load balancer allowing HTTP traffic'
```
Lou1415926: nit:
```suggestion
      'aws:copilot:description': 'A security group for your internal load balancer allowing HTTP traffic from within the VPC'
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -146,6 +179,20 @@ Resources:
       Subnets: [ {{range $id := .VPCConfig.Imported.PublicSubnetIDs}}{{$id}}, {{end}} ]
 {{- else}}
       Subnets: [ {{range $ind, $cidr := .VPCConfig.Managed.PublicSubnetCIDRs}}!Ref PublicSubnet{{inc $ind}}, {{end}} ]
+{{- end}}
+      Type: application
+  InternalLoadBalancer:
+    Metadata:
+      'aws:copilot:description': 'An internal Application Load Balancer to distribute private traffic to your services'
```
Lou1415926: nit: maybe just to be more specific?
```suggestion
      'aws:copilot:description': 'An internal Application Load Balancer to distribute private traffic from within the VPC to your services'
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -146,6 +179,20 @@ Resources:
       Subnets: [ {{range $id := .VPCConfig.Imported.PublicSubnetIDs}}{{$id}}, {{end}} ]
 {{- else}}
       Subnets: [ {{range $ind, $cidr := .VPCConfig.Managed.PublicSubnetCIDRs}}!Ref PublicSubnet{{inc $ind}}, {{end}} ]
+{{- end}}
+      Type: application
+  InternalLoadBalancer:
```
Lou1415926, line 190: What do you think of writing it such that resources related to internal LB are close together? Would it be easier in the future to navigate? This LGTM as well though!!

Lou1415926 (approved):

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -104,7 +108,28 @@ Resources:
...
+    Condition: CreateInternalALB
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      GroupDescription: Access to the internal load balancer
+      SecurityGroupIngress:
+        - CidrIp: 0.0.0.0/0
+          Description: Allow from anyone within the VPC on port 80
+          FromPort: 80
+          IpProtocol: tcp
+          ToPort: 80
```
efekarakus: I wonder if this is what we want to do (anything in the VPC being able to make requests to the load balancer) vs. only allowing resources with the `EnvironmentSecurityGroup` associated to make requests to the LB.

So just like how you have `EnvironmentSecurityGroupIngressFromInternalALB` there would be another one called `InternalALBIngressFromEnvironmentSecurityGroup`.

Would you mind reaching out to the various internal ALB github issues and asking them if they would want only their copilot services to make requests to the ALB or anything in the VPC to have access to it?

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -104,7 +108,28 @@ Resources:
...
+{{- if .VPCConfig.Imported}}
+      VpcId: {{.VPCConfig.Imported.ID}}
+{{- else}}
+      VpcId: !Ref VPC
+{{- end}}
+      Tags:
+        - Key: Name
+          Value: !Sub 'copilot-${AppName}-${EnvironmentName}-ilb'
```
efekarakus: nit:
```suggestion
          Value: !Sub 'copilot-${AppName}-${EnvironmentName}-internal-lb'
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -390,6 +470,26 @@ Outputs:
     Value: !Ref DefaultHTTPTargetGroup
     Export:
       Name: !Sub ${AWS::StackName}-DefaultHTTPTargetGroup
+  InternalLoadBalancerDNSName:
+    Condition: CreateInternalALB
+    Value: !GetAtt InternalLoadBalancer.DNSName
+    Export:
+      Name: !Sub ${AWS::StackName}-InternalLoadBalancerDNS
+  InternalLoadBalancerFullName:
```
efekarakus: Can I ask for a favor? 🥺  maybe let's remove these outputs (for now) and only add them when you make modification to the backend service CFN template so that we can guarantee that only outputs that we need get exported

efekarakus (commented):
Looks great to me! 😍

We would also need to update the `deploy.LatestEnvTemplateVersion` to `v1.10`

internal/pkg/template/templates/environment/cf.yml
efekarakus: I know that a user mentioned that the ALB should be accessible from any instance in the VPC but I think we should start from least-privileged and then expand the scope of this security group. What do you think?

So let's start  with only allowing ingress only from the Environment Security Group. With the environment manifest around the corner, we'll allow folks to customize the security group.
If other clients also want to increase the scope, we can then decide to output the `InternalLoadBalancerSecurityGroup` so that they can write a  `AWS::EC2::SecurityGroupIngress` addon

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -131,6 +150,22 @@ Resources:
...
+      IpProtocol: -1
+      SourceSecurityGroupId: !Ref InternalLoadBalancerSecurityGroup
+  InternalALBIngressFromEnvironmentSecurityGroup:
+    Type: AWS::EC2::SecurityGroupIngress
+    Condition: CreateInternalALB
+    Properties:
+      Description: Ingress from the env security group
+      GroupId: !Ref EnvironmentSecurityGroup
+      IpProtocol: -1
+      SourceSecurityGroupId: !Ref InternalLoadBalancerSecurityGroup
```
efekarakus: these should be the opposite 😂 #copypaste 
```suggestion
      GroupId: !Ref InternalLoadBalancerSecurityGroup
      IpProtocol: -1
      SourceSecurityGroupId: !Ref EnvironmentSecurityGroup
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -150,6 +185,20 @@ Resources:
...
+  InternalLoadBalancer:
+    Metadata:
+      'aws:copilot:description': 'An internal Application Load Balancer to distribute private traffic from within the VPC to your services'
+    Condition: CreateInternalALB
+    Type: AWS::ElasticLoadBalancingV2::LoadBalancer
+    Properties:
+      Scheme: internal
+      SecurityGroups: [ !GetAtt InternalLoadBalancerSecurityGroup.GroupId ]
+{{- if .VPCConfig.Imported}}
+      Subnets: [ {{range $id := .VPCConfig.Imported.PrivateSubnetIDs}}{{$id}}, {{end}} ]
```
efekarakus, line 199: 💭 
Unrelated to this PR: even with this design proposal (https://github.com/aws/copilot-cli/issues/3538#issuecomment-1130316476), we should keep this behavior:
- By default the internal ALB is spread among any private subnets. However, if the flag is provided they can limit it to whatever subnets they want

efekarakus (approved):
Feel free to remove the label once the feedback is addressed!

PR #3534, opened 2022-05-05
chore(manifest): add `http` to environment manifest

Lou1415926:
This PR:
1. Define environment's http field which contains internet configurations related to LB
2. Implement validation to the field
3. Translate it to CFN data


internal/pkg/manifest/env.go
```diff
@@ -114,3 +115,11 @@ type environmentObservability struct {
 func (o environmentObservability) IsEmpty() bool {
 	return o.ContainerInsights == nil
 }
+
+type environmentHTTPConfig struct {
+	Internet internetHTTPConfig `yaml:"internet"`
```
efekarakus: I'm torn about the field name, because we have "public" subnets, `network.placement: 'public'`, and app runner also uses the term "public" for endpoints that have access from the internet (https://aws.amazon.com/blogs/containers/deep-dive-on-aws-app-runner-vpc-networking/). On the other hand, for elb it's `internet-facing`.

I'm leaning towards `public` instead what do you think? for consistency

internal/pkg/manifest/validate_env.go
```diff
@@ -163,3 +163,16 @@ func (c subnetConfiguration) Validate() error {
...
+func (o environmentHTTPConfig) Validate() error {
+	if err := o.Internet.Validate(); err != nil {
+		return fmt.Errorf(`validate "internet": %w`, err)
+	}
+	return nil
+}
+
+// Validate returns nil if internetHTTPConfig is configured correctly.
+func (o internetHTTPConfig) Validate() error {
+	return nil
```
efekarakus, line 183: maybe one type of validation we can do is call `arn.Parse` and if that returns an error we can wrap it? to ensure that the inputs are indeed ACM arns 💭

internal/pkg/manifest/env.go
Lou1415926: So it's `internet` & `internal` vs. `public` & `private` right?

I think what you proposed sound good. Both `network.public` & `http.public` would map to something "with a public IP", & `network.private` & `http.private` map to something without a public IP.

efekarakus: yeah, I think that makes sense to me too

internal/pkg/manifest/validate_env.go
```diff
@@ -163,3 +164,21 @@ func (c subnetConfiguration) Validate() error {
...
+		return fmt.Errorf(`validate "public": %w`, err)
+	}
+	return nil
+}
+
+// Validate returns nil if publicHTTPConfig is configured correctly.
+func (o publicHTTPConfig) Validate() error {
+	for idx, certARN := range o.Certificates {
+		if _, err := arn.Parse(certARN); err != nil {
+			return fmt.Errorf(`validate "certificates[%d]": %w`, idx, err)
```
iamhopaul123, line 180: I like `certificates[%d]` which makes the error msg cleaner.
nit:
```suggestion
			return fmt.Errorf(`parse "certificates[%d]": %w`, idx, err)
```

iamhopaul123 (approved):

internal/pkg/manifest/validate_env.go
Lou1415926: I felt like `validate` better describes what we are doing here though. We are just parsing for the sake of validating 🤔  Are you concerned because you don't want user to think that we are validating certs by upserting records? In this case, what do you think of `validate "certificates[%d] arn": %w`?

iamhopaul123: i just feel like the error msg wrapping should describe the behavior. Since we are just parsing the arn i feel like we should wrap the error like this. But it's ok just a nit.

internal/pkg/manifest/validate_env.go
```diff
@@ -163,3 +164,21 @@ func (c subnetConfiguration) Validate() error {
...
+	if err := o.Public.Validate(); err != nil {
+		return fmt.Errorf(`validate "public": %w`, err)
+	}
+	return nil
+}
+
+// Validate returns nil if publicHTTPConfig is configured correctly.
+func (o publicHTTPConfig) Validate() error {
+	for idx, certARN := range o.Certificates {
+		if _, err := arn.Parse(certARN); err != nil {
```
efekarakus, line 179: nit: we can also check if 

```go
arn, err := arn.Parse(certARN)
arn.Service == "acm"
```

efekarakus (approved):

internal/pkg/manifest/validate_env.go
Lou1415926: RIP will do in the next PR :laughcry:

PR #3539, opened 2022-05-05
chore: add InternalALBWorkloads param

huanjani:
```
if o.ALBEnabled {
  parameters = append(parameters, "InternalALBWorkloads,")
}
```
will be more fleshed out with manifest changes. For now, it's a noop.


internal/pkg/template/workload.go
```diff
@@ -627,6 +627,11 @@ func envControllerParameters(o WorkloadOpts) []string {
 		}
 		parameters = append(parameters, "Aliases,") // YAML needs the comma separator; resolved in EnvContr.
 	}
+	if o.WorkloadType == "Backend Service" {
```
efekarakus, line 630: I can already hear in my head @dannyrandall asking for a `switch` statement

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -37,6 +37,7 @@ const (
 	envParamAppDNSDelegationRoleKey  = "AppDNSDelegationRole"
 	EnvParamAliasesKey               = "Aliases"
 	EnvParamALBWorkloadsKey          = "ALBWorkloads"
+	EnvParamInternalALBWorkloadsKey  = "InternalALBWorkloads"
```
efekarakus, line 40: nit: can this be private

efekarakus (commented):
We're in this odd situation where if we merge #3531 without this PR (and vice-versa) then the deployment will fail because either we're trying to create add the `InternalALBWorkloads` parameter to a stack where the param doesn't exist in the template or the other way around.

My recommendation would be to:
* Potentially merge the two PRs together 😢 
* Or, Add a `Default: ""` to #3531 and then remove it in this PR

internal/pkg/template/workload.go
dannyrandall: lol :joy: they are prettier :sob:

huanjani:
Closing-- changes now in #3531.

PR #3542, opened 2022-05-06
ci: add conventional-commits job to validate PR titles

efekarakus:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,10 @@ name: ci
 
 on:
   pull_request:
+    types:
+      - opened
+      - edited
+      - synchronize
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


efekarakus:
Added the label to not merge until v1.18 release is done - feel free to play around with the PR title to test things out

.github/workflows/ci.yml
```diff
@@ -114,3 +118,20 @@ jobs:
...
+        uses: actions/github-script@v6
+        with:
+          script: |
+            // See https://gist.github.com/marcojahn/482410b728c31b221b70ea6d2c433f0c#file-conventional-commit-regex-md
+            const regex = /^(build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test){1}(\([\w\-\.]+\))?(!)?: ([\w ])+([\s\S]*)/g;
+
+            const pr = context.payload.pull_request;
+            const title = pr.title;
+
+            if (title.match(regex) == null) {
```
Lou1415926, line 135: Previously, if the PR has only one commit, that one commit message needs to be semantic even if the title already is. Do we still need to check that?

efekarakus: nah I thought that was super annoying before 😂 because mergify squashes all commits and instead uses the PR title and body in the commit message:
https://github.com/aws/copilot-cli/blob/762e6cb39cf2571c6ef93b561b4aa3358b6ac53a/.mergify.yml#L65-L68

So we should only be checking the PR title for the best experience

Lou1415926 (approved):

dannyrandall (approved):

PR #3549, opened 2022-05-10
chore(cli): implement env deployer that uploads artifacts

Lou1415926:


internal/pkg/cli/deploy/env.go
```diff
@@ -0,0 +1,68 @@
...
+// UploadArtifacts uploads the deployment artifacts for the environment.
+func (d *envDeployer) UploadArtifacts() (map[string]string, error) {
+	envRegion := d.env.Region
+
+	resources, err := d.getAppRegionalResources()
+	if err != nil {
+		return nil, err
+	}
+	if resources.S3Bucket == "" {
+		log.Errorf("Cannot find the S3 artifact bucket in %s region created by app %s. The S3 bucket is necessary for many future operations. For example, when you need addons to your services.", envRegion, d.app.Name)
```
huanjani: Should we break this up into two or three lines, for width reasons?

huanjani: maybe?
```suggestion
		log.Errorf("Cannot find the S3 artifact bucket in region %s created by app %s. The S3 bucket is necessary for many future operations. For example, when you need addons to your services.", envRegion, d.app.Name)
```

internal/pkg/cli/deploy/env.go
```diff
@@ -0,0 +1,68 @@
...
+		return nil, fmt.Errorf("cannot find the S3 artifact bucket in %s region", envRegion)
```
huanjani: (same as above)
```suggestion
		return nil, fmt.Errorf("cannot find the S3 artifact bucket in region %s", envRegion)
```

internal/pkg/cli/deploy/env.go
```diff
@@ -0,0 +1,68 @@
...
+func (d *envDeployer) getAppRegionalResources() (*stack.AppRegionalResources, error) {
+	if d.appRegionalResources != nil {
+		return d.appRegionalResources, nil
+	}
+	resources, err := d.appCFN.GetAppResourcesByRegion(d.app, d.env.Region)
+	if err != nil {
+		return nil, fmt.Errorf("get app resources in region %s: %w", d.env.Region, err)
+	}
+	return resources, nil
+}
```
huanjani, line 65: tiny nit: add extra line at end of this file and the test file?

internal/pkg/cli/deploy/env.go
iamhopaul123: Why do we need this log msg? I feel it is not a common error unless they remove the bucket manually?

internal/pkg/cli/deploy/env.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/aws/copilot-cli/internal/pkg/aws/s3"
+	"github.com/aws/copilot-cli/internal/pkg/config"
+	"github.com/aws/copilot-cli/internal/pkg/deploy/cloudformation/stack"
+	"github.com/aws/copilot-cli/internal/pkg/manifest"
+	"github.com/aws/copilot-cli/internal/pkg/term/log"
+)
+
+type appResourcesGetter interface {
+	GetAppResourcesByRegion(app *config.Application, region string) (*stack.AppRegionalResources, error)
+	GetRegionalAppResources(app *config.Application) ([]*stack.AppRegionalResources, error)
```
iamhopaul123: Are we using this?

huanjani (approved):
Just a few nits; otherwise, LGTM!

internal/pkg/cli/deploy/env.go
Lou1415926: The log message was moved over from `cli/env_init.go` but sure we can remove it!

iamhopaul123 (approved):

internal/pkg/cli/deploy/env.go
Lou1415926: I have it locally but after I pushed it got removed 🤔

PR #3554, opened 2022-05-11
chore(cli): use env deployer to update an environment

Lou1415926:

### DIFF
--- a/internal/pkg/cli/deploy/env.go
+++ b/internal/pkg/cli/deploy/env.go
@@ -5,22 +5,38 @@ package deploy
 
 import (
 	"fmt"
+	"os"
 
+	"github.com/aws/copilot-cli/internal/pkg/aws/cloudformation"
 	"github.com/aws/copilot-cli/internal/pkg/aws/s3"
 	"github.com/aws/copilot-cli/internal/pkg/config"
 	"github.com/aws/copilot-cli/internal/pkg/deploy/cloudformation/stack"
 	"github.com/aws/copilot-cli/internal/pkg/manifest"
+
+	"github.com/aws/copilot-cli/internal/pkg/aws/partitions"
+	"github.com/aws/copilot-cli/internal/pkg/deploy"
+	termprogress "github.com/aws/copilot-cli/internal/pkg/term/progress"
 )
 
 type appResourcesGetter interface {
 	GetAppResourcesByRegion(app *config.Application, region string) (*stack.AppRegionalResources, error)
 }
 
+type deployEnvironmentInput struct {
+	rootUserARN         string
+	isProduction        bool
+	customResourcesURLs map[string]string
+}
+
+type environmentDeployer interface {
+	UpdateAndRenderEnvironment(out termprogress.FileWriter, env *deploy.CreateEnvironmentInput, opts ...cloudformation.StackOption) error
+}
+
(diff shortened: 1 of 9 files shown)


internal/pkg/deploy/cloudformation/env.go
```diff
@@ -28,35 +28,72 @@ const (
 
 // DeployAndRenderEnvironment creates the CloudFormation stack for an environment, and render the stack creation to out.
 func (cf CloudFormation) DeployAndRenderEnvironment(out progress.FileWriter, env *deploy.CreateEnvironmentInput) error {
```
efekarakus: nit: should we rename this method to 
```suggestion
func (cf CloudFormation) CreateAndRenderEnvironment(out progress.FileWriter, env *deploy.CreateEnvironmentInput) error {
```

internal/pkg/cli/deploy/env.go
```diff
@@ -53,6 +67,34 @@ func (d *envDeployer) UploadArtifacts() (map[string]string, error) {
 	return urls, nil
 }
 
+// DeployEnvironment deploys an environment using CloudFormation.
+func (d *envDeployer) DeployEnvironment(in *deployEnvironmentInput) error {
```
efekarakus, line 71: from the method name it feels like it should do either CreateAndRender or UpdateAndRender depending on whether the stack exists or not - did you have something else in mind?

Lou1415926: The workflow in my mind is:
1. `env init` does `Create` only. It should error out if the env stack already exists, because `env init` is expected to run against only new environments.
2. `env deploy` does `Update` only. It should and will error out if the env stack does not exist. The stack should have been created by `env init`.

This `DeployEnvironment` method is meant to do the latter "update", and I wanted to define another method for the former that's typically called by `env init`:
```go
func (d *envDeployer) BootstrapEnvironment() {
    ...
    input := &deploy.CreateEnvironmentInput {
        Name: d.env.Name,
       App: deploy.AppInformation{
			Name:                d.app.Name,
			Domain:              d.app.Domain,
			AccountPrincipalARN: in.rootUserARN,
		},
		AdditionalTags:       d.app.Tags,
     }
    d.envDeployer.CreateAndRenderEnvironment(os.Stderr, input)
}
```

dannyrandall (approved):

internal/pkg/cli/deploy/env.go
```diff
@@ -5,44 +5,58 @@ package deploy
...
+	"github.com/aws/copilot-cli/internal/pkg/aws/partitions"
+	"github.com/aws/copilot-cli/internal/pkg/deploy"
+	termprogress "github.com/aws/copilot-cli/internal/pkg/term/progress"
 )
 
 type appResourcesGetter interface {
 	GetAppResourcesByRegion(app *config.Application, region string) (*stack.AppRegionalResources, error)
 }
 
+type deployEnvironmentInput struct {
```
efekarakus, line 25: I assume these will become public later when it gets used

efekarakus (approved):

internal/pkg/cli/deploy/env.go
Lou1415926: yupp

PR #3560, opened 2022-05-12
chore(log): add Logger struct

efekarakus:
This new type should make it easy to add a logger to our other packages that can be discarded or unit tested.


internal/pkg/term/log/log.go
```diff
@@ -6,10 +6,17 @@ package log
 
 import (
 	"fmt"
+	"io"
 
 	"github.com/fatih/color"
 )
 
+// Wrapper writers around standard error and standard output that work on windows.
```
huanjani: I know this comment existed before, but would this change make it make more sense?
```suggestion
// Wrap writers around standard error and standard output that work on windows.
```

dannyrandall (approved):

internal/pkg/term/log/log.go
efekarakus: oh sorry these are meant to be "wrapper" like decorated writers.
I'll update the comment so it's not confusing!

iamhopaul123 (approved):

Issue #3556, opened 2022-05-11
Environment upgrade from 1.8.0 -> 1.9.0 failed: AccessDenied when upload template to bucket


TillaTheHun0:
On a recent service deploy, `copilot` CLI attempted to transitively upgrade the Copilot environment from `1.8.0` to `1.9.0`. For past environment upgrades, this has gone off without a hitch. Though this time, an error was received when attempting to upload the new service template to the s3 bucket in the App AWS account:

```
✘ execute "env upgrade --app foo-app --name preview": upgrade environment preview from version v1.8.0 to version v1.9.0: upload manual/templates/***/18918392bc70edd***.yml to bucket stackset-foo-app-inf-pipelinebuiltartifactbuc-1sgrupkj5d4eq****: AccessDenied: Access Denied
	status code: 403, request id: MMDRNV2Y90MTNHH3, host id: rERY8wMubiJ+VpU8I7hd+z/8l1BSCwnwe2uePw5pPwtBflXDkL1TiLSDw2D3b3jeKhxAIE8+vpc=
```

I've obfuscated some of the error above with `****`

I'm using the App AWS credentials, so not sure why i'd be getting an access denied on the s3. Thanks for you help!

huanjani:
Hello, @TillaTheHun0!
I'm so sorry about this. 

Have you by chance made any modifications to your `EnvironmentManagerRole`? Would you mind checking if you have the `PutObjectsToArtifactBucket` [policy](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/environment/partials/environment-manager-role.yml#L221) in your CFN template for your environment stack? 

Thanks!

efekarakus:
Hi @TillaTheHun0 !

To add onto Janice's comment, we're currently in the process of trying to reproduce the issue.

In the mean time to get you unblocked, our hypothesis is that it's because the following permission is missing in the `EnvironmentManagerRole`:
https://github.com/aws/copilot-cli/blob/0a2bd5ac31948ceb9e8f7e8bff67480dc1623c47/internal/pkg/template/templates/environment/partials/environment-manager-role.yml#L225
I think if you make a modification to the EnvironmentManagerRole's policy to add the permission above, it should get you unblocked.

### Hypothesis

[...]

PR #3562, opened 2022-05-12
chore: detect if EnvManagerRole has permissions to upload artifacts

efekarakus:
This PR is step 1 in resolving #3556

We first need to detect if the env manager role is allowed to upload to S3. 
If so, then we're good and can use the env session to upload the template and custom resources.  
Otherwise, in step 2 we'll have to call `UpdateEnvironmentTemplate` by adding the permission to the EnvManagerRole.

### DIFF
--- a/internal/pkg/cli/env_upgrade.go
+++ b/internal/pkg/cli/env_upgrade.go
@@ -6,6 +6,7 @@ package cli
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ssm"
@@ -64,6 +65,7 @@ type envUpgradeOpts struct {
 
 	// Constructors for clients that can be initialized only at runtime.
 	// These functions are overridden in tests to provide mocks.
+	newEnvDeployer      func(conf *config.Environment) (environmentDeployer, error)
 	newEnvVersionGetter func(app, env string) (versionGetter, error)
 	newTemplateUpgrader func(conf *config.Environment) (envTemplateUpgrader, error)
 	newS3               func(conf *config.Environment) (uploader, error)
@@ -91,6 +93,13 @@ func newEnvUpgradeOpts(vars envUpgradeVars) (*envUpgradeOpts, error) {
 		uploader: template.New(),
 		appCFN:   cloudformation.New(defaultSession),
 
+		newEnvDeployer: func(conf *config.Environment) (environmentDeployer, error) {
+			sess, err := sessProvider.FromRole(conf.ManagerRoleARN, conf.Region)
+			if err != nil {
+				return nil, fmt.Errorf("create environment stack deployer session from role %s and region %s: %v", conf.ManagerRoleARN, conf.Region, err)
+			}
+			return cloudformation.New(sess), nil
+		},
 		newEnvVersionGetter: func(app, env string) (versionGetter, error) {
 			d, err := describe.NewEnvDescriber(describe.NewEnvDescriberConfig{
 				App:         app,
@@ -174,6 +183,9 @@ func (o *envUpgradeOpts) Execute() error {
 		if err != nil {
 			return fmt.Errorf("get app resources: %w", err)
 		}
+		if err := o.ensureManagerRoleIsAllowedToUpload(env, resources.S3Bucket); err != nil {
+			return err
+		}
 		s3Client, err := o.newS3(env)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


internal/pkg/cli/env_upgrade.go
```diff
@@ -174,6 +183,9 @@ func (o *envUpgradeOpts) Execute() error {
 		if err != nil {
 			return fmt.Errorf("get app resources: %w", err)
 		}
+		if err := o.ensureManagerRoleIsAllowedToUpload(env, resources.S3Bucket); err != nil {
```
iamhopaul123, line 187: this looks like a patch for a specific scenario 😭

internal/pkg/cli/env_upgrade.go
```diff
@@ -212,6 +224,44 @@ func (o *envUpgradeOpts) listEnvsToUpgrade() ([]*config.Environment, error) {
...
 
+func (o *envUpgradeOpts) ensureManagerRoleIsAllowedToUpload(env *config.Environment, bucketName string) error {
+	cfn, err := o.newEnvDeployer(env)
+	if err != nil {
+		return err
+	}
+	body, err := cfn.EnvironmentTemplate(env.App, env.Name)
+	if err != nil {
+		return err
+	}
```
iamhopaul123, line 236: If we are not using `cfn` and `body` in TODO, maybe we could move it into `isManagerRoleAllowedToUpload`?

internal/pkg/cli/env_upgrade.go
```diff
@@ -212,6 +224,44 @@ func (o *envUpgradeOpts) listEnvsToUpgrade() ([]*config.Environment, error) {
...
+	// TODO(efekarakus): if the role isn't allowed to upload, modify the template to add the permission and call cfn.UpdateEnvironmentTemplate
+	return nil
+}
+
+func (o *envUpgradeOpts) isManagerRoleAllowedToUpload(template, bucketName string) bool {
+	wantedPolicyStatementID := "PutObjectsToArtifactBucket"
+	if !strings.Contains(template, wantedPolicyStatementID) {
+		return false
+	}
+	remainingTemplate := template[strings.Index(template, wantedPolicyStatementID):]
```
iamhopaul123: Why do we get the remaining template every time?

internal/pkg/cli/env_upgrade.go
```diff
@@ -212,6 +224,44 @@ func (o *envUpgradeOpts) listEnvsToUpgrade() ([]*config.Environment, error) {
...
+	}
+	if o.isManagerRoleAllowedToUpload(body, bucketName) {
+		return nil
+	}
+	// TODO(efekarakus): if the role isn't allowed to upload, modify the template to add the permission and call cfn.UpdateEnvironmentTemplate
+	return nil
+}
+
+func (o *envUpgradeOpts) isManagerRoleAllowedToUpload(template, bucketName string) bool {
+	wantedPolicyStatementID := "PutObjectsToArtifactBucket"
```
Lou1415926: Do you think we also want to check that the template version is < v1.18.0?

efekarakus: I kept debating if we should just check if the Template version is < 1.9.0 and if so just add the permission, or go this route by checking specifically the values.

I can update it to use the Metadata.Version instead

internal/pkg/cli/env_upgrade.go
efekarakus: We will use it 👍

internal/pkg/cli/env_upgrade.go
efekarakus: I'll update to Wanxian's recommendation here: https://github.com/aws/copilot-cli/pull/3562#discussion_r871705951

internal/pkg/cli/env_upgrade.go
efekarakus: yeah - we couldn't think of a better alternative unfortunately

internal/pkg/cli/env_upgrade.go
Lou1415926: If we don't check the version, false positive could happen if in the we rearrange the policy statements a little, such that `Put` permissions are grouped into another statement and that we delete `PutArtifactsToS3`, or if we simply renamed the statement. I imagine this is not a big deal though.

(On a side note, I think false negative won't happen unless users have manual modifications to the template that happen to break the assumptions that support this implementation so we're good.)

iamhopaul123 (approved):

Lou1415926 (approved):

TillaTheHun0:
I've no part in this solution to #3556, but I have say that you all are awesome!

The team is super responsive when issues/feature requests are filed, and always provide great feedback, workarounds, and advice. When there is a bug, as an outsider, I can clearly follow the progress of its resolution.

Kudos to the Copilot team! Copilot is such a boon for container development on AWS!

PR #3563, opened 2022-05-12
chore: read environment manifest from workspace and unmarshal

Lou1415926:

### DIFF
--- a/internal/pkg/workspace/workspace.go
+++ b/internal/pkg/workspace/workspace.go
@@ -611,7 +611,7 @@ func (ws *Workspace) manifestNameMatchWithDir(mft namedManifest, mftDirName stri
 		return err
 	}
 	if mftName != mftDirName {
-		return fmt.Errorf(`name of the manifest "%s" and directory "%s" do not match`, mftName, mftDirName)
+		return fmt.Errorf("name of the manifest %q and directory %q do not match", mftName, mftDirName)
 	}
 	return nil
 }
--- a/internal/pkg/workspace/workspace.go
+++ b/internal/pkg/workspace/workspace.go
@@ -9,6 +9,9 @@
 //  │   ├── .workspace                 (workspace summary)
 //  │   ├── my-service
 //  │   │   └── manifest.yml           (service manifest)
+//  |   |   environments
+//  |   |   └── test
+//  │   │       └── manifest.yml       (environment manifest for the environment test)
 //  │   ├── buildspec.yml              (legacy buildspec for the pipeline's build stage)
 //  │   ├── pipeline.yml               (legacy pipeline manifest)
 //  │   ├── pipelines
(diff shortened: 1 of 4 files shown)


internal/pkg/workspace/workspace.go
```diff
@@ -279,12 +283,21 @@ func (ws *Workspace) ReadWorkloadManifest(mftDirName string) (WorkloadManifest,
...
+
+// ReadEnvironmentManifest returns the contents of the environment's manifest under copilot/environments/{name}/manifest.yml.
+func (ws *Workspace) ReadEnvironmentManifest(mftDirName string) (WorkloadManifest, error) {
+	raw, err := ws.read(environmentsDirName, mftDirName, manifestFileName)
 	if err != nil {
 		return nil, err
 	}
-	if mftName != mftDirName {
-		return nil, fmt.Errorf(`name of the manifest "%s" and directory "%s" do not match`, mftName, mftDirName)
+	mft := WorkloadManifest(raw)
```
iamhopaul123: This doesn't look like it should belong to here 🤔

Lou1415926: why? `WorkloadManifest` is just a manifest with `name` and `type` fields

iamhopaul123: Because this is for env manifest? I would assume `WorkloadManifest` the method is for workload manifest only. I think though they have the same struct, we probably want to give the public method a different name.

Lou1415926: updated

iamhopaul123 (approved):

internal/pkg/workspace/workspace.go
```diff
@@ -585,25 +605,55 @@ func (ws *Workspace) ListDockerfiles() ([]string, error) {
 	return dockerfiles, nil
 }
 
+func (ws *Workspace) manifestNameMatchWithDir(mft namedManifest, mftDirName string) error {
```
dannyrandall, line 608: i like this, we should do it for workloads if we don't already

internal/pkg/workspace/workspace.go
```diff
@@ -585,25 +605,55 @@ func (ws *Workspace) ListDockerfiles() ([]string, error) {
...
+	mftName, err := mft.name()
+	if err != nil {
+		return err
+	}
+	if mftName != mftDirName {
+		return fmt.Errorf(`name of the manifest "%s" and directory "%s" do not match`, mftName, mftDirName)
```
dannyrandall: `%q`

internal/pkg/workspace/workspace_test.go
```diff
@@ -878,6 +878,181 @@ func TestWorkspace_WriteAddon(t *testing.T) {
 	}
 }
 
+func TestWorkspace_ReadWorkloadManifest(t *testing.T) {
```
dannyrandall, line 881: :heart:

dannyrandall (approved):

internal/pkg/workspace/workspace.go
Lou1415926: we are!

PR #3550, opened 2022-05-10
refactor: LBWS specific ALB -> generic ALB

dannyrandall:
Refactors ALB code/templates currently specific to LBWS to be generic so it's easier to reuse in a backend service's internal alb. The main changes are:
- Rename parameter constants to be `Workload` constants instead of LBWS
- Pull ALB CF from LBWS CF into its own partial
- Move `manifest.RoutingRuleConfiguration` into it's own file from the LBWS file

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the Apache 2.0 License.

PR #3565, opened 2022-05-12
chore: backend service http support

dannyrandall:
HTTPS support to come in a future PR.

The `alb.yml` custom resource role changes are because currently each service type has their own custom resource rule, but I pulled it out of LBWS and put it in `alb.yml` in #3550. The generated role should stay the same, just generated in a different file.


internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/complex-http-manifest.yml
```diff
@@ -0,0 +1,28 @@
+name: complex-http
```
iamhopaul123: Just to double check: we don't need to update the test file, and adding test data are ok right? Also can we give them a more concrete name like `with-alb` instead of `complex`?

internal/pkg/template/templates/workloads/services/backend/cf.yml
```diff
@@ -74,7 +97,20 @@ Resources:
             Action:
               - sts:AssumeRole
       Path: /
+      ManagedPolicyArns:
+        - !Sub arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
       Policies:
+      {{- if .ALBEnabled}}
+        - PolicyName: "DNSandACMAccess"
```
iamhopaul123: Should we give it a better name?

internal/pkg/template/templates/workloads/services/backend/cf.yml
```diff
@@ -102,18 +138,31 @@ Resources:
...
+      - EnvControllerAction
+      {{- if and .ALBEnabled .HTTPSListener}}
+      - HTTPListenerRuleWithDomain
+      - HTTPSListenerRule
+      {{ else if .ALBEnabled}}
+      - HTTPListenerRule
+      {{- end}}
     Properties:
-{{include "service-base-properties" . | indent 6}}
+      {{- "\n"}}{{ include "service-base-properties" . | indent 6 }}
```
iamhopaul123, line 157: Why do we need "\n" 🤔

dannyrandall: good question: [answer](https://stackoverflow.com/questions/46066524/how-can-i-place-go-template-actions-with-multi-line-output-at-any-indentation-le#answer-46066525)
so that i can indent it and have the template fold correctly in my editor 🙄

internal/pkg/template/templates/workloads/services/backend/cf.yml
dannyrandall: yeah, it's not super representative - it's the same policy name we give in LBWS, so let's leave it to be consistent? long term, we should probably get rid of the custom resource role and just have a single role per lambda

internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/complex-http-manifest.yml
dannyrandall: not 100% sure what you're asking - these are brand new test files, not refactored/moved - is that what you were wondering?

either way, i'll give them better names ✔️

internal/pkg/template/templates/workloads/services/backend/cf.yml
iamhopaul123: It's just we don't have ACM related actions here :think

internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/complex-http-manifest.yml
iamhopaul123: Yes but where's the integ test file that reads those test data? Something like this https://github.com/aws/copilot-cli/blob/8eae6a3b6b2ea0677db64874f682133e0303f7ab/internal/pkg/deploy/cloudformation/stack/lb_web_service_integration_test.go

internal/pkg/template/templates/workloads/services/backend/cf.yml
dannyrandall: yeah....................................................good point haha. i'll change it for the backend service for now

internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/complex-http-manifest.yml
dannyrandall: [`TestBackendService_TemplateAndParamsGeneration`](https://github.com/aws/copilot-cli/pull/3565/files#diff-b5a4ffcc3517c2f2e286fb91ac6988e025a2b3ba7ce13238c6f376c0cd7351a8R509) uses the files

internal/pkg/template/templates/workloads/services/backend/cf.yml
dannyrandall: lmk if you don't like the name :)

iamhopaul123 (approved):

internal/pkg/template/templates/workloads/services/backend/cf.yml
huanjani: can we use `nindent`? That seems nifty!

dannyrandall: we'd have to import https://github.com/Masterminds/sprig 😔

huanjani: 😤

huanjani (approved):
:shipit:

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
```diff
@@ -169,7 +169,64 @@ Resources:
 
 {{- if .ALBEnabled}}
 {{include "alb" . | indent 2}}
+
+  CustomResourceRole:
```
efekarakus, line 173: At a glance the `CustomResourceRole` for lbws and backend service seem to be the same, am I missing something or can we potentially move the role to a partial?

internal/pkg/template/templates/workloads/services/backend/cf.yml
```diff
@@ -74,7 +97,20 @@ Resources:
             Action:
               - sts:AssumeRole
       Path: /
+      ManagedPolicyArns:
+        - !Sub arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
       Policies:
+      {{- if .ALBEnabled}}
+        - PolicyName: "RulePriorityGeneratorAccess"
```
efekarakus, line 104: The `HTTPRulePriorityAction` that's included for backend services I think have the following `ListenerARN`:

https://github.com/aws/copilot-cli/blob/9a479c48371e1f1ae339270219c4c1eb49dbe43d/internal/pkg/template/templates/workloads/partials/cf/http-listener.yml#L7
and

https://github.com/aws/copilot-cli/blob/9a479c48371e1f1ae339270219c4c1eb49dbe43d/internal/pkg/template/templates/workloads/partials/cf/http-listener.yml#L36

We'll want to update these listener ARNs to be for the InternalALB on top of Janice's changes

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
dannyrandall: Previously, the `CustomResourceRole` was in the `alb` partial, but the issue with it there is that the backend service needs it if autoscaling is enabled, regardless of if it needs an ALB or not. I considered putting it in it's own partial, but think the best long term solution (since the worker service uses it as well) is to refactor out the `CustomResourceRole` everywhere and create a lambda-specific role instead. 

I've started that work [here](https://github.com/aws/copilot-cli/compare/mainline...dannyrandall:fix/nlb-autoscaling), but want to finish the internal ALB stuff before merging that.

Thoughts? You're absolutely right that the role is the exact same, so it's something that could be solved sooner by putting it in it's own partial.

dannyrandall: resolved in https://github.com/aws/copilot-cli/pull/3578

internal/pkg/template/templates/workloads/services/backend/cf.yml
dannyrandall: resolved in https://github.com/aws/copilot-cli/pull/3608

PR #3567, opened 2022-05-12
fix: ensure EnvManagerRole template has permissions to upload artifacts

efekarakus:
Related #3556 

Tested `env upgrade` from the following versions:
- [x] `copilot v1.0` 
- [x] `copilot v1.13`
- [x] `copilot v1.17`  

### DIFF
--- a/internal/pkg/cli/env_upgrade.go
+++ b/internal/pkg/cli/env_upgrade.go
@@ -319,14 +319,12 @@ func (o *envUpgradeOpts) grantManagerRolePermissionToUpload(cfn environmentDeplo
 	var errEmptyChangeSet *awscfn.ErrChangeSetEmpty
 	o.prog.Start("Update the environment's manager role with permission to upload artifacts to S3")
 	err := cfn.UpdateEnvironmentTemplate(app, env, updatedBody, execRole)
-	switch {
-	case err != nil && !errors.As(err, &errEmptyChangeSet):
+	if err != nil && !errors.As(err, &errEmptyChangeSet) {
 		o.prog.Stop(log.Serrorln("Unable to update the environment's manager role with upload artifacts permission"))
 		return fmt.Errorf("update environment template with PutObject permissions: %v", err)
-	default:
-		o.prog.Stop(log.Ssuccessln("Updated the environment's manager role with permissions to upload artifacts to S3"))
-		return nil
 	}
+	o.prog.Stop(log.Ssuccessln("Updated the environment's manager role with permissions to upload artifacts to S3"))
+	return nil
 }
 
 func (o *envUpgradeOpts) upgrade(env *config.Environment, app *config.Application,
--- a/internal/pkg/cli/env_upgrade.go
+++ b/internal/pkg/cli/env_upgrade.go
@@ -6,6 +6,7 @@ package cli
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"gopkg.in/yaml.v3"
 
@@ -15,6 +16,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 
+	awscfn "github.com/aws/copilot-cli/internal/pkg/aws/cloudformation"
 	"github.com/aws/copilot-cli/internal/pkg/aws/s3"
 	"github.com/aws/copilot-cli/internal/pkg/aws/sessions"
 	"github.com/aws/copilot-cli/internal/pkg/config"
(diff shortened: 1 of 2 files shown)


internal/pkg/cli/env_upgrade.go
```diff
@@ -266,6 +266,69 @@ func (o *envUpgradeOpts) isManagerRoleAllowedToUpload(body string) (bool, error)
...
+						} `yaml:"PolicyDocument"`
+					} `yaml:"Policies"`
+				} `yaml:"Properties"`
+			} `yaml:"EnvironmentManagerRole"`
+		} `yaml:"Resources"`
+	}
+
+	var tpl Template
+	if err := yaml.Unmarshal([]byte(body), &tpl); err != nil {
+		return fmt.Errorf("unmarshal environment template to find EnvironmentManagerRole policy statement: %v", err)
```
dannyrandall, line 288:
```suggestion
		return fmt.Errorf("unmarshal environment template to find EnvironmentManagerRole policy statement: %w", err)
```

internal/pkg/cli/env_upgrade_test.go
```diff
@@ -201,14 +204,15 @@ func TestEnvUpgradeOpts_Execute(t *testing.T) {
 		given     func(ctrl *gomock.Controller) *envUpgradeOpts
 		wantedErr error
 	}{
-		"should return an error if the environment template version cannot upload artifacts": {
+		"should return an error if the environment template cannot be updated to grand upload artifacts permissions": {
```
dannyrandall:
```suggestion
		"should return an error if the environment template cannot be updated to grant upload artifacts permissions": {
```

internal/pkg/cli/env_upgrade.go
```diff
@@ -266,6 +266,69 @@ func (o *envUpgradeOpts) isManagerRoleAllowedToUpload(body string) (bool, error)
...
+	if len(tpl.Resources.ManagerRole.Properties.Policies) == 0 {
+		return errors.New("unable to find policies for the EnvironmentManagerRole")
+	}
+	// lines and columns are 1-indexed, so we have to substract one from each.
+	statementLineIndex := tpl.Resources.ManagerRole.Properties.Policies[0].Document.Statements.Line - 1
+	numSpaces := tpl.Resources.ManagerRole.Properties.Policies[0].Document.Statements.Column - 1
+	pad := strings.Repeat(" ", numSpaces)
+
+	// Create the additional permissions needed with the appropriate indentation.
+	permissions := fmt.Sprintf(`- Sid: PatchPutObjectsToArtifactBucket
```
Lou1415926, line 299: Noice! Naming it `Patch...` ensures that we can upgrade environment using the env template without fear for "change set has no changes" error.

dannyrandall (approved):

internal/pkg/cli/env_upgrade.go
```diff
@@ -266,6 +266,69 @@ func (o *envUpgradeOpts) isManagerRoleAllowedToUpload(body string) (bool, error)
...
+	o.prog.Start("Update the environment's manager role with permission to upload artifacts to S3")
+	err := cfn.UpdateEnvironmentTemplate(app, env, updatedBody, execRole)
+	switch {
+	case err != nil && !errors.As(err, &errEmptyChangeSet):
+		o.prog.Stop(log.Serrorln("Unable to update the environment's manager role with upload artifacts permission"))
+		return fmt.Errorf("update environment template with PutObject permissions: %v", err)
+	default:
+		o.prog.Stop(log.Ssuccessln("Updated the environment's manager role with permissions to upload artifacts to S3"))
+		return nil
+	}
```
dannyrandall: i'd prob just `if` here since there's only one case

Lou1415926 (approved):
Put on the label so that you can address feedback!

internal/pkg/cli/env_upgrade_test.go
efekarakus: Done!

internal/pkg/cli/env_upgrade.go
efekarakus: Done! totally makes sense, I had a bunch of things before :rip:

internal/pkg/cli/env_upgrade.go
efekarakus: Kept it as is, it's not really important 🤷  but I did that on purpose since the `cli` package is the root pkg so we don't need to surface the error part of the api

internal/pkg/cli/env_upgrade.go
```diff
@@ -266,6 +266,67 @@ func (o *envUpgradeOpts) isManagerRoleAllowedToUpload(body string) (bool, error)
...
+
+	var tpl Template
+	if err := yaml.Unmarshal([]byte(body), &tpl); err != nil {
+		return fmt.Errorf("unmarshal environment template to find EnvironmentManagerRole policy statement: %v", err)
+	}
+	if len(tpl.Resources.ManagerRole.Properties.Policies) == 0 {
+		return errors.New("unable to find policies for the EnvironmentManagerRole")
+	}
+	// lines and columns are 1-indexed, so we have to substract one from each.
```
dannyrandall, line 293: > substract

i finally found one 😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇😇

Issue #1851, opened 2021-01-20
environment  --prod flag usage 


vbarba:
According the command-line help there is a flag
```
--prod                           If the environment contains production services.
```
And I am wondering what this flag actually does. I didn't find anything in the doc

Thanks

efekarakus:
Hi @vbarba !

It's not really used for anything at the moment :D Before v1.0, the idea was that the flag could provide a way for customers to decide to keep certain resources if a prod env is deleted. But that can be achieved already right now within addons, using [`Mappings`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/mappings-section-structure.html) and the `Env` parameter.

In the future, it could provide us a way of listing all your production resources but there is no plan at the moment. It's safe to ignore it :)

vbarba:
ok! thanks!

PR #3569, opened 2022-05-13
chore(cli): implement Execute for `env deploy`

Lou1415926:


dannyrandall (approved):
lgtm☺️

internal/pkg/cli/deploy/env.go
```diff
@@ -67,8 +59,16 @@ func (d *envDeployer) UploadArtifacts() (map[string]string, error) {
 	return urls, nil
 }
 
+// DeployEnvironmentInput contains information used to deploy the environment.
+type DeployEnvironmentInput struct {
+	RootUserARN         string
+	IsProduction        bool
```
huanjani, line 65: Are we keeping this in case users have existing envs that they inited with the now-deprecated `--prod` flag?

Lou1415926: The flag was deprecated in `cli` package but all these wiring behind `cli` wasn't removed (for example, `deploy.CreateEnvironmentInput` still has the field `Prod`). Therefore I thought we wanted to keep the wirings there so that we can revive the flag in the future when we want. This is to keep the existing wirings. 

However, I am not sure whether we simply forgot to remove them or not. Do you think I should remove them altogether?

huanjani: I _think_ we've moved away from a prescribed `prod` setting in favor of giving users a more fine-tuned config-setting experience, but I can't speak for the team!

huanjani: Oh, [this](https://github.com/aws/copilot-cli/issues/1851) has an idea for that flag for the future.

huanjani (approved):
LGTM

PR #3566, opened 2022-05-12
feat: enable `exec` functionality with Windows platform

huanjani:
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the Apache 2.0 License.

PR #3573, opened 2022-05-16
chore(e2e): restore `EnableExec` field for tasks

huanjani:


iamhopaul123 (approved):

iamhopaul123:
Maybe use chore since it is not a bug fix?

Lou1415926:
quick question: why is it "restore"? Was `EnableExec: true` there before?  I asked because I'd be curious why it was removed if it was originally there... Just in case there were some decisions/concerns around it

huanjani:
> quick question: why is it "restore"? Was `EnableExec: true` there before? I asked because I'd be curious why it was removed if it was originally there... Just in case there were some decisions/concerns around it

I mistakenly took it out just recently in #3566.

Lou1415926 (approved):

PR #3575, opened 2022-05-16
refactor: minor refactor to the selector package

Lou1415926:


internal/pkg/term/selector/selector.go
```diff
@@ -89,138 +89,134 @@ var presetSchedules = []prompt.Option{
...
 
-// WsWorkloadLister wraps the method to get workloads in current workspace.
-type WsWorkloadLister interface {
+type wsWorkloadLister interface {
 	ListServices() ([]string, error)
 	ListJobs() ([]string, error)
 	ListWorkloads() ([]string, error)
 }
 
-// WsPipelinesLister is a pipeline lister.
```
iamhopaul123, line 127: Why do we remove comments for some interfaces but not all of them?

internal/pkg/term/selector/selector.go
```diff
@@ -89,138 +89,134 @@ var presetSchedules = []prompt.Option{
...
 	ListActiveAppEnvTasks(opts ecs.ListActiveAppEnvTasksOpts) ([]*awsecs.Task, error)
 	ListActiveDefaultClusterTasks(filter ecs.ListTasksFilter) ([]*awsecs.Task, error)
 }
 
-// Select prompts users to select the name of an application or environment.
-type Select struct {
-	prompt Prompter
-	config ConfigLister
+// AppEnvSelect prompts users to select the name of an application or environment.
+type AppEnvSelect struct {
```
iamhopaul123: nit i feel `selector` is a better name as a noun

iamhopaul123 (approved):

internal/pkg/term/selector/selector.go
Lou1415926: Some interfaces's purpose I think is a little ambiguous, whereas others are clear judging from the signature. I kept the comments for the first group of interfaces.

dannyrandall (approved):

PR #3577, opened 2022-05-16
chore(cli): implement `Validate` and `Ask` for `env deploy`

Lou1415926:


internal/pkg/cli/env_deploy.go
```diff
@@ -85,13 +107,41 @@ func (o *deployEnvOpts) environmentManifest() (*manifest.Environment, error) {
...
+	o.name = name
+	return nil
+}
+
 func (o *deployEnvOpts) cachedTargetEnv() (*config.Environment, error) {
 	if o.targetEnv == nil {
 		env, err := o.store.GetEnvironment(o.appName, o.name)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("get environment %s: %w", o.name, err)
```
efekarakus: nit: can we include the app name here as well

internal/pkg/term/selector/selector.go
```diff
@@ -701,6 +706,51 @@ func (s *WorkspaceSelector) Workload(msg, help string) (wl string, err error) {
 	return selectedWlName, nil
 }
 
+// WSEnvironment fetches all environments belong to the app in the workspace and prompts the user to select one.
+func (s *WorkspaceSelector) WSEnvironment(msg, help string) (wl string, err error) {
```
efekarakus: Can we rename this to the following? since the method is for a `WorkspaceSelector` it feels a bit redundant.
```suggestion
func (s *WorkspaceSelector) Environment(msg, help string) (string, error) {
```

internal/pkg/workspace/workspace.go
```diff
@@ -276,6 +276,31 @@ func (ws *Workspace) listWorkloads(match func(string) bool) ([]string, error) {
...
 }
 
+// ListEnvironments returns the name of the environments in the workspace.
+func (ws *Workspace) ListEnvironments() ([]string, error) {
+	copilotPath, err := ws.copilotDirPath()
+	if err != nil {
+		return nil, err
+	}
+	envPath := filepath.Join(copilotPath, environmentsDirName)
+	files, err := ws.fsUtils.ReadDir(envPath)
```
efekarakus: nit: can we rename `fsUtils` to `fs`? 🥺

efekarakus (commented):
Looks great!

internal/pkg/cli/env_deploy.go
```diff
@@ -37,6 +42,24 @@ type deployEnvOpts struct {
 	unmarshalManifest func(in []byte) (*manifest.Environment, error)
 }
 
+// Validate returns an error for any invalid optional flags.
+func (o *deployEnvOpts) Validate() error {
+	return nil
+}
```
iamhopaul123, line 48: If there's no validation needed for this command maybe
```golang
// Validate is a no-op for this command.
```

internal/pkg/cli/env_deploy.go
```diff
@@ -37,6 +42,24 @@ type deployEnvOpts struct {
...
+
+// Ask prompts for and validates any required flags.
+func (o *deployEnvOpts) Ask() error {
+	if o.appName != "" {
```
iamhopaul123:
```suggestion
	if o.appName == "" {
```
to simplify the `if else`

internal/pkg/term/selector/selector.go
Lou1415926: This is because `WorkspaceSelector` contains `ConfigSelector` which contains `AppEnvSelector` which has a method named `Environment`. Renaming `WSEnv` to `Env` overshadows the `Env` method of `AppEnvSelector`, and will break some other `cli` commands which uses `WorkspaceSelector` to invoke `Environment` method "inherited" from the `AppEnvSelector`... I wanted to refactor it but it'd remove the method "inheritance" between the bunch of selectors.

internal/pkg/cli/env_deploy.go
Lou1415926: This was intended to re-enforce the "validateOrAsk" mindset.

internal/pkg/term/selector/selector.go
efekarakus: ☹️ dang, we shouldn't have embedded the *ConfigSelector to the WorkspaceSelector.

Well - maybe this is a scenario where the name of the type is poor?
What if we had a refactor where we rename `WorkspaceSelector -> LocalWorkloadsSelector` and then have a separate type that's `WorkspaceSelector` that's limited to strictly choosing stuff from a workspace

Lou1415926: Done!

efekarakus (approved):

iamhopaul123 (approved):

PR #3586, opened 2022-05-19
chore(manifest): add fn to transform config.Env to manifest.Env

efekarakus:


internal/pkg/config/env.go
```diff
@@ -40,7 +40,10 @@ type CustomizeEnv struct {
 }
 
 // IsEmpty returns if CustomizeEnv is an empty struct.
-func (c CustomizeEnv) IsEmpty() bool {
+func (c *CustomizeEnv) IsEmpty() bool {
+	if c == nil {
+		return true
+	}
```
huanjani, line 46: nit: can't we do this?
```suggestion
func (c *CustomizeEnv) IsEmpty() bool {
	return c == nil
```

efekarakus, line 46: not quite, because we want this method to still return `true` when the following happens:
```go
env := &CustomizeEnv{} // not nil
require.True(env.IsEmpty())
```

internal/pkg/config/env.go
```diff
@@ -40,7 +40,10 @@ type CustomizeEnv struct {
 }
 
 // IsEmpty returns if CustomizeEnv is an empty struct.
-func (c CustomizeEnv) IsEmpty() bool {
+func (c *CustomizeEnv) IsEmpty() bool {
```
huanjani, line 43: Ooh, is this a chance to use `IsZero`?

huanjani (approved):
LGTM!

internal/pkg/config/env.go
efekarakus: that's a good idea! Maybe we can do it as a follow-up PR? I didn't add any new `IsEmpty` methods so I didn't refactor for `IsZero`

internal/pkg/manifest/env.go
```diff
@@ -41,6 +69,45 @@ type environmentVPCConfig struct {
...
+		cfg.loadImportedVPCConfig(imported)
+	}
+}
+
+func (cfg *environmentVPCConfig) loadAdjustedVPCConfig(vpc *config.AdjustVPC) {
+	cfg.CIDR = ipNetP(vpc.CIDR)
+	cfg.Subnets.Public = make([]subnetConfiguration, len(vpc.PublicSubnetCIDRs))
+	cfg.Subnets.Private = make([]subnetConfiguration, len(vpc.PrivateSubnetCIDRs))
+	hasCustomAZNames := len(vpc.AZs) > 0
+	for i := range vpc.PublicSubnetCIDRs {
```
dannyrandall: will there always be the same number of public and private subnets?

internal/pkg/manifest/env.go
```diff
@@ -41,6 +69,45 @@ type environmentVPCConfig struct {
 	Subnets subnetsConfiguration `yaml:"subnets,omitempty"`
 }
 
+func (cfg *environmentVPCConfig) loadVPCConfig(vpc *config.CustomizeEnv) {
+	if vpc.IsEmpty() {
+		return
+	}
+	if adjusted := vpc.VPCConfig; adjusted != nil {
+		cfg.loadAdjustedVPCConfig(adjusted)
```
paragbhingre, line 77: What is AdjustedVPCConfig?

internal/pkg/manifest/env.go
```diff
@@ -41,6 +69,45 @@ type environmentVPCConfig struct {
...
+	}
+}
+
+func (cfg *environmentVPCConfig) loadImportedVPCConfig(vpc *config.ImportVPC) {
+	cfg.ID = stringP(vpc.ID)
+	cfg.Subnets.Public = make([]subnetConfiguration, len(vpc.PublicSubnetIDs))
+	for i, subnet := range vpc.PublicSubnetIDs {
+		cfg.Subnets.Public[i].SubnetID = stringP(subnet)
+	}
+	cfg.Subnets.Private = make([]subnetConfiguration, len(vpc.PublicSubnetIDs))
```
paragbhingre: `len(vpc.PublicSubnetIDs)` should be `len(vpc.privateSubnetIDs)`?

Danny's question will resolve this question too.

internal/pkg/manifest/env.go
paragbhingre: +1

internal/pkg/manifest/env.go
efekarakus: There are two ways of customizing an environment today with `copilot env init`: 
https://github.com/aws/copilot-cli/blob/c94e2b4e83e663a2a322e18e6bf9bd119b9566d6/internal/pkg/config/env.go#L36-L38

You can either "import" a VPC ID, and subnet IDs that already exists. Or you can let copilot create a new VPC and "adjust" its settings like the CIDR ranges, AZs.

Hope this helps!

internal/pkg/manifest/env.go
efekarakus: That's the case today: 
https://github.com/aws/copilot-cli/blob/c94e2b4e83e663a2a322e18e6bf9bd119b9566d6/internal/pkg/cli/validate.go#L815

The user first chooses the number of AZs, and then they have to specify the CIDR ranges for each subnet type in the AZs. 

But I agree that it'd be safer to write the method without this assumption. I updated the flow **with a more flexible assumption** that the number of AZs selected >= number of CIDR ranges specified. I think that's safer.

internal/pkg/manifest/env.go
```diff
@@ -41,6 +69,49 @@ type environmentVPCConfig struct {
...
+}
+
+func (cfg *environmentVPCConfig) loadAdjustedVPCConfig(vpc *config.AdjustVPC) {
+	cfg.CIDR = ipNetP(vpc.CIDR)
+	hasCustomAZNames := len(vpc.AZs) > 0
+	cfg.Subnets.Public = make([]subnetConfiguration, len(vpc.PublicSubnetCIDRs))
+	cfg.Subnets.Private = make([]subnetConfiguration, len(vpc.PrivateSubnetCIDRs))
+	for i, cidr := range vpc.PublicSubnetCIDRs {
+		cfg.Subnets.Public[i].CIDR = ipNetP(cidr)
+		if hasCustomAZNames {
```
dannyrandall: something like...
```suggestion
		if hasCustomAZNames && len(vpc.AZs) > i {
```
? just so it doesn't panic?

dannyrandall (approved):

internal/pkg/manifest/env.go
```diff
@@ -127,14 +198,28 @@ type environmentObservability struct {
...
 
 type environmentHTTPConfig struct {
 	Public publicHTTPConfig `yaml:"public,omitempty"`
 }
 
+func (cfg *environmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
+	if env.IsEmpty() {
+		return
+	}
+	cfg.Public.Certificates = env.ImportCertARNs
```
iamhopaul123, line 219: What about the internal ALB certs? Will it have a different name in SSM?

internal/pkg/manifest/env.go
```diff
@@ -127,14 +198,28 @@ type environmentObservability struct {
 }
 
 // IsEmpty returns true if there is no configuration to the environment's observability.
-func (o environmentObservability) IsEmpty() bool {
-	return o.ContainerInsights == nil
+func (o *environmentObservability) IsEmpty() bool {
+	return o == nil || o.ContainerInsights == nil
+}
+
+func (o *environmentObservability) loadConfig(tele *config.Telemetry) {
```
iamhopaul123:
```suggestion
func (o *environmentObservability) loadObsConfig(tele *config.Telemetry) {
```

internal/pkg/manifest/http.go
```diff
@@ -83,6 +83,14 @@ func (r *RoutingRuleConfiguration) isEmpty() bool {
 // IPNet represents an IP network string. For example: 10.1.0.0/16
 type IPNet string
 
+func ipNetP(s string) *IPNet {
```
iamhopaul123, line 86: :D

internal/pkg/manifest/doc.go
```diff
@@ -0,0 +1,5 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package manifest provides functionality to create Manifest files.
+package manifest
```
iamhopaul123, line 5: This is slick! We should populate this to the other pkgs.

internal/pkg/manifest/env.go
efekarakus: done, replaced with `if len(vpc.AZs) > i` - we shouldn't need `hasCustomAZNames` then

tbh I think we're pretty safe 😛 it's good that we don't panic, and if there is a mistake in the generated manifest then `Validate` will catch it before deploying:
https://github.com/aws/copilot-cli/blob/e92a4bcd20856e4836a59337cc5b6bed1d214e46/internal/pkg/manifest/validate_env.go#L86-L127

internal/pkg/manifest/env.go
efekarakus: done!

internal/pkg/manifest/env.go
efekarakus: Yeah, once @huanjani adds it to SSM, we will have to update this method.

internal/pkg/manifest/doc.go
efekarakus: I was thinking we can do it everytime we touch a different package, instead of sending a massive refactor PR but I'm fine either way!

iamhopaul123: oh no worries. I was thinking the same thing

efekarakus:
@mergifyio refresh

internal/pkg/manifest/env.go
```diff
@@ -51,60 +121,60 @@ func UnmarshalEnvironment(in []byte) (*Environment, error) {
...
 	// Either the VPC isn't configured, or everything need to be explicitly configured.
-	if !v.managedVPCCustomized() {
+	if !cfg.managedVPCCustomized() {
 		return nil
 	}
-	publicSubnetCIDRs := make([]string, len(v.Subnets.Public))
-	privateSubnetCIDRs := make([]string, len(v.Subnets.Public))
-	azs := make([]string, len(v.Subnets.Public))
+	publicSubnetCIDRs := make([]string, len(cfg.Subnets.Public))
+	privateSubnetCIDRs := make([]string, len(cfg.Subnets.Public))
```
paragbhingre, line 159:
```suggestion
	privateSubnetCIDRs := make([]string, len(cfg.Subnets.Private))
```

paragbhingre: Oops I'm little late

PR #3587, opened 2022-05-20
chore(manifest): make `placement` a stringOrMap

iamhopaul123:


huanjani (approved):
LGTM!

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -580,10 +580,15 @@ func convertNetworkConfig(network manifest.NetworkConfig) template.NetworkOpts {
...
 		SecurityGroups: network.VPC.SecurityGroups,
 	}
-	if network.VPC.Placement == nil {
+	placement := network.VPC.Placement
+	if placement.IsEmpty() {
 		return opts
 	}
-	if *network.VPC.Placement != manifest.PublicSubnetPlacement {
+	if placement.PlacementString == nil {
+		opts.SubnetIDs = placement.PlacementArgs.Subnets
```
efekarakus, line 589: do we want to set `opts.AssignPublicIP = template.DisablePublicIP` here? Otherwise it will be set to true

internal/pkg/manifest/rd_web_svc.go
```diff
@@ -60,15 +61,48 @@ func (c *RequestDrivenWebServiceNetworkConfig) IsEmpty() bool {
...
 
-// RequestDrivenWebServicePlacement represents where to place tasks for a Request-Driven Web Service.
-type RequestDrivenWebServicePlacement Placement
+// RequestDrivenWebServicePlacement represents what types of subnets to place tasks.
+type RequestDrivenWebServicePlacementString PlacementString
+
+// RequestDrivenWebServicePlacementArgOrString represents where to place tasks for a Request-Driven Web Service.
+type RequestDrivenWebServicePlacementArgOrString struct {
+	PlacementString *RequestDrivenWebServicePlacementString
+	PlacementArgs
```
efekarakus: Does this mean you will allow folks to specify the subnet IDs for the VpcConnector resource as well?

internal/pkg/manifest/rd_web_svc.go
```diff
@@ -60,15 +61,48 @@ func (c *RequestDrivenWebServiceNetworkConfig) IsEmpty() bool {
 	return c.VPC.isEmpty()
 }
 
-// RequestDrivenWebServicePlacement represents where to place tasks for a Request-Driven Web Service.
-type RequestDrivenWebServicePlacement Placement
+// RequestDrivenWebServicePlacement represents what types of subnets to place tasks.
+type RequestDrivenWebServicePlacementString PlacementString
```
efekarakus: This is a bit of tangent, but an opportunity to simplify:
I recently learned from Amit on the App Runner team that `placement: public` is actually supported by App Runner and doesn't require a nat gateway in that scenario.

So my point is we can actually completely get rid of `RequestDrivenWebServicePlacementString` and use `PlacementArgOrString` 😍

internal/pkg/deploy/cloudformation/stack/transformers.go
iamhopaul123: Good idea. I'll have a follow-up PR adding the CFN template changes and an integ test

internal/pkg/manifest/rd_web_svc.go
iamhopaul123: Yeah that's right.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -595,10 +601,15 @@ func convertRDWSNetworkConfig(network manifest.RequestDrivenWebServiceNetworkCon
...
+	placement := network.VPC.Placement
+	if placement.IsEmpty() {
+		return opts
+	}
+	if placement.PlacementString == nil {
+		opts.SubnetIDs = placement.PlacementArgs.Subnets
 		return opts
 	}
-	if string(*network.VPC.Placement) == string(manifest.PrivateSubnetPlacement) {
+	if string(*placement.PlacementString) != string(manifest.PublicSubnetPlacement) {
```
efekarakus: nit: I think we don't need the `string()` type conversion anymore

internal/pkg/template/workload.go
```diff
@@ -420,6 +420,7 @@ type DeadLetterQueue struct {
 type NetworkOpts struct {
 	AssignPublicIP string
 	SubnetsType    string
+	SubnetIDs      []string
```
efekarakus: nit: can you add a comment here mentionining `SubnetsType` and `SubnetIDs` are mutually exclusive

efekarakus (approved):

Issue #3538, opened 2022-05-05
[Design] enable internal ALBs for Backend Services


huanjani:
**Add support for a shared internal load balancer with Backend Services**

Allow users to add the [http](https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#http) field, currently limited to internet-facing Load Balanced Web Services, to Backend Services' manifests. The deployment of a Backend Service with `http` specified will trigger the creation of an [internal load balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-create-internal-load-balancer.html) in the [environment](https://aws.github.io/copilot-cli/docs/concepts/environments/) to which the service is deployed. The internal load balancer may be shared by all Backend Services in the environment.
```yaml
name: api
type: Backend Service

http:
  path: '/'
...
```

For users who want to have an https endpoint, they can leverage the new [import existing certificates](https://github.com/aws/copilot-cli/pull/3503) feature. If certs are imported into an environment with a VPC with _only_ private subnets, they will be automatically applied to the _internal_ load balancer.

```
$ copilot env init --import-vpc-id vpc-099x12d2b34abcd47 \
--import-private-subnets subnet-099fafef48fb3c547,subnet-00c9e76f123456e7f \
--import-cert-arns arn:aws:acm:us-east-1:123456789012:certificate/12345678-1234-1234-1234-123456789012 
```
Users can then bring their own hosted zones so that once the load balancer is created, Copilot will insert the alias record into the hosted zone, mapping the alias name to the LB's DNS name.
```yaml
http:
    path: '/'
    alias: example.com # will work only for environments that have imported certificates 
    hosted_zone: Z0873220N255IR3MTNR4 # alias requires a `hosted_zone`
```

When multiple services and aliases are in play, users will need to 1. bring a hosted zone and  provide a _unique_ alias for each service they create or 2. not bring a hosted zone and add records post-release, independently of Copilot.

Related: #2961, #3344.

huanjani:
Q for those interested in this feature: 
Would you want
1. all resources within the VPC to have access to the internal load balancer, or
2. only Copilot services to have access?
Thanks!

rverpillot:
We need internal ALB, but we will have to use a VPC with "public" subnets routed on our on-prem network and "private" subnets for the application only. So the ALB should be deployed on the "public" subnets to be reachable from the on-prem network, whereas ECS services will be deployed on private subnets. 

[...]

PR #3590, opened 2022-05-20
chore: add InternalHTTPListener to env template

huanjani:
Updating #3531 with

- changes resulting from [discussion ](https://github.com/aws/copilot-cli/issues/3538)
- additions for `https` 
- necessary outputs
- minor reordering for organization

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/env_test.go
+++ b/internal/pkg/deploy/cloudformation/stack/env_test.go
@@ -34,7 +34,7 @@ func TestEnv_Template(t *testing.T) {
 					DNSDelegationLambda:    "mockkey2",
 					CustomDomainLambda:     "mockkey4",
 					VPCConfig: template.VPCConfig{
-						Imported: nil,
+						Imported: &template.ImportVPC{},
 						Managed: template.ManagedVPC{
 							CIDR:               DefaultVPCCIDR,
 							PrivateSubnetCIDRs: DefaultPrivateSubnetCIDRs,
--- a/internal/pkg/deploy/cloudformation/stack/env_test.go
+++ b/internal/pkg/deploy/cloudformation/stack/env_test.go
@@ -77,6 +77,8 @@ func TestEnv_Parameters(t *testing.T) {
 	deploymentInput := mockDeployEnvironmentInput()
 	deploymentInputWithDNS := mockDeployEnvironmentInput()
 	deploymentInputWithDNS.App.Domain = "ecs.aws"
+	deploymentInputWithPrivateDNS := mockDeployEnvironmentInput()
+	deploymentInputWithPrivateDNS.ImportCertARNs = []string{"arn:aws:acm:us-east-1:123456789012:certificate/12345678-1234-1234-1234-123456789012"}
 	testCases := map[string]struct {
 		input *deploy.CreateEnvironmentInput
 		want  []*cloudformation.Parameter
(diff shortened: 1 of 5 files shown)


internal/pkg/template/templates/environment/cf.yml
```diff
@@ -284,25 +290,54 @@ Resources:
...
+      'aws:copilot:description': 'An internal load balancer listener to route HTTPS traffic'
+    Type: AWS::ElasticLoadBalancingV2::Listener
+    Condition: ExportInternalHTTPSListener
+    Properties:
+      Certificates:
+{{- if .ImportCertARNs}}
+        - CertificateArn: {{index .ImportCertARNs 0}}
+{{- else}}
+        - CertificateArn: !Ref HTTPSCert
+{{- end}}
```
efekarakus: We don't want to use our HTTPSCert for internal load balancers, the certificate has to be private, so no need for the else statement

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -284,25 +290,54 @@ Resources:
...
+      DefaultActions:
+        - TargetGroupArn: !Ref DefaultHTTPTargetGroup
+          Type: forward
+      LoadBalancerArn: !Ref PublicLoadBalancer
```
```suggestion
      DefaultActions:
        - TargetGroupArn: !Ref DefaultInternalHTTPTargetGroup
          Type: forward
      LoadBalancerArn: !Ref InternalLoadBalancer
```

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-imported-certs-observability.yml
```diff
@@ -407,18 +413,39 @@ Resources:
...
       Port: 80
       Protocol: HTTP
+  InternalHTTPSListener:
+    Metadata:
+      'aws:copilot:description': 'An internal load balancer listener to route HTTPS traffic'
+    Type: AWS::ElasticLoadBalancingV2::Listener
+    Condition: ExportInternalHTTPSListener
+    Properties:
+      Certificates:
+        - CertificateArn: cert-1
```
iamhopaul123, line 435: 🤔

internal/pkg/template/templates/environment/cf.yml
huanjani: 🤦🏼

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -284,25 +290,52 @@ Resources:
...
+  InternalHTTPSListener:
+    Metadata:
+      'aws:copilot:description': 'An internal load balancer listener to route HTTPS traffic'
+    Type: AWS::ElasticLoadBalancingV2::Listener
+    Condition: ExportInternalHTTPSListener
+    Properties:
+      Certificates:
+{{- if .ImportCertARNs}}
+        - CertificateArn: {{index .ImportCertARNs 0}}
+{{- end}}
```
iamhopaul123, line 314: Does this mean if there's no imported cert this one will be an empty value? I thought CFN won't like it then.

I *think* we still need the lambda for the default certificate if they don't import any certs.

huanjani: Thanks for catching this. Added a `CreateInternalHTTPSListener` parameter that is true when certs are imported and no public subnets are imported.

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -158,14 +175,6 @@ Resources:
       GroupId: !Ref EnvironmentSecurityGroup
       IpProtocol: -1
       SourceSecurityGroupId: !Ref InternalLoadBalancerSecurityGroup
-  InternalALBIngressFromEnvironmentSecurityGroup:
```
efekarakus, line 161: Can we still keep the old security group settings?
I'd want us to default to the principle of least privileged although it doesn't meet the customer's usecase. Instead for now, we can Output the `InternalLoadBalancerSecurityGroup` so that people can write their own addons augmenting the permission for the internal ALB.

We can follow up with folks on the github issues to make sure an addon is an okay tradeoff - and very soon they'll be able to customize it with the env manifest

iamhopaul123 (approved):
:shipit:

internal/pkg/deploy/cloudformation/stack/env_test.go
```diff
@@ -185,6 +189,10 @@ func TestEnv_Parameters(t *testing.T) {
 					ParameterKey:   aws.String(envParamNATWorkloadsKey),
 					ParameterValue: aws.String(""),
 				},
+				{
+					ParameterKey:   aws.String(envParamCreateInternalHTTPSListenerKey),
+					ParameterValue: aws.String("false"),
```
efekarakus, line 196: can we add a test case for the conditional? 🥺

efekarakus (approved):
LGTM! feel free to remove the label after addressing feedback

PR #3591, opened 2022-05-20
chore: add subnet placement condition in env cfn template

huanjani:
If subnets are indicated for internal ALB placement, instruct accordingly in the CFN template. `cli` package changes to come.

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/env.go
+++ b/internal/pkg/deploy/cloudformation/stack/env.go
@@ -102,10 +102,10 @@ func (e *EnvStackConfig) Template() (string, error) {
 		ArtifactBucketARN:      e.in.ArtifactBucketARN,
 		ArtifactBucketKeyARN:   e.in.ArtifactBucketKeyARN,
 
-		ImportCertARNs:     e.importCertARNs(),
-		VPCConfig:          e.vpcConfig(),
-		InternalALBSubnets: e.internalALBSubnets(),
-		Telemetry:          e.telemetryConfig(),
+		ImportCertARNs:           e.importCertARNs(),
+		VPCConfig:                e.vpcConfig(),
+		CustomInternalALBSubnets: e.internalALBSubnets(),
+		Telemetry:                e.telemetryConfig(),
 
 		Version:       e.in.Version,
 		LatestVersion: deploy.LatestEnvTemplateVersion,
--- a/internal/pkg/deploy/cloudformation/stack/env.go
+++ b/internal/pkg/deploy/cloudformation/stack/env.go
@@ -102,9 +102,10 @@ func (e *EnvStackConfig) Template() (string, error) {
 		ArtifactBucketARN:      e.in.ArtifactBucketARN,
 		ArtifactBucketKeyARN:   e.in.ArtifactBucketKeyARN,
 
-		ImportCertARNs: e.importCertARNs(),
-		VPCConfig:      e.vpcConfig(),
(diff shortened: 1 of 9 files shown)


internal/pkg/manifest/validate_env.go
```diff
@@ -182,3 +185,13 @@ func (o publicHTTPConfig) Validate() error {
 	}
 	return nil
 }
+
+// Validate returns nil if privateHTTPConfig is configured correctly.
+func (o privateHTTPConfig) Validate() error {
+	for idx, certARN := range o.Certificates {
+		if _, err := arn.Parse(certARN); err != nil {
+			return fmt.Errorf(`validate "certificates[%d]": %w`, idx, err)
```
iamhopaul123: I still think both this and https://github.com/aws/copilot-cli/pull/3591/files#diff-80023f967d4bab363b7e0d95b3ff0467441beb4da4da9c49fbaf918cfb5c26b3L180 should be "parse" instead of "validate" because this function is to validate but the public function we call for catching the error is `arn.Parse`

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -196,6 +197,15 @@ func (e *EnvStackConfig) importCertARNs() []string {
 	return e.in.ImportCertARNs
 }
 
+func (e *EnvStackConfig) internalALBSubnets() []string {
+	// If a manifest is present, it is the only place we look.
+	if e.in.Mft != nil {
+		return e.in.Mft.HTTPConfig.Private.Subnets
+	}
+	// Fallthrough to SSM config.
+	return e.in.InternalALBSubnets
```
iamhopaul123, line 207: Is it even possible 🤔 How could users store this info in SSM using Copilot?

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -102,9 +102,10 @@ func (e *EnvStackConfig) Template() (string, error) {
 		ArtifactBucketARN:      e.in.ArtifactBucketARN,
 		ArtifactBucketKeyARN:   e.in.ArtifactBucketKeyARN,
 
-		ImportCertARNs: e.importCertARNs(),
-		VPCConfig:      e.vpcConfig(),
-		Telemetry:      e.telemetryConfig(),
+		ImportCertARNs:     e.importCertARNs(),
+		VPCConfig:          e.vpcConfig(),
+		InternalALBSubnets: e.internalALBSubnets(),
```
efekarakus: nit: I wonder if this would be easier to read in the template if we renamed it to:
```suggestion
		CustomInternalALBSubnets: e.internalALBSubnets(),
```

efekarakus (approved):
LGTM assuming the SSM change is going to come later

internal/pkg/deploy/cloudformation/stack/env.go
huanjani: 🤔 Because `{{- if .InternalALBSubnets}}` doesn't really distinguish the specified target subnets from regular subnets the internal ALB is being placed on, right? Yeah, that makes sense. Changing.

internal/pkg/deploy/cloudformation/stack/env.go
huanjani: It will be in the next PR. They will pass it in via a flag with `env init`.

iamhopaul123 (approved):

PR #3592, opened 2022-05-20
feat: allow placing workloads in any subnets

iamhopaul123:


internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
```diff
@@ -50,3 +50,9 @@ environments:
     command: crazy diamond
     variables:
       LOG_LEVEL: debug        # Log level for the "test" environment.
+  prod:
+    network:
+      vpc:
+        placement:
+          subnets: ["mockSubnetARN1", "mockSubnetARN2"]
```
efekarakus: nit: there are no ARNs :) just IDs
```suggestion
          subnets: ["mockSubnetID1", "mockSubnetID2"]
```

internal/pkg/manifest/workload.go
```diff
@@ -28,7 +29,13 @@ const (
 )
 
 // All placement options.
-var subnetPlacements = []string{string(PublicSubnetPlacement), string(PrivateSubnetPlacement)}
+var (
+	subnetPlacements           = []string{string(PublicSubnetPlacement), string(PrivateSubnetPlacement)}
+	SubnetPlacementForTemplate = map[PlacementString]string{
+		PrivateSubnetPlacement: template.PrivateSubnetsPlacement,
+		PublicSubnetPlacement:  template.PublicSubnetsPlacement,
+	}
```
efekarakus: Should we put this transformation logic only in the `stack` pkg because that's where it gets used

internal/pkg/manifest/validate.go
```diff
@@ -272,6 +272,12 @@ func (r RequestDrivenWebServiceConfig) Validate() error {
 	if err = r.Network.Validate(); err != nil {
 		return fmt.Errorf(`validate "network": %w`, err)
 	}
+	// TODO: remove this constraint when AppRunner services in public subnets can get internet outbound traffic.
```
efekarakus: nit: can we remove this comment

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
```diff
@@ -84,7 +84,7 @@ Resources:
                 Value: !Ref EnvName
               - Name: COPILOT_SERVICE_NAME
                 Value: !Ref WorkloadName
-{{- if eq .Network.SubnetsType "PrivateSubnets"}}
+{{- if or .Network.SubnetIDs (ne .Network.SubnetsType "")}}
```
efekarakus: should we add a method to `.Network` such as:
```
// IsSet returns true if there are VPC settings configured for launching compute.
func (net NetworkOpts) IsSet() bool
```

iamhopaul123: Hmm I am debating on that because this logic actually only applies to RDWS.

efekarakus: Gotcha that makes sense 💭 
My goal was to reduce the duplicate conditionals in the template as much as possible - maybe that means instead a method in `WorkloadOpts` like this:

```go
func (opt WorkloadOpts) RequiresVPCConnector() bool {
  if opt.WorkloadType != "Request-Driven Web Service" { return false }
  return len(opt.Network.SubnetIDs) > 0 || opt.Network.SubnetsType != ""
}
```
This isn't too bad - I just wish we had specific Opts structs instead of a generic `WorkloadOpts`. I'll let you decide!

efekarakus (approved):

internal/pkg/manifest/validate.go
```diff
@@ -272,6 +272,11 @@ func (r RequestDrivenWebServiceConfig) Validate() error {
 	if err = r.Network.Validate(); err != nil {
 		return fmt.Errorf(`validate "network": %w`, err)
 	}
+	if r.Network.VPC.Placement.PlacementString != nil &&
+		*r.Network.VPC.Placement.PlacementString != PrivateSubnetPlacement {
+		return fmt.Errorf(`placement "%s" is not supported for %s`,
+			string(*r.Network.VPC.Placement.PlacementString), RequestDrivenWebServiceType)
```
```suggestion
		return fmt.Errorf(`placement %q is not supported for %s`,
			*r.Network.VPC.Placement.PlacementString, RequestDrivenWebServiceType)
```

internal/pkg/manifest/validate.go
```diff
@@ -272,6 +272,11 @@ func (r RequestDrivenWebServiceConfig) Validate() error {
 	if err = r.Network.Validate(); err != nil {
 		return fmt.Errorf(`validate "network": %w`, err)
 	}
+	if r.Network.VPC.Placement.PlacementString != nil &&
```
Lou1415926, line 275: Thought we had this!

dannyrandall (approved):

internal/pkg/manifest/validate.go
iamhopaul123: Yeah we had it before. In the previous PR we thought placing vpc connectors in public subnets work, so we removed this validation.

PR #3596, opened 2022-05-23
chore(describe): add Manifest method to EnvDecriber

efekarakus:

### DIFF
--- a/internal/pkg/describe/env.go
+++ b/internal/pkg/describe/env.go
@@ -19,6 +19,8 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/deploy"
 	cfnstack "github.com/aws/copilot-cli/internal/pkg/deploy/cloudformation/stack"
 	"github.com/aws/copilot-cli/internal/pkg/describe/stack"
+	"github.com/aws/copilot-cli/internal/pkg/manifest"
+	"github.com/aws/copilot-cli/internal/pkg/template"
 	"github.com/aws/copilot-cli/internal/pkg/term/color"
 )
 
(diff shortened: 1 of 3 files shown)


Lou1415926 (approved):

internal/pkg/describe/env.go
```diff
@@ -140,7 +168,7 @@ func (d *EnvDescriber) Params() (map[string]string, error) {
 	return descr.Parameters, nil
 }
 
-// Params returns the outputs of the environment stack.
```
iamhopaul123, line 143: 🥲

internal/pkg/describe/env.go
```diff
@@ -131,6 +133,32 @@ func (d *EnvDescriber) Describe() (*EnvDescription, error) {
 	return d.description, nil
 }
 
+// Manifest returns the contents of the manifest used to deploy an environment stack.
+func (d *EnvDescriber) Manifest() ([]byte, error) {
+	tpl, err := d.cfn.StackMetadata()
+	if err != nil {
+		return nil, err
```
iamhopaul123, line 140: We might want to wrap this error 🤔

internal/pkg/describe/env.go
```diff
@@ -131,6 +133,32 @@ func (d *EnvDescriber) Describe() (*EnvDescription, error) {
...
+	}
+
+	metadata := struct {
+		Manifest string `yaml:"Manifest"`
```
iamhopaul123, line 144: Why do we capitalize `manifest`?

internal/pkg/describe/env.go
efekarakus: I didn't on purpose because the error msg from the downstream package already has all the info 👍

internal/pkg/describe/env.go
efekarakus: Otherwise, `yaml.Unmarshal` won't be able to populate the field. The types can be private but the fields must be public

iamhopaul123: Oh i meant `yaml:"Manifest"`

efekarakus: That's how we write it in the CFN template:
```
Metadata:
  Version:
  Manifest:
```

iamhopaul123: I see it looks good since all the other fields are capitalized as well.

iamhopaul123 (approved):

PR #3597, opened 2022-05-23
chore: add hidden --manifest flag to env show

efekarakus:

### DIFF
--- a/internal/pkg/cli/env_show.go
+++ b/internal/pkg/cli/env_show.go
@@ -34,6 +34,7 @@ type showEnvVars struct {
 	name                  string
 	shouldOutputJSON      bool
 	shouldOutputResources bool
+	shouldOutputManifest  bool
 }
 
 type showEnvOpts struct {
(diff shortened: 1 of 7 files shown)


internal/pkg/cli/env_show.go
```diff
@@ -140,17 +139,40 @@ func (o *showEnvOpts) askApp() error {
 	return nil
 }
 
-func (o *showEnvOpts) askEnvName() error {
-	//return if env name is set by flag
+func (o *showEnvOpts) validateApp() error {
+	_, err := o.store.GetApplication(o.appName)
+	if err != nil {
```
iamhopaul123:
```suggestion
	if _, err := o.store.GetApplication(o.appName); err != nil {
```

internal/pkg/cli/env_show.go
```diff
@@ -140,17 +139,40 @@ func (o *showEnvOpts) askApp() error {
...
 	env, err := o.sel.Environment(fmt.Sprintf(envShowNamePrompt, color.HighlightUserInput(o.appName)), envShowHelpPrompt, o.appName)
 	if err != nil {
 		return fmt.Errorf("select environment for application %s: %w", o.appName, err)
 	}
 	o.name = env
+	return nil
+}
+
+func (o *showEnvOpts) validateEnv() error {
+	_, err := o.store.GetEnvironment(o.appName, o.name)
```
iamhopaul123: same here

internal/pkg/cli/env_show.go
```diff
@@ -140,17 +139,40 @@ func (o *showEnvOpts) askApp() error {
 	return nil
 }
 
-func (o *showEnvOpts) askEnvName() error {
-	//return if env name is set by flag
+func (o *showEnvOpts) validateApp() error {
+	_, err := o.store.GetApplication(o.appName)
+	if err != nil {
+		return fmt.Errorf("validate application name '%s': %v", o.appName, err)
```
iamhopaul123: I feel the reason why we need this strange pattern for error wrapping (validate ... but the actual API we call is `GetApplication`) is because we probably need a `validate` package.

iamhopaul123 (approved):

internal/pkg/cli/env_show.go
efekarakus: you think so? maybe a `cli/validate` package isn't a bad idea for us to move all of our `validate.go` logic

internal/pkg/cli/env_show.go
```diff
@@ -140,17 +139,38 @@ func (o *showEnvOpts) askApp() error {
...
 	if err != nil {
 		return fmt.Errorf("select environment for application %s: %w", o.appName, err)
 	}
 	o.name = env
+	return nil
+}
 
+func (o *showEnvOpts) validateEnv() error {
+	if _, err := o.store.GetEnvironment(o.appName, o.name); err != nil {
+		return fmt.Errorf("validate environment name '%s' in application '%s': %v", o.name, o.appName, err)
```
Lou1415926: @dannyrandall told me about using `%q` to quote the strings!

Lou1415926 (approved):

internal/pkg/cli/env_show.go
Lou1415926: Feel free to remove the label when you are ready! Adding it just in case.

PR #3607, opened 2022-05-25
feat(cli): add flag to enable internal ALB subnet placement

huanjani:


internal/pkg/cli/deploy/deploy.go
```diff
@@ -1226,6 +1231,31 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
...
+	}
+	for _, placementSubnet := range d.env.CustomConfig.InternalALBSubnets {
+		for _, importedSubnet := range d.env.CustomConfig.ImportVPC.PrivateSubnetIDs {
+			if placementSubnet == importedSubnet {
+				return nil
+			}
+		}
+		for _, importedSubnet := range d.env.CustomConfig.ImportVPC.PublicSubnetIDs {
+			if placementSubnet == importedSubnet {
+				log.Warningln(`You are placing an internal ALB on a public subnet. Your load balancer
```
efekarakus: Will it? I don't know if there is an actual security concern with placing an internal ALB onto public subnets - it's just a bit odd why anyone would do that

internal/pkg/cli/deploy/deploy.go
```diff
@@ -1226,6 +1231,31 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
 	return nil
 }
 
+func (d backendSvcDeployer) validateInternalALBSubnets() error {
```
efekarakus: nit: pointer receiver for efficiency 🥺 
```suggestion
func (d *backendSvcDeployer) validateInternalALBSubnets() error {
```

efekarakus: Can we potentially err earlier than `deploy` for example when users run `env init --internal-alb-subnets` can we add logic in `cli/validate.go` _then_ to make sure that the subnets passed are indeed within their VPC.

```
if user has --internal-alb-subnets but chooses our default VPC:
   return err saying copilot will place the internal alb in the default private subnets, if they want to control then either import or adjust default settings.

if user has --internal-alb-subnets and chooses to adjust their VPC settings:
   return err if the --internal-alb-subnets is not a subset of their vpc.AZs

if user has --internal-alb-subnets and chooses to import a VPC:
   return err if the --internal-alb-subnets is not a subset of their imported subnets
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -191,8 +191,10 @@ func (e *EnvStackConfig) telemetryConfig() *template.Telemetry {
 
 func (e *EnvStackConfig) importCertARNs() []string {
 	// If a manifest is present, it is the only place we look at.
+	// Look under both public and private.
 	if e.in.Mft != nil {
-		return e.in.Mft.HTTPConfig.Public.Certificates
+		allCerts := append(e.in.Mft.HTTPConfig.Public.Certificates, e.in.Mft.HTTPConfig.Private.Certificates...)
+		return allCerts
```
efekarakus: hmmm does this mean we should really be distinguishing between public imported certs and private imported certs?

I think so - like in the `template.EnvOpts` have two fields `PublicImportedCertARNs` and `PrivateImportedCertARNs` - otherwise, if a customer has both a LBWS and Backend Service with `http` then we'll try to attach public and private certs to both lbs

internal/pkg/cli/deploy/deploy.go
huanjani: Yeah, that's why I made it a warning and not an error; someone may unwittingly do it. @dannyrandall tested it and that's what happened; can you confirm, Danny?

efekarakus: I'd expect it to get a DNS name - but it should be private - is it really accessible from the interwebz?

dannyrandall: Yeah I remembered thinking what you're saying @huanjani, but I just tested it to be sure and an internal LB in a public subnet is not reachable from the internet. Maybe I tested it wrong last week :)

internal/pkg/manifest/env.go
```diff
@@ -217,7 +217,16 @@ func (cfg *environmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
...
-	cfg.Public.Certificates = env.ImportCertARNs
+	if len(env.InternalALBSubnets) != 0 {
+		cfg.Private.Subnets = env.InternalALBSubnets
+	}
+	if env.ImportVPC != nil {
+		if len(env.ImportVPC.PublicSubnetIDs) == 0 {
+			cfg.Private.Certificates = env.ImportCertARNs
+			return
+		}
+		cfg.Public.Certificates = env.ImportCertARNs
```
dannyrandall: i think this is not backwards-compatible with the current version - this requires importing a VPC to be able to use imported certs.

i think the same should go for the private lb - let them use imported certs without importing a VPC.

internal/pkg/cli/deploy/deploy.go
huanjani: Done! In `env_init`.

internal/pkg/manifest/env.go
huanjani: Fixed and added unit tests.

internal/pkg/deploy/cloudformation/stack/env.go
huanjani: Okay, but the MVP behavior is still to consider the certs public unless there are no public subnets, in which case we consider them private...right?

efekarakus: Yup yup - that's the end behavior that we'll maintain for the MVP. 

The comment was just to re-structure the `stack/env.go` and `template` files to split `ImportCertArns` into two so that once we have the env manifest, we have a path for enabling certs on both load balancers.

I'm 🆗 w keeping it like this in this PR so that it doesn't become v large, if we can send the follow-up PR with imported certs split?

internal/pkg/cli/env_init.go
```diff
@@ -368,6 +374,9 @@ func (o *initEnvOpts) validateCustomizedResources() error {
 	if (o.importVPC.isSet() || o.adjustVPC.isSet()) && o.defaultConfig {
 		return fmt.Errorf("cannot import or configure vpc if --%s is set", defaultConfigFlag)
 	}
+	if o.internalALBSubnets != nil && o.adjustVPC.isSet() || o.defaultConfig {
```
efekarakus: I think this conditional will evaluate to true even when the user _only_ specifies the default config and no internal internal alb subnets. I believe we want parentheses here:
```suggestion
	if o.internalALBSubnets != nil && (o.adjustVPC.isSet() || o.defaultConfig) {
```

internal/pkg/cli/env_init.go
```diff
@@ -80,6 +81,9 @@ var (
 	envInitAdjustEnvResourcesSelectOption = "Yes, but I'd like configure the default resources (CIDR ranges, AZs)."
 	envInitImportEnvResourcesSelectOption = "No, I'd like to import existing resources (VPC, subnets)."
 	envInitCustomizedEnvTypes             = []string{envInitDefaultConfigSelectOption, envInitAdjustEnvResourcesSelectOption, envInitImportEnvResourcesSelectOption}
+
+	errDefaultConfigWithSubnetPlacement = errors.New(`to specify internal ALB subnet placement, you must import existing resources, including subnets.
```
```suggestion
	errInternalALBPlacementWithManagedVPC = errors.New(`to specify internal ALB subnet placement, you must import existing resources, including subnets.
```

internal/pkg/cli/env_init.go
```diff
@@ -767,6 +785,33 @@ func (o *initEnvOpts) validateCredentials() error {
 	return nil
 }
 
+func (o *initEnvOpts) validateInternalALBSubnets() error {
+	if o.internalALBSubnets == nil {
+		return nil
+	}
+	isImported := make(map[string]bool)
+	for _, placementSubnet := range o.internalALBSubnets {
+		for _, importedPrivateSubnet := range o.importVPC.PrivateSubnetIDs {
```
efekarakus: nit: just syntactic sugar - I think we can reduce the two for-loops into one:
```suggestion
		for _, subnet := range append(o.importVPC.PrivateSubnetIDs, o.importVPC.PublicSubnetIDs...) {
```

internal/pkg/cli/env_init.go
```diff
@@ -767,6 +785,33 @@ func (o *initEnvOpts) validateCredentials() error {
...
+		for _, importedPublicSubnet := range o.importVPC.PublicSubnetIDs {
+			if placementSubnet == importedPublicSubnet {
+				isImported[placementSubnet] = true
+			}
+		}
+	}
+	if len(isImported) != len(o.internalALBSubnets) {
+		return fmt.Errorf("%s %s %s designated for ALB placement, but %s imported",
+			english.PluralWord(len(o.internalALBSubnets), "subnet", "subnets"),
+			o.internalALBSubnets,
```
efekarakus:
```suggestion
			strings.Join(o.internalALBSubnets, ","),
```

internal/pkg/cli/flag.go
```diff
@@ -278,10 +279,12 @@ To use it for an ECS service, specify --generate-cmd <cluster name>/<service nam
...
-	vpcIDFlagDescription           = "Optional. Use an existing VPC ID."
-	publicSubnetsFlagDescription   = "Optional. Use existing public subnet IDs."
-	privateSubnetsFlagDescription  = "Optional. Use existing private subnet IDs."
-	certsFlagDescription           = "Optional. Apply existing ACM certificates to the internet-facing load balancer."
+	vpcIDFlagDescription              = "Optional. Use an existing VPC ID."
+	publicSubnetsFlagDescription      = "Optional. Use existing public subnet IDs."
+	privateSubnetsFlagDescription     = "Optional. Use existing private subnet IDs."
+	certsFlagDescription              = "Optional. Apply existing ACM certificates to the internet-facing load balancer."
+	internalALBSubnetsFlagDescription = `Optional. Specify subnet placement for an internal load balancer.
+Cannot be specified with --default-config or any of the --override flags.`
```
efekarakus: Should we say something about the default? and that the input is IDs
```suggestion
	internalALBSubnetsFlagDescription = `Optional. Specify subnet IDs for an internal load balancer.
By default, the load balancer will be placed in your private subnets.
Cannot be specified with --default-config or any of the --override flags.`
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -189,21 +189,42 @@ func (e *EnvStackConfig) telemetryConfig() *template.Telemetry {
...
+func (e *EnvStackConfig) importPublicCertARNs() []string {
 	// If a manifest is present, it is the only place we look at.
 	if e.in.Mft != nil {
 		return e.in.Mft.HTTPConfig.Public.Certificates
 	}
 	// Fallthrough to SSM config.
+	if e.in.ImportCertARNs == nil {
+		return nil
+	}
 	return e.in.ImportCertARNs
```
efekarakus: Should this be like this instead?
```go
if len(e.in.ImportVPCConfig.PublicSubnetIDs) == 0 {
   return nil
}
return e.in.ImportCertArns
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -189,21 +189,42 @@ func (e *EnvStackConfig) telemetryConfig() *template.Telemetry {
...
+		return e.in.Mft.HTTPConfig.Private.Certificates
+	}
+	// Fallthrough to SSM config.
+	if e.in.ImportCertARNs == nil {
+		return nil
+	}
+	if e.in.ImportVPCConfig.PublicSubnetIDs == nil {
+		return e.in.ImportCertARNs
+	}
+	return nil
```
efekarakus: nit:
```suggestion
	// Fallthrough to SSM config.
	if len(e.in.ImportVPCConfig.PublicSubnetIDs) == 0 {
		return e.in.ImportCertARNs
	}
	return nil
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -189,21 +189,42 @@ func (e *EnvStackConfig) telemetryConfig() *template.Telemetry {
...
 	// If a manifest is present, it is the only place we look.
 	if e.in.Mft != nil {
-		return e.in.Mft.HTTPConfig.Private.Subnets
+		return e.in.Mft.HTTPConfig.Private.InternalALBSubnets
 	}
 	// Fallthrough to SSM config.
+	if e.in.InternalALBSubnets == nil {
+		return nil
+	}
 	return e.in.InternalALBSubnets
```
efekarakus:
```suggestion
	return e.in.InternalALBSubnets
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -416,7 +416,8 @@ Resources:
 {{- end}}
 {{include "cfn-execution-role" . | indent 2}}
 {{include "environment-manager-role" . | indent 2}}
-{{- if not .ImportCertARNs}}
+{{- if not .PublicImportedCertARNs}}
+{{- if not .PrivateImportedCertARNs}}
```
efekarakus: I think we want to create to create the resource regardless if the customer has imported private certs since they get used for the public lb
```suggestion
{{- if not .PublicImportedCertARNs}}
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -545,7 +547,8 @@ Outputs:
     Description: The role to be assumed by the Cloudformation service when it deploys application infrastructure.
     Export:
       Name: !Sub ${AWS::StackName}-CFNExecutionRoleARN
-{{- if not .ImportCertARNs}}
+{{- if not .PublicImportedCertARNs}}
+{{- if not .PrivateImportedCertARNs}}
```
```suggestion
{{- if not .PublicImportedCertARNs}}
```

internal/pkg/template/templates/environment/partials/environment-manager-role.yml
```diff
@@ -19,14 +19,25 @@ EnvironmentManagerRole:
...
+{{- end}}
+{{- end}}
+{{- if .PrivateImportedCertARNs}}
+        - Sid: ImportedCertificates
+          Effect: Allow
+          Action: [
+            acm:DescribeCertificate
+          ]
+          Resource:
+{{- range $arn := .PrivateImportedCertARNs}}
```
efekarakus: folks should be able to import both public and private certs soon, so I think we can do:
```
{{ if or (.PublicImportedCertARNs .PrivateImportedCertARNs)}}
-  Sid: ImportedCertificates
   Effect: Allow
   Action: [
        acm:DescribeCertificate
   ]
   Resource:
{{- range $arn := .PublicImportedCertARNs}}
          - "{{$arn}}"
{{- end}}
{{- range $arn := .PrivateImportedCertARNs}}
          - "{{$arn}}"
{{- end}}
```

internal/pkg/deploy/cloudformation/stack/env.go
huanjani: 🤔 Those changes would give me nil pointer errors if those fields are nil.

internal/pkg/cli/env_init.go
```diff
@@ -141,10 +145,11 @@ type initEnvVars struct {
...
 
-	importVPC   importVPCVars // Existing VPC resources to use instead of creating new ones.
-	adjustVPC   adjustVPCVars // Configure parameters for VPC resources generated while initializing an environment.
-	telemetry   telemetryVars // Configure observability and monitoring settings.
-	importCerts []string      // Addtional existing ACM certificates to use.
+	importVPC          importVPCVars // Existing VPC resources to use instead of creating new ones.
+	adjustVPC          adjustVPCVars // Configure parameters for VPC resources generated while initializing an environment.
+	telemetry          telemetryVars // Configure observability and monitoring settings.
+	importCerts        []string      // Addtional existing ACM certificates to use.
+	internalALBSubnets []string      // Private subnets to be used for internal ALB placement.
```
dannyrandall:
```suggestion
	internalALBSubnets []string      // Subnets to be used for internal ALB placement.
```

internal/pkg/cli/env_init.go
```diff
@@ -767,6 +785,28 @@ func (o *initEnvOpts) validateCredentials() error {
 	return nil
 }
 
+func (o *initEnvOpts) validateInternalALBSubnets() error {
+	if o.internalALBSubnets == nil {
+		return nil
+	}
```
dannyrandall:
```suggestion
```
can just remove entirely

internal/pkg/manifest/validate_env.go
```diff
@@ -36,6 +36,15 @@ func (e EnvironmentConfig) Validate() error {
 	if err := e.HTTPConfig.Validate(); err != nil {
 		return fmt.Errorf(`validate "http config": %w`, err)
 	}
+
+	if e.HTTPConfig.Private.InternalALBSubnets != nil {
+		if !e.Network.VPC.imported() {
+			return errors.New("in order to specify internal ALB subnet placement, subnets must be imported ")
```
efekarakus: nit: 
```suggestion
			return errors.New("in order to specify internal ALB subnet placement, subnets must be imported")
```

internal/pkg/manifest/validate_env.go
```diff
@@ -198,3 +207,21 @@ func (o privateHTTPConfig) Validate() error {
 	}
 	return nil
 }
+
+func (c EnvironmentConfig) validateSubnets() error {
```
efekarakus: nit:
```suggestion
func (c EnvironmentConfig) validateInternalALBSubnets() error {
```

internal/pkg/deploy/cloudformation/stack/env.go
efekarakus: Wouldn't the above also result into a segfault if `e.in.ImportVPCConfig` is `nil`? I might be misunderstanding something!

What if we do:
```
if e.in.ImportVPCConfig != nil && len(e.in.ImportVPCConfig.PublicSubnetIDs) == 0
```

efekarakus (commented):
LGTM, just one question and then we can :shipit:

internal/pkg/cli/deploy/deploy_test.go
```diff
@@ -1136,6 +1139,9 @@ func TestBackendSvcDeployer_stackConfiguration(t *testing.T) {
 			},
 			Env: &config.Environment{
 				Name: mockEnvName,
+				CustomConfig: &config.CustomizeEnv{
+					InternalALBSubnets: []string{},
+				},
```
dannyrandall: can this not be left unset? (and the one above)

dannyrandall (approved):
lgtm:tada: just a couple of nits😊

internal/pkg/manifest/validate_env.go
huanjani: I did this bc in the manifest the field will just be `subnets` but you're right, this method should be more clearly named.

internal/pkg/cli/env_init.go
huanjani: discussed offline

internal/pkg/cli/env_init.go
```diff
@@ -80,6 +81,9 @@ var (
 	envInitAdjustEnvResourcesSelectOption = "Yes, but I'd like configure the default resources (CIDR ranges, AZs)."
 	envInitImportEnvResourcesSelectOption = "No, I'd like to import existing resources (VPC, subnets)."
 	envInitCustomizedEnvTypes             = []string{envInitDefaultConfigSelectOption, envInitAdjustEnvResourcesSelectOption, envInitImportEnvResourcesSelectOption}
+
+	errInternalALBPlacementWithManagedVPC = errors.New(`to specify internal ALB subnet placement, you must import existing resources, including subnets.
```
iamhopaul123: Can we simplify the error msg by any chance and put more details in a log msg?

internal/pkg/cli/env_init.go
```diff
@@ -368,6 +374,9 @@ func (o *initEnvOpts) validateCustomizedResources() error {
 	if (o.importVPC.isSet() || o.adjustVPC.isSet()) && o.defaultConfig {
 		return fmt.Errorf("cannot import or configure vpc if --%s is set", defaultConfigFlag)
 	}
+	if o.internalALBSubnets != nil && (o.adjustVPC.isSet() || o.defaultConfig) {
```
iamhopaul123:
```suggestion
	if o.internalALBSubnets != nil && !o.importVPC.isSet() {
```
How about this?

internal/pkg/cli/env_init.go
```diff
@@ -767,6 +783,28 @@ func (o *initEnvOpts) validateCredentials() error {
...
+		return nil
+	}
+	isImported := make(map[string]bool)
+	for _, placementSubnet := range o.internalALBSubnets {
+		for _, subnet := range append(o.importVPC.PrivateSubnetIDs, o.importVPC.PublicSubnetIDs...) {
+			if placementSubnet == subnet {
+				isImported[placementSubnet] = true
+			}
+		}
+	}
```
iamhopaul123, line 796: Can we simplify the logic by erring out if we find any subnet within `o.internalALBSubnets` does not appear in the subnets of `importVPC`?

iamhopaul123: nvm i think what you have here is a better user experience! Awesome!

internal/pkg/config/env.go
```diff
@@ -34,17 +34,23 @@ func (e *Environment) HasImportedCerts() bool {
...
 func (c *CustomizeEnv) IsEmpty() bool {
 	if c == nil {
 		return true
 	}
-	return c.ImportVPC == nil && c.VPCConfig == nil && len(c.ImportCertARNs) == 0
+	return c.ImportVPC == nil && c.VPCConfig == nil && len(c.ImportCertARNs) == 0 && len(c.InternalALBSubnets) == 0
+}
+
+// HasInternalALBSubnets returns if subnet placement of an internal ALB has been specified.
+func (e *Environment) HasInternalALBSubnets() bool {
```
iamhopaul123: Seems like this one is not used...Maybe consider adding it when we use it?

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -189,19 +189,34 @@ func (e *EnvStackConfig) telemetryConfig() *template.Telemetry {
...
-func (e *EnvStackConfig) importCertARNs() []string {
+func (e *EnvStackConfig) importPublicCertARNs() []string {
 	// If a manifest is present, it is the only place we look at.
 	if e.in.Mft != nil {
 		return e.in.Mft.HTTPConfig.Public.Certificates
 	}
 	// Fallthrough to SSM config.
+	if e.in.ImportCertARNs == nil {
+		return nil
+	}
```
iamhopaul123: Why do we add this 🥲 I don't see the difference if we remove this block

internal/pkg/cli/env_init.go
huanjani: Yes! Now that we're combining the two sets of subnets, that makes more sense!

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -189,19 +189,34 @@ func (e *EnvStackConfig) telemetryConfig() *template.Telemetry {
...
+func (e *EnvStackConfig) importPrivateCertARNs() []string {
+	// If a manifest is present, it is the only place we look at.
+	if e.in.Mft != nil {
+		return e.in.Mft.HTTPConfig.Private.Certificates
+	}
+	// Fallthrough to SSM config.
+	if e.in.ImportVPCConfig != nil && len(e.in.ImportVPCConfig.PublicSubnetIDs) == 0 {
+		return e.in.ImportCertARNs
+	}
+	return nil
```
iamhopaul123, line 213:
```suggestion
	return e.in.ImportCertARNs
```
I _think_ we might not need to double validate this (as this is handled by env init validate).

internal/pkg/deploy/cloudformation/stack/env.go
iamhopaul123: <strike>Yeah I think should be this based on our last decision on this. Plus maybe adding adequate comment describing why we do this. </strike>

Sorry misread this. Maybe we should error out in `env init` validation instead?

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-imported-certs-observability.yml
```diff
@@ -431,21 +431,12 @@ Resources:
     Type: AWS::ElasticLoadBalancingV2::Listener
     Condition: ExportInternalHTTPSListener
     Properties:
-      Certificates:
-        - CertificateArn: cert-1
```
iamhopaul123, line 435: Hmm why do they disappear? Should we update the integ test instead?

huanjani: `cert-1` and `cert-2` are public:
```
http:
  public:
    certificates:
      - cert-1
      - cert-2
```
so they should disappear there. 

Will add an integ test with private certs in another PR.

internal/pkg/deploy/cloudformation/stack/env.go
huanjani: Here and below: I think this is more about assigning/distinguishing between public/private certs than validating.

internal/pkg/deploy/cloudformation/stack/env.go
huanjani: (see above)

internal/pkg/cli/env_init.go
```diff
@@ -374,8 +371,10 @@ func (o *initEnvOpts) validateCustomizedResources() error {
 	if (o.importVPC.isSet() || o.adjustVPC.isSet()) && o.defaultConfig {
 		return fmt.Errorf("cannot import or configure vpc if --%s is set", defaultConfigFlag)
 	}
-	if o.internalALBSubnets != nil && (o.adjustVPC.isSet() || o.defaultConfig) {
-		return errInternalALBPlacementWithManagedVPC
+	if o.internalALBSubnets != nil && !o.importVPC.isSet() {
+		log.Error(`To specify internal ALB subnet placement, you must import existing resources, including subnets.
+For default config without subnet placement specification, Copilot will place the internal ALB in the generated private subnets.`)
+		return fmt.Errorf("subnets '%s' specified for internal ALB placement, but those subnets not imported", strings.Join(o.internalALBSubnets, ", "))
```
efekarakus:
```suggestion
		return fmt.Errorf("subnets '%s' specified for internal ALB placement, but those subnets are not imported", strings.Join(o.internalALBSubnets, ", "))
```

efekarakus (approved):

internal/pkg/cli/env_init.go
```diff
@@ -767,6 +782,28 @@ func (o *initEnvOpts) validateCredentials() error {
 	return nil
 }
 
+func (o *initEnvOpts) validateInternalALBSubnets() error {
+	if len(o.internalALBSubnets) == 0 {
+		return nil
+	}
+	isImported := make(map[string]bool)
+	for _, placementSubnet := range o.internalALBSubnets {
```
nodir-t, line 790: nit: this is O(N^2). This can be O(N):
1. all `internalALBSubnets` to a new map
2. Go over subnets. If a given subnet is present in the map above, then add to `isImported`

internal/pkg/cli/env_init.go
```diff
@@ -767,6 +782,28 @@ func (o *initEnvOpts) validateCredentials() error {
...
+		for _, subnet := range append(o.importVPC.PrivateSubnetIDs, o.importVPC.PublicSubnetIDs...) {
```
nodir-t, line 791: note that this might mutate the underlying array of `o.importVPC.PrivateSubnetIDs`. 
IMO it should be either [cloned](https://github.com/golang/go/wiki/SliceTricks#copy), or have two loops

internal/pkg/cli/env_init.go
nodir-t: Thinking about this more, this can be simpler than above. The new map can be called `missing` and in the loop (2) remove items from the map. Then in the L797 check if the map isn't empty, i.e. something is still missing

PR #3608, opened 2022-05-25
feat: lbws and backend service `hosted_zone`

dannyrandall:


internal/pkg/cli/deploy/deploy.go
```diff
@@ -907,6 +907,10 @@ func (d *lbSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (*svcS
 	if err != nil {
 		return nil, err
 	}
+
+	if d.lbMft.RoutingRule.HostedZone != nil {
+		return nil, fmt.Errorf("hosted_zone is not supported for Load Balanced Web Services")
```
iamhopaul123:
```suggestion
		return nil, fmt.Errorf(`"hosted_zone" is not supported for Load Balanced Web Services`)
```

internal/pkg/cli/deploy/deploy.go
```diff
@@ -1210,8 +1214,13 @@ func validateAppVersionForAlias(appName string, appVersionGetter versionGetter)
...
 
 func (d *backendSvcDeployer) validateALBRuntime() error {
-	if d.backendMft.RoutingRule.EmptyOrDisabled() || d.backendMft.RoutingRule.Alias.IsEmpty() || !d.env.HasImportedCerts() {
+	switch {
+	case d.backendMft.RoutingRule.EmptyOrDisabled():
 		return nil
+	case !d.backendMft.RoutingRule.Alias.IsEmpty() && !d.env.HasImportedCerts():
+		return fmt.Errorf(`setting "alias" requires an environment with imported certs`)
+	case d.backendMft.RoutingRule.HostedZone != nil && d.backendMft.RoutingRule.Alias.IsEmpty():
+		return fmt.Errorf(`setting "hosted_zone" requires "alias"`)
```
iamhopaul123: Feel like it would be better to use negative sentences for error msgs.

internal/pkg/cli/deploy/deploy.go
```diff
@@ -1210,8 +1214,13 @@ func validateAppVersionForAlias(appName string, appVersionGetter versionGetter)
 }
 
 func (d *backendSvcDeployer) validateALBRuntime() error {
-	if d.backendMft.RoutingRule.EmptyOrDisabled() || d.backendMft.RoutingRule.Alias.IsEmpty() || !d.env.HasImportedCerts() {
+	switch {
+	case d.backendMft.RoutingRule.EmptyOrDisabled():
 		return nil
+	case !d.backendMft.RoutingRule.Alias.IsEmpty() && !d.env.HasImportedCerts():
```
iamhopaul123: what if the env has imported certs but alias is not set. Are we good with that?

internal/pkg/cli/deploy/deploy.go
iamhopaul123: I don't see a reason why hosted zone can't be used for LBWS. I think hosted zone can only be used when the env has imported certs, which requires alias.

Lou1415926: Applies only if we still want to keep `hosted_zone` out of lbws's manifest: Do you think LBWS's manifest `Validate()` is a better place to do this?

internal/pkg/cli/deploy/deploy.go
```diff
@@ -1210,8 +1214,13 @@ func validateAppVersionForAlias(appName string, appVersionGetter versionGetter)
...
+		return fmt.Errorf(`setting "alias" requires an environment with imported certs`)
+	case d.backendMft.RoutingRule.HostedZone != nil && d.backendMft.RoutingRule.Alias.IsEmpty():
```
Lou1415926: Same question: perhaps  manifest `Validate` is a better home for them? What do ya think

dannyrandall: yeah, that looks like a better place - i'll move these over there 😊  same for the lbws one above.

internal/pkg/cli/deploy/deploy.go
dannyrandall: i agree there's no reason why LBWS shouldn't support `hosted_zone`. however, i just want to focus on support for backend service for now, and we can bring it to LBWS once that's working.

internal/pkg/cli/deploy/deploy.go
dannyrandall: i think so? it would make sense to me to just use http in that case

internal/pkg/cli/deploy/deploy.go
dannyrandall: discussed offline, will add lbws hosted zone support now

internal/pkg/cli/deploy/deploy.go
dannyrandall: see if you like the new error better

internal/pkg/cli/deploy/deploy.go
iamhopaul123: But then what domain do we use for them 🤔

dannyrandall: umm... just the default lb alias? so just a listener rule without any `host-header` restriction?

though i'm realizing that could cause a weird situation where an A-record points to the LB and then this listener rule picks it up 🤔

dannyrandall: let's just avoid that situation until we can make the rule priority lambda smarter and just block the alias not set/imported certs situation. i'll make that change :)

internal/pkg/cli/deploy/deploy.go
```diff
@@ -1210,8 +1210,16 @@ func validateAppVersionForAlias(appName string, appVersionGetter versionGetter)
...
+	switch {
+	case d.backendMft.RoutingRule.Alias.IsEmpty() && d.env.HasImportedCerts():
+		return &errSvcWithNoALBAliasDeployingToEnvWithImportedCerts{
+			name:    d.name,
+			envName: d.env.Name,
+		}
+	case d.backendMft.RoutingRule.Alias.IsEmpty():
 		return nil
+	case !d.env.HasImportedCerts():
+		return fmt.Errorf(`cannot specify "alias" in an environment without imported certs`)
```
efekarakus, line 1222: What do you think of re-structuring like this:
```suggestion
	case d.backendMft.RoutingRule.Alias.IsEmpty() && !d.env.HasImportedCerts():
		return fmt.Errorf(`cannot specify "alias" in an environment without imported certs`)
	default:
		return nil
```

efekarakus (approved):
Looks good to me 👍

internal/pkg/cli/deploy/deploy.go
dannyrandall: problem is we want the case where the certs are imported and the alias is not empty to fall through to the checks below, but i'm happy to make the case more verbose like 
```suggestion
	case d.backendMft.RoutingRule.Alias.IsEmpty():
		return nil
	case !d.backendMft.RoutingRule.Alias.IsEmpty() && !d.env.HasImportedCerts():
		return fmt.Errorf(`cannot specify "alias" in an environment without imported certs`)
```
if you think that helps with readability😊

internal/pkg/manifest/validate.go
```diff
@@ -206,6 +206,11 @@ func (b BackendServiceConfig) Validate() error {
 	if err = b.ImageOverride.Validate(); err != nil {
 		return err
 	}
+	if !b.RoutingRule.EmptyOrDisabled() {
```
Lou1415926: My understanding of `Validate` is that - ideally `Validate` should be called upon every single fields no matter what - like whether the field is empty or not. 

The use of this conditional implies that `b.RoutingRule.isEmpty` and `b.RoutingRule.Disabled` are valid inputs for backend `http` - which should have been judged by `Validate`.

I see that this is probably because backend `http` has different validation rules than lbws `http`: lbws requires `path` to be non-nil unless `http: false`, while backend doesn't since its `http` isn't enabled by default. 

Do you think this probably indicates that they are not really the same type of `RoutingRuleOrConfig` 🤔.

What do you think of something like:
```go
type RoutingRuleOrConfig struct { /*same as today*/ }
type LBWSRoutingRuleOrConfig RoutingRuleOrConfig
type BackendRoutingRuleOrConfig RoutingRuleOrConfig

func (r RoutingRuleOrConfig) Validate() error {
   // Validate except for `path`
}

func (r HTTPRoutingRuleOrConfig) Validate() error {
   if err := RoutingRuleConfiguration(r).Validate(); err != nil {
      return err
   }
   if r.Path == nil {
            missingField: "path",
         }
   }
   return nil
}

func (r BackendRoutingRuleOrConfig) Validate() error {
   if err := RoutingRuleConfiguration(r).Validate(); err != nil {
      return err
   }
   return nil
```

huanjani (approved):
Looks great!

PR #3612, opened 2022-05-26
chore: check service manifest compatibility with environment version

Lou1415926:
This PR checks service manifest compatibility with environment stack version. The CLI errors out and asks user to run `copilot env upgrade` when it detects that service manifest intends to use a feature that's not yet available on their environment template.


internal/pkg/cli/svc_deploy.go
```diff
@@ -364,13 +366,40 @@ func workloadManifest(in *workloadManifestInput) (interface{}, error) {
...
+	for _, f := range features {
+		if _, ok := available[f]; !ok {
+			logMsg := fmt.Sprintf(`Your manifest configuration requires your environment %q to have the feature %q available. 
+The least environment version that supports the feature is %s.`, envName, template.FriendlyFeatureName[f], deploy.LeastVersionForFeature[f])
+			currVersion, err := env.Version()
+			if err == nil {
+				logMsg += fmt.Sprintf(" Your environment is on %s.\n", currVersion)
+			}
+			logMsg += fmt.Sprintf(`Please upgrade your environment by running %s.`, color.HighlightCode(fmt.Sprintf("copilot env upgrade --name %s", envName)))
+			log.Errorln(logMsg)
```
huanjani, line 396: Love the customization of this log message!

internal/pkg/cli/svc_deploy.go
```diff
@@ -364,13 +366,40 @@ func workloadManifest(in *workloadManifestInput) (interface{}, error) {
...
+			return fmt.Errorf("environment %q is not on the versions that support the feature %q", envName, template.FriendlyFeatureName[f])
```
huanjani:
```suggestion
			return fmt.Errorf("environment %q is not on a version that supports the %q feature", envName, template.FriendlyFeatureName[f])
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -364,13 +366,40 @@ func workloadManifest(in *workloadManifestInput) (interface{}, error) {
...
 	if err := envMft.Validate(); err != nil {
 		return nil, fmt.Errorf("validate manifest against environment %s: %s", in.envName, err)
 	}
 	return envMft, nil
 }
 
+func isManifestCompatibleWithEnvironment(mft manifest.WorkloadManifest, envName string, env versionFeatureGetter) error {
+	availableFeatures, err := env.AvailableFeatures()
+	if err != nil {
+		return fmt.Errorf("get available features of the environment %s stack: %w", envName, err)
```
huanjani: nit:
```suggestion
		return fmt.Errorf("get available features of the %s environment stack: %w", envName, err)
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -364,13 +366,40 @@ func workloadManifest(in *workloadManifestInput) (interface{}, error) {
...
+	}
+	exists := struct{}{}
+	available := make(map[string]struct{})
```
huanjani, line 380: cool!

internal/pkg/cli/svc_deploy.go
```diff
@@ -364,13 +366,40 @@ func workloadManifest(in *workloadManifestInput) (interface{}, error) {
...
+	features := mft.RequiredEnvironmentFeatures()
+	for _, f := range features {
+		if _, ok := available[f]; !ok {
+			logMsg := fmt.Sprintf(`Your manifest configuration requires your environment %q to have the feature %q available. 
+The least environment version that supports the feature is %s.`, envName, template.FriendlyFeatureName[f], deploy.LeastVersionForFeature[f])
+			currVersion, err := env.Version()
+			if err == nil {
+				logMsg += fmt.Sprintf(" Your environment is on %s.\n", currVersion)
+			}
+			logMsg += fmt.Sprintf(`Please upgrade your environment by running %s.`, color.HighlightCode(fmt.Sprintf("copilot env upgrade --name %s", envName)))
```
efekarakus: Should this be the following based on our previous discussion 💭 
```suggestion
			logMsg += fmt.Sprintf(`Please upgrade your environment by running %s.`, color.HighlightCode(fmt.Sprintf("copilot env deploy --name %s", envName)))
```

internal/pkg/cli/interfaces.go
```diff
@@ -419,6 +419,11 @@ type envDescriber interface {
 	PublicCIDRBlocks() ([]string, error)
 }
 
+type versionFeatureGetter interface {
```
efekarakus: nit: feel free to ignore it - I was just trying to think of a descriptive name for the interface
```suggestion
type versionCompatibilityChecker interface {
```

internal/pkg/template/env.go
```diff
@@ -15,6 +15,22 @@ const (
 	fmtEnvCFSubTemplatePath = "environment/partials/%s.yml"
 )
 
+// Latest available feature names.
```
```suggestion
// Latest available env-controller managed feature names.
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -364,13 +366,40 @@ func workloadManifest(in *workloadManifestInput) (interface{}, error) {
...
+	exists := struct{}{}
+	available := make(map[string]struct{})
+	for _, f := range availableFeatures {
+		available[f] = exists
+	}
+
+	features := mft.RequiredEnvironmentFeatures()
+	for _, f := range features {
+		if _, ok := available[f]; !ok {
+			logMsg := fmt.Sprintf(`Your manifest configuration requires your environment %q to have the feature %q available.
```
efekarakus: I'm trying to think if we can get to a stage where if the engineers forget to update the `FriendlyFeatureName` abd `LeastVersionForFeature` maps then we still log nicely. For example, if the friendly feature name is not in the map then default to "InternalALBWorkloads".

Otherwise, an alternative I think would be adding a local integ tests that ensure that our env stack parameter and the template maps are always in parity - that would also be safe

internal/pkg/deploy/env.go
```diff
@@ -17,6 +18,17 @@ const (
 	LatestEnvTemplateVersion = "v1.10.0"
 )
 
+// LatestAvailableFeatures is a list of latest available feature names.
+var LatestAvailableFeatures = []string{template.ALBFeatureName, template.EFSFeatureName, template.NATFeatureName, template.InternalALBFeatureName}
```
efekarakus: I wonder if there is a more descriptive name here 🤔 "Latest" doesn't bring a whole lot we can drop it all together.
```go
func AvailableEnvFeatures() []string
```
nit: func is preferred over an exported global variable so that a consumer package cannot update the variable

efekarakus: Is there a benefit in having this list separate than a `template.AvailableEnvFeatures`? (basically having everything in the `template` pkg)

I'd like to limit the number of packages we need to touch if possible to enable an env feature

internal/pkg/manifest/storage.go
```diff
@@ -175,3 +175,13 @@ type AuthorizationConfig struct {
 func (a *AuthorizationConfig) IsEmpty() bool {
 	return a.IAM == nil && a.AccessPointID == nil
 }
+
+func efsFeatureRequired(s Storage) bool {
```
efekarakus: nit: was there a reason you preferred a function over a method?

dannyrandall (approved):
lgtm!

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -153,6 +153,21 @@ func (s *LoadBalancedWebService) MarshalBinary() ([]byte, error) {
 	return content.Bytes(), nil
 }
 
+// RequiredEnvironmentFeatures returns environment features that are required for this manifest.
+func (s *LoadBalancedWebService) RequiredEnvironmentFeatures() []string {
```
efekarakus, line 157: Do we need to include alias in here as well?

internal/pkg/manifest/backend_svc.go
```diff
@@ -78,6 +78,21 @@ func (s *BackendService) MarshalBinary() ([]byte, error) {
...
+		features = append(features, template.InternalALBFeatureName)
+	}
+	if aws.StringValue((*string)(s.Network.VPC.Placement.PlacementString)) == string(PrivateSubnetPlacement) {
+		features = append(features, template.NATFeatureName)
+	}
+	if efsFeatureRequired(s.Storage) {
+		features = append(features, template.EFSFeatureName)
+	}
+	return features
+}
```
efekarakus: what do you think of a flow like this:
```go
var features []string
features = append(features, s.RoutingRule.requiredEnvFeatures()...)
features = append(features, s.Network.requiredEnvFeatures()...)
features = append(features, s.Storage.requiredEnvFeatures()...)
return features
```

internal/pkg/template/env.go
```diff
@@ -15,6 +15,22 @@ const (
...
+const (
+	ALBFeatureName         = "ALBWorkloads"
+	EFSFeatureName         = "EFSWorkloads"
+	NATFeatureName         = "NATWorkloads"
+	InternalALBFeatureName = "InternalALBWorkloads"
+)
+
+// FriendlyFeatureName maps each feature name to a user-friendly feature name.
+var FriendlyFeatureName = map[string]string{
```
```suggestion
// FriendlyEnvFeatureName maps each env controller managed parameter name to a user-friendly feature name.
var FriendlyEnvFeatureName = map[string]string{
```

internal/pkg/cli/svc_deploy.go
iamhopaul123: Why do we use `struct{}` instead of `bool`?

iamhopaul123 (commented):
LGTM overall +1 to @efekarakus's comments

internal/pkg/cli/svc_deploy.go
Lou1415926: Because struct{} take 0 byte

iamhopaul123: After some investigation i think I am onboard now! But not for the memory reason which I think the benefit is so small but for the readability reason: it helps to avoid the meaningless "tri-state boolean" brought by `map[string]bool`.

internal/pkg/manifest/lb_web_svc.go
Lou1415926: `alias` is a part of `RoutingRule`!

If `!s.RoutingRule.Disabled` (i.e. it is not `http:false` in the manifest), then we will need ALB whether or not alias is being used.

internal/pkg/manifest/backend_svc.go
Lou1415926: The concern was that the same configuration may mean different things for each workload type.

For example, an empty `http:` means ALB enabled for an LBWS and the opposite to a backend service. That was why I was letting the workload struct to decide what required features they want, instead of delegating to it to the subfields.

But yeah I think perhaps common logic can be delegated like you suggested. Will update!

internal/pkg/deploy/env.go
Lou1415926: > Is there a benefit in having this list separate than a template.AvailableEnvFeatures? (basically having everything in the template pkg)

Implementation-wise, not much. I just thought `AvailableEnvFeatures` are features available to use in a deployed env stack, hence `deploy`. Plust it's close to where `LatestEnvTemplateVersion` lives.  

Though it can also be argued that `template` is the right package since `template` determines the template for a deployed env. This also comes with the benefit of easier to maintain than the other route. Changing 👍🏼

internal/pkg/cli/svc_deploy.go
Lou1415926: Added a local integ test - let me know what you think!

internal/pkg/template/env.go
```diff
@@ -15,6 +15,35 @@ const (
...
+// Latest available env-controller managed feature names.
+const (
+	ALBFeatureName         = "ALBWorkloads"
+	EFSFeatureName         = "EFSWorkloads"
+	NATFeatureName         = "NATWorkloads"
+	InternalALBFeatureName = "InternalALBWorkloads"
+)
+
+// FriendlyEnvFeatureName maps each env-controller managed parameter name to a user-friendly feature name.
+var FriendlyEnvFeatureName = map[string]string{
```
efekarakus: nit: we can move these to functions as well instead of global vars

internal/pkg/template/env_local_integration_test.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+)
+
+func TestEnv_AvailableEnvFeatures(t *testing.T) {
```
efekarakus, line 15: yayy v nice!

efekarakus (approved):

PR #3613, opened 2022-05-27
chore(customresource): add new pkg to upload cfn custom resources

efekarakus:


internal/pkg/deploy/upload/customresource/customresource.go
```diff
@@ -0,0 +1,185 @@
...
+)
+
+// CustomResource represents a CloudFormation custom resource backed by a Lambda function.
+type CustomResource struct {
+	name  string
+	files []file
+}
+
+// FunctionName is the name of the Lambda function.
+func (cr CustomResource) FunctionName() string {
```
Lou1415926: My concern is that the value can get large with the list of files, in which case it might not be the most efficient to use value receiver. Do you think we can use a pointer receiver here?

internal/pkg/deploy/upload/customresource/customresource.go
```diff
@@ -0,0 +1,185 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("upload custom resource %q: %w", cr.FunctionName(), err)
+		}
+		urls[cr.FunctionName()] = url
+	}
+	return urls, nil
+}
+
+func buildCustomResources(fs template.Reader, pathForFn map[string]string) ([]CustomResource, error) {
+	var crs []CustomResource
```
Lou1415926: tiny little nit: perhaps in this case we could allocate the capacity here since it's known!

Lou1415926 (approved):

internal/pkg/deploy/upload/customresource/customresource.go
efekarakus: Done for both!

dannyrandall (approved):

PR #3619, opened 2022-05-31
chore(cli): allow `--internal-alb-subnet` flag to be used independently

huanjani:
This tiny change reverts the logic that allows users to specify subnets for internal ALB placement without importing subnets WITH FLAGS. If the `--internal-alb-subnets` flag is used and no `--override` or `--default-config` flags are used, rather than erroring out, we should proceed to prompt for env vpc resource imports.

### DIFF
--- a/internal/pkg/cli/env_init.go
+++ b/internal/pkg/cli/env_init.go
@@ -371,7 +371,7 @@ func (o *initEnvOpts) validateCustomizedResources() error {
 	if (o.importVPC.isSet() || o.adjustVPC.isSet()) && o.defaultConfig {
 		return fmt.Errorf("cannot import or configure vpc if --%s is set", defaultConfigFlag)
 	}
-	if o.internalALBSubnets != nil && !o.importVPC.isSet() {
+	if o.internalALBSubnets != nil && (o.adjustVPC.isSet() || o.defaultConfig) {
 		log.Error(`To specify internal ALB subnet placement, you must import existing resources, including subnets.
 For default config without subnet placement specification, Copilot will place the internal ALB in the generated private subnets.`)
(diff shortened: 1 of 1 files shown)


internal/pkg/cli/env_init.go
```diff
@@ -371,7 +371,7 @@ func (o *initEnvOpts) validateCustomizedResources() error {
 	if (o.importVPC.isSet() || o.adjustVPC.isSet()) && o.defaultConfig {
 		return fmt.Errorf("cannot import or configure vpc if --%s is set", defaultConfigFlag)
 	}
-	if o.internalALBSubnets != nil && !o.importVPC.isSet() {
+	if o.internalALBSubnets != nil && (o.adjustVPC.isSet() || o.defaultConfig) {
```
efekarakus, line 374: tests don't need to be updated? 🤔

huanjani: No-- there are tests to make sure the `--internal-alb-subnets` flag _isn't_ used with `--default-config` or the `--override` flags, but when I mistakenly accepted the change that made the validation more restrictive (not allowing `--internal-alb-subnets` unless `--import` flags were used), I didn't add tests because I didn't realize that the change was bigger than just streamlining the logic.

efekarakus (approved):

internal/pkg/cli/env_init.go
iamhopaul123: Oh that was my bad. I thought internal ALB can't be used unless they import VPC

iamhopaul123: But right they could do that without the flag for sure.

dannyrandall (approved):

Issue #3469, opened 2022-04-15
Try once for Scheduled Jobs 


dmathewwws:
Hi,

Working with scheduled jobs (and really enjoying it) and appreciate the new easy turn off feature for scheduling a job.

One thing that I run into when I have a new scheduled job is that I want to deploy it and run it right away to test it works (and then I go in and change the schedule to the actual longterm schedule, once I've tested it runs as expected).

I would love an option to test a scheduled job once immediately after its deployed.

efekarakus:
Related https://github.com/aws/copilot-cli/issues/2429#issuecomment-1067394185

Thanks @dmathewwws for the feature request, it totally makes sense. I agree it would be super nice to have a command like this:
```
copilot job run
```
Similar to `task run` to invoke the step function!

kangere:
@efekarakus I would like to work on this issue, will you be taking PRs for this particular issue?

Lou1415926:
Hi, @kangere! Thank you very much! That would be awesome ❤️ 

Feel free to look at the guide [here](https://github.com/aws/copilot-cli/blob/mainline/CONTRIBUTING.md) and comment in this Issue with your design / additional thoughts. If you have any blueprint implementation plan feel free to post it here as well!

kangere:
Hi, @Lou1415926 , I propose creating a `job invoke` command that would take a name and environment, for example

`copilot job invoke --name jobname --env envronment`

This command would have to be called after a call to `job deploy` since the job infrastructure needs to have been provisioned first

Lou1415926:
@kangere Thank you very much for the proposal!

> This command would have to be called after a call to job deploy since the job infrastructure needs to have been provisioned first

This totally makes sense to me!

A similar example would be `copilot svc log --name api --env test`. To see the log, `api` and `test` must have been an existing service/environment, respectively.  We validate these assumptions [here](https://github.com/aws/copilot-cli/blob/5cd9981a26c0a0eb1b1aa02c4f9f3ac6bfe3ee21/internal/pkg/cli/svc_logs.go#L210) and [here](https://github.com/aws/copilot-cli/blob/5cd9981a26c0a0eb1b1aa02c4f9f3ac6bfe3ee21/internal/pkg/cli/svc_logs.go#L216).

We might want to do the same for this command!

---

>  I propose creating a `job invoke` command

The flags `--name` and `--env` sounds great to me. I'd say we can also add the app flag `--app`. 

[...]

PR #3620, opened 2022-06-01
chore(task): access task run command's exit code

paragbhingre:
This PR addresses the issue #2525 

This PR adds an ability to fail the `copilot task run --follow` command with non zero exit code when essential container of the task fails.  
Screen Shot 2022-05-31 at 5 57 30 PM


internal/pkg/cli/interfaces.go
```diff
@@ -384,6 +384,7 @@ type taskStackManager interface {
 
 type taskRunner interface {
 	Run() ([]*task.Task, error)
+	CheckNonZeroExitCode(tasks []*task.Task) error
```
iamhopaul123:
```suggestion
	CheckNonZeroExitCode([]*task.Task) error
```
nit

internal/pkg/cli/task_run.go
```diff
@@ -681,6 +681,9 @@ Did you tag your secrets with the "copilot-application" and "copilot-environment
 		if err := o.displayLogStream(); err != nil {
 			return err
 		}
+		if nonZeroExitCodeErr := o.runner.CheckNonZeroExitCode(tasks); nonZeroExitCodeErr != nil {
```
iamhopaul123:
```suggestion
		if err := o.runner.CheckNonZeroExitCode(tasks); err != nil {
```
it is ok to just use `err` here because its scope is very small (only for this condition)

internal/pkg/task/config_runner.go
```diff
@@ -104,3 +107,65 @@ func (r *ConfigRunner) validateDependencies() error {
 
 	return nil
 }
+
+// CheckNonZeroExitCode returns the status of the containers part of the given tasks.
+func (r *ConfigRunner) CheckNonZeroExitCode(tasks []*Task) error {
+	var essentialContainers []string
```
iamhopaul123: We could do very similar logic as https://github.com/aws/copilot-cli/pull/3620/files#diff-f4d4181f6b85ffcf797c84472ae4742e03524f9e77606345fe3c77c2629b9b9dR126 !

internal/pkg/task/config_runner.go
```diff
@@ -104,3 +107,65 @@ func (r *ConfigRunner) validateDependencies() error {
...
+	taskARNs := make([]string, len(tasks))
+	for idx, task := range tasks {
+		taskARNs[idx] = task.TaskARN
+	}
+	describedTasks, describeErr := r.Starter.DescribeTasks(r.Cluster, taskARNs)
+	if describeErr != nil {
+		return describeErr
+	}
+	// Today we only use one container inside a stand-alone task, shall we keep the looping logic
+	// active for multiple containers or only add this logic when we have multiple containers inside a stand-alone task in the future.
```
iamhopaul123: We might not need these comments 🤔

internal/pkg/task/config_runner.go
```diff
@@ -104,3 +107,65 @@ func (r *ConfigRunner) validateDependencies() error {
...
+					taskID,
+					aws.Int64Value(container.ExitCode)))
+			}
+		}
+	}
+	return nil
+}
+
+// contains checks if a string is present in a slice
+func contains(s []string, str string) bool {
```
iamhopaul123: Can we use [`map[string]struct{}`](https://github.com/aws/copilot-cli/pull/3612/files#diff-887bb70e74d71d9ecbfe078d768a70bdd2f02043501ba27c8b254ac3a102e799R380) to avoid for loop?

internal/pkg/task/config_runner.go
dannyrandall: i think this could be simpler if this was `map[string]bool` instead (maybe like `isEssential`?)

internal/pkg/task/config_runner.go
```diff
@@ -104,3 +107,65 @@ func (r *ConfigRunner) validateDependencies() error {
...
+	}
+
+	taskARNs := make([]string, len(tasks))
+	for idx, task := range tasks {
+		taskARNs[idx] = task.TaskARN
+	}
+	describedTasks, describeErr := r.Starter.DescribeTasks(r.Cluster, taskARNs)
+	if describeErr != nil {
+		return describeErr
+	}
```
iamhopaul123: Seems like we can make this block running in parallel with https://github.com/aws/copilot-cli/pull/3620/files#diff-f4d4181f6b85ffcf797c84472ae4742e03524f9e77606345fe3c77c2629b9b9dR114-R124

internal/pkg/task/config_runner.go
```diff
@@ -104,3 +107,65 @@ func (r *ConfigRunner) validateDependencies() error {
...
+	for _, v := range s {
+		if v == str {
+			return true
+		}
+	}
+	return false
+}
+
+// HumanString returns a friendly colorized summary of the container description.
+func HumanString(containerName string, taskName string, exitCode int64) string {
```
iamhopaul123: We could define a custom error type instead. For example: https://github.com/aws/copilot-cli/blob/507747027c5ff3007b3c99e1f9c33aad74a63566/internal/pkg/manifest/errors.go#L57

internal/pkg/task/env_runner.go
```diff
@@ -156,3 +157,48 @@ func containsString(s []string, search string) bool {
 	}
 	return false
 }
+
+// CheckNonZeroExitCode returns the status of the containers part of the given tasks.
+func (r *EnvRunner) CheckNonZeroExitCode(tasks []*Task) error {
```
dannyrandall, line 164: wonder if there'd be a good way to call a common function between this one and the one in `config_runner`🤔, it looks pretty similar

iamhopaul123, line 164: I feel this function should be placed in the `ecs` package instead. Can we move this function to https://github.com/aws/copilot-cli/tree/507747027c5ff3007b3c99e1f9c33aad74a63566/internal/pkg/ecs? And then we can call it in both config runner and env runner to reduce code duplication?

dannyrandall (commented):
it would be cool if the process (`copilot task run`) exit code could match the exit code of the task instead of just printing it, might make scripting easier 🤔

internal/pkg/task/config_runner.go
paragbhingre: This comment is for the team to decide if we should keep the looping logic even when we know that there is always going to be only single container in the task? I will remove this comment before merging this PR

internal/pkg/task/config_runner.go
paragbhingre: Do you think we need to parallelize these two blocks because there is only one call to DescribeTask and TaskDefinition is happening and for loops are also going for max 5 times as there are not much stand-alone tasks that customers deploy and task definition will also have only one container per task. let me know what do you think about it.

iamhopaul123: That's fair. I think we can leave it as is

internal/pkg/task/config_runner.go
paragbhingre: We have used `map[string]bool` for instead of `essentialContainers []string` so we do not need this function anymore.

paragbhingre:
Addressed all the comments except this [one](https://github.com/aws/copilot-cli/pull/3620#discussion_r887315081). Working on it.

internal/pkg/cli/task_run.go
```diff
@@ -681,6 +681,10 @@ Did you tag your secrets with the "copilot-application" and "copilot-environment
 		if err := o.displayLogStream(); err != nil {
 			return err
 		}
+		if exitCode, err := o.runner.CheckNonZeroExitCode(tasks); err != nil {
+			fmt.Println(err)
```
Lou1415926: Do we intend to print the error out to users? If so what do you think of `log.Error`?

internal/pkg/task/config_runner.go
```diff
@@ -104,3 +104,43 @@ func (r *ConfigRunner) validateDependencies() error {
 
 	return nil
 }
+
+// CheckNonZeroExitCode returns the status of the containers part of the given tasks.
+func (r *ConfigRunner) CheckNonZeroExitCode(tasks []*Task) (int, error) {
+	essentialContainers := make(map[string]bool)
+	taskDefName := fmt.Sprintf("copilot-%s", r.GroupName)
+	taskDefinition, err := r.Starter.TaskDefinition(taskDefName)
```
Lou1415926:
```suggestion
	taskDefinition, err := r.Starter.TaskDefinition(taskFamilyName(r.GroupName))
```

internal/pkg/cli/task_run.go
```diff
@@ -681,6 +681,10 @@ Did you tag your secrets with the "copilot-application" and "copilot-environment
...
+			os.Exit(exitCode)
```
Lou1415926: This probably doesn't work super nicely with the current implementation of `CheckNonZeroExitCode`. It could result in a state where an `err` is printed out, while the program exits with `0`, which wouldn't make sense.

Perhaps we can change the method signature for `CheckNonZeroExitCode` to
```
func (r *ConfigRunner) CheckNonZeroExitCode(*int, err)
```
We return `nil, err` when we have an error, return `&exitcode, err`  or `&exitcode, nil` when we have the exit code.

Lou1415926: I also keep debating on whether we should `os.Exit` with the task error code. While I see the potential benefit it has for script users, I'm not sure if it's a faithful representation of the program execution.

Would love to see what folks think!

internal/pkg/task/config_runner.go
```diff
@@ -104,3 +104,43 @@ func (r *ConfigRunner) validateDependencies() error {
...
+	if err != nil {
+		return 0, fmt.Errorf("get task definition %s of a task", taskDefName)
```
Lou1415926:
```suggestion
		return 0, fmt.Errorf("get task definition %s: %w", taskDefName, err)
```

Lou1415926 (commented):
Haven't look at the whole PR but I have a few comment!

Also, +1 to https://github.com/aws/copilot-cli/pull/3620#discussion_r887314234

internal/pkg/cli/task_run.go
paragbhingre: Yes we are planning on displaying error to the customer instead of just silently failing the task. `log.error` makes sense.

internal/pkg/cli/task_run.go
paragbhingre: One thing we can do here is to exit only when the exit code is greater than `0`

internal/pkg/task/env_runner.go
paragbhingre: I have moved the common functionality to the `ecs` package

internal/pkg/task/config_runner.go
```diff
@@ -36,6 +35,9 @@ type ConfigRunner struct {
 	// Must not be nil if using default subnets.
 	VPCGetter VPCGetter
 
+	// Figures non-zero exit code of the task
+	NonZeroExitCodeGetter NonZeroExitCodeGetter
```
Lou1415926, line 39: Where is this initialized 🤔 ? Does this not produce a nil-pointer panic?

internal/pkg/cli/task_run.go
dannyrandall: I'm in favor of including the exit code from the task in `task run` - `docker run` [does the same thing](https://docs.docker.com/engine/reference/run/#exit-status). However, `kubectl run` does only [return the exit code if a some flags are set](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands) (`-i --restart=Never`), and I'd be fine with an approach like a flag that we could add in a separate PR. 

Either way, I think that the exit code should be set in main.go, not here. We could return a custom error type with an exit code in it or something like that.

Lou1415926: +1 to that exit code should be wrapped inside of the custom error

paragbhingre: As @dannyrandall suggested we are now displaying error message at main.go level and also exiting the process with the non zero exit code from there itself. 

Also for your concern @Lou1415926 re: `func (r *ConfigRunner) CheckNonZeroExitCode(*int, err)` function signature. So we have implemented `ExitCode()` function on `ErrExitCode` which we call in main.go directly so no need to keep passing this exit code param from bottom to up until main.go. So we have removed that int from the signature now. The new [commit](https://github.com/aws/copilot-cli/pull/3620/commits/f158d30d8675137365c334f4528f9fd18ff7df6f#diff-92e031f8d969d8ae755c21b2887e498847e62ecab3c96303dcf3696499aa48a0R23-R44) will have these changes.

cmd/copilot/main.go
```diff
@@ -20,6 +20,10 @@ type actionRecommender interface {
 	RecommendActions() string
 }
 
+type exitCodeError interface {
```
Lou1415926, line 23: nice!

internal/pkg/cli/task_run.go
```diff
@@ -260,10 +260,11 @@ func (o *runTaskOpts) configureRunner() (taskRunner, error) {
...
 
-			VPCGetter:            vpcGetter,
-			ClusterGetter:        ecs.New(o.sess),
-			Starter:              ecsService,
-			EnvironmentDescriber: d,
+			VPCGetter:             vpcGetter,
+			ClusterGetter:         ecs.New(o.sess),
+			Starter:               ecsService,
+			EnvironmentDescriber:  d,
+			NonZeroExitCodeGetter: ecs.New(o.sess),
```
Lou1415926: We can reuse the same client for both `ClusterGetter` and `NonZeroExitCodeChecker`:

```suggestion
   ecsClient := ecs.New(o.sess)
   ...
			ClusterGetter:    ecsClient,
			Starter:               ecsService,
			EnvironmentDescriber:  d,
			NonZeroExitCodeGetter:ecsClient,
```

internal/pkg/ecs/ecs.go
```diff
@@ -465,3 +466,37 @@ func (c Client) stateMachineARN(app, env, job string) (string, error) {
 	}
 	return stateMachineARN, nil
 }
+
+// NonZeroExitCode returns the non zero exit code of the task if any.
+func (c Client) NonZeroExitCode(taskARNs []string, groupName string, cluster string) error {
```
Lou1415926: nit:
We could probably remove `groupName` from parameters. We could get the `taskDefinitionArn` from the response of `DescribeTasks`, and call `DescribeTasks` with the arn. This could also prevent the mistaken usage where the `groupName` doesn't really point to the real task definition that the `taskARNs` are using.

What do you think of this:
```go
// HasNonZeroExitCode returns an error if at least one of the tasks exited with a non-zero exit code. It assumes that all tasks are built on the same task definition.
func (c Client) HasNonZeroExitCode(taskARNs []string, cluster string) error {
	tasks, err := c.ecsClient.DescribeTasks(cluster, taskARNs)
	if err != nil {
		return err
	}

	if len(tasks) == 0 {
		return errors.New("describe tasks: none of the tasks %s exist", strings.Join(taskARNs, ","))
	}

	taskDefinitonARN := aws.StringValue(tasks[0].TaskDefinitionArn)
	taskDefinition, err := c.ecsClient.TaskDefinition(taskDefinitonARN)
	if err != nil {
		return fmt.Errorf("get task definition %s: %w", taskDefinitonARN, err)
	}

	isContainerEssential := make(map[string]bool)
	for _, container := range taskDefinition.ContainerDefinitions {
		isContainerEssential[aws.StringValue(container.Name)] = aws.BoolValue(container.Essential)	}
	...
}
```

Lou1415926: Just to put my thoughts ⬇️  here 💭 

---

I think oit could be probably argued that this method could be a method of [`ECS`](https://github.com/aws/copilot-cli/blob/93f3ad4c87aa805a9aa96057c09d9f8180d55690/internal/pkg/aws/ecs/ecs.go#L48) instead. What the method does is basically: given a list of task ARNs, check whether at least one of them exited with a non-zero exit code. This isn't Copilot-specific, you could pass in any list of task ARNs and do the same.

On the other hand, It could be argue that the assumption of `all tasks are built on the same task definition` is a Copilot-ty assumption.  Therefore, I am not against of keeping it in the `ecs` package.

paragbhingre: Yeah agree to your recommendation, we can avoid duplicate information from passing to HasNonZeroExitCode and make the code cleaner.

internal/pkg/ecs/ecs.go
```diff
@@ -467,27 +467,31 @@ func (c Client) stateMachineARN(app, env, job string) (string, error) {
...
-func (c Client) NonZeroExitCode(taskARNs []string, groupName string, cluster string) error {
-	essentialContainers := make(map[string]bool)
-	taskDefName := fmt.Sprintf("copilot-%s", groupName)
-	taskDefinition, err := c.ecsClient.TaskDefinition(taskDefName)
+// HasNonZeroExitCode returns an error if at least one of the tasks exited with a non-zero exit code. It assumes that all tasks are built on the same task definition.
+func (c Client) HasNonZeroExitCode(taskARNs []string, cluster string) error {
+	tasks, err := c.ecsClient.DescribeTasks(cluster, taskARNs)
 	if err != nil {
-		return fmt.Errorf("get task definition %s: %w", taskDefName, err)
+		return err
```
Lou1415926: I think we can probably wrap it here:
```suggestion
		return fmt.Errorf("describe tasks %s: err", taskARNs, err)
```

internal/pkg/task/config_runner.go
```diff
@@ -36,6 +35,9 @@ type ConfigRunner struct {
 	// Must not be nil if using default subnets.
 	VPCGetter VPCGetter
 
+	// Figures non-zero exit code of the task
```
Lou1415926: nit:
```suggestion
	// Figures non-zero exit code of the task.
```

internal/pkg/ecs/errors.go
```diff
@@ -12,3 +12,17 @@ type ErrMultipleContainersInTaskDef struct {
...
 }
+
+type ErrExitCode struct {
+	containerName string
+	taskId        string
+	exitCode      int
+}
+
+func (e *ErrExitCode) Error() string {
+	return fmt.Sprintf("Container %s in task %s exited with status code %d", e.containerName, e.taskId, e.exitCode)
```
Lou1415926:
```suggestion
	return fmt.Sprintf("container %s in task %s exited with status code %d", e.containerName, e.taskId, e.exitCode)
```

internal/pkg/task/env_runner.go
```diff
@@ -49,6 +49,9 @@ type EnvRunner struct {
 	ClusterGetter        ClusterGetter
 	Starter              Runner
 	EnvironmentDescriber EnvironmentDescriber
+
+	// Figures non-zero exit code of the task
```
Lou1415926, line 53: nit:
```suggestion
	// Figures non-zero exit code of the task.
```

internal/pkg/task/config_runner.go
```diff
@@ -104,3 +106,12 @@ func (r *ConfigRunner) validateDependencies() error {
 
 	return nil
 }
+
+// CheckNonZeroExitCode returns the status of the containers part of the given tasks.
+func (r *ConfigRunner) CheckNonZeroExitCode(tasks []*Task) error {
```
Lou1415926, line 111: Leaving the comment here just for future reference, no action item for ya!

At a latter point we should refactor this package such that:
1. Instead of returning tasks, `Run()` should return cache the tasks.
2. This function will no longer need to have `tasks` as the parameter. It should use the cache `tasks` instead.
3. Same for `cluster` for `EnvRunner`. We should cache it, instead of getting cluster twice, once in `Run()`, once in `CheckExitCode`.

The above mentioned refactoring is related to: https://github.com/aws/copilot-cli/issues/3469#issuecomment-1124393633

Lou1415926 (approved):

internal/pkg/ecs/ecs.go
```diff
@@ -465,3 +466,41 @@ func (c Client) stateMachineARN(app, env, job string) (string, error) {
...
+
+// HasNonZeroExitCode returns an error if at least one of the tasks exited with a non-zero exit code. It assumes that all tasks are built on the same task definition.
+func (c Client) HasNonZeroExitCode(taskARNs []string, cluster string) error {
+	tasks, err := c.ecsClient.DescribeTasks(cluster, taskARNs)
+	if err != nil {
+		return fmt.Errorf("describe tasks %s: %w", taskARNs, err)
+	}
+
+	if len(tasks) == 0 {
+		return fmt.Errorf("describe tasks: none of the tasks %s exist", strings.Join(taskARNs, ","))
```
iamhopaul123: nit: this is not an error returned by `DescribeTasks`.
```suggestion
		return fmt.Errorf("cannot find tasks %s", strings.Join(taskARNs, ", "))
```

internal/pkg/ecs/ecs.go
```diff
@@ -465,3 +466,41 @@ func (c Client) stateMachineARN(app, env, job string) (string, error) {
...
+	}
+
+	isContainerEssential := make(map[string]bool)
+	for _, container := range taskDefinition.ContainerDefinitions {
+		isContainerEssential[aws.StringValue(container.Name)] = aws.BoolValue(container.Essential)
+	}
+
+	for _, describedTask := range tasks {
+		for _, container := range describedTask.Containers {
+			if isContainerEssential[aws.StringValue(container.Name)] && aws.Int64Value(container.ExitCode) > 0 {
```
iamhopaul123:
```suggestion
			if isContainerEssential[aws.StringValue(container.Name)] && aws.Int64Value(container.ExitCode) != 0 {
```

internal/pkg/ecs/errors.go
```diff
@@ -12,3 +12,17 @@ type ErrMultipleContainersInTaskDef struct {
 func (e *ErrMultipleContainersInTaskDef) Error() string {
 	return fmt.Sprintf("found more than one container in task definition: %s", e.taskDefIdentifier)
 }
+
+type ErrExitCode struct {
```
iamhopaul123, line 17: Does this need to be public? If so could you add comment to the public struct?

iamhopaul123 (approved):

internal/pkg/ecs/errors.go
paragbhingre: Added a comment

PR #3622, opened 2022-06-01
chore: hosted zone fixes

dannyrandall:

### DIFF
--- a/internal/pkg/template/templates/environment/cf.yml
+++ b/internal/pkg/template/templates/environment/cf.yml
@@ -523,11 +523,6 @@ Outputs:
     Value: !GetAtt InternalLoadBalancer.DNSName
     Export:
       Name: !Sub ${AWS::StackName}-InternalLoadBalancerDNS
-  InternalLoadBalancerFullName:
-    Condition: CreateInternalALB
-    Value: !GetAtt InternalLoadBalancer.LoadBalancerFullName
-    Export:
-      Name: !Sub ${AWS::StackName}-InternalLoadBalancerFullName
   InternalLoadBalancerHostedZone:
     Condition: CreateInternalALB
     Value: !GetAtt InternalLoadBalancer.CanonicalHostedZoneID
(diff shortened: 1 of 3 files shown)


internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
```diff
@@ -33,7 +33,7 @@ LoadBalancerDNSAlias:
       - Name: {{$alias}}
         Type: A
         AliasTarget:
-          {{- if eq .WorkloadType "Backend Service"}}
+          {{- if eq $.WorkloadType "Backend Service"}}
```
iamhopaul123, line 36: 🤝👴

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -523,6 +523,16 @@ Outputs:
     Value: !GetAtt InternalLoadBalancer.DNSName
     Export:
       Name: !Sub ${AWS::StackName}-InternalLoadBalancerDNS
+  InternalLoadBalancerFullName:
```
efekarakus, line 526: Are these exports planning to be used somewhere? 🤔

iamhopaul123 (approved):

internal/pkg/template/templates/environment/cf.yml
dannyrandall: half of them are 😄 `InternalLoadBalancerHostedZone` is already used [here](https://github.com/aws/copilot-cli/pull/3622/files#diff-ebcd538b6b6f76a2a339b6bcf0d0dbc47b1e1be7bbbcf7ba01c8338e48fefc4dR37) and `InternalLoadBalancerFullName` will be needed if we support autoscaling by requests (see [here](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/workloads/services/lb-web/cf.yml#L100)). I just added it today so that people won't have to upgrade their env when (if?) we add that feature, but I'm fine to remove that one.

dannyrandall: chatted with @iamhopaul123 offline, we'll remove `InternalLoadBalancerFullName` until we need it

dannyrandall: nvm lol

efekarakus (approved):

PR #3623, opened 2022-06-01
test: enable running a single e2e test with `e2e-dryrun` target

efekarakus:
You can now run a single e2e test in your AWS account with:
```
make e2e-dryrun test=<the test suit name>
```
For example, `make e2e-dryrun test=multi-env-app`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,6 +120,16 @@ e2e: build-e2e
 	@echo "Running E2E Tests" &&\
 	docker run --privileged -v ${HOME}/.aws:/home/.aws -e "HOME=/home" copilot/e2e:latest
 
+.PHONY: e2e-dryrun
+e2e-dryrun: build # Sample command "make e2e-dryrun test=multi-env-app" to run the test suit under "e2e/multi-env-app"
+	@echo "Install ginkgo"
+	go install github.com/onsi/ginkgo/ginkgo@latest
+	@echo "Setup credentials"
+	./scripts/e2e-dryrun-creds.sh
+	@echo "Run the $(test) test"
+	cd e2e/$(test) && DRYRUN=true ginkgo -v -r
+	cd -
+
 .PHONY: tools
 tools:
 	@echo "Installing custom resource dependencies" &&\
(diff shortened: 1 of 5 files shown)


scripts/e2e-dryrun-creds.sh
```diff
@@ -0,0 +1,19 @@
+#! /usr/bin/env python3
+import subprocess
+
+cmd = subprocess.run(['aws', 'configure', 'list-profiles'], stdout=subprocess.PIPE)
+profiles = cmd.stdout.decode('utf-8').split('\n')
+
+for e2eprofile in ['e2etestenv', 'e2eprodenv']:
+  if e2eprofile in profiles:
+    continue
+  print(f'Profile [{e2eprofile}] is required but does not exist. Copying your [default] access key id and secret key...')
```
iamhopaul123, line 10: I wonder if we should just error out and prompt them to set up any missing profile and use different accounts/regions as possible for `e2etestenv` and `e2eprodenv`. Otherwise, even if they run the multiple env suite successfully within the same account/region, it doesn't mean multi-env scenario is ok 🤔

Makefile
```diff
@@ -120,6 +120,16 @@ e2e: build-e2e
 	@echo "Running E2E Tests" &&\
 	docker run --privileged -v ${HOME}/.aws:/home/.aws -e "HOME=/home" copilot/e2e:latest
 
+.PHONY: e2e-dryrun
```
iamhopaul123, line 123: awesome!

scripts/e2e-dryrun-creds.sh
efekarakus: I kept debating about this too! 

I'm not super sure because I want the dryrun to be easy to run and cover the majority of the test. Having to create 3 AWS accounts to do `make e2e-dryrun` seems cumbersome.  
If the engineer already has the `e2etestenv` and `e2eprodenv` profiles in separate accounts then we won't override those creds and the envs will be created in separate accounts 💭 only the e2e test won't check if they in separate accounts.

I'm down to change it though if folks prefer the prompt route

iamhopaul123 (approved):

Lou1415926 (approved):
super nice! quality of life ⭐

PR #3624, opened 2022-06-01
chore: autoscaling by `requests` or `response_time` for backend services 

dannyrandall:
...only when an internal ALB is enabled:blush:


internal/pkg/manifest/validate.go
```diff
@@ -209,6 +209,12 @@ func (b BackendServiceConfig) Validate() error {
 	if err = b.RoutingRule.Validate(); err != nil {
 		return fmt.Errorf(`validate "http": %w`, err)
 	}
+	if b.RoutingRule.IsEmpty() && (b.Count.AdvancedCount.Requests != nil || b.Count.AdvancedCount.ResponseTime != nil) {
```
iamhopaul123, line 212: :+1:

internal/pkg/template/templates/workloads/partials/cf/autoscaling.yml
```diff
@@ -296,4 +296,56 @@ AutoScalingPolicy{{logicalIDSafe $topic.Service}}{{logicalIDSafe $topic.Name}}Ev
 {{- end }}{{/* range $topic := .Subscribe.Topics */}}
 {{- end }}{{/* if .Subscribe */}}
 
-{{- end }}{{/* if .Autoscaling.QueueDelay */}}
\ No newline at end of file
+{{- end }}{{/* if .Autoscaling.QueueDelay */}}
+
+{{- if .Autoscaling.Requests}}
```
iamhopaul123, line 301: How about for worker service?

efekarakus: I wonder if we need something like this to make our validations more specific:

```go
type backendTaskConfigValidator TaskConfig
type workerTaskConfigValidator TaskConfig

type backendCountValidator Count 
type workerCountValidator Count

type (v backendCountValidator) Validate() error {
  if !v.QueueScaling().IsEmpty() {
      return errors.New("boom")
  }
  ...
}

if (v workerCountValidator) Validate() error {
   if v.Requests != nil {
       return errors.New("boom2")
   }
}
```

and to use it:
```go
func (b BackendServiceConfig) Validate() error {
  ...
  if err := backendTaskConfigValidator(b.TaskConfig).Validate(); err != nil { ... }
  ...
}
```

I understand that it's a massive change though, and I'm okay with addressing it as a follow-up PR.

Today, the same problem I think exists for LBWS and BackendServices if users specify `queue_delay` for example.

efekarakus (commented):
LGTM - let me know what y'all think around validation

internal/pkg/template/templates/workloads/partials/cf/autoscaling.yml
```diff
@@ -296,4 +296,56 @@ AutoScalingPolicy{{logicalIDSafe $topic.Service}}{{logicalIDSafe $topic.Name}}Ev
...
+    PolicyName: !Join ['-', [!Ref WorkloadName, ALBSumRequestCountPerTarget, ScalingPolicy]]
+    PolicyType: TargetTrackingScaling
+    ScalingTargetId: !Ref AutoScalingTarget
+    TargetTrackingScalingPolicyConfiguration:
+      CustomizedMetricSpecification:
+        Dimensions:
+          - Name: LoadBalancer
+            {{- if eq .WorkloadType "Backend Service"}}
+            Value: !GetAtt EnvControllerAction.InternalLoadBalancerFullName
+            {{- else}}
```
huanjani, line 314: Just curious why we didn't have `requests`/`response_time` for Worker Services previously...?

efekarakus: They are not behind load balancers - requests and response_time are metrics that are emitted by ALB - worker services receive requests from queues ✌️

huanjani: So then this should be an `{{- if...}}` not an `{{- else}}` because the worker service cfn template includes this partial?

internal/pkg/template/templates/workloads/partials/cf/autoscaling.yml
dannyrandall: Chatted offline - we'll wait to make this refactor in a follow up PR.

internal/pkg/template/templates/workloads/partials/cf/autoscaling.yml
dannyrandall: Those cases will never occur - `Validate()` will error out if `requests` or `response_time` are set on a Worker Service (see [here](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/manifest/validate.go#L832-L879)) - similar to how `queue_delay` is only valid for worker services, but not explicitly gated by type in the cfn template.

huanjani: Gotcha, gotcha. Thanks!

huanjani (approved):
LGTM! :shipit:

iamhopaul123 (approved):

PR #3626, opened 2022-06-02
chore: allow non-http svcs in env w/ cert(s)

huanjani:
In an environment with imported certs, not all backend services should have to be associated with the internal load balancer. This limits the alias-cert validation to backend services with `http` specified.

### DIFF
--- a/internal/pkg/cli/deploy/deploy.go
+++ b/internal/pkg/cli/deploy/deploy.go
@@ -944,7 +944,6 @@ func (d *backendSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (
 	if err != nil {
 		return nil, err
 	}
-
 	if err := d.validateALBRuntime(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


internal/pkg/cli/deploy/deploy.go
```diff
@@ -1211,6 +1210,8 @@ func validateAppVersionForAlias(appName string, appVersionGetter versionGetter)
 
 func (d *backendSvcDeployer) validateALBRuntime() error {
 	switch {
+	case d.backendMft.RoutingRule.IsEmpty():
+		return nil
```
Lou1415926: What do you think of:
```go
if d.backendMft.RoutingRule.IsEmpty() {
   return nil
}
switch {
...
}
```
leaving `switch` to handle enumerating the 4 possible cases for alias x imported cert  only.

Lou1415926 (approved):

efekarakus (approved):
LGTM, applied do-not-merge to address Wanxian's feedback

Issue #3165, opened 2022-01-04
Configuring scaling cooldown periods in the manifest


efekarakus:
### Description

Discussed in https://github.com/aws/copilot-cli/discussions/3161

### Proposal

If the same values need to be applied to all scaling policies:
```yaml
count:
  range: 1-3
  cooldown:   # Optional.
    in: 3m        # Defaults to 2m
    out: 1m30s    # Defaults to 1m
  cpu_percentage: 40
  memory_percentage: 60
```
The `cooldown` field can only be specified if `range` is specified.

If individual policies need to be configured:
```yaml
count:
  range: 1-3
  cpu_percentage:
     value: 40   
     cooldown:
       in: 3m          
  memory_percentage: 60
```
The individual scaling in and out cooldowns take precedence over global ones

efekarakus:
This feature is now out in v1.20 🚀 
Release notes: https://github.com/aws/copilot-cli/releases/tag/v1.20.0 !
Blog post: https://aws.github.io/copilot-cli/blogs/release-v120/

PR #3629, opened 2022-06-03
chore: add scale-in and scale-out configuration to svc manifest

CaptainCarpensir:
The full functionality is unfinished, but this allows copilot-cli to read autoscaling cooldown values from the service manifest. Currently the values are stored in the program, but are not written into the cloudformation template, and so will not be reflected in the deployed service.

Addresses #3165


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -197,10 +197,18 @@ func convertAutoscaling(a manifest.AdvancedCount) (*template.AutoscalingOpts, er
 		MaxCapacity: &max,
 	}
 	if a.CPU != nil {
-		autoscalingOpts.CPU = aws.Float64(float64(*a.CPU))
+		if a.CPU.Value != nil {
+			autoscalingOpts.CPU = aws.Float64(float64(*a.CPU.Value))
+		} else {
+			autoscalingOpts.CPU = aws.Float64(float64(*a.CPU.ScaledResource.Value))
+		}
```
iamhopaul123: Maybe it'll be safer to
```suggestion
		if a.CPU.Value != nil {
			autoscalingOpts.CPU = aws.Float64(float64(*a.CPU.Value))
		}
		if a.CPU.ScaledResource.Value != nil {
			autoscalingOpts.CPU = aws.Float64(float64(*a.CPU.ScaledResource.Value))
		}
```

internal/pkg/manifest/svc.go
```diff
@@ -172,23 +196,52 @@ func (c *Count) Desired() (*int, error) {
 // Percentage represents a valid percentage integer ranging from 0 to 100.
 type Percentage int
 
+// Resource represents a resource that has autoscaling configurations.
+type Resource struct {
```
iamhopaul123: Can we make this struct name to be more expressive? How about `AdvancedScalingConfigOrPercentage`?

internal/pkg/manifest/svc.go
```diff
@@ -172,23 +196,52 @@ func (c *Count) Desired() (*int, error) {
...
+	Value          *Percentage
+	ScaledResource AdvancedResource // mutually exclusive with Value
+}
+
+// AdvancedResource represents the allocation of a resource implementing scaling cooldown.
+type AdvancedResource struct {
```
iamhopaul123:
```suggestion
// AdvancedScalingConfig represents advanced configurable options for a scaling policy.
type AdvancedScalingConfig struct {
```

internal/pkg/manifest/svc.go
```diff
@@ -204,11 +257,11 @@ func (a *AdvancedCount) hasAutoscaling() bool {
...
 	switch a.workloadType {
 	case LoadBalancedWebServiceType:
-		return []string{"cpu_percentage", "memory_percentage", "requests", "response_time"}
+		return []string{"cpu_percentage", "memory_percentage", "requests", "response_time", "cooldown"}
 	case BackendServiceType:
-		return []string{"cpu_percentage", "memory_percentage"}
+		return []string{"cpu_percentage", "memory_percentage", "cooldown"}
 	case WorkerServiceType:
-		return []string{"cpu_percentage", "memory_percentage", "queue_delay"}
+		return []string{"cpu_percentage", "memory_percentage", "queue_delay", "cooldown"}
```
iamhopaul123: I don't think we should add `cooldown` here. Same for `hasScalingFieldsSet`. Reason is `cooldown` does not enable any scaling policy and users are not supposed to just set `cooldown` without setting any auto scaling policy. Instead, in `validate.go` we probably want to make sure users set at least one of the `validScalingFields` if they set `cooldown`.

internal/pkg/manifest/svc.go
```diff
@@ -217,18 +270,20 @@ func (a *AdvancedCount) validScalingFields() []string {
...
 	default:
-		return a.CPU != nil || a.Memory != nil || a.Requests != nil || a.ResponseTime != nil || !a.QueueScaling.IsEmpty()
+		return a.CPU != nil || a.Memory != nil || a.Requests != nil ||
+			a.ResponseTime != nil || !a.QueueScaling.IsEmpty() || a.Cooldown != nil
 	}
 }
 
 func (a *AdvancedCount) unsetAutoscaling() {
 	a.Range = Range{}
+	a.Cooldown = nil
```
iamhopaul123: This is a good one!

internal/pkg/manifest/svc_test.go
```diff
@@ -416,16 +436,20 @@ func TestCount_UnmarshalYAML(t *testing.T) {
...
 			inContent: []byte(`count:
   range: 1-10
-  cpu_percentage: 70
-  memory_percentage: 80
+  cpu_percentage:
+    value: 70
+    cooldown:
+      in: 1m
+      out: 1m
+  memory_percentage: 70
```
iamhopaul123, line 444: awesome!

internal/pkg/manifest/validate.go
```diff
@@ -850,6 +850,28 @@ func (a AdvancedCount) Validate() error {
 		}
 	}
 
+	// Validate combinations with cooldown
+	if a.CPU != nil {
```
iamhopaul123: Nice validation. Also we probably can simplify the logic by checking if `a.cooldown` is set first. If it is not set then we don't even need to go through each advanced scaling policy setting.

internal/pkg/manifest/validate.go
```diff
@@ -875,6 +897,33 @@ func (p Percentage) Validate() error {
...
+	if r.IsEmpty() {
+		return nil
+	}
+	if err := r.Value.Validate(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Validation is a no-op for Cooldown
```
iamhopaul123: nit
```suggestion
// Validation is a no-op for Cooldown.
```

iamhopaul123 (commented):
Mostly LGTM. Doesn't look like a PR from a first time contributor with few golang experience. Good job!

internal/pkg/manifest/svc.go
```diff
@@ -172,23 +196,52 @@ func (c *Count) Desired() (*int, error) {
...
+}
+
 // AdvancedCount represents the configurable options for Auto Scaling as well as
 // Capacity configuration (spot).
 type AdvancedCount struct {
 	Spot         *int           `yaml:"spot"` // mutually exclusive with other fields
 	Range        Range          `yaml:"range"`
-	CPU          *Percentage    `yaml:"cpu_percentage"`
-	Memory       *Percentage    `yaml:"memory_percentage"`
+	Cooldown     *Cooldown      `yaml:"cooldown"`
```
Lou1415926: quick question: any reason why we prefer `*Cooldown` over `Cooldown` here? (same for `CPU` and `Memory` fields)

I think normally we'd want to keep members of `struct` type non-pointer, and members of basic types (bool, numeric, string) with pointer - to work with `mergo`, reduce nil-pointer check and prevent nil pointer panic downstream!

internal/pkg/manifest/svc.go
```diff
@@ -148,6 +148,30 @@ func (c *Count) UnmarshalYAML(value *yaml.Node) error {
...
+		}
+	}
+
+	if !r.ScalingConfig.IsEmpty() {
+		// Successfully unmarshalled ScalingConfig fields, return
+		return nil
+	}
+
+	if err := value.Decode(&r.Value); err != nil {
+		return errUnmarshalCountOpts
```
Lou1415926:
```suggestion
		return errors.New(`unable to unmarshal into int or composite-style map`))
```
Since this is not unmarshalling the `count` field 😛

internal/pkg/manifest/svc.go
```diff
@@ -148,6 +148,30 @@ func (c *Count) UnmarshalYAML(value *yaml.Node) error {
 	return nil
 }
 
+// UnmarshalYAML overrides the default YAML unmarshaling logic for the ScalingConfigOrPercentage
+// struct, allowing it to perform more complex unmarshaling behavior.
+// This method implements the yaml.Unmarshaler (v3) interface.
+func (r *ScalingConfigOrPercentage) UnmarshalYAML(value *yaml.Node) error {
```
Lou1415926, line 154: We probably need to implement a transformer for this struct so that environment override applies nicely: https://github.com/aws/copilot-cli/blob/c5fb60645ccef3fab823a36a551d49c927716f2b/internal/pkg/manifest/transform.go.

I'd also be fine with implementing the transformer in an upcoming PR. But just to explain a little why we need it:

Right now, likely both of these manifests ⬇️  
```yaml
count:
  range: 1-3
  cpu_percentage: 30    

environments:
  test:
    count:
      range: 1-3
      cpu_percentage:
        value: 40   
        cooldown:
           in: 3m 
```
and 
```yaml
count:
  range: 1-3
  cpu_percentage:
        value: 40   
        cooldown:
           in: 3m   

environments:
  test:
    count:
      range: 1-3
      cpu_percentage: 30
```

would result in a manifest struct that both has:
```go
AdvancedCount.CPU.Value = 30
AdvancedCount.CPU.ScalingConfig.Value = 40
AdvancedCount.CPU.ScalingConfig.Cooldown.ScaleInCooldown = 3m
```
That is, both `AdvancedCount.CPU.Value` and `AdvancedCount.CPU.ScalingConfig` are non-empty, while at most one of them should be non-empty. It'd be then ambiguous which - `AdvancedCount.CPU.Value` or `AdvancedCount.CPU.ScalingConfig` - contains the value that the user truly want to use (in the first case, `AdvancedCount.CPU.ScalingConfig`  would be the true answer; in the second case, `AdvancedCount.CPU.Value`; but we wouldn't know if they are both non-empty)

Let me know if this make sense!

CaptainCarpensir: This was the next step for what I was going to do with the feature, thanks for letting me know something I'll need to take into consideration 👍

internal/pkg/manifest/svc.go
CaptainCarpensir: To be quite honest, I wrote as many things as pointers because I couldn't think of any reasons they wouldn't need to be, just for the memory advantage. I'll try to restructure them then to make it work with `mergo` better. 👍

internal/pkg/deploy/cloudformation/stack/transformers_test.go
```diff
@@ -411,8 +414,20 @@ func Test_convertAutoscaling(t *testing.T) {
...
 		mockRequests     = 1000
 		mockResponseTime = 512 * time.Millisecond
-		mockCPU          = manifest.Percentage(70)
-		mockMem          = manifest.Percentage(80)
+		perc             = manifest.Percentage(70)
+		timeMinute       = time.Second * 60
+		mockCPU          = manifest.ScalingConfigOrPercentage{
+			ScalingConfig: manifest.AdvancedScalingConfig{
+				Value: &perc,
+				Cooldown: &manifest.Cooldown{
```
paragbhingre: Are we considering the fact that The `cooldown` field can only be specified if `range` is specified somewhere?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -196,11 +196,21 @@ func convertAutoscaling(a manifest.AdvancedCount) (*template.AutoscalingOpts, er
 		MinCapacity: &min,
 		MaxCapacity: &max,
 	}
-	if a.CPU != nil {
-		autoscalingOpts.CPU = aws.Float64(float64(*a.CPU))
+	if !a.CPU.IsEmpty() {
```
iamhopaul123: looks like this condition doesn't bring much benefit 🤔

internal/pkg/manifest/validate.go
```diff
@@ -850,16 +850,42 @@ func (a AdvancedCount) Validate() error {
...
+			}
+		}
+		if !a.CPU.IsEmpty() {
+			if !a.CPU.ScalingConfig.IsEmpty() {
+				return &errFieldMutualExclusive{
+					firstField:  "cooldown",
+					secondField: "cpu_percentage/cooldown",
+				}
+			}
+		}
```
iamhopaul123: Hmm I don't think they should be mutually exclusive with each other. The top level `cooldown` should be a general one where as any sub level ones could override its behavior.

internal/pkg/manifest/validate.go
```diff
@@ -850,16 +850,42 @@ func (a AdvancedCount) Validate() error {
...
+	}
+
 	// Validate individual custom autoscaling options.
 	if err := a.QueueScaling.Validate(); err != nil {
 		return fmt.Errorf(`validate "queue_delay": %w`, err)
 	}
-	if a.CPU != nil {
+	if !a.CPU.IsEmpty() {
```
iamhopaul123: we probably don't need this because of https://github.com/aws/copilot-cli/pull/3629/files#diff-5fc13795122fe030799a09ef956a8b66cec3d40bf1a81266323edbf3626b5685R906

internal/pkg/manifest/svc.go
```diff
@@ -172,23 +196,57 @@ func (c *Count) Desired() (*int, error) {
...
-	ResponseTime *time.Duration `yaml:"response_time"`
-	QueueScaling QueueScaling   `yaml:"queue_delay"`
+	Spot         *int                      `yaml:"spot"` // mutually exclusive with other fields
+	Range        Range                     `yaml:"range"`
+	Cooldown     Cooldown                  `yaml:"cooldown"`
+	CPU          ScalingConfigOrPercentage `yaml:"cpu_percentage"`
+	Memory       ScalingConfigOrPercentage `yaml:"memory_percentage"`
+	Requests     *int                      `yaml:"requests"`
+	ResponseTime *time.Duration            `yaml:"response_time"`
+	QueueScaling QueueScaling              `yaml:"queue_delay"`
```
iamhopaul123, line 227: We need cool down support for these scaling policies as well :)

CaptainCarpensir: Will add these in a separate PR after the cooldown is working fully for the general, cpu, and memory fields.

iamhopaul123 (approved):
LGTM :shipit:

internal/pkg/manifest/transform_test.go
```diff
@@ -638,6 +641,62 @@ func TestAdvancedCountTransformer_Transformer(t *testing.T) {
...
+	perc := Percentage(80)
+	mockConfig := AdvancedScalingConfig{
+		Value: &perc,
+	}
+	testCases := map[string]struct {
+		original func(s *ScalingConfigOrPercentage)
+		override func(s *ScalingConfigOrPercentage)
+		wanted   func(s *ScalingConfigOrPercentage)
+	}{
+		"percentage set to nil if advanced config value is not nil": {
```
Lou1415926: seems like the two test cases name should be swapped!

Lou1415926, line 654: seems like the two test cases name should be swapped!

internal/pkg/manifest/validate.go
```diff
@@ -875,6 +882,33 @@ func (p Percentage) Validate() error {
...
+
+// Validate returns nil if AdvancedScalingConfig is configured correctly.
+func (r AdvancedScalingConfig) Validate() error {
+	if r.IsEmpty() {
+		return nil
+	}
+	if err := r.Value.Validate(); err != nil {
+		return err
+	}
+	return nil
```
Lou1415926: tiny nit:
```suggestion
	return r.Cooldown.Validate()
```
Even though it's a no-op, I think it makes sense to still make this function call here. Otherwise in the future, when we want to add some validation to `Cooldown`, we could forget to add `r.Cooldown.Validate` here.

Lou1415926: tiny nit:
```suggestion
	return r.Cooldown.Validate()
```
I think it makes sense to still make this function call here because `AdvancedScalingConfig` doesn't know whether cooldown's validate is a no-op or not.

internal/pkg/manifest/validate.go
```diff
@@ -850,20 +850,27 @@ func (a AdvancedCount) Validate() error {
 		}
 	}
 
+	// Validate combinations with cooldown
+	if !a.Cooldown.IsEmpty() {
```
Lou1415926, line 871: Is this manifest valid:
```
count:
  range: 1-3
  cooldown:   # Optional.
    in: 3m        # Defaults to 2m
    out: 1m30s    # Defaults to 1m
  cpu_percentage: 
    value: 40   
    cooldown:
       in: 3m    
  memory_percentage: 60
```

Lou1415926, line 871: Is this manifest valid?
```
count:
  range: 1-3
  cooldown:   # Optional.
    in: 3m        # Defaults to 2m
    out: 1m30s    # Defaults to 1m
  cpu_percentage: 
    value: 40   
    cooldown:
       in: 3m    
```
where cooldown is specified in both levels?

CaptainCarpensir: Both the above manifests are valid, and the expected output should be that the lower level overwrites the upper level. I'm slightly uncertain about how it's done in code because as I'm writing the next part of the feature in `deploy/cloudformation/stack/transformers.go` the logic here is dependent on the order the lines are executed, but I don't know if there's a different way (best practices) that avoids this.

`
	// Write from generalized cooldown first
	generalCooldown := convertCooldown(a.Cooldown)
	CPUCooldown = generalCooldown
	MemCooldown = generalCooldown

	// If both generalized cooldown and specific cooldowns are specified, cooldown is overwritten by lower level config
	if cooldown := convertCooldown(a.CPU.ScalingConfig.Cooldown); cooldown != nil {
		CPUCooldown = cooldown
	}
	if cooldown := convertCooldown(a.Memory.ScalingConfig.Cooldown); cooldown != nil {
		MemCooldown = cooldown
	}

	autoscalingOpts.CPUCooldown = CPUCooldown
	autoscalingOpts.MemCooldown = MemCooldown`

You can see in the above conversation that iamhopaul123 specified they should not be mutually exclusive:
https://github.com/aws/copilot-cli/pull/3629#discussion_r890361447

Lou1415926 (approved):
Put on a DNM label if you want to address the nits. Feel free to remove it when you are ready!

Issue #3627, opened 2022-06-03
[Feature] Use many aliases by service


rverpillot:
Could be possible to use many aliases by service ? Something like this:
```

http:
  path: '/'
  aliases:
    - alias: myapp.aws-private.com
      hosted_zone_id: Z0873220N255IR3MTNR4
    - alias: myapp.corp-private.com

```
The goal is to add listener rules for different domain names.

With this feature we could migrate applications and still using existing URLs by creating DNS CNAME (and importing existing certificates as well).

Today:
`myapp.corp-private.com (A)` --> `10.x.x.x`

Tomorrow:
`myapp.corp-private.com (CNAME)` --> `myapp.aws-private.com (A)` --> `ALB`

PR #3636, opened 2022-06-06
chore(manifest): add advanced alias type to alias

iamhopaul123:
Related to #3627

### DIFF
--- a/internal/pkg/manifest/http.go
+++ b/internal/pkg/manifest/http.go
@@ -100,7 +100,10 @@ func (a *AdvancedAlias) isEmpty() bool {
 
 // Alias is a custom type which supports unmarshaling "http.alias" yaml which
 // can either be of type advancedAlias slice or type stringSliceOrString.
-type Alias advancedAliasSliceOrStringSliceOrString
+type Alias struct {
+	AdvancedAliases     []AdvancedAlias
+	StringSliceOrString stringSliceOrString
+}
 
 // IsEmpty returns empty if Alias is empty.
 func (a *Alias) IsEmpty() bool {
@@ -111,16 +114,26 @@ func (a *Alias) IsEmpty() bool {
 // struct, allowing it to perform more complex unmarshaling behavior.
 // This method implements the yaml.Unmarshaler (v3) interface.
 func (a *Alias) UnmarshalYAML(value *yaml.Node) error {
-	if err := unmarshalYAMLToAdvancedAliasSliceOrStringSliceOrString(
-		(*advancedAliasSliceOrStringSliceOrString)(a), value); err != nil {
-		return errUnmarshalAlias
+	if err := value.Decode(&a.AdvancedAliases); err != nil {
+		switch err.(type) {
+		case *yaml.TypeError:
+			break
+		default:
+			return err
+		}
 	}
-	return nil
+
+	if len(a.AdvancedAliases) != 0 {
+		// Unmarshaled successfully to s.StringSlice, unset s.String, and return.
+		a.StringSliceOrString = stringSliceOrString{}
+		return nil
+	}
+	return unmarshalYAMLToStringSliceOrString(&a.StringSliceOrString, value)
 }
 
 // ToAdvancedAliasSlice converts an Alias to a slice of advancedAliasSlice.
 func (a *Alias) ToStringSlice() ([]string, error) {
-	s := (*advancedAliasSliceOrStringSliceOrString)(a)
+	s := (*Alias)(a)
 	if len(s.AdvancedAliases) != 0 {
 		aliases := make([]string, len(s.AdvancedAliases))
 		for i, advancedAlias := range s.AdvancedAliases {
@@ -137,7 +150,7 @@ func (a *Alias) ToStringSlice() ([]string, error) {
 
 // ToString converts an Alias to a string.
 func (a *Alias) ToString() string {
-	s := (*advancedAliasSliceOrStringSliceOrString)(a)
+	s := (*Alias)(a)
 	if len(s.AdvancedAliases) != 0 {
 		aliases := make([]string, len(s.AdvancedAliases))
 		for i, advancedAlias := range s.AdvancedAliases {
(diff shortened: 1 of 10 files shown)


internal/pkg/manifest/http.go
```diff
@@ -88,38 +88,65 @@ func ipNetP(s string) *IPNet {
 	return &ip
 }
 
+// AdvancedAlias represents advanced alias configuration.
+type AdvancedAlias struct {
+	Alias      *string `yaml:"value"`
```
efekarakus: Can we use `name` to match the AWS API 🥺 ? https://aws.amazon.com/premiumsupport/knowledge-center/alias-resource-record-set-route53-cli/
```suggestion
	Alias      *string `yaml:"name"`
```

internal/pkg/manifest/workload.go
```diff
@@ -264,11 +264,38 @@ func (c *CommandOverride) ToStringSlice() ([]string, error) {
 	return out, nil
 }
 
+type advancedAliasSliceOrStringSliceOrString struct {
+	AdvancedAliases     []AdvancedAlias
+	StringSliceOrString stringSliceOrString
+}
```
efekarakus: Is there a reason why you preferred this route vs something like this:
```go
type Alias struct {
   value string       // mutually exclusive with advanced
   advanced []MultiAlias
}

type MultiAlias struct {
    value string  // mutually exclusive with advanced
    advanced []AliasWithHostedZone
}

type AliasWithHostedZone struct {
   name string
   hostedZone string
}
```

iamhopaul123: I think this way is slightly better from reusability's perspective. We only have one alias field in the manifest, and we have multiple fields that are `stringSliceOrString`. If we have multiple alias fields in the manifest and one of them uses `Alias` whereas one of them uses `AliasWithHostedZone` then I would agree with the example above.

efekarakus: gotcha! 
OK sounds good, what do you think of just getting rid of `advancedAliasSliceOrStringSliceOrString` and instead have:

```go
type Alias struct {
	AdvancedAliases     []AdvancedAlias  // Mutually exclusive with Simple.
	Simple stringSliceOrString
}
```
My concern is that "advancedAliasSliceOrStringSliceOrString" is really difficult to read 😂 it took me a while to realize it's "advancedAliasSlice" OR "StringSliceOrString" vs. "advancedAliasSliceOrStringSlice" OR "String"

internal/pkg/manifest/validate.go
```diff
@@ -694,7 +694,28 @@ func (h NLBHealthCheckArgs) Validate() error {
 }
 
 // Validate returns nil if Alias is configured correctly.
-func (Alias) Validate() error {
+func (a Alias) Validate() error {
```
Lou1415926, line 703: It's not quite valid yet for RDWS to have ⬇️  right?
```
http:
   alias:
      - name: blah
         hosted_zone: blah
```
Do you think it's worthwhile validating it here?

internal/pkg/manifest/http.go
```diff
@@ -88,38 +88,81 @@ func ipNetP(s string) *IPNet {
...
+	if err := unmarshalYAMLToStringSliceOrString(&a.StringSliceOrString, value); err != nil {
 		return errUnmarshalAlias
 	}
 	return nil
 }
 
-// ToStringSlice converts an Alias to a slice of string using shell-style rules.
-func (e *Alias) ToStringSlice() ([]string, error) {
-	out, err := toStringSlice((*stringSliceOrString)(e))
+// ToAdvancedAliasSlice converts an Alias to a slice of advancedAliasSlice.
```
Lou1415926:
```suggestion
// ToStringSlice converts an Alias to a slice of string.
```

internal/pkg/manifest/http.go
```diff
@@ -88,38 +88,81 @@ func ipNetP(s string) *IPNet {
...
+func (a *Alias) ToStringSlice() ([]string, error) {
+	s := (*Alias)(a)
```
Lou1415926: Do we  still need this?

internal/pkg/manifest/http.go
```diff
@@ -88,38 +88,81 @@ func ipNetP(s string) *IPNet {
...
-	return out, nil
+	return aliases, nil
 }
 
 // ToString converts an Alias to a string.
-func (e *Alias) ToString() string {
-	if e.String != nil {
-		return aws.StringValue(e.String)
+func (a *Alias) ToString() string {
+	s := (*Alias)(a)
```
Lou1415926: samesies!

internal/pkg/manifest/http_test.go
```diff
@@ -49,6 +50,83 @@ func TestRoutingRuleConfigOrBool_Disabled(t *testing.T) {
 	}
 }
 
+func TestAlias_UnmarshalYAML(t *testing.T) {
```
Lou1415926, line 53: nice!

internal/pkg/manifest/transform.go
```diff
@@ -98,9 +99,36 @@ func (t buildArgsOrStringTransformer) Transformer(typ reflect.Type) func(dst, sr
 	}
 }
 
+type aliasTransformer struct{}
+
+// Transformer returns custom merge logic for Alias's fields.
+func (t aliasTransformer) Transformer(typ reflect.Type) func(dst, src reflect.Value) error {
+	if !typ.ConvertibleTo(reflect.TypeOf(Alias{})) {
```
Lou1415926: nit: this could just be
```suggestion
	if typ != reflect.TypeOf(Alias{}) {
		return nil
	}
```
since we don't have any type that's aliased to `Alias` type. (If we have any in the future, we are not 100% now sure whether we'd want to apply this transformer).

Lou1415926 (approved):
LGTM! Just a few nits

internal/pkg/manifest/validate.go
iamhopaul123: For RDWS the alias is just a string https://aws.github.io/copilot-cli/docs/manifest/rd-web-service/#http-alias 🤔

iamhopaul123: But that does remind me that we might need to do that for NLB.

efekarakus (approved):

PR #3638, opened 2022-06-06
chore(stack): preserve previous values for env-controller-managed env stack parameters 

Lou1415926:
This PR contains two changes.

Env stack parameters in `Parameters` section can be grouped into those that are controlled by env-controller, and those that are not. The first group of parameter  should only be updated by env controller. This means that the env stack should explicitly assign them `ParameterValue` when these parameters are added to the stack *for the first time*; otherwise, it should `UsePreviousValue`.

An additional change is simplifying part of the changes made by #3528. After v1.14.0, container insights is either "enabled" or "disabled", instead of empty. This behavior should be kept.


internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -174,19 +174,24 @@ func (e *EnvStackConfig) managedVPC() template.ManagedVPC {
...
 	if e.in.Mft != nil {
-		if e.in.Mft.Observability.IsEmpty() {
-			return nil
+		var enabled bool
+		if aws.BoolValue(e.in.Mft.Observability.ContainerInsights) {
+			enabled = true
 		}
 		return &template.Telemetry{
-			EnableContainerInsights: e.in.Mft.Observability.ContainerInsights,
+			EnableContainerInsights: enabled,
```
efekarakus: nit:
```go
return &template.Telemetry{
    EnableContainerInsights: aws.BoolValue(e.in.Mft.Observability.ContainerInsights),
}
```

internal/pkg/deploy/cloudformation/env.go
```diff
@@ -159,67 +155,49 @@ func (cf CloudFormation) UpgradeEnvironment(in *deploy.CreateEnvironmentInput) e
...
-			paramKey := aws.StringValue(param.ParameterKey)
-			if _, ok := paramSet[paramKey]; !ok {
-				continue
-			}
-			paramSet[paramKey] = param
-		}
-		var params []*awscfn.Parameter
-		for _, param := range paramSet {
-			params = append(params, param)
+		// Generally, we want to:
```
efekarakus: nit: maybe drop the "Generally"
```suggestion
		// We want to:
```

internal/pkg/deploy/cloudformation/env.go
```diff
@@ -247,3 +225,99 @@ func (cf CloudFormation) upgradeEnvironment(in *deploy.CreateEnvironmentInput, t
...
+	}
+
+	var params []*awscfn.Parameter
+	for _, param := range curr {
+		params = append(params, param)
+	}
+	return params, nil
+}
+
+func transformEnvControllerParameters(new *awscfn.Parameter, old *awscfn.Parameter) *awscfn.Parameter {
```
efekarakus: What's the reason where we want to use the new param values for EnvControllerManaged parameters vs. `UsePreviousValue`? Can we add it as a comment to this function?

internal/pkg/deploy/cloudformation/env.go
```diff
@@ -247,3 +225,99 @@ func (cf CloudFormation) upgradeEnvironment(in *deploy.CreateEnvironmentInput, t
...
+	}
+	old := make(map[string]*awscfn.Parameter)
+	for _, p := range oldParams {
+		old[aws.StringValue(p.ParameterKey)] = p
+	}
+
+	// Remove or transform each of the current parameters.
+	for k, p := range curr {
+		transformed := transform(p, old[k])
+		if transformed == nil {
```
efekarakus: I wonder if we should change the signature of `transform` to be:
`transform func(new awscfn.Parameter, old *awscfn.Parameter) *awscfn.Parameter`

Since `if new == nil` can never happen in the `transform` functions, we can get rid of one of the if-statements.

internal/pkg/deploy/cloudformation/env.go
```diff
@@ -247,3 +225,99 @@ func (cf CloudFormation) upgradeEnvironment(in *deploy.CreateEnvironmentInput, t
...
+	}
+	return params, nil
+}
+
+func transformEnvControllerParameters(new *awscfn.Parameter, old *awscfn.Parameter) *awscfn.Parameter {
+	if new == nil {
+		return nil
+	}
+	if old == nil {
+		return new
```
efekarakus: can we drop a comment in the various transform functions like this 🥺 :
```suggestion
	if old == nil {
		return new // The ParamKey doesn't exist in the old stack, use the new value.
```

efekarakus (commented):
Looks good, mostly adding some clarification comments

internal/pkg/deploy/cloudformation/env.go
Lou1415926: Responding here but will add ⬇️  as the comment as well:

One of your suggested comment explain exactly when we want to "use the new param value".
> // The ParamKey doesn't exist in the old stack, use the new value.

Right after `env init`, the env stack's `Parameter` section doesn't have parameters such as `ALBWorkloads`. At the first time users run `env deploy` after `env init`, these parameters are "new parameters that do not exist in the old stack", in which case, we can't `UsePreviousValue`.

This is the only time we set the param value for env-controller-managed params from env stack. We want to `UsePreviousValue` otherwise.

internal/pkg/deploy/cloudformation/env.go
iamhopaul123: Or maybe just 
```suggestion
```
and move the comment to `transformParameters`

iamhopaul123 (approved):
Overall LGTM +1 to efe's fb.

efekarakus (approved):

PR #3640, opened 2022-06-08
chore(customresource): add ArtifactPath method

efekarakus:


internal/pkg/deploy/upload/customresource/customresource.go
```diff
@@ -54,31 +54,44 @@ var (
...
 
-// Zip creates a zip archive from all the files in the custom resource and returns a reader for its content.
-func (cr *CustomResource) Zip() (io.Reader, error) {
+// ArtifactPath returns the S3 key for the zipped custom resource object.
+func (cr *CustomResource) ArtifactPath() string {
+	return artifactpath.CustomResource(strings.ToLower(cr.FunctionName()), cr.zip.Bytes())
+}
+
+// Zip returns a reader for the zip archive from all the files in the custom resource.
+func (cr *CustomResource) Zip() io.Reader {
```
Lou1415926: Do these methods need to be exported ?

internal/pkg/aws/s3/s3.go
```diff
@@ -155,6 +155,11 @@ func ParseURL(url string) (bucket string, key string, err error) {
 	return
 }
 
+// URL returns a virtual-hosted–style S3 url for the object stored at key in a bucket created in the specified region.
+func URL(region, bucket, key string) string {
```
Lou1415926, line 161: I guess this function is getting referenced in an upcoming PR is that correct 👀 ?

Lou1415926 (approved):

internal/pkg/aws/s3/s3.go
```diff
@@ -155,6 +155,11 @@ func ParseURL(url string) (bucket string, key string, err error) {
...
+	return fmt.Sprintf("https://%s.s3.%s.amazonaws.com/%s", bucket, region, key)
```
iamhopaul123: potential partition problem here? maybe not a big deal because we don't have any customers there

iamhopaul123 (approved):

internal/pkg/deploy/upload/customresource/customresource.go
efekarakus: Done!

internal/pkg/aws/s3/s3.go
efekarakus: Done! I only did this for the CN partition because I couldn't figure out if for other paritions the URL changes X_X

PR #3642, opened 2022-06-08
chore(deploy): use customresource pkg to UploadArtifacts 

efekarakus:
We call the new `customresource` path only behind a feature flag so that we donot upload these lambda functions for now.

Also refactors to rename receivers for LBWS deployers and validation method names.

### DIFF
--- a/internal/pkg/cli/deploy/deploy.go
+++ b/internal/pkg/cli/deploy/deploy.go
@@ -1017,7 +1017,7 @@ func (d *lbWebSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (*s
 	if err := d.validateALBRuntime(); err != nil {
 		return nil, err
 	}
-	if err := d.validateNLBWSRuntime(); err != nil {
+	if err := d.validateNLBRuntime(); err != nil {
 		return nil, err
 	}
 	var opts []stack.LoadBalancedWebServiceOption
@@ -1374,7 +1374,7 @@ func (d *lbWebSvcDeployer) validateALBRuntime() error {
 	return fmt.Errorf("cannot specify http.alias when application is not associated with a domain and env %s doesn't import one or more certificates", d.env.Name)
 }
 
-func (d *lbWebSvcDeployer) validateNLBWSRuntime() error {
+func (d *lbWebSvcDeployer) validateNLBRuntime() error {
 	if d.lbMft.NLBConfig.Aliases.IsEmpty() {
 		return nil
 	}
--- a/internal/pkg/cli/deploy/deploy.go
+++ b/internal/pkg/cli/deploy/deploy.go
@@ -1014,7 +1014,7 @@ func (d *lbWebSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (*s
 	if err != nil {
 		return nil, err
 	}
-	if err := d.validateALBWSRuntime(); err != nil {
+	if err := d.validateALBRuntime(); err != nil {
 		return nil, err
 	}
 	if err := d.validateNLBWSRuntime(); err != nil {
@@ -1343,7 +1343,7 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
 	return nil
 }
 
-func (d *lbWebSvcDeployer) validateALBWSRuntime() error {
+func (d *lbWebSvcDeployer) validateALBRuntime() error {
 	if d.lbMft.RoutingRule.Alias.IsEmpty() {
 		if d.env.HasImportedCerts() {
 			return &errSvcWithNoALBAliasDeployingToEnvWithImportedCerts{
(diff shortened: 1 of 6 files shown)


internal/pkg/cli/deploy/deploy.go
```diff
@@ -812,6 +905,20 @@ func (d *workloadDeployer) uploadArtifactsToS3(in *uploadArtifactsToS3Input) (up
 	}, nil
 }
 
+func (d *workloadDeployer) uploadCustomResources(workloadType string, customResources func(template.Reader) ([]*customresource.CustomResource, error)) (map[string]string, error) {
```
iamhopaul123: do we need this parameter `workloadType`? It seems like it is only used for error msg.

internal/pkg/cli/deploy/env.go
```diff
@@ -49,11 +58,32 @@ func (d *envDeployer) UploadArtifacts() (map[string]string, error) {
...
+		return d.s3.ZipAndUpload(bucket, key, objects...)
 	})
 	if err != nil {
-		return nil, fmt.Errorf("upload custom resources to bucket %s: %w", resources.S3Bucket, err)
+		return nil, fmt.Errorf("upload custom resources to bucket %s: %w", bucket, err)
+	}
+	return urls, nil
+}
+
+func (d *envDeployer) uploadCustomResources(bucket string) (map[string]string, error) {
```
iamhopaul123, line 77: Why do we need this new `uploadCustomResources` 🤔

efekarakus: For environments? they're basically the same except that the functions upload the custom resources in different paths.

internal/pkg/cli/deploy/deploy.go
efekarakus: Do you have a recommendation for an alternative way of writing it? 🥺 otherwise the error messages won't be specific enough

internal/pkg/cli/deploy/deploy.go
```diff
@@ -499,6 +507,91 @@ func (d *workloadDeployer) UploadArtifacts() (*UploadArtifactsOutput, error) {
...
+		return nil, err
+	}
+	if !d.uploadCustomResourceFlag {
+		return out, nil
+	}
+	urls, err := d.uploadCustomResources(manifest.LoadBalancedWebServiceType, customresource.LBWS)
+	if err != nil {
+		return nil, err
+	}
+	out.CustomResourceURLs = urls
```
dannyrandall: what if these were in the `workloadDeployer.UploadArtifacts()` func instead? is there a reason for splitting it by workload type? they just seem pretty similar now
L502:
```
if !d.uplaodCustomResourceFlag {
    return &UploadArtifcatsOutput{...}
}

switch d.mft.(type) {
case lbws:
    out.CustomResourceURLs, err = d.uploadCustomResources(lbwsType, lbws)
// etc
}
if err != nil {
}

return out
```

internal/pkg/cli/deploy/deploy.go
efekarakus: Got rid of it!

internal/pkg/cli/deploy/deploy.go
efekarakus: I did something slightly different, I've added the customResourcesFunc as a field in the various service deployers and reduced the duplication across the board.

Let me know what you think!

internal/pkg/cli/deploy/deploy_test.go
```diff
@@ -128,6 +160,139 @@ func TestWorkloadDeployer_UploadArtifacts(t *testing.T) {
...
 		},
+		"should retrieve Load Balanced Web Service custom resource URLs": {
+			inUploadCustomResourcesFlag: true,
+			mock: func(m *deployMocks) {
+				// Ignore addon uploads.
+				m.mockTemplater.EXPECT().Template().Return("", &addon.ErrAddonsNotFound{})
+
+				// Ensure all custom resources were uploaded.
+				crs, err := customresource.LBWS(fakeTemplateFS())
+				require.NoError(t, err)
```
dannyrandall, line 171: will this fail the parent test, not the subtest? (idk what the output looks like, but if it's clear which test failed, then idc)

dannyrandall (approved):
lgtm! just one nit question

internal/pkg/cli/deploy/deploy.go
dannyrandall: looks great :meow_love:

internal/pkg/cli/deploy/deploy_test.go
efekarakus: yeah it's pretty awkward that it causes the parent test to fail :rip: we should take `*testing.T` as an input, let me see if I can fix that real quick

iamhopaul123 (approved):
:shipit:

PR #3643, opened 2022-06-08
feat(template): allow multiple alias hosted zones

### LINKED ISSUE #3627: [Feature] Use many aliases by service

iamhopaul123:
resolves #3627 

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
+++ b/internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
@@ -189,6 +189,11 @@ func (s *LoadBalancedWebService) Template() (string, error) {
 		}
 	}
 
+	hostedZoneToAliases := convertHostedZone(s.manifest.RoutingRule.RoutingRuleConfiguration)
+	if len(hostedZoneToAliases) != 0 && !s.certImported {
+		return "", fmt.Errorf("cannot specify alias hosted zones when env certificates are managed by Copilot")
+	}
+
 	var deregistrationDelay *int64 = aws.Int64(60)
 	if s.manifest.RoutingRule.DeregistrationDelay != nil {
 		deregistrationDelay = aws.Int64(int64(s.manifest.RoutingRule.DeregistrationDelay.Seconds()))
@@ -246,6 +251,7 @@ func (s *LoadBalancedWebService) Template() (string, error) {
 		Observability: template.ObservabilityOpts{
 			Tracing: strings.ToUpper(aws.StringValue(s.manifest.Observability.Tracing)),
 		},
+		HostedZone: hostedZoneToAliases,
 	})
 	if err != nil {
 		return "", err
--- a/internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
+++ b/internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
@@ -189,8 +189,11 @@ func (s *LoadBalancedWebService) Template() (string, error) {
 		}
 	}
 
-	hostedZoneToAliases := convertHostedZone(s.manifest.RoutingRule.RoutingRuleConfiguration)
-	if len(hostedZoneToAliases) != 0 && !s.certImported {
+	aliasesFor, err := convertHostedZone(s.manifest.RoutingRule.RoutingRuleConfiguration)
+	if err != nil {
+		return "", err
+	}
+	if len(aliasesFor) != 0 && !s.certImported {
 		return "", fmt.Errorf("cannot specify alias hosted zones when env certificates are managed by Copilot")
 	}
 
@@ -251,7 +254,7 @@ func (s *LoadBalancedWebService) Template() (string, error) {
 		Observability: template.ObservabilityOpts{
 			Tracing: strings.ToUpper(aws.StringValue(s.manifest.Observability.Tracing)),
 		},
-		HostedZone: hostedZoneToAliases,
+		HostedZoneAliases: aliasesFor,
 	})
 	if err != nil {
 		return "", err
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -116,28 +116,30 @@ func convertContainerHealthCheck(hc manifest.ContainerHealthCheck) *template.Con
 	}
 }
 
-func convertHostedZone(m manifest.RoutingRuleConfiguration) template.HostedZoneToAliases {
-	hostedZoneToAliases := make(map[string][]string)
+func convertHostedZone(m manifest.RoutingRuleConfiguration) (template.AliasesForHostedZone, error) {
+	aliasesFor := make(map[string][]string)
 	defaultHostedZone := m.HostedZone
-	for _, alias := range m.Alias.AdvancedAliases {
-		if alias.HostedZone != nil {
-			hostedZoneToAliases[*alias.HostedZone] = append(hostedZoneToAliases[*alias.HostedZone], *alias.Alias)
-			continue
+	if len(m.Alias.AdvancedAliases) != 0 {
+		for _, alias := range m.Alias.AdvancedAliases {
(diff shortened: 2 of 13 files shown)


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -116,6 +116,30 @@ func convertContainerHealthCheck(hc manifest.ContainerHealthCheck) *template.Con
 	}
 }
 
+func convertHostedZone(m manifest.RoutingRuleConfiguration) template.HostedZoneToAliases {
+	hostedZoneToAliases := make(map[string][]string)
```
efekarakus: nit: (I typically use `For` when creating maps. I first saw this from Martin Fowler's refactor book)
```suggestion
func convertHostedZone(m manifest.RoutingRuleConfiguration) template.AliasesForHostedZone {
	aliasesFor := make(map[string][]string)
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -116,6 +116,30 @@ func convertContainerHealthCheck(hc manifest.ContainerHealthCheck) *template.Con
...
+	defaultHostedZone := m.HostedZone
```
efekarakus, line 121: Should we still allow a top level `hosted_zone` field? I thought we would restrict it to be strictly under `alias`.

I think restricting it makes sense to me because I can't think of many usecases where the same hosted zone is going to be used by multiple aliases in a single service

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
```diff
@@ -20,16 +20,17 @@ LoadBalancerDNSAlias:
       AliasTarget:
         HostedZoneId: !GetAtt EnvControllerAction.PublicLoadBalancerHostedZone
         DNSName: !GetAtt EnvControllerAction.PublicLoadBalancerDNSName
-{{- else if .HostedZoneID}}
-LoadBalancerDNSAlias:
+{{- else}}
+{{- range $hostedZoneID, $aliases := .HostedZone}}
```
efekarakus: I'm probably missing something but doesn't this mean that _there has to be at least one_ `hosted_zone` field under `http`?

like what if I have:
```yaml
http:
  alias: [example.com, api.example.com]
``` 
Wouldn't this `{{range}}` block try to add the `defaultHostedZone := m.HostedZone` right now?

internal/pkg/template/workload.go
```diff
@@ -485,7 +488,7 @@ type WorkloadOpts struct {
 	ServiceDiscoveryEndpoint string
 	HTTPVersion              *string
 	ALBEnabled               bool
-	HostedZoneID             *string
+	HostedZone               HostedZoneToAliases
```
efekarakus: nit
```suggestion
	HostedZoneAliases               AliasesForHostedZone
```

internal/pkg/deploy/cloudformation/stack/transformers.go
iamhopaul123: Hmm the most common example I could think of is `example.com`, `www.example.com`, and `foobar.com`?

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
iamhopaul123: If it is
```yaml
http:
  alias: [example.com, api.example.com]
```
no A record will be added for them using the CFN resource (could be they imported certs and they want to manage A records on their own, or could be they use Copilot managed aliases that will be handled by the lambda).

if it is
```yaml
http:
  alias: [example.com, api.example.com]
  hosted_zone: ABCD123
```
then it applies to both `example.com` and `api.example.com`

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -116,6 +116,30 @@ func convertContainerHealthCheck(hc manifest.ContainerHealthCheck) *template.Con
...
+	for _, alias := range m.Alias.AdvancedAliases {
+		if alias.HostedZone != nil {
+			hostedZoneToAliases[*alias.HostedZone] = append(hostedZoneToAliases[*alias.HostedZone], *alias.Alias)
+			continue
+		}
+		if defaultHostedZone != nil {
+			hostedZoneToAliases[*defaultHostedZone] = append(hostedZoneToAliases[*defaultHostedZone], *alias.Alias)
+		}
+	}
```
Lou1415926, line 133: Can we return `hostedZoneToAliases` already at this point? Can we assume `AdvancedAliases` and `StringSliceOrString` won't be non-empty at the same time?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -116,6 +116,30 @@ func convertContainerHealthCheck(hc manifest.ContainerHealthCheck) *template.Con
...
+	if defaultHostedZone == nil {
+		return hostedZoneToAliases
+	}
+	for _, alias := range m.Alias.StringSliceOrString.StringSlice {
+		hostedZoneToAliases[*defaultHostedZone] = append(hostedZoneToAliases[*defaultHostedZone], alias)
+	}
+	if m.Alias.StringSliceOrString.String != nil {
+		hostedZoneToAliases[*defaultHostedZone] = append(hostedZoneToAliases[*defaultHostedZone], *m.Alias.StringSliceOrString.String)
+	}
```
Lou1415926: Do you think we can use `m.Alias.ToStringSlice` here to simplify this a little bit?

internal/pkg/deploy/cloudformation/stack/transformers.go
efekarakus: OK boss, `example.com` and `www.example.com` in the same hosted zone makes sense to me. We can keep it!

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
efekarakus: ok I see - in `convertHostedZone` if there is no `hosted_zone` field then we return an empty map that's not nil. But since the map is empty this range-block does nothing 👍 we're good!

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -189,8 +189,11 @@ func (s *LoadBalancedWebService) Template() (string, error) {
 		}
 	}
 
-	hostedZoneToAliases := convertHostedZone(s.manifest.RoutingRule.RoutingRuleConfiguration)
-	if len(hostedZoneToAliases) != 0 && !s.certImported {
+	aliasesFor, err := convertHostedZone(s.manifest.RoutingRule.RoutingRuleConfiguration)
+	if err != nil {
+		return "", err
+	}
+	if len(aliasesFor) != 0 && !s.certImported {
```
Lou1415926, line 196: Likely I'm missing something: does backend service need this validation?

Lou1415926 (approved):
:shipit:

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
iamhopaul123: Not really because for backend service we don't add A record by lambda

huanjani (approved):
This looks great. Thank you so much for implementing this (and so quickly)!

PR #3645, opened 2022-06-08
chore(session): try retrieving credentials within 10s if not err

paragbhingre:


internal/pkg/config/app.go
```diff
@@ -74,8 +76,18 @@ func (s *Store) UpdateApplication(application *Application) error {
 
 // GetApplication fetches an application by name. If it can't be found, return a ErrNoSuchApplication
 func (s *Store) GetApplication(applicationName string) (*Application, error) {
+
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	defer cancel()
```
efekarakus: **My main concern with this fix is that we don't provide any hints to the customer on what might be the issue.** After 3s they will receive a timeout error, but it won't be clear that it's because they are missing env vars or don't have credentials for their `[default]` or `AWS_PROFILE`.


For example, if I comment out my default credentials and use the AWS CLI, I get a nice error:
```
$  aws sts get-caller-identity                                                                    main

Unable to locate credentials. You can configure credentials by running "aws configure".
```
We should strive for a similar experience.

efekarakus: My understanding from the error that we get:
```
✘ get application demo configuration: get application demo: NoCredentialProviders: no valid providers in chain
caused by: EnvAccessKeyNotFound: failed to find credentials in the environment.
SharedCredsLoad: failed to load profile, .
EC2RoleRequestError: no EC2 instance role found
caused by: RequestError: send request failed
caused by: Get "http://169.254.169.254/latest/meta-data/iam/security-credentials/": dial tcp 169.254.169.254:80: connect: host is down
/Users/karakuse/Development/copilot-cli/bin/local/copilot svc init  0.02s user 0.01s system 0% cpu 5.555 total
```

The step that takes the longest time is while it's trying to see if it can fetch EC2 IAM role credentials 💭 

If I comment our custom timeouts:
https://github.com/aws/copilot-cli/blob/0bea71053c1f260ad4f832d7cb550353941f5d51/internal/pkg/aws/sessions/sessions.go#L26-L28

Then we receive an error much faster. Whereas now it takes 8*30s = 4 mins.
Maybe, we can change our defaults or create a test "default" session with very short timeouts (2 retries, 3 sec timeout) and see if we get an error.

internal/pkg/aws/sessions/errors.go
```diff
@@ -24,3 +24,21 @@ func (e *errMissingRegion) RecommendActions() string { // implements new actionR
...
+type errCredProviderTimeout struct{}
+
+// Implements error interface.
+func (e *errCredProviderTimeout) Error() string {
+	return "retrieving credentials timeout"
+}
+
+// RecommendActions returns recommended actions to be taken after the error.
+// Implements main.actionRecommender interface.
+func (e *errCredProviderTimeout) RecommendActions() string { // implements new actionRecommender interface.
```
efekarakus: nit: 
```suggestion
func (e *errCredProviderTimeout) RecommendActions() string {
```

internal/pkg/aws/sessions/errors.go
```diff
@@ -24,3 +24,21 @@ func (e *errMissingRegion) RecommendActions() string { // implements new actionR
...
+	return fmt.Sprintf(`It looks like your AWS configuration/credentials are missing.
+- We recommend including your configuration in the "~/.aws/config" & "~/.aws/credentials" file.
```
efekarakus: I really wish there was a way for us to figure out which provider is taking forever 😭 

Maybe, we can point a bit more towards what's missing?
```
It looks like your default AWS credential chain is missing or misconfigured:
https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html#specifying-credentials
- We recommend including your credentials in the shared credentials file.
- Alternatively, ...
```
or
```
It looks your [<profile name>] is missing or misconfigured.
...
```

internal/pkg/aws/sessions/sessions.go
```diff
@@ -202,3 +204,14 @@ func (p *Provider) userAgentHandler() request.NamedHandler {
...
+func validateCredentials(sess *session.Session) (credentials.Value, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), credsTimeout)
+	defer cancel()
+
+	v, credErr := sess.Config.Credentials.GetWithContext(ctx)
+	if credErr != nil {
+		return credentials.Value{}, credErr
+	}
+	return v, nil
+}
```
efekarakus: We need a way of distinguishing if the error is due to a timeout or if it's something else that went wrong
```go
func credsValue(sess *session.Session) (credentials.Value, error) {
	ctx, cancel := context.WithTimeout(context.Background(), credsTimeout)
	defer cancel()

	v, err := sess.Config.Credentials.GetWithContext(ctx)
	if err != nil {
		if errors.Is(err, context.DeadlineExceeded) {
			return credentials.Value{}, errCredProviderTimeout{}
		}
		return credentials.Value{}, err
	}
	return v, nil
}
```

internal/pkg/aws/sessions/sessions.go
```diff
@@ -155,6 +155,11 @@ func (p *Provider) defaultSession() (*session.Session, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	if _, credErr := validateCredentials(sess); credErr != nil {
```
efekarakus: can we also include this for the `Profile` function 🥺

efekarakus (commented):
Awesome! The message returned from `RecommendAction` was great. I tested it out locally and it reads great 😍

internal/pkg/aws/sessions/errors.go
paragbhingre: Yeah, then everything would have been so easy to just take that provider and tell the customer about the failure.

internal/pkg/aws/sessions/sessions.go
paragbhingre: Done

internal/pkg/aws/sessions/errors.go
```diff
@@ -24,3 +24,22 @@ func (e *errMissingRegion) RecommendActions() string { // implements new actionR
...
+// RecommendActions returns recommended actions to be taken after the error.
+// Implements main.actionRecommender interface.
+func (e *errCredProviderTimeout) RecommendActions() string {
+	return fmt.Sprintf(`It looks like your %s chain is missing or misconfigured:
+https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html#specifying-credentials
+- We recommend including your credentials in the shared credentials file.
+- Alternatively, you can also set credentials throguh 
+	* Environment Variables
+	* EC2 Instance Metadata (credentials only)
+More information: https://aws.github.io/copilot-cli/docs/credentials/`, color.HighlightCode("default AWS credential"))
```
efekarakus: If the err occurs in the `FromProfile` method then it's not default cred chain.

Perhaps we can have a field in the type like this:
```go
//errCredProviderTimeout indicates a timeout occurred while fetching credentials for a session.
// The zero value errCredProviderTimeout{} represents the default AWS credential chain.
type errCredProviderTimeout struct{
    profile string // Optional. If the credentials were created from a profile. 
}
```

internal/pkg/aws/sessions/sessions.go
```diff
@@ -202,3 +225,14 @@ func (p *Provider) userAgentHandler() request.NamedHandler {
...
+
+func (v *validator) ValidateCredentials(sess *session.Session) (credentials.Value, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), credsTimeout)
+	defer cancel()
+
+	val, err := sess.Config.Credentials.GetWithContext(ctx)
+	if err != nil {
+		return credentials.Value{}, err
+	}
+	return val, nil
```
efekarakus:
```suggestion
	return sess.Config.Credentials.GetWithContext(ctx)
```

internal/pkg/aws/sessions/sessions.go
```diff
@@ -104,6 +114,12 @@ func (p *Provider) FromProfile(name string) (*session.Session, error) {
 	if aws.StringValue(sess.Config.Region) == "" {
 		return nil, &errMissingRegion{}
 	}
+	if _, credErr := p.sessionValidator.ValidateCredentials(sess); credErr != nil {
+		if strings.Contains(credErr.Error(), "context deadline exceeded") {
```
efekarakus: Does this not work?
```suggestion
		if errors.Is(err, context.DeadlineExceeded) {
```

internal/pkg/aws/sessions/sessions_test.go
```diff
@@ -5,6 +5,8 @@ package sessions
 
 import (
 	"errors"
+	sessionMocks "github.com/aws/copilot-cli/internal/pkg/aws/sessions/mocks"
```
efekarakus: nit: 
```suggestion
	"github.com/aws/copilot-cli/internal/pkg/aws/sessions/mocks"
```

internal/pkg/aws/sessions/sessions_test.go
```diff
@@ -171,6 +173,13 @@ func TestProvider_FromProfile(t *testing.T) {
 	})
 
 	t.Run("region information present", func(t *testing.T) {
+
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		m := sessionMocks.NewMocksessionValidator(ctrl)
+		m.EXPECT().ValidateCredentials(gomock.Any()).Return(credentials.Value{}, nil)
```
efekarakus, line 182: with this new mock can we write a test case for 

`m.EXPECT().ValidateCredentials(gomock.Any()).Return(credentials.Value{}, context.DeadlineExceeded)` to see if we wrap the error appropriately?

efekarakus (commented):
Awesome, ty for adding the mock!

internal/pkg/aws/sessions/sessions.go
paragbhingre: Yeah it doesn't work that is why I had to do the string comparison way

paragbhingre: Also it is little confusing how sometimes we get ```DeadlineExceed``` error and other times we get 
```
NoCredentialProviders: no valid providers in chain
caused by: EnvAccessKeyNotFound: failed to find credentials in the environment.
SharedCredsLoad: failed to load profile, .
EC2RoleRequestError: no EC2 instance role found
caused by: EC2MetadataError: failed to make EC2Metadata request
```

So we need to check for both the errors.

internal/pkg/aws/sessions/sessions.go
```diff
@@ -104,6 +114,12 @@ func (p *Provider) FromProfile(name string) (*session.Session, error) {
 	if aws.StringValue(sess.Config.Region) == "" {
 		return nil, &errMissingRegion{}
 	}
+	if _, credErr := p.sessionValidator.ValidateCredentials(sess); credErr != nil {
```
dannyrandall:
```suggestion
	if _, err := p.sessionValidator.ValidateCredentials(sess); err != nil {
```
nit

internal/pkg/aws/sessions/sessions.go
dannyrandall: looks like this is why it doesn't work 😔  https://github.com/aws/aws-sdk-go/issues/2820

internal/pkg/aws/sessions/sessions.go
```diff
@@ -39,7 +40,14 @@ type Provider struct {
 	defaultSess *session.Session
 
 	// Metadata associated with the provider.
-	userAgentExtras []string
+	userAgentExtras  []string
+	sessionValidator sessionValidator
+}
+
+type validator struct{}
```
dannyrandall: nit: move this down to 228

internal/pkg/aws/sessions/sessions_test.go
```diff
@@ -183,12 +193,48 @@ func TestProvider_FromProfile(t *testing.T) {
...
+		// WHEN
+		provider := &Provider{
+			sessionValidator: m,
+		}
+
+		sess, err := provider.FromProfile("walk-like-an-egyptian")
+
+		// THEN
+		require.NotNil(t, err)
+		require.EqualError(t, errors.New("retrieving credentials timeout"), err.Error())
```
dannyrandall:
```suggestion
		require.Error(t, err)
		require.EqualError(t, err, "retrieving credentials timeout")
```
nit. could probably just get rid of `require.Error` entirely

dannyrandall (approved):
looks great!😊 just a couple of nits

internal/pkg/aws/sessions/errors.go
```diff
@@ -24,3 +24,24 @@ func (e *errMissingRegion) RecommendActions() string { // implements new actionR
...
+
+// Implements error interface.
+func (e *errCredProviderTimeout) Error() string {
+	return "retrieving credentials timeout"
+}
+
+// RecommendActions returns recommended actions to be taken after the error.
+// Implements main.actionRecommender interface.
+func (e *errCredProviderTimeout) RecommendActions() string {
+	return fmt.Sprintf(`It looks like your %s chain is missing or misconfigured:
```
AnthonyFoiani-at: As someone relatively new to AWS, I wouldn't immediately know what a "chain" is, in this context.

Maybe:

> Unable to obtain credentials for profile '%s'; it might be missing or misconfigured.

?

(came here from https://github.com/aws/copilot-cli/issues/3685 )

paragbhingre: @AnthonyFoiani-at Thank you for your recommendation. We have considered your ask and changed error message to something which will be easy to read for the people who are new to AWS . Please take a look at the error message in [here](https://github.com/aws/copilot-cli/pull/3645/files#diff-f890a37782b4a1c1ebff2e66278f12e089474a58144dbd9484552857f5e19d84R42-R53) and let us know if it makes sense to you.

internal/pkg/aws/sessions/errors.go
```diff
@@ -24,3 +24,37 @@ func (e *errMissingRegion) RecommendActions() string { // implements new actionR
...
+// Implements error interface.
+func (e *errCredProviderTimeout) Error() string {
+	return "retrieving credentials timeout"
+}
+
+// RecommendActions returns recommended actions to be taken after the error.
+// Implements main.actionRecommender interface.
+func (e *errCredProviderTimeout) RecommendActions() string {
+	var msg string
+	if e.profile == "default" {
```
efekarakus: it's not just the the "default" profile, we want to display this message if for example they use env vars as their credentials but it's misconfigured.

```go
msg := `It looks like your credential settings are misconfigured or missing:
https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html#specifying-credentials
- We recommend including your credentials in the shared credentials file.
- Alternatively, you can also set credentials through
	* Environment Variables
	* EC2 Instance Metadata (credentials only)
More information: https://aws.github.io/copilot-cli/docs/credentials/`

if e.profile != "" {
   msg = fmt.Sprintf(`It looks like your profile [%s] is misconfigured or missing:
https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html#specifying-credentials
- We recommend including your credentials in the shared credentials file.
- Alternatively, you can also set credentials through 
	* Environment Variables
	* EC2 Instance Metadata (credentials only)
More information: https://aws.github.io/copilot-cli/docs/credentials/`, e.profile)
}

return msg
```

internal/pkg/aws/sessions/sessions.go
```diff
@@ -104,6 +112,12 @@ func (p *Provider) FromProfile(name string) (*session.Session, error) {
 	if aws.StringValue(sess.Config.Region) == "" {
 		return nil, &errMissingRegion{}
 	}
+	if _, err := p.sessionValidator.ValidateCredentials(sess); err != nil {
+		if strings.Contains(err.Error(), "context deadline exceeded") || strings.Contains(err.Error(), "NoCredentialProviders") {
+			return nil, &errCredProviderTimeout{name}
```
efekarakus: If we return `errCredProviderTimeout` when the real error is "NoCredentialProviders", we lose some troubleshooting information. For example, we will print the error `"retrieving credentials timeout"` when in reality the creds didn't timeout, they just quickly failed to fetch.

Can we differentiate between the two?

```go
   if isCredRetrievalErr(err):
      return nil, &errCredRetrieval{profile: name, parentErr: err}
```

and we have the following instead of `errCredProviderTimeout`:
```go
type errCredRetrieval struct {
	profile string
	parentErr error
}
func (e *errCredRetrieval) Error() string {
	return e.parentErr
}
func (e *errCredRetrieval) RecommendActions() string {
        // same as before
}
```
```go
func isCredRetrievalErr(err error) bool {
    return strings.Contains(err.Error(), "context deadline exceeded") || strings.Contains(err.Error(), "NoCredentialProviders")
}
```

internal/pkg/aws/sessions/errors.go
paragbhingre: Okay I get it now. Changed.

internal/pkg/aws/sessions/sessions.go
paragbhingre: This makes sense to me.

internal/pkg/aws/sessions/errors.go
```diff
@@ -24,3 +24,36 @@ func (e *errMissingRegion) RecommendActions() string { // implements new actionR
...
+
+// Implements error interface.
+func (e *errCredRetrieval) Error() string {
+	return e.parentErr.Error()
+}
+
+// RecommendActions returns recommended actions to be taken after the error.
+// Implements main.actionRecommender interface.
+func (e *errCredRetrieval) RecommendActions() string {
+	msg := `It looks like your credential settings are misconfigured or missing:
```
efekarakus: nit: it looks like only the first sentence changes
```go
notice := "It looks like your credential settings are misconfigured or missing"
if e.profile != "" {
   notice = fmt.Sprintf("It looks like your profile [%s] is misconfigured or missing", e.profile)
}
return fmt.Sprintf(`%s:
https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html#specifying-credentials
- We recommend including your credentials in the shared credentials file.
- Alternatively, you can also set credentials through 
	* Environment Variables
	* EC2 Instance Metadata (credentials only)
More information: https://aws.github.io/copilot-cli/docs/credentials/`, notice)
```

internal/pkg/aws/sessions/sessions.go
```diff
@@ -202,3 +223,15 @@ func (p *Provider) userAgentHandler() request.NamedHandler {
...
+
+type validator struct{}
+
+func (v *validator) ValidateCredentials(sess *session.Session) (credentials.Value, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), credsTimeout)
+	defer cancel()
+	return sess.Config.Credentials.GetWithContext(ctx)
+}
+
+func isCredRetrievalErr(err error) bool {
```
efekarakus: nit: can we move this to `errors.go`

internal/pkg/aws/sessions/sessions.go
```diff
@@ -155,6 +169,13 @@ func (p *Provider) defaultSession() (*session.Session, error) {
 	if err != nil {
 		return nil, err
 	}
+	if _, err = p.sessionValidator.ValidateCredentials(sess); err != nil {
+		if isCredRetrievalErr(err) {
+			return nil, &errCredRetrieval{profile: "profile", parentErr: err}
```
efekarakus: Why are we setting this to `"profile"`? In this scenario, shouldn't it be:
```suggestion
			return nil, &errCredRetrieval{parentErr: err}
```
So that the default notice is shown.

paragbhingre: Oops I mistakenly wrote this profile thing. I have removed it now.

efekarakus (approved):

PR #3646, opened 2022-06-08
feat(alb): add flag to `env init` for adding internal ALB ingress

huanjani:
For environments that will have internal ALBs, passing in the `--internal-alb-allow-vpc-ingress` flag when running `copilot env init` will add inbound rule(s) for port 80 and/or port 443 when Backend Services with `http` (port 80) or `http.alias` (ports 80 and 443) are deployed in them.

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/env.go
+++ b/internal/pkg/deploy/cloudformation/stack/env.go
@@ -43,6 +43,7 @@ const (
 	envParamNATWorkloadsKey                = "NATWorkloads"
 	envParamCreateHTTPSListenerKey         = "CreateHTTPSListener"
 	envParamCreateInternalHTTPSListenerKey = "CreateInternalHTTPSListener"
+	envParamAllowVPCIngressKey             = "AllowVPCIngressForInternalALB"
 	EnvParamServiceDiscoveryEndpoint       = "ServiceDiscoveryEndpoint"
 
 	// Output keys.
@@ -93,7 +94,6 @@ func (e *EnvStackConfig) Template() (string, error) {
 		}
 		mft = string(out)
 	}
-
 	content, err := e.parser.ParseEnv(&template.EnvOpts{
 		AppName:                  e.in.App.Name,
 		DNSCertValidatorLambda:   dnsCertValidator,
@@ -106,6 +106,7 @@ func (e *EnvStackConfig) Template() (string, error) {
 		PrivateImportedCertARNs:  e.importPrivateCertARNs(),
 		VPCConfig:                e.vpcConfig(),
 		CustomInternalALBSubnets: e.internalALBSubnets(),
+		AllowVPCIngress:          e.in.AllowVPCIngress,
 		Telemetry:                e.telemetryConfig(),
 
 		Version:       e.in.Version,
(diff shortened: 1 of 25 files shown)


internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -233,7 +234,10 @@ func (e *EnvStackConfig) Parameters() ([]*cloudformation.Parameter, error) {
 	if len(e.in.ImportCertARNs) != 0 && len(e.in.ImportVPCConfig.PublicSubnetIDs) == 0 {
 		internalHTTPSListener = "true"
 	}
-
+	allowVPCIngress := "false"
+	if e.in.AllowVPCIngress {
+		allowVPCIngress = "true"
+	}
```
Lou1415926: Why do we prefer using `Parameters` over using go template rendering?

Lou1415926 (commented):
LGTM overall but I have one small question!

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -34,21 +34,31 @@ Parameters:
   CreateInternalHTTPSListener:
     Type: String
     AllowedValues: [true, false]
+  AllowVPCIngressForInternalALB:
```
iamhopaul123: Hmm any reason why we want to add a new param instead of using go template? I am worried about the fact that we add a new parameter that's relatively less likely to be used by users.

internal/pkg/deploy/cloudformation/stack/env.go
efekarakus: _I think_ (correct me if I'm wrong) if we don't use `Parameters` we will have to store the `--internal-alb-allow-vpc-ingress` value in SSM.
Otherwise, `env upgrade` will not have the `AllowVPCIngress` field set to `true` while upgrading and the SecurityGroupIngresses will get removed.

I am okay with storing it in SSM, but we will have to touch `upgradeEnvironment` to pass the value.
It's a bit more work that way, but it allows us to save a Parameter

Lou1415926: ehh that's true I think you are right. Let's stay away from SSM :this-is-fine-fire:

This could potentially be removed as a parameter to address [the concern raised by PH](https://github.com/aws/copilot-cli/pull/3646#discussion_r892939368) because this configuration (allow ingress) can be read from the env manifest.

efekarakus: But we have to store it at all times in SSM right? otherwise with `env show --manifest` we won't be able to generate a manifest file that says the security group should allow ingress from the VPC.

And if we're going to store in SSM might as well at that point not use the Parameter 🤔

Lou1415926: Good catch! Yup `env show --manifest` is going to 
1. Read from template `Metadata`
2. Read from SSM if 1. fails

It could be argued that `env show --manifest` can read the parameter values of `AllowVPCIngressForInternalALB` for this configuration specifically - that'd be ugly code, not desirable, but allows for faster/simpler implementation at this moment right now.

efekarakus: @huanjani @Lou1415926 @iamhopaul123 ok updated the existing PR with a new commit to use SSM and Go templates.

My only concern is not having the change done in https://github.com/aws/copilot-cli/blob/a8eca9d6bf7375b18ff94f65f4681c828d9d3cee/internal/pkg/manifest/env.go#L29 because we have to think about how the manifest would look like for it.
But we should at least be able to use this

Lou1415926 (approved):
LGTM!

iamhopaul123 (approved):

PR #3650, opened 2022-06-09
chore: uri for backend services on `svc deploy`

dannyrandall:


internal/pkg/describe/uri.go
```diff
@@ -46,22 +54,19 @@ func NewReachableService(app, svc string, store ConfigStoreSvc) (ReachableServic
 }
 
 // URI returns the LBWebServiceURI to identify this service uniquely given an environment name.
-func (d *LBWebServiceDescriber) URI(envName string) (string, error) {
+func (d *LBWebServiceDescriber) URI(envName string) (string, URIAccessType, error) {
 	svcDescr, err := d.initECSServiceDescribers(envName)
 	if err != nil {
-		return "", err
+		return "", 0, err
```
Lou1415926: nit: Instead of returning `0`, what do you think of returning a const `URIAccessTypeNone` or something like that?

internal/pkg/describe/uri.go
```diff
@@ -234,23 +277,27 @@ type nlbURI struct {
...
 
 func (u *LBWebServiceURI) String() string {
+	uris := u.albURI.uris()
+	for _, dnsName := range u.nlbURI.DNSNames {
+		uris = append(uris, fmt.Sprintf("%s:%s", dnsName, u.nlbURI.Port))
+	}
+	return english.OxfordWordSeries(uris, "or")
+}
+
+func (u *albURI) uris() []string {
```
Lou1415926: nit: do you think this can be named instead `string()` - that it's stringify-ing an `albURI` struct?

Lou1415926 (approved):
LGTM!

internal/pkg/describe/uri.go
```diff
@@ -13,13 +13,21 @@ import (
...
+
 var (
 	fmtSvcDiscoveryEndpointWithPort = "%s.%s:%s" // Format string of the form {svc}.{endpoint}:{port}
 )
 
-// ReachableService represents a service describe that has an endpoint.
+// ReachableService represents a service describer that has an endpoint.
 type ReachableService interface {
-	URI(env string) (string, error)
+	URI(env string) (string, URIAccessType, error)
```
iamhopaul123: Can we return a wrapped-up type so that we don't need to return three values for this function?

iamhopaul123: 🥺

internal/pkg/cli/svc_deploy.go
```diff
@@ -416,19 +416,22 @@ func (o *deploySvcOpts) uriRecommendedActions() ([]string, error) {
...
 	}
-	uri, err := describer.URI(o.envName)
+	uri, typ, err := describer.URI(o.envName)
 	if err != nil {
 		return nil, fmt.Errorf("get uri for environment %s: %w", o.envName, err)
 	}
 
 	network := "over the internet."
-	if o.svcType == manifest.BackendServiceType {
+	switch typ {
```
huanjani: nice

internal/pkg/describe/uri.go
efekarakus: I'm also a little worried with the 3 return types

I haven't read the PR so take my feedback with a grain of salt 😂 but I wonder if we can have something like this: https://go.dev/play/p/UeAeOWHH7nT 💭

dannyrandall: lmk if you have a better idea for the struct name :)

iamhopaul123 (approved):

PR #3657, opened 2022-06-10
chore(manifest): add validation and error message for when invalid autoscaling fields set

CaptainCarpensir:
Added a new error type in `errors.go` and create validation for the errors to give a clearer response when invalid scaling fields are set. An example of this is that copilot-cli will now give a descriptive error message when you try to set the `queue_delay` field in the manifest for a `Load Balanced Web Service`.


internal/pkg/manifest/svc.go
```diff
@@ -285,6 +285,29 @@ func (a *AdvancedCount) hasScalingFieldsSet() bool {
 	}
 }
 
+func (a *AdvancedCount) getInvalidFieldsSet() []string {
```
iamhopaul123, line 288: Can we do this in the other way around by reusing `validScalingFields()`? So that by default we throw incompatible error unless it is specifically defined as a valid one for this type of workload.

This can be a good reminder if we forget to update this function if we add more supports for some scaling policies in the future.

CaptainCarpensir: That would certainly simplify it. I made it like this because I wanted it to respond by saying exactly what scaling fields that are set are invalid, but users have the resources and are likely able to figure this out regardless. So in hindsight I agree that it's worth making the code simpler to have a slightly less informative error message. Will change 👍

CaptainCarpensir: Though looking more at it, how should we store all the invalid scaling fields? There's a list of which ones are valid for each type, but not a comprehensive list of scaling fields in total. Should I make a constant array of them to compare against?

iamhopaul123: Yeah i think also the reminder thing is what I am mostly interested in too. Because if we forget to update `getInvalidFieldsSet` then most likely users will not be aware of the issue whereas if we do it the other way around, before shipping this feature we would know because it errors out

CaptainCarpensir:
Penghao and I have looked at some alternative ways to do this. Please don't review this PR right now! I'll remove the `do-not-merge` label when I've pushed a new implementation to the branch that's ready for review.

internal/pkg/manifest/validate_test.go
```diff
@@ -1730,6 +1730,21 @@ func TestAdvancedCount_Validate(t *testing.T) {
...
+					Value: (*IntRangeBand)(aws.String("1-10")),
+				},
+				CPU: mockConfig,
+				QueueScaling: QueueScaling{
+					AcceptableLatency: durationp(10 * time.Second),
+					AvgProcessingTime: durationp(1 * time.Second),
+				},
+				workloadType: LoadBalancedWebServiceType,
+			},
+			wantedError: fmt.Errorf(`autoscaling field "queue_delay" is invalid with workload type Load Balanced Web Service`),
```
iamhopaul123, line 1745: Could you add an additional unit test case for multiple invalid fields?

internal/pkg/manifest/errors.go
```diff
@@ -72,6 +72,20 @@ func (e *errFieldMustBeSpecified) Error() string {
...
+	invalidFields []string
+	workloadType  string
+}
+
+func (e *errInvalidAutoscalingFieldsWithWkldType) Error() string {
+	quoted := make([]string, len(e.invalidFields))
+	for i, f := range e.invalidFields {
+		quoted[i] = strconv.Quote(f)
+	}
+	return fmt.Sprintf(`autoscaling %s %s %s invalid with workload type %s`, english.PluralWord(len(e.invalidFields), "field", "fields"),
```
iamhopaul123: Awesome!

internal/pkg/manifest/svc.go
```diff
@@ -272,17 +273,49 @@ func (a *AdvancedCount) validScalingFields() []string {
...
+	scalingFieldNames := []string{"CPU", "Memory", "Requests", "ResponseTime", "QueueScaling"}
+	validFieldsSet := []string{}
+	invalidFieldsSet := []string{}
+
+	validTags := a.validScalingFields()
+	typ := reflect.ValueOf(*a).Type()
+
+	for _, name := range scalingFieldNames {
+		currStructField, ok := typ.FieldByName(name)
+		if ok {
```
iamhopaul123: Maybe
```suggestion
		if !ok {
		  continue
		}
		...
```

internal/pkg/manifest/svc.go
```diff
@@ -272,17 +273,49 @@ func (a *AdvancedCount) validScalingFields() []string {
...
+			zero := reflect.Zero(currStructField.Type).Interface()
+			value := reflect.ValueOf(a).Elem().FieldByName(name).Interface()
+			if zero != value {
```
iamhopaul123: Similarly to reduce indented lines.
```suggestion
			if zero == value {
			  continue
			}
			...
```

internal/pkg/manifest/svc.go
```diff
@@ -272,17 +273,49 @@ func (a *AdvancedCount) validScalingFields() []string {
...
-	switch a.workloadType {
-	case LoadBalancedWebServiceType:
-		return !a.CPU.IsEmpty() || !a.Memory.IsEmpty() || !a.Requests.IsEmpty() || !a.ResponseTime.IsEmpty()
-	case BackendServiceType:
-		return !a.CPU.IsEmpty() || !a.Memory.IsEmpty() || !a.Requests.IsEmpty() || !a.ResponseTime.IsEmpty()
-	case WorkerServiceType:
-		return !a.CPU.IsEmpty() || !a.Memory.IsEmpty() || !a.QueueScaling.IsEmpty()
-	default:
-		return !a.CPU.IsEmpty() || !a.Memory.IsEmpty() || !a.Requests.IsEmpty() || !a.ResponseTime.IsEmpty() || !a.QueueScaling.IsEmpty()
+func (a *AdvancedCount) getValidScalingFieldsSet() ([]string, error) {
```
iamhopaul123:
```suggestion
func (a *AdvancedCount) validWkldScalingFields() ([]string, error) {
```

internal/pkg/manifest/svc.go
```diff
@@ -272,17 +273,49 @@ func (a *AdvancedCount) validScalingFields() []string {
...
+				}
+				if !isValid {
+					invalidFieldsSet = append(invalidFieldsSet, tag)
+				}
+			}
+		}
+	}
+
+	if len(invalidFieldsSet) > 0 {
+		return validFieldsSet,
```
iamhopaul123: It is a little weird for the return value to be non nil when an error is thrown. Maybe return nil?

internal/pkg/manifest/svc.go
```diff
@@ -272,17 +273,49 @@ func (a *AdvancedCount) validScalingFields() []string {
...
+			&errInvalidAutoscalingFieldsWithWkldType{
+				invalidFields: invalidFieldsSet,
+				workloadType:  a.workloadType,
+			}
 	}
+	return validFieldsSet, nil
+}
+
+func (a *AdvancedCount) hasScalingFieldsSet() bool {
+	fields, _ := a.getValidScalingFieldsSet()
```
iamhopaul123: We could do the error handling here.

internal/pkg/manifest/svc.go
CaptainCarpensir: I like continuing if the bool value returns not ok in the previous comment, but personally I find it less readable to have a bunch of cases that are for skipping, so in this case I'd prefer to keep nesting the if statement. Lmk if you really disagree with this, otherwise I'll keep it the same for now. 
This is what it looks like currently:
```
for _, name := range scalingFieldNames {
	currStructField, ok := typ.FieldByName(name)
	if !ok {
		continue
	}

	zero := reflect.Zero(currStructField.Type).Interface()
	value := reflect.ValueOf(a).Elem().FieldByName(name).Interface()

	if zero != value {
		tag := currStructField.Tag.Get("yaml")
		isValid := false

		for _, v := range validTags {
			if tag == v {
				validFieldsSet = append(validFieldsSet, tag)
				isValid = true
			}
		}
		if !isValid {
			invalidFieldsSet = append(invalidFieldsSet, tag)
		}
	}
}
```

internal/pkg/manifest/svc.go
CaptainCarpensir: How would you propose to do this? Make `hasScalingFieldsSet()` return both a bool and an error?

internal/pkg/manifest/svc.go
CaptainCarpensir: This change sounds less descriptive to me of what the function does. It makes it sound like it would then return the same information as `validScalingFields()` because that function already has workload implied in its name.

internal/pkg/manifest/svc.go
iamhopaul123: yeah because when `getValidScalingFieldsSet` throws an error, it doesn't make sense to proceed i guess 🤔

CaptainCarpensir: Hm, I agree with you, but running it down the line, this change makes a lot of the cases where `hasScalingFieldSet()` and `hasAutoscaling()` clunkier to use. This is because they're both generally used in one line boolean statements, and returning two values would end up making these all a lot more complicated. It ends up adding lines to check this error in multiple places, whereas the current implementation just checks for the error once.

CaptainCarpensir: Is there a way to maintain the easy to read one-line statements while ignoring the error check?

huanjani:
Awesome!
🤔 Just curious if docs updates for autoscaling cooldown support are coming in a future PR...? 🙏🏼

internal/pkg/manifest/svc.go
dannyrandall: do we use the returned `[]string` anywhere...? or just the `error`

Lou1415926: I'm not against the use of `reflect`, but am in general a little concerned. This package needs to be handled delicately (😂) to avoid scary panic. It also reduces code readability thus potentially maintainability. I know we've used it in `manifest/transform.go` but it was sort of impossible to avoid in that case :(. Ideally, we want to avoid using `reflect` as much as we can.

I can see why you use `reflect`: if we were to check invalid scaling fields for each workload type, then every time we introduce a new scaling field, we will need to add additional lines for every single workload type.

Just to brainstorm, I have this alternative: https://go.dev/play/p/R1_hJ2unLvX that I myself am not quite fond of because of the drilling (see how `advancedCountType` is drilled all the way from `TaskConfig` to `AdvancedCount`).

Ultimately I think the current solution is fine and am happy to approve after some nits! Just wanted to figure out if there's a better way to handle it.

CaptainCarpensir: Yes, the valid scaling fields are used as a replacement for `hasScalingFieldsSet()`. They're just returned as a slice in case any future uses would need to have information on specifically which ones are set.

Then, the point of changing this to use reflect was to consolidate the places where changes would need to be made when adding new fields. I feel like if we're too concerned about the potential dangers of the reflect package, then the initial implementation I made would be better suited than the drilling. The initial implementation just hard-coded the invalid fields. Once again it's not ideal, but in all, that solution is the most readable and simple. The `reflect` package however minimizes future changes, though it does add more complications and risk if the file were overhauled in some way.

Overall I think the trade-off between readability and reproducibility is worth it in this case, and am in favor of keeping the reflect package in this instance. I also think that one drawback to this solution was that it was harder to implement, but as I've already written the code for it, the initial implementation is no longer an issue.

internal/pkg/manifest/errors.go
```diff
@@ -72,6 +73,18 @@ func (e *errFieldMustBeSpecified) Error() string {
...
+type errInvalidAutoscalingFieldsWithWkldType struct {
+	invalidFields []string
+	workloadType  string
+}
+
+func (e *errInvalidAutoscalingFieldsWithWkldType) Error() string {
+	if len(e.invalidFields) == 1 {
+		return fmt.Sprintf("autoscaling field %v is invalid with workload type %v", e.invalidFields[0], e.workloadType)
+	}
+	return fmt.Sprintf("autoscaling field %v are invalid with workload type %v", strings.Join(e.invalidFields, "/"), e.workloadType)
```
Lou1415926: or...
```go
return fmt.Errorf("autoscaling %s %v %s invalid with workload type %v", 
english.PluralWord(len(e.invalidField), "field", "field"), 
strings.Join(e.invalidFields, "/"), 
english.PluralWord(len(e.invalidField), "is", "are")
```

dannyrandall: this but `english.PluralWord(len(e.invalidField), "field", "fields")` :blush:

dannyrandall (approved):

internal/pkg/manifest/errors.go
```diff
@@ -72,6 +72,16 @@ func (e *errFieldMustBeSpecified) Error() string {
...
 }
 
+type errInvalidAutoscalingFieldsWithWkldType struct {
+	invalidFields []string
+	workloadType  string
+}
+
+func (e *errInvalidAutoscalingFieldsWithWkldType) Error() string {
+	return fmt.Sprintf("autoscaling %v %v %v invalid with workload type %v", english.PluralWord(len(e.invalidFields), "field", "fields"),
+		english.WordSeries(e.invalidFields, "and"), english.PluralWord(len(e.invalidFields), "is", "are"), e.workloadType)
```
Lou1415926:
```suggestion
		english.WordSeries(template.QuoteSliceFunc(e.invalidFields), "and"), english.PluralWord(len(e.invalidFields), "is", "are"), e.workloadType)
```

Lou1415926 (approved):
just one nit! feel free to remove the label when you're ready!

PR #3666, opened 2022-06-14
fix: render sidecar configuration without requiring port

### LINKED ISSUE #3665: [Bug] "command" parameter on sidecars appears to be ignored
As far as I can tell the `command` parameter on sidecars (documented [here](https://aws.github.io/copilot-cli/docs/developing/sidecars/#command)) doesn't actually override the command in the image.

To test this I added a sidecar with this simple definition:

```
  test:
    image: alpine
    command: "while [ ! ]; do echo 'test' && sleep 1; done"
```

I would expect it to hang around perpetually spitting "test" into stdout.

What happened instead was the sidecar exited immediately, indicating that my command wasn't being passed to the container.

Also, changing the command parameter and trying to deploy again netted a "stack has no changes" error, so it seems like the command parameter is just completely ignored.

Lou1415926:
Fix #3665.

### DIFF
--- a/internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
+++ b/internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
@@ -44,8 +44,8 @@
 {{- if $sidecar.Essential}}
   Essential: {{$sidecar.Essential}}
 {{- end}}
+{{include "image-overrides" . | indent 2}}  
 {{- if $sidecar.Port}}
-{{include "image-overrides" . | indent 2}}
   PortMappings:
     - ContainerPort: {{$sidecar.Port}}
     {{- if $sidecar.Protocol}}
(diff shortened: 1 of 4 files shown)


internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
```diff
@@ -44,8 +44,8 @@
 {{- if $sidecar.Essential}}
   Essential: {{$sidecar.Essential}}
 {{- end}}
+{{include "image-overrides" . | indent 2}}
```
efekarakus, line 47: :cool_crying:

Would it be possible to add another sidecar here without `port` to make sure we have a case covering it 🥺: 
https://github.com/aws/copilot-cli/blob/a633641af728dadba33aa93bbb53491d5898037d/internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-manifest.yml#L81-L89

iamhopaul123 (approved):

efekarakus (approved):

PR #3667, opened 2022-06-14
feat: autoscaling cooldown support for all autoscaling fields

CaptainCarpensir:
Add autoscaling cooldown support for Requests, ResponseTime, and QueueScaling fields.

Finishes feature for #3165 


internal/pkg/manifest/transform.go
```diff
@@ -284,23 +287,23 @@ func (t advancedCountTransformer) Transformer(typ reflect.Type) func(dst, src re
 	}
 }
 
-type scalingConfigOrPercentageTransformer struct{}
+type scalingConfigOrTTransformer[T ~int | time.Duration] struct{}
```
iamhopaul123, line 290: awesome!

internal/pkg/template/workload.go
```diff
@@ -332,15 +332,18 @@ type Cooldown struct {
...
+	MaxCapacity      *int
+	CPU              *float64
+	Memory           *float64
+	Requests         *float64
+	ResponseTime     *float64
+	CPUCooldown      Cooldown
+	MemCooldown      Cooldown
+	ReqCooldown      Cooldown
+	RespTimeCooldown Cooldown
+	QDelayCooldown   Cooldown
```
iamhopaul123:
```suggestion
	QueueDelayCooldown   Cooldown
```

iamhopaul123 (approved):
🎉

internal/pkg/manifest/validate.go
```diff
@@ -964,13 +970,13 @@ func (qs QueueScaling) Validate() error {
...
 	}
-	if qs.AcceptableLatency == nil {
+	if qs.AcceptableLatency == nil && qs.AvgProcessingTime != nil {
 		return &errFieldMustBeSpecified{
 			missingField:      "acceptable_latency",
 			conditionalFields: []string{"msg_processing_time"},
 		}
 	}
-	if qs.AvgProcessingTime == nil {
+	if qs.AvgProcessingTime == nil && qs.AcceptableLatency != nil {
```
dannyrandall, line 979: i think this case is the same as the above one?

internal/pkg/template/templates/workloads/partials/cf/autoscaling.yml
```diff
@@ -263,6 +263,7 @@ PermissionToInvokeBacklogPerTaskCalculatorLambda:
     SourceArn: !GetAtt BacklogPerTaskScheduledRule.Arn
 
 {{- $acceptableBacklog := .Autoscaling.QueueDelay.AcceptableBacklogPerTask }}
+{{- $queueDelayCooldown := .Autoscaling.QueueDelayCooldown }}
```
dannyrandall, line 266: is there a reason for doing this here but not for `.Autoscaling.ReqCooldown` and `.Autoscaling.RespTimeCooldown`?

internal/pkg/manifest/svc.go
```diff
@@ -196,16 +196,16 @@ func (c *Count) Desired() (*int, error) {
 // Percentage represents a valid percentage integer ranging from 0 to 100.
 type Percentage int
 
-// ScalingConfigOrPercentage represents a resource that has autoscaling configurations or a default percentage.
-type ScalingConfigOrPercentage struct {
-	Value         *Percentage
-	ScalingConfig AdvancedScalingConfig // mutually exclusive with Value
+// ScalingConfigOrT represents a resource that has autoscaling configurations or a generic value.
+type ScalingConfigOrT[T ~int | time.Duration] struct {
```
dannyrandall, line 200: sweet!

internal/pkg/template/templates/workloads/partials/cf/autoscaling.yml
CaptainCarpensir: Two reasons, first is that it's used in two blocks of code for the `QueueDelayCooldown`, whereas `ReqCooldown` and `RespTimeCooldown` are only used in one block. The second reason is that because of how the go template package works, I don't think you can access the `QueueDelayCooldown` without having a variable defined for it (something to do with [this](https://github.com/aws/copilot-cli/blob/9383fb10a96695186260f6045ade6fc4d50017b2/internal/pkg/template/templates/workloads/partials/cf/autoscaling.yml#L289) line)
As such, I personally don't think it's necessary to define a variable for these (which should also include `MemCooldown` and `CPUCooldown` because they're only used in the same local area.

internal/pkg/manifest/validate.go
CaptainCarpensir: Since `IsEmpty()` for queue scaling now includes checking the cooldown, I have to specify that the `AcceptableLatency` isn't nil (and vise versa) as it should be valid for only the cooldown to be defined (it just won't do anything).

internal/pkg/manifest/transform.go
huanjani: 😍

internal/pkg/manifest/validate.go
dannyrandall: ah, gotcha - i was missing the `!=` nil switch :)

internal/pkg/template/templates/workloads/partials/cf/autoscaling.yml
dannyrandall: ah i see - i _think_ you can access the fields within the `range` by doing `$.Autoscaling.QueueDelayCooldown`. (adding a `$` to the beginning). i don't have any strong preferences either way - whichever you prefer!

dannyrandall (approved):
added dnm so that you can make changes if you want. if not feel free to remove the label!

PR #3668, opened 2022-06-14
feat: backend service alb http predictable alias

dannyrandall:
When using HTTP (not HTTPS) for a backend service with an internal ALB, create a hosted zone in form `env.app.internal` and insert an A record `svc.env.app.internal` into the hosted zone pointing to the internal ALB.

Also updates the backend service describer to pull the URI from the listener rule.


internal/pkg/describe/lb_web_service.go
```diff
@@ -32,11 +32,12 @@ const (
 	envOutputInternalLoadBalancerDNSName = "InternalLoadBalancerDNSName"
```
iamhopaul123, line 32: Is this still useful now that we don't display internal ALB DNS name at all.

iamhopaul123 (approved):
env version bump pls 🙏

internal/pkg/describe/lb_web_service.go
dannyrandall: i think we'll still need to fall back on it if someone has deployed an internal alb with http using the version prior to this change - the service won't have any listener rules configured with host header until they redeploy the service.

iamhopaul123: Do we display that in svc show or anywhere currently? because iirc url is only used after deployment or svc show. After deployment is probably fine. but it's ok im good for leaving it.

dannyrandall: yeah it's used in `svc show`, so that would fail for some services if we removed it

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -357,6 +357,23 @@ Resources:
         - CertificateArn: {{$arn}}
 {{- end}}
 {{- end}}
+  {{- if not .PrivateImportedCertARNs}}
+  InternalWorkloadsHostedZone:
+    Metadata:
+      'aws:copilot:description': 'Default hosted zone for workloads exposed with a private ALB'
```
efekarakus: Is there any chance we can render this description w the endpoint?
```suggestion
      'aws:copilot:description': 'A hosted zone for http://{{.Env}}.{{.App}}.internal for backends behind a private ALB'
```

internal/pkg/template/templates/workloads/partials/cf/http-listener.yml
```diff
@@ -1,3 +1,23 @@
+{{- if eq $.WorkloadType "Backend Service"}}
+LoadBalancerInternalDNSAlias:
+  Metadata:
+    'aws:copilot:description': 'Load balancer alias for service'
```
efekarakus: 😂 similarly can we possibly write the endpoint here:
```suggestion
    'aws:copilot:description': 'Alias for http://{{.Name}}.{{.Env}}.{{.App}.internal to the internal load balancer'
```

internal/pkg/template/templates/workloads/partials/cf/http-listener.yml
```diff
@@ -19,14 +39,23 @@ HTTPListenerRule:
...
-          {{- range $sourceIP := .AllowedSourceIps}}
-          - {{$sourceIP}}
-          {{- end}}
-    {{- end}}
+            {{- range $sourceIP := .AllowedSourceIps}}
+            - {{$sourceIP}}
+            {{- end}}
+      {{- end}}
+      {{- if eq $.WorkloadType "Backend Service"}}
+      - Field: 'host-header'
```
efekarakus, line 50: If a client today (v1.19) has a backend manifest like this:
```
http:
  path: /api
```
and uses the LB dns name to make requests `http://internal-name-123456789.region.elb.amazonaws.com/api` then due to the host-header config these requests won't match the rule. 

Looking at this fancy fancy API: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-listenerrule-hostheaderconfig.html

Should we accept the InternalALBDNS as well as an input to `Values` so that it can match both

efekarakus (commented):
yayayay

internal/pkg/deploy/cloudformation/stack/backend_svc_test.go
```diff
@@ -219,6 +219,9 @@ Outputs:
 				m.EXPECT().Read(envControllerPath).Return(&template.Content{Buffer: bytes.NewBufferString("something")}, nil)
 				m.EXPECT().ParseBackendService(gomock.Any()).DoAndReturn(func(actual template.WorkloadOpts) (*template.Content, error) {
 					require.Equal(t, template.WorkloadOpts{
+						AppName:      "phonetool",
```
efekarakus, line 222: wow I am glad the DoAndReturn got merged first 🤣

internal/pkg/template/templates/workloads/partials/cf/http-listener.yml
dannyrandall: Great catch! I've added the internal ALB DNS to `Values` :blush:.

efekarakus (approved):
😍

PR #3669, opened 2022-06-14
chore: make windows bin md5 sum file name match others

dannyrandall:
All of the other md5 sums of our output resources are named `filename.md5` except for windows, which has `.exe` stripped off the filename. We don't strip off the `.exe` for the signature files (`.asc`) on windows, so this makes the windows md5 sum match the windows signature file and the rest (linux & mac) of the resources.

### DIFF
--- a/.release/buildspec_stage.yml
+++ b/.release/buildspec_stage.yml
@@ -22,7 +22,7 @@ phases:
       - |
         for artifact in `cat ./$GIT_COMMIT_ID.manifest`
         do
-            md5FileName="${artifact%.exe}"
+            md5FileName="${artifact}"
             md5sum $artifact | sed 's/ .*//' > "$md5FileName.md5"
             echo "Generated m5d for $artifact: $(cat $md5FileName.md5)"
             aws s3 cp $artifact "s3://$STAGE_BUCKET/$artifact" || exit 1
--- a/.release/buildspec_integ.yml
+++ b/.release/buildspec_integ.yml
@@ -36,7 +36,7 @@ phases:
             do
                 bn="$(basename $artifact)"
                 echo "bn: $bn"
-                md5FileName="${{bn%.exe}}.md5"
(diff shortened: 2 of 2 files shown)


.release/buildspec_stage.yml
```diff
@@ -22,7 +22,7 @@ phases:
       - |
         for artifact in `cat ./$GIT_COMMIT_ID.manifest`
         do
-            md5FileName="${artifact%.exe}"
```
efekarakus, line 25: im so confused - we used to have ".exe" at the end of all md5 files? but that's clearly not the case here: https://github.com/aws/copilot-cli/releases what am I misunderstanding?

efekarakus, line 25: im so confused - we used to have ".exe" at the end of all md5 files? but that's clearly not the case here: https://github.com/aws/copilot-cli/releases what am I misunderstanding?

dannyrandall: so for some reason on the windows file, we trim off the `.exe`, but only for the md5 - in the `.asc` file we leave it on. i just want them to be the same :sob:. 

so today our outputs are:
```
copilot-windows.exe
copilot-windows.exe.asc
copilot-windows.md5
```

after this change (theoretically):
```
copilot-windows.exe
copilot-windows.exe.asc
copilot-windows.exe.md5
```

efekarakus: ok my bad! I didn't realize "%.exe" removed ".exe" 😂 now evreything makes sense

efekarakus: ok my bad! I didn't realize "%.exe" removed ".exe" 😂 now evreything makes sense

efekarakus (approved):

huanjani (approved):

iamhopaul123 (approved):

PR #3670, opened 2022-06-14
chore(stack): add CustomResourcesURL to RuntimeConfig

efekarakus:


internal/pkg/deploy/cloudformation/stack/backend_svc_test.go
```diff
@@ -217,41 +217,45 @@ Outputs:
...
+							StartPeriod: aws.Int64(0),
+							Timeout:     aws.Int64(10),
+						},
+						HostedZoneAliases: make(template.AliasesForHostedZone),
+						HTTPHealthCheck: template.HTTPHealthCheckOpts{
+							HealthCheckPath: manifest.DefaultHealthCheckPath,
+							GracePeriod:     aws.Int64(manifest.DefaultHealthCheckGracePeriod),
+						},
+						DeregistrationDelay: aws.Int64(60), // defaults to 60
+						CustomResources:     make(map[string]template.S3ObjectLocation),
```
efekarakus, line 236: The only thing added is this field. I've just changed the tests to use `DoAndReturn` and use `require.Equal` to compare objects as it works better than `gomock`.

`Gomock` would return an error when the field's location is different than expected

dannyrandall: great change this was so tedious before lol

dannyrandall (approved):
🥳

Lou1415926 (approved):
LGTM!

Issue #3654, opened 2022-06-10
avoid docker engine check when unnecessary


huanjani:
If users are bringing their own built image, they should not have to have their docker engine running. 

Currently, they can use the `--image` flag or follow the prompt for image triggered by no docker engine detected.

However, in the case that a user has an existing manifest with the `image` field populated, they can't get past our prompts and errors about docker engines. 

If they don't use the `--image` flag and then hit 'Enter' when prompted for it, we should trigger `RequireNonEmpty()` instead of erroring out about the docker engine. 

If we find an existing manifest, we should parse the `image` value out of it, as we do for `healthcheck`. 

Related: #3639

PR #3671, opened 2022-06-15
fix(cli): remove docker dependency from svc init

### LINKED ISSUE #3639: Remove docker dependency
We don't use copilot to build docker images (images are built in another CI process), but copilot fails on `copilot svc init` because docker is missing: `Cannot connect to the Docker daemon at unix:///var/run/docker.sock`

Lou1415926:
Fixes #3639 


huanjani:
Related: #3654

internal/pkg/cli/svc_init.go
```diff
@@ -288,7 +289,7 @@ func (o *initSvcOpts) Execute() error {
 		}
 	}
 	// If the user passes in an image, their docker engine isn't necessarily running, and we can't do anything with the platform because we're not building the Docker image.
-	if o.image == "" {
+	if o.image == "" && !o.manifestExists {
```
paragbhingre, line 292: Does Copilot today assumes if manifest is provided then image.location has to be there? Just trying to understand if we need to check for manifest itself or image.location field within manifest also?

Lou1415926: Given user input on image, dockerfile, etc., `svc init` creates a manifest for users to start with. I do not feel like it's a requirement that they have to have provided an image during `svc init` - either through prompts, flags, or the existing manifest. 🤔

iamhopaul123 (approved):

internal/pkg/cli/job_init.go
```diff
@@ -222,7 +224,7 @@ func (o *initJobOpts) Execute() error {
 		}
 	}
 	// If the user passes in an image, their docker engine isn't necessarily running, and we can't do anything with the platform because we're not building the Docker image.
-	if o.image == "" {
+	if o.image == "" && !o.manifestExists {
```
efekarakus, line 227: Can we also fix the first scenario here: https://github.com/aws/copilot-cli/issues/3639#issuecomment-1150406758 (where users enter empty string "" on Image prompt) and the CLI blows up. We should validate that empty string isn't valid 🥺

Lou1415926: > We should validate that empty string isn't valid 🥺

Could it be that they just want to write the `image` field later directly in the manifest? Basically I'm a little hesitant on making `image` basically a required information at `svc init` 🤔

efekarakus: I feel like --dockerfile xor --image should be required if there isn't a manifest 🤔 
They can still change it afterwards, but doing `copilot deploy` after a `svc init` should ideally just work. Was there a scenario that you had in mind?

Lou1415926: > doing copilot deploy after a svc init should ideally just work.

This makes sense to me

> Was there a scenario that you had in mind?

Perhaps if the user wants to build & use their own image and they haven't built the image they want to use yet 👀. I personally also feel more comfortable editing configurations in an editor, instead of in CLI prompts.

A clear pro of having them required though is like you said about the deployment flow, and it reenforces that `image` is a required manifest configuration.

internal/pkg/cli/svc_init.go
```diff
@@ -393,7 +393,7 @@ func (o *initSvcOpts) askImage() error {
 		return nil
 	}
 
-	var validator prompt.ValidatorFunc
+	validator := validateNonEmptyString
```
efekarakus: sorry my bad - I should have let you know! I think we have a validator already in the `prompt` pkg: 
https://github.com/aws/copilot-cli/blob/cc94d5ea457366a738dcb347468b29b09876b46f/internal/pkg/term/prompt/prompt.go#L343-L345

efekarakus (approved):

PR #3677, opened 2022-06-16
chore: delete no longer used inlined ZipFiles

efekarakus:


huanjani (approved):
:shipit:

internal/pkg/template/workload.go
```diff
@@ -510,6 +510,7 @@ type WorkloadOpts struct {
 	HTTPVersion              *string
 	ALBEnabled               bool
 	HostedZoneAliases        AliasesForHostedZone
+	CredentialsParameter     string
```
Lou1415926, line 513: qq: perhaps i missed something - what is this for?

internal/pkg/template/templates/workloads/partials/cf/vpc-connector.yml
```diff
@@ -24,7 +24,7 @@ EnvironmentSecurityGroupIngressFromServiceSecurityGroup:
 
 VpcConnector:
   Type: AWS::AppRunner::VpcConnector
-{{- if ne .EnvControllerLambda ""}}
+{{- if gt (len (envControllerParams .)) 0}}
```
Lou1415926, line 27: I see, for RDWS , `{{- if ne .EnvControllerLambda ""}}` and `{{- if gt (len (envControllerParams .)) 0}}` are effectively equivalent because both are true iff network placement is private.

internal/pkg/template/workload.go
efekarakus: The parameter already existed at l 527, I just moved it up :D 

It's the private registry creds I believe

Lou1415926: woop me blind!

Lou1415926 (approved):

Issue #3664, opened 2022-06-13
[Design] Remove default security groups


gautam-nutalapati:
Proposal for #3387

Use `deny_default_security_group` key to indicate service should not use default security group.

Example:
```
    network:
      vpc:
        placement: 'private'
        security_groups: ['sg-xxxxxxxxxxxxxxxxx']
        deny_default_security_group: true

```
```
    network:
      vpc:
        deny_default_security_group: true
```

- When `deny_default_security_group` is `false`, Current behavior is not modified and Default EnvironmentSecurityGroup is applied.
- When `deny_default_security_group` or `vpc` OR `network` keys are not present in yaml, `deny_default_security_group` is `false` by default. Current behavior is not modified, and Default EnvironmentSecurityGroup is applied.
- When `deny_default_security_group` is `true`, Default EnvironmentSecurityGroup is NOT applied.

Lou1415926:
Hello @gautam-nutalapati ! Thank you very much for the proposal!

I see that you are looking to remove the `EnvironmentSecurityGroup` from a service. If I'm not mistaken, you'd like to scope down the service's permissions that's set up by security group.

With this sg removed, the service would not be able to access a majority of the environment resources, neither would it be accessible from most of these environment resources. Does this align with your use case?

An alternative I could think of is modifying the ingress of the `EnvironmentSecurityGroup` 🤔 . This would affect all services deployed in the environment, though.

gautam-nutalapati:
Hi @Lou1415926!

Your understanding is correct. We would like to scope down service permissions. 
We want to prevent egress, and only for a few services. For rest of the services, svc-to-svc and egress communications are allowed (current approach of Copilot).

As this is for subset of services, a `service specific security group` that allows `ingress` from `EnvironmentSecurityGroup`, and `egress` to VPC endpoints is created in add-ons. We can also allow `egress` to `EnvironmentSecurityGroup` if needed (just not our use case).

This can be also useful to control which services can talk with each other. (avoiding `EnvironmenSecurityGroup` for services that just need to talk to each other) etc.

We don't want to prevent egress for all services. 
But if users want to prevent egress completely, like you said, we can modify `EnvironmenSecurityGroup` security group. 
Or use this flag along with a custom `EnvironmenSecurityGroup` specified in `network/vpc/security_groups`.

I think service specific flag approach is more flexible.
Update: 
[...]

Issue #3387, opened 2022-03-22
Remove default security groups


honeydew-health:
The security policy in our organization to restrict outbound traffic to specific IP addresses. Therefore we would like to be able to remove the default security group setup by Copilot because it allows outbound traffic to everywhere.

Please could this feature can be considered to be added in a future update?

Thanks, Roger

Lou1415926:
Hello @honeydew-health !

Thank you very much for the request! This sounds good to me. 

Just to confirm, you are looking to modify the security group with name ending with "EnvironmentSecurityGroup", is that correct?

For a temporary workaround, would it work for you if you manually update the security group outbound rule through console or aws-cli?

honeydew-health:
Hello @Lou1415926 
Thanks for getting back to me. Yes I'm referring to the security group with 'description' ending  "EnvironmentSecurityGroup".

Yes as a temporary workaround we could remove the default security group using CLI or web console, but this is not ideal as we use CodePilot to automate deployment of app updates to our containers when changes are pushed to GitHub. So ideally we could set a value in manifest.yml to prevent creating the default security groups.

PR #3682, opened 2022-06-21
feat: Add ability to deny environment security group

gautam-nutalapati:
Related to https://github.com/aws/copilot-cli/issues/3387
 - This change adds option to deny EnvironmentSecurityGroup which is applied by default on copilot service
 - Default security group is still applied by default like before

Design https://github.com/aws/copilot-cli/issues/3664
Thank you @Lou1415926 for your input on the detailed design!!


internal/pkg/manifest/workload.go
```diff
@@ -441,14 +441,78 @@ func (p *PlacementArgs) isEmpty() bool {
...
 type PlacementString string
 
+// SecurityGroupsIDsOrConfig represents security groups attached to task. It supports unmarshalling
+// yaml which can either be of type SecurityGroupsConfig or a list of strings.
+type SecurityGroupsIDsOrConfig struct {
+	SecurityGroupIds     []string `yaml:"security_groups"`
+	SecurityGroupsConfig SecurityGroupsConfig
+}
+
+func (s *SecurityGroupsIDsOrConfig) IsEmpty() bool {
```
iamhopaul123: Can we have a comment for public func? Same thing below.

internal/pkg/manifest/workload.go
```diff
@@ -441,14 +441,78 @@ func (p *PlacementArgs) isEmpty() bool {
...
+	return (s.SecurityGroupIds == nil || len(s.SecurityGroupIds) == 0) && s.SecurityGroupsConfig.isEmpty()
```
iamhopaul123:
```suggestion
	return len(s.SecurityGroupIds) == 0 && s.SecurityGroupsConfig.isEmpty()
```
since `s.SecurityGroupIds` is `[]string`. Same thing below.

internal/pkg/manifest/workload.go
```diff
@@ -441,14 +441,78 @@ func (p *PlacementArgs) isEmpty() bool {
...
+	}
+
+	if !s.SecurityGroupsConfig.isEmpty() {
+		// Unmarshalled successfully to s.SecurityGroupsConfig, unset s.SecurityGroupIds, and return.
+		s.SecurityGroupIds = nil
+		return nil
+	}
+
+	if err := value.Decode(&s.SecurityGroupIds); err != nil {
+		return errUnmarshalPlatformOpts
```
iamhopaul123: Can we give it a dedicated error msg?

iamhopaul123 (approved):
🎉

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
```diff
@@ -47,7 +47,9 @@ NetworkConfiguration:
         - Fn::ImportValue: !Sub '${AppName}-${EnvName}-{{.Network.SubnetsType}}'
     {{- end}}
     SecurityGroups:
+      {{- if not .Network.DenyDefaultSecurityGroup}}
```
Lou1415926, line 50: Nice! Just to put my my thoughts here in case others are wondering.

The deployment will error out if there turns out to be zero security groups (i.e. the default is denied, no custom sg, no NLB and no addon sg - should be rare), because the template is malformed with `SecurityGroups: ` field having `null` value.

However, if we remove `SecurityGroups` field (so template isn't malformed and the deployment goes through), the [default security group of the VPC will be used](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/default-custom-security-groups.html), which I think we want to avoid. So I think keeping `SecurityGroups` field here and just let it error out if no sg is left is the better option!

internal/pkg/template/template_integration_test.go
```diff
@@ -420,6 +420,28 @@ func TestTemplate_ParseNetwork(t *testing.T) {
      - Fn::ImportValue: !Sub '${AppName}-${EnvName}-EnvironmentSecurityGroup'
      - "sg-1bcf1d5b"
      - "sg-asdasdas"
+`,
+		},
+		"should render AWS VPC configuration without default environment security group": {
```
Lou1415926, line 425: Thanks for adding this test!

internal/pkg/manifest/workload.go
```diff
@@ -441,14 +441,78 @@ func (p *PlacementArgs) isEmpty() bool {
...
+
+func (s *SecurityGroupsIDsOrConfig) IsEmpty() bool {
+	return (s.SecurityGroupIds == nil || len(s.SecurityGroupIds) == 0) && s.SecurityGroupsConfig.isEmpty()
+}
+
+// SecurityGroupsConfig represents which security groups are attached to a task
+// and if default security group is applied.
+type SecurityGroupsConfig struct {
+	SecurityGroups []string `yaml:"groups"`
+	DenyDefault    bool     `yaml:"deny_default"`
```
Lou1415926: I think we probably want to use `*bool` for this field so that environment override works properly. 
```
network:
  vpc:
    security_groups: 
      deny_default: true
      groups: ['sg-1']

environments:
  test:
    network:
      vpc:
        security_groups:
          deny_default: false
          groups: ['sg-2']
```
The expected manifest after applying environment override should be something like:
```
network.vpc.securityGroups.DenyDefault = false
network.vpc.securityGroups.Groups = ['sg-2']
```
However, using `bool` instead of `*bool` will result in
```
network.vpc.securityGroups.DenyDefault = true
network.vpc.securityGroups.Groups = ['sg-2']
```
because `mergo` is told "not to override with zero value".

internal/pkg/manifest/workload.go
```diff
@@ -441,14 +441,78 @@ func (p *PlacementArgs) isEmpty() bool {
 // PlacementString represents what types of subnets (public or private subnets) to place tasks.
 type PlacementString string
 
+// SecurityGroupsIDsOrConfig represents security groups attached to task. It supports unmarshalling
+// yaml which can either be of type SecurityGroupsConfig or a list of strings.
+type SecurityGroupsIDsOrConfig struct {
+	SecurityGroupIds     []string `yaml:"security_groups"`
```
Lou1415926: Do we need the tag? Since we already have `"security_groups"` YAML tag for `SecurityGroupsIDsOrConfig` I worried this would require a manifest of
```
network:
  vpc:
    security_groups:
      security_groups: "..."
```
for it to unmarshal.

Lou1415926: What do you think of renaming the fields to
```
type SecurityGroupsIDsOrConfig struct {
  IDs []string
  AdvancedConfig SecurityGroupsConfig
}

type vpcConfig struct {
  SecurityGroups SecurityGroupsIDOrConfig
}
```
so that in callers it reads
```
v := &vpcConfig{...}
ids := v.SecurityGroups.ID
config := v.SecurityGroups.AdvancedConfig
```

internal/pkg/manifest/validate.go
```diff
@@ -1277,6 +1280,19 @@ func (p PlacementString) Validate() error {
 	return fmt.Errorf(`"placement" %s must be one of %s`, string(p), strings.Join(subnetPlacements, ", "))
 }
 
+// Validate is a no-op for SecurityGroupsIDsOrConfig.
```
Lou1415926, line 1283: Thanks for adding the `Validate`!

Lou1415926 (commented):
Yayy thank you very much for the. contribution! I have a few comments below but nothing major!

internal/pkg/manifest/workload.go
gautam-nutalapati: changed the func to private. It was not required as public.

internal/pkg/manifest/workload.go
gautam-nutalapati: change is made in this file at all places.

internal/pkg/manifest/workload.go
gautam-nutalapati: done

internal/pkg/manifest/workload.go
gautam-nutalapati: Thank you for pointing this out! It was a problem indeed. It is changed to `*bool`.
I am newbie in goLang. Please let me know if the way I wrote the code/tests for this change is in par with coding structure that copilot follows.

internal/pkg/manifest/workload.go
gautam-nutalapati: suggested change sounds better. They are changed as suggested.

internal/pkg/manifest/lb_web_svc_test.go
```diff
@@ -727,7 +736,60 @@ func TestLoadBalancedWebService_ApplyEnv(t *testing.T) {
...
+					},
+				},
+				Environments: map[string]*LoadBalancedWebServiceConfig{
+					"prod-iad": {
+						Network: NetworkConfig{
+							VPC: vpcConfig{
+								SecurityGroups: SecurityGroupsIDsOrConfig{
+									AdvancedConfig: SecurityGroupsConfig{
+										SecurityGroups: []string{"sg-456", "sg-700"},
+										DenyDefault:    &trueValue,
```
Lou1415926: nit: we can use `aws.Bool` to take the pointer! I've also added suggested changes to related places so that you can just apply the changes if you want 👀 !
```suggestion
										DenyDefault:    aws.Bool(true),
```

internal/pkg/manifest/lb_web_svc_test.go
```diff
@@ -727,7 +736,60 @@ func TestLoadBalancedWebService_ApplyEnv(t *testing.T) {
...
+				LoadBalancedWebServiceConfig: LoadBalancedWebServiceConfig{
+					Network: NetworkConfig{
+						VPC: vpcConfig{
+							Placement: PlacementArgOrString{
+								PlacementString: placementStringP(PublicSubnetPlacement),
+							},
+							SecurityGroups: SecurityGroupsIDsOrConfig{
+								AdvancedConfig: SecurityGroupsConfig{
+									SecurityGroups: []string{"sg-456", "sg-700"},
+									DenyDefault:    &trueValue,
```
Lou1415926: nit: related to using `aws.Bool` to take the pointer
```suggestion
									DenyDefault:    aws.Bool(true),
```

internal/pkg/manifest/lb_web_svc_test.go
```diff
@@ -278,6 +278,7 @@ func TestLoadBalancedWebService_ApplyEnv(t *testing.T) {
 		mockConfig = ScalingConfigOrT[Percentage]{
 			Value: &perc,
 		}
+		trueValue = true
```
Lou1415926: nit: related to using `aws.Bool` to take the pointer

```suggestion
```

internal/pkg/manifest/workload.go
```diff
@@ -37,11 +37,12 @@ var (
...
 	ErrAppRunnerInvalidPlatformWindows = errors.New("Windows is not supported for App Runner services")
 
-	errUnmarshalBuildOpts     = errors.New("unable to unmarshal build field into string or compose-style map")
-	errUnmarshalPlatformOpts  = errors.New("unable to unmarshal platform field into string or compose-style map")
-	errUnmarshalPlacementOpts = errors.New("unable to unmarshal placement field into string or compose-style map")
-	errUnmarshalCountOpts     = errors.New(`unable to unmarshal "count" field to an integer or autoscaling configuration`)
-	errUnmarshalRangeOpts     = errors.New(`unable to unmarshal "range" field`)
+	errUnmarshalBuildOpts         = errors.New("unable to unmarshal build field into string or compose-style map")
+	errUnmarshalPlatformOpts      = errors.New("unable to unmarshal platform field into string or compose-style map")
+	errUnmarshalSecurityGroupOpts = errors.New(`unable to unmarshal "security groups" field into slice of strings or compose-style map`)
```
Lou1415926: nit:
```suggestion
	errUnmarshalSecurityGroupOpts = errors.New(`unable to unmarshal "security_groups" field into slice of strings or compose-style map`)
```

internal/pkg/manifest/workload.go
```diff
@@ -441,14 +442,82 @@ func (p *PlacementArgs) isEmpty() bool {
...
+
+	if err := value.Decode(&s.IDs); err != nil {
+		return errUnmarshalSecurityGroupOpts
+	}
+	return nil
+}
+
+// GetIDs returns security groups from SecurityGroupsIDsOrConfig that are attached to task.
+// nil is returned if no security groups are specified.
+func (s *SecurityGroupsIDsOrConfig) GetIDs() []string {
```
Lou1415926, line 494: nit:

Referencing the `Getters` naming convention in [Effective Go](https://go.dev/doc/effective_go#Getters), what do you think of renaming it to `IDs`?

internal/pkg/manifest/workload.go
```diff
@@ -441,14 +442,82 @@ func (p *PlacementArgs) isEmpty() bool {
...
+	if !s.AdvancedConfig.isEmpty() && len(s.AdvancedConfig.SecurityGroups) > 0 {
+		return s.AdvancedConfig.SecurityGroups
+	}
+	if len(s.IDs) > 0 {
+		return s.IDs
+	}
+	return nil
```
Lou1415926: nit: what do you think of
```suggestion
	if !s.AdvancedConfig.isEmpty() {
	    return s.AdvancedConfig.SecurityGroups
	}
	return s.IDs
```
which should be effectively the same as what you have right now!

Just to share my personal thoughts on why I made the suggestion: the case where both `IDs` and `AdvancedConfig` are non-empty sounds like a bug from the upper stream (bugs in functions that are designed to be executed in an earlier stage e.g. in `ApplyEnv`, `Validation`. etc.). In (downtream) util methods like this, I feel like it's reasonable to assume that the upper stream implementation is correct - and that either `IDs` or `AdvancedConfig`, not both, is non-empty. One major pros imo is that it'd be easier to locate a bug if a bug does exists in the upper stream... Just my own thoughts though!

internal/pkg/manifest/workload.go
```diff
@@ -441,14 +442,82 @@ func (p *PlacementArgs) isEmpty() bool {
...
+}
+
+// IsDefaultSecurityGroupDenied returns true if DenyDefault is set to true
+// in SecurityGroupsIDsOrConfig.AdvancedConfig. Otherwise, false is returned.
+func (s *SecurityGroupsIDsOrConfig) IsDefaultSecurityGroupDenied() bool {
+	if !s.AdvancedConfig.isEmpty() && s.AdvancedConfig.DenyDefault != nil {
+		return *s.AdvancedConfig.DenyDefault
+	}
+	return false
```
Lou1415926: nit: this should be effectively the same as yours but is more aligned with the rest of the codebase! `aws.BoolValue` returns the referenced value given a pointer, and `false` if the pointer is `nil`

```suggestion
	if !s.AdvancedConfig.isEmpty() {
		return aws.BoolValue(s.AdvancedConfig.DenyDefault)
	}
	return false
```

internal/pkg/manifest/workload_test.go
```diff
@@ -649,7 +651,117 @@ func TestNetworkConfig_IsEmpty(t *testing.T) {
 	}
 }
 
+func TestSecurityGroupsConfig_UtilityFuncForSecurityGroups(t *testing.T) {
```
Lou1415926: nit:
```suggestion
func TestSecurityGroupsConfig_IDs(t *testing.T) {
```
or 
```suggestion
func TestSecurityGroupsConfig_GetIDs(t *testing.T) {
```

internal/pkg/manifest/workload_test.go
```diff
@@ -649,7 +651,117 @@ func TestNetworkConfig_IsEmpty(t *testing.T) {
...
+					SecurityGroups: []string{"group", "group1"},
+				},
+			},
+			wantedSecurityGroups: []string{"group", "group1"},
+		},
+		"nil returned when security groups in map are empty": {
+			data: SecurityGroupsIDsOrConfig{
+				AdvancedConfig: SecurityGroupsConfig{
+					SecurityGroups: []string{},
+				},
```
Lou1415926, line 675: also just nit: would this be a better example of what could take place?
```suggestion
				AdvancedConfig: SecurityGroupsConfig{
					DenyDefault: aws.Bool(bool)
				},
```

Since in go, a field of type `[]sting` gets initialized as `nil` instead of `[]string{}`, a non-nil slice with a length of 0.

internal/pkg/manifest/workload_test.go
```diff
@@ -649,7 +651,117 @@ func TestNetworkConfig_IsEmpty(t *testing.T) {
...
+			// WHEN
+			sgs := tc.data.GetIDs()
+
+			// THEN
+			require.Equal(t, tc.wantedSecurityGroups, sgs)
+		})
+	}
+}
+
+func TestSecurityGroupsConfig_UtilityFuncForDefaultSecurityGroup(t *testing.T) {
```
Lou1415926: nit:
```suggestion
func TestSecurityGroupsConfig_IsDefaultSecurityGroupDenied(t *testing.T) {
```

internal/pkg/manifest/workload_test.go
```diff
@@ -649,7 +651,117 @@ func TestNetworkConfig_IsEmpty(t *testing.T) {
...
+	var (
+		falseValue = false
+		trueValue  = true
+	)
+	testCases := map[string]struct {
+		data              SecurityGroupsIDsOrConfig
+		IsDefaultSGDenied bool
```
Lou1415926: nit: just to align with the rest of the tests in the codebase
```suggestion
                 in SecurityGroupsIDsOrConfig
		wanted bool
```

internal/pkg/manifest/workload_test.go
```diff
@@ -649,7 +651,117 @@ func TestNetworkConfig_IsEmpty(t *testing.T) {
...
+	}{
+		"default security group is applied when no vpc security config is present": {
+			data:              SecurityGroupsIDsOrConfig{},
+			IsDefaultSGDenied: false,
```
Lou1415926: tiny tiny nit:
```suggestion
			IsDefaultSGDenied: false,
```
`data` is going to be initialized as `SecurityGroupsIDsOrConfig{}` so no need to assign it!

internal/pkg/manifest/workload_test.go
```diff
@@ -649,7 +651,117 @@ func TestNetworkConfig_IsEmpty(t *testing.T) {
...
+					SecurityGroups: []string{"1"},
+				},
+			},
+			IsDefaultSGDenied: false,
+		},
+		"default security group is applied when deny_default is false in SG config": {
+			data: SecurityGroupsIDsOrConfig{
+				AdvancedConfig: SecurityGroupsConfig{
+					SecurityGroups: []string{"1"},
+					DenyDefault:    &falseValue,
```
Lou1415926: nit: similar to the other `aws.Bool` nits!
```suggestion
					DenyDefault:    aws.Bool(false),
```

internal/pkg/manifest/workload_test.go
```diff
@@ -649,7 +651,117 @@ func TestNetworkConfig_IsEmpty(t *testing.T) {
...
+		"default security group is applied when security group array is specified": {
+			data: SecurityGroupsIDsOrConfig{
+				IDs: []string{"1"},
+			},
+			IsDefaultSGDenied: false,
+		},
+		"default security group is not applied when default_deny is true": {
+			data: SecurityGroupsIDsOrConfig{
+				AdvancedConfig: SecurityGroupsConfig{
+					DenyDefault: &trueValue,
```
Lou1415926:
```suggestion
					DenyDefault: aws.Bool(true),
```

internal/pkg/manifest/workload_test.go
```diff
@@ -679,7 +791,51 @@ network:
...
 					},
-					SecurityGroups: []string{"sg-1234", "sg-4567"},
+					SecurityGroups: SecurityGroupsIDsOrConfig{
+						IDs:            []string{"sg-1234", "sg-4567"},
+						AdvancedConfig: SecurityGroupsConfig{},
+					},
+				},
+			},
+		},
+		"unmarshal is successful for security groups specified in config": {
```
Lou1415926, line 790: love these tests!! super nice to visualize how manifest is going to be unmarshalled!

Lou1415926 (approved):
Thank you again for addressing my last comments! I just have a few more nits mostly just to align with the codebase. 15 comments seem a lot but I promise they are just nits and renaming!

Approved with the DNM label in case you'd like to address the nits. Also, please let me know if you need any help with the feedback! I'd be happy to jump in and add additional commits to the PR ❤️ !

internal/pkg/manifest/workload.go
gautam-nutalapati: @Lou1415926 suggestion sounds good but there's a conflict.
changing method to `IDs` conflicts with [field](https://github.com/gautam-nutalapati/copilot-cli/blob/83c3e587ac68c18ec7fab4ba0e6428774aafcf60/internal/pkg/manifest/workload.go#L448) in struct:
```
type SecurityGroupsIDsOrConfig struct {
	IDs            []string
	AdvancedConfig SecurityGroupsConfig
}
```
I can't change IDs to lowercase as its accessed in [test](https://github.com/gautam-nutalapati/copilot-cli/blob/83c3e587ac68c18ec7fab4ba0e6428774aafcf60/internal/pkg/deploy/cloudformation/stack/backend_svc_test.go#L386) of `stack` module.

I can make test refer to AdvancedConfig, and change `IDs` field to `ids`, and then rename the method as you suggested, but that looks hacky.
Or we can rename IDs for `identifiers`. 
Let me know what sounds better, or any other options that'll be good.
Update:
I am ok to leave it as `GetIDs` if it's not a critical thing.

gautam-nutalapati:
@Lou1415926 Thank you so much for nits. I applied all suggested changes, except the change to `GetIDs`.

internal/pkg/manifest/workload.go
Lou1415926: > I can't change IDs to lowercase as its accessed in [test](https://github.com/gautam-nutalapati/copilot-cli/blob/83c3e587ac68c18ec7fab4ba0e6428774aafcf60/internal/pkg/deploy/cloudformation/stack/backend_svc_test.go#L386) of stack module.

Yeah not only that, but also that the `mergo` package used for environment overrides only works for exported field. If it was private it'd just skip over it without merging 😓.
 
I think we can keep the `GetIDs` naming then!!  It could also be argued that the naming conventions was for `Getter` that gets private fields, which is not particularly what we are doing here. Thank you for digging into it!

Lou1415926 (approved):
yayy again thank you so much for the contribution!

PR #3684, opened 2022-06-22
chore(cli): build env deploy into the cli

Lou1415926:


internal/pkg/cli/deploy/env.go
```diff
@@ -51,6 +54,37 @@ type envDeployer struct {
 	uploadCustomResourceFlag bool
 }
 
+type NewEnvDeployerInput struct {
+	App             *config.Application
+	Env             *config.Environment
+	SessionProvider *sessions.Provider
+}
+
+func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
```
paragbhingre, line 68: Can we add a comment to this public function? Also, where are we calling this function from?

internal/pkg/cli/env_deploy.go
```diff
@@ -42,6 +50,50 @@ type deployEnvOpts struct {
...
+		return nil, err
+	}
+	store := config.NewSSMStore(identity.New(defaultSess), ssm.New(defaultSess), aws.StringValue(defaultSess.Config.Region))
+	ws, err := workspace.New()
+	if err != nil {
+		return nil, fmt.Errorf("new workspace: %w", err)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("initiate a new environment deployer: %w", err)
+	}
```
paragbhingre: How does second if will execute if there is an error while creating a new workspace?
And I think second if looks incorrect as it says _initiate a new environment deployer_ which doesn't seem like happening around this code and looks like happening down on line 79

internal/pkg/cli/deploy/env.go
Lou1415926: https://github.com/aws/copilot-cli/blob/0b94c7f4a57b4dd5724331352626c7908553fb0b/internal/pkg/cli/env_deploy.go#L88 Here 😄

internal/pkg/cli/env_deploy.go
Lou1415926: I forgot to remove the second error handling 😄

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -78,6 +78,7 @@ func (e *EnvStackConfig) Template() (string, error) {
 		return "", err
 	}
 
+	// TODO: remove all these after we are able to migrate to the new upload workflow.
```
dannyrandall:
```suggestion
	// TODO(Lou1415926): remove all these after we are able to migrate to the new upload workflow.
```
😜

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -234,12 +235,11 @@ func (e *EnvStackConfig) internalALBSubnets() []string {
...
 func (e *EnvStackConfig) Parameters() ([]*cloudformation.Parameter, error) {
 	httpsListener := "false"
-	if len(e.in.ImportCertARNs) != 0 || e.in.App.Domain != "" {
+	if len(e.importPublicCertARNs()) != 0 || e.in.App.Domain != "" {
 		httpsListener = "true"
 	}
 	internalHTTPSListener := "false"
-	if len(e.in.ImportCertARNs) != 0 && e.in.ImportVPCConfig != nil &&
-		len(e.in.ImportVPCConfig.PublicSubnetIDs) == 0 {
+	if len(e.importPrivateCertARNs()) != 0 {
```
dannyrandall, line 242: yay! 🎉

internal/pkg/deploy/cloudformation/stack/env_test.go
```diff
@@ -264,7 +264,7 @@ func TestEnv_Parameters(t *testing.T) {
 				},
 				{
 					ParameterKey:   aws.String(envParamCreateHTTPSListenerKey),
-					ParameterValue: aws.String("true"),
+					ParameterValue: aws.String("false"),
```
dannyrandall, line 267: how was this passing before🤔

dannyrandall (approved):

internal/pkg/cli/deploy/deploy.go
```diff
@@ -121,12 +121,6 @@ type publicCIDRBlocksGetter interface {
 	PublicCIDRBlocks() ([]string, error)
 }
 
-type customResourcesUploader interface {
```
iamhopaul123, line 124: Is this file only related to workload deployment? If so maybe we can think about renaming the file

internal/pkg/deploy/cloudformation/stack/env_test.go
Lou1415926: it was related to the change here: https://github.com/aws/copilot-cli/blob/e86f180596c1210f1ae2bb79d44c0e8a4844278e/internal/pkg/deploy/cloudformation/stack/env.go#L238

which was previously:
```go
if len(e.in.ImportCertARNs) != 0 || e.in.App.Domain != ""
```

So when you have imported certs, and 0 public subnet (i.e. imported private cert), this parameter still renders a value of `true` - which, I think, but not sure of, was supposed to be false.

iamhopaul123 (approved):

internal/pkg/deploy/cloudformation/stack/env_test.go
dannyrandall: yeah, i agree it should be false - i'm just confused why it wasn't false _before_ also since [`mockDeployEnvironmentInput`](https://github.com/Lou1415926/copilot-cli/blob/e86f180596c1210f1ae2bb79d44c0e8a4844278e/internal/pkg/deploy/cloudformation/stack/env_test.go#L386) doesn't have any imported certs... anyways. Not a big deal. I'm just glad it's right😊

internal/pkg/cli/deploy/deploy.go
Lou1415926: addressed!

internal/pkg/cli/env_deploy.go
Lou1415926: removed and addressed!

PR #3686, opened 2022-06-23
chore: priority lambda should handle root rule

iamhopaul123:


cf-custom-resources/lib/alb-rule-priority-generator.js
```diff
@@ -77,10 +79,11 @@ let report = function (
  * priorities, and then returns max + 1.
  *
  * @param {string} listenerArn the ARN of the ALB listener.
+ * @param {bool} isRootPath if it is the root listener rule path.
 
  * @returns {number} The next available ALB listener rule priority.
  */
-const calculateNextRulePriority = async function (listenerArn) {
+const calculateNextRulePriority = async function (listenerArn, isRootPath) {
```
efekarakus: Can we split this function into two?
```
calculateNextDefaultRulePathPriority
calculateNextRulePathPriority
```

cf-custom-resources/lib/alb-rule-priority-generator.js
```diff
@@ -97,24 +100,42 @@ const calculateNextRulePriority = async function (listenerArn) {
...
-    nextRulePriority = max + 1;
+  var nextRulePriority;
+  if (isRootPath) {
+    nextRulePriority = maxPriorityForRootRule;
+    if (rules.length > 0) {
+      // Take the min rule priority, and decrement it by 1.
+      const rulePriorities = rules.map((rule) => {
+        if (
+          rule.Priority === "default" ||
+          rule.Priority < minPriorityForRootRule
```
efekarakus: Instead of having a range of `[minPriorityForRootRule, maxPriorityForRootRule]` can we instead 
do something like
```js
const minDefaultRulePriority = rules.filter(isDefaultRule).reduce(minPriority);
```
the `isDefaultRule` function will check if the rulepath is `"/*"` and the `minPriority` is a reducer to grab only the smallest priority among them.

This way we don't have to have a range.

cf-custom-resources/lib/alb-rule-priority-generator.js
```diff
@@ -97,24 +100,42 @@ const calculateNextRulePriority = async function (listenerArn) {
...
+        ) {
+          // Ignore the root rule's priority.
+          // Ignore the non root rule's priority.
+          return maxPriorityForRootRule + 1;
+        }
+        return parseInt(rule.Priority);
+      });
+      nextRulePriority = Math.min(...rulePriorities) - 1;
```
efekarakus: Is it possible that we end up with `Math.min = 0` if there are no rules with a "/"? For example, if the user has set up only manifests with `/api` or `/frontend`

cf-custom-resources/lib/alb-rule-priority-generator.js
```diff
@@ -97,24 +100,42 @@ const calculateNextRulePriority = async function (listenerArn) {
...
-      }
-      return parseInt(rule.Priority);
-    });
-    const max = Math.max(...rulePriorities);
-    nextRulePriority = max + 1;
+  var nextRulePriority;
+  if (isRootPath) {
+    nextRulePriority = maxPriorityForRootRule;
+    if (rules.length > 0) {
+      // Take the min rule priority, and decrement it by 1.
```
efekarakus: Should we add more explanation here on why we're doing this for root paths?
Like a sample scenario why we want the root paths to have the lowest priorities.

cf-custom-resources/lib/alb-rule-priority-generator.js
iamhopaul123: If there's no default rule then we won't get to line 119 right 🤔

cf-custom-resources/lib/alb-rule-priority-generator.js
```diff
@@ -97,24 +99,68 @@ const calculateNextRulePriority = async function (listenerArn) {
...
+    const rulesResponse = await elb
+      .describeRules({
+        ListenerArn: listenerArn,
+        Marker: marker,
+      })
+      .promise();
+
+    rules = rules.concat(rulesResponse.Rules);
+    marker = rulesResponse.NextMarker;
+  } while (marker);
```
dannyrandall: maybe break this out into it's own func?

cf-custom-resources/lib/alb-rule-priority-generator.js
```diff
@@ -97,24 +99,68 @@ const calculateNextRulePriority = async function (listenerArn) {
     marker = rulesResponse.NextMarker;
   } while (marker);
 
-  let nextRulePriority = 1;
+  var nextRulePriority = 1;
```
dannyrandall: why `var`...?

iamhopaul123: <strike>to make it consistent? also i find `let` pretty confusing...</strike>

Updated actually `var` is more confusing.

cf-custom-resources/lib/alb-rule-priority-generator.js
```diff
@@ -4,8 +4,10 @@
 
 const aws = require("aws-sdk");
 
-// priorityForRootRule is the max priority number that's always set for the listener rule that matches the root path "/"
-const priorityForRootRule = "50000";
+// minPriorityForRootRule is the min priority number for the the root path "/".
+const minPriorityForRootRule = 48000;
+// maxPriorityForRootRule is the max priority number for the the root path "/".
+const maxPriorityForRootRule = 50000;
```
dannyrandall, line 10: i wonder if it would be nice to leave the lowest priority open and set our max to 49999? kinda like how we skip the highest priority (0)?

cf-custom-resources/lib/alb-rule-priority-generator.js
efekarakus: yup you're right sorry!

efekarakus (approved):

cf-custom-resources/lib/alb-rule-priority-generator.js
dannyrandall: nvm, might be a breaking change

dannyrandall: nvm, _might_ be a breaking change

dannyrandall (approved):

PR #3689, opened 2022-06-24
docs: minor error on SNS subscription policy v 1.16 blog

aws-rafams:
Filter policy acts at the topic level not queue level https://aws.github.io/copilot-cli/docs/manifest/worker-service/#topic-filter-policy

### DIFF
--- a/site/content/blogs/release-v116.en.md
+++ b/site/content/blogs/release-v116.en.md
@@ -158,7 +158,7 @@ subscribe:
         store:
             - example_corp
         event:
-            - anything-but: order_canceled
+            - order_canceled
         price_usd:
             - numeric:
               - ">="
--- a/site/content/blogs/release-v116.en.md
+++ b/site/content/blogs/release-v116.en.md
@@ -154,10 +154,7 @@ subscribe:
   topics:
     - name: ordersTopic
       service: api
-  queue:
-    dead_letter:
-      tries: 5
-    filter_policy:
+      filter_policy:
         store:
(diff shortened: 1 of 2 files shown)


dannyrandall (commented):
Great catch!:tada: Now that I'm looking at it, is this statement also wrong?
L145
> However, let’s say you need to create a new worker that only processes messages of canceled orders with price value of more than $100.

I wonder if to make that true, we should change the `filter_policy.event` to be
```
filter_policy:
  event:
    - order_canceled
```
instead. What do you think?

aws-rafams:
> Great catch!🎉 Now that I'm looking at it, is this statement also wrong? L145
> 
> > However, let’s say you need to create a new worker that only processes messages of canceled orders with price value of more than $100.
> 
> I wonder if to make that true, we should change the `filter_policy.event` to be
> 
> ```
> filter_policy:
>   event:
>     - order_canceled
> ```
> 
> instead. What do you think?

You are indeed correct, with the current statement, it would process messages of orders that are not canceled. ([Docs for reference](https://docs.aws.amazon.com/sns/latest/dg/string-value-matching.html#string-anything-but-matching)). To be coherent with the scenario the anything-but-matching should be removed and the policy should look like the one you proposed. Nice catch as well!

Lou1415926 (approved):
Sharp eyes! Thank you @aws-rafams and @dannyrandall!

efekarakus (approved):

PR #3692, opened 2022-06-24
feat: add job run command

kangere:
This PR adds a `job run` command which can be used to execute scheduled jobs

Addresses #3469 


kangere:
I had to make some manual changes to resolve the merge conflicts (I haven't worked on this for a while)

internal/pkg/aws/stepfunctions/stepfunctions_test.go
```diff
@@ -68,3 +69,53 @@ func TestStepFunctions_StateMachineDefinition(t *testing.T) {
 		})
 	}
 }
+
+func TestStepFunctions_Execute(t *testing.T) {
```
Lou1415926, line 73: Thanks for the tests!!

internal/pkg/aws/stepfunctions/stepfunctions.go
```diff
@@ -39,3 +40,15 @@ func (s *StepFunctions) StateMachineDefinition(stateMachineARN string) (string,
...
 	return aws.StringValue(out.Definition), nil
 }
+
+//Execute starts a state machine execution
+func (s *StepFunctions) Execute(stateMachineARN string) error {
+	_, err := s.client.StartExecution(&sfn.StartExecutionInput{
+		StateMachineArn: aws.String(stateMachineARN),
+	})
+	if err != nil {
+		return fmt.Errorf("execute state machine: %w", err)
```
Lou1415926: Do you think this may make the error message more helpful:
```suggestion
		return fmt.Errorf("execute state machine %s: %w", stateMachineARN, err)
```

internal/pkg/cli/interfaces.go
```diff
@@ -644,6 +644,11 @@ type workloadTemplateGenerator interface {
 		*clideploy.GenerateCloudFormationTemplateOutput, error)
 }
 
+
+type Runner interface {
```
Lou1415926: Seems like we can keep this interface private:
```suggestion
type runner interface {
```

internal/pkg/runner/job/job_runner.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/copilot-cli/internal/pkg/aws/cloudformation"
+	"github.com/aws/copilot-cli/internal/pkg/aws/stepfunctions"
+)
+
+type JobExecutor interface {
```
Lou1415926: Same here and the other interfaces in this package! We can keep this unexported if it's not referenced in other packages:
```suggestion
type jobExecutor interface {
```

internal/pkg/runner/job/job_runner.go
```diff
@@ -0,0 +1,86 @@
...
+		App:            opts.App,
+		Env:            opts.Env,
+		Job:            opts.Job,
+	}
+
+}
+
+func (r *JobRunner) Run() error {
+
+	resources, err := r.StackRetriever.StackResources(fmt.Sprintf("%s-%s-%s", r.App, r.Env, r.Job))
```
Lou1415926: For stack name, perhaps we can use this function https://github.com/aws/copilot-cli/blob/efe855432195e92927263dc2643a2d6e35ed8c9c/internal/pkg/deploy/cloudformation/stack/name.go#L28  so that it reuses the code and deal with potentially truncated stack name.

internal/pkg/runner/job/job_runner.go
```diff
@@ -0,0 +1,86 @@
...
+
+	if err != nil {
+		return fmt.Errorf("describe stack error: %v", err)
```
Lou1415926: Similar to the other comment about including info error messages to make them more helpful:
```suggestion
		return fmt.Errorf("describe stack %s: %v", NameForService(r.App, r.Env, r.Job), err)
```

internal/pkg/runner/job/job_runner.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	var stateMachineARN string
+
+	for _, resource := range resources {
+
+		if *resource.ResourceType == "AWS::StepFunctions::StateMachine" {
```
Lou1415926: We can use `aws.StringValue(resource.ResourceType)` which deals with nil-pointers automatically:
```suggestion
		if aws.StringValue(resource.ResourceType) == "AWS::StepFunctions::StateMachine" {
```

Lou1415926: In addition, what do you think of defining a const
```
const CFNStackResourceTypeStateMachine = "AWS::StepFunctions::StateMachine"
```
and use it here?

Perhaps the constant can be defined around here: https://github.com/aws/copilot-cli/blob/efe855432195e92927263dc2643a2d6e35ed8c9c/internal/pkg/deploy/cloudformation/cloudformation.go#L39

internal/pkg/runner/job/job_runner.go
```diff
@@ -0,0 +1,86 @@
...
+			stateMachineARN = *resource.PhysicalResourceId
```
Lou1415926: Samesies! For peace of mind we can do:

```suggestion
			stateMachineARN = aws.StringValue(resource.PhysicalReourceId)
```

Lou1415926 (commented):
Yayy this is amazing - thanks a ton for writing up the tests and of course the contribution!! I have only some minor comments.

internal/pkg/runner/job/job_runner.go
kangere: This function takes in a service name as one of its parameters, would it still work if you pass in the job name?

Lou1415926: Yes! The function name is a bit outdated as it really should be `NameForWorkload` 😅 

Service and job's names are both determined using this function: https://github.com/aws/copilot-cli/blob/966c97f0f0824dace2ab0ebbd76441aa664b6e20/internal/pkg/deploy/cloudformation/stack/workload.go#L131

kangere: okay, my latest commit should address this

internal/pkg/aws/stepfunctions/stepfunctions.go
```diff
@@ -39,3 +40,15 @@ func (s *StepFunctions) StateMachineDefinition(stateMachineARN string) (string,
 
 	return aws.StringValue(out.Definition), nil
 }
+
+//Execute starts a state machine execution
+func (s *StepFunctions) Execute(stateMachineARN string) error {
```
efekarakus: nit:
```suggestion
//Execute starts a state machine execution.
func (s *StepFunctions) Execute(arn string) error {
```

internal/pkg/cli/interfaces.go
```diff
@@ -644,6 +644,10 @@ type workloadTemplateGenerator interface {
 		*clideploy.GenerateCloudFormationTemplateOutput, error)
 }
 
+type simpleRunner interface {
```
efekarakus: Is it possible to rename the existing `runner` interface to `execRunner`, and `simpleRunner` to `runner` please? 🥺 🙏

internal/pkg/cli/job_run.go
```diff
@@ -0,0 +1,192 @@
...
+	opts := &jobRunOpts{
+		jobRunVars:  vars,
+		configStore: configStore,
+		sel:         selector.NewDeploySelect(prompt.New(), configStore, deployStore),
+		ws:          selector.NewLocalWorkloadSelector(prompter, configStore, ws),
+	}
+
+	opts.initRunner = func() {
+		opts.runner = jobRunner.NewJobRunner(&jobRunner.JobRunnerConfig{
+			Sess: defaultSess,
```
efekarakus: 1. For running the job, we should use the credentials from the `EnvManagerRole` instead of the default session. This is because the `default` session might point to an AWS account different than the environment's account where the step function state machine is located.

Can we do something similar to the following to fetch the session from the env manager role:
https://github.com/aws/copilot-cli/blob/a90c9a56909c0addf3e22d1aa9c3c48358799d45/internal/pkg/cli/svc_exec.go#L186-L192

efekarakus: 2. Instead of passing in the `Sess` to the `jobRunner` is it possible to instead use dependency injection?

```go
opts.initRunner = func() error {
   sess, err := o.envSession()
   if err != nil {
      return err
   }
   cfn := cloudformation.New(sess)
   sfn := stepfunctions.New(sess)
   opts.runner =  jobrunner.NewJobRunner({
       Executor: sfn,
       StackRetriever: cfn,
       ...
   })
   return nil
}
```

internal/pkg/cli/job_run.go
```diff
@@ -0,0 +1,192 @@
...
+
+func (o *jobRunOpts) askJobName() error {
+	if o.jobName != "" {
+		if _, err := o.configStore.GetJob(o.appName, o.jobName); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	name, err := o.ws.Job("Select a job from your workspace", "")
```
efekarakus: Can we select the `Job` and `Environment` from a `ConfigSelector` instead of the workspace?
We should be able to run `job run` even if the job manifest file isn't available locally.

```
sel: selector.NewConfigSelector(prompt.New(), configStore),
```

internal/pkg/cli/job_run.go
```diff
@@ -0,0 +1,192 @@
...
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ssm"
+	"github.com/aws/copilot-cli/internal/pkg/aws/identity"
+	"github.com/aws/copilot-cli/internal/pkg/aws/sessions"
+	"github.com/aws/copilot-cli/internal/pkg/config"
+	"github.com/aws/copilot-cli/internal/pkg/deploy"
+	jobRunner "github.com/aws/copilot-cli/internal/pkg/runner/job"
```
efekarakus: nit: can we rename the package to be `runner/jobrunner` instead?
```suggestion
	"github.com/aws/copilot-cli/internal/pkg/runner/jobrunner"
```

internal/pkg/runner/job/job_runner.go
```diff
@@ -0,0 +1,88 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//Package job provides support for invoking scheduled jobs
```
efekarakus: nit: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences
> Comments should begin with the name of the thing being described and end in a period:
```suggestion
// Package jobrunner provides support for invoking scheduled jobs.
```

internal/pkg/runner/job/job_runner.go
```diff
@@ -0,0 +1,88 @@
...
+
+type jobExecutor interface {
+	Execute(stateMachineARN string) error
+}
+
+type StackRetriever interface {
+	StackResources(name string) ([]*cloudformation.StackResource, error)
+}
+
+type JobRunner struct {
```
efekarakus: Can we add doc comments to all exported names please 🥺 
https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

internal/pkg/runner/job/job_runner.go
```diff
@@ -0,0 +1,88 @@
...
+	Executor jobExecutor
+
+	//App Env and Job name to retrieve cloudformation stack
+	App string
+	Env string
+	Job string
+
+	StackRetriever StackRetriever
```
efekarakus: all of the fields in the `JobRunner` can be private I think :)

internal/pkg/runner/job/job_runner.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+type JobRunnerConfig struct {
+	App  string
+	Env  string
+	Job  string
+	Sess *session.Session
+}
+
+func NewJobRunner(opts *JobRunnerConfig) *JobRunner {
```
efekarakus: nit:
```suggestion
func New(cfg *Config) *JobRunner {
```

efekarakus (commented):
Thank you so much @kangere ! this is amazing work 😍 

Please let me know if you need any help with the feedback, we'd be happy to jump in and add additional commits to the PR

internal/pkg/cli/interfaces.go
kangere: sure, that makes sense

internal/pkg/cli/job_run.go
kangere: I am not opposed to that idea as I am not as familiar with how sessions work but in the svc_exec.go file you linked the default session is used for initialization purposes, should this have been done differently here ?
https://github.com/aws/copilot-cli/blob/a90c9a56909c0addf3e22d1aa9c3c48358799d45/internal/pkg/cli/svc_exec.go#L63-L70

internal/pkg/runner/job/job_runner.go
kangere: I added comments, but not too sure of the wording, feel free to edit them

internal/pkg/cli/job_run.go
kangere: This should be addressed in my latest commit

internal/pkg/cli/job_run.go
efekarakus: Hi @kangere !

I've addressed all the feedback, there is one more small action that needs to do before we merge the change.
I've added a TODO in `job_run.go` which I'll handle tomorrow morning.

Thank you so much for the contribution! There shouldn't be any more action items from your end :)

kangere: Hi @efekarakus 
Thanks for all the feedback and help. I appreciate it.

I hope I was able to address the comments and feedback well enough. 

I also hope to make many more future contributions!

efekarakus: > I also hope to make many more future contributions!

We would love to, thank you @kangere !

The PR is at a reviewable state, I'll wait for feedback from my teammates and then we can merge it

internal/pkg/cli/job_deploy.go
```diff
@@ -35,12 +35,12 @@ type deployJobOpts struct {
 	ws                   wsWlDirReader
 	unmarshal            func(in []byte) (manifest.WorkloadManifest, error)
 	newInterpolator      func(app, env string) interpolator
-	cmd                  runner
+	cmd                  execRunner
 	sessProvider         *sessions.Provider
+	envUpgradeCmd        actionCommand
```
Lou1415926:
```suggestion
```

because of #3698 !

Lou1415926 (approved):
LGTM thank you so much!

efekarakus (approved):

PR #3688, opened 2022-06-23
chore: retrieve imported certificates from manifest and ssm

Lou1415926:
This PR reads the deployed environment's manifest before attempting reading from SSM configuration for environment for imported certs.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the Apache 2.0 License.

PR #3698, opened 2022-06-28
chore: replace auto-upgrade of environments with compatibility checkers

Lou1415926:
Previously, Copilot automatically upgrades an environment during workload's `deploy` commands if the environment is not on the latest template version.

With this PR, it stops doing that, and instead will check if the workload's manifest is attempting to use a feature that is not yet supported by the environment's template; if yes, then Copilot errors out and asks user to upgrade their environment.

Similar logic applies to the "pipeline + `svc package`" workflow. 

Note: this PR can be merged only after #3688 .


internal/pkg/cli/svc_deploy.go
```diff
@@ -467,6 +469,21 @@ func (o *deploySvcOpts) getTargetApp() (*config.Application, error) {
...
+}
+
+func (e *errManifestIncompatibleWithEnvironment) Error() string {
+	return fmt.Sprintf("environment %q is not on a version that supports the %q feature", e.envName, template.FriendlyEnvFeatureName(e.missingFeature))
+}
+
+// RecommendActions returns recommended actions to be taken after the error.
+// Implements main.actionRecommender interface.
+func (e *errManifestIncompatibleWithEnvironment) RecommendActions() string {
+	return fmt.Sprintf("You can upgrade your environment by running %s.\n", color.HighlightCode(fmt.Sprintf("copilot env deploy --name %s", e.envName)))
```
```suggestion
	return fmt.Sprintf("You can upgrade your environment template by running %s.\n", color.HighlightCode(fmt.Sprintf("copilot env deploy --name %s", e.envName)))
```

internal/pkg/cli/svc_package.go
```diff
@@ -195,14 +194,14 @@ func (o *packageSvcOpts) Execute() error {
 	if err != nil {
 		return nil
 	}
-	appTemplates, err := o.getSvcTemplates(targetEnv)
```
efekarakus, line 198: :doge:

internal/pkg/cli/svc_package.go
```diff
@@ -396,9 +404,9 @@ func (o *packageSvcOpts) getTargetEnv() (*config.Environment, error) {
 	return o.targetEnv, nil
 }
 
-// RecommendActions is a no-op for this command.
+// RecommendActions suggests recommended actions before the packaged template is used for deployment.
 func (o *packageSvcOpts) RecommendActions() error {
-	return nil
+	return isManifestCompatibleWithEnvironment(o.appliedManifest.(manifest.WorkloadManifest), o.envName, o.envFeaturesDescriber)
```
efekarakus: nit: What do you think of renaming this fn to be like this:
```suggestion
	return validateManifestCompatiblityWithEnv(o.appliedManifest.(manifest.WorkloadManifest), o.envName, o.envFeaturesDescriber)
```
Just because `is*` makes me think that the return type is a `bool`

efekarakus: Also, should we actually error here? I think recommending them to run `env deploy` by logging to `stderr` is totally fair and printing the template to `stdout`. I just don't know if  **`svc`**`package` should fail if the template is incompatible w the env

```
err := isManifestCompatibleWithEnvironment(...)
if err != nil {
   log.Errorln(err)
}
return nil
```

internal/pkg/template/templates/cicd/buildspec.yml
```diff
@@ -20,13 +20,8 @@ phases:
     commands:
       - ls -l
       - export COLOR="false"
-      # First, upgrade the cloudformation stack of every environment in the pipeline.
       - pipeline=$(cat $CODEBUILD_SRC_DIR/{{.ManifestPath}} | ruby -ryaml -rjson -e 'puts JSON.pretty_generate(YAML.load(ARGF))')
       - pl_envs=$(echo $pipeline | jq -r '.stages[].name')
-      - >
```
efekarakus, line 26: 🥳 yayayya

internal/pkg/cli/svc_package.go
Lou1415926: I was mainly concerned with the pipeline flow: if we don't exit with a non-zero code, the build stage will go through, and the deployment stage will follow. The deployment will then fail because of the incompatibility.

Do you think we can use a flag such as `copilot svc package --check-compatibility` to do this? Just like how we handle `--upload-assets`.

efekarakus: Oh! That's smart, ok let's keep it as is.

It would be awesome if the feature just works without people having to update their buildspec.

iamhopaul123 (approved):

efekarakus (approved):

PR #3702, opened 2022-06-29
fix: Add quotes for env-vars and secrets for `task run --generate-cmd`

### LINKED ISSUE #3683: `task run --generate-cmd` gives invalid output if an env var has commas in it
Example of the error:

```bash
❯ copilot task run --env-vars a="1,2,3",b="b"
✘ invalid argument "a=1,2,3,b=b" for "--env-vars" flag: 2 must be formatted as key=value
```

As far as I've been able to determine, the error message comes from the [pflag library](https://pkg.go.dev/github.com/spf13/pflag@v1.0.5#section-readme) but this seems to actually be a bug in `--generate-cmd`... the quotes that `--generate-cmd` wraps the values in get stripped out by the shell, so their effect is to shell-quote but not to delimit env vars to copilot. You can see this more simply by running echo:

```bash
❯ echo --env-vars a="1,2,3",b="b"
--env-vars a=1,2,3,b=b
```

I tried escaping the quotes, thinking maybe copilot could start escaping them, but that fails too:

```bash
❯ copilot task run --env-vars a=\"1,2,3\",b=\"b\"       
✘ invalid argument "a=\"1,2,3\",b=\"b\"" for "--env-vars" flag: parse error on line 1, column 3: bare " in non-quoted-field
```

What _does_ work is passing each env var as its own arg, rather than one arg separated by commas:

```bash
❯ copilot task run --env-vars "a=1,2,3" --env-vars "b=b"
```

I'm not sure what the preferred fix would be here, though. Thoughts?

codekitchen:
fix quoting of env vars and secrets in `task run --generate-cmd` output. also quote them in the generated yaml cf template.

Fixes #3683

A couple comments on this:

1. I had to go further than the solution we originally talked about @efekarakus -- once I started testing the fix against our actual app, I realized it didn't fully fix our use case which is having JSON env var values. We need to do further shell escaping/quoting to get JSON passed back in to copilot correctly. I decided maybe wrapping the entire option in single quotes is best?
2. This shell escaping is necessarily shell specific. I tested this in sh, bash, zsh, fish, it works fine there, but I don't know of any reasonable way to guarantee escaping across all common shells. I know copilot supports windows as well, I don't know much about windows shells, is this likely to work there? If not I'm not sure what the fix might be.

### DIFF
--- a/internal/pkg/ecs/run_task_request.go
+++ b/internal/pkg/ecs/run_task_request.go
@@ -299,6 +299,9 @@ func fmtStringMapToString(m map[string]string) (string, error) {
 
 	// Then for shell escaping, wrap the entire argument in single quotes
 	// and escape any internal single quotes.
-	final = strings.ReplaceAll(final, `'`, `'\''`)
-	return `'` + final + `'`, nil
+	return shellQuote(final), nil
+}
+
+func shellQuote(s string) string {
+	return `'` + strings.ReplaceAll(s, `'`, `'\''`) + `'`
 }
--- a/internal/pkg/ecs/run_task_request.go
+++ b/internal/pkg/ecs/run_task_request.go
@@ -168,7 +168,7 @@ func RunTaskRequestFromJob(client JobDescriber, app, env, job string) (*RunTaskR
 }
 
 // String stringifies a RunTaskRequest.
-func (r RunTaskRequest) CLIString() string {
+func (r RunTaskRequest) CLIString() (string, error) {
 	output := []string{"copilot task run"}
 	if r.executionRole != "" {
 		output = append(output, fmt.Sprintf("--execution-role %s", r.executionRole))
@@ -191,11 +191,19 @@ func (r RunTaskRequest) CLIString() string {
 	}
 
 	if r.envVars != nil && len(r.envVars) != 0 {
-		output = append(output, fmt.Sprintf("--env-vars %s", fmtStringMapToString(r.envVars)))
+		vars, err := fmtStringMapToString(r.envVars)
+		if err != nil {
+			return "", err
+		}
+		output = append(output, fmt.Sprintf("--env-vars %s", vars))
 	}
 
 	if r.secrets != nil && len(r.secrets) != 0 {
-		output = append(output, fmt.Sprintf("--secrets %s", fmtStringMapToString(r.secrets)))
+		secrets, err := fmtStringMapToString(r.secrets)
+		if err != nil {
+			return "", err
+		}
+		output = append(output, fmt.Sprintf("--secrets %s", secrets))
 	}
 
 	if r.networkConfiguration.Subnets != nil && len(r.networkConfiguration.Subnets) != 0 {
@@ -218,7 +226,7 @@ func (r RunTaskRequest) CLIString() string {
 		output = append(output, fmt.Sprintf("--cluster %s", r.cluster))
 	}
 
-	return strings.Join(output, " \\\n")
+	return strings.Join(output, " \\\n"), nil
 }
 
 func containerInformation(taskDef *awsecs.TaskDefinition, containerName string) (*containerInfo, error) {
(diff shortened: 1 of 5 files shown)


internal/pkg/ecs/run_task_request.go
```diff
@@ -271,7 +274,11 @@ func fmtStringMapToString(m map[string]string) string {
...
 	for _, k := range keys {
-		output = append(output, fmt.Sprintf(`%s="%v"`, k, m[k]))
+		// params are parsed as csv, so we escape double quotes by doubling them up
+		v := strings.ReplaceAll(m[k], `"`, `""`)
+		// need to allow single quotes through the shell escaping
+		v = strings.ReplaceAll(v, `'`, `'\''`)
+		output = append(output, fmt.Sprintf(`"%s=%s"`, k, v))
 	}
-	return strings.Join(output, ",")
+	return `'` + strings.Join(output, ",") + `'`
```
efekarakus: I wonder if we can leverage `csv.NewWriter` instead? 
Here is a sample snippet: https://go.dev/play/p/aH4TUM7o4jo

I don't know if this causes any issues with the shell interpreting quotes.

efekarakus (commented):
Thank you so much @codekitchen ! ❤️

internal/pkg/ecs/run_task_request.go
codekitchen: that's a great idea thanks, I'll push up a change to use `csv` instead. We want to apply the shell escaping _after_ the csv formatting, so it should work here.

codekitchen: ok I've pushed up that change @efekarakus

it's nice that the csv lib only wraps in double quotes when necessary, it cleans up the --generate-cmd output a bit

internal/pkg/ecs/run_task_request.go
```diff
@@ -260,18 +269,36 @@ func containerInformation(taskDef *awsecs.TaskDefinition, containerName string)
...
+		return "", err
 	}
-	return strings.Join(output, ",")
+	w.Flush()
+	final := strings.TrimSuffix(buf.String(), "\n")
+
+	// Then for shell escaping, wrap the entire argument in single quotes
+	// and escape any internal single quotes.
+	final = strings.ReplaceAll(final, `'`, `'\''`)
+	return `'` + final + `'`, nil
```
efekarakus: nit: 
can we move this to a separate function

```go
func shellQuote(s string) string { 
    return `'` + strings.ReplaceAll(s, `'`, `'\''`) + `'`
}
```

efekarakus (approved):
Looks great! Thank you very much, just a small nit. Once addressed, we can remove the label and it will automerge

internal/pkg/template/templates/task/cf.yml
```diff
@@ -57,14 +57,14 @@ Resources:
           Name: !Ref TaskName{{if .EnvVars}}
           Environment:{{range $name, $value := .EnvVars}}
           - Name: {{$name}}
-            Value: {{$value}}{{end}}{{end}}
+            Value: {{$value | printf "%q"}}{{end}}{{end}}
```
Lou1415926, line 60: Awesome this is much needed! For example without it, `{""key1"":""val1"",""key2"":5}` would get interpreted as a JSON object instead of a JSON string.

Lou1415926 (approved):
Thank you very much for fixing this bug 🙇🏼‍♀️ !!

internal/pkg/template/templates/task/cf.yml
codekitchen: yeah I ran into that during testing of the other portion of this change! haha

internal/pkg/ecs/run_task_request.go
codekitchen: sure, done

PR #3705, opened 2022-06-30
test(e2e): add 'isolated' tests for internal ALB

huanjani:


e2e/isolated/isolated_test.go
```diff
@@ -0,0 +1,247 @@
...
+		})
+		It("app show includes app name", func() {
+			appShowOutput, err := cli.AppShow(appName)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(appShowOutput.Name).To(Equal(appName))
+			Expect(appShowOutput.URI).To(BeEmpty())
+		})
+	})
+
+	Context("when deploying resources to be imported", func() {
```
iamhopaul123, line 44: It seems like we delete the stack after app delete. Shall we create the vpc stack before doing app init? Because otherwise the stack we want to delete might not exist (we could ignore the error but not ideal 🤔 )

Here we just need to double check if the VPC stack is there before moving on

e2e/isolated/isolated_test.go
```diff
@@ -0,0 +1,247 @@
...
+				cmd.Stdout = os.Stdout
+				cmd.Stderr = os.Stderr
+				Expect(cmd.Stderr).NotTo(BeNil())
+			})
+		})
+
+		Context("when `curl`ing the LB DNS from within the container", func() {
+			It("session manager should be installed", func() {
+				// Use custom SSM plugin as the public version is not compatible to Alpine Linux.
+				err := client.BashExec("chmod +x ./session-manager-plugin")
```
iamhopaul123: How does it work if we don't have `session-manager-plugin` in this suite?

iamhopaul123: Also instead of using exec, can we just create another frontend service and then ping the ALB endpoint in the code? So when we ping the frontend service, it pings the internal ALB to see if it gets ok. That way we also test if internal ALB can work well with public ALB (also we don't need to include session manager plugin in this suite).

e2e/isolated/backend/Dockerfile
```diff
@@ -0,0 +1,34 @@
+# We specify the base image we need for our
+# go application
+FROM golang:1.15 AS builder
```
dannyrandall, line 3: ancient lol

e2e/isolated/isolated_suite_test.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+var _ = BeforeSuite(func() {
+	vpcStackName = fmt.Sprintf("e2e-isolated-vpc-stack-%d", time.Now().Unix())
+	vpcStackTemplatePath = "file://vpc.yml"
+	ecsCli, err := client.NewCLI()
+	Expect(err).NotTo(HaveOccurred())
+	cli = ecsCli
+	aws = client.NewAWS()
+	appName = fmt.Sprintf("e2e-isolated-%d", time.Now().Unix())
```
dannyrandall: maybe share a `time.Now()` with `vpcStackName`? so they could be somewhat correlated if we have to manually delete resources?

e2e/isolated/isolated_suite_test.go
```diff
@@ -0,0 +1,67 @@
...
+	vpcConfig = client.EnvInitRequestVPCConfig{
+		CIDR:               "10.1.0.0/16",
+		PrivateSubnetCIDRs: "10.1.2.0/24,10.1.3.0/24",
+		PublicSubnetCIDRs:  "10.1.0.0/24,10.1.1.0/24",
```
dannyrandall: isn't it only supposed to have private subnets...?

e2e/isolated/isolated_test.go
```diff
@@ -0,0 +1,247 @@
...
+				Expect(envShowErr).NotTo(HaveOccurred())
+			})
+			It("should now have an internal ALB", func() {
+				Expect(envShowOutput.Resources).To(ContainElement(HaveKeyWithValue("type", "AWS::ElasticLoadBalancingV2::LoadBalancer")))
+			})
+		})
+
+		Context("when `curl`ing the LB DNS", func() {
+			It("is not reachable", func() {
+				cmd := osExec.Command("curl", fmt.Sprintf("http://%s.%s.%s.internal", svcName, envName, appName))
```
dannyrandall: why curl instead of `http.Get()`?

e2e/isolated/backend/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/aws/copilot-cli/e2e/isolated/back-end
+
+go 1.15
+
+require github.com/julienschmidt/httprouter v1.3.0 // indirect
```
efekarakus: huh not important but i wonder why it thinks this is //indirect

e2e/isolated/backend/main.go
```diff
@@ -0,0 +1,28 @@
...
+package main
+
+import (
+	"log"
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+)
+
+var message = "hello world"
```
efekarakus: nit
```suggestion
```

e2e/isolated/vpc.yml
```diff
@@ -0,0 +1,165 @@
...
+  VPC:
+    Type: AWS::EC2::VPC
+    Properties:
+      CidrBlock: 10.0.0.0/16
+      EnableDnsHostnames: true
+      EnableDnsSupport: true
+      InstanceTenancy: default
+  InternetGateway:
+    Type: AWS::EC2::InternetGateway
+  InternetGatewayAttachment:
```
efekarakus: Can we remove the `InternetGateway` and `IGWAttachment` resources? this way there is not even a possibility of a route that leads to the public internet

e2e/isolated/vpc.yml
```diff
@@ -0,0 +1,165 @@
...
+      VpcEndpointType: Interface
+      PrivateDnsEnabled: true
+      SubnetIds:
+        - !Ref PrivateSubnetOne
+        - !Ref PrivateSubnetTwo
+      ServiceName: !Sub com.amazonaws.${AWS::Region}.ec2messages
+      VpcId: !Ref VPC
+      SecurityGroupIds:
+        - !Ref VPCEndpointSecurityGroup
+  EC2VPCEndpoint:
```
efekarakus: ooh what is the ec2 and ec2messages VPC endpoint used by do you know?

e2e/isolated/isolated_suite_test.go
efekarakus: it looks like `vpcConfig` is actually not used in the test, can we delete the variable?

e2e/isolated/isolated_test.go
```diff
@@ -0,0 +1,247 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package isolated_test
+
+import (
+	"errors"
+	"fmt"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	osExec "os/exec"
```
efekarakus: nit: https://rakyll.org/style-packages/#renames-should-follow-the-same-rules
```suggestion
	osexec "os/exec"
```

e2e/isolated/isolated_test.go
```diff
@@ -0,0 +1,247 @@
...
+		})
+
+		It("should not return an error", func() {
+			Expect(initErr).NotTo(HaveOccurred())
+		})
+		It("svc init should create a svc manifest", func() {
+			Expect("./copilot/backend/manifest.yml").Should(BeAnExistingFile())
+		})
+		It("should write 'http' and private placement to the manifest", func() {
+			f, err := os.OpenFile("./copilot/backend/manifest.yml", os.O_WRONLY|os.O_APPEND, 0644)
```
efekarakus: nit: this definitely works, no need to change - I just wanted to point out an alternative. If you want you can also add an existing `copilot/backend/manifest.yml` file so that the CLI doesn't create a new one. This way you don't need to write the string at the end of the file :)

e2e/isolated/isolated_test.go
```diff
@@ -0,0 +1,247 @@
...
+			Expect(svc.SvcName).To(Equal(svcName))
+			Expect(svc.AppName).To(Equal(appName))
+			Expect(len(svc.Configs)).To(Equal(1))
+			Expect(svc.Configs[0].Environment).To(Equal(envName))
+			Expect(svc.Configs[0].CPU).To(Equal("256"))
+			Expect(svc.Configs[0].Memory).To(Equal("512"))
+			Expect(svc.Configs[0].Port).To(Equal("80"))
+		})
+
+		Context("when running `svc logs`", func() {
```
efekarakus: nit: personally, I think it's okay to get rid of this test case - the main thing that we want to test is making sure that the service and ALB is strictly internal. The test suite already covers that 👍 , anything beyond that kinda increases the chance that the e2e test batch will be flaky

e2e/isolated/isolated_test.go
```diff
@@ -0,0 +1,247 @@
...
+					AppName: appName,
+					EnvName: envName,
+				})
+			})
+			It("should not return an error", func() {
+				Expect(envShowErr).NotTo(HaveOccurred())
+			})
+			It("should now have an internal ALB", func() {
+				Expect(envShowOutput.Resources).To(ContainElement(HaveKeyWithValue("type", "AWS::ElasticLoadBalancingV2::LoadBalancer")))
+			})
```
efekarakus: nit:
```suggestion
			It("show show internal ALB", func() {
				envShowOutput, envShowErr = cli.EnvShow(&client.EnvShowRequest{
					AppName: appName,
					EnvName: envName,
				})
				Expect(envShowErr).NotTo(HaveOccurred())
				Expect(envShowOutput.Resources).To(ContainElement(HaveKeyWithValue("type", "AWS::ElasticLoadBalancingV2::LoadBalancer")))
			})
```

e2e/isolated/isolated_test.go
efekarakus: +1 I'm also not sure if `curl` is installed in the container img that's running the test

e2e/isolated/isolated_test.go
efekarakus: I like that this test suite imports a VPC with no IGW so we get to have at least a suite where we can add logic to make sure copilot behaves nicely in isolated environments 😍 
Up to y'all, but I would prefer keeping it as is.

e2e/isolated/isolated_test.go
```diff
@@ -0,0 +1,247 @@
...
+				AppName:   appName,
+				Resources: true,
+			})
+		})
+
+		It("should not return an error", func() {
+			Expect(svcShowError).NotTo(HaveOccurred())
+		})
+
+		It("should return correct configuration", func() {
```
efekarakus, line 160: can we add an additional test case for `svc status` to ensure the task successfully spun up?
similar to here:
https://github.com/aws/copilot-cli/blob/1caa364fd59f1bab4e0bc6a09d1d271fff24f5ab/e2e/multi-svc-app/multi_svc_app_test.go#L254

e2e/isolated/isolated_test.go
huanjani: Oh, I like that more. Changing!

e2e/isolated/vpc.yml
huanjani: I thought they were necessary for Systems Manager, but it appears that for this use case, we don't need them. Deleted.

e2e/isolated/isolated_suite_test.go
```diff
@@ -0,0 +1,67 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var cli *client.CLI
+var aws *client.AWS
+var appName string
+var vpcStackName string
+var vpcStackTemplatePath string
+var vpcImport client.EnvInitRequestVPCImport
```
dannyrandall, line 21: move to other test file? i don't see it used here

dannyrandall (approved):
lgtm!

e2e/isolated/isolated_test.go
```diff
@@ -0,0 +1,217 @@
...
+			Expect(svc.SvcName).To(Equal(svcName))
+			Expect(svc.AppName).To(Equal(appName))
+			Expect(len(svc.Configs)).To(Equal(1))
+			Expect(svc.Configs[0].Environment).To(Equal(envName))
```
iamhopaul123, line 164: Maybe check if the URL is the same as https://github.com/aws/copilot-cli/pull/3705/files#diff-6a53edf684761ba19fb41f20ae1ae12766e6fed7c288a46f646a5f9331b11449R211 ?

e2e/isolated/isolated_suite_test.go
huanjani: This seems to be how the other e2e tests have their vars, so I just followed suit(e🤣).

e2e/isolated/isolated_suite_test.go
```diff
@@ -0,0 +1,67 @@
...
+*/
+func Test_Isolated(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Isolated Suite")
+}
+
+var _ = BeforeSuite(func() {
+	vpcStackName = fmt.Sprintf("e2e-isolated-vpc-stack-%d", timeNow)
+	vpcStackTemplatePath = "file://vpc.yml"
+	ecsCli, err := client.NewCLI()
```
efekarakus: janice! 😡 excuse me
```suggestion
	copilot, err := client.NewCLI()
```

e2e/isolated/isolated_suite_test.go
```diff
@@ -0,0 +1,67 @@
...
+	Expect(err).NotTo(HaveOccurred())
+	err = aws.WaitStackCreateComplete(vpcStackName)
+	Expect(err).NotTo(HaveOccurred())
+})
+
+var _ = AfterSuite(func() {
+	_, err := cli.AppDelete()
+	Expect(err).NotTo(HaveOccurred())
+	// Delete VPC stack.
+	err = aws.DeleteStack(vpcStackName)
```
efekarakus, line 55: do we need to wait for the stack deletion to complete?

efekarakus (approved):

e2e/isolated/isolated_suite_test.go
huanjani: Added.

PR #3706, opened 2022-06-30
chore(deploy): define configurations to deploy a bootstrap stack for an environment

Lou1415926:

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/env.go
+++ b/internal/pkg/deploy/cloudformation/stack/env.go
@@ -129,109 +129,6 @@ func (e *EnvStackConfig) Template() (string, error) {
 	return content.String(), nil
 }
 
-func (e *EnvStackConfig) vpcConfig() template.VPCConfig {
-	return template.VPCConfig{
-		Imported: e.importVPC(),
-		Managed:  e.managedVPC(),
-	}
-}
-
-func (e *EnvStackConfig) importVPC() *template.ImportVPC {
-	// If a manifest is present, it is the only place we look at.
-	if e.in.Mft != nil {
-		return e.in.Mft.Network.VPC.ImportedVPC()
-	}
-
-	// Fallthrough to SSM config.
-	if e.in.ImportVPCConfig == nil {
-		return nil
-	}
-	return &template.ImportVPC{
-		ID:               e.in.ImportVPCConfig.ID,
-		PublicSubnetIDs:  e.in.ImportVPCConfig.PublicSubnetIDs,
-		PrivateSubnetIDs: e.in.ImportVPCConfig.PrivateSubnetIDs,
-	}
-}
-
-func (e *EnvStackConfig) managedVPC() template.ManagedVPC {
-	defaultManagedVPC := template.ManagedVPC{
-		CIDR:               DefaultVPCCIDR,
-		PublicSubnetCIDRs:  DefaultPublicSubnetCIDRs,
-		PrivateSubnetCIDRs: DefaultPrivateSubnetCIDRs,
-	}
-	// If a manifest is present, it is the only place we look at.
-	if e.in.Mft != nil {
-		if v := e.in.Mft.Network.VPC.ManagedVPC(); v != nil {
-			return *v
-		}
-		return defaultManagedVPC
-	}
-
-	// Fallthrough to SSM config.
-	if e.in.AdjustVPCConfig == nil {
-		return defaultManagedVPC
-	}
-	return template.ManagedVPC{
-		CIDR:               e.in.AdjustVPCConfig.CIDR,
-		AZs:                e.in.AdjustVPCConfig.AZs,
-		PublicSubnetCIDRs:  e.in.AdjustVPCConfig.PublicSubnetCIDRs,
-		PrivateSubnetCIDRs: e.in.AdjustVPCConfig.PrivateSubnetCIDRs,
-	}
-}
-
-func (e *EnvStackConfig) telemetryConfig() *template.Telemetry {
-	// If a manifest is present, it is the only place we look at.
-	if e.in.Mft != nil {
-		return &template.Telemetry{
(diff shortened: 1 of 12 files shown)


internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -129,109 +129,6 @@ func (e *EnvStackConfig) Template() (string, error) {
 	return content.String(), nil
 }
 
-func (e *EnvStackConfig) vpcConfig() template.VPCConfig {
```
Lou1415926, line 132: These deletions of `EnvStackConfig` (`vpcConfig`, `importVPC`, ..., `internalALBSubnets`) are not real: I just moved further down in this file. Nothing is changed.

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -340,3 +295,106 @@ func (e *EnvStackConfig) ToEnv(stack *cloudformation.Stack) (*config.Environment
 		ExecutionRoleARN: stackOutputs[envOutputCFNExecutionRoleARN],
 	}, nil
 }
+
+func (e *EnvStackConfig) vpcConfig() template.VPCConfig {
```
Lou1415926, line 299: These are also fake news.

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -318,9 +215,67 @@ func (e *EnvStackConfig) StackName() string {
...
+
+// SerializedParameters returns the CloudFormation stack's parameters serialized
+// to a YAML document annotated with comments for readability to users.
+func (e *BootstrapEnvStackConfig) SerializedParameters() (string, error) {
+	// No-op for now.
+	return "", nil
+}
+
+// Tags returns the tags that should be applied to the bootstrap CloudFormation stack.
+func (e *BootstrapEnvStackConfig) Tags() []*cloudformation.Tag {
```
iamhopaul123, line 267: Do we need to explicitly define these functions?

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -318,9 +215,67 @@ func (e *EnvStackConfig) StackName() string {
...
+// StackName returns the name of the CloudFormation stack (based on the app and env names).
+func (e *BootstrapEnvStackConfig) StackName() string {
+	return (*EnvStackConfig)(e).StackName()
+}
+
 // ToEnv inspects an environment cloudformation stack and constructs an environment
-// struct out of it (including resources like ECR Repo)
-func (e *EnvStackConfig) ToEnv(stack *cloudformation.Stack) (*config.Environment, error) {
+// struct out of it (including resources like ECR Repo).
+func (e *BootstrapEnvStackConfig) ToEnv(stack *cloudformation.Stack) (*config.Environment, error) {
```
iamhopaul123, line 278: Why do we need to make this change 🤔

internal/pkg/template/templates/environment/partials/cfn-execution-role.yml
```diff
@@ -5,9 +5,6 @@ CloudformationExecutionRole:
     'aws:copilot:description': 'An IAM Role for AWS CloudFormation to manage resources'
   DeletionPolicy: Retain
   Type: AWS::IAM::Role
-{{- if not .VPCConfig.Imported}}
-  DependsOn: VPC
-{{- end}}
```
iamhopaul123, line 10: I can't remember why we need this dependsOn. Is there anything to worry about if we remove it 🤔

internal/pkg/deploy/cloudformation/stack/env.go
Lou1415926: Yes so that it implements `stackConfig`  interface.

internal/pkg/deploy/cloudformation/stack/env.go
Lou1415926: Because the only usage of `ToEnv` is to retrieve  information such as `EnvManagerRole` and `CFNExecutionRole` right after `env init` deploys the stack. I think `BootstrapEnvStackConfig` is a better receiver for the method than `EnvStackConfig`. 

Nothing is changed really whether `BootstrapEnvStackConfig` is the receiver or `EnvStackConfig `. Just thought the former makes more sense to me!

internal/pkg/template/templates/environment/partials/cfn-execution-role.yml
Lou1415926: I've discussed with several founders 🤣 and the decision was to remove it cuz there is no reason to keep it if no one knows why it's there.

iamhopaul123 (approved):

internal/pkg/template/templates/environment/bootstrap-cf.yml
```diff
@@ -0,0 +1,28 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+Description: CloudFormation environment template for infrastructure shared among Copilot workloads.
+Metadata:
+  Version: v1.10.1
```
efekarakus: Should we take the actual version as input here?

efekarakus (commented):
Looks great! Just a small question

internal/pkg/template/templates/environment/bootstrap-cf.yml
```diff
@@ -0,0 +1,28 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+Description: CloudFormation environment template for infrastructure shared among Copilot workloads.
```
efekarakus: nit: can we make the description like this as well:
```suggestion
Description: CloudFormation environment bootstrap template with the necessary roles to create envs and workloads.
```

efekarakus (approved):

Issue #3701, opened 2022-06-28
[Design] Copilot CloudFront Integration


CaptainCarpensir:
This issue is a design proposal to add CloudFront integration into Copilot.

CloudFront is a content delivery network which allows users to deploy their applications globally, and securely. Introducing Copilot support for this eases development, and gives users more use cases for Copilot.

### Problem Statement

Currently, Copilot has no internal integration with global content delivery.  In order to setup these features, users have to work with AWS APIs or Consoles outside of Copilot. To continue the cycle of build, release, and operate, we want to deliver CloudFront support with Copilot for users who wish to introduce CloudFront into their applications without requiring them to spend time on creating the CloudFront infrastructure manually.

### Proposal

Adding `cdn:` field to environment manifests being introduced in #3522.

Users will configure CloudFront through a field in the environment manifest. All cases and customization will be handled in this field. Copilot will handle all the infrastructure work of setting up a CloudFront deployment for the user. The user will then be able to configure how the CloudFront distribution interacts with the rest of their application via subfields of the `cdn`. 

Initially we plan to create a default CloudFront distribution by specifying `cdn`.
```
cdn: true
```

Then, for further customization, fields which cover `http` and https support, as well as `tls_termination` and http/https `redirect`.
```
cdn
  http:
    path: "api" # This is part of the subdomain of the CloudFront/Route 53 domain name
    redirect: true
```
```
cdn:
  tls_termination: true
```

Then we also plan to support additional `static` S3 bucket origins.
```
cdn:
  static:
    # Must be a bucket regional domain name
    path: "static" # Default is *
    location: cf-s3-ecs-demo-bucket.s3.us-west-2.amazonaws.com
```

### Milestones

- First, CloudFront will be generated as a distribution in front the Application Load Balancers generated by a deployed LBWS.
- Second, we will allow users to hook previously created S3 buckets onto the CloudFront distribution.
- Lastly, we will allow the user to specify using TLS termination at CloudFront.

These features cover the initial design of CloudFront integration with Copilot, which has been requested before in #1313 

---
We hope this feature will benefit you, and we're happy to receive any feedback you have!

afgallo:
Hi @CaptainCarpensir I'd love to see this feature added to Copilot. We currently achieve a similar outcome by creating the CF distribution via an add-on. If I may suggest something else, the ability to lock down the ALB to accept connections from CF only would be amazing too. More info here: https://github.com/aws/copilot-cli/discussions/3699

[...]

PR #3707, opened 2022-07-01
chore(aws): add function for ec2 api to retrieve prefix lists

CaptainCarpensir:
Add a function to retrieve aws managed prefix list ID queried by name. This will allow Copilot to retrieve prefix lists to set security group ingress to only allow traffic from CloudFront.

Related #3701

### DIFF
--- a/internal/pkg/aws/ec2/ec2.go
+++ b/internal/pkg/aws/ec2/ec2.go
@@ -14,8 +14,9 @@ import (
 )
 
 const (
-	defaultForAZFilterName  = "default-for-az"
-	internetGatewayIDPrefix = "igw-"
+	defaultForAZFilterName   = "default-for-az"
+	internetGatewayIDPrefix  = "igw-"
+	cloudFrontPrefixListName = "com.amazonaws.global.cloudfront.origin-facing"
 
 	// TagFilterName is the filter name format for tag filters
 	TagFilterName = "tag:%s"
@@ -439,8 +440,8 @@ func (idx *routeTableIndex) IsPublicSubnet(subnetID string) bool {
 	return idx.mainTable.HasIGW()
 }
 
-// ManagedPrefixListId returns the PrefixListId of a prefix list queried by name.
-func (c *EC2) ManagedPrefixListId(prefixListName string) (*string, error) {
+// managedPrefixListId returns the DescribeManagedPrefixListsOutput of a query by name.
+func (c *EC2) managedPrefixList(prefixListName string) (*ec2.DescribeManagedPrefixListsOutput, error) {
 	prefixListOutput, err := c.client.DescribeManagedPrefixLists(&ec2.DescribeManagedPrefixListsInput{
 		Filters: []*ec2.Filter{
 			{
@@ -454,17 +455,28 @@ func (c *EC2) ManagedPrefixListId(prefixListName string) (*string, error) {
 		return nil, fmt.Errorf("describe managed prefix list with name %s: %w", prefixListName, err)
 	}
 
+	return prefixListOutput, nil
+}
+
+// CloudFrontManagedPrefixListId returns the PrefixListId of the associated cloudfront prefix list as a *string.
+func (c *EC2) CloudFrontManagedPrefixListId() (*string, error) {
+	prefixListsOutput, err := c.managedPrefixList(cloudFrontPrefixListName)
+
+	if err != nil {
+		return nil, err
+	}
+
 	var ids []*string
-	for _, v := range prefixListOutput.PrefixLists {
+	for _, v := range prefixListsOutput.PrefixLists {
 		ids = append(ids, v.PrefixListId)
 	}
 
 	if len(ids) == 0 {
-		return nil, fmt.Errorf("cannot find any prefix list with name: %s", prefixListName)
+		return nil, fmt.Errorf("cannot find any prefix list with name: %s", cloudFrontPrefixListName)
 	}
 
 	if len(ids) > 1 {
-		return nil, fmt.Errorf("found more than one prefix list with the name %s: %v", prefixListName, ids)
+		return nil, fmt.Errorf("found more than one prefix list with the name %s: %v", cloudFrontPrefixListName, ids)
 	}
 
 	return ids[0], nil
--- a/internal/pkg/aws/ec2/ec2.go
+++ b/internal/pkg/aws/ec2/ec2.go
@@ -37,6 +37,7 @@ type api interface {
 	DescribeNetworkInterfaces(input *ec2.DescribeNetworkInterfacesInput) (*ec2.DescribeNetworkInterfacesOutput, error)
 	DescribeRouteTables(input *ec2.DescribeRouteTablesInput) (*ec2.DescribeRouteTablesOutput, error)
 	DescribeAvailabilityZones(input *ec2.DescribeAvailabilityZonesInput) (*ec2.DescribeAvailabilityZonesOutput, error)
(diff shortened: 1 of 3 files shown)


internal/pkg/aws/ec2/ec2.go
```diff
@@ -437,3 +438,25 @@ func (idx *routeTableIndex) IsPublicSubnet(subnetID string) bool {
 	}
 	return idx.mainTable.HasIGW()
 }
+
+// GetManagedPrefixListId returns the PrefixListId of a prefix list queried by name.
+func (c *EC2) GetManagedPrefixListId(prefixListName string) (*string, error) {
```
iamhopaul123: Maybe 
```suggestion
func (c *EC2) ManagedPrefixListID(prefixListName string) (*string, error) {
```
https://go.dev/doc/effective_go#Getters

internal/pkg/aws/ec2/ec2.go
```diff
@@ -437,3 +438,25 @@ func (idx *routeTableIndex) IsPublicSubnet(subnetID string) bool {
...
+	prefixListOutput, err := c.client.DescribeManagedPrefixLists(&ec2.DescribeManagedPrefixListsInput{
```
iamhopaul123, line 445: <strike>Do we need to handle pagination here? An example would be https://github.com/aws/copilot-cli/blob/109552990cdeeabb1c271b75c64af93734eac5a1/internal/pkg/aws/ecs/ecs.go#L253</strike>

Nvm we also don't need pagination. If the list is too long apparently something wrong happened since we just expect one candidate after the filtering.

internal/pkg/aws/ec2/ec2.go
```diff
@@ -437,3 +438,25 @@ func (idx *routeTableIndex) IsPublicSubnet(subnetID string) bool {
...
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("prefix-list-name"),
+				Values: aws.StringSlice([]string{prefixListName}),
+			},
+		},
+	})
+
+	if err != nil {
```
iamhopaul123, line 454: <strike>Can we handle the error right after calling the API?</strike> sorry misread that

internal/pkg/aws/ec2/ec2.go
```diff
@@ -437,3 +438,25 @@ func (idx *routeTableIndex) IsPublicSubnet(subnetID string) bool {
...
+		return nil, fmt.Errorf("query returned error: %w", err)
+	}
+
+	if len(prefixListOutput.PrefixLists) != 1 {
+		return nil, fmt.Errorf("query `%s` did not return one prefix list", prefixListName)
```
iamhopaul123: I would suggest splitting this error into two: 0 and more than 1, for clearer error msg. Like "cannot find any prefix list with the name %s" and "found more than one prefix list with the name %s: %v" and %v is the prefix list IDs we got.

internal/pkg/aws/ec2/ec2.go
```diff
@@ -437,3 +438,25 @@ func (idx *routeTableIndex) IsPublicSubnet(subnetID string) bool {
...
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("prefix-list-name"),
+				Values: aws.StringSlice([]string{prefixListName}),
+			},
+		},
+	})
+
+	if err != nil {
+		return nil, fmt.Errorf("query returned error: %w", err)
```
Lou1415926: nit: we could provide more context wrapping the error:
```suggestion
		return nil, fmt.Errorf("describe managed prefix list with name %s : %w", prefixListName, err)
```

internal/pkg/aws/ec2/ec2.go
```diff
@@ -437,3 +438,33 @@ func (idx *routeTableIndex) IsPublicSubnet(subnetID string) bool {
...
+	}
+
+	if len(prefixListOutput.PrefixLists) == 0 {
+		return nil, fmt.Errorf("cannot find any prefix list with name: %s", prefixListName)
+	}
+
+	if len(prefixListOutput.PrefixLists) > 1 {
```
Lou1415926: Out of curiosity - when would this happen and why do we want exactly one `PrefixLists`?

internal/pkg/aws/ec2/ec2.go
```diff
@@ -454,8 +454,16 @@ func (c *EC2) GetManagedPrefixListId(prefixListName string) (*string, error) {
...
-		return nil, fmt.Errorf("query `%s` did not return one prefix list", prefixListName)
+	if len(prefixListOutput.PrefixLists) == 0 {
+		return nil, fmt.Errorf("cannot find any prefix list with name: %s", prefixListName)
+	}
+
+	if len(prefixListOutput.PrefixLists) > 1 {
+		var ids []*string
+		for _, v := range prefixListOutput.PrefixLists {
+			ids = append(ids, v.PrefixListId)
+		}
```
iamhopaul123: nit: we could do this out side of the `if` and check `ids`. If the length is 1 then we return `ids[0]`

internal/pkg/aws/ec2/ec2.go
CaptainCarpensir: We mostly want one `PrefixList` to simplify handling the ID in logic for the stack. We can't return a list of ID's because we have no way to confirm which ID is the one we need (in the event of a query returning multiple). If I were to return all of them, we'd have to return the list of prefix lists in their entirety to maybe have the information to separate them, so I thought it would be simpler to require querying for a unique ID, and to return an error if the query doesn't result in pulling a unique one.

Now I don't think it would ever happen, at least not in this use case, so it's mostly just a case for future proofing.

iamhopaul123: Yeah it's a managed prefix list by EC2 and you cannot even delete it on your own. Ideally there should be only one for cloudfront per region and they always have the same id same name for every account (same region).

Lou1415926: Gotcha. My concern was just that, if the expectation of "exactly one prefix list" is unique to CloudFront, perhaps we should handle it in the caller that calls `ManagedPrefixListID(cloudFrontPrefixListName)` instead of here. No biggie.  Happy to approve - if it happen in the future that a caller to this method expect it to faithfully return all prefix list ids, we can do a refactor which is not going to be difficult.

Lou1415926 (approved):
LGTM! Adding the DNM label just in case you have anything to add / address regarding my comment. Feel free to keep it as is and just remove the label!

internal/pkg/aws/ec2/ec2.go
CaptainCarpensir: Sounds good to me

internal/pkg/aws/ec2/ec2.go
```diff
@@ -437,3 +439,45 @@ func (idx *routeTableIndex) IsPublicSubnet(subnetID string) bool {
...
+
+	if err != nil {
+		return nil, fmt.Errorf("describe managed prefix list with name %s: %w", prefixListName, err)
+	}
+
+	return prefixListOutput, nil
+}
+
+// CloudFrontManagedPrefixListId returns the PrefixListId of the associated cloudfront prefix list as a *string.
+func (c *EC2) CloudFrontManagedPrefixListId() (*string, error) {
```
iamhopaul123:
```suggestion
func (c *EC2) CloudFrontManagedPrefixListID() (*string, error) {
```

iamhopaul123 (approved):

PR #3708, opened 2022-07-01
chore: add DeployedJob selector method

bvtujo:
- chore: add DeployedJob method to DeploySelector and refactor filters to be job/svc agnostic.
- chore: fix mockgen
- test: add unit tests for DeployedJob()

This also fixes the makefile as far as I can tell. 


internal/pkg/term/selector/selector.go
```diff
@@ -534,9 +545,91 @@ func (s *CFTaskSelector) Task(msg, help string, opts ...GetDeployedTaskOpts) (st
 	return choice, nil
 }
 
+// DeployedJob has the user select a deployed job. Callers can provide either a particular environment,
+//// a particular service to filter on, or both.
```
paragbhingre:
```suggestion
// a particular service to filter on, or both.
```

internal/pkg/term/selector/selector.go
```diff
@@ -484,6 +485,16 @@ func (s *DeployedService) String() string {
 	return fmt.Sprintf("%s (%s)", s.Svc, s.Env)
 }
 
+// DeployedJob contains the job name and environment name of the deployed job.
+type DeployedJob struct {
+	Job string
+	Env string
+}
+
+func (j *DeployedJob) String() string {
```
paragbhingre, line 499: can we put comment to this public function?

internal/pkg/term/selector/selector.go
paragbhingre:
```suggestion
// a particular job to filter on, or both.
```

paragbhingre (approved):
LGTM🎉 just few tiny nits

internal/pkg/term/selector/selector.go
```diff
@@ -534,9 +545,91 @@ func (s *CFTaskSelector) Task(msg, help string, opts ...GetDeployedTaskOpts) (st
 	return choice, nil
 }
 
+// DeployedJob has the user select a deployed job. Callers can provide either a particular environment,
+//// a particular service to filter on, or both.
+func (s *DeploySelector) DeployedJob(msg, help string, app string, opts ...GetDeployedWorkloadOpts) (*DeployedJob, error) {
```
iamhopaul123, line 567: It looks like this method has a lot of duplication as `DeployedService`. I wonder if we should instead add a `deployedWorkloads` and having both `DeployedServices` and `DeployedJobs` to use it.

iamhopaul123 (approved):

Issue #1689, opened 2020-11-17
Enable additional IAM policies for pipeline's "build" stage


danieloldberg:
Is there a way to add additional policies to copilot roles?
I currently have a need to add IAM policy to enable codebuild role to pull down from CodeArtifact.
Tried to find something in the docs but couldn't find anything.

efekarakus:
Heya @danieloldberg !

Hmm that's interesting, that'd be a new feature requests for us to enable additional policies to the build stage.

We have a way of adding additional IAM policies to the Task Role through the additional AWS resources feature: https://aws.github.io/copilot-cli/docs/developing/additional-aws-resources/ but that doesn't seem to be what you're asking for.

sekamaneka:
I also need to pull something from an existing S3 bucket in the codebuild stage. I didn't find a way to do that. Adding IAM policies would help me very much!

davivcgarcia:
I was also looking for this! My use case is to have the Build stage (CodeBuild) to get values from SSM/ParameterStore or SecretsManager, and for that I need to have specific IAM permissions added to CodeBuild, such as `ssm:GetParameters` or `secretsmanager:GetSecretValue`.

gautam-nutalapati:
> I was also looking for this! My use case is to have the Build stage (CodeBuild) to get values from SSM/ParameterStore or SecretsManager, and for that I need to have specific IAM permissions added to CodeBuild, such as `ssm:GetParameters` or `secretsmanager:GetSecretValue`.

Same here, we need this feature to get docker credentials stored as secrets in build pipeline. Driving force behind this is 'rate limiting implemented by docker hub'
Update: In case it's useful for anyone else, We ended up using SSM to store secrets for now. As copilot BuildProjectRole contains SSM READ permissions, it works perfectly.

gautam-nutalapati:
Related https://github.com/aws/copilot-cli/issues/2755

Lou1415926:
Closing this because #3709 resovles this!

Issue #2546, opened 2021-07-01
Access secrets manager in buildspec results in access denied.


h5aaimtron:
We want to use docker hub to get around the rate limiting issue for image pulls. Currently, when copilot provisions the IAM role, it does not add secrets manager access for the codebuild. Can this be updated to all this behavior as pulling from docker hub and the rate limit issue are pretty common now.

Lou1415926:
This totally makes sense to me. Thank you very much for the suggestion!! 

For the time being, you can modify the build project role so that it has the permission to read your secrets manager resources. You might want to make the modification directly on the resource, as opposed to via modifying CloudFormation template, as the template will get reset next time you run `copilot pipeline update`.

h5aaimtron:
@Lou1415926 Yeah, I modified the IAM role to give it access to the secrets manager, modified the BuildProject adding the environment vars pointing to the secrets, and finally added the docker login in the buildspec. It all worked fine, but yeah I just thought this was cumbersome.

efekarakus:
Related https://github.com/aws/copilot-cli/issues/2755

PR #3709, opened 2022-07-01
feat(pipeline): allow additional policy to build role

### LINKED ISSUE #2755: Feature request: build project role addon policies
It is well documented that you can create an addon with a ManagedPolicy resource and an Output definition and this policy will be injected into the ECS Task Role. As documented on this page: https://aws.github.io/copilot-cli/docs/developing/additional-aws-resources/#what-does-an-addon-template-look-like

> If you need to add additional policies to your ECS task role, you can define an IAM ManagedPolicy resource in your template that holds the permissions for your task, and then add it as an Output. The permission will be injected into your ECS Task Role.

I am working with pipelines and wish to add permissions to the CodeBuild Build Project Role so I can grant, for example, cross-account CodeArtifact access and allow the Build Project Role to perform a login similar to:

`aws codeartifact login --tool pip --repository artifacts --domain <domain> --domain-owner <owner>`

I asked about this in the the project discussion: https://github.com/aws/copilot-cli/discussions/2535

It seems this is not something that is currently supported.

Proposal:
Offer a way to configure Outputs for addon templates that can attach to the Build Project Role and continue to support Task Role injection as well.

paragbhingre:
Resolves #2755


internal/pkg/template/templates/cicd/pipeline_cfn.yml
```diff
@@ -124,6 +124,22 @@ Resources:
           {{- end }} {{/* endif ne .Source.OutputArtifactFormat "" */}}{{- end }} {{/* endif ne .Source.ProviderName "GitHubV1" */}}
       Roles:
         - !Ref BuildProjectRole
+  {{- if isSet .Build.AdditionalPolicy }}
```
dannyrandall:
```suggestion
  {{- if .Build.AdditionalPolicy }}
```
i don't think you need `isSet`

internal/pkg/manifest/pipeline.go
```diff
@@ -178,8 +183,9 @@ type Source struct {
 
 // Build defines the build project to build and test image.
 type Build struct {
-	Image     string `yaml:"image"`
-	Buildspec string `yaml:"buildspec,omitempty"`
+	Image            string            `yaml:"image"`
+	Buildspec        string            `yaml:"buildspec,omitempty"`
+	AdditionalPolicy *AdditionalPolicy `yaml:"additional_policy,omitempty"`
```
dannyrandall: what would you think about just having this as `yaml.Node` or `map[string]any` and marshaling it directly into the pipeline template? and just letting cloudformation do the validation?

iamhopaul123: Agreed. I think right now it is pretty much the same as what CFN has (we don't do any abstraction or setting default values for them). We could just let them specify [the json policy document](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-policy.html#cfn-iam-policy-policydocument) as a string.

paragbhingre: The only thing I thought of is CFN template errors are little difficult for the customers to understand and solve that is why I kept it this way that we will guide customers with more user friendly errors.

iamhopaul123: Yeah but additional policies is a pretty advanced feature and most likely those users who need to specify additional policies they already have the json policy document handy.

internal/pkg/deploy/pipeline_test.go
```diff
@@ -241,6 +241,18 @@ func TestPipelineBuild_Init(t *testing.T) {
 				BuildspecPath:   "some/path",
 			},
 		},
+		"additional policy is empty": {
```
efekarakus, line 271: should we have a test case where we are marshaling the `mfBuild.AdditionalPolicy`

internal/pkg/deploy/pipeline.go
```diff
@@ -99,9 +101,18 @@ func (b *Build) Init(mfBuild *manifest.Build, mfDirPath string) {
 	if strings.Contains(image, "aarch64") {
 		environmentType = "ARM_CONTAINER"
 	}
+	if mfBuild != nil && mfBuild.AdditionalPolicy.Content != nil {
```
efekarakus:
```suggestion
	if mfBuild != nil && !mfBuild.AdditionalPolicy.IsZero() {
```

internal/pkg/template/templates/cicd/pipeline_cfn.yml
```diff
@@ -124,6 +124,14 @@ Resources:
           {{- end }} {{/* endif ne .Source.OutputArtifactFormat "" */}}{{- end }} {{/* endif ne .Source.ProviderName "GitHubV1" */}}
       Roles:
         - !Ref BuildProjectRole
+  {{- if .Build.AdditionalPolicy }}
+  BuildProjectAdditionalPolicy:
+    Type: 'AWS::IAM::Policy'
+    Properties:
+      PolicyName: !Sub ${AWS::StackName}-BuildProjectAdditionalPolicy
+  {{indent 4 .Build.AdditionalPolicy}}
```
efekarakus: If I'm understanding correctly, this is assume that clients will write the pipeline manifest like this:
```yaml
build:
  additional_policy:
    PolicyDocument:
      Statement: ...
```

It's not super clear why `Roles` and `PolicyName` cannot be set in `additional_policy`. Can we instead change the struct to be something like this?
```go
type Build struct {
   // ..
   AdditionalPolicy struct {
      Document yaml.Node `yaml:"document",omitempty`
   } `yaml:"additional_policy,omitempty"`
}
```
The above makes it clear to me that only `PolicyDocument` will be customizable by clients

efekarakus (commented):
Yay awesome, this will provide a lot more flexibility to customers! 🎉 

Is it also possible to add a local integration test in for example https://github.com/aws/copilot-cli/blob/1caa364fd59f1bab4e0bc6a09d1d271fff24f5ab/internal/pkg/deploy/cloudformation/stack/bb_pipeline_integration_test.go#L25 to make sure the additonal policy is rendered in the template?

internal/pkg/template/templates/cicd/pipeline_cfn.yml
paragbhingre: Yeah this looks clean, I will change it.

internal/pkg/deploy/pipeline.go
```diff
@@ -99,9 +101,18 @@ func (b *Build) Init(mfBuild *manifest.Build, mfDirPath string) {
 	if strings.Contains(image, "aarch64") {
 		environmentType = "ARM_CONTAINER"
 	}
+	if mfBuild != nil && !mfBuild.AdditionalPolicy.Document.IsZero() {
+		additionalPolicy, err := yaml.Marshal(&mfBuild.AdditionalPolicy.Document)
+		if err != nil {
+			return fmt.Errorf("marshal pipeline manifest to embed in template: %v", err)
```
efekarakus:
```suggestion
			return fmt.Errorf("marshal `additional_policy.PolicyDocument` in pipeline manifest: %v", err)
```

internal/pkg/template/templates/cicd/pipeline_cfn.yml
```diff
@@ -124,6 +124,15 @@ Resources:
...
         - !Ref BuildProjectRole
+  {{- if .Build.AdditionalPolicyDocument }}
+  BuildProjectAdditionalPolicy:
+    Type: 'AWS::IAM::Policy'
+    Properties:
+      PolicyName: !Sub ${AWS::StackName}-BuildProjectAdditionalPolicy
+      PolicyDocument:
+{{.Build.AdditionalPolicyDocument | indent 8}}
+      Roles:
+        - !Ref BuildProjectRole {{- end}}
```
efekarakus: nit:
```suggestion
        - !Ref BuildProjectRole
  {{- end}}
```

efekarakus (commented):
Looks great! We can :shipit: after fixing the integ test

internal/pkg/template/templates/cicd/pipeline_cfn.yml
```diff
@@ -124,6 +124,16 @@ Resources:
...
+  {{- if .Build.AdditionalPolicyDocument }}
+  BuildProjectAdditionalPolicy:
+    Type: 'AWS::IAM::Policy'
+    Properties:
+      PolicyName: !Sub ${AWS::StackName}-BuildProjectAdditionalPolicy
+      PolicyDocument:
+{{.Build.AdditionalPolicyDocument | indent 8}}
+      Roles:
+        - !Ref BuildProjectRole
+{{- end}}
```
efekarakus:
```suggestion
  {{- end}}
```

efekarakus (approved):

internal/pkg/deploy/pipeline.go
```diff
@@ -99,9 +101,18 @@ func (b *Build) Init(mfBuild *manifest.Build, mfDirPath string) {
 	if strings.Contains(image, "aarch64") {
 		environmentType = "ARM_CONTAINER"
 	}
+	if mfBuild != nil && !mfBuild.AdditionalPolicy.Document.IsZero() {
+		additionalPolicy, err := yaml.Marshal(&mfBuild.AdditionalPolicy.Document)
```
dannyrandall, line 105: do we need to pass the pointer to the document? could we just do `mfBuild.AdditionalPolicy.Document`?

internal/pkg/deploy/pipeline_test.go
```diff
@@ -203,6 +204,17 @@ func TestPipelineBuild_Init(t *testing.T) {
...
+	yamlNode := yaml.Node{}
+	policyDocument := []byte(`
+  Statement:
+    Action: '*'
+    Effect: Allow
+    Resource: '*'
+  Version: 2012-10-17`)
+
+	if err := yaml.Unmarshal(policyDocument, &yamlNode); err != nil {
+		t.Errorf("unmarshal error:  %v", err)
```
dannyrandall: let's do `require.NoError(t, err)`, just to be consistent with the other failures

internal/pkg/deploy/cloudformation/stack/dd_pipeline_integration_test.go
```diff
@@ -0,0 +1,86 @@
...
+				KeyArn:     "arn:aws:kms:us-west-2:1111:key/abcd",
+			},
+		},
+		AdditionalTags: nil,
+	})
+
+	actual, err := serializer.Template()
+	require.NoError(t, err, "template should have rendered successfully")
+	actualInBytes := []byte(actual)
+	m1 := make(map[interface{}]interface{})
```
dannyrandall:
```suggestion
	m1 := make(map[any]any)
```

dannyrandall (approved):
just some small things but looks good!!

huanjani:
Resolves https://github.com/aws/copilot-cli/issues/1689, https://github.com/aws/copilot-cli/issues/2546, https://github.com/aws/copilot-cli/discussions/2535, https://github.com/aws/copilot-cli/discussions/2104

internal/pkg/deploy/pipeline_test.go
paragbhingre: I think we need `require.Error(t, err)` as we are going inside if only when error occurs. What do you think about it?

internal/pkg/deploy/pipeline.go
paragbhingre: Actually when we `unmarshal` the whole manifest we point to the `Pipeline` variable in which we store the unmarshalled manifest. And while we `Marshal` is back into bytes then also we need pointer to the same variable. So indeed we need `&mfBuild.AdditionalPolicy.Document` else we get empty response from `Marshal`. 

Does that make sense?

internal/pkg/deploy/pipeline_test.go
dannyrandall: i think you can remove the if check and just do something like this:
```
require.NoError(t, yaml.Unmarshal(policyDocument, &yamlNode))
```

Issue #3522, opened 2022-05-03
[Design] Environment manifest


Lou1415926:
This issue is a design proposal to introduce manifest files for environments. 

By introducing a declarative configuration similar to service or job manifests, environments will gain all the benefits of infrastructure-as-code: code reviews, easy to replicate and mutate. The manifest file will also allow Copilot to surface more environment configuration over time.

### Problem statement
Today, environments can only be configured during creation with `copilot env init <flags>`. It’s not possible to toggle features without re-creating the environment (such as `--container-insights`). This becomes especially problematic as Copilot surfaces new environment configuration fields such as [`--import-cert-arns`](https://github.com/aws/copilot-cli/pull/3503) (https://github.com/aws/copilot-cli/pull/3503) that users cannot enable in their existing environments. Furthermore, commands cannot be code reviewed to ensure the organization standards are met.

#### New deploy workflow
Running `copilot env init -n <name>` will now generate a manifest file for the environment under:
```
copilot/
└── environments/
    └── test/
        └── manifest.yml
```
After peeking and optionally making modifications to the manifest file, users will be able to deploy their environments using a new command: `copilot env deploy`.
**Note: this means that once this feature is released, `env init` will no longer automatically deploy environments. Instead, clients will have to run `env deploy` to create or update environment stacks.**

In order to view the CloudFormation template that will be generated, users can run `copilot env package`.

In order to generate a manifest file for existing environments, a new flag will be introduced to `copilot env show --manifest` that will print the manifest file representing their existing environment. Therefore, users with existing environments can create a manifest using `env show —manifest` instead of having deleting and re-creating the environment for the sake of manifest.

#### Manifest specification
The environment manifest file will initially ship with just a parity with the existing `env init` flags:
```yaml
name: prod
type: Environment
network:
    vpc:
        id: 'vpc-12345'           # Import.
        cidr: '10.0.0.0/16'       # Configure.
        subnets:
            public:
                - id: 'subnet-11111'    # Import.
                  cidr: '10.0.0.0/24'   # Configure.
                  az: 'us-east-2a'      # Configure.
                - id: 'subnet-22222'    # Import.
                  cidr: '10.0.1.0/24'   # Configure.
                  az: 'us-east-2b'      # Configure.
            private:
                - id: 'subnet-33333'    # Import.
                  cidr: '10.0.3.0/24'   # Configure.
                  az: 'us-east-2a'      # Configure.
                - id: 'subnet-44444'    # Import.
[...]

Issue #3510, opened 2022-04-28
Multiple docker builds causing slow build


srbartlett:
Hi, I am looking for guidance on the build pipeline step.  Firstly, thanks for providing such a great tool chain.  Life is better with Copilot 👍  

After upgrading from 1.5 to 1.16, I rebuilt the pipeline and have noticed my build step takes significantly longer to complete.  I have x2 environments and x2 services (LBS and Backend Service).

The build step runs `docker build` four times.  Is it necessary to build an image per environment per services? I would have through a docker build for each service which is tagged for the different environments would suffice.  I am missing something?

iamhopaul123:
Hello @srbartlett. Thank you for your love for Copilot! As for the build time, it is a great question. I think the reason is because previously, for each service we use the same image for all the environments it deploys to. However, after v1.16, we use different images. In order to speed up the image building, you could modify your buildspec so that for each service can use the same image for all the environments it deploys to (see [here](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/cicd/buildspec.yml#L57)).

However, I think for the second build for the same image. For example, in your case, LBS image for the second environment, though it requires another `docker build`, it should be real quick because it will get from cache like the following

```
Step 1/4 : FROM public.ecr.aws/nginx/nginx:latest
--
213 | ---> fa5269854a5e
214 | Step 2/4 : COPY nginx.conf /etc/nginx/nginx.conf
215 | ---> Using cache
216 | ---> c5e3c195a161
217 | Step 3/4 : RUN mkdir -p /www/data/frontend
218 | ---> Using cache
219 | ---> a7c736f8961b
220 | Step 4/4 : COPY index.html /www/data/frontend
221 | ---> Using cache
222 | ---> 39bd699b2b7d
```

Could you help me to verify if it is the same for your pipeline?

srbartlett:
Hi @iamhopaul123 ,  thanks for your comment.

Regarding you first point to modify the [buildspec](https://github.com/aws/copilot-cli/blob/76aa0e420bebca8efb6a6b9fd7c8a50b1fc0a420/internal/pkg/template/templates/cicd/buildspec.yml#L57), I am correct to assume if I remove ${env} from tag it will only build the image once?  If so, I'll give it a go.

Regarding the cache, yes subsequent builds use the cache.  In fact, I added `cache_from` to my manifest which helps the initial build.  

[...]

PR #3717, opened 2022-07-05
chore: add buildspec for environment pipelines

efekarakus:
Related #3522

### DIFF
--- a/internal/pkg/template/templates/cicd/env/buildspec.yml
+++ b/internal/pkg/template/templates/cicd/env/buildspec.yml
@@ -1,0 +1,36 @@
+# Buildspec runs in the build stage of your environment pipeline to generate the environment CloudFormation stack config.
+version: 0.2
+phases:
+  install:
+    runtime-versions:
+      ruby: 2.6
+    commands:
+      - echo "cd into $CODEBUILD_SRC_DIR"
+      - cd $CODEBUILD_SRC_DIR
+      # Download the copilot linux binary. You can update this version.
+      - wget -q {{.BinaryS3BucketPath}}/copilot-linux-{{.Version}}
+      - mv ./copilot-linux-{{.Version}} ./copilot-linux
+      - chmod +x ./copilot-linux
+  build:
+    commands:
+      - echo "Run your tests"
+      # - make test
+  post_build:
+    commands:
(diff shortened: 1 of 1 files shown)


internal/pkg/template/templates/cicd/env/buildspec.yml
```diff
@@ -0,0 +1,36 @@
...
+      - mv ./copilot-linux-{{.Version}} ./copilot-linux
+      - chmod +x ./copilot-linux
+  build:
+    commands:
+      - echo "Run your tests"
+      # - make test
+  post_build:
+    commands:
+      - ls -l
+      - export COLOR="false"
```
dannyrandall, line 21: `export CI=true`? or is that only for docker output

dannyrandall (approved):

internal/pkg/template/templates/cicd/env/buildspec.yml
```diff
@@ -0,0 +1,37 @@
...
+      - export CI="true"
+      - pipeline=$(cat $CODEBUILD_SRC_DIR/{{.ManifestPath}} | ruby -ryaml -rjson -e 'puts JSON.pretty_generate(YAML.load(ARGF))')
+      - stages=$(echo $pipeline | jq -r '.stages[].name')
+      # Generate the cloudformation templates.
+      - >
+        for env in $stages; do
+          ./copilot-linux env package -n $env --output-dir './infrastructure' --upload-assets;
```
dannyrandall, line 28: could we put the generated files somewhere else? like `/infrastructure`? i know we'd need to fix it for services as well, but it would be nice if we didn't create anything in the users repo to avoid https://github.com/aws/copilot-cli/issues/3510

huanjani (approved):
Putting the label back on so you can address Danny's question.

internal/pkg/template/templates/cicd/env/buildspec.yml
efekarakus: That's a great idea! Although it seems to be more of a concern for workloads rather than environments because Docker isn't needed for env pipelines 💭

dannyrandall: good point😊

PR #3718, opened 2022-07-05
chore: allow cloudfront distribution creation via env manifest

CaptainCarpensir:
Allows CloudFront to be deployed in front of a load balancer with no certificate/domain set. A following PR will be made to allow the user to configure the environment ALB to allow public traffic to the load balancer or not while CDN is enabled.

PR as part of design in #3701 which satisfies requirements 1, 2, 4, 6, and 12. Milestone 1 still not met with this PR.


internal/pkg/cli/deploy/env.go
```diff
@@ -104,6 +111,24 @@ func (d *envDeployer) UploadArtifacts() (map[string]string, error) {
 	return d.legacyUploadCustomResources(resources.S3Bucket)
 }
 
+func (d *envDeployer) getPrefixListId(in *DeployEnvironmentInput) (*string, error) {
```
```suggestion
func (d *envDeployer) cfManagedPrefixListID(in *DeployEnvironmentInput) (*string, error) {
```

internal/pkg/cli/deploy/env.go
```diff
@@ -104,6 +111,24 @@ func (d *envDeployer) UploadArtifacts() (map[string]string, error) {
...
+	if in.Manifest != nil {
+		if in.Manifest.CDN.PublicIngressAllowed() {
+			return nil, nil
+		}
+	}
+
+	if d.ec2 == nil {
+		return nil, nil
+	}
```
iamhopaul123: We don't have to do this. we just need to make sure we always use `NewEnvDeployer` to initialize the struct.

internal/pkg/cli/deploy/env.go
```diff
@@ -104,6 +111,24 @@ func (d *envDeployer) UploadArtifacts() (map[string]string, error) {
 	return d.legacyUploadCustomResources(resources.S3Bucket)
 }
 
+func (d *envDeployer) getPrefixListId(in *DeployEnvironmentInput) (*string, error) {
+	if in.Manifest != nil {
```
iamhopaul123: We probably don't need this validation because the manifest should be there all the time (@Lou1415926 is it true?). Otherwise we might update the stack and unexpectedly remove the ingress control because of that.

internal/pkg/template/env.go
```diff
@@ -103,10 +104,17 @@ type EnvOpts struct {
...
 	Telemetry                *Telemetry
 
-	LatestVersion      string
+	CDNConfig CDNConfig
+
+	LatestVersion string
 	SerializedManifest string // Serialized manifest used to render the environment template.
 }
 
+type CDNConfig struct {
```
iamhopaul123: Comment pls

internal/pkg/template/env.go
```diff
@@ -103,10 +104,17 @@ type EnvOpts struct {
 	AllowVPCIngress          bool
 	Telemetry                *Telemetry
 
-	LatestVersion      string
+	CDNConfig CDNConfig
```
iamhopaul123:
```suggestion
	CDNConfig *CDNConfig
```
Then maybe we don't need `EnableCDN bool` at all?

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -0,0 +1,24 @@
...
+      Origins:
+        - CustomOriginConfig:
+            OriginProtocolPolicy: http-only # Will need to change with TLS termination and Alias usage
+          DomainName: !GetAtt PublicLoadBalancer.DNSName
+          Id: !Sub 'copilot-${AppName}-${EnvironmentName}-origin'
+      # ViewerCertificate: # Used for Aliases
+      #   AcmCertificateArn: # Imported certificate goes here
+      #   CloudFrontDefaultCertificate: true # If importing certificate, set to false
+      #   MinimumProtocolVersion: TLSv1 # Default for now, may need to change later
+      #   SsLSupportMethod: sni-only # Changing this can incur costs to the user
```
iamhopaul123, line 24: do we need to get rid of these?

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -296,6 +297,33 @@ func (e *BootstrapEnvStackConfig) ToEnv(stack *cloudformation.Stack) (*config.En
 	}, nil
 }
 
+func (e *EnvStackConfig) cdnConfig() template.CDNConfig {
```
iamhopaul123: nit we could return `*template.CDNConfig` to simplify the logic in this function and use

```
{{- if .CDNConfig}}
```

in the template file.

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -296,6 +297,33 @@ func (e *BootstrapEnvStackConfig) ToEnv(stack *cloudformation.Stack) (*config.En
...
+	var (
+		enabled      bool
+		prefixListId *string
+	)
+
+	if e.in.Mft != nil {
```
iamhopaul123: same here

internal/pkg/manifest/env.go
```diff
@@ -69,6 +71,73 @@ type environmentVPCConfig struct {
...
+	}
+	return false
+}
+
+// CDNEnabled returns whether a CDN configuration has been enabled in the environment manifest.
+func (e *environmentCDNConfig) CDNEnabled() bool {
+	if e.EnableCDN != nil {
+		if *e.EnableCDN {
+			return true
+		}
```
iamhopaul123:
```suggestion
		return *e.EnableCDN
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -296,6 +297,33 @@ func (e *BootstrapEnvStackConfig) ToEnv(stack *cloudformation.Stack) (*config.En
...
+			enabled = true
+			prefixListId = e.in.PrefixListID
+		}
+
+		// CDN Config specified by basic boolean enabler
+		if e.in.Mft.CDN.EnableCDN != nil {
+			if *e.in.Mft.CDN.EnableCDN {
+				enabled = true
+			}
+		}
```
iamhopaul123: I know your current implementation is the optimal one (with fewest if-else). However, since it doesn't really take much longer (could be ignored) and might make it more readable and less repetitive code we could do
```suggestion
                enable = e.in.Mft.CDN.CDNEnabled()
                if !e.in.Mft.CDN.CDNConfig.IsEmpty() {
                    prefixListId = e.in.PrefixListID
                }
```
since stack pkg doesn't really care how mft package is implemented

internal/pkg/template/env.go
```diff
@@ -59,6 +59,7 @@ func LeastVersionForFeature(feature string) string {
 var (
 	// Template names under "environment/partials/".
 	envCFSubTemplateNames = []string{
+		"cdn-resources",
```
iamhopaul123, line 62: oh nice actually this is a good finding

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -56,6 +56,9 @@ Conditions:
   HasAliases:
     !Not [!Equals [ !Ref Aliases, "" ]]
 Resources:
+{{- if .CDNConfig.EnableCDN}}
```
iamhopaul123: Can we make CFN changes a separate PR 🙏 There are a lot to digest and by making another PRs we could potentially reduce the back-and-forth

internal/pkg/manifest/env.go
```diff
@@ -69,6 +71,73 @@ type environmentVPCConfig struct {
...
+		return true
+	}
+
+	return false
+}
+
+// UnmarshalYAML overrides the default YAML unmarshaling logic for the environmentCDNConfig
+// struct, allowing it to perform more complex unmarshaling behavior.
+// This method implements the yaml.Unmarshaler (v3) interface.
+func (e *environmentCDNConfig) UnmarshalYAML(value *yaml.Node) error {
```
iamhopaul123: We might need to update the transformer file since this is a union type?

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -0,0 +1,24 @@
+cloudfrontdistribution:
```
efekarakus: nit: to be consistent with our other logical IDs
```suggestion
CloudFrontDistribution:
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -100,6 +103,14 @@ Resources:
     Type: AWS::EC2::SecurityGroup
     Properties:
       GroupDescription: Access to the public facing load balancer
+       # Change load balancer to only accept ingress from CloudFront when prefix list specified
+{{- if .CDNConfig.PrefixListID}}
+      SecurityGroupIngress:
```
efekarakus: Should this be a list? https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html#cfn-ec2-securitygroup-securitygroupingress
```suggestion
      SecurityGroupIngress:
        - SourcePrefixListId:
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -100,6 +103,14 @@ Resources:
...
+        SourcePrefixListId: {{.CDNConfig.PrefixListID}}
+        FromPort: 0
+        IpProtocol: tcp
+        ToPort: 0
```
efekarakus: Is there a way of narrowing this down to be more specific for example can we do:
```yaml
        FromPort: 443
        IpProtocol: tcp
        ToPort: 443 # and another rule ToPort: 80
```

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -0,0 +1,24 @@
...
+  Metadata:
+    'aws:copilot:description': 'A CloudFront distribution for global content delivery'
+  Condition: CreateALB
+  Type: AWS::CloudFront::Distribution
+  Properties:
+    DistributionConfig:
+      DefaultCacheBehavior:
+        AllowedMethods: ["GET", "HEAD", "OPTIONS", "PUT", "PATCH", "POST", "DELETE"]
+        CachePolicyId: 4135ea2d-6df8-44a3-9df3-4b5a84be39ad
```
efekarakus: 😮 How is this value determined?

internal/pkg/manifest/env.go
```diff
@@ -69,6 +71,73 @@ type environmentVPCConfig struct {
 	Subnets subnetsConfiguration `yaml:"subnets,omitempty"`
 }
 
+type environmentCDNConfig struct {
+	EnableCDN *bool
+	CDNConfig AdvancedCDNConfig // mutually exclusive with EnableCDN
```
```suggestion
	Enabled *bool
	Advanced AdvancedCDNConfig // mutually exclusive with Enabled
```

internal/pkg/manifest/env.go
```diff
@@ -69,6 +71,73 @@ type environmentVPCConfig struct {
...
+}
+
+// AdvancedCDNConfig represents an advanced configuration for a Content Delivery Network.
+type AdvancedCDNConfig struct {
+	PrefixListIngress *bool `yaml:"public_ingress_allowed"`
```
efekarakus: I wonder if this field will be clear enough to customers, what do you think of flipping the statement so that the field is by default false?
```suggestion
	LimitIngressToCDN *bool `yaml:"limit_origin_ingress"` 
```

internal/pkg/manifest/env.go
```diff
@@ -69,6 +71,73 @@ type environmentVPCConfig struct {
...
+func (e *environmentCDNConfig) PublicIngressAllowed() bool {
+	if e.CDNConfig.PrefixListIngress != nil {
+		return *e.CDNConfig.PrefixListIngress
+	}
+	if e.EnableCDN != nil {
+		if *e.EnableCDN {
+			return true
+		}
+	}
+	return false
```
efekarakus:
```suggestion
	return aws.BoolValue(cfg.Enabled)
```

internal/pkg/manifest/env.go
```diff
@@ -69,6 +71,73 @@ type environmentVPCConfig struct {
...
+	CDNConfig AdvancedCDNConfig // mutually exclusive with EnableCDN
+}
+
+// AdvancedCDNConfig represents an advanced configuration for a Content Delivery Network.
+type AdvancedCDNConfig struct {
+	PrefixListIngress *bool `yaml:"public_ingress_allowed"`
+}
+
+// IsEmpty returns whether AdvancedCDNConfig is empty.
+func (a *AdvancedCDNConfig) IsEmpty() bool {
```
efekarakus: nit: can we rename all Config receivers to
```suggestion
func (cfg *AdvancedCDNConfig) IsEmpty() bool {
```

internal/pkg/manifest/env_test.go
```diff
@@ -617,3 +656,79 @@ func TestEnvironmentObservability_IsEmpty(t *testing.T) {
...
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			got := tc.in.PublicIngressAllowed()
+			require.Equal(t, tc.wanted, got)
+		})
+	}
+}
+
+func TestEnvironmentCDNConfig_CDNEnabled(t *testing.T) {
```
efekarakus, line 848: nice test cases!

internal/pkg/template/templates/environment/partials/cdn-resources.yml
CaptainCarpensir: The cache policy id is just a global id given by AWS, so I'm fairly certain can be hardcoded in this code.
[Here](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-cache-policies.html) is more about this.

CaptainCarpensir: The cache policy id is just a global id given by AWS, so I'm fairly certain can be hardcoded in this code.
[Here](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-cache-policies.html) is more about this.

internal/pkg/manifest/env.go
CaptainCarpensir: Will keep in mind as part of why I'm now splitting this PR is so that I can move this field to the environment ALB (in environmentHTTPConfig) so as to more accurately reflect where the security group resides. And yes, personally I think it's better to have public ingress disabled by default, so I'll change this for the next PR after splitting this one.

internal/pkg/template/templates/environment/partials/cdn-resources.yml
efekarakus: Nice! can you add this as a comment next to the value?
```suggestion
        CachePolicyId: 4135ea2d-6df8-44a3-9df3-4b5a84be39ad # See https://go.aws/3bJid3k
```

internal/pkg/template/templates/environment/cf.yml
CaptainCarpensir: For this and the below suggestion, it's relevant to the prefix list, so I'll address it in the next PR, and am removing these changes from the current one to reduce the scope. See comment below that as well.

internal/pkg/manifest/env_test.go
```diff
@@ -801,3 +819,65 @@ func TestEnvironmentObservability_IsEmpty(t *testing.T) {
...
+				Enabled: aws.Bool(true),
+			},
+			wanted: true,
+		},
+		// "enabled via struct": {
+		// 	in: environmentCDNConfig{
+		// 		CDNConfig: AdvancedCDNConfig{},
+		// 	},
+		// 	wanted: true,
+		// },
```
iamhopaul123: 🤔  can we get rid of unused code?

internal/pkg/manifest/env.go
```diff
@@ -98,6 +100,57 @@ type environmentVPCConfig struct {
 	Subnets subnetsConfiguration `yaml:"subnets,omitempty"`
 }
 
+type environmentCDNConfig struct {
+	Enabled   *bool
+	CDNConfig AdvancedCDNConfig // mutually exclusive with Enabled
+}
+
+// AdvancedCDNConfig represents an advanced configuration for a Content Delivery Network.
+type AdvancedCDNConfig struct{}
```
iamhopaul123: Does this have to be public?

internal/pkg/template/env.go
```diff
@@ -103,10 +104,16 @@ type EnvOpts struct {
...
 
+	CDNConfig *CDNConfig // If nil, no cdn is to be used
+
 	LatestVersion      string
 	SerializedManifest string // Serialized manifest used to render the environment template.
 }
 
+// CDNConfig represents a Content Delivery Network deployed by CloudFront.
+type CDNConfig struct {
+}
```
iamhopaul123: very nit
```suggestion
type CDNConfig struct {}
```

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -0,0 +1,24 @@
...
+      DefaultCacheBehavior:
+        AllowedMethods: ["GET", "HEAD", "OPTIONS", "PUT", "PATCH", "POST", "DELETE"]
+        CachePolicyId: 4135ea2d-6df8-44a3-9df3-4b5a84be39ad # See https://go.aws/3bJid3k
+        TargetOriginId: !Sub 'copilot-${AppName}-${EnvironmentName}-origin'
+        ViewerProtocolPolicy: allow-all # Will need to change with TLS termination and Alias usage
+      Enabled: true
+      IPV6Enabled: true
+      Origins:
+        - CustomOriginConfig:
+            OriginProtocolPolicy: http-only # Will need to change with TLS termination and Alias usage
```
iamhopaul123: Why is this `http-only`? Does that mean CF will always terminate TLS for users?

internal/pkg/manifest/env_test.go
CaptainCarpensir: Yeah, was there to remind myself to add back that test case when CDNConfig has fields to specify, but as there are other things to change as well like `IsEmpty()` it's not needed.

CaptainCarpensir: There's similarly commented code in `transform_test.go` that I can get rid of too

internal/pkg/template/templates/environment/cf.yml
CaptainCarpensir: With a max weight of 60, the cloudfront prefix list contains 55 of this, and so we can only have one security group for all of cloudfront.
https://docs.aws.amazon.com/vpc/latest/userguide/working-with-aws-managed-prefix-lists.html#aws-managed-prefix-list-weights
The only way I can think of to limit the port range is by specifying a max of 443 and a min of 80, but that seems pointless, as we then allow each port between then. I'd rather just allow cloudfront access to all ports in that case.

internal/pkg/manifest/env.go
```diff
@@ -98,6 +100,57 @@ type environmentVPCConfig struct {
...
+
+// advancedCDNConfig represents an advanced configuration for a Content Delivery Network.
+type advancedCDNConfig struct{}
+
+// IsEmpty returns whether environmentCDNConfig is empty.
+func (cfg *environmentCDNConfig) IsEmpty() bool {
+	return cfg.Enabled == nil && cfg.CDNConfig.IsEmpty()
+}
+
+// IsEmpty returns whether AdvancedCDNConfig is empty.
```
iamhopaul123: Maybe specify this is a no-op now

internal/pkg/manifest/env.go
```diff
@@ -98,6 +100,57 @@ type environmentVPCConfig struct {
...
+// struct, allowing it to perform more complex unmarshaling behavior.
+// This method implements the yaml.Unmarshaler (v3) interface.
+func (cfg *environmentCDNConfig) UnmarshalYAML(value *yaml.Node) error {
+	if err := value.Decode(&cfg.CDNConfig); err != nil {
+		switch err.(type) {
+		case *yaml.TypeError:
+			break
+		default:
+			return err
+		}
```
iamhopaul123:
```suggestion
		var yamlTypeErr *yaml.TypeError
		if !errors.As(err, &yamlTypeErr) {
			return err
		}
```

iamhopaul123 (approved):

internal/pkg/manifest/transform_test.go
```diff
@@ -1124,3 +1124,55 @@ func TestSecretTransformer_Transformer(t *testing.T) {
...
 	}
 }
+
+func TestEnvironmentCDNConfigTransformer_Transformer(t *testing.T) {
+	testCases := map[string]struct {
+		original func(cfg *environmentCDNConfig)
+		override func(cfg *environmentCDNConfig)
+		wanted   func(cfg *environmentCDNConfig)
+	}{
+		// "enabled set to nil if cdnconfig is not empty": {
```
efekarakus: can we delete the unused comments? 🥺

internal/pkg/deploy/cloudformation/stack/env.go
efekarakus: 👍 can we just return `nil` for now so that it's guaranteed that we don't accidentally include the `cdn` partial to the environment template?

efekarakus (commented):
Looks great! 👍 to Penghao's feedback and 2 small requests, and then we can :shipit:

efekarakus (approved):

PR #3722, opened 2022-07-06
chore(manifest): marshal environment manifest

Lou1415926:


internal/pkg/template/templates/environment/manifest.yml
```diff
@@ -0,0 +1,77 @@
...
+    cidr: {{$vpc.CIDR }}
+    {{- end }}
+    {{- if not $vpc.Subnets.IsEmpty}}{{- $subnets := $vpc.Subnets}}
+    subnets:
+      {{- if $subnets.Public}}
+      public:
+        {{- range $subnet := $subnets.Public}}
+        - {{- if $subnet.SubnetID}} id: {{$subnet.SubnetID}} 
+          {{- else if $subnet.CIDR}} cidr: {{$subnet.CIDR}}
+          {{if $subnet.AZ -}} az: {{$subnet.AZ}} {{- end}}
```
efekarakus, line 30: is this  `{{- end}}` for `subnet.AZ`? I think I'm confused why there isn't an `{{- end}}` after the {{- else if}}

efekarakus: Why did you prefer removing the trailing spaces after, rather than before?
```suggestion
          {{- if $subnet.AZ}} az: {{$subnet.AZ}} {{- end}}
```

internal/pkg/template/templates/environment/manifest.yml
```diff
@@ -0,0 +1,77 @@
...
+    certificates: {{fmtStringSlice $publicHTTP.Certificates}}
+    {{- end}}
+  {{- end}}
+  {{- if not .HTTPConfig.Private.IsEmpty}}{{$privateHTTP := .HTTPConfig.Private}}
+  private:
+    {{- if $privateHTTP.Certificates}}
+    certificates: {{fmtStringSlice $privateHTTP.Certificates}}
+    {{- end}}
+    {{- if $privateHTTP.InternalALBSubnets}}
+    subnets: {{$privateHTTP.InternalALBSubnets}}
```
efekarakus, line 65: sweet, we need to retrofit `--internal-alb-allow-vpc-ingress` in a separate PR right?

internal/pkg/manifest/validate_env.go
```diff
@@ -189,8 +189,8 @@ func (cfg environmentHTTPConfig) Validate() error {
 }
 
 // Validate returns nil if publicHTTPConfig is configured correctly.
-func (o publicHTTPConfig) Validate() error {
-	for idx, certARN := range o.Certificates {
+func (c publicHTTPConfig) Validate() error {
```
efekarakus: nit: can we rename all receivers to `cfg`? (same in env.go)

internal/pkg/manifest/testdata/environment-import-vpc.yml
```diff
@@ -0,0 +1,28 @@
+# The manifest for the "test" environment.
+# Read the full specification for the "Environment" type at:
+#  https://aws.github.io/copilot-cli/docs/manifest/environment/
+
+# Your environment name will be used in naming your resources like VPC, cluster, etc.
```
efekarakus, line 5: omg it's happening!!! 😍

internal/pkg/manifest/testdata/environment-default.yml
```diff
@@ -0,0 +1,21 @@
...
+#  https://aws.github.io/copilot-cli/docs/manifest/environment/
+
+# Your environment name will be used in naming your resources like VPC, cluster, etc.
+Name: test
+Type: Environment
+
+# Import your own VPC and subnets or configure how they should be created.
+# network:
+#   vpc:
+#     id:
```
efekarakus, line 12: This would be so awesome, if we can write eventually
```suggestion
# network:
#   vpc:
#     azs: 3
```

efekarakus (commented):
Looks so great 🔥

internal/pkg/template/templates/environment/manifest.yml
Lou1415926: Because that trims the `\n` (also a whitespace) from the last line as well 😢

internal/pkg/template/templates/environment/manifest.yml
Lou1415926: yeah! we don't have a environment field for it now, we will need a separate PR to add the field + unmarshal + marshal + validate it.

internal/pkg/manifest/testdata/environment-default.yml
Lou1415926: agreed! Currently this manifest is a strictly minimal to achieve parity with existing flags, therefore `azs` isn't there yet. After we roll out `vpc.azs` we can update the default commented-out manifest with it!

efekarakus (approved):

internal/pkg/manifest/env.go
```diff
@@ -229,7 +273,17 @@ type publicHTTPConfig struct {
...
+func (cfg publicHTTPConfig) IsEmpty() bool {
+	return len(cfg.Certificates) == 0
+}
+
 type privateHTTPConfig struct {
 	InternalALBSubnets []string `yaml:"subnets,omitempty"`
 	Certificates       []string `yaml:"certificates,omitempty"`
 }
+
+// IsEmpty returns true if there are customization to the internal ALB.
```
huanjani: I could be confused, but did you mean this?
```suggestion
// IsEmpty returns true if there is no customization to the internal ALB.
```

internal/pkg/manifest/env.go
```diff
@@ -229,7 +273,17 @@ type publicHTTPConfig struct {
 	Certificates []string `yaml:"certificates,omitempty"`
 }
 
+// IsEmpty returns true if there are customization to the public ALB.
```
huanjani: same as below

huanjani:
```suggestion
// IsEmpty returns true if there is no customization to the public ALB.
```

internal/pkg/manifest/testdata/environment-adjust-vpc.yml
```diff
@@ -0,0 +1,32 @@
+# The manifest for the "test" environment.
```
huanjani, line 1: So pretty!

huanjani (approved):
So exciting!

PR #3724, opened 2022-07-06
chore(cli): let `env init` deploys only bootstrap resources

Lou1415926:
This PR modifies `env init`'s behaviors:
1. It deploys environment stack with bootstrap resources only (currently, it means `CFNExecutionRole` and `EnvManagerRole`). Other environment resources such as VPC are deployed in `env deploy`, not `env init`.
2. SSM parameter no longer contains information about user customization to the environment, such as `importVPC` and `adjustVPC`; instead, it only contains basic information about the app and env, and the bootstrap resources.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the Apache 2.0 License.

PR #3725, opened 2022-07-06
chore(cli): let `env init` writes an environment manifest

Lou1415926:
Lets `env init` writes an initial environment manifest to `/copilot/environments/<env_name>/manifest.yml`, given user inputs (either from prompts or flags).

NOTE: depends on #3724. Marked "WIP" until #3724 is merged.

### DIFF
--- a/internal/pkg/cli/env_init.go
+++ b/internal/pkg/cli/env_init.go
@@ -171,6 +171,9 @@ type initEnvOpts struct {
 	uploader     customResourcesUploader
 
 	sess *session.Session // Session pointing to environment's AWS account and region.
+
+	// Cached variables.
+	wsAppName string
 }
 
 func newInitEnvOpts(vars initEnvVars) (*initEnvOpts, error) {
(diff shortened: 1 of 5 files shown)


internal/pkg/cli/env_init.go
```diff
@@ -655,8 +659,45 @@ To recreate the environment, please run:
 	return nil
 }
 
-func (o *initEnvOpts) deployEnv(app *config.Application,
-	artifactBucketARN, artifactBucketKeyARN string) error {
+func (o *initEnvOpts) importVPCConfig() *config.ImportVPC {
```
Lou1415926, line 659: These two `importVPCConfig()` and `adjustVPCConfig()` are existing methods that were removed in #3724 for static check. No changes are made implementation-wise!

internal/pkg/cli/env_init.go
```diff
@@ -187,6 +197,11 @@ func newInitEnvOpts(vars initEnvVars) (*initEnvOpts, error) {
 	}
 
 	prompter := prompt.New()
+
+	ws, err := workspace.New()
+	if err != nil {
+		return nil, fmt.Errorf("workspace cannot be created: %w", err)
```
iamhopaul123:
```suggestion
		return nil, fmt.Errorf("create workspace: %w", err)
```

iamhopaul123 (approved):
Do we consider moving the init process to `initialize` pkg 🤔

Lou1415926:
> Do we consider moving the init process to initialize pkg 🤔

Yes this is on my to do list (and part of the refactor is already done). Right now I haven't done this because I think this is easier for reviewers to read the diff and be confident about the correctness.

efekarakus (approved):
yayay

efekarakus:
added the label for PH's feedback

PR #3726, opened 2022-07-06
ci: update golangci-lint to v1.46.2

rclinard-amzn:
This re-enables the "staticcheck" lint that was disabled due to upgrading to Go 1.18.

structcheck is still disabled since it has not been upgraded to support Go 1.18 in golangci-lint. See https://github.com/golangci/golangci-lint/issues/2649 for more details.

This PR also fixes some of the lint issues that were missed in the past few months.


dannyrandall (approved):
perfect😍

efekarakus (approved):
yay, thank you!

PR #3727, opened 2022-07-06
chore(manifest): add from_tag for workload

iamhopaul123:


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -653,8 +653,9 @@ func convertEFSConfiguration(in manifest.EFSVolumeConfiguration) *template.EFSVo
 	}
 }
 
-func convertNetworkConfig(network manifest.NetworkConfig) template.NetworkOpts {
+func convertNetworkConfig(network manifest.NetworkConfig, subnetIDsFromTags []string) template.NetworkOpts {
 	if network.IsEmpty() {
+		fmt.Println("123")
```
paragbhingre: Can we remove this extra print statements?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -669,21 +670,27 @@ func convertNetworkConfig(network manifest.NetworkConfig) template.NetworkOpts {
 
 	placement := network.VPC.Placement
 	if placement.IsEmpty() {
+		fmt.Println("1234")
```
paragbhingre: Same here

internal/pkg/cli/deploy/svc.go
```diff
@@ -151,6 +152,10 @@ type configDescriber interface {
 	Manifest() ([]byte, error)
 }
 
+type SubnetIDsGetter interface {
```
paragbhingre: Can we make this interface name small case as not using outside this package? And if planning to use it outside of this package then we can put up a comment for this one.

internal/pkg/manifest/workload_test.go
```diff
@@ -389,6 +389,60 @@ func TestPlacementArgOrString_UnmarshalYAML(t *testing.T) {
...
+		"success with string slice": {
+			inContent: []byte(`subnets: ["id1", "id2"]`),
+			wantedStruct: SubnetListOrArgs{
+				IDs: []string{"id1", "id2"},
+			},
+		},
+		"success with args": {
+			inContent: []byte(`subnets:
+  from_tags:
+    foo: bar`),
```
paragbhingre, line 432: Just trying to understand the realtime scenario here. What will be actual values if not `foo: bar`?
Will it be `subnetName: subnetId`?

internal/pkg/manifest/workload.go
```diff
@@ -430,17 +434,71 @@ func (p *PlacementArgOrString) UnmarshalYAML(value *yaml.Node) error {
...
+// This method implements the yaml.Unmarshaler (v3) interface.
+func (s *SubnetListOrArgs) UnmarshalYAML(value *yaml.Node) error {
+	if err := value.Decode(&s.SubnetArgs); err != nil {
+		switch err.(type) {
+		case *yaml.TypeError:
+			fmt.Println(err)
+			break
+		default:
+			return err
+		}
```
dannyrandall: - use `errors.As` instead https://pkg.go.dev/errors
- remove `fmt.Println`

internal/pkg/manifest/workload_test.go
paragbhingre: Okay I got it. I checked the `FilterForTags` method and understood the use case

internal/pkg/cli/deploy/svc.go
```diff
@@ -997,10 +1011,32 @@ func (d *workloadDeployer) runtimeConfig(in *StackRuntimeConfiguration) (*stack.
...
 		Region:                   d.env.Region,
+		SubnetIDs:                subnetIDs,
 		CustomResourcesURL:       in.CustomResourceURLs,
 	}, nil
 }
 
+func (d *workloadDeployer) subnetIDsFromTags() ([]string, error) {
+	subnetFilterTags := subnetTags(d.mft)
+	var subnetIDs []string
+	if subnetFilterTags != nil {
```
dannyrandall:
```suggestion
	if subnetFilterTags == nil {
	    return nil, nil
	}
	var subnetIDs []string
```

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -187,9 +187,15 @@ func (s *LoadBalancedWebService) BuildArgs(wsRoot string) *DockerBuildArgs {
 
 // EnvFile returns the location of the env file against the ws root directory.
 func (s *LoadBalancedWebService) EnvFile() string {
+	s.ContainerPlatform()
```
dannyrandall: why do this...?

internal/pkg/deploy/cloudformation/stack/transformers.go
iamhopaul123: 😮‍💨 i thought i removed them

internal/pkg/manifest/lb_web_svc.go
iamhopaul123: i guess my IDE was broken yesterday. i tried to quit it by force a couple of times 😮‍💨

internal/pkg/aws/ec2/ec2.go
```diff
@@ -50,6 +50,17 @@ type Filter struct {
 	Values []string
 }
 
+// FilterForTags takes a key and optional values to construct an EC2 filter.
+func FilterForTags(key string, values ...string) Filter {
```
efekarakus, line 55: Nice, this is a really clean function signature 👍 😍

internal/pkg/manifest/workload.go
```diff
@@ -331,10 +330,8 @@ func (b *BuildArgsOrString) isEmpty() bool {
 // This method implements the yaml.Unmarshaler (v3) interface.
 func (b *BuildArgsOrString) UnmarshalYAML(value *yaml.Node) error {
 	if err := value.Decode(&b.BuildArgs); err != nil {
-		switch err.(type) {
-		case *yaml.TypeError:
-			break
-		default:
+		var yamlTypeErr *yaml.TypeError
```
Lou1415926, line 352: :+1_cat

efekarakus (commented):
Posting here as well for increase visibility. We were discussing offline how we can make adding the `from_tags` option to other fields easier and here is our code specification proposal:

https://gist.github.com/efekarakus/5859927876e87528f14117cfbd779833

internal/pkg/cli/job_deploy.go
```diff
@@ -44,6 +45,7 @@ type deployJobOpts struct {
 	// cached variables
 	targetApp       *config.Application
 	targetEnv       *config.Environment
+	envSess         *session.Session
```
bvtujo, line 48: 😰 were we not using the correct session for job deploy this entire time>>>

internal/pkg/manifest/loader.go
```diff
@@ -0,0 +1,75 @@
...
+type subnetIDsGetter interface {
+	SubnetIDs(filters ...ec2.Filter) ([]string, error)
+}
+
+type loader interface {
+	load() error
+}
+
+// DynamicWorkloadManifest represents a dynamically populated workload manifest.
+type DynamicWorkloadManifest struct {
```
bvtujo, line 20: this makes so much sense as the next evolution of the manifest types

bvtujo (approved):
nice work boss, looks good to me

internal/pkg/aws/ec2/ec2.go
```diff
@@ -50,6 +51,17 @@ type Filter struct {
...
+	if len(values) == 0 {
+		return Filter{Name: tagKeyFilter, Values: []string{key}}
+	}
```
dannyrandall, line 58: could we add a test case for this?

internal/pkg/cli/svc_deploy.go
```diff
@@ -371,10 +380,13 @@ func workloadManifest(in *workloadManifestInput) (manifest.WorkloadManifest, err
 	}
 	envMft, err := mft.ApplyEnv(in.envName)
 	if err != nil {
-		return nil, fmt.Errorf("apply environment %s override: %s", in.envName, err)
+		return nil, fmt.Errorf("apply environment %s override: %w", in.envName, err)
 	}
 	if err := envMft.Validate(); err != nil {
-		return nil, fmt.Errorf("validate manifest against environment %s: %s", in.envName, err)
+		return nil, fmt.Errorf("validate manifest against environment %s: %w", in.envName, err)
```
dannyrandall: (just bc i just saw a very similar line in another PR)
```suggestion
		return nil, fmt.Errorf("validate manifest against environment %q: %w", in.envName, err)
```

internal/pkg/manifest/loader.go
```diff
@@ -0,0 +1,75 @@
...
+	return &DynamicWorkloadManifest{
+		mft: mft,
+		newSubnetIDsGetter: func(s *session.Session) subnetIDsGetter {
+			return ec2.New(s)
+		},
+	}
+}
+
+// Manifest returns the manifest content.
+func (s *DynamicWorkloadManifest) Manifest() interface{} {
```
dannyrandall: nit
```suggestion
func (s *DynamicWorkloadManifest) Manifest() any {
```

internal/pkg/manifest/workload.go
```diff
@@ -430,17 +447,87 @@ func (p *PlacementArgOrString) UnmarshalYAML(value *yaml.Node) error {
...
+type SubnetListOrArgs struct {
+	IDs []string
+	SubnetArgs
+}
+
+func (s *SubnetListOrArgs) isEmpty() bool {
+	return len(s.IDs) == 0 && s.SubnetArgs.isEmpty()
+}
+
+type dynamicSubnets struct {
```
dannyrandall, line 470: i'm not seeing where this gets used...

internal/pkg/manifest/loader.go
bvtujo: whoa when did this keyword appear!?

internal/pkg/manifest/workload.go
bvtujo: [Line 60 of manifest/loader.go](https://github.com/aws/copilot-cli/pull/3727/files#diff-6272e76645f1b500c7be4973be9e3d455c5f1aeb5350aea1890782dfec9eba20R60)

internal/pkg/manifest/loader.go
dannyrandall: [1.18 😊](https://go.dev/doc/go1.18)

internal/pkg/cli/job_deploy.go
iamhopaul123: it is actually used [here](https://github.com/aws/copilot-cli/blob/36a67d6d3d2068721138851b0f644f6bd6126fd9/internal/pkg/cli/job_deploy.go#L154)

internal/pkg/aws/ec2/ec2.go
```diff
@@ -333,7 +344,9 @@ func (c *EC2) subnets(filters ...Filter) ([]*ec2.Subnet, error) {
 		}
 		subnets = append(subnets, response.Subnets...)
 	}
-
+	if len(subnets) == 0 {
+		return nil, fmt.Errorf("cannot find any subnets")
```
Lou1415926, line 348:
```suggestion
		return nil, errors.New("cannot find any subnets")
```

internal/pkg/deploy/cloudformation/stack/lb_grpc_web_service_integration_test.go
```diff
@@ -48,11 +47,11 @@ func TestGrpcLoadBalancedWebService_Template(t *testing.T) {
 		require.NoError(t, err)
 		envMft, err := mft.ApplyEnv(tc.envName)
 		require.NoError(t, err)
-
 		err = envMft.Validate()
 		require.NoError(t, err)
+		content := envMft.Manifest()
```
Lou1415926, line 52: Do you thing we also want to add the `.Load` step to these local integration test? I feel like otherwise they wouldn't be testing the right thing because in real executions their output template would be one that has gone through `.Load`.

internal/pkg/manifest/transform.go
```diff
@@ -20,10 +20,11 @@ var defaultTransformers = []mergo.Transformers{
 	imageTransformer{},
 	buildArgsOrStringTransformer{},
 	aliasTransformer{},
-	stringSliceOrStringTransformer{},
+	StringSliceOrStringTransformer{},
```
Lou1415926, line 23: Does this transformer need to be exported?

internal/pkg/task/env_runner.go
```diff
@@ -23,8 +23,8 @@ const (
 
 // Names for tag filters
 var (
-	tagFilterNameForApp = fmt.Sprintf(ec2.TagFilterName, deploy.AppTagKey)
-	tagFilterNameForEnv = fmt.Sprintf(ec2.TagFilterName, deploy.EnvTagKey)
+	FmtTagFilterForApp = fmt.Sprintf(ec2.FmtTagFilter, deploy.AppTagKey)
```
Lou1415926, line 26: I might be missing something - do they need to be exported?

internal/pkg/manifest/workload.go
```diff
@@ -252,36 +261,45 @@ func (e *EntryPointOverride) ToStringSlice() ([]string, error) {
...
+type StringSliceOrString struct {
 	String      *string
 	StringSlice []string
 }
 
-func (s *stringSliceOrString) isEmpty() bool {
+// UnmarshalYAML overrides the default YAML unmarshaling logic for the StringSliceOrString
+// struct, allowing it to perform more complex unmarshaling behavior.
+// This method implements the yaml.Unmarshaler (v3) interface.
+func (s *StringSliceOrString) UnmarshalYAML(value *yaml.Node) error {
```
Lou1415926, line 288: This is essentially the same method as `unmarshalYAMLToStringSliceOrString(s *StringSliceOrString, value *yaml.Node) error`. Maybe we can keep this one and remove that one. Then we can replace all 
```go
unmarshalYAMLToStringSliceOrString((*stringSliceOrString)(e), value)
```
to
```
(*stringSliceOrString)(e).UnmarshalYAML(value)
```

internal/pkg/manifest/workload.go
```diff
@@ -430,17 +447,87 @@ func (p *PlacementArgOrString) UnmarshalYAML(value *yaml.Node) error {
...
+	}
+	if len(dyn.cfg.IDs) > 0 {
+		return nil
+	}
+	var filters []ec2.Filter
+	for k, v := range dyn.cfg.FromTags {
+		values := v.StringSlice
+		if v.String != nil {
+			values = []string{aws.StringValue(v.String)}
+		}
```
Lou1415926, line 488: Does 
```go
values := v.toStringSlice()
```
do what you want here?

internal/pkg/manifest/workload.go
```diff
@@ -430,17 +447,87 @@ func (p *PlacementArgOrString) UnmarshalYAML(value *yaml.Node) error {
...
+	return len(s.IDs) == 0 && s.SubnetArgs.isEmpty()
+}
+
+type dynamicSubnets struct {
+	cfg    *SubnetListOrArgs
+	client subnetIDsGetter
+}
+
+// Load populates the subnet's IDs field if the client is using tags.
+func (dyn *dynamicSubnets) load() error {
```
Lou1415926, line 476: I see this method is tested in the `DynamicWorkload`'s `Load`'s test. I think it makes sense. In the future, when we have more and more loader though, we probably would want to test each individual loader separately 💭

dannyrandall (approved):

internal/pkg/manifest/workload.go
```diff
@@ -60,21 +63,27 @@ type WorkloadManifest interface {
 	ApplyEnv(envName string) (WorkloadManifest, error)
 	Validate() error
 	RequiredEnvironmentFeatures() []string
+	Load(sess *session.Session) error
+	Manifest() interface{}
+}
+
+type workloadManifest interface {
+	Validate() error
```
Lou1415926, line 71: Could this be an unexported function as well?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -671,15 +671,16 @@ func convertNetworkConfig(network manifest.NetworkConfig) template.NetworkOpts {
...
+		}
+		opts.SubnetsType = subnetPlacementForTemplate[*placement.PlacementString]
 		return opts
 	}
-	if *placement.PlacementString == manifest.PrivateSubnetPlacement {
-		opts.AssignPublicIP = template.DisablePublicIP
-	}
-	opts.SubnetsType = subnetPlacementForTemplate[*placement.PlacementString]
+	opts.AssignPublicIP = template.DisablePublicIP
+	opts.SubnetsType = ""
```
Lou1415926, line 682: tiny nit: is this line here just to be sure we aren't setting `SubnetsType`? we could probably remove it

Lou1415926 (commented):
Late to the party but 😂 adding the comments just for your reference

internal/pkg/deploy/cloudformation/stack/lb_grpc_web_service_integration_test.go
iamhopaul123: If we add it it is not local anymore lol

Lou1415926: It is not local only if it has `subnet.from_tags` and not `subnet.ids` right?

iamhopaul123: `Load` is supposed to load dynamic contents that requires API calls. If the test itself is supposed to be run locally then it doesn't make sense to me to add `.Load` step 🤔

Lou1415926: I feel like the only reason why these tests have all these `Validate`, `ApplyEnv` calls is because we didn't have a better way to mimic how `cli` package would have manipulate the manifest, than to just hardcoding-ly repeating the codes in `cli`. Without calling `Load`, we are not effectively "mimicking" it.

Another way that I think about it is that I think these integration tests were meant to test that "given a manifest input, do I eventually get an expected template". Without calling `Load` like `cli` does, we can't be confident with our test result.

internal/pkg/deploy/cloudformation/stack/transformers.go
iamhopaul123: this is actually to unset the subnet types since it is contradictory to subnetIDs.

Lou1415926: oh lol didn't realize it was set to `template.PublicSubnetsPlacement` by default. I feel like this function probably also needs some refactoring in the future 😂

internal/pkg/deploy/cloudformation/stack/lb_grpc_web_service_integration_test.go
iamhopaul123: that's true. IMO we need to
1. remove some of our existing integ tests because we have e2e tests now
2. make current local integ tests to be integ tests because we can't generate expected template without making internet.

PR #3729, opened 2022-07-07
chore(templates): Make all IAM policies have a consistent Version declaration and avoid the outdated 2008-10-17 policy language

rclinard-amzn:
This ensures that our CloudFormation templates use the 2012-10-17 IAM policy version.

Since 2008-10-17 is the default policy version, this also adds the Version key to the policies that were missing it.

Based on https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_version.html, the 2008-10-17 policy should not be used for new policies.

This should not change any behavior, but is likely good practice. Since customers might use Copilot's generated CloudFormation templates as a reference, we want to make sure our generated templates follow best practices recommended by other AWS services.

### DIFF
--- a/e2e/multi-pipeline/s3template.yml
+++ b/e2e/multi-pipeline/s3template.yml
@@ -29,7 +29,7 @@ Resources:
     Type: AWS::S3::BucketPolicy
     Properties:
       PolicyDocument:
-        Version: 2012-10-17
+        Version: '2012-10-17'
         Statement:
           - Sid: ForceHTTPS
             Effect: Deny
@@ -52,7 +52,7 @@ Resources:
         - Grants CRUD access to the S3 bucket ${Bucket}
         - { Bucket: !Ref e2epipelineaddonBucket }
       PolicyDocument:
-        Version: 2012-10-17
+        Version: '2012-10-17'
         Statement:
           - Sid: S3ObjectActions
             Effect: Allow
--- a/internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-staging.stack.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-staging.stack.yml
@@ -117,6 +117,7 @@ Resources: # If a bucket URL is specified, that means the template exists.
     Type: AWS::IAM::Role
     Properties:
       AssumeRolePolicyDocument:
+        Version: '2012-10-17'
         Statement:
           - Effect: Allow
             Principal:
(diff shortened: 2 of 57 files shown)


internal/pkg/deploy/cloudformation/stack/testdata/workloads/windows-svc-test.stack.yml
```diff
@@ -108,6 +108,7 @@ Resources: # If a bucket URL is specified, that means the template exists.
     Type: AWS::IAM::Role
     Properties:
       AssumeRolePolicyDocument:
+        Version: '2012-10-17'
```
paragbhingre, line 111: Nice, thanks for adding it to the test data as well :)

internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-staging.stack.yml
```diff
@@ -117,6 +117,7 @@ Resources: # If a bucket URL is specified, that means the template exists.
     Type: AWS::IAM::Role
     Properties:
       AssumeRolePolicyDocument:
+        Version: '2012-10-17'
```
paragbhingre, line 120: Now that I think I curious how these tests were passing without `Version` before? 
Isn't our actual code generates cfn template that has `Version`? interesting... 

No wonder so many integ tests are failing

rclinard-amzn: It looks like having / not having Version was synced between the templates and tests, so they passed because the expected output had the Version data based on where it occurred in the actual output. The integration test failures happened since I only looked at 100 policies out of 290 policies the first time around, which I fixed in the most recent commit.

rclinard-amzn:
This PR has been updated so that now every IAM policy in the CF template YML files has a consistent `Version: '2012-10-17'` declaration. This avoids the version being parsed as a date, even though CF seems to handle this properly.

e2e/multi-pipeline/s3template.yml
```diff
@@ -29,7 +29,7 @@ Resources:
     Type: AWS::S3::BucketPolicy
     Properties:
       PolicyDocument:
-        Version: 2012-10-17
+        Version: '2012-10-17'
```
iamhopaul123, line 32: Why do we add `''` 🤔

efekarakus (approved):

e2e/multi-pipeline/s3template.yml
rclinard-amzn: Without the quotes, the version is parsed as a Date instead of a String. This didn't seem quite right to me, since the policy language version is defined to be a string, even though CloudFormation appears to handle this properly.

This just makes it so that all of the policy language declarations are consistently defined with single quotes to avoid any of that weirdness.

iamhopaul123 (approved):

PR #3730, opened 2022-07-07
chore: don't print internal alb dns name on `svc deploy`

dannyrandall:
For backwards compatibility, we [set both the internal ALB DNS name and the "friendly" DNS name](https://github.com/aws/copilot-cli/blob/74d034755279b4d28b3c6f4b61c9a34aa2e6513b/internal/pkg/template/templates/workloads/partials/cf/http-listener.yml#L51-L58) that we generate as host headers for the service listener rule on internal ALBs. This change makes it so that we only show the friendly DNS name on `svc deploy`.

### DIFF
--- a/internal/pkg/describe/uri.go
+++ b/internal/pkg/describe/uri.go
@@ -179,7 +179,7 @@ func (d *BackendServiceDescriber) URI(envName string) (URI, error) {
 				return URI{}, err
 			}
 			if !albURI.HTTPS && len(albURI.DNSNames) > 1 {
-				albURI, _ = albDescr.removeEnvDNSName(albURI)
+				albURI = albDescr.bestEffortRemoveEnvDNSName(albURI)
 			}
 			return URI{
 				URI:        english.OxfordWordSeries(albURI.strings(), "or"),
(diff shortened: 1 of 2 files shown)


internal/pkg/describe/uri.go
```diff
@@ -278,6 +281,21 @@ func (d *albDescriber) uri() (albURI, error) {
 	}, nil
 }
 
+func (d *albDescriber) removeEnvDNSName(albURI albURI) (albURI, error) {
+	envOutputs, err := d.envDescriber.Outputs()
+	if err != nil {
+		return albURI, fmt.Errorf("get stack outputs for environment %s: %w", d.env, err)
+	}
+	lbDNSName := envOutputs[d.envDNSNameKey]
```
paragbhingre, line 289: just a suggestion, if we can reuse this same code that is being written in [envDNSName](https://github.com/aws/copilot-cli/blob/8edeba09ac557bedd8ed4df615e630827de99252/internal/pkg/describe/uri.go#L226)
We can send empty path string if we do not have that value available here. And we can then use the returned value of `envDNSName` here directly as `albURI.DNSNames`

paragbhingre (approved):
Just one suggestion otherwise everything else looks good to me :)

internal/pkg/describe/uri.go
dannyrandall: i thought about using that function but i felt wrong putting in an empty path 😅. since this ends up being about the same length anyways, i decided to just duplicate it, but if you feel strongly i'm happy to change it 😊

paragbhingre: hmm you are right that is not a really big function, so I guess I am fine 👍

internal/pkg/describe/uri.go
```diff
@@ -178,6 +178,9 @@ func (d *BackendServiceDescriber) URI(envName string) (URI, error) {
 			if err != nil {
 				return URI{}, err
 			}
+			if !albURI.HTTPS && len(albURI.DNSNames) > 1 {
+				albURI, _ = albDescr.removeEnvDNSName(albURI)
```
iamhopaul123: If we decided to ignore the error maybe we should do it in `removeEnvDNSName` and change its name to `bestEffort...` or add a comment to the function (since I guess it would only be used here)?

dannyrandall: good idea, fixed 🫡

iamhopaul123 (approved):

PR #3731, opened 2022-07-07
feat: add `env package` command

efekarakus:
Related #3522

### DIFF
--- a/cmd/copilot/template/macro.go
+++ b/cmd/copilot/template/macro.go
@@ -40,11 +40,28 @@ func h2(text string) string {
 
 func code(text string) string {
 	lines := strings.Split(text, "\n")
+
+	var startCodeBlockIdx, codeBlockPadding int
 	for i, line := range lines {
 		if strings.HasPrefix(strings.TrimSpace(line), "/code ") {
 			codeIndex := strings.Index(line, "/code ")
 			lines[i] = line[:codeIndex] +
 				termcolor.HighlightCode(strings.ReplaceAll(line[codeIndex:], "/code ", ""))
+		}
+		if strings.HasPrefix(strings.TrimSpace(line), "/startcodeblock") {
+			startCodeBlockIdx = i
+			codeBlockPadding = strings.Index(line, "/startcodeblock")
+		}
+		if strings.HasPrefix(strings.TrimSpace(line), "/endcodeblock") {
+			colored := termcolor.HighlightCodeBlock(strings.Join(lines[startCodeBlockIdx+1:i], "\n"))
+			coloredLines := strings.Split(colored, "\n")
+			for j, val := range coloredLines {
+				padding := ""
+				if j == 0 || j == len(coloredLines)-1 {
+					padding = strings.Repeat(" ", codeBlockPadding)
+				}
+				lines[startCodeBlockIdx+j] = padding + val
+			}
 		}
 	}
 	return strings.Join(lines, "\n")
--- a/internal/pkg/cli/env_package.go
+++ b/internal/pkg/cli/env_package.go
@@ -5,6 +5,14 @@ package cli
 
 import (
 	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/aws/copilot-cli/internal/pkg/cli/deploy"
+	"github.com/spf13/afero"
+
+	"github.com/aws/copilot-cli/internal/pkg/manifest"
 
 	"github.com/aws/copilot-cli/internal/pkg/term/color"
 	"github.com/aws/copilot-cli/internal/pkg/term/log"
@@ -28,13 +36,30 @@ type packageEnvVars struct {
 	uploadAssets bool
 }
 
+type discardFile struct{}
+
+func (df discardFile) Write(p []byte) (n int, err error) {
+	return io.Discard.Write(p)
+}
+
+func (df discardFile) Close() error {
+	return nil //noop
+}
+
 type packageEnvOpts struct {
 	packageEnvVars
 
 	// Dependencies.
-	cfgStore store
-	ws       wsEnvironmentReader
-	sel      wsEnvironmentSelector
+	cfgStore     store
+	ws           wsEnvironmentReader
+	sel          wsEnvironmentSelector
+	caller       identityService
+	fs           afero.Fs
+	tplWriter    io.WriteCloser
+	paramsWriter io.WriteCloser
+
+	newInterpolator func(appName, envName string) interpolator
+	newEnvDeployer  func() (envPackager, error)
 
 	// Cached variables.
 	appCfg *config.Application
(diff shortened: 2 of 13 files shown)


internal/pkg/cli/env_package.go
```diff
@@ -0,0 +1,298 @@
...
+
+	name, err := o.sel.LocalEnvironment("Select an environment manifest from your workspace", "")
+	if err != nil {
+		return fmt.Errorf("select environment: %w", err)
+	}
+	o.envName = name
+	return nil
+}
+
+func (o *packageEnvOpts) environmentManifest() (*manifest.Environment, error) {
```
Lou1415926: Do you think it's worthy uniting this method with the following? https://github.com/aws/copilot-cli/blob/74d034755279b4d28b3c6f4b61c9a34aa2e6513b/internal/pkg/cli/env_deploy.go#L138

internal/pkg/cli/env_package.go
```diff
@@ -0,0 +1,298 @@
...
+
+func (o *packageEnvOpts) setWriters() error {
+	if o.outputDir == "" {
+		return nil
+	}
+	if err := o.fs.MkdirAll(o.outputDir, 0755); err != nil {
+		return fmt.Errorf("create directory %q: %w", o.outputDir, err)
+	}
+
+	path := filepath.Join(o.outputDir, fmt.Sprintf("env-%s.yml", o.envName))
```
Lou1415926: An extreme edge case would be having a workload called "env". Running `svc package --name env --env test` would write the output to `env-test.params.json`  - because it's formatted as `<svc>-<env>.params.json` - which collides with this naming scheme. 

I don't think we need to worry about this too much though but just to call it out!

internal/pkg/cli/env_package.go
```diff
@@ -0,0 +1,298 @@
...
+	packageEnvVars
+
+	// Dependencies.
+	cfgStore     store
+	ws           wsEnvironmentReader
+	sel          wsEnvironmentSelector
+	caller       identityService
+	fs           afero.Fs
+	tplWriter    io.WriteCloser
+	paramsWriter io.WriteCloser
```
Lou1415926, line 59: We can use `WriteCloser` for `svc package`'s writers as well! In a later refactor

Lou1415926 (approved):

internal/pkg/cli/env_package.go
iamhopaul123: To avoid this edge case can we simply change the env artifacts naming pattern to `env_%s.yaml`?

internal/pkg/cli/env_package.go
```diff
@@ -0,0 +1,298 @@
...
+
+// buildEnvPkgCmd builds the command for printing an environment CloudFormation stack configuration.
+func buildEnvPkgCmd() *cobra.Command {
+	vars := packageEnvVars{}
+	cmd := &cobra.Command{
+		Use:   "package",
+		Short: "Prints the AWS CloudFormation template of an environment.",
+		Long:  `Prints the CloudFormation stack template and configuration used to deploy an environment.`,
+		Example: `
+  Print the CloudFormation template for the "prod" environment.
```
iamhopaul123: nit: should we be consistent about "print" or "prints"? Maybe
```suggestion
		Short: "Print the AWS CloudFormation template of an environment.",
		Long:  `Print the CloudFormation stack template and configuration used to deploy an environment.`,
		Example: `
  Print the CloudFormation template for the "prod" environment.
```

internal/pkg/cli/interfaces.go
```diff
@@ -658,3 +658,8 @@ type envDeployer interface {
 	DeployEnvironment(in *clideploy.DeployEnvironmentInput) error
 	UploadArtifacts() (map[string]string, error)
 }
+
+type envPackager interface {
+	GenerateCloudFormationTemplate(in *clideploy.DeployEnvironmentInput) (*clideploy.GenerateCloudFormationTemplateOutput, error)
```
iamhopaul123, line 667: How about
```suggestion
type packager interface {
	GenerateCloudFormationTemplate(in *clideploy.GenerateCloudFormationTemplateInput) (*clideploy.GenerateCloudFormationTemplateOutput, error)
```
💭

cmd/copilot/template/macro.go
```diff
@@ -40,12 +40,29 @@ func h2(text string) string {
...
 	lines := strings.Split(text, "\n")
+
+	var startCodeBlockIdx, codeBlockPadding int
 	for i, line := range lines {
 		if strings.HasPrefix(strings.TrimSpace(line), "/code ") {
 			codeIndex := strings.Index(line, "/code ")
 			lines[i] = line[:codeIndex] +
 				termcolor.HighlightCode(strings.ReplaceAll(line[codeIndex:], "/code ", ""))
 		}
+		if strings.HasPrefix(strings.TrimSpace(line), "/startcodeblock") {
```
iamhopaul123, line 51: This is awesome!

internal/pkg/cli/env_package.go
Lou1415926: then it may have 
```
/infrastructure
   /svc-env.yaml
   /env_test.yaml
```
The inconsistency on the usage of `-` and `_` is a bit bothering 😓

iamhopaul123: Right right but it seems like we need to choose from two poisons

efekarakus: what do you think of 
`./infrastructure/environments/env-test.yaml`

we can then generate the pipeline manifest with this path in mind

iamhopaul123: yea but then the output dir isn't really the one they specify, which is not consistent with the workload behavior. And could end up being `./infra/env/environments/env-test.yaml` if they specify the output dir to be `./infra/env/`. I would prefer to leave as is...

internal/pkg/cli/env_package.go
efekarakus: Done!

internal/pkg/cli/env_package.go
efekarakus: Renamed the file format to `%s.env.yml` and `%s.env.params.json`

internal/pkg/cli/env_package.go
efekarakus: Done! updated in this pr

internal/pkg/cli/env_package.go
efekarakus: Done!

internal/pkg/cli/interfaces.go
efekarakus: I couldn't really do that because the template inputs for environments and workloads are pretty different :rip:

Lou1415926 (approved):

iamhopaul123 (approved):

PR #3732, opened 2022-07-07
refactor(deploy): swap the implementation to upload custom resources for env stack

Lou1415926:
This wasn't done previously  in #3684 because `env init` was using the old implementation and swapping to this new implementation without updating `env init` would break that command.

Now that `env init` is only deploying bootstrap stack after #3730, it's safe to swap the implementation now.
 


iamhopaul123 (approved):

efekarakus (approved):
Awesome, ty!

PR #3737, opened 2022-07-08
chore: allow cloud front prefix list to restrict security group access via environment manifest

CaptainCarpensir:
Allow users to specify prefix lists, and to restrict ingress traffic to the ALB via the environment manifest. Adds a new field to the PublicHTTPConfig field `restrict_alb_ingress_to_cf` for this behavior.

Feature for part of #3701 


internal/pkg/manifest/env.go
```diff
@@ -320,13 +322,14 @@ func (cfg *environmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
 	cfg.Public.Certificates = env.ImportCertARNs
 }
 
-type publicHTTPConfig struct {
-	Certificates []string `yaml:"certificates,omitempty"`
+type PublicHTTPConfig struct {
+	LimitToCFIngress *bool    `yaml:"restrict_alb_ingress_to_cf"`
```
iamhopaul123: Can we make this more extensible maybe something like this?
```yaml
http:
  public:
    security_groups:
      ingress:
        from_cdn: true
```

internal/pkg/manifest/env.go
```diff
@@ -320,13 +322,14 @@ func (cfg *environmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
 	cfg.Public.Certificates = env.ImportCertARNs
 }
 
-type publicHTTPConfig struct {
-	Certificates []string `yaml:"certificates,omitempty"`
+type PublicHTTPConfig struct {
```
iamhopaul123, line 340: Comment pls

internal/pkg/cli/deploy/env.go
```diff
@@ -105,6 +113,37 @@ func (d *envDeployer) uploadCustomResources(bucket string) (map[string]string, e
...
+		prefixListIDs = append(prefixListIDs, *cfManagedPrefixListId)
+	}
+
+	return prefixListIDs, nil
+}
+
+func (d *envDeployer) cfManagedPrefixListId(in *DeployEnvironmentInput) (*string, error) {
+	// Check if ingress is allowed from cloudfront
+	if in.Manifest != nil {
+		if !aws.BoolValue(in.Manifest.HTTPConfig.Public.LimitToCFIngress) {
```
```suggestion
	if in.Manifest == nil || !aws.BoolValue(in.Manifest.HTTPConfig.Public.LimitToCFIngress) {
	  return nil, nil
	}
```

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
```diff
@@ -34,6 +34,7 @@ type: Environment
 # All these comments should be deleted.
 http:
   public:
+    restrict_alb_ingress_to_cf: false
```
iamhopaul123: Can we make it true so that it can be covered in the integ test?

internal/pkg/manifest/validate_env.go
```diff
@@ -37,6 +37,10 @@ func (e environmentConfig) Validate() error {
 		return fmt.Errorf(`validate "http config": %w`, err)
 	}
 
+	if aws.BoolValue(e.HTTPConfig.Public.LimitToCFIngress) && !e.CDNConfig.CDNEnabled() {
+		return errors.New("CDN must be enabled to limit security group ingress to CloudFront")
```
iamhopaul123, line 41: Can we use https://github.com/aws/copilot-cli/blob/45bdcc3549aa7a85ec3f08170ee44e5006ebe323/internal/pkg/manifest/errors.go#L57

CaptainCarpensir: This doesn't quite cover the error case. The error is for when `cdn` is either not specified _or_ disabled and the `from_cdn` is enabled.  The `errFieldMustBeSpecified` doesn't care about the values of the field, only that they're dependent on each other.

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
CaptainCarpensir: How do we do this? It will need to query for a prefix list to test against, and I don't know how to not write that explicity into the `template-with-imported-certs-observability.yml` file

internal/pkg/manifest/validate_env.go
iamhopaul123: Sorry i was about to say https://github.com/aws/copilot-cli/blob/45bdcc3549aa7a85ec3f08170ee44e5006ebe323/internal/pkg/manifest/errors.go#L75

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
iamhopaul123: right right it is controlled by https://github.com/aws/copilot-cli/pull/3737/files#diff-031da705d16a975e87d8537638fac3cc2e4dfbc8411ff9b184cb7b28666914b9R101 instead. Can we specify `PrefixListIDs` [here](https://github.com/aws/copilot-cli/blob/45bdcc3549aa7a85ec3f08170ee44e5006ebe323/internal/pkg/deploy/cloudformation/stack/env_integration_test.go#L45)?

internal/pkg/manifest/validate_env.go
CaptainCarpensir: That doesn't work either for the same reasons 🤔 
I don't think `manifest/errors.go` has any existing error type that covers this specific validation error

iamhopaul123: Gotcha

iamhopaul123 (approved):

internal/pkg/cli/deploy/env.go
```diff
@@ -33,6 +35,10 @@ type environmentDeployer interface {
 	UpdateAndRenderEnvironment(out termprogress.FileWriter, env *deploy.CreateEnvironmentInput, opts ...cloudformation.StackOption) error
 }
 
+type prefixListGetter interface {
+	CloudFrontManagedPrefixListID() (*string, error)
```
efekarakus: should the API be ?
```suggestion
	CloudFrontManagedPrefixListID() (string, error)
```
And return an `ErrPrefixListIDNotFound` in case it's not there instead of of `nil` for a string.

internal/pkg/cli/deploy/env.go
```diff
@@ -105,6 +113,33 @@ func (d *envDeployer) uploadCustomResources(bucket string) (map[string]string, e
 	return urls, nil
 }
 
+func (d *envDeployer) prefixLists(in *DeployEnvironmentInput) ([]string, error) {
```
efekarakus: Was there a reason why you preferred a `[]string` instead of just `string, error`?

internal/pkg/cli/deploy/env.go
```diff
@@ -133,6 +172,7 @@ func (d *envDeployer) DeployEnvironment(in *DeployEnvironmentInput) error {
 		CustomResourcesURLs:  in.CustomResourcesURLs,
 		ArtifactBucketARN:    s3.FormatARN(partition.ID(), resources.S3Bucket),
 		ArtifactBucketKeyARN: resources.KMSKeyARN,
+		PrefixListIDs:        prefixListIDs,
```
efekarakus: Reading the field name it's not clear to me what the prefixes are _for_? What do you think of renaming to `EC2PrefixListIDs` or `CIDRPrefixListIDs` (same for the field name in `template`)

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
```diff
@@ -32,8 +32,11 @@ name: test
 type: Environment
 # Create the public ALB with certificates attached.
 # All these comments should be deleted.
+cdn: true
 http:
   public:
+    ingress:
```
efekarakus: Would you mind syncing with @Lou1415926 to make sure the security groups experience is consistent with her changes: https://github.com/aws/copilot-cli/pull/3763/files#diff-0fc2e38bf10b33570f95dbb18d51eb694ff20caa1e87860c18c32d800a6b7266?

Just reading that scheme it sounds like a more consistent format for the two changes is (feel free to change it):
```yaml
http:
  public:
     security_groups:
        restrict_from_cdn: true
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -96,24 +96,60 @@ Resources:
           Value: disabled
           {{- end}}
 {{- end}}
-  PublicLoadBalancerSecurityGroup:
+  PublicHTTPLoadBalancerSecurityGroup:
```
efekarakus, line 100: hmmm is this a safe change? 🤔 

Does renaming the logic id of a resource break environment updates? For example if I have already deployed an environment with `PublicLoadBalancerSecurityGroup` and now try `env deploy` with this new template does that succeed?

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -216,7 +260,7 @@ Resources:
     Type: AWS::ElasticLoadBalancingV2::LoadBalancer
     Properties:
       Scheme: internet-facing
-      SecurityGroups: [ !GetAtt PublicLoadBalancerSecurityGroup.GroupId ]
+      SecurityGroups: [ !GetAtt PublicHTTPLoadBalancerSecurityGroup.GroupId,  !GetAtt PublicHTTPSLoadBalancerSecurityGroup.GroupId]
```
efekarakus: We are suddenly increasing the surface or requests that a load balancer accepts (allowing port 443 requests), is it possible to include this sec group only if HTTPS is required?

internal/pkg/cli/deploy/env.go
CaptainCarpensir: I had this return a `[]string` so that it can handle the logic of adding any more prefix lists that we may want use for in the future. That's then why there's a separation b/w the functions `prefixList` and `cfManagedPrefixListId`

internal/pkg/template/templates/environment/cf.yml
CaptainCarpensir: After testing, yeah this does succeed at least in a basic circumstance.

internal/pkg/cli/deploy/env.go
CaptainCarpensir: The reason I don't think I can do this, at least with how I've currently arranged it, is that the function `cfManagedPrefixListId` should return nil when the ingress isn't enabled for cloudfront, as that function is what handles checking if it's enabled. If I make `cfManagedPrefixListId` return a `string`, then it can't handle the situation where it's not enabled, and as such shouldn't return any id, but also shouldn't error out the creation of the `CreateEnvironmentInput` struct. This change could be possible then if I moved the checks to the function `prefixLists` but I feel that isn't a good solution as over time it causes bloat to the function.

internal/pkg/template/templates/environment/cf.yml
iamhopaul123: It is not increasing since we have 443 open now no matter if it is https or not, but I do agree we should improve this behavior and make it optional :+1:

internal/pkg/cli/deploy/env.go
CaptainCarpensir: As a follow up TLDR, to do this I'll need to move the validation to check if the CloudFront prefix list is needed to the function `prefixLists`. I personally prefer the check to be in `cfManagedPrefixListId` because it removes potential bloat from the more general function as more potential prefix lists can be used, but I'm happy to move that check over if you think it'd be better overall.

iamhopaul123 (approved):

CaptainCarpensir:
I realize that I changed a field name that we just released to make it more clear in the most recent commit to `allow_from_vpc` rather than just `from_vpc`. Unfortuantely I don't think we can push this since it's a breaking change, but I also regret not recognizing that I don't think the name `from_vpc` is clear enough, especially under the context of adding it to a struct which has both restricting and allowing fields.

internal/pkg/cli/deploy/env.go
rclinard-amzn: Looking at the current implementation of CloudFrontManagedPrefixList in ec2.go, I wasn't able to see a case where it returned `nil` for the string without there being an error condition, and it does in fact return an error for when the prefix list ID isn't there - though not as a separate error type. `cfManagedPrefixListId` could always make it a `*string` if needed separate from this function?

https://github.com/aws/copilot-cli/blob/a57ecbd9203bf3e51eee195e39d60a85e9ddb86f/internal/pkg/aws/ec2/ec2.go#L462-L483

rclinard-amzn:
"ingress from vpc" reads naturally to me and does imply the "allow" part, so it might be a fine name by itself. How about instead moving away from the "restrict" terminology? `only_from_cdn` might be a more clear name and could avoid the ambiguity, and would read similarly naturally as well - "(allow) ingress only from cdn".

internal/pkg/cli/deploy/env.go
bvtujo: I think we should write code for the use case that's already here, instead of what may happen in the future. In many cases a small refactor (changing the method signature from `(string, error)` to `([]string, error)` is better than pushing code that, while more general, isn't used to its full extent. Is there another prefixList you can think of that we might imminently need?

rclinard-amzn:
Per discussion in-person between myself, Austin, and Aiden - we think that modeling this as the following might work better:

```
http:
  public:
    security_group:
      ingress:
        restrict_to: [cdn]
```

This would accommodate adding additional prefix lists (or exposing prefix lists to the user) in a way that wouldn't generate ambiguity as would happen if we ended up with two fields along the lines of `only_from_...`, and without changing the current default behavior of the public `ingress` field of allowing all inbound Internet traffic by default.

internal/pkg/cli/deploy/env.go
CaptainCarpensir: This isn't a small refactor at all, as this goes all the way up the chain starting from the template to where the prefix list is retrieved in `cli/deploy/env.go`. It also affects decisions I've made with how the manifest is structured, so is completely integral to this entire PR.

internal/pkg/cli/deploy/env.go
```diff
@@ -34,13 +36,18 @@ type environmentDeployer interface {
...
+}
+
 type envDeployer struct {
 	app *config.Application
 	env *config.Environment
 
 	// Dependencies to upload artifacts.
 	templateFS template.Reader
 	s3         uploader
+	ec2        prefixListGetter
```
Lou1415926: what do you think of renaming the field to `prefixListGetter` instead of `ec2` just to better describe what it does? (`s3` isn't a great example I need to refactor it to a better name 😅 )

internal/pkg/cli/deploy/env.go
```diff
@@ -109,6 +117,30 @@ func (d *envDeployer) uploadCustomResources(bucket string) (map[string]string, e
...
+		if err != nil {
+			return nil, err
+		}
+		prefixListIDs = append(prefixListIDs, cfManagedPrefixListId)
+	}
+
+	return prefixListIDs, nil
+}
+
+func (d *envDeployer) cfManagedPrefixListId(in *DeployEnvironmentInput) (string, error) {
```
Lou1415926: What is `in` passed in for 👀 ?

internal/pkg/cli/deploy/env_test.go
```diff
@@ -257,6 +261,42 @@ func TestEnvDeployer_DeployEnvironment(t *testing.T) {
...
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			wantedError: fmt.Errorf("retrieve CloudFront managed prefix list id: some error"),
+		},
+		"prefix list not retrieved when manifest not present": {
```
Lou1415926, line 288: I think this scenario wouldn't happen at all: environment manifest is required for both  `env deploy` and `env package`. We can keep this test case - I just wanted to add so that other reviewers know!

internal/pkg/cli/deploy/env.go
```diff
@@ -109,6 +117,30 @@ func (d *envDeployer) uploadCustomResources(bucket string) (map[string]string, e
 	return urls, nil
 }
 
+func (d *envDeployer) prefixLists(in *DeployEnvironmentInput) ([]string, error) {
+	var prefixListIDs []string
+
+	// Check if ingress is allowed from cloudfront
+	if in.Manifest != nil && aws.BoolValue(in.Manifest.HTTPConfig.Public.SecurityGroupConfig.Ingress.RestrictiveIngress.CDNIngress) {
```
Lou1415926: We can flip the conditional and return early, such as:
```
if aws.BoolValue(in.Manifest.HTTPConfig.Public.SecurityGroupConfig.Ingress.RestrictiveIngress.CDNIngress) == false {
  return "", nil
}
cfManagedPrefixListID, err := ...
```

internal/pkg/cli/deploy/env.go
```diff
@@ -109,6 +117,30 @@ func (d *envDeployer) uploadCustomResources(bucket string) (map[string]string, e
...
+		cfManagedPrefixListId, err := d.cfManagedPrefixListId(in)
```
Lou1415926: nit:
```suggestion
		cfManagedPrefixListID, err := d.cfManagedPrefixListId(in)
```

internal/pkg/deploy/env.go
```diff
@@ -38,6 +38,7 @@ type CreateEnvironmentInput struct {
 	ImportCertARNs     []string              // Optional configuration if users want to import certificates.
 	InternalALBSubnets []string              // Optional configuration if users want to specify internal ALB placement.
 	AllowVPCIngress    bool                  // Optional configuration to allow access to internal ALB from ports 80/443.
+	CIDRPrefixListIDs  []string              // Optional configuration to specify public security group ingress based on prefix lists
```
Lou1415926: nit:
```suggestion
	CIDRPrefixListIDs  []string              // Optional configuration to specify public security group ingress based on prefix lists.
```

internal/pkg/manifest/validate_env.go
```diff
@@ -206,14 +210,22 @@ func (cfg environmentHTTPConfig) Validate() error {
...
-func (cfg publicHTTPConfig) Validate() error {
+// Validate returns nil if PublicHTTPConfig is configured correctly.
+func (cfg PublicHTTPConfig) Validate() error {
 	for idx, certARN := range cfg.Certificates {
 		if _, err := arn.Parse(certARN); err != nil {
 			return fmt.Errorf(`parse "certificates[%d]": %w`, idx, err)
 		}
 	}
-	return nil
+	if cfg.SecurityGroupConfig.Ingress.VPCIngress != nil {
```
Lou1415926, line 220: yayy nice thank you!!  I'm also planning to refactor it so that the underlying data struct for `http.public.security_groups.ingress` is not the same as `http.private.security_groups.ingress`. We can keep this validation for now until the refactor happens.

internal/pkg/manifest/validate_env.go
```diff
@@ -223,9 +235,17 @@ func (cfg privateHTTPConfig) Validate() error {
 			return fmt.Errorf(`parse "certificates[%d]": %w`, idx, err)
 		}
 	}
+	if !cfg.SecurityGroupsConfig.Ingress.RestrictiveIngress.IsEmpty() {
```
Lou1415926, line 238: Same as above! No action item for ya.

internal/pkg/template/templates/environment/cf.yml
Lou1415926: My concern is that it may delete the old security group and create a new one. If we run `copilot env deploy`  on an environment that still has `PublicLoadBalancerSecurityGroup` and not `PublicHTTPLoadBalancerSecurityGroup` yet (i.e. before your PR change), we'd probably see the following activities in `Event` tab:
```
PublicLoadBalancerSecurityGroup DELETE_COMPLETE
PublicLoadBalancerSecurityGroup DELETE_IN_PROGRESS
...
<stack_name> UPDATE_COMPLETE_CLEANUP_IN_PROGRESS
...
PublicHTTPLoadBalancerSecurityGroup CREATE_COMPLETE
PublicHTTPLoadBalancerSecurityGroup CREATE_IN_PROGRESS
...
<stack_name> UPDATE_IN_PROGRESS
```
Although it's probably not a problem to recreate the security group 🤔  but we should be careful about it if it introduces unnecessary resource updates just because the logical ID changes.

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -97,24 +97,60 @@ Resources:
...
+        {{- end}}
+{{- if .VPCConfig.Imported}}
+      VpcId: {{.VPCConfig.Imported.ID}}
+{{- else}}
+      VpcId: !Ref VPC
+{{- end}}
+      Tags:
+        - Key: Name
+          Value: !Sub 'copilot-${AppName}-${EnvironmentName}-lb'
+  PublicHTTPSLoadBalancerSecurityGroup:
```
Lou1415926, line 131: qq: I see we are trying to add this ingress only if the ALB is listening for HTTPS on 443. This makes sense. Any other reasons why we are doing this now?

internal/pkg/cli/deploy/env.go
efekarakus: The big chain `in.Manifest.HTTPConfig.Public.SecurityGroupConfig.Ingress.RestrictiveIngress.CDNIngress` reminds me of the Law of Demeter to me (https://en.wikipedia.org/wiki/Law_of_Demeter#In_object-oriented_programming). It doesn't quite apply because it's not a chain of methods, but regardless coupling `deploy` to the internal structure of the `manifest` would make future refactoring a bit difficult.  
To be honest, this is not the first place where we violate the rule the `stack/transformers.go` is another good example. 

Instead, what do you think of a code structure like this:
```go
func (mft *Environment) IsIngressRestrictedToCDN() bool
```

internal/pkg/cli/deploy/env.go
efekarakus: do we need to take `in` as input?

internal/pkg/cli/deploy/env.go
```diff
@@ -109,6 +117,30 @@ func (d *envDeployer) uploadCustomResources(bucket string) (map[string]string, e
...
+		prefixListIDs = append(prefixListIDs, cfManagedPrefixListId)
+	}
+
+	return prefixListIDs, nil
+}
+
+func (d *envDeployer) cfManagedPrefixListId(in *DeployEnvironmentInput) (string, error) {
+	id, err := d.ec2.CloudFrontManagedPrefixListID()
+	if err != nil {
+		return "", fmt.Errorf("retrieve CloudFront managed prefix list id: %s", err)
```
```suggestion
		return "", fmt.Errorf("retrieve CloudFront managed prefix list id: %w", err)
```

internal/pkg/cli/deploy/env.go
```diff
@@ -188,6 +220,10 @@ func (d *envDeployer) buildStackInput(in *DeployEnvironmentInput) (*deploy.Creat
 	if err != nil {
 		return nil, err
 	}
+	cidrPrefixListIDs, err := d.prefixLists(in)
```
efekarakus: nit: I wonder if we should rename `prefixLists` to `cidrPrefixLists` ?

internal/pkg/template/env.go
```diff
@@ -105,12 +105,13 @@ type EnvOpts struct {
...
-	PrivateImportedCertARNs  []string
-	CustomInternalALBSubnets []string
-	AllowVPCIngress          bool
-	Telemetry                *Telemetry
+	VPCConfig                     VPCConfig
+	PublicFacingCIDRPrefixListIDs []string
+	PublicImportedCertARNs        []string
+	PrivateImportedCertARNs       []string
+	CustomInternalALBSubnets      []string
+	AllowVPCIngress               bool
```
efekarakus, line 113: I feel like we're in dire need of creating a type to group related resources together instead of inline individual fields similar to what we did with `VPCConfig`.

```go
   SG          SecurityGroupConfig 
   CertARNs    []string
}

type PrivateHTTPConfig struct {
   SG          SecurityGroupConfig
   CertARNs    []string
}

type SecurityGroupConfig struct {
   PrefixListIDs    []string
   AllowAllIngress  bool
}

type env struct {
   ...
   PublicHTTP  PublicHTTPConfig
   PrivateHTTP PrivateHTTConfig
}
```

efekarakus: If you want we can merge it as is, but we should send a follow-up refactor PR after this one

internal/pkg/template/templates/environment/cf.yml
CaptainCarpensir: Splitting the security group is needed because one security group has a max count of prefix lists of 60, but each cloud front prefix list counts for 55 of those, so in order to still only allow http and https traffic, we need to split them into two different SGs.

iamhopaul123: To add on @CaptainCarpensir. One other benefit is that we don't always open HTTPS 443 for users anymore if they don't enable https.

internal/pkg/cli/deploy/env.go
CaptainCarpensir: relic of a past change, thanks for spotting that!

internal/pkg/template/templates/environment/cf.yml
efekarakus: Similarly, I'm also still worried because I don't have a clear understanding of the sequence of steps that occur.

> After testing, yeah this does succeed at least in a basic circumstance.

@CaptainCarpensir what did you observe on an env existing stack with a LBWS already enabled?
Does it:
1. Create the new `PublicHTTPLoadBalancerSecurityGroup` 
2. Update `PublicLoadBalancer`
3. Delete `PublicLoadBalancerSecurityGroup`

Is that the sequence? If it's the other way around (what Wanxian wrote) then the users will experience downtime because the load balancer for a brief period won't accept any incoming traffic.

internal/pkg/manifest/env.go
```diff
@@ -328,13 +330,46 @@ func (cfg *environmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
 	cfg.Public.Certificates = env.ImportCertARNs
 }
 
-type publicHTTPConfig struct {
-	Certificates []string `yaml:"certificates,omitempty"`
+// ALBSecurityGroupsConfig represents security group configuration settings for an ALB.
+type ALBSecurityGroupsConfig struct {
```
efekarakus, line 346: nit: can we move this type after `PublicHTTPConfig`

CaptainCarpensir: Did both ways testing, going from `PublicLoadBalancerSecurityGroup` to `PublicHTTPLoadBalancerSecurityGroup` and back the other way successfully creates the resource, updates the load balancer, and then deletes the unneeded resource in order.

internal/pkg/template/env.go
CaptainCarpensir: I agree, I think I'd rather do this merge as is, and then quickly make a refactor PR for this.

internal/pkg/cli/deploy/env.go
```diff
@@ -109,6 +116,31 @@ func (d *envDeployer) uploadCustomResources(bucket string) (map[string]string, e
...
+	if in.Manifest == nil || !in.Manifest.IsIngressRestrictedToCDN() {
+		return nil, nil
+	}
+	cfManagedPrefixListID, err := d.cfManagedPrefixListID()
+	if err != nil {
+		return nil, err
+	}
+	cidrPrefixListIDs = append(cidrPrefixListIDs, cfManagedPrefixListID)
+
+	return cidrPrefixListIDs, nil
```
efekarakus, line 132: nit:
```suggestion

	// Check if ingress is allowed from cloudfront
	if in.Manifest == nil || !in.Manifest.IsIngressRestrictedToCDN() {
		return nil, nil
	}
	id, err := d.cfManagedPrefixListID()
	if err != nil {
		return nil, err
	}
	return []string{id}, nil
```

efekarakus (approved):

PR #3740, opened 2022-07-08
fix: Resolve a number of issues preventing unit tests from passing on Windows

rclinard-amzn:
This PR fixes the unit test suite on Windows. There are two primary classes of errors fixed:

## Assumptions about the path separator being `/`

 Test cases always assuming `/` path separators in messages, rather than expecting `\` on Windows where needed. Fixed by adding calls to `filepath.FromSlashes` where necessary.

## Problematic calls to `filepath.Rel`

There were various issues caused by test cases using bogus paths in mocks, that then lead to calls to functions such as `filepath.Rel`. On Windows, unlike Unix-based OSes, two arbitrary paths might not be able to be made relative to each other, since there are multiple potential roots to the filesystem (`C:\`, `D:\`, etc) rather than there being one (`/`).

This was addressed by replacing a number of helper functions along the lines of `relPath` / etc with two centrally-defined functions on Workspace, for the two use cases in the codebase for creating relative paths:

- `RelCwd`: Returns a path relative to the current working directory, for displaying to the user.
- `RelWsRoot`: Returns a path relative to the current workspace root, for storing a stable form of a path in configuration files and for similar purposes.

Then, since Workspace is almost always mocked, tests could handle the bogus paths in mocked calls rather than the bogus paths leaking through to OS-specific filesystem calls.

---


internal/pkg/cli/job_init_test.go
```diff
@@ -65,7 +66,7 @@ func TestJobInitOpts_Validate(t *testing.T) {
 			setupMocks: func(m initJobMocks) {
 				m.mockStore.EXPECT().GetApplication("phonetool").Return(&config.Application{}, nil)
 			},
-			wantedErr: errors.New("open hello/Dockerfile: file does not exist"),
+			wantedErr: errors.New("open " + filepath.FromSlash("hello/Dockerfile") + ": file does not exist"),
```
huanjani: nit (here and elsewhere):
```suggestion
			wantedErr: fmt.Errorf("open %s: file does not exist", filepath.FromSlash("hello/Dockerfile")),
```

rclinard-amzn: Should be fixed!

huanjani (approved):
This looks great! Thank you so much for taking this on and improving our codebase.

internal/pkg/workspace/workspace.go
```diff
@@ -82,6 +82,13 @@ func New() (*Workspace, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Make sure the working directory is an absolute path, most of the functions here assume that it is (and that
+	// the derived copilotDir is absolute as well)
+	workingDir, err = filepath.Abs(workingDir)
```
efekarakus: Can we replace the comment by renaming the field name? Also reading [os.Getwd](https://pkg.go.dev/os#Getwd) it sounds like the stdlib returns the abs path already, so we can avoid this extra function call 🥳 

```
wdAbsPath: wd,
```

internal/pkg/workspace/workspace.go
```diff
@@ -515,26 +522,62 @@ func (ws *Workspace) Path() (string, error) {
...
 
-// Rel returns the path relative to the workspace root.
-func (ws *Workspace) Rel(fullPath string) (string, error) {
+// RelWsRoot returns the path relative to the workspace root.
+// The input path is allowed to be either relative to the current
+// working directory or absolute.
+//
+// This is useful for storing a file path in configuration.
+//
+// Prefer RelCwd for displaying paths to users.
```
efekarakus: nit: can we remove the comments indicating _how_ a client should use the method? I try to stick to _what_ in the doc comments, and _why_ in the implementation comments.
```suggestion
```

internal/pkg/workspace/workspace.go
```diff
@@ -515,26 +522,62 @@ func (ws *Workspace) Path() (string, error) {
...
+func (ws *Workspace) RelWsRoot(path string) (string, error) {
```
efekarakus: Should we keep the old method signature? reading `ws.Rel(somePath)` seems fairly reasonable I think to me to hint at a relative path from the workspace.

internal/pkg/workspace/workspace.go
```diff
@@ -515,26 +522,62 @@ func (ws *Workspace) Path() (string, error) {
...
+// The input path is allowed to be either relative to the current
+// working directory or absolute.
+//
+// This is useful when displaying a file path to the user.
+//
+// You probably want to use RelWsRoot for most other cases.
+// In the common case (the user running commands from the workspace root)
+// these functions have the same behavior, which can be a dangerous
+// source of false assumptions because that is not guaranteed!
+func (ws *Workspace) RelCwd(fullPath string) (string, error) {
```
efekarakus: At a glance, I can't tell why this needs to belong to the `Workspace` object 💭 I understand `Rel` because it's relative to the workspace root but this could be also written as a function in the cli pkg:
```go
package cli

func relCwd(path string) (string, error) {
   if !filepath.IsAbs(path) { return filepath.Clean(path), nil }
   wd, err := os.Getwd()
   if err != nil {
      return "", err
   }
   return filepath.Rel(wd, path), nil
}
```

to make the above easy to unit test in commands:
```go
// in cmd.go
package cli 

func relCwd(getWd() (string, error), path string) (string, error) { ... }
type Cmd struct {
   getWd func() (string, error)
}
func NewCmd() *Cmd {
   return &Cmd{ getWd: os.Getwd }
}

// in cmd_test.go
testCmd := &Cmd{
  getWd: func() { return "root", nil }
}
```

internal/pkg/workspace/workspace.go
```diff
@@ -515,26 +522,62 @@ func (ws *Workspace) Path() (string, error) {
...
+
+// copilotDirPath tries to find the current app's copilot directory from the workspace working directory.
 func (ws *Workspace) copilotDirPath() (string, error) {
 	if ws.copilotDir != "" {
 		return ws.copilotDir, nil
 	}
-	// Are we in the application directory?
+	// Are we in the application's copilot directory already?
 	inCopilotDir := filepath.Base(ws.workingDir) == CopilotDirName
 	if inCopilotDir {
```
efekarakus:
```suggestion
	// Are we inside a "copilot" directory?
	if inCopilotDir := := filepath.Base(ws.workingDir) == CopilotDirName; inCopilotDir {
```

internal/pkg/cli/task_run_test.go
```diff
@@ -232,14 +232,14 @@ func TestTaskRunOpts_Validate(t *testing.T) {
 			inDockerfilePath: "world/hello/Dockerfile",
 			isDockerfileSet:  true,
 
-			wantedError: errors.New("invalid `--dockerfile` path: open world/hello/Dockerfile: file does not exist"),
+			wantedError: fmt.Errorf("invalid `--dockerfile` path: open %s: file does not exist", filepath.FromSlash("world/hello/Dockerfile")),
```
efekarakus, line 235: ty for fixing all these tests 😍

internal/pkg/workspace/workspace.go
rclinard-amzn: I'm not sure about this change. Saying "a 'copilot' directory" sorta loses the intent conveyed in the comment, the intent that this code is meant to determine whether we are in *the* single copilot directory for the application. While it is true that what the code *does* is that it detects if we are in a directory named copilot, in the case where we're in a directory that the user happened to name "copilot" (but is not the workspace's copilot folder) this code will incorrect, and someone who wants to figure out how to fix it would be guided by the *intent* in the comment.

How about something like this?

```suggestion
	// Are we in the application's copilot directory already?
	//
	// Note: This code checks for *any* directory named "copilot", but this might not work
	// correctly if we're in some subdirectory of the app that the user might have happened
	// to name "copilot". It's not clear if there's a good way to avoid that problem, though it
	// doesn't seem to be a terribly likely issue.
	if filepath.Base(ws.workingDir) == CopilotDirName {
```

efekarakus: Sounds good, thanks Ryan!

internal/pkg/workspace/workspace.go
rclinard-amzn: The proposed field name might introduce an ambiguity between whether it's the "working directory" (wd) or "workspace directory" (wd). Maybe `workingDirAbs`?

rclinard-amzn:
I think all those comments should be resolved!

I decided to create a `pathDisplayer` interface instead of the proposed approach, since the only reason these commands needed to care about the working directory and Rel and such is so that they can display a path to the user. Having a dedicated interface to do that keeps things easily testable & mockable without having to unnecessarily tie the functionality to Workspace like you stated.

Thanks for the feedback!

internal/pkg/cli/cli.go
```diff
@@ -139,3 +126,26 @@ func indentListItem(multiline string) string {
 	}
 	return strings.Join(prefixedLines, "\n")
 }
+
+// CwdPathDisplayer displays paths to the user relative to their current working directory.
+type CwdPathDisplayer struct {
+	workingDir string
+}
+
+func NewCwdPathDisplayer() (*CwdPathDisplayer, error) {
```
efekarakus: I wonder if we can potentially get rid of the additional `DisplayPath` interface if we just store the `workingDir` string in the commands instead?
For example:
```go
type initEnvOpts {
   // other fields...
   cwdPath string // Absolute path to the current working dir.
}

func newInitEnvOpts(vars initEnvVars) (*initEnvOpts, error) {
   // initialize other dependencies...
   wd, err := os.Getwd()
   if err != nil {
      return nil, err
   }
   return &initEnvOpts{
       // assign other fields...
       cwdPath: wd,
   }
}

func (o *initEnvOpts) writeManifest() {
   p, err := relPath(o.cwdPath, mftPath) // Call a function instead.
}
```

I think it achieves the same benefits (ease of testing since unit tests can set a value for `cwdPath`) while getting rid of an additional type `CwdPathDisplayer` and interface 💭

internal/pkg/cli/env_init.go
```diff
@@ -816,6 +823,7 @@ func (o *initEnvOpts) tryDeletingEnvRoles(app, env string) {
 	}
 }
 
+// writeManifest
```
efekarakus: nit: can we remove the comment :)
```suggestion
```

internal/pkg/cli/env_init.go
```diff
@@ -270,11 +277,11 @@ func (o *initEnvOpts) Execute() error {
 	}
 
 	// 1. Write environment manifest.
-	mftPath, err := o.writeManifest()
+	mftDisplayedPath, err := o.writeManifest()
```
efekarakus: nit: in Go (compared to Java 😜) folks tend to gravitate towards shorter var names https://github.com/golang/go/wiki/CodeReviewComments#variable-names 
Personally, my fav article is from this ex-sr. principal at AWS: https://a-nickels-worth.blogspot.com/2016/04/a-guide-to-naming-variables.html

in this scenario, the field name is nice and descriptive so this variable can be just:
```go
	path, err := o.writeManifest()
	// if err != nil { ... }
	o.mftDisplayedPath = path
```

internal/pkg/cli/interfaces.go
```diff
@@ -249,10 +249,36 @@ type wsPipelineManifestReader interface {
...
+	// No guarantees are given on the stability of the path across runs, all that is
+	// guaranteed is that the displayed path is visually pleasing & meaningful for a
+	// user.
+	//
+	// This path should not be stored in configuration files or used in any way except
+	// for being displayed to the user.
+	DisplayPath(path string) (string, error)
+}
+
+type wsPathRelativizer interface {
```
efekarakus: nit: Since the `Rel(path) (string, error)` doesn't have a `workspace` specific type in it we can make the interface name a bit more generic
```suggestion
type relPath interface {
```

internal/pkg/cli/interfaces.go
```diff
@@ -249,10 +249,36 @@ type wsPipelineManifestReader interface {
...
+	// Rel returns the path relative to the workspace root.
+	// The input path is allowed to be either relative to the current
+	// working directory or absolute.
+	//
+	// This is useful for storing a file path in configuration.
+	//
+	// Prefer another function for displaying a path to a user, as the
+	// workspace directory might be different from the current working
+	// directory.
```
efekarakus: Similar to above, the interface can match other types besides a `*Workspace`, what do you think of just a single line:
```suggestion
	// Rel returns the path relative from the object to the target path.
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -640,6 +647,15 @@ func (o *initPipelineOpts) createPipelineManifest() error {
 		manifestExists = true
 		o.manifestPath = e.FileName
 	}
+
+	// Rel might give us a path not relative to the current working directory.
+	// So we want a path relative to the current working directory to display to the
+	// user.
+	manifestPathDisplay, err := o.pathDisplayer.DisplayPath(o.manifestPath)
```
efekarakus: For these implementation comments (not public comments), I like to think of how we can get rid of them by refactoring the code. For example, if we do the suggestion above the code becomes:
```
mftPath, err := relPath(o.cwdPath, o.manifestPath) 
```
To me this is conveying that we are going to display to the use the relative path from cwd to the provided manfiestPath.

internal/pkg/initialize/workload.go
```diff
@@ -215,7 +213,7 @@ func (w *WorkloadInitializer) initService(props *ServiceProps) (string, error) {
 		manifestExists = true
 		manifestPath = e.FileName
 	}
-	manifestPath, err = relPath(manifestPath)
+	manifestPath, err = w.Ws.Rel(manifestPath)
```
efekarakus: is this a bug fix 🤩  ?

Where the `manifestPath` was previous from the `cwd` which could have been not the root of the workspace?

internal/pkg/workspace/workspace.go
```diff
@@ -78,14 +78,16 @@ func New() (*Workspace, error) {
 	fsUtils := &afero.Afero{Fs: fs}
 	logger := log.Infof
 
-	workingDir, err := os.Getwd()
+	// note: Getwd returns an absolute path.
```
efekarakus: nit: the variable name makes this clear, can we remove the comment 🥺

efekarakus (commented):
Yay thank you Ryan! Let me know what you think of the feedback below 👇

internal/pkg/cli/cli.go
rclinard-amzn: The principal concern here is that this would mean that test cases still end up calling `filepath.Rel`. The root problem is that there exist input strings to `Rel` on Windows that cannot possibly be relativized to *any* given directory on Windows even though they might work on Linux.

I saw that you had already suggested this approach but because of the issues in the previous paragraph, it is not a simple solution. It doesn't matter if we can change the working directory that gets passed to `relPath` because no matter what the value of `cwdPath` might be, when `relPath` is called with `relPath(cwdPath, "/some/bogus/test/path")` that will *always* result in an error on Windows because the second path is not a valid path on the host filesystem. On Linux and Mac, of course, path rules are super lenient because `/` is the root and the separator, and every other character than `/` is a valid file name character, so these test cases do not cause problems on Linux and Mac. So, if I were to go forward with this approach, I would have to modify a large number of path strings in a few large test files, which would be a substantial amount of work.

But when I look at the details of these test cases, it's clear that for the most part, they are already mostly mocked away from touching the filesystem. So the fact that filepath.Rel was still getting called feels wrong - since these paths and filesystem accesses that are seemingly supposed to be mocked away are leaking through to the host filesystem - and that doesn't seem quite right to me.

This was why I used the `pathDisplayer` interface approach. Since the only current need for making a path relative to the current working directory is a detail of displaying paths to users, this felt like a better point of extensibility because it means that in test cases we can operate on file paths that are invalid on the current host file system, because we are able to know that the only significance of the result of calling DisplayPath is that it will be printed to the console - not that it will then be used to access any files on the file system. It makes it obvious what the purpose of computing the relative path is.

If you are still confident that I should remove the pathDisplayer interface I am willing to do so, but this is why I think it's a reasonable approach to the problem caused by bogus file paths in test cases.

internal/pkg/cli/env_init.go
rclinard-amzn: Good catch, not sure where that came from!

internal/pkg/initialize/workload.go
rclinard-amzn: I think so - the path seems to escape out of the function instead of just being displayed! Looking at it again, though, the path is also displayed to the user so I'll clean that up too so that the user gets the right current directory-relative path and the caller gets the correct workspace-relative path.

internal/pkg/cli/env_init.go
rclinard-amzn: Got it! It'll take some getting used to but I'll try to keep that in mind in the future.

internal/pkg/cli/interfaces.go
rclinard-amzn: Good catch, I didn't think to rename the interface when I was removing the other method from it!

internal/pkg/cli/interfaces.go
rclinard-amzn: Makes sense. I think DisplayPath still deserves a sizable comment but since DisplayPath documents the necessary pitfalls the long comments are not needed.

However, since this interface's handling of relative paths is notably different than filepath.Rel, I'll leave that documented as a comment, if that is reasonable.

internal/pkg/cli/pipeline_init.go
rclinard-amzn: I just shortened the variable name and deleted the comment according to your other comments, since the code still makes sense with those changes!

internal/pkg/cli/cli.go
efekarakus: Thanks for the explanation Ryan!
I'm still not inclined to using a `pathDisplayer` interface and field as is because reading `o.pathDisplayer.DisplayPath(summary.Path)` isn't clear to me what i'm expected to get. 

What if we allowed overriding `filepath.Rel` by embedding it as a field in our types?
```go
type initEnvOpts struct {
    cwd       string
    relPath   func(string, string) (string, error)  // Overriden in tests to work with Windows filepaths.
}

func newInitEnvOpts(vars initEnvVars) (*initEnvOpts, error) { {
   wd, err := os.Getwd()
   if err != nil {
      return nil, err
   }
   return &initEnvOpts{
       // assign other fields...
       cwd:     wd,
       relPath: relPath,
   }
}
```

rclinard-amzn: That makes sense. I will say the part about it not being fully clear about the expected result is partially by design - since the function is meant to just return a path that is appropriate to display to the user in the CLI. Though, that might be Java-ish approach carrying over - is it preferred in Go to avoid these kinds of general abstractions?

That proposed approach would work, but the main concern with it is that it would make the test cases look something like this:

```go
opts.cwd = "/test/"
opts.relPath = func(base string, fullPath string) (string, error) {
    if base == "/test/" && fullPath == "/test/Dockerfile" {
         return "Dockerfile";
    } else {
        panic("Unexpected relPath call: (" + base + ", " + fullPath + ")")
    }
}
```

Which seems a whole lot less nice to look at than:

```go
mockPathDisplayer = mocks.newMockPathDisplayer(ctrl)
opts.pathDisplayer = mockPathDisplayer
mockPathDisplayer.EXPECT().DisplayPath("/test/Dockerfile").Return("Dockerfile", nil)
```

Especially since the first two lines of the second case are shared with all test cases, while with the previous approach that function has to be repeated for every test case in the table.

efekarakus: >is it preferred in Go to avoid these kinds of general abstractions?

It doesn't have anything to do with the language, my concern is that we are introducing an abstraction to make testing on Windows possible. Since DisplayPath is just for pretty-printing information and isn't critical what if we just define the error out of existence?

```go
func displayPath(base, target string) string {
	if !filepath.IsAbs(target) {
		return filepath.Clean(target)
	}

	rel, err := filepath.Rel(base, target)
	if err != nil {
		// No path from base to target available, return target as is.
		return filepath.Clean(target)
	}
	return rel
}
```
This way we don't have to touch the tests either

rclinard-amzn: Okay, that approach makes sense, I will implement it! Thanks for all the feedback 👍

rclinard-amzn:
I think that should be fixed! Hopefully no more PR rewrites are needed now 😅

internal/pkg/cli/cli.go
```diff
@@ -136,25 +136,26 @@ func quoteStringSlice(in []string) []string {
...
+// No guarantees are given on the stability of the path across runs, all that is
+// guaranteed is that the displayed path is visually pleasing & meaningful for a
+// user.
+//
+// This path should not be stored in configuration files or used in any way except
+// for being displayed to the user.
+func displayPath(target string) string {
+	base, err := os.Getwd()
+
+	if err != nil || !filepath.IsAbs(target) {
```
efekarakus, line 151: nit: we can move the `os.Getwd` call after the if-block since `base` isn't used in the if-statement

efekarakus (approved):
🥳

PR #3745, opened 2022-07-11
chore(manifest): environment manifest does not require az if VPC is adjusted

Lou1415926:

### DIFF
--- a/internal/pkg/manifest/env.go
+++ b/internal/pkg/manifest/env.go
@@ -246,7 +246,7 @@ func (cfg *environmentVPCConfig) ManagedVPC() *template.ManagedVPC {
 	}
 	publicSubnetCIDRs := make([]string, len(cfg.Subnets.Public))
 	privateSubnetCIDRs := make([]string, len(cfg.Subnets.Public))
-	azs := make([]string, len(cfg.Subnets.Public))
+	var azs []string
 
 	// NOTE: sort based on `az`s to preserve the mappings between azs and public subnets, private subnets.
 	// For example, if we have two subnets defined: public-subnet-1 ~ us-east-1a, and private-subnet-1 ~ us-east-1a.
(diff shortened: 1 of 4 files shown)


internal/pkg/manifest/env.go
```diff
@@ -247,17 +251,20 @@ func (cfg *environmentVPCConfig) ManagedVPC() *template.ManagedVPC {
...
 		return aws.StringValue(cfg.Subnets.Private[i].AZ) < aws.StringValue(cfg.Subnets.Private[j].AZ)
 	})
 	for idx, subnet := range cfg.Subnets.Public {
 		publicSubnetCIDRs[idx] = aws.StringValue((*string)(subnet.CIDR))
 		privateSubnetCIDRs[idx] = aws.StringValue((*string)(cfg.Subnets.Private[idx].CIDR))
 		azs[idx] = aws.StringValue(subnet.AZ)
 	}
+	if azs[0] == "" {
+		azs = nil
+	}
```
iamhopaul123: this looks so weird...can we just make `var azs []string`?

internal/pkg/manifest/validate_env.go
```diff
@@ -104,28 +104,34 @@ func (cfg environmentVPCConfig) validateManagedVPC() error {
 	)
 	var exists = struct{}{}
 	for idx, subnet := range cfg.Subnets.Public {
-		if aws.StringValue((*string)(subnet.CIDR)) == "" || aws.StringValue(subnet.AZ) == "" {
-			return fmt.Errorf(`validate "subnets": public[%d] must include "cidr" and "az" fields if the vpc is configured`, idx)
+		if aws.StringValue((*string)(subnet.CIDR)) == "" {
+			return fmt.Errorf(`validate "subnets": public[%d] must include "cidr" if the vpc is configured`, idx)
```
iamhopaul123: can we use https://github.com/aws/copilot-cli/blob/8e6e7a17bb045bba643b2084d8b12b22fca447c8/internal/pkg/manifest/errors.go#L57 ?

internal/pkg/manifest/env.go
Lou1415926: updated!

internal/pkg/manifest/validate_env.go
```diff
@@ -235,3 +241,15 @@ func (c environmentConfig) validateInternalALBSubnets() error {
 	}
 	return nil
 }
+
+func areSetsEqual[T comparable](a map[T]struct{}, b map[T]struct{}) bool {
```
efekarakus, line 257: one of these days, we should add a `sets` pkg like how we have a `graphs` pkg

efekarakus (approved):
:shipit: applied label in case you'd like to address Penghao's feedback

internal/pkg/manifest/validate_env.go
Lou1415926: done!

internal/pkg/manifest/validate_env.go
iamhopaul123: Why is very basic `set()` not supported in Go 😤

iamhopaul123 (approved):

internal/pkg/manifest/validate_env.go
dannyrandall: this'll exist soon (hopefully)! https://pkg.go.dev/golang.org/x/exp/maps#Equal

PR #3747, opened 2022-07-11
chore(templates): renders `deployments` for each stage in the initial pipeline manifest

Lou1415926:
This PR:
1. Refactor unit tests in `manifest` package such that those test suites that are actually testing the integration of `NewX()` and `MarshalBinary` are moved to a new test file.
2. Add go-template rendering in to pipeline manifest YAML so that it renders `stage.deployments` if given.


internal/pkg/manifest/initial_manifest_integration_test.go
```diff
@@ -0,0 +1,413 @@
...
+			tpl, err := manifest.MarshalBinary()
+			require.NoError(t, err)
+
+			// THEN
+			require.Equal(t, string(wantedBytes), string(tpl))
+		})
+	}
+}
+
+func TestPipelineManifest_InitialManifest_Integration(t *testing.T) {
```
Lou1415926, line 337: Although GH has hidden all the diffs in this file, this test suite is actually relevant.

internal/pkg/template/templates/cicd/pipeline.yml
```diff
@@ -25,6 +25,12 @@ source:
 stages:{{range .Stages}}
     - # The name of the environment.
       name: {{.Name}}
+      {{- if .Deployments}}{{range $actionName, $action := .Deployments}}
+      deployments:
+        {{$actionName}}:
+          template_path: {{$action.TemplatePath}}
+          template_config: {{$action.TemplateConfig}}
+      {{- end}}{{end}}
```
efekarakus, line 33: ooh given this, what's the reason for having the `env/pipeline.yml` aren't they almost identical?

Lou1415926: I forgot to delete it 😅  I created that file before I realized I could just use `pipeline.yml`

efekarakus (approved):

huanjani (approved):
LGTM!

Issue #3719, opened 2022-07-05
Allow customizing the `EnvironmentSecurityGroup` with the environment manifest


efekarakus:
## Ask
For compliance users, customers might require customizing the ingress and egress of the `EnvironmentSecurityGroup`:
https://github.com/aws/copilot-cli/blob/a28a4682b4bf53e616c7840077c30e6224170498/internal/pkg/template/templates/environment/cf.yml#L138
For example, folks might want to remove the [default egress rule](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html#aws-properties-ec2-security-group--examples--Remove_the_default_rule) that gets generated by CloudFormation.

### Sample manifest

```yaml
name: test
type: 'Environment'
security_group:
  ingress:
  - <config>
  egress:
  - <config>
```

craigjbass:
This would be useful - in particular, to mitigate against supply chain attacks having services running in a public zone, accessible via an ALB but unable to initiate egress (e.g. initiate an outbound TCP connection or send UDP packets).

paragbhingre:
Hello @craigjbass,
You are right, we can very well stop the outbound connections by setting following rule in your environment manifest. 

```
network:
  vpc:
    security_group:
      egress:
        - ip_protocol: -1
          ports: 0-65536
          cidr: 127.0.0.1/32
```

[Cloudformation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html#aws-properties-ec2-security-group--examples--Remove_Default_Rule) suggests setting above mentioned egress rule in order to remove the default egress applied by the cfn and restrict traffic to just the localhost `(127.0.0.1/32)`.

This feature will go out in the next copilot release.

dannyrandall:
This feature is now released in v1.21.0! https://github.com/aws/copilot-cli/releases/tag/v1.21.0 🎉

Issue #2557, opened 2021-07-05
Allow additional security groups to the load balancer


mvn-bachhuynh-dn:
Hi guys,
I really happy when we have an options to limit IP to access `dev` env.

Because It only added in to the rule of a lister rule without delete default rule
You could see that in my screeshot
![image](https://user-images.githubusercontent.com/16200158/124476123-acf90580-ddcc-11eb-8786-a579933297cf.png)
and here is my manifest.yml
```
name: copilot-test-svc
type: Load Balanced Web Service
http:
  path: '/'
  allowed_source_ips: ["10.24.34.0/23"]
image:
  build: Dockerfile
  port: 80

cpu: 256       # Number of CPU units for the task.
memory: 512    # Amount of memory in MiB used by the task.
count: 1       # Number of tasks that should be running in your service.
exec: true     # Enable running commands in your container.

network:
        vpc:
                placement: private
```
and as you may know, only 5 conditions are allowed 
![image](https://user-images.githubusercontent.com/16200158/124476821-82f41300-ddcd-11eb-87bf-f01417f089af.png)

So, it seems this option is useless to me because I have more 5 IP to allow.

So, If copilot could allow attaching existed security group of ELB maybe more effective!

Thank you!

mvn-bachhuynh-dn:
Thank you @efekarakus ,
Hope it will be updated soon.
Beside attaching existed security group of ELB, 
Let find another way like as proactive configure it in manifest.yml to change the default security group rule of an ELB.

Thank you!

efekarakus:
>  Let find another way like as proactive configure it in manifest.yml to change the default security group rule of an ELB.


Can you elaborate? Are you looking into modifying the following security group:
https://github.com/aws/copilot-cli/blob/bb688026410283f03e1ec500238e48d45450ba61/templates/environment/versions/cf-v1.4.0.yml#L70-L87

What would you like to change?

mvn-bachhuynh-dn:
Hi @efekarakus, yes it is!
I want copilot could modify CIDR default when creating by manifest.
Currently, the security group is allow any (0.0.0.0/0) port 80/443 to public ELB.
Some case like as building develop environments, I need to IP limits.
So I could proactively config my IP in manifest.

proposal:
```
elb:
   security_group:
      ingress:
          cidrip: ["IP1", ...,"IP10"]
          fromport: <Default getting from expose of dockerfile, or setting from copilot command>
[...]

PR #3749, opened 2022-07-12
chore(manifest): add security_group to env manifest

paragbhingre:
This PR fix the issue #3719 

Now customers can define security group rules for EnvironmentSecurityGroup (e.g. ingress/egress) in their env manifest 

Sample manifest template with security group rules is given below 
```
network:
  vpc:
    security_group:
      ingress:
        IpProtocol: tcp
        FromPort: 80
        ToPort: 80
        CidrIp: 0.0.0.0/0
      egress:
        IpProtocol: tcp
        FromPort: 80
        ToPort: 80
        CidrIp: 0.0.0.0/0
```


internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -128,6 +135,30 @@ func (e *EnvStackConfig) Template() (string, error) {
 	return content.String(), nil
 }
 
+func getSecurityGroupConfig(mft *manifest.Environment) (*template.SecurityGroupConfig, error) {
+	var ingress string
+	if mft != nil && !mft.Network.VPC.SecurityGroupConfig.Ingress.IsZero() {
+		out, err := yaml.Marshal(mft.Network.VPC.SecurityGroupConfig.Ingress)
+		if err != nil {
+			return &template.SecurityGroupConfig{}, fmt.Errorf("marshal security group from environment manifest to embed in template: %v", err)
```
bvtujo: Can we distinguish these errors so that folks know whether it's the ingress or egress that's malformed? (unless there is a more descriptive error including the field name that's wrapped here. 

(also nit: can we wrap errors with %w, rather than %v?

bvtujo:
```suggestion
			return &template.SecurityGroupConfig{}, fmt.Errorf("marshal security group ingress from environment manifest to embed in template: %w", err)
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -128,6 +135,30 @@ func (e *EnvStackConfig) Template() (string, error) {
...
+		}
+		ingress = strings.TrimSpace(string(out))
+	}
+
+	var egress string
+	if mft != nil && !mft.Network.VPC.SecurityGroupConfig.Egress.IsZero() {
+		out, err := yaml.Marshal(mft.Network.VPC.SecurityGroupConfig.Egress)
+		if err != nil {
+			return &template.SecurityGroupConfig{}, fmt.Errorf("marshal security group from environment manifest to embed in template: %v", err)
```
bvtujo:
```suggestion
			return &template.SecurityGroupConfig{}, fmt.Errorf("marshal security group egress from environment manifest to embed in template: %w", err)
```

internal/pkg/manifest/env.go
```diff
@@ -64,9 +64,13 @@ type environmentNetworkConfig struct {
...
 
 type environmentVPCConfig struct {
-	ID      *string              `yaml:"id"`
-	CIDR    *IPNet               `yaml:"cidr"`
-	Subnets subnetsConfiguration `yaml:"subnets,omitempty"`
+	ID                  *string              `yaml:"id"`
+	CIDR                *IPNet               `yaml:"cidr"`
+	Subnets             subnetsConfiguration `yaml:"subnets,omitempty"`
+	SecurityGroupConfig struct {
+		Ingress yaml.Node `yaml:"ingress"`
```
bvtujo: Sorry I'm dumb, is there a reason this is an inline struct rather than its own datatype like `subnetsConfiguration`?

bvtujo (commented):
couple of dumb questions but this looks great!

internal/pkg/manifest/env.go
paragbhingre: There isn't any specific reason behind using inline struct other than readability and less code. I can very well separate it out into its own datatype, if you want me to.

internal/pkg/deploy/cloudformation/stack/env.go
paragbhingre: So do you suggest adding ingress and egress words in the error message to distinguish or you want me to do something more than that? Sorry just confirming if I understood your comment right.

bvtujo: basically just changing the error message. this may not even be necessary since some of the errors I'm thinking of may not be able to be caught by the unmarshaler.

internal/pkg/manifest/env.go
bvtujo: I think maybe just for consistency's sake with the rest of the codebase let's use a separate struct

bvtujo (approved):

bvtujo:
🚀

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
```diff
@@ -61,6 +61,57 @@ observability:
 			}(),
 			wantedFileName: "template-with-imported-certs-observability.yml",
 		},
+
+		"generate template with embedded manifest file with custom security groups rules added by te customer": {
```
huanjani: nit:
```suggestion
		"generate template with embedded manifest file with custom security groups rules added by the customer": {
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -112,6 +120,30 @@ func (e *EnvStackConfig) Template() (string, error) {
...
+
+	var egress string
+	if mft != nil && !mft.Network.VPC.SecurityGroupConfig.Egress.IsZero() {
+		out, err := yaml.Marshal(mft.Network.VPC.SecurityGroupConfig.Egress)
+		if err != nil {
+			return &template.SecurityGroupConfig{}, fmt.Errorf("marshal security group egress from environment manifest to embed in template: %w", err)
+		}
+		egress = strings.TrimSpace(string(out))
+	}
+	return &template.SecurityGroupConfig{
```
huanjani, line 150: So if the Ingress/Egress fields are empty, this will return a `*template.SecurityGroupConfig` object with nil values for ingress and egress?

Would it be more efficient to return earlier if `IsZero`, as that's the more common use case?

huanjani:
Looks good! 
Are there docs showing users how to specify their rules correctly?
Would you mind changing the example in your PR description to match the changes from your latest commit?
Feel free to remove the label when you're ready!

huanjani (approved):
Just a couple small comments!

efekarakus:
Is it possible to hold on to merging this PR until after the release please 🥺 

If we can we'd like to halt on env template CFN changes unless it's about the env manifest. Once released, we can merge the PR!

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -152,6 +152,14 @@ Resources:
       Tags:
         - Key: Name
           Value: !Sub 'copilot-${AppName}-${EnvironmentName}-env'
+{{- if and .SecurityGroupConfig .SecurityGroupConfig.Ingress }}
+      SecurityGroupIngress:
+{{.SecurityGroupConfig.Ingress | indent 8}}
+{{- end }}
+{{- if and .SecurityGroupConfig .SecurityGroupConfig.Egress }}
+      SecurityGroupEgress:
+{{.SecurityGroupConfig.Egress | indent 8}}
```
iamhopaul123, line 162: Ideally we don't put `yaml.node` for them because then it wholly depends on customers to write CFN template which is very similar to what addons should do. As a manifest field, it is usually with some level of abstraction (integrated logic). For example, [`allow_vpc_ingress`](https://github.com/aws/copilot-cli/pull/3763/files#diff-4c63a9ce04347fa71ffd19988ac84f3eca7aacf36c414bb9e8f95c011b0f622dR350) and [`from_cdn`](https://github.com/aws/copilot-cli/pull/3737/files#diff-4c63a9ce04347fa71ffd19988ac84f3eca7aacf36c414bb9e8f95c011b0f622dR339)

internal/pkg/template/env.go
```diff
@@ -118,6 +119,12 @@ type VPCConfig struct {
 	Managed  ManagedVPC
 }
 
+// SecurityGroupConfig holds the fields to import security group config
+type SecurityGroupConfig struct {
+	Ingress string
```
bvtujo, line 133: So does the marshaling logic depend on the fact that customers have to specify the exact same keys in the manifest vs in CF? 

For example, the manifest config 
```yaml
network:
  vpc:
    security_group:
      ingress:
        - IpProtocol: tcp
          FromPort: 0
          ToPort: 65535
        - IpProtocol: tcp
          FromPort: 1
          ToPort: 6
      egress:
        - IpProtocol: tcp
          FromPort: 0
          ToPort: 65535`
```
maps to the CloudFormation:
```yaml
      SecurityGroupIngress:
        - IpProtocol: tcp
          FromPort: 0
          ToPort: 65535
        - IpProtocol: tcp
          FromPort: 1
          ToPort: 6
      SecurityGroupEgress:
        - IpProtocol: tcp
          FromPort: 0
          ToPort: 65535
```

but this means that the customer has to write manifest keys in two separate capitalization styles, which may be confusing. Are there any other places in the manifest that we directly use CF keys versus remapping them to their snake case counterparts (`protocol`, `from_port`, and `to_port`)?

paragbhingre:
To address everyone's concern (including @bvtujo , @iamhopaul123 ) about accepting security_group as `yaml.Node` or other concern as if we should be allowing customers to specify CFN template inside manifest, I would like to mention that as per my discussion with @efekarakus we are going to implement security_group rules with restricted fields only. That means we will be not using `yaml.Node` neither we allow customers to define CFN fields inside manifest. The expected input from the customer would be, 

```
security_groups:
 ingress:
    - cidr:
      from_port:
      to_port:
      ip_protocol: 
```

We will target just CIDR for now which will solve some use cases such as mentioned [here](https://github.com/aws/copilot-cli/issues/2557#issuecomment-877566362)

rclinard-amzn:
It looks like this doesn't pass CI, likely due to a broken merge?

PR #3750, opened 2022-07-12
chore: impl `GenerateTemplate` for environments

efekarakus:

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/env.go
+++ b/internal/pkg/deploy/cloudformation/stack/env.go
@@ -25,8 +25,9 @@ type envReadParser interface {
 // EnvStackConfig is for providing all the values to set up an
 // environment stack and to interpret the outputs from it.
 type EnvStackConfig struct {
-	in     *deploy.CreateEnvironmentInput
-	parser envReadParser
+	in         *deploy.CreateEnvironmentInput
+	prevParams []*cloudformation.Parameter
+	parser     envReadParser
 }
 
 const (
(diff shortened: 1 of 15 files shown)


internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -198,6 +212,64 @@ func (e *EnvStackConfig) StackName() string {
 	return NameForEnv(e.in.App.Name, e.in.Name)
 }
 
+// transformParameters removes or transforms each of the current parameters and does not add any new parameters.
```
efekarakus, line 215: this section is an exact copy of https://github.com/aws/copilot-cli/blob/41856491bfda15f4eec1df7372d07c65a181b89d/internal/pkg/deploy/cloudformation/env.go#L262-L315 

We'll send follow-up PRs after this to deprecate "env upgrade" all that code will be deleted

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -198,6 +212,64 @@ func (e *EnvStackConfig) StackName() string {
...
+	for _, f := range template.AvailableEnvFeatures() {
+		isEnvControllerManaged[f] = exists
+	}
+	if _, ok := isEnvControllerManaged[aws.StringValue(new.ParameterKey)]; !ok {
+		return &new
+	}
+	return &cloudformation.Parameter{
+		ParameterKey: new.ParameterKey,
+
+		// Ideally, we would set `UsePreviousValue: true` unfortunately CodePipeline template config does not support it.
```
Lou1415926, line 267: 🥲

Lou1415926: If we are doing this for both `package` and `deploy` - one downside is that this would prevent `env deploy` to be a potential workaround for folks that want to parallelize workloads pipeline and environments pipeline. However I think this is fine because it makes a lot less sense to parallelize these two pipelines imo.

Lou1415926 (approved):

internal/pkg/deploy/cloudformation/stack/env.go
efekarakus: Ugh yeah you're totally right.

We could also have two different constructors :mildpanic: to decide how we generate parameters:

```
stack.EnvCfgForCodePipeline(in, oldParams) // Does not use UsePreviousValue
stack.EnvCfg(in, oldParams) // UsesPreviousValue
```

internal/pkg/deploy/cloudformation/stack/env.go
Lou1415926: Am I understanding correctly that https://github.com/aws/copilot-cli/blob/8ab524dc945cd80f1889d44b7860ee44cc5170c3/internal/pkg/deploy/cloudformation/env.go#L48 this is going to use `NewEnvStackFromExistingStack` too?

efekarakus: Yup that's the plan, as discussed here: https://github.com/aws/copilot-cli/pull/3750#discussion_r919444192 `UpdateAndRenderEnvironment` can invoke a slightly different constructor that will leverage `UsePreviousValue: true` just like today

dannyrandall (approved):
lgtm!

PR #3751, opened 2022-07-12
chore(cli)!: init environments pipeline from `pipeline init`

Lou1415926:


internal/pkg/utils/strings.go
```diff
@@ -0,0 +1,15 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package utils
+
+import "strconv"
+
+// QuoteStringSlice adds quotation marks around each string in the string slice.
+func QuoteStringSlice(in []string) []string {
```
paragbhingre: I think this function and func[ QuoteSliceFunc ](https://github.com/aws/copilot-cli/blob/8ab524dc945cd80f1889d44b7860ee44cc5170c3/internal/pkg/template/template_functions.go#L96) doing the same functionality. Do you think we can use the existing function? 

Also I think this is the right place of this string manipulation function but just that the code is getting duplicated.

internal/pkg/cli/pipeline_init.go
```diff
@@ -89,6 +100,7 @@ type initPipelineVars struct {
 	repoURL           string
 	repoBranch        string
 	githubAccessToken string
+	typ               string
```
paragbhingre: Can we rename this to something like `pipelineType`? Because it looks incomplete as soon as we look at it.

internal/pkg/utils/strings.go
```diff
@@ -0,0 +1,15 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package utils
```
efekarakus: 🥺 https://dave.cheney.net/practical-go/presentations/qcon-china.html#_avoid_package_names_like_base_common_or_util 🥺 

The general concern of util style pkgs is that they become the dumping ground for all sorts of stuff ( here is a zesty tweet from a sr principal at AWS: https://twitter.com/ZeGloP/status/1544799399850549248) 

In this scenario, we can either:
1. ✅  just have the duplication   in both packages (this is my preferred option)
2. Or rename the pkg to `internal/pkg/strings/quote` and rename the fn to `quote.StringSlice`

internal/pkg/cli/pipeline_init.go
```diff
@@ -221,12 +237,32 @@ func (o *initPipelineOpts) Execute() error {
 	//   - stage names (environments)
 	//   - enable/disable transition to prod envs
 	log.Infoln()
-	if err := o.createPipelineManifest(); err != nil {
-		return err
+	switch o.pipelineType {
+	case pipelineTypeWorkloads:
```
efekarakus: Let me know if you think this is overengineering, I'm totally ok with keeping what we have.

When I see a `switch` statement, I immediately think is there a way of leveraging duck-typing. For example here:
```go
type pipelineInitializer interface {
  WriteManifest() error
  WriteBuildspec() error
}

// implements pipelineInitializer
type workloadPipelineInitiliazer {
  cmd *initPipelineOpts
}
func (ini *workloadPipelineInitiliazer) WriteManifest() error {...}
func (ini *workloadPipelineInitializer) WriteBuildspec() error {...}

// implements pipelineInitializer
type envPipelineInitializer {
  cmd *initPipelineOpts
}
func (ini *envPipelineInitializer) WriteManifest() error {...}
func (ini *envPipelineInitializer) WriteBuildspec() error {...}

// factory
func newPipelineInitializer(cmd *initPipelineOpts) pipelineInitializer
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -318,6 +354,35 @@ func (o *initPipelineOpts) askPipelineName() error {
...
+		}
+		return fmt.Errorf("invalid pipeline type %q; must be one of %s", o.pipelineType, english.WordSeries(utils.QuoteStringSlice(pipelineTypes), "or"))
+	}
+
+	typ, err := o.prompt.SelectOption("What type of continuous delivery pipeline is this?",
+		"A pipeline can be set up to deploy either your workloads or your environments",
+		[]prompt.Option{
+			{
+				Value: pipelineTypeWorkloads,
+				Hint:  "Deploy the workloads in your workspace",
```
efekarakus: nit: We can use slightly different wording in the hint
```suggestion
				Hint:  "Deploy the services or jobs in your workspace",
```

internal/pkg/cli/pipeline_init.go
```diff
@@ -817,6 +891,6 @@ func buildPipelineInitCmd() *cobra.Command {
 	_ = cmd.Flags().MarkHidden(githubAccessTokenFlag)
 	cmd.Flags().StringVarP(&vars.repoBranch, gitBranchFlag, gitBranchFlagShort, "", gitBranchFlagDescription)
 	cmd.Flags().StringSliceVarP(&vars.environments, envsFlag, envsFlagShort, []string{}, pipelineEnvsFlagDescription)
-
+	cmd.Flags().StringVar(&vars.pipelineType, pipelineTypeFlag, "", pipelineTypeFlagDescription)
```
efekarakus: should we take a short flag like `-t, --svc-type` similar to `svc init`?

efekarakus (commented):
Thank you Wanxian!!

internal/pkg/cli/pipeline_init.go
Lou1415926: using `-p` for shorthand because `-t` is already taken by github-token flag :sad-panda:

internal/pkg/cli/pipeline_init.go
```diff
@@ -318,6 +400,35 @@ func (o *initPipelineOpts) askPipelineName() error {
...
+	if o.pipelineType != "" {
+		for _, typ := range pipelineTypes {
+			if o.pipelineType == typ {
+				return nil
+			}
+		}
+		return fmt.Errorf("invalid pipeline type %q; must be one of %s", o.pipelineType, english.WordSeries(quoteStringSlice(pipelineTypes), "or"))
+	}
+
+	typ, err := o.prompt.SelectOption("What type of continuous delivery pipeline is this?",
```
bvtujo, line 413: Adding a new prompt means that saved pipeline init commands which executed previously will now require an additional user input, which means we may have to mark this release as containing a breaking change.

Just want to make sure we know to highlight it in the release docs if that's the case.

bvtujo (commented):
Looks awesome! Nice work on the duck typed factories.

bvtujo (approved):

internal/pkg/utils/strings.go
Lou1415926: Adopted option 1 here: https://github.com/aws/copilot-cli/pull/3751#discussion_r920389517!

efekarakus (approved):

internal/pkg/cli/pipeline_init.go
Lou1415926: 👍🏼 ! Fortunately (typically) `copilot pipeline init` should not be scripted and hopefully nobody does that :crylaugh:

Issue #3546, opened 2022-05-10
Package addons CloudFormation templates before deployments


efekarakus:
## Ask

Allow writing templates in the `addons/` directory that can refer to local artifacts (local paths).  
Copilot should upload local artifacts, such as source code for an AWS Lambda function or a Swagger file for an AWS API Gateway REST API, to an S3 bucket. Following that, it should write an edited version of the template replacing references to local artifacts with the S3 location of the artifacts. Finally, this transformed template should be uploaded to the S3 bucket and used for the `AddonsStack`.

This functionality should be in parity with the [`aws cloudformation package`](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/package.html) command.
Gitter thread: https://gitter.im/aws/copilot-cli?at=627a8caaeaf3267f8553fef6

### Example

Let's imagine I have the following workspace structure:
```
.
├── copilot/
│   └── api/
│       ├── manifest.yml
│       └── addons/
│           └── ddb.yml
└── src/
    ├── ...
    └── lambda/
        └── stream-replicator.py

```
In the `ddb.yml` template, I'd like to refer to the `src/lambda/stream-replicator.py` function directly in the template. Otherwise, I'd need to run `aws cloudformation package --template-file copilot/api/addons/ddb.yml` before `copilot deploy`. Instead, I want copilot to take over this burden:
```yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Resources:
  ProcessDynamoDBStream:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: /src/lambda/stream-replicator.py  # NEW
      ...
```

Copilot should zip the file and upload it to the environment regional S3 bucket:
https://github.com/aws/copilot-cli/blob/2d1669446263c8ca49bd35b30e220ebca956b2ad/internal/pkg/deploy/env.go#L29

This is already where Copilot uploads pipeline, and manual deployment artifacts.

Then Copilot will transform the template such that the Serverless Function's `CodeUri` has the s3 URL as value.

### Specification
We should have parity with the AWS CLI:
> 

        BodyS3Location property for the AWS::ApiGateway::RestApi resource
        Code property for the AWS::Lambda::Function resource
        CodeUri property for the AWS::Serverless::Function resource
        DefinitionS3Location property for the AWS::AppSync::GraphQLSchema resource
        RequestMappingTemplateS3Location property for the AWS::AppSync::Resolver resource
[...]

PR #3754, opened 2022-07-12
feat: package addon assets

dannyrandall:
Includes support for the following resources:
```
AWS::ApiGateway::RestApi
AWS::Lambda::Function
AWS::Lambda::LayerVersion
AWS::AppSync::GraphQLSchema
AWS::AppSync::Resolver
AWS::AppSync::FunctionConfiguration
AWS::ElasticBeanstalk::ApplicationVersion
AWS::Glue::Job
AWS::StepFunctions::StateMachine
AWS::CodeCommit::Repository
```

Support for `AWS::CloudFormation::Stack` resources and the `AWS::Include` transform coming in future PRs.

Relevant: #3546


internal/pkg/addon/addons.go
```diff
@@ -93,10 +122,19 @@ func (a *Addons) Template() (string, error) {
 			return "", err
 		}
 	}
+
+	if a.UploadAssets {
+		mergedTemplate, err = a.packageLocalArtifacts(mergedTemplate)
+		if err != nil {
+			return "", fmt.Errorf("package local artifacts: %w", err)
```
iamhopaul123: nit: we don't have to wrap up the error msg for private functions.

internal/pkg/addon/addons.go
```diff
@@ -64,6 +85,14 @@ func New(wlName string) (*Addons, error) {
 // If the addons directory doesn't exist, it returns the empty string and
 // ErrAddonsDirNotExist.
 func (a *Addons) Template() (string, error) {
+	a.templateOnce.Do(func() {
```
iamhopaul123: Why do we need this 🤔

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,302 @@
...
+	if err != nil {
+		return fmt.Errorf("upload asset: %w", err)
+	}
+
+	bucket, key, err := s3.ParseURL(url)
+	if err != nil {
+		return fmt.Errorf("parse s3 url: %w", err)
+	}
+
+	fmt.Printf("Uploaded %s to s3 at: %s\n", assetPath, s3.Location(bucket, key))
```
iamhopaul123: Do we need this 🤔

internal/pkg/aws/s3/s3.go
```diff
@@ -170,6 +170,10 @@ func URL(region, bucket, key string) string {
 	return fmt.Sprintf("https://%s.s3.%s.amazonaws.%s/%s", bucket, region, tld, key)
 }
 
+func Location(bucket, key string) string {
```
iamhopaul123, line 174: comment pls and same for the others

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,302 @@
...
+
+		propsNode, ok := res["Properties"]
+		if !ok || propsNode.Kind != yaml.MappingNode {
+			continue
+		}
+
+		transforms, ok := transformInfoFor[typeNode.Value]
+		if !ok {
+			continue
+		}
```
iamhopaul123: nit this can be moved to line 120

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,302 @@
...
+	Property           []string
+	BucketNameProperty string
+	ObjectKeyProperty  string
+
+	ForceZip bool
+}
+
+// TODO(dnrnd) AWS::Include.Location
+// TODO(dnrnd) AWS::CloudFormation::Stack
+var transformInfoFor = map[string][]transformInfo{
```
iamhopaul123, line 79: Can we give it a comment describing what these are for?

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,302 @@
...
+)
+
+type uploader interface {
+	Upload(bucket, key string, data io.Reader) (string, error)
+}
+
+type transformInfo struct {
+	Property           []string
+	BucketNameProperty string
+	ObjectKeyProperty  string
```
iamhopaul123: We need more explanation on these. It took me a while to understand property is a list of string but they are not in parallel with each other but a path chain instead. and `BucketNameProperty` `ObjectKeyProperty` are only for end node schema names.

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,302 @@
...
+		var ok bool
+		node, ok = mapNode[key]
+		if !ok || (i+1 != len(tr.Property) && node.Kind != yaml.MappingNode) {
+			return nil // no error if the property doesn't exist
+		}
+		mapNode = mappingNode(node)
+	}
+
+	if node == nil || node.Kind != yaml.ScalarNode {
+		// only transform if the node is a scalar node
```
iamhopaul123: Does this only happen when CFN changes their schema? should we give a warning instead? At least users would be aware of the malfunctioning and let us know

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+func (a *Addons) transformProperty(properties *yaml.Node, tr transformInfo) error {
+	mapNode := mappingNode(properties)
+	var node *yaml.Node
+	for i, key := range tr.Property {
+		var ok bool
+		node, ok = mapNode[key]
+		if !ok || (i+1 != len(tr.Property) && node.Kind != yaml.MappingNode) {
+			return nil // no error if the property doesn't exist
```
iamhopaul123: same here should we give a warning instead?

internal/pkg/addon/addons.go
dannyrandall: good catch, just me pre-optimizing for multiple threads calling `Template()` 🙄 i'll remove it

internal/pkg/addon/package.go
dannyrandall: both the aws cli and sam cli print out where they are uploading files...it's definitely not a need, but i thought it might be nice. i'm fine to remove it if you think it's better to not surface it.

internal/pkg/addon/package.go
dannyrandall: done!

internal/pkg/addon/package.go
dannyrandall: see what you think about the descriptions :)

internal/pkg/addon/addons.go
dannyrandall: why not though 🥺

iamhopaul123: Well it is to prevent the error wrapping becoming too verbose. However, it is not a must thing. In some circumstance it makes more sense to wrap errors from private functions to provide more context (but usually the inner problem is the package is too ambiguous and everything coupled tightly together)

internal/pkg/aws/s3/s3.go
dannyrandall: think i got all of them! 🤞

internal/pkg/addon/package.go
dannyrandall: cleaned it up a bit, should be better. it would have happened if some cases where we don't want to transform, like they had already specified an s3 location like this:
```
Properties:
  Code: # not a yaml.ScalerNode
    S3Bucket: bucket
    S3Key: key
```

internal/pkg/addon/addons.go
dannyrandall: problem is if one of the errors isn't wrapped in the chain, then you can't call `errors.As/Is` any errors underneath that. but you're right - unless we want users (of the package) to be able to discriminate on implementation errors from this package, we shouldn't wrap. it would lead to tighter coupling between this package and someone using it. i'll remove it 😊

internal/pkg/addon/package.go
dannyrandall: i don't think so - it doesn't necessarily mean they've misconfigured or anything. Just means that for whatever reason the property isn't there. maybe like
```
Type: AWS::CodeCommit::Repository
Properties:
  RandomProp1: value
  Code:
    SomeOtherProp: value
```
Just because `S3` isn't present, doesn't necessarily mean their configuration is invalid (just an ex, it might be required for this resource, idk)

dannyrandall:
I'm going to split this PR into 2 or 3 smaller PRs.

PR #3756, opened 2022-07-13
chore(cli): add the flag to force an `env deploy` update

Lou1415926:


internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -70,6 +72,15 @@ func NewEnvStackConfig(input *deploy.CreateEnvironmentInput) *EnvStackConfig {
 	}
 }
 
+// NewEnvConfigFromExistingStack returns a CloudFormation stack configuration for updating an environment.
+func NewEnvConfigFromExistingStack(in *deploy.CreateEnvironmentInput, prevForceUpdateID string) *EnvStackConfig {
```
dannyrandall: i feel like calling it `prevForceUpdateID` here (and in the `EnvStackConfig` struct) is a little confusing, bc it is the current forceUpdateID that the env stack is currently on. i see what you mean though😊 

maybe just change it to just `ForceUpdateID`?

dannyrandall (approved):

internal/pkg/template/env.go
```diff
@@ -108,6 +113,7 @@ type EnvOpts struct {
 
 	LatestVersion      string
 	SerializedManifest string // Serialized manifest used to render the environment template.
+	ForceUpdateID      string
```
CaptainCarpensir, line 116: Maybe make any occurences of `ForceUpdateID` a pointer to reflect that not every deployment is forced? I know "" covers that currently, but it may work better if it's needed to be used in the future to check if it's nil.

O/w LGTM

CaptainCarpensir (approved):

internal/pkg/cli/env_deploy.go
```diff
@@ -215,5 +217,6 @@ Deploy an environment named "test".
 	}
 	cmd.Flags().StringVarP(&vars.appName, appFlag, appFlagShort, tryReadingAppName(), appFlagDescription)
 	cmd.Flags().StringVarP(&vars.name, nameFlag, nameFlagShort, "", envFlagDescription)
+	cmd.Flags().BoolVar(&vars.forceNewUpdate, forceFlag, false, forceFlagDescription)
```
efekarakus: I think we need to update the `forceFlagDescription` right now it's specific to services. Something like:
```go
const forceEnvDeployFlagDescription = "Optional. Force update the environment stack template." 
```

internal/pkg/deploy/cloudformation/env.go
```diff
@@ -238,6 +254,9 @@ func (cf CloudFormation) toUploadedStack(artifactBucketARN string, stackConfig S
 }
 
 func (cf CloudFormation) waitAndDescribeStack(stackName string) (*cloudformation.StackDescription, error) {
+	if cf.cachedDeployedStack != nil {
+		return cf.cachedDeployedStack, nil
```
efekarakus: is there a scenario where caching can be problematic? are there any dependent functions that you think would want the latest value. For example, if we cache the `StackStatus` will never be updated

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -18,15 +19,16 @@ import (
...
 	ParseEnvBootstrap(data *template.EnvOpts, options ...template.ParseOption) (*template.Content, error)
 }
 
 // EnvStackConfig is for providing all the values to set up an
 // environment stack and to interpret the outputs from it.
 type EnvStackConfig struct {
-	in     *deploy.CreateEnvironmentInput
-	parser envReadParser
+	in                *deploy.CreateEnvironmentInput
+	PrevForceUpdateID string
```
efekarakus: can this be private?

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -642,3 +642,6 @@ Outputs:
     Description: The ID of the Copilot-managed EFS filesystem.
     Export:
       Name: !Sub ${AWS::StackName}-FilesystemID
+  DeploymentController:
```
efekarakus: What do you think of naming this out like this?
```suggestion
  LastForceDeployID: 
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -85,6 +96,15 @@ func (e *EnvStackConfig) Template() (string, error) {
 		}
 		mft = string(out)
 	}
+
+	forceUpdateID := e.PrevForceUpdateID
+	if e.in.ForceUpdate {
+		id, err := uuid.NewRandom()
```
efekarakus, line 92: 💭 random thoughts - no action item:

I was thinking if we need to have a way of making this ID more understandable like using the git commit SHA so that they can figure out what caused this template change. But I realized that we have `env show --manifest` now ✅ so they can always figure what caused the template change

internal/pkg/cli/deploy/env.go
```diff
@@ -184,6 +190,7 @@ func (d *envDeployer) buildStackInput(in *DeployEnvironmentInput) (*deploy.Creat
 		ArtifactBucketARN:    s3.FormatARN(partition.ID(), resources.S3Bucket),
 		ArtifactBucketKeyARN: resources.KMSKeyARN,
 		Mft:                  in.Manifest,
+		ForceUpdate:          in.ForceNewUpdate,
```
efekarakus, line 203: It's kinda awkward that this field is set for `GenerateCloudFormationTemplate` but not for `DeployEnvironment` where that field gets populated instead in `UpdateAndRenderEnvironment` 💭 

What do you think of moving the logic to outside of `UpdateAndRenderEnvironment` into `DeployEnvironment` so that we do all the initialization in a single pkg?

internal/pkg/template/env.go
Lou1415926: I've been actually torn between whether to make it a pointer! 

If it is a pointer, we'd probably have in `cf.yml`:
```
{{- if .ForceUpdateID}}
DeploymentController:
  Value: {{.ForceUpdateID}}
{{- end}}
```

The UX would (probably) be:
1. I as a user, have been running `copilot env deploy` without the flag ➡️  no `DeploymentController` in my stack
2. One day, I ran `copilot env deploy --force` ➡️  `DeploymentController` shows up ➡️  I guess it makes sense because I used `--force`.
3. I then ran `copilot env deploy` ➡️  `DeploymentController` is still there though the value doesn't change ➡️  Confused! I thought `DeploymentController` only shows up when I `--force`!

The pros is that for people that will never ever use `--force` flag, this crazy output won't ever get added to their template. However I'm worried about 3.

On the other hand, if we keep it as it, without the pointer. The UX would be:
4. I as a user, have been running `copilot env deploy` without the flag ➡️  `DeploymentController` has always been in my stack.
5. One day, I ran `copilot env deploy --force` ➡️  `DeploymentController` value changes ➡️  I guess it makes sense because I used `--force`.
6. I then ran `copilot env deploy` ➡️  `DeploymentController` is still there though the value doesn't change ➡️  I'm familiar with this output because It's been always there. I guess it just changes value when I run `--force`. 

The cons is that it's going to be there even if the user never uses `--force`. However, it doesn't introduce the "irreversible flip" as described in 3.

internal/pkg/deploy/cloudformation/env.go
Lou1415926: > are there any dependent functions that you think would want the latest value

There are currently two dependent calls, one is getting the old parameters, the other is getting for force update ID. Between these two, I think it's safe to assume that no matter which is the first to invoke this function, the second call follows immediately such that the cached stack is the latest stack.

I do see your concern though! I felt like retrieving old parameters and force update ID do not warrant two API calls. I can refactor the functions such that we don't need to cache while still maintaining just one API call.

internal/pkg/template/env.go
CaptainCarpensir: I can see that concern. Would it then be possible to do something like resetting the ID when `--force` is unused?

internal/pkg/deploy/cloudformation/stack/env.go
Lou1415926: what do ya think of `lastForceUpdateID`? I get that `prev` sounds like it's not going to be the same value anymore though in reality it just depends on `in.ForceUpdate`. But maybe `last` sounds better?

dannyrandall: i think `last` is perf 👌

internal/pkg/cli/deploy/env.go
Lou1415926: done!

internal/pkg/template/env.go
Lou1415926: This will create an unintentional force update :( The way it works is that changing the `Outputs` section can force the stack to update, even when there is "no change in changeset". Updating the value of `LastForceUpdateID`, adding `LastForceUpdateID` and removing `LastForceUpdateID` all count as "changing the `Outputs` section, thus creating a force update (unless there are indeed changes in change set)

internal/pkg/cli/deploy/env.go
```diff
@@ -18,19 +18,20 @@ import (
...
-	termprogress "github.com/aws/copilot-cli/internal/pkg/term/progress"
 )
 
 type appResourcesGetter interface {
 	GetAppResourcesByRegion(app *config.Application, region string) (*stack.AppRegionalResources, error)
 }
 
 type environmentDeployer interface {
-	UpdateAndRenderEnvironment(out termprogress.FileWriter, env *deploy.CreateEnvironmentInput, opts ...cloudformation.StackOption) error
+	UpdateAndRenderEnvironment(out progress.FileWriter, conf deploycfn.StackConfiguration, bucketARN string, opts ...cloudformation.StackOption) error
```
efekarakus, line 32: Why did you prefer to stop taking the `env *deploy.CreateEnvInput` as input? is it that we would need to pass the old params and that's ugly too?

internal/pkg/cli/env_deploy.go
```diff
@@ -134,8 +137,16 @@ func (o *deployEnvOpts) Execute() error {
...
+		ForceNewUpdate:      o.forceNewUpdate,
 		RawManifest:         rawMft,
 	}); err != nil {
+		var errEmptyChangeSet *awscfn.ErrChangeSetEmpty
+		if errors.As(err, &errEmptyChangeSet) {
+			log.Errorf(`It seems like your update does not introduce immediate resource changes. 
+This may because the resources being modified are not created yet until deemed necessary by the services in your environment.
+In this case, you can try %s. This will deploy the modified template, even if there is no immediate changes.
+`, color.HighlightCode("copilot env deploy --force"))
+		}
```
efekarakus:
```suggestion
			log.Errorf(`Your update does not introduce immediate resource changes. 
This may be because the resources are not created until they are deemed 
necessary by a service deployment.

In this case, you can run %s to push a modified template, even if there are no immediate changes.
`, color.HighlightCode("copilot env deploy --force"))
		}
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -642,3 +642,6 @@ Outputs:
     Description: The ID of the Copilot-managed EFS filesystem.
     Export:
       Name: !Sub ${AWS::StackName}-FilesystemID
+  LastForceDeployID:
+    Value: {{quote .ForceUpdateID}}
+    Description: Optionally force the stack to update when no immediate resource change is present.
```
```suggestion
    Description: Optionally force the template to update when no immediate resource change is present.
```

efekarakus (approved):

internal/pkg/cli/deploy/env.go
Lou1415926: After addressing https://github.com/aws/copilot-cli/pull/3756#discussion_r920566795, `env` is no longer needed in this method because it's only used to create `conf deploycfn.StackConfiguration` which is now passed in as a parameter (as the result of offloading those initiation to `cli/deploy/env.go`

PR #3757, opened 2022-07-13
chore: add logic in underlying libraries for job logs

bvtujo:
- chore: add job logs logic and validation
- chore: update cloudwatchlogs to allow limiting log streams fetched and custom prefixes
- chore: rename service.go to workload.go in logging pkg
- chore: add logic for fetching state machine logs to workload
- chore: add unit tests for workload logging pkg

Adds unit tests and changes for the cloudwatchlogs and logging libraries necessary for job logs, specifically for getting one or N executions at a time, optionally including state machine logs. 


internal/pkg/aws/cloudwatchlogs/cloudwatchlogs_test.go
```diff
@@ -237,6 +238,146 @@ func TestLogEvents(t *testing.T) {
...
+			logGroupName:   "mockLogGroup",
+			logStreamLimit: aws.Int64(1),
+			mockcloudwatchlogsClient: func(m *mocks.Mockapi) {
+				m.EXPECT().DescribeLogStreams(&cloudwatchlogs.DescribeLogStreamsInput{
+					LogGroupName: aws.String("mockLogGroup"),
+					Descending:   aws.Bool(true),
+					OrderBy:      aws.String("LastEventTime"),
+					Limit:        aws.Int64(1),
+				}).Return(&cloudwatchlogs.DescribeLogStreamsOutput{
+					LogStreams: []*cloudwatchlogs.LogStream{
```
huanjani, line 294: I may be reading this wrong, but as in the next test, should this return more than one log stream so that we can confirm that only one was fetched?

internal/pkg/cli/job_logs.go
```diff
@@ -23,6 +23,9 @@ import (
 
 const (
 	jobAppNamePrompt = "Which application does your job belong to?"
+
+	jobLogNamePrompt     = "Which job's logs would you like to show?"
+	jobLogNameHelpPrompt = "The logs of a deployed job will be shown."
```
huanjani: nit: maybe something like this, bc this is the name prompt?
```suggestion
	jobLogNameHelpPrompt = "The logs of the indicated deployed job will be shown."
```

internal/pkg/aws/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -46,11 +46,13 @@ type LogEventsOutput struct {
 // LogEventsOpts wraps the parameters to call LogEvents.
 type LogEventsOpts struct {
 	LogGroup            string
-	LogStreams          []string // If nil, retrieve logs from all log streams.
+	LogStreamPrefixes   []string // If nil, retrieve logs from all log streams.
```
iamhopaul123: Maybe
```suggestion
	LogStreamPrefixesFilter   []string
```

internal/pkg/logging/workload.go
```diff
@@ -49,18 +51,22 @@ type WriteLogEventsOpts struct {
 	TaskIDs   []string
 	// OnEvents is a handler that's invoked when logs are retrieved from the service.
 	OnEvents func(w io.Writer, logs []HumanJSONStringer) error
+	// LogStreamLimit is an optional parameter for jobs and tasks to speed up CW queries
+	// involving multiple log streams.
+	LogStreamLimit *int64
```
iamhopaul123: Is this going to be a new flag or how do users specify this?

internal/pkg/aws/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -60,13 +62,15 @@ func New(s *session.Session) *CloudWatchLogs {
...
-func (c *CloudWatchLogs) logStreams(logGroup string, logStreams ...string) ([]string, error) {
-	resp, err := c.client.DescribeLogStreams(&cloudwatchlogs.DescribeLogStreamsInput{
+// logStreams returns all name of the log streams in a log group with optional limit and prefix filters.
+func (c *CloudWatchLogs) logStreams(logGroup string, logStreamLimit *int64, logStreamPrefixes ...string) ([]string, error) {
+	describeLogStreamsInput := &cloudwatchlogs.DescribeLogStreamsInput{
 		LogGroupName: aws.String(logGroup),
 		Descending:   aws.Bool(true),
 		OrderBy:      aws.String(cloudwatchlogs.OrderByLastEventTime),
-	})
+		Limit:        logStreamLimit,
```
iamhopaul123: I feel like we should do pagination and do the filtering for each page and then apply this limit after that (quit querying the next page if limit is reached). Otherwise it could happen that `logStreamLimit` is 20 whereas the none of the first 20 log streams with the latest events matches the log stream prefixes, resulting no log stream returned.

internal/pkg/logging/workload.go
bvtujo: this is specified by the `--last` flag; as in `job logs for the last N executions`.

internal/pkg/aws/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -60,38 +62,55 @@ func New(s *session.Session) *CloudWatchLogs {
...
+				continue
+			}
+			streams = append(streams, name)
+		}
+		if len(logStreamPrefixes) != 0 {
+			logStreamNames = append(logStreamNames, filterStringSliceByPrefix(streams, logStreamPrefixes)...)
+		} else {
+			logStreamNames = append(logStreamNames, streams...)
+		}
+		if len(logStreamNames) > logStreamLimit {
```
iamhopaul123: nit
```suggestion
		if len(logStreamNames) >= logStreamLimit {
```

internal/pkg/aws/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -147,6 +166,13 @@ func truncateEvents(limit int, events []*Event) []*Event {
 	return events[len(events)-limit:] // Only grab the last N elements where N = limit
 }
 
+func truncateStreams(limit int, streams []string) []string {
+	if len(streams) <= limit {
+		return streams
+	}
+	return streams[:len(streams)-limit]
```
iamhopaul123: Should this be
```suggestion
	return streams[:limit]
```

huanjani (approved):
LGTM!

iamhopaul123 (approved):
🚢

PR #3759, opened 2022-07-13
chore: describe public alb change to allow cloudfront display

CaptainCarpensir:
Allow CloudFront URI to be displayed on top of public ALB. I also changed the struct name from `albURI` to `accessURI` as it better fits the description when the URI is the access point to the  environment, rather than just to the ALB. When PR for prefix lists #3737 is merged, I'll then modify these files to allow to disable showing the ALB DNS altogether.

Addresses #3701 


internal/pkg/template/templates/environment/cf.yml
```diff
@@ -642,3 +642,12 @@ Outputs:
     Description: The ID of the Copilot-managed EFS filesystem.
     Export:
       Name: !Sub ${AWS::StackName}-FilesystemID
+{{- if .CDNConfig}}
+  CloudFrontDomainName:
+    Condition: CreateALB
+    Value: !GetAtt CloudFrontDistribution.DomainName
+    Description: The domain name of an associated Cloud Front Distribution.
+    Export:
+      Name: !Sub ${AWS::StackName}-CloudFrontDomainName
```
efekarakus: If the Output won't be used by other stacks (like if we can't think of a reason that it would be used in addons) we can remove the `Export` field for now and add it later if anyone requests it:
```suggestion
```

internal/pkg/describe/uri.go
```diff
@@ -91,13 +91,13 @@ func (d *LBWebServiceDescriber) URI(envName string) (URI, error) {
...
-			envDNSNameKey:   envOutputPublicLoadBalancerDNSName,
+			envDNSTypeKey:   envOutputPublicLoadBalancerDNSName,
 		}
-		albURI, err := albDescr.uri()
+		publicURI, err := albDescr.uri()
 		if err != nil {
 			return URI{}, err
 		}
-		uri.albURI = albURI
+		uri.publicURI = publicURI
```
efekarakus: For variables, types, fns the recommendation from Go is to avoid stutter: https://github.com/golang/go/wiki/CodeReviewComments#package-names
`uri.publicURI` is a bit redundant compared to `uri.public`

internal/pkg/describe/uri.go
```diff
@@ -220,37 +220,41 @@ type albDescriber struct {
 	svcDescriber    ecsDescriber
 	envDescriber    envDescriber
 	initLBDescriber func(string) (lbDescriber, error)
-	envDNSNameKey   string
+	envDNSTypeKey   string
```
efekarakus: What do you think of renaming this to
```suggestion
	albCfnOutputName   string // The DNS name for the public or private ALB.
```

internal/pkg/describe/uri.go
```diff
@@ -172,17 +172,17 @@ func (d *BackendServiceDescriber) URI(envName string) (URI, error) {
 				svcDescriber:    svcDescr,
 				envDescriber:    envDescr,
 				initLBDescriber: d.initLBDescriber,
-				envDNSNameKey:   envOutputInternalLoadBalancerDNSName,
+				envDNSTypeKey:   envOutputInternalLoadBalancerDNSName,
 			}
-			albURI, err := albDescr.uri()
+			publicURI, err := albDescr.uri()
```
efekarakus: Should this be private URI? because it's from the internal load balancer

internal/pkg/describe/uri.go
```diff
@@ -220,37 +220,41 @@ type albDescriber struct {
...
+func (d *albDescriber) envDNSName(path string) (publicURI, error) {
 	envOutputs, err := d.envDescriber.Outputs()
 	if err != nil {
-		return albURI{}, fmt.Errorf("get stack outputs for environment %s: %w", d.env, err)
+		return publicURI{}, fmt.Errorf("get stack outputs for environment %s: %w", d.env, err)
 	}
-	return albURI{
-		DNSNames: []string{envOutputs[d.envDNSNameKey]},
+	dnsNames := []string{envOutputs[d.envDNSTypeKey]}
+	if cfDNS, ok := envOutputs[envOutputCloudFrontDomainName]; ok {
```
efekarakus: Do we display only the CloudFrontDomainName is the security group ingress is restricted to cloudfront? Is there a way of achieving that 🥺

internal/pkg/describe/uri.go
```diff
@@ -220,37 +220,41 @@ type albDescriber struct {
 	svcDescriber    ecsDescriber
 	envDescriber    envDescriber
 	initLBDescriber func(string) (lbDescriber, error)
-	envDNSNameKey   string
+	envDNSTypeKey   string
 }
 
-func (d *albDescriber) envDNSName(path string) (albURI, error) {
+func (d *albDescriber) envDNSName(path string) (publicURI, error) {
```
efekarakus: is the return type always a "publicURI" ? since a `*BackendServiceDescriber` also calls this method?

To me this seems to indicate that `publicURI` is potentially the wrong refactor for `albURI`.

I see that we have `LBWebServiceURI` that's composed of, can we potentially add a new one for CloudFront?
```go
        albURI albURI
	nlbURI nlbURI
	cloudFrontURI cloudFrontURI
```

internal/pkg/describe/uri.go
iamhopaul123: For backend service, the URI should be private instead of public 🤔

internal/pkg/describe/uri.go
```diff
@@ -91,13 +91,13 @@ func (d *LBWebServiceDescriber) URI(envName string) (URI, error) {
 			svcDescriber:    svcDescr,
 			envDescriber:    envDescr,
 			initLBDescriber: d.initLBDescriber,
-			envDNSNameKey:   envOutputPublicLoadBalancerDNSName,
+			envDNSTypeKey:   envOutputPublicLoadBalancerDNSName,
```
iamhopaul123: Why do we rename this?

internal/pkg/describe/uri.go
iamhopaul123, line 232: Use `albDescriber` to get cf dns looks wild. Maybe rename `albDescriber`?

internal/pkg/describe/uri_test.go
```diff
@@ -542,7 +542,7 @@ func TestLBWebServiceURI_String(t *testing.T) {
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
 			uri := &LBWebServiceURI{
-				albURI: albURI{
+				publicURI: publicURI{
```
iamhopaul123: can we add a test case for dns name being as cf domain name?

internal/pkg/describe/uri.go
```diff
@@ -316,13 +320,13 @@ func (d *RDWebServiceDescriber) URI(envName string) (URI, error) {
...
 // LBWebServiceURI represents the unique identifier to access a load balanced web service.
 type LBWebServiceURI struct {
-	albURI albURI
-	nlbURI nlbURI
+	publicURI publicURI
+	nlbURI    nlbURI
 }
 
-type albURI struct {
+type publicURI struct {
```
iamhopaul123:
```suggestion
type accessURI struct {
```

internal/pkg/describe/uri.go
CaptainCarpensir: Yes there is! I'm planning on doing this in one more PR because the PrefixList PR isn't merged yet.

internal/pkg/describe/uri.go
CaptainCarpensir: The issue with this was that I found it added a lot of complexity. Most of the time I spent making this originally was spent thinking about whether to add a new `cfURI` or make it part of the `albURI`. The reason I decided to make it part of the `albURI` was that the schema already reflected very well how the URI for CloudFront would interact with describe. Essentially it's tacked on to the public facing load balancer, and I wanted that to directly reflect in the struct design. Yesterday I spent a lot of time figuring out how this would work, and unfortunately I forgot some of the more direct reasons as to why I prefer how it currently functions. In hindsight I should've written my thoughts down 😔 but I trust the decision I made yesterday.

As penghao pointed out, renaming to `accessURI` fits it fairly well. If you're not satisfied with this though, I'll spend the time to look into this refactor again, and write up some thoughts about taking either direction so we can have a more conclusive solution.

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -642,6 +642,12 @@ Outputs:
     Description: The ID of the Copilot-managed EFS filesystem.
     Export:
       Name: !Sub ${AWS::StackName}-FilesystemID
+{{- if .CDNConfig}}
+  CloudFrontDomainName:
+    Condition: CreateALB
```
dannyrandall, line 648: aren't we only using cf if a customer enables it for their environment? i feel like the condition `CreateALB` doesn't quite fit for that, but maybe i'm missing something 🤔

internal/pkg/describe/uri.go
```diff
@@ -85,19 +85,19 @@ func (d *LBWebServiceDescriber) URI(envName string) (URI, error) {
...
 	var uri LBWebServiceURI
 	if albEnabled {
-		albDescr := &albDescriber{
-			svc:             d.svc,
-			env:             envName,
-			svcDescriber:    svcDescr,
-			envDescriber:    envDescr,
-			initLBDescriber: d.initLBDescriber,
-			envDNSNameKey:   envOutputPublicLoadBalancerDNSName,
+		albOrCdnDescr := &albOrCdnDescriber{
```
iamhopaul123: `albOrCdn` looks strange...maybe just `uriDescriber`?

internal/pkg/describe/uri.go
```diff
@@ -214,43 +214,47 @@ func (d *BackendServiceDescriber) URI(envName string) (URI, error) {
...
-	envDescriber    envDescriber
-	initLBDescriber func(string) (lbDescriber, error)
-	envDNSNameKey   string
+type albOrCdnDescriber struct {
+	svc              string
+	env              string
+	svcDescriber     ecsDescriber
+	envDescriber     envDescriber
+	initLBDescriber  func(string) (lbDescriber, error)
+	albCfnOutputName string // The DNS name for the public or private ALB.
```
dannyrandall:
```suggestion
	albCFNOutputName string // The DNS name for the public or private ALB.
```
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

internal/pkg/describe/uri.go
```diff
@@ -166,23 +166,23 @@ func (d *BackendServiceDescriber) URI(envName string) (URI, error) {
...
 	for _, res := range resources {
 		if res.LogicalID == svcStackResourceALBTargetGroupLogicalID {
-			albDescr := &albDescriber{
-				svc:             d.svc,
-				env:             envName,
-				svcDescriber:    svcDescr,
-				envDescriber:    envDescr,
-				initLBDescriber: d.initLBDescriber,
-				envDNSNameKey:   envOutputInternalLoadBalancerDNSName,
+			albOrCdnDescr := &albOrCdnDescriber{
```
dannyrandall:
```suggestion
			albOrCDNDescr := &albOrCDNDescriber{
```
same here😊

iamhopaul123 (approved):

internal/pkg/template/templates/environment/cf.yml
CaptainCarpensir: CloudFront is an environment resource, but there's no reason to create it until there's an existing ALB to put it in front of, which is dictated by a deploy service. That's we we have the check `{{- if .CDNConfig}}` to check if CF is enabled, and then the condition `CreateALB` to make sure the resource is then only created once it's needed. Only when both of those are true will the resource actually be created.

dannyrandall: oh 🤦 i saw that 🙄

internal/pkg/describe/uri.go
```diff
@@ -265,35 +269,35 @@ func (d *albDescriber) uri() (albURI, error) {
...
-	return albURI{
+	return accessURI{
 		HTTPS:    httpsEnabled,
 		DNSNames: dnsNames,
 		Path:     path,
 	}, nil
 }
 
-func (d *albDescriber) bestEffortRemoveEnvDNSName(albURI albURI) albURI {
+func (d *uriDescriber) bestEffortRemoveEnvDNSName(accessURI accessURI) accessURI {
```
dannyrandall:
```suggestion
func (d *uriDescriber) bestEffortRemoveALBDNSName(accessURI accessURI) accessURI {
```

small nit, maybe we should change this func name along with the struct key change

dannyrandall (approved):

PR #3760, opened 2022-07-13
chore: embed the raw env manifest instead of yaml.Marshal

efekarakus:


internal/pkg/deploy/cloudformation/stack/env_integration_test.go
```diff
@@ -102,10 +103,14 @@ type: Environment
 			envStack := stack.NewEnvStackConfig(tc.input)
 			actual, err := envStack.Template()
 			require.NoError(t, err, "serialize template")
+			fmt.Println(actual)
```
paragbhingre: Can we remove this print statement?

iamhopaul123 (approved):

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
efekarakus: 😂 done!

dannyrandall (approved):

Issue #3672, opened 2022-06-15
task run `env_file` support


cdunn:
Services have the ability to set env_file's
```
environments:
 production:
   env_file: production.env
...
```

Is there any way to translate this to `task run` besides enumerating them inline with `--env-vars`?

Lou1415926:
Hello @cdunn ! This would be a feature request for us. Currently `task run` doesn't support that and there isn't a great workaround :( 

If you have a service, and you want to spin up a task basically copy-and-pasting the service's main container set up, you could run `copilot task run --generate-cmd`. The environment variables used in the service's main container should be populated in the command that's generated by `copilot task run --generate-cmd`.

dannyrandall:
This feature is now released in v1.21.0! https://github.com/aws/copilot-cli/releases/tag/v1.21.0 🎉

cdunn:
@dannyrandall awesome! super useful, thanks

PR #3762, opened 2022-07-14
chore(cfn): Add support for EnvFileARN in task templates

rclinard-amzn:
This is just a small and straightforward PR that adds support for passing the EnvFileARN in task templates, but does not actually add any user-facing options to add env files. Supporting the actual `--env-file` option will be for a future PR.

Relates to https://github.com/aws/copilot-cli/issues/3672 (part one in a series of PRs)


internal/pkg/template/templates/task/cf.yml
```diff
@@ -89,9 +99,46 @@ Resources:
...
         - !Sub 'arn:${AWS::Partition}:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy'
-
-      {{- if or .SSMParamSecrets .SecretsManagerSecrets}}
       Policies:
+        - !If
+          # Optional IAM permission required by ECS task def env file
+          # https://docs.aws.amazon.com/AmazonECS/latest/developerguide/taskdef-envfiles.html#taskdef-envfiles-iam
+          # Example EnvFileARN: arn:aws:s3:::stackset-demo-infrastruc-pipelinebuiltartifactbuc-11dj7ctf52wyf/manual/1638391936/env
+          - HasEnvFile
+          - PolicyName:
```
paragbhingre, line 109: How about keeping this PolicyName consistent with other policyNames for example
 workloads?
For example the one in [executionrole.yml](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/workloads/partials/cf/executionrole.yml#L14)

internal/pkg/deploy/cloudformation/stack/task.go
```diff
@@ -27,6 +27,7 @@ const (
 	taskTaskRoleParamKey       = "TaskRole"
 	taskCommandParamKey        = "Command"
 	taskEntryPointParamKey     = "EntryPoint"
+	taskEnvFileARN             = "EnvFileARN"
```
paragbhingre:
```suggestion
	taskEnvFileARNParamKey             = "EnvFileARN"
```

internal/pkg/deploy/cloudformation/stack/task.go
```diff
@@ -115,6 +116,10 @@ func (t *taskStackConfig) Parameters() ([]*cloudformation.Parameter, error) {
 			ParameterKey:   aws.String(taskEntryPointParamKey),
 			ParameterValue: aws.String(strings.Join(t.EntryPoint, ",")),
 		},
+		{
+			ParameterKey:   aws.String(taskEnvFileARN),
```
paragbhingre:
```suggestion
			ParameterKey:   aws.String(taskEnvFileARNParamKey),
```

paragbhingre (commented):
This looks amazing 🎉  Great job!
Just suggested couple nits.

internal/pkg/template/templates/task/cf.yml
rclinard-amzn: I considered this but AppName and other parameters weren't available so I removed them from the substitution. Should I add them as parameters then?

paragbhingre:
Can we also add the issue number in the description of this PR so that other folks can get the better idea about the requirement?

internal/pkg/template/templates/task/cf.yml
paragbhingre: Hmm, right. App and Env will not be available in the `--default` scenario so whatever you have done is fine.

internal/pkg/template/templates/task/cf.yml
```diff
@@ -89,9 +99,46 @@ Resources:
...
+                        - !Ref AWS::Partition
+                        - ":s3:::"
+                        - !Select [
+                          0,
+                          !Split [
+                            "/",
+                            !Select [ 5, !Split [ ":", !Ref EnvFileARN ] ],
+                          ],
+                        ]
+          - !Ref AWS::NoValue
```
iamhopaul123, line 141: Is this to guarantee the correctness when there's no policies under `Policies`?

iamhopaul123 (commented):
Awesome! LGTM mostly with just one question.

internal/pkg/template/templates/task/cf.yml
rclinard-amzn: I mostly just copied this but I believe it is needed since we're passing the EnvFileARN as a parameter instead of using a template (so now the Policies will always be written since the EnvFileARN could be changed without the template content itself changing). Thanks for the the review!

rclinard-amzn:
Thanks for the feedback! I think all comments have been addressed so I'd really appreciate a re-review!

iamhopaul123 (approved):
Looks good!

paragbhingre (approved):
Ship it 🎉

PR #3758, opened 2022-07-13
chore: make `env upgrade` a no op

efekarakus:
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the Apache 2.0 License.

PR #3763, opened 2022-07-14
chore: retrofit `--allow-vpc-ingress` flag to manifest

Lou1415926:


internal/pkg/deploy/cloudformation/env_test.go
```diff
@@ -5,352 +5,16 @@ package cloudformation
...
 	awscfn "github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/copilot-cli/internal/pkg/aws/cloudformation"
-	"github.com/aws/copilot-cli/internal/pkg/deploy"
 	"github.com/aws/copilot-cli/internal/pkg/deploy/cloudformation/mocks"
-	"github.com/aws/copilot-cli/internal/pkg/template"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
 )
 
-func TestCloudFormation_UpgradeEnvironment(t *testing.T) {
```
Lou1415926: These shouldn't be part of the change after I'm able to rebase on #3758.

internal/pkg/manifest/env.go
```diff
@@ -336,11 +336,20 @@ func (cfg publicHTTPConfig) IsEmpty() bool {
 }
 
 type privateHTTPConfig struct {
-	InternalALBSubnets []string `yaml:"subnets,omitempty"`
-	Certificates       []string `yaml:"certificates,omitempty"`
+	InternalALBSubnets   []string             `yaml:"subnets,omitempty"`
+	Certificates         []string             `yaml:"certificates,omitempty"`
+	SecurityGroupsConfig securityGroupsConfig `yaml:"security_groups,omitempty"`
```
paragbhingre, line 343: Just wanted to bring to your notice that my PR to add security_group to env manifest also have similar [secuirty_group](https://github.com/aws/copilot-cli/pull/3749/files#diff-4c63a9ce04347fa71ffd19988ac84f3eca7aacf36c414bb9e8f95c011b0f622dR101) field in the `vpc`. 
I was just concerned about having the same naming with just extra `s` at the end of `security_groups` will be little confusing for the customers.  What do you think about this?

paragbhingre (commented):
I just have one concern otherwise looks good to me.

internal/pkg/manifest/env.go
Lou1415926: This is going to read
```
http:
   private:
      security_groups: # configs for the sg used on internal ALB
```
Whereas yours iirc would be
```
security_group:
   # environmentsg configs
```
so I guess we're probably fine!!

paragbhingre: Yes you are correct. I was just making sure that 2 almost similar parameters shouldn't be confusing to the customers. But as you said they read differently in different sections should be okay.

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -105,7 +105,7 @@ func (e *EnvStackConfig) Template() (string, error) {
 		PrivateImportedCertARNs:  e.importPrivateCertARNs(),
 		VPCConfig:                e.vpcConfig(),
 		CustomInternalALBSubnets: e.internalALBSubnets(),
-		AllowVPCIngress:          e.in.AllowVPCIngress, // TODO(jwh): fetch AllowVPCIngress from Manifest or SSM.
```
iamhopaul123, line 108: Do we need to remove `AllowVPCIngress` from `e.in`?

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-imported-certs-observability.yml
```diff
@@ -6,7 +6,7 @@ Metadata:
     name: test
     type: Environment
     observability: {container_insights: true}
-    http: {public: {certificates: [cert-1, cert-2]}}
+    http: {public: {certificates: [cert-1, cert-2]}, private: {security_groups: {allow_vpc_ingress: true}}}
```
iamhopaul123: should we do
```yaml
http:
  private:
    security_groups:
      ingress:
        allow_vpc: true
```

internal/pkg/manifest/env.go
iamhopaul123: This is related to https://github.com/aws/copilot-cli/pull/3737/files#diff-4c63a9ce04347fa71ffd19988ac84f3eca7aacf36c414bb9e8f95c011b0f622dR327 . it is just `from_cdn` won't be valid for `private`.

```yaml
http:
  public:
    security_groups:
      ingress:
        from_cdn: true
```

internal/pkg/deploy/cloudformation/stack/env.go
paragbhingre: +1

Lou1415926: we can't because we need to feed from `e.in` to manifest for `env show --manifest`

iamhopaul123 (approved):

paragbhingre (approved):
Looks good to me 🎉

PR #3767, opened 2022-07-14
chore: make `env show --manifest > manifest.yml` not show null values

dannyrandall:
…if we are generating from ssm

also don't use flow style!


efekarakus (approved):
😍 ty Danny!

iamhopaul123 (approved):

PR #3769, opened 2022-07-14
chore(templates): determine http listener rule priority base on http listener

Lou1415926:
The problem with using https listener to determine the priority for `HTTPListenerRuleWithDomain` (a listener rule that's imposed on the HTTP listener, not HTTPS) is, when a service converts from HTTP to using HTTPS, the priority of `HTTPListenerRuleWithDomain` is not necessarily valid.


internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
```diff
@@ -102,7 +115,7 @@ HTTPListenerRuleWithDomain:
     {{- else}}
     ListenerArn: !GetAtt EnvControllerAction.HTTPListenerArn
     {{- end}}
-    Priority: !GetAtt HTTPSRulePriorityAction.Priority # Same priority as HTTPS Listener
+    Priority: !GetAtt HTTPRulePriorityAction.Priority # Same priority as HTTPS Listener
```
dannyrandall:
```suggestion
    Priority: !GetAtt HTTPRulePriorityAction.Priority
```

dannyrandall (approved):
great fix!

efekarakus (approved):
Huh how come this didn't cause a problem before?

iamhopaul123:
> Huh how come this didn't cause a problem before?

Because there's no way for them to switch from http <-> https

Issue #3298, opened 2022-02-18
Cannot deploy service due to error "Response object is too long"


khageshsaini:
Hi,

We have close to 100 services in an application. We are using copilot to deploy these in an environment. We are not able to deploy new services and recieve the error "Response object is too long" in **Resources** section with **Logical ID - EnvControllerAction**. 

I am suspecting this is due to this [Cloudformation restriction ](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-responses.html) on size of custom resource response objects.

Few Observations:
 - EnabledFeatures Key in the Output section for env stack is a very long string when deploying large number of services. The string even contains duplicates. Not sure if we can get rid of it.
- We can shorten the size of each service name but we may eventually hit the size restrictions.

Please let us know the resolution or a workaround for the time being.

efekarakus:
Hi @khageshsaini !

Apologies that you ran into the CloudFormation (CFN) size limits 🙇  
I'll provide an explanation first for the `EnabledFeatures` output and other CFN parameters in the environment stack in order to explain the workarounds.

## Background
Today, Copilot environments have the following CFN Parameters:
* `ALBWorkloads`: comma-separated service names that require a load balancer
* `EFSWorkloads`: comma-separated service names that require a file system
* `NATWorkloads`: comma-separated service names that are placed in private subnets

When a service is deployed, Copilot will populate these parameters in order to create the resource. For example, if a "Load Balanced Web Service" named "frontend" is deployed, then the `ALBWorkloads` parameter has the value `frontend`. If another "Load Balanced Web Service" named "api" is deployed, then the `ALBWorkloads` has the value `frontend,api`.
 This method ensures that when "frontend" is deleted the load balancer is not removed. Only when both services are deleted, the `ALBWorkloads` parameter becomes the empty string resulting in the load balancer to be deleted in order to save cost and not pay for idle resources.

Finally, the `EnabledFeatures` output is just a concatenation of `${ALBWorkloads},${EFSWorkloads},${NATWorkloads}`. This is why you're seeing duplicates, it's expected if you have a service that requires both a load balancer and NAT gateways for example. The output `EnabledFeatures` is necessary because CloudFormation won't update a stack unless a resource or output is modified, this is our little hack to ensure that even if a shared resource already exists we can keep updating the Parameters.

## Mitigations

### ⭐️ [preferred] Option 1: Update environment parameters with a single value

[...]

PR #3771, opened 2022-07-15
chore(templates): trigger env controller when there is an alias change

Lou1415926:
In addition to the title, also renaming the logical ID of the HTTP rule priority action that is used for `HTTPListenerRuleWithDomain`. This listener rule shouldn't share the same priority action as `HTTPListenerRule`.

### DIFF
--- a/internal/pkg/template/templates/environment/cf.yml
+++ b/internal/pkg/template/templates/environment/cf.yml
@@ -634,7 +634,7 @@ Outputs:
       Name: !Sub ${AWS::StackName}-SubDomain
 {{- end}}
   EnabledFeatures:
-    Value: !Sub '${ALBWorkloads},${InternalALBWorkloads},${EFSWorkloads},${NATWorkloads}'
+    Value: !Sub '${ALBWorkloads},${InternalALBWorkloads},${EFSWorkloads},${NATWorkloads},${Aliases}'
     Description: Required output to force the stack to update if mutating feature params, like ALBWorkloads, does not change the template.
   ManagedFileSystemID:
     Condition: CreateEFS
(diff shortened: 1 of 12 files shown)


dannyrandall (approved):

iamhopaul123 (approved):

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -634,7 +634,7 @@ Outputs:
       Name: !Sub ${AWS::StackName}-SubDomain
 {{- end}}
   EnabledFeatures:
-    Value: !Sub '${ALBWorkloads},${InternalALBWorkloads},${EFSWorkloads},${NATWorkloads}'
+    Value: !Sub '${ALBWorkloads},${InternalALBWorkloads},${EFSWorkloads},${NATWorkloads},${Aliases}'
```
efekarakus, line 637: A long time ago, we had a chat w @iamhopaul123 because I was concerned that adding `Aliases` to Outputs will cause us to quickly hit the custom resource response limit of 4,096 bytes: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-responses.html for the `EnvControllerAction`.
This is because the `EnvControllerAction`'s response contains all the Outputs of the env stack.

This has happened in the past: https://github.com/aws/copilot-cli/issues/3298 where the client deployed over 100 services in an environment.

Can I ask two action items please 🥺:

1. Can we validate if an output value can be longer than 4096 bytes? To ensure that v large environments can continue being deployed.
2. Can we change the code for the  `EnvControllerAction` to ignore the `EnabledFeatures` output from the stack? This should make it v likely that we won't hit the response output limit for custom resources.

PR #3773, opened 2022-07-15
perf(cloudwatchlogs): improve filterStringSliceByPrefix

iamhopaul123:
Improve my crappy implementation 🤦


internal/pkg/aws/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -158,17 +157,48 @@ func initGetLogEventsInput(opts LogEventsOpts) *cloudwatchlogs.GetLogEventsInput
...
+		for _, char := range str {
+			val, ok := node.children[char]
+			if !ok {
+				break
 			}
+			if val.hasPrefix {
+				matches = append(matches, str)
+				break
+			}
+			node = val
```
Lou1415926: nit: maybe we can make this a method of `trieNode`?

internal/pkg/aws/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -158,17 +157,48 @@ func initGetLogEventsInput(opts LogEventsOpts) *cloudwatchlogs.GetLogEventsInput
...
-		for _, prefix := range prefixes {
-			if strings.HasPrefix(candidate, prefix) {
-				m[candidate] = true
+func filterStringSliceByPrefix(all, prefixes []string) []string {
+	trie := buildTrie(prefixes)
+	var matches []string
+	for _, str := range all {
+		node := trie
+		for _, char := range str {
+			val, ok := node.children[char]
```
rclinard-amzn:
```suggestion
			child, ok := node.children[char]
```
(inspired by this document Efe sent to me last week: https://a-nickels-worth.blogspot.com/2016/04/a-guide-to-naming-variables.html#orgheadline10)

internal/pkg/aws/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -158,17 +157,48 @@ func initGetLogEventsInput(opts LogEventsOpts) *cloudwatchlogs.GetLogEventsInput
 
 // Example: if the prefixes is []string{"a"} and all is []string{"a", "b", "ab"}
 // then it returns []string{"a", "ab"}.
```
rclinard-amzn: The edge case of an empty prefix string isn't handled in this implementation, and this is a change from the previous implementation, but since this is a private method that doesn't deal with the edge case it doesn't matter too much. It's worth documenting though in case the method gets copied / moved to a more general location. 
```suggestion
// then it returns []string{"a", "ab"}. Empty string prefixes are not supported.
```

internal/pkg/aws/cloudwatchlogs/cloudwatchlogs.go
rclinard-amzn: Seconding - trieNode.find seems appropriate

internal/pkg/aws/cloudwatchlogs/cloudwatchlogs.go
iamhopaul123: good catch!

rclinard-amzn (approved):
Looks good to me! 👍

dannyrandall (approved):

PR #3774, opened 2022-07-15
chore(custom-resources): ignore certain outputs from env stack in env controller

Lou1415926:
Addressing https://github.com/aws/copilot-cli/pull/3771#discussion_r922325007

### DIFF
--- a/cf-custom-resources/lib/env-controller.js
+++ b/cf-custom-resources/lib/env-controller.js
@@ -16,6 +16,9 @@ const updateStackWaiter = {
 
 const AliasParamKey = "Aliases";
 
+// Per the doc at https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-responses.html
+// The size of the response body should not exceed 4096 bytes.
+// Therefore, we should ignore any outputs that we don't need. g
 let ignoredEnvOutputs = new Set(["EnabledFeatures", "LastForceDeployID"]);
 
 /**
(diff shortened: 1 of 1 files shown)


iamhopaul123 (approved):

cf-custom-resources/lib/env-controller.js
```diff
@@ -308,7 +310,9 @@ const updateAliases = function (cfnAliases, workload, aliases) {
 const getExportedValues = function (stack) {
   const exportedValues = {};
   stack.Outputs.forEach((output) => {
-    exportedValues[output.OutputKey] = output.OutputValue;
+    if (!ignoredEnvOutputs.has(output.OutputKey)) {
+      exportedValues[output.OutputKey] = output.OutputValue;
+    }
```
efekarakus: nit:
```suggestion
    if (ignoredEnvOutputs.has(output.OutputKey)) {
      return
    }
    exportedValues[output.OutputKey] = output.OutputValue;
```

cf-custom-resources/lib/env-controller.js
```diff
@@ -16,6 +16,8 @@ const updateStackWaiter = {
 
 const AliasParamKey = "Aliases";
 
+let ignoredEnvOutputs = new Set(["EnabledFeatures", "LastForceDeployID"]);
```
efekarakus, line 22: Can you add a comment w a link to the 4096 bytes limit and why we're ignoring it here? 🥺

efekarakus (commented):
Awesome, ty!

efekarakus (approved):

cf-custom-resources/lib/env-controller.js
```diff
@@ -16,6 +16,11 @@ const updateStackWaiter = {
 
 const AliasParamKey = "Aliases";
 
+// Per the doc at https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-responses.html
+// The size of the response body should not exceed 4096 bytes.
+// Therefore, we should ignore any outputs that we don't need. g
```
dannyrandall:
```suggestion
// the size of the response body should not exceed 4096 bytes.
// Therefore, we should ignore any outputs that we don't need.
```

dannyrandall (approved):

PR #3779, opened 2022-07-15
chore(templates): fix the condition of `CustomDomainAction`

Lou1415926:
As well as enhancing the integration test that makes sure all env-controller managed parameters are added as a feature.

### DIFF
--- a/internal/pkg/template/templates/environment/partials/custom-resources.yml
+++ b/internal/pkg/template/templates/environment/partials/custom-resources.yml
@@ -35,7 +35,7 @@ HTTPSCert:
 CustomDomainAction:
   Metadata:
     'aws:copilot:description': 'Add an A-record to the hosted zone for the domain alias'
-  Condition: HasAliases
+  Condition: ManagedAliases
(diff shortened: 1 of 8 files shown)


internal/pkg/template/templates/environment/partials/custom-resources.yml
```diff
@@ -35,7 +35,7 @@ HTTPSCert:
 CustomDomainAction:
   Metadata:
     'aws:copilot:description': 'Add an A-record to the hosted zone for the domain alias'
-  Condition: HasAliases
+  Condition: ManagedAliases
```
efekarakus, line 38: is it possible that we still want this action if one LBWS has no `alias` but another LBWS has an `alias`?

Lou1415926: then `Aliases` would be non-empty and `HasAliases` would render true, as long as the dns is delegated (i.e. `copilot app init --domain` was executed

efekarakus (approved):

iamhopaul123 (approved):

PR #3792, opened 2022-07-18
chore: e2e app runner `svc logs` fix

dannyrandall:
Fixes an e2e test failure caused by filtering app runner log streams to have to begin with `copilot/`, but they don't 😊 

### DIFF
--- a/internal/pkg/logging/workload.go
+++ b/internal/pkg/logging/workload.go
@@ -173,7 +173,9 @@ func (s *WorkloadClient) WriteLogEvents(opts WriteLogEventsOpts) error {
 		LogStreamLimit:      opts.LogStreamLimit,
 	}
 
-	logEventsOpts.LogStreamPrefixFilters = s.logStreams(opts.TaskIDs)
+	if opts.TaskIDs != nil {
+		logEventsOpts.LogStreamPrefixFilters = s.logStreams(opts.TaskIDs)
+	}
 
 	for {
 		logEventsOutput, err := s.eventsGetter.LogEvents(logEventsOpts)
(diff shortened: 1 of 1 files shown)


internal/pkg/logging/workload.go
```diff
@@ -173,7 +173,9 @@ func (s *WorkloadClient) WriteLogEvents(opts WriteLogEventsOpts) error {
 		LogStreamLimit:      opts.LogStreamLimit,
 	}
 
-	logEventsOpts.LogStreamPrefixFilters = s.logStreams(opts.TaskIDs)
+	if s.logStreamNamePrefix != "" {
```
Lou1415926, line 178: Since app runner client is the only workload client that doesn't have `logStreamNamePrefix` set up, this chain holds true: `logStreamNamePrefix == "" ➡️  it's app runner ➡️  don't filter by "copilot" prefix`.

I think this totally makes sense as a quick workaround! However we should probably refactor it later so that it works without assuming `logStreamNamePrefix == "" ➡️  it's app runner`. This is effectively true but is not necessarily true in terms of logic, which may make it confusing to read and maintain in the future. Do you think we want to add a `TODO` comment on this, explain why there is a `if` block and what we need to do about it in the future?

internal/pkg/logging/workload.go
```diff
@@ -173,7 +173,9 @@ func (s *WorkloadClient) WriteLogEvents(opts WriteLogEventsOpts) error {
...
+		logEventsOpts.LogStreamPrefixFilters = s.logStreams(opts.TaskIDs)
+	}
```
efekarakus, line 180: I think our mistake is that we are returning a `*WorkloadClient` for both apprunner and ecs causing the type to have a mix of fields that don't make sense for one service type 💭 .

efekarakus (approved):

Lou1415926 (approved):

internal/pkg/logging/workload.go
dannyrandall: great point - i added a TODO saying what this does along with @efekarakus's suggestion to have a separate logging client for App Runner services.

PR #3794, opened 2022-07-18
feat: enable job logs command

bvtujo:
- chore: update cloudwatchlogs to allow limiting log streams fetched and custom prefixes
- chore: add logic for fetching state machine logs to workload
- chore: add unit tests for workload logging pkg
- chore: make things work for ci
- chore: address feedback
- feat: add job logs logic
- chore: add logic in underlying libraries for job logs (#3757)
- fix logic bugs in log stream sorting
- chore: make IncludeStateMachine a logging option, not a struct member
- feat: enable job logs command and add tests

Adds job logs command and fixes a bug which randomized the sort order of log streams.


internal/pkg/aws/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -185,16 +195,13 @@ func initGetLogEventsInput(opts LogEventsOpts) *cloudwatchlogs.GetLogEventsInput
 // Example: if the prefixes is []string{"a"} and all is []string{"a", "b", "ab"}
 // then it returns []string{"a", "ab"}.
 func filterStringSliceByPrefix(all, prefixes []string) (res []string) {
-	m := make(map[string]bool)
 	for _, candidate := range all {
 		for _, prefix := range prefixes {
 			if strings.HasPrefix(candidate, prefix) {
-				m[candidate] = true
+				res = append(res, candidate)
```
Lou1415926: I see this is getting sort of addressed in #3773!

internal/pkg/cli/job_logs.go
```diff
@@ -263,7 +278,13 @@ Displays logs from specific task IDs.
 	cmd.Flags().BoolVar(&vars.follow, followFlag, false, followFlagDescription)
 	cmd.Flags().DurationVar(&vars.since, sinceFlag, 0, sinceFlagDescription)
 	cmd.Flags().IntVar(&vars.limit, limitFlag, 0, limitFlagDescription)
+	cmd.Flags().IntVar(&vars.last, lastFlag, 6, lastFlagDescription)
```
Lou1415926: why do we want 6 for the default?

internal/pkg/cli/job_logs.go
```diff
@@ -151,38 +157,43 @@ func (o *jobLogsOpts) Ask() error {
...
 
-	deployedJob, err := o.sel.DeployedJob(jobLogNamePrompt, jobLogNameHelpPrompt, o.appName, selector.WithEnv(o.envName), selector.WithName(o.name))
+	err := o.logsSvc.WriteLogEvents(logging.WriteLogEventsOpts{
+		Follow:                  o.follow,
+		Limit:                   limit,
+		EndTime:                 o.endTime,
+		StartTime:               o.startTime,
+		TaskIDs:                 o.taskIDs,
+		OnEvents:                eventsWriter,
+		LogStreamLimit:          logStreamLimit,
```
Lou1415926, line 187: I realized I've never noticed this parameter existed before :homer-disappearing !  If `logStreamLimit`, is it guaranteed that the one and only one log stream that gets fetched is the log stream that's produced by the latest call? Could it be the log stream that's produced by the second lastest call? (sorry for the repetition of words 😂 )

bvtujo: It's the most recent that Cloudwatch has to offer; I don't know that we can control the delay between a container coming up and the log stream for that container appearing. But I assume that there will be few times when someone needs to immediately view the logs from the last execution less than 10 seconds after the job is triggered, and this case can be covered by
```sh
copilot job run && copilot job logs --follow
``` 
or eventually
```sh
copilot job run --follow
```

site/content/docs/commands/job-logs.en.md
```diff
@@ -0,0 +1,62 @@
+# job logs
+```console
+$ copilot job logs
+```
+
+## What does it do?
+
+`copilot job logs` displays the logs of a deployed service.
```
rclinard-amzn: This might need to be changed in the code as well
```suggestion
`copilot job logs` displays the logs of a deployed job.
```

site/content/docs/commands/job-logs.en.md
```diff
@@ -0,0 +1,62 @@
...
+  -e, --env string              Name of the environment.
+      --follow                  Optional. Specifies if the logs should be streamed.
+  -h, --help                    help for logs
+      --include-state-machine   Optional. Include logs from the state machine executions.
+      --json                    Optional. Output in JSON format.
+      --last int                Optional. The number of executions of the scheduled job for which
+                                logs should be shown. (default 3)
+      --limit int               Optional. The maximum number of log events returned. Default is 10
+                                unless any time filtering flags are set.
+  -n, --name string             Name of the service.
```
rclinard-amzn: This might need to be changed in the code as well
```suggestion
  -n, --name string             Name of the job.
```

site/content/docs/commands/job-logs.en.md
```diff
@@ -0,0 +1,62 @@
...
+                                Defaults to all logs. Only one of end-time / follow may be used.
+  -e, --env string              Name of the environment.
+      --follow                  Optional. Specifies if the logs should be streamed.
+  -h, --help                    help for logs
+      --include-state-machine   Optional. Include logs from the state machine executions.
+      --json                    Optional. Output in JSON format.
+      --last int                Optional. The number of executions of the scheduled job for which
+                                logs should be shown. (default 3)
+      --limit int               Optional. The maximum number of log events returned. Default is 10
+                                unless any time filtering flags are set.
```
rclinard-amzn, line 24: Out of curiosity, is the default actually 10? It seems to be set to 0 in the code but I haven't read the code close enough to figure out whether it gets defaulted to 10 later.

internal/pkg/cli/cli.go
```diff
@@ -23,8 +23,8 @@ import (
 )
 
 const (
-	svcAppNamePrompt     = "Which application does your service belong to?"
-	svcAppNameHelpPrompt = "An application groups all of your services and jobs together."
```
iamhopaul123, line 27: 🤦

internal/pkg/aws/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -101,15 +101,16 @@ func (c *CloudWatchLogs) logStreams(logGroup string, logStreamLimit int, logStre
 			break
 		}
 	}
+	logStreamNames = truncateStreams(logStreamLimit, logStreamNames)
 
-	return truncateStreams(logStreamLimit, logStreamNames), nil
+	// Make sure that the streams are returned with the latest events last so they're printed to terminal in chronological order.
+	return reverseStreams(logStreamNames), nil
```
iamhopaul123: Hmm I think I would expect last events to be first for logs 🤔

internal/pkg/aws/cloudwatchlogs/cloudwatchlogs.go
iamhopaul123: This function is updated now so that we can deal with the matching more efficiently.

internal/pkg/cli/job_logs.go
```diff
@@ -151,38 +157,43 @@ func (o *jobLogsOpts) Ask() error {
...
+	if o.last != 0 {
+		logStreamLimit = o.last
 	}
 
-	if o.name != "" {
-		if _, err := o.configStore.GetJob(o.appName, o.name); err != nil {
-			return err
-		}
+	if o.includeStateMachineLogs {
+		logStreamLimit *= 2
```
iamhopaul123: Is this a safe assumption that we can make 💭 If so we might need to add some comments explaining the reason. Also I feel like we should deal with the logic within the logging pkg, because otherwise we are passing duplicated info around.

site/content/docs/commands/job-logs.en.md
iamhopaul123: This is the one https://github.com/aws/copilot-cli/blob/e1083b91cd96a0c3123789ac8638b6e181540ffe/internal/pkg/logging/workload.go#L23

internal/pkg/aws/cloudwatchlogs/cloudwatchlogs.go
bvtujo: I wrote a long digression on why this is necessary, but I guess though it actually doesn't matter. Later on, in `LogEvents()`, we sort all the logs by event time so the order of streams doesn't matter, just that they are the most recent N streams chronologically.

internal/pkg/cli/job_logs.go
bvtujo: Good idea. I guess we can use more jobby abstractions in `logging` since it's not the low-level aws package.

site/content/docs/concepts/jobs.en.md
```diff
@@ -41,4 +41,18 @@ Since Copilot uses CloudFormation under the hood, all the resources created are
 Scheduled Jobs are composed of an AmazonECS Task Definition, Task Role, Task Execution Role, 
 a Step Function State Machine for retrying on failures, and finally an Event Rule to trigger the state machine.
 
- 
\ No newline at end of file
+### Where are my job's logs?
+
+Checking your job logs is easy as well. Running `copilot job logs` will show the most recent logs of your job. You can follow your logs live with the `--follow` flag,
```
Lou1415926: nit:
```suggestion
Checking your job logs is easy as well. Running [`copilot job logs`](../commands/job-log.en.md) will show the most recent logs of your job. You can follow your logs live with the `--follow` flag,
```

internal/pkg/cli/job_logs.go
```diff
@@ -198,54 +205,58 @@ func (o *jobLogsOpts) getTargetEnv() (*config.Environment, error) {
...
   /code $ copilot job logs -n my-job -e test
   Displays logs in the last hour.
   /code $ copilot job logs --since 1h
-  Displays logs from 2006-01-02T15:04:05 to 2006-01-02T15:05:05.
-  /code $ copilot job logs --start-time 2006-01-02T15:04:05+00:00 --end-time 2006-01-02T15:05:05+00:00
-Displays logs from specific task IDs.
+  Displays logs from the last execution of the job.
+  /code $ copilot job logs --last 1
+  Displays logs from specific task IDs.
   /code $ copilot job logs --tasks 709c7eae05f947f6861b150372ddc443,1de57fd63c6a4920ac416d02add891b9
```
Lou1415926: nit: maybe we can update this example as 
```suggestion
/code $ copilot job logs --tasks 709c7ea,1de57fd
```
I took this from the doc - it looks cleaner and is consistent with the doc!

site/content/docs/commands/job-logs.en.md
```diff
@@ -0,0 +1,62 @@
...
+
+```console
+$ copilot job logs --last 4
+```
+
+Displays logs from specific task IDs
+```console
+$ copilot job logs --tasks 709c7ea,1de57fd
+```
+
```
Lou1415926:
```suggestion
Displays logs in real time.
```console
$ copilot job logs --follow
```


```

File: internal/pkg/logging/workload_test.go [hunk 13]
```diff
@@ -165,12 +165,13 @@ firelens_log_router/fcfe4 10.0.0.00 - - [01/Jan/1970 01:01:01] "WARN some warnin
...
 		"success with state machine included": {
 			includeStateMachine: true,
+			last:                1,
 			setupMocks: func(m workloadLogsMocks) {
 				gomock.InOrder(
 					m.logGetter.EXPECT().LogEvents(gomock.Any()).
 						Do(func(param cloudwatchlogs.LogEventsOpts) {
 							require.Equal(t, param.LogStreamPrefixFilters, []string{"copilot/", "states"})
-							require.Equal(t, param.Limit, aws.Int64(10))
+							require.Equal(t, param.LogStreamLimit, 2)
```
Lou1415926: are we still interested in what's passed in `param.Limit`? If so, what do you think of
```go
require.Equal(t, param.LogStreamPrefixFilters, []string{"copilot/", "states"})
require.Equal(t, param,Limit aws.Int(10))
require.Equal(t, param.LogStreamLimit, 2)
```

Lou1415926 (approved):

iamhopaul123 (approved):

internal/pkg/logging/workload_test.go
bvtujo: Good idea; i've added a check here and a new test to ensure that some log stream limit behavior is working properly.

Lou1415926 (approved):
LGTM! Feel free to remove the label when you are ready (I did it but then added it back just in case you wanted to add anything)

PR #3801, opened 2022-07-22
chore: fix broken deploy

iamhopaul123:
Also addresses fb from https://github.com/aws/copilot-cli/pull/3727#pullrequestreview-1032263086


internal/pkg/manifest/validate_env.go
```diff
@@ -216,26 +216,45 @@ func (cfg publicHTTPConfig) Validate() error {
...
+	return s.Ingress.validate()
+}
+
+// validate returns nil if ingress is configured correctly.
+func (i ingress) validate() error {
+	if i.isEmpty() {
+		return nil
+	}
 	return nil
 }
```
rclinard-amzn, line 246: Is this for a future PR? It doesn't seem to do anything right now 🤔

rclinard-amzn (approved):
The changes here look good! I'm having a hard time figuring out what the functional changes were here though since most of the diff seems to be renames & refactors.

internal/pkg/manifest/validate_env.go
iamhopaul123: We have an auditor checking if every field with complex type implements `validate()` or not. Just in case we don't validate our manifest inputs.

internal/pkg/cli/job_deploy.go
```diff
@@ -46,7 +46,7 @@ type deployJobOpts struct {
 	targetApp       *config.Application
 	targetEnv       *config.Environment
 	envSess         *session.Session
-	appliedManifest interface{}
+	appliedManifest manifest.WorkloadManifest
```
Lou1415926: This reminds me that we probably want to rename `manifest.WorkloadManifest` to `manifest.DynamicWorkloadManifest`: the interface now requires the implementation of `Load()`, which makes it dynamic. The reason why I thought of this was because I found myself having a hard time trying to wrap my head around on how `DynamicWorkloadManifest` implements `WorkloadManifest`, while it has a member field that's of type `workloadManifest`  which actually doesn't implement `WorkloadManifest` though spelled the same  😵

internal/pkg/manifest/workload.go
```diff
@@ -64,7 +64,7 @@ type WorkloadManifest interface {
 	Validate() error
 	RequiredEnvironmentFeatures() []string
 	Load(sess *session.Session) error
-	Manifest() interface{}
+	Manifest() any
```
Lou1415926, line 67: what do you think of ⬇️ 
```suggestion
	Manifest() workloadManifest
```

internal/pkg/manifest/workload.go
```diff
@@ -230,27 +230,30 @@ type ImageOverride struct {
...
+type EntryPointOverride stringSliceOrShellString
 
 // CommandOverride is a custom type which supports unmarshalling "command" yaml which
 // can either be of type string or type slice of string.
-type CommandOverride StringSliceOrString
+type CommandOverride stringSliceOrShellString
 
 // UnmarshalYAML overrides the default YAML unmarshalling logic for the EntryPointOverride
 // struct, allowing it to perform more complex unmarshalling behavior.
 // This method implements the yaml.Unmarshaler (v3) interface.
```
Lou1415926: tiny nit:
```suggestion
// struct, allowing it to perform more complex unmarshalling behavior.
// This method implements the yaml.Unmarshaler (v3) interface.
```
```suggestion
// struct, allowing it to be unmarshalled into a string slice or a string.
// This method implements the yaml.Unmarshaler (v3) interface.
```

internal/pkg/manifest/workload.go
```diff
@@ -261,15 +264,15 @@ func (e *EntryPointOverride) ToStringSlice() ([]string, error) {
 // struct, allowing it to perform more complex unmarshaling behavior.
```
Lou1415926, line 264:
```suggestion
// struct, allowing it to be unmarshalled into a string slice or a string.

```

Lou1415926 (approved):
Adding a DNM label in case you'd like to address the comment!

internal/pkg/manifest/workload.go
iamhopaul123: I think it is probably fine to leave as `any` since it really needs the manifest itself rather than the interface. It doesn't make sense for a public method to return a private interface (because the client in the CLI package can't call any of the private methods there). The reason why we need `workloadManifest` in the dynamic struct is because we need to call the method under the hood.

PR #3802, opened 2022-07-22
test: add load to local integ tests

iamhopaul123:


rclinard-amzn (commented):
I'm having a hard time figuring out what the implications of this change are - from what I can see, calling `Load` might result in an EC2 API call in some cases through `dynamicSubnets.load` indirectly being called? Is that known/intended?

iamhopaul123:
> I'm having a hard time figuring out what the implications of this change are - from what I can see, calling `Load` might result in an EC2 API call in some cases through `dynamicSubnets.load` indirectly being called? Is that known/intended?

Your understanding is correct. That's exactly the reason why it was not there before. And the reason why I decided to add it back is because we want to make sure no API calls if we don't set any field that will cause API calls in the manifest (of course if in the test manifest lets say we enable `from_tag` then this local integ test will fail, which is expected).

dannyrandall (approved):

huanjani (approved):

rclinard-amzn:
Looks like this conflicts! It does have two approvals so it'll be automatically merged once they're resolved it looks like.

PR #3804, opened 2022-07-22
feat: Add --env-file flag to `task run`

### LINKED ISSUE #3672: task run `env_file` support

rclinard-amzn:
Fixes https://github.com/aws/copilot-cli/issues/3672


internal/pkg/cli/deploy/svc.go
```diff
@@ -929,7 +929,8 @@ func (d *workloadDeployer) pushEnvFileToS3Bucket(in *pushEnvFilesToS3BucketInput
 		return "", fmt.Errorf("read env file %s: %w", path, err)
 	}
 	reader := bytes.NewReader(content)
-	url, err := in.uploader.Upload(d.resources.S3Bucket, artifactpath.EnvFiles(path, content), reader)
+	// use filepath.Base to prevent cryptic errors in the ecs agent for paths like "..\magic.env"
+	url, err := in.uploader.Upload(d.resources.S3Bucket, artifactpath.EnvFiles(filepath.Base(path), content), reader)
```
efekarakus: nice! should we instead move filepath.Base` into the `artifactpath.EnvFiles` function so that we don't have to update it in multiple locations?

internal/pkg/cli/flag.go
```diff
@@ -278,6 +279,7 @@ Must be either "MySQL" or "PostgreSQL".`
 	taskRoleFlagDescription      = "Optional. The ARN of the role for the task to use."
 	executionRoleFlagDescription = "Optional. The ARN of the role that grants the container agent permission to make AWS API calls."
 	envVarsFlagDescription       = "Optional. Environment variables specified by key=value separated by commas."
+	envFileFlagDescription       = "Optional. A path to an environment variable (.env) file with each line being of the form of VARIABLE=VALUE. Values specified with --env-vars take precedence over --env-file."
```
efekarakus: nit: can we split the description into multiple lines? we typically try to maintain a single line printed to the terminal to be <= 80 chars as that's the default terminal width.

```suggestion
	envFileFlagDescription       = `Optional. A path to an environment variable (.env) file. 
Each line should be of the form of VARIABLE=VALUE. 
Values specified with --env-vars take precedence over --env-file.`
```

internal/pkg/cli/task_run.go
```diff
@@ -400,6 +424,12 @@ func (o *runTaskOpts) Validate() error {
 		}
 	}
 
+	if o.envFile != "" {
+		if filepath.Ext(o.envFile) != envFileExt {
+			return fmt.Errorf("environment file %s specified in --%s must have a %s file extension", o.envFile, envFileFlag, envFileExt)
```
efekarakus, line 425: really nice error messages!

internal/pkg/cli/task_run.go
```diff
@@ -945,6 +981,77 @@ func (o *runTaskOpts) deploy() error {
...
+	region := aws.StringValue(o.sess.Config.Region)
+
+	app, err := o.store.GetApplication(o.appName)
+	if err != nil {
+		return "", fmt.Errorf("get application from store: %w", err)
+	}
+
+	// bootstrap S3 bucket
+	o.spinner.Start(fmt.Sprintf(fmtTaskRunResourcesStart, color.HighlightUserInput(o.appName)))
+	err = o.deployer.AddPipelineResourcesToApp(app, region)
```
efekarakus: 1. If the users are  running `copilot task run` in a Copilot application and environment with `--app` and `--env` won't the S3 bucket already exist? Can we reuse the existing environment bucket in that scenario?

2. It's possible that folks run `copilot task run` outside of a Copilot application with `--default` for example. In that scenario `AppPipelineResourcesToApp` will fail. 
Personally, I think I'm okay shipping this feature and restricting the `--env-file` flag to only work with `--app` and `--env` at first by returning an error in `Validate`. If anybody asks to expand it, we can add it on later.

internal/pkg/cli/task_run.go
```diff
@@ -945,6 +981,77 @@ func (o *runTaskOpts) deploy() error {
...
+	// push env file
+	o.spinner.Start(fmt.Sprintf(fmtTaskRunEnvUploadStart, color.HighlightUserInput(o.envFile)))
+	envFileARN, err := o.pushEnvFileToS3(app, region)
+	if err != nil {
+		o.spinner.Stop(log.Serrorf(fmtTaskRunEnvUploadFailed, color.HighlightUserInput(o.envFile)))
+		return "", err
+	}
+	o.spinner.Stop(log.Ssuccessf(fmtTaskRunEnvUploadComplete, color.HighlightUserInput(o.envFile)))
+
+	o.envFileARN = envFileARN
```
efekarakus: nit: is there any benefit in caching it? it doesn't seem like this method will be called multiple times

internal/pkg/cli/task_run.go
```diff
@@ -945,6 +981,77 @@ func (o *runTaskOpts) deploy() error {
...
+	if err != nil {
+		return "", fmt.Errorf("parse s3 url: %w", err)
+	}
+	// The app and environment are always within the same partition.
+	partition, err := partitions.Region(aws.StringValue(o.sess.Config.Region)).Partition()
+	if err != nil {
+		return "", err
+	}
+	envFileARN := s3.FormatARN(partition.ID(), fmt.Sprintf("%s/%s", bucket, key))
+	return envFileARN, nil
```
dannyrandall: nit
```suggestion
	return s3.FormatARN(partition.ID(), fmt.Sprintf("%s/%s", bucket, key)), nil
```

dannyrandall (commented):
+1 to efe's comments but otherwise looks good!

internal/pkg/cli/task_run.go
rclinard-amzn: This actually is called multiple times! When no pre-built image is provided, it first deploys the task stack to create the repo and everything, then builds the image & pushes it to the ECR repo, and then deploys the task stack again. Since this function is called from deploy(), it'll be called twice in that case otherwise.

internal/pkg/cli/task_run.go
rclinard-amzn: > If the users are running copilot task run in a Copilot application and environment with --app and --env won't the S3 bucket already exist? Can we reuse the existing environment bucket in that scenario?

I thought this was the case, but when I ran in a bare application without any services initialized I found that this was necessary to get things working.

I can also add the warning for that second case after testing tomorrow! 👍

internal/pkg/cli/deploy/svc.go
rclinard-amzn: Done!

internal/pkg/cli/task_run.go
rclinard-amzn: Thanks for taking a look here! The second point was indeed a potential source of bad UX.

1. I couldn't find any other app-wide or env-wide S3 buckets that get created other than the one created with the function that I call here.
2. I added some additional checks here as suggested, though I have to return the error in `Ask` instead:

```
PS C:\Users\rclinard\Projects\testing\magic-test> C:\Users\rclinard\Projects\copilot-cli\bin\local\copilot.exe task run --env-file .\magic.env --default
Couldn't find any applications in this region and account. Try initializing one with `copilot app init`
X no S3 bucket available to hold environment file .\magic.env specified in --env-file: ask for application: no apps found
```

internal/pkg/cli/task_run.go
```diff
@@ -572,14 +602,20 @@ func (o *runTaskOpts) Ask() error {
...
 		if err := o.askAppName(); err != nil {
 			return err
 		}
 		if err := o.askEnvName(); err != nil {
 			return err
 		}
+	} else if o.envFile != "" {
+		if err := o.askAppName(); err != nil {
+			return fmt.Errorf("no S3 bucket available to hold environment file %s specified in --%s: %w", o.envFile, envFileFlag, err)
+		}
```
efekarakus: Should we immediately error here instead of asking for the app name?

My understanding from reading is that if the user specifies `task run --default --env-file .env` we will ignore that they provided `--default` and ask for the app name. That might be surprising for folks:

```go
return fmt.Errorf("--%s is only supported for tasks that can run in an environment with --%s and --%s", envFileFlag, appFlag, envFlag)
```

internal/pkg/cli/task_run.go
efekarakus: > I thought this was the case, but when I ran in a bare application without any services initialized I found that this was necessary to get things working.

I might be misunderstanding but after running `copilot env init` we call `cfn.AddEnvToApp` and a stackset instance is created with the `PipelineBuiltArtifactBucket` resource in it.

So we still shouldn't need to call `AddPipelineResourcesToApp` here if an env is initialized 🤔 .

internal/pkg/cli/task_run.go
paragbhingre: +1, When I try running task run with default settings I get error in get application.

internal/pkg/template/artifactpath/artifactpath.go
```diff
@@ -39,7 +40,8 @@ func CFNTemplate(key string, content []byte) string {
 // EnvFiles returns the path to store an env file artifact with sha256 of the content..
 // Example: manual/env-files/key/sha.env.
 func EnvFiles(key string, content []byte) string {
-	return path.Join(s3ArtifactDirName, s3ArtifactEnvFilesDirName, key, fmt.Sprintf("%x.env", sha256.Sum256(content)))
+	// use filepath.Base to prevent cryptic errors in the ecs agent for paths like "..\magic.env"
```
paragbhingre, line 50: Nice 👍🏼

internal/pkg/cli/task_run.go
rclinard-amzn: I think I see what you are saying. I might have only run `copilot app init` which doesn't imply `copilot env init`. So you're proposing that `task run --env-file` just require that both an app & env have already been initialized instead of trying to set up the pipeline resources itself?

internal/pkg/cli/task_run.go
rclinard-amzn: That error message might be unintuitive since if you don't specify `--app` and `--env` it'll ask anyways - how about something like:

```go
return fmt.Errorf("--%s is not compatible with --%s because it requires an S3 bucket provided from an app & environment", envFileFlag, defaultFlag)
```

Or, should I just use `cmd.MarkFlagsMutuallyExclusive(envFileFlag, defaultFlag)`?

internal/pkg/cli/task_run.go
efekarakus: yup that's right!

internal/pkg/cli/task_run.go
efekarakus: > Or, should I just use cmd.MarkFlagsMutuallyExclusive(envFileFlag, defaultFlag)?

sgtm!

rclinard-amzn:
I though about this more and I realized that there's a much nicer way to approach this. Instead of all the fragility of relying on an app-associated or env-associated S3 bucket, I can instead create an S3 bucket as part of the task, and then use the same approach as is used for ECR repo creation / uploading a built image to ECR. It'll also mean that env files for the task will get cleaned up & deleted when the task gets deleted. The only downside is that it'll require the double-deploy like what is needed with built images, but it seems like CFN handles that just fine so it's not too huge of a concern. I'll look into implementing that instead.

rclinard-amzn:
Okay! The code has been reworked quite a bit now, but this should completely fix the issue where uploading env files wouldn't work outside of an app/env, instead of trying to make that into an error case.

internal/pkg/cli/task_delete.go
```diff
@@ -389,10 +394,22 @@ func (o *deleteTaskOpts) clearECRRepository() error {
 	err = o.newImageRemover(defaultSess).ClearRepository(ecrRepoName)
 	if err != nil {
 		o.spinner.Stop(log.Serrorln("Error emptying ECR repository."))
-		return fmt.Errorf("clear ECR repository for task %s: %w", o.name, err)
+		return fmt.Errorf("empty ECR repository for task %s: %w", o.name, err)
+	}
+
+	o.spinner.Stop(log.Ssuccessf("Emptied ECR repository for task %s.\n", color.HighlightUserInput(o.name)))
```
dannyrandall, line 400: good catch!:tada:

internal/pkg/template/templates/task/cf.yml
```diff
@@ -255,7 +256,48 @@ Resources:
     Properties:
       LogGroupName: !Join ['', ["/copilot/", !Ref TaskName]]
       RetentionInDays: !Ref LogRetention
+  S3Bucket:
```
dannyrandall, line 259: maybe i'm not quite piecing everything together from the different threads, so if you could help summarize that would be great😅: why create a new bucket instead of using the app bucket?

internal/pkg/template/templates/task/cf.yml
```diff
@@ -255,7 +256,48 @@ Resources:
...
+          - ServerSideEncryptionByDefault:
+              SSEAlgorithm: AES256
+      LifecycleConfiguration:
+        Rules:
+          # .env files are only needed on the initial RunTask call and are not needed after that.
+          # This prevents them from piling up (hopefully it does not take 1 day to build the docker image).
+          - Id: DeleteEnvFilesRule
+            Status: Enabled
+            Prefix: 'manual/env-files'
+            ExpirationInDays: 1
```
dannyrandall, line 277: this is smart!

internal/pkg/template/templates/task/cf.yml
dannyrandall: ok actually i think i figured it out - you can do `copilot task run` without having created an app :)

dannyrandall: so follow-up question which i think is true, but just to verify: running `task delete` will
1. empty the s3 bucket
2. delete the cf stack, which will delete the s3 bucket

internal/pkg/cli/task_run.go
```diff
@@ -647,6 +673,18 @@ func (o *runTaskOpts) Execute() error {
 		return err
 	}
 
+	update := false
+
```
efekarakus: nit: the slightly more idiomatic way of writing it is just:
```suggestion
	var shouldUpdate bool
```

internal/pkg/deploy/task.go
```diff
@@ -45,6 +45,8 @@ type TaskStackInfo struct {
 	Env       string
 
 	RoleARN string
+
+	S3Bucket string
```
efekarakus: nit: 
```suggestion
	BucketName string
```

internal/pkg/cli/task_run.go
```diff
@@ -647,6 +673,18 @@ func (o *runTaskOpts) Execute() error {
...
+	if o.envFile != "" {
+		envFileARN, err := o.deployEnvFile()
```
efekarakus, line 679: Just to make sure, it is indeed guaranteed that the S3 bucket exists at this point right? because above this line we call `o.deployTaskResources()`.

I wanted to make sure that existing task stacks (not just new ones) won't suddenly fail if they use this flag

efekarakus (commented):
Looks great! just a nit and question 🚀

internal/pkg/template/templates/task/cf.yml
rclinard-amzn: That's all correct I think!

internal/pkg/cli/task_run.go
rclinard-amzn: I believe so. At this point, we have already deployed the CFN stack once, with the S3 bucket and ECR repository unconditionally available, and then the built image & env file are pushed if needed, and then the CFN stack is deployed again to add references to the newly uploaded data. So existing deployed stacks will be updated to have an S3 bucket no matter what, and this logic is sound for ECR, so it should work for S3 too.

rclinard-amzn:
I think all comments have been addressed! Thanks for the feedback!

dannyrandall (approved):
great work!

efekarakus (approved):

PR #3810, opened 2022-07-25
feat: configure environment security group from env manifest

paragbhingre:
This PR fix the issue https://github.com/aws/copilot-cli/issues/3719

Related to [#3749](https://github.com/aws/copilot-cli/pull/3749)

Now customers can define security group rules for EnvironmentSecurityGroup (e.g. ingress/egress) in their env manifest

Sample manifest template with security group rules is given below

```
network:
  vpc:
    security_group:
      ingress:
        - ip_protocol: tcp
           ports: 0-65535
           cidr: 0.0.0.0/0
      egress:
        - ip_protocol: tcp
           ports: 80
           cidr: 0.0.0.0/0
```


rclinard-amzn:
Should this be classified as `feat`?

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -192,6 +197,36 @@ func (e *EnvStackConfig) Parameters() ([]*cloudformation.Parameter, error) {
 	return e.transformParameters(currParams, e.prevParams, transformEnvControllerParameters)
 }
 
+func getSecurityGroupConfig(mft *manifest.Environment) (*template.SecurityGroupConfig, error) {
```
efekarakus: should this be in the `stack/transformers.go` file? and renamed `convertEnvSecurityGroupCfg` to be consistent with the rest of package

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -192,6 +197,36 @@ func (e *EnvStackConfig) Parameters() ([]*cloudformation.Parameter, error) {
...
+
+			return &template.SecurityGroupConfig{
+				Ingress: ingress,
+				Egress:  egress,
+			}, nil
+		} else {
+			return &template.SecurityGroupConfig{}, fmt.Errorf("security group config is invalid")
+		}
+	} else {
+		return &template.SecurityGroupConfig{}, nil
```
efekarakus: was there a reason why you return an empty pointer? typically we'll return the zero value of the type so this would be:
```suggestion
		return nil, nil
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -192,6 +197,36 @@ func (e *EnvStackConfig) Parameters() ([]*cloudformation.Parameter, error) {
...
+				egress[idx].ToPort = egressValue.ToPort
+				egress[idx].FromPort = egressValue.FromPort
+			}
+
+			return &template.SecurityGroupConfig{
+				Ingress: ingress,
+				Egress:  egress,
+			}, nil
+		} else {
+			return &template.SecurityGroupConfig{}, fmt.Errorf("security group config is invalid")
```
efekarakus:
```suggestion
			return nil, fmt.Errorf("security group config is invalid")
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -192,6 +197,36 @@ func (e *EnvStackConfig) Parameters() ([]*cloudformation.Parameter, error) {
 	return e.transformParameters(currParams, e.prevParams, transformEnvControllerParameters)
 }
 
+func getSecurityGroupConfig(mft *manifest.Environment) (*template.SecurityGroupConfig, error) {
+	if mft != nil && !mft.Network.VPC.SecurityGroupConfig.IsEmpty() {
```
efekarakus: if you have a large set of nested-ifs that's usually a hint at using guard clauses: https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -192,6 +197,36 @@ func (e *EnvStackConfig) Parameters() ([]*cloudformation.Parameter, error) {
...
+		if !mft.Network.VPC.SecurityGroupConfig.IsValid() {
```
efekarakus: shouldn't this be the opposite? also wouldn't validation be handled in the `validate() error` methods in the manifest?

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -192,6 +197,36 @@ func (e *EnvStackConfig) Parameters() ([]*cloudformation.Parameter, error) {
...
+			var ingress = make([]template.SecurityGroupRules, len(mft.Network.VPC.SecurityGroupConfig.Ingress))
+			var egress = make([]template.SecurityGroupRules, len(mft.Network.VPC.SecurityGroupConfig.Egress))
```
efekarakus: I mentioned the Law of Demeter in this PR: https://github.com/aws/copilot-cli/pull/3737#discussion_r929180846 what do you think of having the following method:

```go
package manifest

// EnvSecurityGroup returns the security group config if the user has set any values. 
// If there is no env security group settings, then returns nil and false.
func (mft *EnvironmentConfig) EnvSecurityGroup() (*SecurityGroupConfig, bool) 
```

internal/pkg/manifest/env.go
```diff
@@ -95,9 +95,47 @@ type environmentNetworkConfig struct {
...
+	Egress  []securityGroupRules `yaml:"egress,omitempty"`
+}
+
+// IsEmpty returns true if there is no security group rule defined.
+func (sgc securityGroupConfig) IsEmpty() bool {
+	return len(sgc.Ingress) == 0 && len(sgc.Egress) == 0
+}
+
+type securityGroupRules struct {
+	CidrIp     string `yaml:"cidr"`
```
efekarakus: nit: https://github.com/golang/go/wiki/CodeReviewComments#initialisms
```suggestion
	CidrIP     string `yaml:"cidr"`
```

internal/pkg/manifest/env.go
```diff
@@ -95,9 +95,47 @@ type environmentNetworkConfig struct {
...
+	FromPort   int    `yaml:"from_port"`
+	IpProtocol string `yaml:"ip_protocol"`
+	ToPort     int    `yaml:"to_port"`
+}
+
+//IsValid checks for valid security group config.
+func (sgc securityGroupConfig) IsValid() bool {
```
efekarakus: I don't get this method 😅 if the user has set any configuration we return false?

internal/pkg/manifest/env.go
```diff
@@ -95,9 +95,47 @@ type environmentNetworkConfig struct {
...
+	Ingress []securityGroupRules `yaml:"ingress,omitempty"`
+	Egress  []securityGroupRules `yaml:"egress,omitempty"`
+}
+
+// IsEmpty returns true if there is no security group rule defined.
+func (sgc securityGroupConfig) IsEmpty() bool {
+	return len(sgc.Ingress) == 0 && len(sgc.Egress) == 0
+}
+
+type securityGroupRules struct {
```
efekarakus:
```suggestion
type securityGroupRule struct {
```

internal/pkg/manifest/env.go
```diff
@@ -95,9 +95,47 @@ type environmentNetworkConfig struct {
...
+	SecurityGroupConfig securityGroupConfig  `yaml:"security_group,omitempty"`
+}
+
+type securityGroupConfig struct {
+	Ingress []securityGroupRules `yaml:"ingress,omitempty"`
+	Egress  []securityGroupRules `yaml:"egress,omitempty"`
+}
+
+// IsEmpty returns true if there is no security group rule defined.
+func (sgc securityGroupConfig) IsEmpty() bool {
```
efekarakus:
```suggestion
func (cfg securityGroupConfig) IsEmpty() bool {
```

internal/pkg/manifest/env.go
```diff
@@ -95,9 +95,47 @@ type environmentNetworkConfig struct {
...
+		if egress.IsEmpty() {
+			return true
+		}
+	}
+	return false
+}
+
+//IsEmpty checks for required parameters in the security group rules.
+func (sgr securityGroupRules) IsEmpty() bool {
+	return sgr.CidrIp == "" || sgr.IpProtocol == ""
```
efekarakus: what about the other fields?

internal/pkg/template/env.go
```diff
@@ -147,6 +149,19 @@ type Telemetry struct {
 	EnableContainerInsights bool
 }
 
+// SecurityGroupConfig holds the fields to import security group config
+type SecurityGroupConfig struct {
+	Ingress []SecurityGroupRules
+	Egress  []SecurityGroupRules
+}
+
+type SecurityGroupRules struct {
```
efekarakus: https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

```suggestion
type SecurityGroupRule struct {
```

internal/pkg/template/env.go
```diff
@@ -112,6 +112,8 @@ type EnvOpts struct {
 	AllowVPCIngress          bool
 	Telemetry                *Telemetry
 
+	SecurityGroupConfig *SecurityGroupConfig
```
efekarakus: Can we instead have this under `VPCConfig` and not be a pointer so that it is always set

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -153,6 +153,20 @@ Resources:
       Tags:
         - Key: Name
           Value: !Sub 'copilot-${AppName}-${EnvironmentName}-env'
+{{- if and .SecurityGroupConfig .SecurityGroupConfig.Ingress }}
+      SecurityGroupIngress: {{range $ind, $securityRule := .SecurityGroupConfig.Ingress}}
+        - IpProtocol: {{$securityRule.IpProtocol}}
+          FromPort: {{$securityRule.FromPort}}
+          ToPort: {{$securityRule.ToPort}}
+          CidrIp: {{$securityRule.CidrIp}}
+{{end }}{{- end}}
```
```suggestion
{{- if (gt (len .SecurityGroupConfig.Ingress) 0)}}
      SecurityGroupIngress: 
      {{- range $ind, $securityRule := .SecurityGroupConfig.Ingress}}
        - IpProtocol: {{$securityRule.IpProtocol}}
          FromPort: {{$securityRule.FromPort}}
          ToPort: {{$securityRule.ToPort}}
          CidrIp: {{$securityRule.CidrIp}}
      {{- end }}
{{- end}}
```

internal/pkg/deploy/cloudformation/stack/env.go
paragbhingre: You are right I should have had this as nil. But there is a concern that I have which I have mentioned [here](https://github.com/aws/copilot-cli/pull/3810#discussion_r929559430)

internal/pkg/manifest/env.go
paragbhingre: Other two fields that we allow are `to_port` and `from_port` which are not mandatory fields. If customer doesn't set those fields then default value of those fields are `0`

internal/pkg/template/env.go
paragbhingre: Sure I can have this under `VPCConfig` but the only challenge is that we will need to return empty object from [`getSecurityGroupConfig`](https://github.com/aws/copilot-cli/pull/3810/files#diff-b2b0270288b7891d9a0645336e6267dd39f5b3f4a6b3bf1f6ef19bd4c2806195R201) method like `template.SecurityGroupConfig{}` as it does not allow us to send `nil` value in that case which you were recommending [here](https://github.com/aws/copilot-cli/pull/3810#discussion_r929380118)

internal/pkg/deploy/cloudformation/stack/env.go
paragbhingre: I have added this method to `env_validate.go` as you mentioned.

internal/pkg/manifest/env.go
paragbhingre: Hmm, actually I was using condition in [here](https://github.com/aws/copilot-cli/pull/3810/files#diff-b2b0270288b7891d9a0645336e6267dd39f5b3f4a6b3bf1f6ef19bd4c2806195R203) which was making sense because of negation but the usage was not right.  I have changed it to read it correctly now.

internal/pkg/deploy/cloudformation/stack/env.go
paragbhingre: Sounds good to me I have added this method to check if env security group config exist.

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -189,6 +189,24 @@ Resources:
       Tags:
         - Key: Name
           Value: !Sub 'copilot-${AppName}-${EnvironmentName}-env'
+{{- if (gt (len .VPCConfig.SecurityGroupConfig.Ingress) 0)}}
+      SecurityGroupIngress:
+      {{range $ind, $securityRule := .VPCConfig.SecurityGroupConfig.Ingress}}
```
Lou1415926: If we don't need to use `$idx` we could also:
```suggestion
      {{range $securityRule := .VPCConfig.SecurityGroupConfig.Ingress}}
```

internal/pkg/template/env.go
```diff
@@ -124,8 +124,9 @@ type EnvOpts struct {
 type CDNConfig struct{}
 
 type VPCConfig struct {
-	Imported *ImportVPC // If not-nil, use the imported VPC resources instead of the Managed VPC.
-	Managed  ManagedVPC
+	Imported            *ImportVPC // If not-nil, use the imported VPC resources instead of the Managed VPC.
+	Managed             ManagedVPC
+	SecurityGroupConfig SecurityGroupConfig
```
Lou1415926: Same concern! Whether we consider environment security group a part of VPC config or not 🤔

internal/pkg/manifest/env.go
```diff
@@ -102,9 +102,56 @@ type environmentNetworkConfig struct {
 }
 
 type environmentVPCConfig struct {
-	ID      *string              `yaml:"id,omitempty"`
-	CIDR    *IPNet               `yaml:"cidr,omitempty"`
-	Subnets subnetsConfiguration `yaml:"subnets,omitempty"`
+	ID                  *string              `yaml:"id,omitempty"`
+	CIDR                *IPNet               `yaml:"cidr,omitempty"`
+	Subnets             subnetsConfiguration `yaml:"subnets,omitempty"`
+	SecurityGroupConfig securityGroupConfig  `yaml:"security_group,omitempty"`
```
Lou1415926, line 112: Do we consider environment security group a part of VPC config? From my understanding, it should be on its own.

internal/pkg/manifest/env.go
```diff
@@ -102,9 +102,56 @@ type environmentNetworkConfig struct {
...
+func (cfg *EnvironmentConfig) EnvSecurityGroup() (*securityGroupConfig, bool) {
+	if isEmpty := cfg.Network.VPC.SecurityGroupConfig.IsEmpty(); !isEmpty {
+		return &cfg.Network.VPC.SecurityGroupConfig, true
+	}
+	return nil, false
+}
+
+//IsEmpty checks for required parameters in the security group rules.
+func (cfg securityGroupRule) IsEmpty() bool {
+	return cfg.CidrIP == "" || cfg.IpProtocol == ""
```
Lou1415926: Should this be an `&&`?

internal/pkg/manifest/validate_env.go
```diff
@@ -68,6 +68,9 @@ func (cfg environmentVPCConfig) validate() error {
 	if err := cfg.Subnets.validate(); err != nil {
 		return fmt.Errorf(`validate "subnets": %w`, err)
 	}
+	if err := cfg.SecurityGroupConfig.validate(); err != nil {
+		return fmt.Errorf(`validate "security group rules": %w`, err)
```
Lou1415926: I think the error wrapping format we are trying to follow in `Validate` is: `validate "field_name": error`. Applied here it'd be `validate "security_group": %w`.

internal/pkg/manifest/env.go
```diff
@@ -102,9 +102,56 @@ type environmentNetworkConfig struct {
...
+
+type securityGroupRule struct {
+	CidrIP     string `yaml:"cidr"`
+	FromPort   int    `yaml:"from_port"`
+	IpProtocol string `yaml:"ip_protocol"`
+	ToPort     int    `yaml:"to_port"`
+}
+
+//IsValid checks for valid security group config.
+func (cfg securityGroupConfig) IsValid() bool {
```
Lou1415926: Two questions!
1. Why don't we inline these logic in `func (cfg securityGroupConfig) validate() error`?
2. Do you have an example of an invalid manifest, according to this method?

internal/pkg/manifest/env.go
paragbhingre: None of these should be empty as they are required parameters hence `or` is required

Lou1415926: "They shouldn't be empty" this should have been handled in `Validate`. `Validate` should validate that any required field is not empty.

What you have here would result in the following:
```
c := securityGroupRule {
   CidrIP: "ip"
}
// c.IsEmpty() is true
```
which seems wrong because `securityGroupRule` is apparently non-empty 😢  

I feel like we should make a distinction between "what is required" and "what is empty". Missing a required field doesn't mean the structure is empty; instead, i think it means the input is invalid. On the other hand, all fields being empty in a struct may mean the struct is empty.

internal/pkg/manifest/env.go
paragbhingre: 1. I think your recommendation make sense to inline `IsValid` inside `validate()`
2. I have added couple of tests demonstrating the invalid input.

internal/pkg/manifest/env.go
paragbhingre: ~As per the discussion renaming this method as `Validate() error` for the less confusion.~ 
added `validate () error` for `securityGroupRule` in the validate.go, and by removing `IsEmpty()` completely.

internal/pkg/manifest/env.go
paragbhingre: I have took this field out of the VPC as per discussion and setting it on Network level instead under the VPC.

internal/pkg/manifest/validate_env.go
```diff
@@ -81,6 +83,32 @@ func (cfg environmentVPCConfig) validate() error {
 	return nil
 }
 
+// validate returns nil if securityGroupRule has all the required parameters set.
+func (cfg securityGroupRule) validate() error {
+	if cfg.CidrIP == "" {
+		return fmt.Errorf(`cidr field is required`)
```
Lou1415926: nit: we could use this error: https://github.com/aws/copilot-cli/blob/3fbc49d39f5e489dbbbf674744cb4b32e9a64d13/internal/pkg/manifest/errors.go#L58

internal/pkg/manifest/validate_env.go
```diff
@@ -81,6 +83,32 @@ func (cfg environmentVPCConfig) validate() error {
...
+		return fmt.Errorf(`ip_protocol field is required`)
+	}
+	return nil
+}
+
+// validate returns nil if securityGroupConfig is configured correctly.
+func (cfg securityGroupConfig) validate() error {
+	for _, ingress := range cfg.Ingress {
+		if err := ingress.validate(); err != nil {
+			return err
```
Lou1415926: nit: we could wrap the err like this: https://github.com/aws/copilot-cli/blob/3fbc49d39f5e489dbbbf674744cb4b32e9a64d13/internal/pkg/manifest/validate_env.go#L87

to provide more information to users.

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -189,6 +189,24 @@ Resources:
...
+          FromPort: {{$securityRule.FromPort}}
+          ToPort: {{$securityRule.ToPort}}
+          CidrIp: {{$securityRule.CidrIP}}
+      {{end }}
+{{- end}}
+  {{- if (gt (len .SecurityGroupConfig.Egress) 0)}}
+      SecurityGroupEgress:
+      {{range $securityRule := .SecurityGroupConfig.Egress}}
+        - IpProtocol: {{$securityRule.IpProtocol}}
+          FromPort: {{$securityRule.FromPort}}
```
Lou1415926, line 205: What if we have a manifest like this:
```yaml
security_group:
  ingress:
    - ip_protocol: tcp
      cidr_ip: 10.0.0.0/24
``` 


The manifest will pass the validation, but the template would be:
```yaml
  SecurityGroupIngress:
     - IpProtocol: tcp
        FromPort: 0
        ToPort: 0
         CidrIp: 10.0.0.0/24
```

Should this be expected? That we automatically open up the ingress on port 0?

internal/pkg/manifest/env.go
```diff
@@ -107,6 +108,32 @@ type environmentVPCConfig struct {
 	Subnets subnetsConfiguration `yaml:"subnets,omitempty"`
 }
 
+type securityGroupConfig struct {
+	Ingress []securityGroupRule `yaml:"ingress,omitempty"`
+	Egress  []securityGroupRule `yaml:"egress,omitempty"`
+}
+
+// IsEmpty returns true if there is no security group rule defined.
+func (cfg securityGroupConfig) IsEmpty() bool {
```
Lou1415926: Does this method need to be exported?

internal/pkg/manifest/env.go
```diff
@@ -107,6 +108,32 @@ type environmentVPCConfig struct {
...
+	return len(cfg.Ingress) == 0 && len(cfg.Egress) == 0
+}
+
+type securityGroupRule struct {
+	CidrIP     string `yaml:"cidr"`
```
Lou1415926: This is sort of related to my other comment in `templates/environment/cf.yml`. I keep debating whether in general we will need to standardize these basic fields to have a pointer, so that we can distinguish an input of zero value from an absent input in env manifest, that is for example, to tell
```
- to_port: 0
  ip_protocol: "tcp"
```
from
```
- ip_protocol: "tcp"
```

If we use `*int`, then we would be able to; if we use `int` (like what you are doing now), then we won't be able to tell. 

For service manifest, we need to be able to tell them apart so that environment overrides are applied correctly; but this is not a thing in environment manifest. 

Up until now, it seems like for env manifest, it's not entirely necessary for us to distinguish zero-value vs. absent except for `bool` fields. But I don wonder though whether we should force every basic field to be a pointer field just to standardize it 💭

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -354,6 +354,31 @@ type networkLoadBalancerConfig struct {
 	appDNSName           *string
 }
 
+func convertEnvSecurityGroupCfg(mft *manifest.Environment) template.SecurityGroupConfig {
+	securityGroupConfig, isSecurityConfigSet := mft.EnvSecurityGroup()
+	if isSecurityConfigSet {
```
efekarakus: nit: we can flip the conditional to reduce nesting
```suggestion
	if !isSecurityConfigSet {
		return template.SecurityGroupConfig{}
	}
	ingress := make(// remaining code...)
```

internal/pkg/template/templates/environment/cf.yml
paragbhingre: So when I had tried setting same things on AWS console then it sets port range to 0 by default so I just tried setting the similar experience here.

internal/pkg/manifest/env.go
```diff
@@ -85,10 +85,11 @@ func (e *Environment) MarshalBinary() ([]byte, error) {
...
 type EnvironmentConfig struct {
-	Network       environmentNetworkConfig `yaml:"network,omitempty,flow"`
-	Observability environmentObservability `yaml:"observability,omitempty,flow"`
-	HTTPConfig    EnvironmentHTTPConfig    `yaml:"http,omitempty,flow"`
-	CDNConfig     environmentCDNConfig     `yaml:"cdn,omitempty,flow"`
+	Network             environmentNetworkConfig `yaml:"network,omitempty,flow"`
+	Observability       environmentObservability `yaml:"observability,omitempty,flow"`
+	HTTPConfig          EnvironmentHTTPConfig    `yaml:"http,omitempty,flow"`
+	CDNConfig           environmentCDNConfig     `yaml:"cdn,omitempty,flow"`
+	SecurityGroupConfig securityGroupConfig      `yaml:"security_group,omitempty"`
```
efekarakus: Sorry Parag this is my bad! would you mind moving back this field to under what you had originally `network.vpc.`?

The main reasoning behind it is that:
1. Consistency between the existing [`network.vpc.security_groups`](https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#network-vpc-security-groups) field as this sg affects the connectivity of all services in the VPC.
2. We will be adding also `http.public.security_group` so there is a nice symmetry by being two levels deep

internal/pkg/template/templates/environment/cf.yml
efekarakus: Yeah! I'm also curious what's the behavior today with CFN when the `FromPort` and `ToPort` are omitted - do they translate to 0 behind the scenes if you open it up in the console?

efekarakus (commented):
Looks great! just some small feedback

internal/pkg/template/templates/environment/cf.yml
paragbhingre: So the fun fact is AWS console by default populate `0` in the port range when you try to create security group rule and CFN sets `All` as port range when you do not specify from_port or to_port. I am sure `All` means 0-65535. But if customers doesn't specify to_port and/or from_port then we can either set ToPort to 0 and FromPort to 65535.

What do you think about this @efekarakus  and @Lou1415926 ?

internal/pkg/manifest/env.go
paragbhingre: Actually it was required before when we had not implemented the Demeter law suggested by Efe but now I think we don't need to export this I will change it to lower case.

internal/pkg/manifest/env.go
paragbhingre: I had a conversation with @Lou1415926 about this and we concluded that security group of the environment is not a part of `VPC` but it's a part of environment itself so it shouldn't go under `VPC` config and be on it's own in the env manifest, hence I made changes accordingly. 
What are your thoughts on this?

internal/pkg/manifest/env.go
paragbhingre: That is a good call out actually. In the env manifest we do not have a consistent way of implementing this particularly thing. Currently, I have changed ToPort and FromPort to be pointers so that we can figure if customer has actually set this parameters in the manifest or not? And if not then set them as `0` or `65535`  whatever we decide in our other conversation [here](https://github.com/aws/copilot-cli/pull/3810#discussion_r931520948)

internal/pkg/manifest/env.go
paragbhingre: As per internal discussion with the team we are concluding this as -

`security_group` should be inside `network.vpc` as `EnvironmentSecurityGroup` will get applied to all the services in the environment in the particular vpc. And it also aligns well with the `http.public.security_groups`  that we have in the env manifest. So `network.vpc`'s security_group is for all the services in the environment and `http.public`'s security_groups is applied to the load balancer.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -354,6 +354,52 @@ type networkLoadBalancerConfig struct {
...
+		toPort = *cfg.PortsConfig.Port
+		fromPort = *cfg.PortsConfig.Port
+	} else {
+		fromPort, toPort, err = cfg.PortsConfig.Ports.Parse()
+		if err != nil {
+			return 0, 0, err
+		}
+	}
+	return fromPort, toPort, nil
+}
```
efekarakus: What do you think of moving this to be a method on `*manifest.SecurityGroupRule` instead? This way `IsEmpty` and `Parse` can become private methods.
```go
// Ports returns the from and to ports of a security group rule.
func (r manifest.SecurityGroupRule) Ports() (from, to int, err error) {
  if r.PortsConfig.Ports.isEmpty() {
    return *cfg.PortsConfig.Port, *cfg.PortsConfig.Port, nil // a single value is provided for ports. 
  }
  return cfg.PortsConfig.Ports.parse()
}
```

internal/pkg/manifest/env.go
```diff
@@ -102,9 +109,98 @@ type environmentNetworkConfig struct {
...
+type securityGroupConfig struct {
+	Ingress []SecurityGroupRule `yaml:"ingress,omitempty"`
+	Egress  []SecurityGroupRule `yaml:"egress,omitempty"`
+}
+
+func (cfg securityGroupConfig) isEmpty() bool {
+	return len(cfg.Ingress) == 0 && len(cfg.Egress) == 0
+}
+
+// SecurityGroupRule holds the security group ingress and egress configs
```
efekarakus: nit: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences
```suggestion
// SecurityGroupRule holds the security group ingress and egress configs.
```

internal/pkg/manifest/env.go
```diff
@@ -102,9 +109,98 @@ type environmentNetworkConfig struct {
...
+type SecurityGroupRule struct {
+	CidrIP      string      `yaml:"cidr"`
+	PortsConfig PortsConfig `yaml:"ports"`
+	IpProtocol  string      `yaml:"ip_protocol"`
+}
+
+// Ports is a custom type which supports unmarshaling yaml which
+// can either be of type int or type PortsRangeBand.
```
efekarakus:
```suggestion
// PortsConfig represents a range of ports [from:to] inclusive. 
// The simple form allow represent from and to ports as a single value, whereas the advanced form is for different values.
```

internal/pkg/manifest/env.go
```diff
@@ -102,9 +109,98 @@ type environmentNetworkConfig struct {
...
+type PortsConfig struct {
+	Port  *int            // 0 is a valid value, so we want the default value to be nil.
+	Ports *PortsRangeBand // Mutually exclusive with port.
+}
+
+// PortsRangeBand is a number range with to and from port values.
```
efekarakus:
```suggestion
// PortsRangeBand represents a range of from and to port values.
// For example, "0-65536".
```

internal/pkg/manifest/env.go
```diff
@@ -102,9 +109,98 @@ type environmentNetworkConfig struct {
...
+type PortsRangeBand string
+
+// Parse parses Ports string and returns the from and to port values.
+// For example: 1-100 returns	1 and 100.
+func (str PortsRangeBand) Parse() (fromPort int, toPort int, err error) {
+	ports := strings.Split(string(str), "-")
```
efekarakus: should we check if the length is equal to 2? otherwise it's possible that "0-1-2" is a valid band

internal/pkg/manifest/env.go
```diff
@@ -102,9 +109,98 @@ type environmentNetworkConfig struct {
...
+	toPort, err = strconv.Atoi(ports[1])
+	if err != nil {
+		return 0, 0, fmt.Errorf("cannot convert to_port value %s to integer", ports[1])
+	}
+	return fromPort, toPort, nil
+}
+
+// IsEmpty returns whether PortsRangeBand is empty.
+func (cfg *PortsRangeBand) IsEmpty() bool {
+	return cfg == nil
```
efekarakus: I think we might need to check the value as well otherwise an `""` string would be non-nil 
```suggestion
	return cfg == nil || len(*cfg) == ""
```

internal/pkg/manifest/env.go
```diff
@@ -102,9 +109,98 @@ type environmentNetworkConfig struct {
...
+}
+
+// IsEmpty returns whether PortsConfig is empty.
+func (cfg *PortsConfig) IsEmpty() bool {
+	return cfg.Port == nil && cfg.Ports == nil
```
efekarakus:
```suggestion
	return cfg.Port == nil && cfg.Ports.isEmpty()
```

internal/pkg/manifest/validate_env.go
```diff
@@ -81,6 +84,56 @@ func (cfg environmentVPCConfig) validate() error {
 	return nil
 }
 
+// validate returns nil if securityGroupRule has all the required parameters set.
+func (cfg SecurityGroupRule) validate() error {
+	if cfg.CidrIP == "" {
+		return fmt.Errorf(`cidr`)
```
efekarakus: ? can we instead return like this:
```suggestion
		return &errFieldMustBeSpecified{
				missingField: "cidr",
			}
```
It's very strange to read an error message that just returns the `"cidr"`

internal/pkg/manifest/validate_env.go
```diff
@@ -81,6 +84,56 @@ func (cfg environmentVPCConfig) validate() error {
...
+		return cfg.Ports.validate()
+	}
+	return nil
+}
+
+// validate checks if PortsRangeBand is set correctly.
+func (str PortsRangeBand) validate() error {
+	ports := strings.Split(string(str), "-")
+	if len(ports) != 2 {
+		return fmt.Errorf("invalid ports value %s. Should be in format of ${from_port}-${to_port}", string(str))
```
efekarakus: Typically we don't use capital letters in error messages or have punctuation: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

Would you mind updating here and:
https://github.com/aws/copilot-cli/blob/b69af51c405418199c9ca46ad293208b984202b6/internal/pkg/manifest/validate.go#L1034

to be more of the format:
https://github.com/aws/copilot-cli/blob/b69af51c405418199c9ca46ad293208b984202b6/internal/pkg/manifest/validate.go#L792

Thank you 🥺

internal/pkg/manifest/validate_env.go
```diff
@@ -81,6 +84,56 @@ func (cfg environmentVPCConfig) validate() error {
...
+	return nil
+}
+
+// validate returns nil if securityGroupConfig is configured correctly.
+func (cfg securityGroupConfig) validate() error {
+	for idx, ingress := range cfg.Ingress {
+		if err := ingress.validate(); err != nil {
+			return fmt.Errorf(`validate ingress[%d]: %w`, idx, &errFieldMustBeSpecified{
+				missingField: err.Error(),
+			})
```
efekarakus: After addressing the feedback above of moving `errFieldMustBeSpecified` to the individual validate() methods, we can then do:
```suggestion
			return fmt.Errorf(`validate ingress[%d]: %w`, idx, err)
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -189,6 +189,24 @@ Resources:
       Tags:
         - Key: Name
           Value: !Sub 'copilot-${AppName}-${EnvironmentName}-env'
+{{- if and .VPCConfig.SecurityGroupConfig (gt (len .VPCConfig.SecurityGroupConfig.Ingress) 0)}}
+      SecurityGroupIngress:
+      {{range $securityRule := .VPCConfig.SecurityGroupConfig.Ingress}}
```
efekarakus: To remove extra new lines (same for egress)
```suggestion
      {{- range $securityRule := .VPCConfig.SecurityGroupConfig.Ingress}}
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -189,6 +189,24 @@ Resources:
...
+        - IpProtocol: {{$securityRule.IpProtocol}}
+          FromPort: {{$securityRule.FromPort}}
+          ToPort: {{$securityRule.ToPort}}
+          CidrIp: {{$securityRule.CidrIP}}
+      {{end }}
```
```suggestion
      {{- end }}
```

efekarakus (commented):
Yay! the port range changes look awesome!

internal/pkg/manifest/env.go
paragbhingre: 🙈

internal/pkg/manifest/env.go
paragbhingre: We are already checking that before even getting to this level. I mean we are checking that in validate. So if customer provides "0-1-2" then it will throw an error which is mentioned in this [test](https://github.com/aws/copilot-cli/pull/3810/files#diff-f06fdcd3f96407fd662bf9b5e635c012e13124fcaedf292652acb93788b65d90R134)

internal/pkg/manifest/env.go
paragbhingre: Do you mean `*cfg == "`. Because I am not sure how we will check the length of the PortsRangeBand

internal/pkg/manifest/validate_env.go
paragbhingre: Sure

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: That's such a great idea. I like this one. I will make the change. :)
But I can only make `Parse` method private as `IsEmpty` is used in the validate_env.

internal/pkg/manifest/env.go
```diff
@@ -102,9 +109,107 @@ type environmentNetworkConfig struct {
...
+// PortsConfig represents a range of ports [from:to] inclusive.
+// The simple form allow represent from and to ports as a single value, whereas the advanced form is for different values.
+type PortsConfig struct {
+	Port  *int            // 0 is a valid value, so we want the default value to be nil.
+	Ports *PortsRangeBand // Mutually exclusive with port.
+}
+
+// PortsRangeBand represents a range of from and to port values.
+// For example, "0-65536".
+type PortsRangeBand string
```
Lou1415926: qq: Do these structs have to be exported? `SecurityGroupRule`, `PortsConfig` and `PortsRangeBand`, etc.

Lou1415926: Also, what do you think of reusing `IntRangeBand` instead of creating `PortsRangeBand`?

internal/pkg/manifest/env.go
```diff
@@ -102,9 +109,107 @@ type environmentNetworkConfig struct {
...
+}
+
+// IsEmpty returns whether PortsRangeBand is empty.
+func (cfg *PortsRangeBand) IsEmpty() bool {
+	return cfg == nil || *cfg == ""
+}
+
+// IsEmpty returns whether PortsConfig is empty.
+func (cfg *PortsConfig) IsEmpty() bool {
+	return cfg.Port == nil && cfg.Ports.IsEmpty()
```
Lou1415926: What about
```go
func (cfg *PortsConfig) IsEmpty() bool {
	return cfg.Port == nil && cfg.Ports == nil
```
and remove `func (cfg *PortsRangeBand) IsEmpty() bool`?

internal/pkg/manifest/env.go
```diff
@@ -102,9 +109,107 @@ type environmentNetworkConfig struct {
...
+}
+
+// Ports returns the from and to ports of a security group rule.
+func (r SecurityGroupRule) Ports() (from, to int, err error) {
+	if r.PortsConfig.Ports.IsEmpty() {
+		return *r.PortsConfig.Port, *r.PortsConfig.Port, nil // a single value is provided for ports.
```
Lou1415926: perhaps
```suggestion
		return aws.IntValue(r.PortsConfig.Port), aws.IntValue(r.PortsConfig.Port), nil // a single value is provided for ports.
```
for the actual safeguard to prevent nil pointer panic.

internal/pkg/manifest/env.go
```diff
@@ -102,9 +109,107 @@ type environmentNetworkConfig struct {
...
+	}
+	return r.PortsConfig.Ports.parse()
+}
+
+// UnmarshalYAML overrides the default YAML unmarshaling logic for the Ports
+// struct, allowing it to perform more complex unmarshaling behavior.
+// This method implements the yaml.Unmarshaler (v3) interface.
+func (cfg *PortsConfig) UnmarshalYAML(value *yaml.Node) error {
+	_, err := strconv.Atoi(value.Value) // detmines if input is integer? if err then we decode range values
```
Lou1415926: qq: is this check necessary? If the input isn't an integer, `value.Decode(&cfg.Port)` would have returned an error anyway, right?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -354,6 +354,39 @@ type networkLoadBalancerConfig struct {
...
+	var egress = make([]template.SecurityGroupRule, len(securityGroupConfig.Egress))
+	for idx, ingressValue := range securityGroupConfig.Ingress {
+		ingress[idx].IpProtocol = ingressValue.IpProtocol
+		ingress[idx].CidrIP = ingressValue.CidrIP
+		if fromPort, toPort, err := ingressValue.Ports(); err != nil {
+			return nil, err
+		} else {
+			ingress[idx].ToPort = toPort
+			ingress[idx].FromPort = fromPort
+		}
```
```suggestion
		if fromPort, toPort, err := ingressValue.Ports(); err != nil {
			return nil, err
		} 
		ingress[idx].ToPort = toPort
		ingress[idx].FromPort = fromPort
```

internal/pkg/manifest/env.go
```diff
@@ -102,9 +109,107 @@ type environmentNetworkConfig struct {
...
+	PortsConfig PortsConfig `yaml:"ports"`
+	IpProtocol  string      `yaml:"ip_protocol"`
+}
+
+// PortsConfig represents a range of ports [from:to] inclusive.
+// The simple form allow represent from and to ports as a single value, whereas the advanced form is for different values.
+type PortsConfig struct {
+	Port  *int            // 0 is a valid value, so we want the default value to be nil.
+	Ports *PortsRangeBand // Mutually exclusive with port.
+}
```
Lou1415926: what do you think of these naming:
```go
// SecurityGroupRule holds the security group ingress and egress configs.
type SecurityGroupRule struct {
	CidrIP      string      `yaml:"cidr"`
	Ports        PortsConfig `yaml:"ports"`
	IpProtocol  string      `yaml:"ip_protocol"`
}

// PortsConfig represents a range of ports [from:to] inclusive.
// The simple form allow represent from and to ports as a single value, whereas the advanced form is for different values.
type PortsConfig struct {
	Value  *int            // 0 is a valid value, so we want the default value to be nil.
	Range *PortsRangeBand // Mutually exclusive with port.
}
```

so that when called, it reads
```go
var cfg SecurityGroupRule
cfg.Ports.Value
cfg.Ports.Range
```

internal/pkg/manifest/env.go
paragbhingre: Actually this particular method has been used almost at 4 places other than this so instead of writing `PortsRangeBand == nil || *PortsRangeBand == ""` 4 times it's better to have this method in place. Also we need to check `*PortsRangeBand == ""` other wise customers can provide `PortsRangeBand = ""` and we will still not consider this as empty and fail one of the use cases.

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: I have not defined `fromPort` and `toPort` outside of the `if` scope so that `fromPort` and `toPort` will only work in either `if` or `else`. But I am happy to declare them separately outside of the `if` scope and then do what you are suggesting. Let me know if you want me to do that.

internal/pkg/manifest/env.go
paragbhingre: Yeah this looks clean, I like it. I will do this change. The only problem I already have method called `Ports` on the `SecurityGroupRule`. I will need to change the name of that method because it seems we can not have both field and method on the same struct.

internal/pkg/manifest/env.go
paragbhingre: if customer somehow defines `ports: "10"` then unmarshaling fails. Also when I write an integration test, `ports` are considered as string and then it unmarshals `ports: 10` into `PortsRangeBand` and the integ test fails. Hence converting it into int before unmarshal it.

internal/pkg/manifest/env.go
paragbhingre: Very good question, even I was wondering if I need these structs to be exported but then I could see almost all the structs are kept public in env manifest file. But you are right I should change them if they are not exported. 

Also I did not use `IntRangeBand` purposefully because `IntRangeBand` has different definition of `validate` than what we needed for the port configs.

internal/pkg/manifest/env.go
Lou1415926: What about swap the order, trying to unmarshal into int first then string?

internal/pkg/manifest/env.go
Lou1415926: I'm not sure which part of `IntRangeBand`'s [`validate` logic](https://github.com/aws/copilot-cli/blob/f04f52f4e856bbb12687720594b9ced2d15ba329/internal/pkg/manifest/validate.go#L1029) can't be reused here 🤔 I might be missing something obvious, but it seems to me that both struct's `validate` are just checking if they are formatted as `min-max`...

paragbhingre: So the [error message](https://github.com/aws/copilot-cli/blob/f04f52f4e856bbb12687720594b9ced2d15ba329/internal/pkg/manifest/validate.go#L1034) that validate function returns is different than what we need. Customer might get confused by the min and max wordings in the error message. Other than that I guess everything is fine. It's just that we return human readable errors for `strconv.Atoi`.  Let me know if you still want me to use `IntRangeBand`

internal/pkg/manifest/env.go
paragbhingre: I tried doing that as well. But the problem is that if I unmarshal into int first then it sets the value of the for example `0-65335` as 0 in the int variable and then int variable becomes non empty and wrong unmarshaling happens. So that is why I am trying to convert input into int 1st and if it is not convertible then we know for sure that it's a range.

internal/pkg/manifest/env.go
```diff
@@ -116,88 +113,62 @@ type environmentVPCConfig struct {
...
+	return cfg.Port == nil && cfg.Ports == nil
 }
 
-// Ports returns the from and to ports of a security group rule.
-func (r SecurityGroupRule) Ports() (from, to int, err error) {
-	if r.PortsConfig.Ports.IsEmpty() {
-		return *r.PortsConfig.Port, *r.PortsConfig.Port, nil // a single value is provided for ports.
+// GetPorts returns the from and to ports of a security group rule.
+func (r securityGroupRule) GetPorts() (from, to int, err error) {
+	if r.Ports.Ports == nil {
```
Lou1415926: nit: `r.Ports.Ports` reads a little awkward 😂 :homer-disappearing  what do you think of `r.Ports.Range`?

internal/pkg/manifest/validate_env.go
```diff
@@ -96,27 +95,23 @@ func (cfg SecurityGroupRule) validate() error {
...
-	ports := strings.Split(string(str), "-")
-	if len(ports) != 2 {
-		return fmt.Errorf("invalid ports value %s: valid port format is ${from_port}-${to_port}", string(str))
+	if cfg.Ports != nil {
+		if err := cfg.Ports.validate(); err != nil {
+			if strings.Contains(err.Error(), "invalid range value") {
+				return fmt.Errorf("invalid ports value %s: valid port format is ${from_port}-${to_port}", *cfg.Ports)
+			}
+			return err
+		}
```
Lou1415926: nit: what do you think of flipping the conditional
```suggestion
        if cfg.Ports == nil {
           retur nil
        }
	// check `cfg.Ports.validate()`
}
```

Lou1415926: Another nit that I'd have is that we could think of a more robust way to match the invalid range value error. Consider if one day we decide to change the error message for `IntRangeBand`'s `validate`, from `"invalid range value..."` to  something like`"range %s is invalid"`, it's likely that we'd forget to update it here 😰 . 

It is not entirely necessary for us to address the nit here. Just sharing the ⬇️ code snippets for your thoughts 💭 

An alternative that can deal with this & does not involve too much modification on top of what we have here is:
```go
// In validate.go
type errInvalidRange struct {
	value       string
	validFormat string
}

func (e *errInvalidRange) Error() string {
	return fmt.Sprintf("invalid range value %s: valid format is %s", e.value, e.validFormat)
}

// validate returns nil if IntRangeBand is configured correctly.
func (r IntRangeBand) validate() error {
	....
	// Valid minMax example: ["1-2", "1", "2"]
	if len(minMax) != 3 {
		return &errInvalidRange{
			value:       str,
			validFormat: "${min}-${max}",
		}
	}
        ...
}
```
then here:
```go
// In validate_env.go
func (cfg portsConfig) validate() error {
  ...
 if err := cfg.Ports.validate(); err != nil {
      var targetErr *errInvalidRange
      if errors.As(err, &targetErr) {
	      return &errInvalidRange{
		      value:       aws.StringValue((*string)(cfg.Ports)),
		      validFormat: "${from_port}-${to_port}",
	      }
      }
      return err
    }
    return nil
}
```

Lou1415926 (approved):
Just some nits and thoughts!

internal/pkg/manifest/env.go
paragbhingre: Hahah, I know it looks little weird. 
I will change it Range.

internal/pkg/manifest/validate_env.go
paragbhingre: Hmm totally make sense to me. We need robust way to handle such kinds of errors and your solution looks good to me. I will make changes in this PR itself to add this robust change now only in copilot.

efekarakus (approved):
Looks great! 🚀

PR #3811, opened 2022-07-25
chore: only create single `addon` service during deploy/package

dannyrandall:
Before this change, an addon template has been generated twice during `svc deploy` and `svc package` - once when generating the addons template to be uploaded, and once to computing the outputs from the addons stack to use while generating the workload template.

To enable https://github.com/aws/copilot-cli/issues/3546, we only want to generate the template _once_, and then reuse the generated template. As part of generating the addons template, we'll upload local files to S3 and replace the local file references with S3 locations (in an upcoming PR).


internal/pkg/addon/addons.go
```diff
@@ -93,10 +105,12 @@ func (a *Addons) Template() (string, error) {
 			return "", err
 		}
 	}
+
```
efekarakus, line 108: :excuseme:

internal/pkg/cli/deploy/svc.go
```diff
@@ -279,6 +286,11 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 	}, nil
 }
 
+// AddonsStackBuilder returns the StackBuilder for this workload's addons.
+func (w *workloadDeployer) AddonsStackBuilder() StackBuilder {
```
efekarakus: Was there a reason why you preferred returning an interface? 
> “Go interfaces generally belong in the package that uses values of the interface type, not the package that implements those values”
> https://www.efekarakus.com/golang/2019/12/29/working-with-interfaces-in-go.html

Can we return `*addon.Addons` and let the client define the `StackBuilder` interface? (see comment below for how to do unit-testing)
```go
func (w *workloadDeployer) Addons() *addon.Addons
```

internal/pkg/cli/svc_package.go
```diff
@@ -201,7 +189,11 @@ func (o *packageSvcOpts) Execute() error {
 	if err != nil {
 		return nil
 	}
-	svcTemplates, err := o.getSvcTemplates(targetEnv)
+	generator, err := o.getTplGenerator(targetEnv)
+	if err != nil {
+		return err
+	}
+	svcTemplates, err := o.getSvcTemplates(generator)
```
efekarakus: nit: What do you think of these names?
```go
	deployer, err := o.getStackDeployer(targetEnv)
	if err != nil {
		return err
	}
	stack, err := o.getStackConfig(deployer)
	if err != nil {
		return err
	}
	o.writeAndClose(o.templateWriter, stack.template)
	o.writeAndClose(o.templateConfigWriter, stack.templateConfig)
```

internal/pkg/cli/svc_package.go
```diff
@@ -211,13 +203,14 @@ func (o *packageSvcOpts) Execute() error {
 	if err := o.writeAndClose(o.paramsWriter, svcTemplates.configuration); err != nil {
 		return err
 	}
-	addonsTemplate, err := o.getAddonsTemplate()
-	// return nil if addons not found.
-	var notFoundErr *addon.ErrAddonsNotFound
-	if errors.As(err, &notFoundErr) {
-		return nil
-	}
+	addonsTemplate, err := generator.AddonsStackBuilder().Template()
```
efekarakus: I assume we are returning a `StackBuilder` interface to make unit-testing easier. An alternative that I've seen before is to provide a fake constructor to make testing easier for other packages (similar to [`afero.MemMapFs`](https://pkg.go.dev/github.com/spf13/afero#NewMemMapFs))

My suggestion is to do something like this:
```go
package addon

// MemAddons creates an *Addons object using dummy template and params values instead of reading the files from the workspace.
func MemAddons(tpl, params string) *Addons {
   return &Addons{
      cachedTemplate: tpl,
      params: params,
   }
}
```

Then in your unit tests:
```go
mockGenerator.EXPECT().Addons().Return(addons.MemAddons("mytemplate", "myparams"), nil)
```


This pattern is also mentioned in the [software engineering at google book](https://abseil.io/resources/swe-book/html/ch13.html#faking-id00042).

efekarakus (commented):
Looks great! Let me know what you think of the suggestions below 👇

internal/pkg/addon/addons.go
dannyrandall: i just need some space 😤

internal/pkg/cli/deploy/svc.go
dannyrandall: The main reason I went this route was to avoid other packages (`pkg/cli`) needing an explicit dependency on `pkg/addon`, instead being able to only depend on `pkg/cli/deploy`, which they already do.

The problem is that `workloadDeployer` is _already an interface_ in `pkg/cli`, so it'll need a new function like `Addons() *addon.Addons` on that interface.

I'm fine switching - i also don't love breaking accept-interfaces-return-structs; we'll just have to have an explicit dependency on `pkg/addon` in `pkg/cli` and `pkg/cli/deploy` (which is fine 😊)

dannyrandall: actually - i think i have a better solution! instead, i'll just change the function to `AddonsTemplate() (string, error)` instead.

efekarakus: :shipit:

internal/pkg/cli/svc_package.go
dannyrandall: should be resolved by https://github.com/aws/copilot-cli/pull/3811#discussion_r929391909

internal/pkg/cli/svc_package.go
dannyrandall: see what you think of the new names!

internal/pkg/cli/svc_package.go
```diff
@@ -306,29 +293,32 @@ func (o *packageSvcOpts) configureClients() error {
 }
 
 type wkldCfnTemplates struct {
```
efekarakus: should we just rename this type as well? it's pretty generic to work outside of workloads
```suggestion
type cfnStackConfig struct {
```

internal/pkg/cli/svc_package.go
```diff
@@ -356,7 +346,10 @@ func (o *packageSvcOpts) getSvcTemplates(env *config.Environment) (*wkldCfnTempl
 	if err != nil {
 		return nil, fmt.Errorf("generate workload %s template against environment %s: %w", o.name, o.envName, err)
 	}
-	return &wkldCfnTemplates{stack: output.Template, configuration: output.Parameters}, nil
+	return &wkldCfnTemplates{
+			template:   output.Template,
+			parameters: output.Parameters},
+		nil
```
efekarakus: nit: wut is this `gofmt` valid?

efekarakus (approved):

internal/pkg/cli/svc_package.go
```diff
@@ -306,29 +293,32 @@ func (o *packageSvcOpts) configureClients() error {
...
 		ws:           o.ws,
 		unmarshal:    o.unmarshal,
-		sess:         o.envSess,
 	})
 	if err != nil {
 		return nil, err
 	}
 	o.appliedManifest = mft
-	generator, err := o.newTplGenerator(o)
+	generator, err := o.newStackGenerator(o)
```
Lou1415926, line 313: nit:
```suggestion
	return o.newStackGenerator(o)
```

internal/pkg/cli/svc_package.go
dannyrandall: lol yes, but i'll move it up to the other line haha

internal/pkg/deploy/cloudformation/stack/lb_web_service_integration_test.go
```diff
@@ -61,6 +64,13 @@ func TestLoadBalancedWebService_Template(t *testing.T) {
 	manifestBytes, err := ioutil.ReadFile(path)
 	require.NoError(t, err)
 	for name, tc := range testCases {
+		ctrl := gomock.NewController(t)
```
Lou1415926, line 67: I'm just wondering instead of mock addon whether we should use `addons.New()` in the local integration test so that it simulates the real work flow as much as possible?

Lou1415926 (approved):
LGTM! Feel free to remove the label when ur ready!

internal/pkg/cli/svc_package.go
dannyrandall: resolved in https://github.com/aws/copilot-cli/pull/3820/commits/ea721cc75996dbf65858faa8b81e6da2e53878a4!

internal/pkg/deploy/cloudformation/stack/lb_web_service_integration_test.go
dannyrandall: changed in https://github.com/aws/copilot-cli/pull/3820/commits/ea721cc75996dbf65858faa8b81e6da2e53878a4!

PR #3813, opened 2022-07-26
chore: correctly display available uris with prefix lists

CaptainCarpensir:
Allows an alb uri to not be displayed when not accessible because of usage of prefix lists

Part of #3701 


internal/pkg/template/templates/environment/cf.yml
```diff
@@ -694,6 +694,15 @@ Outputs:
...
     Description: The domain name of an associated Cloud Front Distribution.
+{{- end}}
+{{- if .PublicFacingCIDRPrefixListIDs}}
+  PublicALBAccessible:
+    Condition: CreateALB
+    Value: false
+{{- else}}
+  PublicALBAccessible:
+    Condition: CreateALB
+    Value: true
```
iamhopaul123:
```suggestion
  PublicALBAccessible:
    Condition: CreateALB
{{- if .PublicFacingCIDRPrefixListIDs}}
    Value: false
{{- else}}
    Value: true
```

iamhopaul123 (approved):

internal/pkg/describe/uri.go
```diff
@@ -228,7 +228,10 @@ func (d *uriDescriber) envDNSName(path string) (accessURI, error) {
 	if err != nil {
 		return accessURI{}, fmt.Errorf("get stack outputs for environment %s: %w", d.env, err)
 	}
-	dnsNames := []string{envOutputs[d.albCFNOutputName]}
+	dnsNames := []string{}
```
dannyrandall:
```suggestion
	var dnsNames []string
```
nit

dannyrandall (approved):

PR #3819, opened 2022-07-26
fix: no extra port mapping in main container when nlb target is a sidecar

### LINKED ISSUE #3768: Load Balanced Web Service: incompatibility when nlb and sidecars are used together
Specifying nlb with a sidecar generates a cf template with the same `ContainerPort` in multiple `ContainerDefinitions` sections, which is an invalid configuration. When attempting to deploy the example service below, the following error is returned:

> Resource handler returned message: "Invalid request provided: Create TaskDefinition: TCP container port '9901' is used multiple times in task when networkMode=awsvpc.

```
name: sidecar-with-nlb

http: false

nlb:
  port: 80/tcp
  target_container: envoy

# Configuration for your containers and service.
image:
  location: public.ecr.aws/hashicorp/envoy-alpine:latest
  port: 10000

cpu: 256       # Number of CPU units for the task.
memory: 512    # Amount of memory in MiB used by the task.
count: 1       # Number of tasks that should be running in your service.

sidecars:
  envoy:
    # envoy admin port
    port: 9901
    image: public.ecr.aws/hashicorp/envoy-alpine:latest
```

Here is an excerpt from the cf template provided by `copilot svc package`
```
      ContainerDefinitions:
        - Name: !Ref WorkloadName
          Image: !Ref ContainerImage

          PortMappings:
            - ContainerPort: 9901
              Protocol: tcp
```

It appears that the cf partial below is not sidecar aware, and assumes that the nlb target is always the main workload container. 

https://github.com/aws/copilot-cli/blob/2ac6b01cdeacc0a0fb81ffdcaf731ba2bfc4cb82/internal/pkg/template/templates/workloads/partials/cf/workload-container.yml#L33

For now we are working around this with a task override as follows:

```
taskdef_overrides:
  - path: ContainerDefinitions[0].PortMappings
    value:
        Protocol: tcp
```

Lou1415926:
Fix #3768 


internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
```diff
@@ -29,17 +29,14 @@
...
-{{- if .NLB}}
-  {{if ne .NLB.Listener.TargetPort .NLB.MainContainerPort}} {{/*No need to add additional port if the target port is the same as image port*/}}
-    - ContainerPort: {{.NLB.Listener.TargetPort}}
-      {{- if eq .NLB.Listener.Protocol "UDP" }}
-      Protocol: udp
-      {{- else }}
-      Protocol: tcp
-      {{- end }}
+{{- if .NLB}} {{ $nlbListener := .NLB.Listener }} 
+  {{- if and (eq $nlbListener.TargetContainer .WorkloadName) (ne $nlbListener.TargetPort .NLB.MainContainerPort)}}
```
efekarakus, line 33: :ack: this looks good, what about this feedback from the issue:

> It appears that the cf partial below is not sidecar aware, and assumes that the nlb target is always the main workload container.

To me it sounded like the NLB target group is pointing to the wrong port?

Lou1415926: With a manifest like this:
```yaml
nlb:
  port: 80/tcp
  target_container: envoy

image:
  location: blah
  port: 10000

sidecars:
  envoy:
    # envoy admin port
    port: 9901
    image: public.ecr.aws/hashicorp/envoy-alpine:latest
```
The NLB target group's target port will be 9901, not 10000, with the current implementation and as expected!

My understanding for the quoted comment was specifically for the code:
https://github.com/aws/copilot-cli/blob/2ac6b01cdeacc0a0fb81ffdcaf731ba2bfc4cb82/internal/pkg/template/templates/workloads/partials/cf/workload-container.yml#L33

The (bugged) code assume the target container is main container, and add the target port to main container's port mapping if it's not the same as the container port. It fails to take into consideration the possibility that the target container could be a sidecar, in which case, it does not need to add the extra port mapping to the main container.

efekarakus: Thanks for the explanation, totally makes sense!

efekarakus (approved):

dannyrandall (approved):

PR #3820, opened 2022-07-26
chore: package addons

dannyrandall:
Doesn't include logic to read and upload local files to S3 (future PR).

Part of #3546.


internal/pkg/addon/addons.go
```diff
@@ -46,18 +46,25 @@ type Addons struct {
...
+
+	bucket   string
+	uploader uploader
+
+	UploadAssets bool
 }
 
 // New creates an Addons object given a workload name.
-func New(wlName string) (*Addons, error) {
+func New(wlName string, bucket string, uploader uploader) (*Addons, error) {
```
efekarakus: What do you think of creating two constructors?

```suggestion
// New creates a raw addons config without any modifications to the template.
func New(workload string) (*Addons, error) {

// NewPackager returns an addons config that can upload and swap local assets to s3.
func NewPackager(workload, bucket string, s3 Uploader) (*Addons, error) {
```

dannyrandall: great idea! added 😊

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// packagePropertyConfig defines how to package a particular property in a cloudformation resource.
+// There are two ways replacements occur. Given a resource configuration like:
+//  MyResource:
+//    Type: AWS::Resource::Type
+//    Properties:
+//      <Property>: file/path
+//
+// Without BucketNameProprty and ObjectKeyProperty, `file/path` is directly replaced with
```
efekarakus: nit:
```suggestion
// Without BucketNameProperty and ObjectKeyProperty, `file/path` is directly replaced with
```

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,243 @@
...
+
+	"github.com/aws/copilot-cli/internal/pkg/aws/s3"
+	"gopkg.in/yaml.v3"
+)
+
+type uploader interface {
+	Upload(bucket, key string, data io.Reader) (string, error)
+}
+
+// packagePropertyConfig defines how to package a particular property in a cloudformation resource.
```
efekarakus, line 19: Really nice explanation

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,243 @@
...
+	// after uploading an asset to S3. If this and ObjectKeyProperty are empty,
+	// a submap will not be created and an S3 location URI will replace value of Property.
+	BucketNameProperty string
+
+	// ObjectKeyProperty represents the key in a submap of Property, created
+	// after uploading an asset to S3. If this and BucketNameProperty are empty,
+	// a submap will not be created and an S3 location URI will replace value of Property.
+	ObjectKeyProperty string
+
+	// ForceZip will force a zip file to be created even if the given file URI
```
efekarakus: By "given file URI" do you mean the "give file path"?

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,243 @@
...
+	// points to a file. Directories are always zipped.
+	ForceZip bool
+}
+
+// resourcePackageConfig maps a CloudFormation resource type to configuration
+// for how to transform it's properties.
+//
+// TODO(dnrnd) AWS::Include.Location
```
efekarakus, line 76: Any reason why you decided to omit this one?
Similarly, I couldn't find entries for the `AWS::Serverless` resources afaik we should be able to support those as well:
- `AWS::Serverless::Function`
- `AWS::Serverless::LayerVersion`
- `AWS::Serverless::Application`
- `AWS::Serverless::Api`
- `AWS::Serverless::StateMachine`

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,243 @@
...
+	// after uploading an asset to S3. If this and BucketNameProperty are empty,
+	// a submap will not be created and an S3 location URI will replace value of Property.
+	ObjectKeyProperty string
+
+	// ForceZip will force a zip file to be created even if the given file URI
+	// points to a file. Directories are always zipped.
+	ForceZip bool
+}
+
+// resourcePackageConfig maps a CloudFormation resource type to configuration
```
efekarakus, line 69: nit: should we include a link here to: https://docs.aws.amazon.com/cli/latest/reference/cloudformation/package.html to mention that we should be in sync

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,243 @@
...
+	BucketNameProperty string
+
+	// ObjectKeyProperty represents the key in a submap of Property, created
+	// after uploading an asset to S3. If this and BucketNameProperty are empty,
+	// a submap will not be created and an S3 location URI will replace value of Property.
+	ObjectKeyProperty string
+
+	// ForceZip will force a zip file to be created even if the given file URI
+	// points to a file. Directories are always zipped.
+	ForceZip bool
```
efekarakus, line 62: Are we planning to adhere to this part of the specification?
>For most resources, if you don't specify a path, the command zips and uploads the current working directory.
> https://docs.aws.amazon.com/cli/latest/reference/cloudformation/package.html

Personally, I wonder if this leads to clients accidentally uploading more than they would want to. I'd be down to return an `error` if the value is an empty string instead.

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,243 @@
...
+			Property:           []string{"BodyS3Location"},
+			BucketNameProperty: "Bucket",
+			ObjectKeyProperty:  "Key",
+		},
+	},
+	"AWS::Lambda::Function": {
+		{
+			Property:           []string{"Code"},
+			BucketNameProperty: "S3Bucket",
+			ObjectKeyProperty:  "S3Key",
```
efekarakus, line 89: Do you know if the `aws cloudformation package` command populates any of the other S3 fields like `S3ObjectVersion`?

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,243 @@
...
+		{
+			Property:           []string{"Code", "S3"},
+			BucketNameProperty: "Bucket",
+			ObjectKeyProperty:  "Key",
+			ForceZip:           true,
+		},
+	},
+}
+
+func (a *Addons) packageTemplate(tmpl *cfnTemplate) (*cfnTemplate, error) {
```
efekarakus: 💭 This feels like it should be a method on `*cfnTemplate` as it's mutating its internal structure. What do you think of a signature like this:
```go
func (tpl *cfnTemplate) package(bucket string, s3 uploader) error {
  tpl.bucket = bucket
  tpl.s3 = s3
  // remaining code...
}
```

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,243 @@
...
+		resType := yamlMapGet(node, "Type").Value
+		confs, ok := resourcePackageConfig[resType]
+		if !ok {
+			continue
+		}
+
+		props := yamlMapGet(node, "Properties")
+		for _, conf := range confs {
+			if err := a.packageProperty(props, conf); err != nil {
+				return nil, fmt.Errorf("transform property %s property for %s: %w", name, conf.Property, err)
```
efekarakus: What do you think of this rewording?
```suggestion
				return nil, fmt.Errorf("package property %q of %q: %w",  strings.Join(conf.Property, "."), name, err)
```

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,243 @@
...
+		if node.Content[i].Value == key {
+			return node.Content[i+1]
+		}
+	}
+
+	return &yaml.Node{}
+}
+
+func (a *Addons) packageProperty(properties *yaml.Node, conf packagePropertyConfig) error {
+	node := properties
```
efekarakus: nit:
```suggestion
func (a *Addons) packageProperty(resourceProperties *yaml.Node, pkgCfg packagePropertyConfig) error {
	targetNode := resourceProperties
```

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,243 @@
...
+	})
+}
+
+// isFilePath returns true if the path URI doesn't have a
+// a schema indicating it's an s3 or http URI.
+func isFilePath(path string) bool {
+	if path == "" || strings.HasPrefix(path, "s3://") || strings.HasPrefix(path, "http://") || strings.HasPrefix(path, "https://") {
+		return false
+	}
+
```
efekarakus, line 266: I assume we'll be checking if the path exists under the workspace in a separate PR?

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,243 @@
...
+	if !isFilePath(node.Value) {
+		return nil
+	}
+
+	obj, err := a.uploadAddonAsset(node.Value, conf.ForceZip)
+	if err != nil {
+		return fmt.Errorf("upload asset: %w", err)
+	}
+
+	if len(conf.BucketNameProperty) == 0 && len(conf.ObjectKeyProperty) == 0 {
```
efekarakus: nit: should this be a method on `packagePropertyConfig`?
```go
func (cfg packagePropertyConfig) hasOverrideProps() bool
```

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,243 @@
...
+		node.Value = s3.Location(obj.Bucket, obj.Key)
+		return nil
+	}
+
+	return node.Encode(map[string]string{
```
efekarakus: nice! so this will replace the existing `node.Value`?

internal/pkg/addon/package_test.go
```diff
@@ -0,0 +1,187 @@
...
+    Metadata:
+      "testKey": "testValue"
+    Type: AWS::ApiGateway::RestApi
+    Properties:
+      BodyS3Location:
+        Bucket: mockBucket
+        Key: TODO
+`,
+		},
+		"AWS::AppSync::Resolver, multiple replacements in one resource": {
```
efekarakus, line 131: nice! I wonder if we should have a task to add `integ` tests that call [`cfn.ValidateTemplate`](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/Welcome.html) for each transformed resource type to ensure that our transformed keys are valid

internal/pkg/cli/deploy/svc.go
```diff
@@ -203,13 +203,14 @@ func (d *workloadDeployer) cachedEnvironmentConfig() (*manifest.Environment, err
...
-	Mft             interface{} // Interpolated, applied, and unmarshaled manifest.
-	RawMft          []byte      // Content of the manifest file without any transformations.
+	SessionProvider   *sessions.Provider
+	Name              string
+	App               *config.Application
+	Env               *config.Environment
+	ImageTag          string
+	Mft               interface{} // Interpolated, applied, and unmarshaled manifest.
+	RawMft            []byte      // Content of the manifest file without any transformations.
+	UploadAddonAssets bool
```
efekarakus, line 213: Do we need an additional flag? I thought that `svc package --upload-assets` would basically force the local paths to be uploaded

internal/pkg/addon/package.go
dannyrandall: yeah - just adding it in a future PR! since it can appear in more places - not just properties of resources - it's going to be processed differently.

---

I was planning on adding `AWS:Serverless` resources (they'd be configured the same way), but in my testing with `AWS::Serverless::Function` in an addon, I always get the error:
```
 Template format error: Unrecognized resource types: [AWS::Serverless::Function]
```
Similarly, I also get the same error with a `AWS::Serverless::Function` while using the aws cli to run `cloudformation deploy`. From that, I _think_ the `AWS::Serverless` resources only work when using SAM CLI. (https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification.html)

internal/pkg/addon/package.go
dannyrandall: yes - i'll say that instead :)

internal/pkg/addon/package.go
dannyrandall: 😇

internal/pkg/addon/package.go
dannyrandall: 🤦‍♂️ (see https://github.com/aws/copilot-cli/pull/3820/commits/77629eca24a5618dac62bb94fefe8b052d20aa4c)

internal/pkg/addon/package.go
dannyrandall: Two answers here:
1. We are not planning on adhering to that part of the spec - I agree with your concern.
2. `ForceZip: true` means that if you write an `AWS::Lambda::Function` with `Code: ./index.js`, we'll create a zip from that file instead of uploading it directly (i.e., to `s3://bucket/index.js`), because lambda requires s3 objects to be zip archives.

internal/pkg/addon/package.go
dannyrandall: great idea!

internal/pkg/addon/package.go
efekarakus: oh! It should work if you add the following to your template:
```
Transform: AWS::Serverless-2016-10-31
```
CFN will then run a macro behind the scenes to transform the template and expand the resources into regular service resources

internal/pkg/addon/package.go
dannyrandall: yes sir🫡 (gotta call `os.Open()` for all of them 😊)

internal/pkg/addon/package.go
dannyrandall: correct! ☺️

internal/pkg/addon/package.go
dannyrandall: oh🙄 then yes i will add them :joy: 

(also just tested and you're right, it works 😊)

internal/pkg/cli/svc_deploy.go
```diff
@@ -118,13 +118,14 @@ func newSvcDeployer(o *deploySvcOpts) (workloadDeployer, error) {
...
-		Mft:             content,
-		RawMft:          raw,
+		SessionProvider:   o.sessProvider,
+		Name:              o.name,
+		App:               targetApp,
+		Env:               o.targetEnv,
+		ImageTag:          o.imageTag,
+		Mft:               content,
+		RawMft:            raw,
+		UploadAddonAssets: false, // TODO(dnrnd): change to true to enable packaging addons
```
Lou1415926, line 128: qq before I review the rest: will this be a flag in `deploy` and `package` commands or will this always be true?

dannyrandall: this feature will not add any new flags. on `svc deploy`, it will always upload the local assets. for `svc package` it will respect the `--upload-assets` flag.

internal/pkg/cli/deploy/svc.go
dannyrandall: does this help? https://github.com/aws/copilot-cli/pull/3820#discussion_r931466274. i don't think we need any additional flags

internal/pkg/addon/package.go
dannyrandall: So this was interesting. The AWS CLI sorta supports object versions, but it doesn't actually ever happen😊.

They [configure versions on resources](https://github.com/aws/aws-cli/blob/ef7b4d565370a3eb0a81fac2f520c8baa4e9284c/awscli/customizations/cloudformation/artifact_exporter.py#L372), but the [uploader](https://github.com/aws/aws-cli/blob/ef7b4d565370a3eb0a81fac2f520c8baa4e9284c/awscli/customizations/s3uploader.py#L129-L148) only ever returns a [bucket and key](https://github.com/aws/aws-cli/blob/ef7b4d565370a3eb0a81fac2f520c8baa4e9284c/awscli/customizations/s3uploader.py#L168-L170).

Tested using `--force-upload` on the AWS CLI and a bucket with object versioning enabled, and verified new object versions _were_ showing up in S3 without the cli including the version property on the result template.

---

also, a side discovered bug in the cli that we avoid (you'll see in the next PR) for resources that `ForceZip`, they copy it to a temp directory and zip the directory which results in the hash of the file _always_ changing. we "force upload" every time anyways, so an object is always uploaded to s3 anyways, but ours will keep the same object key 😊

dannyrandall: In terms of "Should we support object versions" - I don't have any strong preferences there...

internal/pkg/addon/package.go
dannyrandall: although sadly we can't name a function `package` 😔

internal/pkg/addon/package_test.go
dannyrandall: this would be cool, i'll add it as a todo item before we wrap this feature up

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,276 @@
...
+			BucketNameProperty: "Bucket",
+			ObjectKeyProperty:  "Key",
+		},
+	},
+	"AWS::Serverless::StateMachine": {
+		{
+			Property:           []string{"DefinitionUri"},
+			BucketNameProperty: "Bucket",
+			ObjectKeyProperty:  "Key",
+			ForceZip:           true,
```
efekarakus: Oh why is this one `ForceZip`? compared to `AWS::StepFunctions::StateMachine`

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,276 @@
...
+		{
+			Property:           []string{"Code", "S3"},
+			BucketNameProperty: "Bucket",
+			ObjectKeyProperty:  "Key",
+			ForceZip:           true,
+		},
+	},
+}
+
+func (t *cfnTemplate) packageTemplate(a *Addons) error {
```
efekarakus: nit:
```suggestion
func (t *cfnTemplate) pkg(a *Addons) error {
```

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,276 @@
...
+	for name, node := range resources {
+		resType := yamlMapGet(node, "Type").Value
+		confs, ok := resourcePackageConfig[resType]
+		if !ok {
+			continue
+		}
+
+		props := yamlMapGet(node, "Properties")
+		for _, conf := range confs {
+			if err := a.packageProperty(props, conf); err != nil {
```
efekarakus, line 199: this ping-pong between `addons -> cfnTemplate.package -> addons.packageProperty` is kinda funny 🤔 I can't really think of a better way so :shipit:

internal/pkg/addon/package_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+)
+
+type addonMocks struct {
+	// uploader *mocks.Mockuploader
+	// ws *mocks.MockworkspaceReader
+}
```
efekarakus: nit: would you mind deleting the dead-code in the test file (although it'll be populate soon)?

efekarakus (approved):

internal/pkg/addon/package.go
dannyrandall: yeah...i didn't like adding 2 arguments to every function `(bucket, uploader)`, so i considered making `Addons` meet this 
```
type bucketUploader interface {
  upload(key, data)
}
```
and that way i'd only pass one which wasn't as bad but also a lot of extra stuff for.....not much benefit/readability imo 😊

internal/pkg/addon/package.go
dannyrandall: 1000 iq, i didn't think of that

internal/pkg/addon/package.go
dannyrandall: 😅

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,280 @@
...
+		},
+	},
+}
+
+func (t *cfnTemplate) pkg(a *Addons) error {
+	resources := mappingNode(&t.Resources)
+
+	for name, node := range resources {
+		resType := yamlMapGet(node, "Type").Value
+		confs, ok := resourcePackageConfig[resType]
```
Lou1415926, line 192: tiny nit: perhaps
```suggestion
		confs, canPackage := resourcePackageConfig[resType]
```
or something like `isSupported`

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,280 @@
...
+//        <BucketNameProperty>: bucket
+//        <ObjectKeyProperty>: hash
+type packagePropertyConfig struct {
+	// Property is the key in a cloudformation resource's 'Properties' map to be packaged.
+	// Nested properties are represented by multiple keys in the slice, so the field
+	//  Properties:
+	//    Code:
+	//      S3: ./file-name
+	// is represented by []string{"Code", "S3"}.
+	Property []string
```
Lou1415926: nit: what do you think of `PropertyPath` ?

internal/pkg/addon/package.go
```diff
@@ -0,0 +1,280 @@
...
+		{
+			Property:           []string{"Code", "S3"},
+			BucketNameProperty: "Bucket",
+			ObjectKeyProperty:  "Key",
+			ForceZip:           true,
+		},
+	},
+}
+
+func (t *cfnTemplate) pkg(a *Addons) error {
```
Lou1415926, line 187: It's a little 🤯  to me that `cfnTemplate` is produced by addons' `Template` method but then `pkg` is called on `cfnTemplate` with addons as its parameter. I think this is fine, but it does make me think if there's other ways to do it to remove the back and forth...

Do we really need the whole `Addons` in the parameter? I started to think the implementation is general enough that it doesn't necessarily only work for addons (although that's the only use case for now) Do you think it's possible to generalize it to something like
```
func (t *cfnTemplate) pkg(p uploader)
```
Again, no biggie. If the change is too large we don't need to do it now. Just a thought 💭

internal/pkg/addon/package_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+)
+
+type addonMocks struct {
+}
+
+func TestPackage(t *testing.T) {
```
Lou1415926, line 19: Cool tests! Super nice to see how it works.

internal/pkg/deploy/cloudformation/stack/dd_pipeline_integration_test.go
```diff
@@ -37,7 +36,7 @@ func TestDD_Pipeline_Template(t *testing.T) {
 	var build deploy.Build
 
 	if err = build.Init(pipelineMft.Build, ""); err != nil {
-		fmt.Errorf("primting err: %v", err)
```
Lou1415926, line 40: How did you find out 😂

internal/pkg/cli/deploy/svc.go
```diff
@@ -241,10 +242,18 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
...
 	}
-	addonsSvc, err := addon.New(in.Name)
+
+	s3Client := s3.New(envSession)
+	var addonsSvc stackBuilder
+	if in.UploadAddonAssets {
+		addonsSvc, err = addon.NewPackager(in.Name, resources.S3Bucket, s3Client)
+	} else {
+		addonsSvc, err = addon.New(in.Name)
+	}
```
Lou1415926, line 252: 💭 💭 Feel free to ignore. I just want to share my thoughts!

What if we had
```go
package addon

// Template now returns the *Template instead of string.
func (a *Addon) Template() (*Template, error)

func (tpl *Template) String() string

func (tpl *Template) Package(s3 Uploader, bucket string) error
```

and then we could have 
```go
package cli/deploy

func (d *workloadDeployer) pushAddonsTemplateToS3Bucket(in *pushAddonsTemplateToS3BucketInput) (string, error)
     tmpl, err := d.addons.Template()
     err := tmpl.Package(in.uploader, d.resources.S3Bucket)
     reader := strings.NewReader(tmpl.String())
     url, err := in.uploader.Upload(d.resources.S3Bucket, artifactpath.Addons(d.name, []byte(tmpl.String())), reader)
     return url, nil
}
```
This decouples the `pkg` step from `read` step and `swap` step, and lets the caller decide when to `pkg`, and when to `swap`. 

What poked me into seeking for an alternative was the fact that we are implicitly setting how `addons` behave by using different constructors. Here, we use different constructors to initialize an `addons` because  in some downstream calls we want `addons` to/not to package assets. In the long run, this may become difficult to debug, because for example, to understand what a method call does we'd need to trace all the way to when the receiver was constructed. I do believe there are other corners in this codebase that do the same thing - as long as we can keep track, it's no biggie. 

Just some food for thoughts! The current approach also LGTM!

internal/pkg/cli/svc_package.go
```diff
@@ -311,11 +312,7 @@ func (o *packageSvcOpts) getStackGenerator(env *config.Environment) (workloadSta
 		return nil, err
 	}
 	o.appliedManifest = mft
-	generator, err := o.newStackGenerator(o)
```
Lou1415926, line 314: :doge thanks for remembering :doge

Lou1415926 (approved):
LGTM! Awesome work - can't wait to :shipit: ! Adding the label so that you can address my smaller nits if you want. Feel free to ignore the large ones 😇 !

internal/pkg/deploy/cloudformation/stack/dd_pipeline_integration_test.go
dannyrandall: i don't remember :joy: i think it showed up as a lint error

internal/pkg/cli/deploy/svc.go
dannyrandall: i think this sounds great!! i'm going to leave it as is for now - and get all of the functionality merged, and then revisit this and make it cleaner 😊

internal/pkg/addon/package.go
dannyrandall: same as https://github.com/aws/copilot-cli/pull/3820?notification_referrer_id=NT_kwDOAKE7RLM0MDc1NzUwNjQwOjEwNTY2NDY4#discussion_r932542758 😊

PR #3826, opened 2022-07-28
chore: zip and upload addon assets

dannyrandall:
This change adds logic to read local files referenced in an addon resource and (zip and) upload them to S3.

Part of https://github.com/aws/copilot-cli/issues/3546.

### DIFF
--- a/internal/pkg/addon/addons.go
+++ b/internal/pkg/addon/addons.go
@@ -12,6 +12,7 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/template"
 	"github.com/aws/copilot-cli/internal/pkg/workspace"
 	"github.com/dustin/go-humanize/english"
+	"github.com/spf13/afero"
 	"gopkg.in/yaml.v3"
 )
 
@@ -49,6 +50,8 @@ type Addons struct {
 
 	bucket   string
 	uploader uploader
+	wsPath   string
+	fs       *afero.Afero
 }
 
 // New creates an Addons struct given a workload name.
@@ -57,10 +60,15 @@ func New(wlName string) (*Addons, error) {
 	if err != nil {
 		return nil, fmt.Errorf("workspace cannot be created: %w", err)
 	}
+	wsPath, err := ws.Path()
+	if err != nil {
+		return nil, fmt.Errorf("get workspace path: %w", err)
+	}
 	return &Addons{
 		wlName: wlName,
 		parser: template.New(),
 		ws:     ws,
+		wsPath: wsPath,
 	}, nil
 }
 
@@ -75,6 +83,9 @@ func NewPackager(wlName string, bucket string, uploader uploader) (*Addons, erro
 
 	addons.bucket = bucket
 	addons.uploader = uploader
+	addons.fs = &afero.Afero{
+		Fs: afero.NewOsFs(),
+	}
 	return addons, nil
 }
 
--- a/internal/pkg/addon/addons.go
+++ b/internal/pkg/addon/addons.go
@@ -36,6 +36,7 @@ var (
 type workspaceReader interface {
 	ReadAddonsDir(svcName string) ([]string, error)
 	ReadAddon(svcName, fileName string) ([]byte, error)
+	Path() (string, error)
 }
 
 // Addons represents additional resources for a workload.
@@ -60,15 +61,10 @@ func New(wlName string) (*Addons, error) {
 	if err != nil {
 		return nil, fmt.Errorf("workspace cannot be created: %w", err)
 	}
-	wsPath, err := ws.Path()
-	if err != nil {
-		return nil, fmt.Errorf("get workspace path: %w", err)
-	}
 	return &Addons{
 		wlName: wlName,
 		parser: template.New(),
 		ws:     ws,
-		wsPath: wsPath,
 	}, nil
 }
 
@@ -79,6 +75,11 @@ func NewPackager(wlName string, bucket string, uploader uploader) (*Addons, erro
 	addons, err := New(wlName)
 	if err != nil {
 		return nil, err
+	}
+
+	addons.wsPath, err = addons.ws.Path()
+	if err != nil {
+		return nil, fmt.Errorf("get workspace path: %w", err)
 	}
 
 	addons.bucket = bucket
--- a/internal/pkg/addon/package.go
+++ b/internal/pkg/addon/package.go
@@ -4,11 +4,19 @@
 package addon
 
 import (
+	"archive/zip"
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"io"
+	"io/fs"
+	"path"
+	"path/filepath"
 	"strings"
 
 	"github.com/aws/copilot-cli/internal/pkg/aws/s3"
+	"github.com/aws/copilot-cli/internal/pkg/template/artifactpath"
 	"gopkg.in/yaml.v3"
 )
 
@@ -273,8 +281,129 @@ type s3Object struct {
 }
 
 func (a *Addons) uploadAddonAsset(assetPath string, forceZip bool) (s3Object, error) {
+	// make path absolute from wsPath
+	if !path.IsAbs(assetPath) {
+		assetPath = path.Join(a.wsPath, assetPath)
+	}
+
+	info, err := a.fs.Stat(assetPath)
+	if err != nil {
+		return s3Object{}, err
+	}
+
+	var asset asset
+	if forceZip || info.IsDir() {
+		asset, err = a.zipAsset(assetPath)
+	} else {
+		asset, err = a.fileAsset(assetPath)
+	}
+	if err != nil {
+		return s3Object{}, fmt.Errorf("create asset: %w", err)
+	}
+
+	s3Path := artifactpath.AddonAsset(a.wlName, asset.hash)
+
+	url, err := a.uploader.Upload(a.bucket, s3Path, asset.data)
+	if err != nil {
+		return s3Object{}, fmt.Errorf("upload %s to s3 bucket %s: %w", assetPath, a.bucket, err)
+	}
+
+	bucket, key, err := s3.ParseURL(url)
+	if err != nil {
+		return s3Object{}, fmt.Errorf("parse s3 url: %w", err)
+	}
+
+	fmt.Printf("Uploaded %s to s3 at: %s\n", assetPath, s3.Location(bucket, key))
 	return s3Object{
-		Bucket: a.bucket,
-		Key:    "TODO",
+		Bucket: bucket,
+		Key:    key,
+	}, nil
+}
+
+type asset struct {
+	data io.Reader
+	hash string
+}
+
+// zipAsset creates an asset from the directory or file specified by root
+// where the data is the compressed zip archive, and the hash is
+// a hash of each files name, permission, and content.
+func (a *Addons) zipAsset(root string) (asset, error) {
+	buf := &bytes.Buffer{}
+	z := zip.NewWriter(buf)
+	defer z.Close()
+
+	hash := sha256.New()
+
+	if err := a.fs.Walk(root, func(path string, info fs.FileInfo, err error) error {
+		switch {
+		case err != nil:
+			return err
+		case info.IsDir():
+			return nil
+		}
+
+		fname, err := filepath.Rel(root, path)
+		switch {
+		case err != nil:
+			return fmt.Errorf("rel: %w", err)
+		case fname == ".":
+			fname = info.Name()
+		}
+
+		f, err := a.fs.Open(path)
+		if err != nil {
+			return fmt.Errorf("open: %w", err)
+		}
+		defer f.Close()
+
+		header, err := zip.FileInfoHeader(info)
+		if err != nil {
+			return fmt.Errorf("create zip file header: %w", err)
+		}
+
+		header.Name = fname
+		header.Method = zip.Deflate
+
+		zf, err := z.CreateHeader(header)
+		if err != nil {
+			return fmt.Errorf("create zip file: %w", err)
+		}
+
+		// include the file name and permissions as part of the hash
+		hash.Write([]byte(fmt.Sprintf("%s %s", fname, info.Mode().String())))
+		_, err = io.Copy(io.MultiWriter(zf, hash), f)
+		return err
+	}); err != nil {
+		return asset{}, err
+	}
+
+	return asset{
+		data: buf,
+		hash: hex.EncodeToString(hash.Sum(nil)),
+	}, nil
+}
+
+// fileAsset creates an asset from the file specified by path.
+// The data is the content of the file, and the hash is the
+// a hash of the file content.
+func (a *Addons) fileAsset(path string) (asset, error) {
+	hash := sha256.New()
+	buf := &bytes.Buffer{}
+
+	f, err := a.fs.Open(path)
+	if err != nil {
+		return asset{}, fmt.Errorf("open: %w", err)
+	}
+	defer f.Close()
+
+	_, err = io.Copy(io.MultiWriter(buf, hash), f)
+	if err != nil {
+		return asset{}, fmt.Errorf("copy: %w", err)
+	}
+
+	return asset{
+		data: buf,
+		hash: hex.EncodeToString(hash.Sum(nil)),
 	}, nil
 }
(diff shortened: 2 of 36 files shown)


internal/pkg/addon/addons.go
```diff
@@ -73,8 +77,16 @@ func NewPackager(wlName string, bucket string, uploader uploader) (*Addons, erro
 		return nil, err
 	}
 
+	addons.wsPath, err = addons.ws.Path()
```
efekarakus, line 80: should we move these initializations to `New` instead? after addressing Wanxian's refactor this constructor will be gone

internal/pkg/addon/package.go
```diff
@@ -273,8 +281,129 @@ type s3Object struct {
 }
 
 func (a *Addons) uploadAddonAsset(assetPath string, forceZip bool) (s3Object, error) {
+	// make path absolute from wsPath
+	if !path.IsAbs(assetPath) {
+		assetPath = path.Join(a.wsPath, assetPath)
```
efekarakus: For our windows users  
```suggestion
		assetPath = filepath.Join(a.wsPath, assetPath)
```

internal/pkg/addon/package.go
```diff
@@ -273,8 +281,129 @@ type s3Object struct {
...
+	}
+
+	info, err := a.fs.Stat(assetPath)
+	if err != nil {
+		return s3Object{}, err
```
efekarakus: can we wrap the error? 🥺

internal/pkg/addon/package.go
```diff
@@ -273,8 +281,129 @@ type s3Object struct {
...
+	}
+
+	var asset asset
+	if forceZip || info.IsDir() {
+		asset, err = a.zipAsset(assetPath)
+	} else {
+		asset, err = a.fileAsset(assetPath)
+	}
```
efekarakus: nit: just a different way of writing it:
```go
var readAssetFn func(string) (asset, error) = a.fileAsset
if forceZip || info.IsDir() {
   readAssetFn = a.zipAsset
}
asset, err := readAssetFn(path)
```

internal/pkg/addon/package.go
```diff
@@ -273,8 +281,129 @@ type s3Object struct {
...
+	if err != nil {
+		return s3Object{}, fmt.Errorf("create asset: %w", err)
+	}
+
+	s3Path := artifactpath.AddonAsset(a.wlName, asset.hash)
+
```
efekarakus:
```suggestion
```

internal/pkg/addon/package.go
```diff
@@ -273,8 +281,129 @@ type s3Object struct {
...
+	if err != nil {
+		return s3Object{}, fmt.Errorf("upload %s to s3 bucket %s: %w", assetPath, a.bucket, err)
+	}
+
+	bucket, key, err := s3.ParseURL(url)
+	if err != nil {
+		return s3Object{}, fmt.Errorf("parse s3 url: %w", err)
+	}
+
+	fmt.Printf("Uploaded %s to s3 at: %s\n", assetPath, s3.Location(bucket, key))
```
efekarakus:
```suggestion
```

internal/pkg/addon/package.go
```diff
@@ -273,8 +281,129 @@ type s3Object struct {
...
+}
+
+type asset struct {
+	data io.Reader
+	hash string
+}
+
+// zipAsset creates an asset from the directory or file specified by root
+// where the data is the compressed zip archive, and the hash is
+// a hash of each files name, permission, and content.
```
efekarakus: Would you like to add an explanation for why it's not the hash of the zip file itself?

internal/pkg/addon/package.go
```diff
@@ -273,8 +281,129 @@ type s3Object struct {
...
+func (a *Addons) zipAsset(root string) (asset, error) {
+	buf := &bytes.Buffer{}
+	z := zip.NewWriter(buf)
+	defer z.Close()
+
+	hash := sha256.New()
+
+	if err := a.fs.Walk(root, func(path string, info fs.FileInfo, err error) error {
```
efekarakus, line 332: is there a `WalkDir` in afero? https://pkg.go.dev/path/filepath#Walk according to the docs it's more efficient

internal/pkg/addon/package.go
```diff
@@ -273,8 +281,129 @@ type s3Object struct {
...
+		switch {
+		case err != nil:
+			return err
+		case info.IsDir():
+			return nil
```
efekarakus, line 337: just to make although we return `nil`, `Walk` will still traverse down the directory right?

internal/pkg/addon/package.go
```diff
@@ -273,8 +281,129 @@ type s3Object struct {
...
+		}
+
+		fname, err := filepath.Rel(root, path)
```
efekarakus, line 340: I am prob dumb but isn't `path == filepath.Rel(root,path)`?

internal/pkg/addon/package.go
```diff
@@ -273,8 +281,129 @@ type s3Object struct {
...
+		}
+		defer f.Close()
+
+		header, err := zip.FileInfoHeader(info)
+		if err != nil {
+			return fmt.Errorf("create zip file header: %w", err)
+		}
+
+		header.Name = fname
+		header.Method = zip.Deflate
```
efekarakus, line 360: 👍 nice! For my edification, do we this because now the file is going to be written to the zip archive? like I wonder why the `CreateHeader` method doesn't do this by default internally

internal/pkg/addon/package.go
```diff
@@ -273,8 +281,129 @@ type s3Object struct {
...
+
+		zf, err := z.CreateHeader(header)
+		if err != nil {
+			return fmt.Errorf("create zip file: %w", err)
```
efekarakus: nit: should we inclue the `path` here in the msg

internal/pkg/addon/package.go
```diff
@@ -273,8 +281,129 @@ type s3Object struct {
...
+		}
+
+		// include the file name and permissions as part of the hash
+		hash.Write([]byte(fmt.Sprintf("%s %s", fname, info.Mode().String())))
+		_, err = io.Copy(io.MultiWriter(zf, hash), f)
```
efekarakus, line 369: very cool! 🚀 the whole flow of this function is really nice 👍

internal/pkg/addon/package.go
```diff
@@ -273,8 +281,129 @@ type s3Object struct {
...
+
+		header, err := zip.FileInfoHeader(info)
+		if err != nil {
+			return fmt.Errorf("create zip file header: %w", err)
+		}
+
+		header.Name = fname
+		header.Method = zip.Deflate
+
+		zf, err := z.CreateHeader(header)
```
efekarakus: nit: It's already readable but I thought I could share a guideline that I think is neat: https://a-nickels-worth.blogspot.com/2016/04/a-guide-to-naming-variables.html#orgheadline12 where we make the variable names longer the further away they're declared.

So for example, overhere it would be:
```suggestion
archive := zip.NewWriter(buf)
// ... above
		zf, err := archive.CreateHeader(header)
```

internal/pkg/addon/package_test.go
```diff
@@ -23,12 +32,41 @@ func TestPackage(t *testing.T) {
 		bucket = "mockBucket"
 	)
 
+	lambdaZipHash := sha256.New()
+	indexZipHash := sha256.New()
+	indexFileHash := sha256.New()
```
efekarakus, line 37: im dumb are these hash sums supposed to be compared somewhere 😅

internal/pkg/template/artifactpath/artifactpath.go
```diff
@@ -30,6 +31,12 @@ func Addons(key string, content []byte) string {
 	return path.Join(s3ArtifactDirName, s3ArtifactAddonsDirName, key, fmt.Sprintf("%x.yml", sha256.Sum256(content)))
 }
 
+// AddonAsset returns the path to store an addon asset file.
+// Example: manual/addons/addonKey/assets/hash.
+func AddonAsset(addonKey, hash string) string {
```
efekarakus: nit: just to make it clear the key is a workload name
```suggestion
// Example: manual/addons/frontend/assets/ba7816bf8f01.
func AddonAsset(key, hash string) string {
```

efekarakus (commented):
Looks great! 🤩 just some small things

internal/pkg/addon/package.go
dannyrandall: interesting! i like it!

internal/pkg/addon/package.go
dannyrandall: sadly, no 😔

internal/pkg/addon/package.go
dannyrandall: correct! you have to return [`fs.SkipDir`](https://pkg.go.dev/io/fs#SkipDir) to skip the directory but continue traversing (any other error stops entirely)

internal/pkg/addon/package_test.go
dannyrandall: they are used to create the `S3Path`s [here](https://github.com/aws/copilot-cli/pull/3826/files#diff-0344a1783d235315005ac2a895dc6548066195e3312402fa9850f698072420f9R57-R59) and then those are compared in the mock functions!

internal/pkg/addon/package.go
dannyrandall: which one are you asking about? `zip.Deflate`? maybe this answers your question 😊 
> FileInfoHeader creates a partially-populated FileHeader from an fs.FileInfo. Because fs.FileInfo's Name method returns only the base name of the file it describes, it may be necessary to modify the Name field of the returned header to provide the full path name of the file. If compression is desired, callers should set the FileHeader.Method field; it is unset by default.

internal/pkg/addon/addons.go
dannyrandall: moving it causes all of the `deploy/cloudformation/stack` integ tests to fail - they're not configured to "be" in a copilot workspace. i was just planning on leaving it until the refactor and we'll call `Path()` when we call `Package()`.

internal/pkg/addon/package.go
dannyrandall: not quite! `path` includes the root in the path, so if you take the filesystem from `package_test.go`:
```
//  .
//  ├─ lambda
//  │  ├─ index.js (contains lambda function)
//  ┴  └─ test.js (empty)
```
If you called `Walk("lambda")` (meaning the cfn for the resource is something like `Code: lambda/`), `path == "/lambda/index.js"`, but `Rel() == "index.js"`, which is the path we want in the zip.

Another example - if `lambda` was in directory named `parent`, and the `root` was `parent/lambda`, `path` would be `parent/lambda/index.js`, but we want the folder they passed in to be the "root" of the zip archive.

internal/pkg/addon/package_test.go
efekarakus: Oh I see what you mean!
We are validating that the right hashes are generated here: 
```
.Upload(bucket, indexZipS3Path /* ensure the right hash. */, gomock.Any())
```

internal/pkg/addon/package_test.go
```diff
@@ -23,12 +32,41 @@ func TestPackage(t *testing.T) {
...
+	io.MultiWriter(lambdaZipHash, indexZipHash).Write([]byte("index.js " + info.Mode().String()))
+	io.MultiWriter(f, lambdaZipHash, indexZipHash, indexFileHash).Write([]byte(`exports.handler = function(event, context) {}`))
+
+	f2, _ := fs.Create("/lambda/test.js")
+	info, _ = f2.Stat()
+	lambdaZipHash.Write([]byte("test.js " + info.Mode().String()))
+
+	lambdaZipS3Path := artifactpath.AddonAsset(wlName, hex.EncodeToString(lambdaZipHash.Sum(nil)))
+	indexZipS3Path := artifactpath.AddonAsset(wlName, hex.EncodeToString(indexZipHash.Sum(nil)))
+	indexFileS3Path := artifactpath.AddonAsset(wlName, hex.EncodeToString(indexFileHash.Sum(nil)))
```
efekarakus: For unit tests, I try to follow this rule as long as it doesn't make the test too difficult too difficult to follow:
>The lesson is clear: in test code, stick to straight-line code over clever logic, and consider tolerating some duplication when it makes the test more descriptive and meaningful.
> https://abseil.io/resources/swe-book/html/ch12.html#donapostrophet_put_logic_in_tests

I think the logic using the *ZipHash variables is fair, otherwise the reader of the test won't have any idea how the hashes are generated. But for the s3 paths, we can be slightly more explicit to ensure we're uploading to the correct location:
```go
wantedLambdaZipPath := fmt.Sprintf("manual/addons/mock-wl/assets/%s", hex.EncodeToString(lambdaZipHash.Sum(nil))) 
```

efekarakus (approved):

internal/pkg/addon/package.go
```diff
@@ -273,8 +280,128 @@ type s3Object struct {
...
+	if err != nil {
+		return s3Object{}, fmt.Errorf("upload %s to s3 bucket %s: %w", assetPath, a.bucket, err)
+	}
+
+	bucket, key, err := s3.ParseURL(url)
+	if err != nil {
+		return s3Object{}, fmt.Errorf("parse s3 url: %w", err)
+	}
+
 	return s3Object{
```
Lou1415926: what about `template.S3ObjectLocation`?

internal/pkg/addon/package.go
```diff
@@ -273,8 +280,128 @@ type s3Object struct {
...
+			return err
+		case info.IsDir():
+			return nil
+		}
+
+		fname, err := filepath.Rel(root, path)
+		switch {
+		case err != nil:
+			return fmt.Errorf("rel: %w", err)
+		case fname == ".":
```
Lou1415926: Does this happen when a `forceZip` is applied on a single file (as opposed to dir)?

internal/pkg/addon/package_test.go
```diff
@@ -23,12 +32,41 @@ func TestPackage(t *testing.T) {
...
+	io.MultiWriter(lambdaZipHash, indexZipHash).Write([]byte("index.js " + info.Mode().String()))
+	io.MultiWriter(f, lambdaZipHash, indexZipHash, indexFileHash).Write([]byte(`exports.handler = function(event, context) {}`))
+
+	f2, _ := fs.Create("/lambda/test.js")
+	info, _ = f2.Stat()
+	lambdaZipHash.Write([]byte("test.js " + info.Mode().String()))
+
+	lambdaZipS3Path := fmt.Sprintf("manual/addons/mock-wl/assets/%s", hex.EncodeToString(lambdaZipHash.Sum(nil)))
+	indexZipS3Path := fmt.Sprintf("manual/addons/mock-wl/assets/%s", hex.EncodeToString(indexZipHash.Sum(nil)))
+	indexFileS3Path := fmt.Sprintf("manual/addons/mock-wl/assets/%s", hex.EncodeToString(indexFileHash.Sum(nil)))
```
Lou1415926, line 59: I see. Am I summarizing correctly that
1. These three variables are the "correct answer" that we want to check the hash against
2. `lambdaZipS3Path` is the correct hash for resources that wants to upload the whole `/lambda` folder
3. `indexZipS3Path` is the correct hash for resources that wants to zip and upload the `/lambda/index.js` file
4. `indexFileS3Path` is the correct hash for resources that want to just upload the `/lambda.index.js` file (without zipping)

Lou1415926 (commented):
Looks greatt!!!! just one or two questions

internal/pkg/addon/package.go
dannyrandall: correct! i'll add a note about that :)

internal/pkg/addon/package_test.go
dannyrandall: yeah, that's correct. these three paths should cover all potential cases of file hashes (folder, file, and zipped file)

Lou1415926: Nice! Nothing to address here - I just want to share an alternative:
```go
m.uploader.Expect.Upload(bucket, path).DoAndReturn(func(bucket string, path string) s3Location {
   // Compare got path with expected hash.
   require.Equal(indexHash, path)
})
```
This is more verbose and results in longer test files than what you have without making real difference. However I guess the one nice thing on its semantic message to other developers: it conveys the intention to "validate the hash instead of doing simple mocks" here.

Lou1415926 (approved):

PR #3827, opened 2022-07-28
build: upgrade to go 1.18.4

Lou1415926:


.release/buildspec.yml
```diff
@@ -6,8 +6,8 @@ phases:
       nodejs: 12
     commands:
       - 'cd $HOME/.goenv && git pull --ff-only && cd -'
-      - 'goenv install 1.18.0'
-      - 'goenv global 1.18.0'
+      - 'goenv install 1.18.4'
```
efekarakus: :shipit: for folks reviewing, I always check if a version is available first here: https://github.com/syndbg/goenv/tree/master/plugins/go-build/share/go-build

There has been instances in the past where a particular version wasn't available in the repo

efekarakus (approved):

.release/buildspec.yml
dannyrandall: maybe we should add a link to that page 🤔

dannyrandall (approved):

PR #3830, opened 2022-07-29
build: run integration tests in a dedicated account

Lou1415926:

### DIFF
--- a/.release/buildspec_integ.yml
+++ b/.release/buildspec_integ.yml
@@ -22,9 +22,9 @@ phases:
       - echo `git rev-parse HEAD` # Do not delete; for pipeline logging purposes
       - cd $CODEBUILD_SRC_DIR
       - export GOPATH=/go
-      - mkdir -p /tmp/.aws
-      - printf "[default]\nregion=us-west-2\n> /tmp/.aws/config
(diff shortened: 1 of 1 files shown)


.release/buildspec_integ.yml
```diff
@@ -22,6 +22,9 @@ phases:
       - echo `git rev-parse HEAD` # Do not delete; for pipeline logging purposes
       - cd $CODEBUILD_SRC_DIR
       - export GOPATH=/go
+      - mkdir -p /tmp/.aws
```
efekarakus: how do we tell the AWS SDK to read the config file under here?
In the `buildspec_e2e.yml`, we have this: `docker run --privileged -v /tmp/.aws:/home/.aws -e "HOME=/home"`

I wonder if the `AWS_SHARED_CREDENTIALS_FILE` env var works with the aws sdk go to change the location otherwise we can copy to $HOME

Lou1415926: or I can just set the `AWS_ACCESS_KEY` env vars since we have only one profile for integ test anyway

efekarakus (approved):

dannyrandall (approved):

PR #3836, opened 2022-08-01
refactor(cloudformation): move fileWriter from methods to struct field

efekarakus:
Previously, each method in the `deploy/cloudformation` pkg took `os.Stderr` as a parameter to render events. 

This refactor moves the parameter to a struct field instead, and makes the progress tracker optional by defaulting to a `discardFile` as the target file for progress events.

### DIFF
--- a/internal/pkg/cli/deploy/svc.go
+++ b/internal/pkg/cli/deploy/svc.go
@@ -55,7 +55,6 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/repository"
 	"github.com/aws/copilot-cli/internal/pkg/term/color"
 	"github.com/aws/copilot-cli/internal/pkg/term/log"
-	"github.com/aws/copilot-cli/internal/pkg/term/progress"
 	termprogress "github.com/aws/copilot-cli/internal/pkg/term/progress"
 	"github.com/aws/copilot-cli/internal/pkg/workspace"
 )
@@ -131,7 +130,7 @@ type snsTopicsLister interface {
 }
 
 type serviceDeployer interface {
-	DeployService(out progress.FileWriter, conf cloudformation.StackConfiguration, bucketName string, opts ...awscloudformation.StackOption) error
+	DeployService(conf cloudformation.StackConfiguration, bucketName string, opts ...awscloudformation.StackOption) error
 }
 
 type serviceForceUpdater interface {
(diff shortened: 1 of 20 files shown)


internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -130,6 +150,10 @@ func New(sess *session.Session) CloudFormation {
 		appStackSet: stackset.New(sess),
 		s3Client:    s3.New(sess),
 		region:      aws.StringValue(sess.Config.Region),
+		console:     new(discardFile),
```
dannyrandall, line 153: nit: (just my preference)
```suggestion
		console:     &discardFile{},
```

dannyrandall (approved):
looks great!

internal/pkg/cli/deploy/svc.go
```diff
@@ -277,7 +276,7 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 		s3Client:           s3Client,
 		addons:             addonsSvc,
 		imageBuilderPusher: imageBuilderPusher,
-		deployer:           cloudformation.New(envSession),
+		deployer:           cloudformation.New(envSession, cloudformation.WithProgressTracker(os.Stderr)),
```
bvtujo, line 279: I really like this pattern; good refactor

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -130,6 +150,10 @@ func New(sess *session.Session) CloudFormation {
...
+	}
+	for _, opt := range opts {
+		opt(&client)
 	}
 	return client
```
bvtujo, line 158: Just a question for my own purposes: why do we use a struct receiver and not a pointer receiver for the cf methods? Do we want to disallow the cfn client from modifying itself?

bvtujo (approved):

internal/pkg/deploy/cloudformation/cloudformation.go
efekarakus: I have no idea why we did it that way! It's a good question, tbh if I were to write it today I'd make it a pointer receiver to slightly save on memory but if I had to guess your answer is correct!

PR #3837, opened 2022-08-02
chore: add cert-replicator lambda

iamhopaul123:
Note that it won't work for cert import case. I'll add a new field for cert import in the next PR.

### DIFF
--- a/cf-custom-resources/lib/cert-replicator.js
+++ b/cf-custom-resources/lib/cert-replicator.js
@@ -1,0 +1,324 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/* jshint node: true */
+/* jshint esversion: 8 */
+
+"use strict";
+
+const aws = require("aws-sdk");
+
+const defaultSleep = function (ms) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+};
+
+// These are used for test purposes only
+let defaultResponseURL;
+let defaultLogGroup;
+let defaultLogStream;
+let waiter;
+let sleep = defaultSleep;
+let maxAttempts = 10;
+
+/**
+ * Upload a CloudFormation response object to S3.
+ *
+ * @param {object} event the Lambda event payload received by the handler function
+ * @param {object} context the Lambda context received by the handler function
+ * @param {string} responseStatus the response status, either 'SUCCESS' or 'FAILED'
+ * @param {string} physicalResourceId CloudFormation physical resource ID
+ * @param {object} [responseData] arbitrary response data object
+ * @param {string} [reason] reason for failure, if any, to convey to the user
+ * @returns {Promise} Promise that is resolved on success, or rejected on connection error or HTTP error response
+ */
+let report = function (
+  event,
+  context,
+  responseStatus,
+  physicalResourceId,
+  responseData,
+  reason
+) {
+  return new Promise((resolve, reject) => {
+    const https = require("https");
+    const { URL } = require("url");
+
+    let responseBody = JSON.stringify({
+      Status: responseStatus,
+      Reason: reason,
+      PhysicalResourceId: physicalResourceId || context.logStreamName,
+      StackId: event.StackId,
+      RequestId: event.RequestId,
+      LogicalResourceId: event.LogicalResourceId,
+      Data: responseData,
+    });
+
+    const parsedUrl = new URL(event.ResponseURL || defaultResponseURL);
+    const options = {
+      hostname: parsedUrl.hostname,
+      port: 443,
+      path: parsedUrl.pathname + parsedUrl.search,
+      method: "PUT",
+      headers: {
+        "Content-Type": "",
+        "Content-Length": responseBody.length,
+      },
+    };
+
+    https
+      .request(options)
+      .on("error", reject)
+      .on("response", (res) => {
+        res.resume();
+        if (res.statusCode >= 400) {
+          reject(new Error(`Error ${res.statusCode}: ${res.statusMessage}`));
+        } else {
+          resolve();
+        }
+      })
+      .end(responseBody, "utf8");
+  });
+};
+
+/**
+ * Replicate a public certificate from AWS Certificate Manager in the target region.
+ *
+ * @param {string} requestId the CloudFormation request ID
+ * @param {string} appName the application name
+ * @param {string} envName the environment name
+ * @param {string} certArn arn for the certificate to replicate from
+ * @param {string} envRegionAcm acm client in environment region
+ * @param {string} targetRegionAcm acm client in target region
+ * @returns {string} ARN of the replicated certificate
+ */
+const replicateCertificate = async function (
+  requestId,
+  appName,
+  envName,
+  certArn,
+  envRegionAcm,
+  targetRegionAcm
+) {
+  const { Certificate } = await envRegionAcm
+    .describeCertificate({
+      CertificateArn: certArn,
+    })
+    .promise();
+  const domainName = Certificate.DomainName;
+  const sans = Certificate.SubjectAlternativeNames;
+
+  const crypto = require("crypto");
+  return targetRegionAcm
+    .requestCertificate({
+      DomainName: domainName,
+      SubjectAlternativeNames: sans,
+      IdempotencyToken: crypto
+        .createHash("sha256")
+        .update(requestId)
+        .digest("hex")
+        .substr(0, 32),
+      ValidationMethod: "DNS",
+      Tags: [
+        {
+          Key: "copilot-application",
+          Value: appName,
+        },
+        {
+          Key: "copilot-environment",
+          Value: envName,
+        },
+      ],
+    })
+    .promise();
+};
+
+/**
+ * Deletes a certificate from AWS Certificate Manager (ACM) by its ARN.
+ * Specifically, if it is the last certificate attaching to the listener, it will also remove the CNAME records
+ * for validation in all the root, app, and env hosted zones in parallel.
+ * If the certificate does not exist, the function will return normally.
+ *
+ * @param {string} arn The certificate ARN
+ * @param {string} targetRegionAcm acm client in target region
+ */
+const deleteCertificate = async function (arn, acm) {
+  try {
+    console.log(`Waiting for certificate ${arn} to become unused`);
+
+    let inUseByResources;
+    let options;
+
+    for (let attempt = 0; attempt < maxAttempts; attempt++) {
+      const { Certificate } = await acm
+        .describeCertificate({
+          CertificateArn: arn,
+        })
+        .promise();
+
+      inUseByResources = Certificate.InUseBy || [];
+      options = Certificate.DomainValidationOptions || [];
+      let ok = false;
+      for (const option of options) {
+        if (!option.ResourceRecord) {
+          ok = false;
+          break;
+        }
+        ok = true;
+      }
+      if (!ok || inUseByResources.length) {
+        // Deleting resources can be quite slow - so just sleep 30 seconds between checks.
+        await sleep(30000);
+      } else {
+        break;
+      }
+    }
+    if (inUseByResources.length) {
+      throw new Error(
+        `Certificate still in use after checking for ${maxAttempts} attempts.`
+      );
+    }
+
+    await acm
+      .deleteCertificate({
+        CertificateArn: arn,
+      })
+      .promise();
+  } catch (err) {
+    if (err.name !== "ResourceNotFoundException") {
+      throw err;
+    }
+  }
+};
+
+const validateCertificate = async function (certificateARN, acm) {
+  await acm
+    .waitFor("certificateValidated", {
+      // Wait up to 9 minutes and 30 seconds
+      $waiter: {
+        delay: 30,
+        maxAttempts: 19,
+      },
+      CertificateArn: certificateARN,
+    })
+    .promise();
+};
+
+const acmClient = function (region) {
+  const acm = new aws.ACM({
+    region,
+  });
+  if (waiter) {
+    // Used by the test suite, since waiters aren't mockable yet
+    acm.waitFor = waiter;
+  }
+  return acm;
+};
+
+/**
+ * Main certificate replicator handler, invoked by Lambda
+ */
+exports.certificateReplicateHandler = async function (event, context) {
+  let responseData = {};
+  let physicalResourceId = event.PhysicalResourceId;
+  const props = event.ResourceProperties;
+  const [targetRegion, envRegion, certArn] = [
+    props.TargetRegion,
+    props.EnvRegion,
+    props.CertificateArn,
+  ];
+
+  const envRegionAcm = acmClient(envRegion);
+  const targetRegionAcm = acmClient(targetRegion);
+
+  try {
+    let response = {};
+    switch (event.RequestType) {
+      case "Create":
+      case "Update":
+        response = await replicateCertificate(
+          event.RequestId,
+          props.AppName,
+          props.EnvName,
+          certArn,
+          envRegionAcm,
+          targetRegionAcm
+        );
+        responseData.Arn = physicalResourceId = response.CertificateArn;
+        await validateCertificate(response.CertificateArn, targetRegionAcm);
+        break;
+      case "Delete":
+        // If the resource didn't create correctly, the physical resource ID won't be the
+        // certificate ARN, so don't try to delete it in that case.
+        if (physicalResourceId.startsWith("arn:")) {
+          await deleteCertificate(physicalResourceId, targetRegionAcm);
+        }
+        break;
+      default:
+        throw new Error(`Unsupported request type ${event.RequestType}`);
+    }
+    await report(event, context, "SUCCESS", physicalResourceId, responseData);
+  } catch (err) {
+    console.log(`Caught error ${err}.`);
+    await report(
+      event,
+      context,
+      "FAILED",
+      physicalResourceId,
+      null,
+      `${err.message} (Log: ${defaultLogGroup || context.logGroupName}/${
+        defaultLogStream || context.logStreamName
+      })`
+    );
+  }
+};
+
+/**
+ * @private
+ */
+exports.withDefaultResponseURL = function (url) {
+  defaultResponseURL = url;
+};
+
+/**
+ * @private
+ */
+exports.withWaiter = function (w) {
+  waiter = w;
+};
+
+/**
+ * @private
+ */
+exports.withSleep = function (s) {
+  sleep = s;
+};
+
+/**
+ * @private
+ */
+exports.reset = function () {
+  sleep = defaultSleep;
+  waiter = undefined;
+  maxAttempts = 10;
+};
+
+/**
+ * @private
+ */
+exports.withMaxAttempts = function (ma) {
+  maxAttempts = ma;
+};
+
+/**
+ * @private
+ */
+exports.withDefaultLogStream = function (logStream) {
+  defaultLogStream = logStream;
+};
+
+/**
+ * @private
+ */
+exports.withDefaultLogGroup = function (logGroup) {
+  defaultLogGroup = logGroup;
+};
--- a/cf-custom-resources/lib/cert-replicator.js
+++ b/cf-custom-resources/lib/cert-replicator.js
@@ -45,7 +45,10 @@ let report = function (
     let responseBody = JSON.stringify({
       Status: responseStatus,
       Reason: reason,
-      PhysicalResourceId: physicalResourceId,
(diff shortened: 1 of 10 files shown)


cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,324 @@
...
+ * Specifically, if it is the last certificate attaching to the listener, it will also remove the CNAME records
+ * for validation in all the root, app, and env hosted zones in parallel.
+ * If the certificate does not exist, the function will return normally.
+ *
+ * @param {string} arn The certificate ARN
+ * @param {string} targetRegionAcm acm client in target region
+ */
+const deleteCertificate = async function (arn, acm) {
+  try {
+    console.log(`Waiting for certificate ${arn} to become unused`);
```
Lou1415926: Is this log crucial? I have the impression that we want to log as little as possible because it's costs a lot.

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,324 @@
...
+      );
+    }
+
+    await acm
+      .deleteCertificate({
+        CertificateArn: arn,
+      })
+      .promise();
+  } catch (err) {
+    if (err.name !== "ResourceNotFoundException") {
```
Lou1415926, line 172: If this error is thrown by `describeCertificate`, what do you think of removing this large `try-catch` block, and instead do
```go
var certificate
try {
  { Certificate: certificate } = await acm
        .describeCertificate({
          CertificateArn: arn,
        })
        .promise();
} catch (errr) {
     if (err.name !== "ResourceNotFoundException")
}
```

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,324 @@
...
+          ok = false;
+          break;
+        }
+        ok = true;
+      }
+      if (!ok || inUseByResources.length) {
+        // Deleting resources can be quite slow - so just sleep 30 seconds between checks.
+        await sleep(30000);
+      } else {
+        break;
```
Lou1415926: nit: what about
```suggestion
      if (ok && inUseByResources.length === 0) {
         break;
      }
       // Deleting resources can be quite slow - so just sleep 30 seconds between checks.
        await sleep(30000);
  
```

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,324 @@
...
+    for (let attempt = 0; attempt < maxAttempts; attempt++) {
+      const { Certificate } = await acm
+        .describeCertificate({
+          CertificateArn: arn,
+        })
+        .promise();
+
+      inUseByResources = Certificate.InUseBy || [];
+      options = Certificate.DomainValidationOptions || [];
+      let ok = false;
```
Lou1415926: Dump question 🖐🏼 ! It seems like `ok` means:
1. Either there are no validation option left
2. Or there are, but all validation option does not have `ResourceRecord` (which is a CNAME record that gets upsert to validate a domain)

Can you help me understand why we need this to decide whether we can delete the certificate?

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,324 @@
...
+      CertificateArn: certificateARN,
+    })
+    .promise();
+};
+
+const acmClient = function (region) {
+  const acm = new aws.ACM({
+    region,
+  });
+  if (waiter) {
```
Lou1415926: seems to be mockable now! https://github.com/aws/copilot-cli/blob/0a64bf63801d18d00dc84da00fc1904a90076700/cf-custom-resources/test/nlb-cert-validator-test.js#L360

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,324 @@
...
+  }
+  return acm;
+};
+
+/**
+ * Main certificate replicator handler, invoked by Lambda
+ */
+exports.certificateReplicateHandler = async function (event, context) {
+  let responseData = {};
+  let physicalResourceId = event.PhysicalResourceId;
```
Lou1415926, line 196: Just double checking, when it's `CREATE` (i.e. no physical resource exists yet), `event.PhysicalResourceId` will be `null` or `""` right?

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,324 @@
...
+        // If the resource didn't create correctly, the physical resource ID won't be the
+        // certificate ARN, so don't try to delete it in that case.
+        if (physicalResourceId.startsWith("arn:")) {
+          await deleteCertificate(physicalResourceId, targetRegionAcm);
+        }
+        break;
+      default:
+        throw new Error(`Unsupported request type ${event.RequestType}`);
+    }
+    await report(event, context, "SUCCESS", physicalResourceId, responseData);
```
Lou1415926, line 234: We probably want to error out if this runs more than 15 minutes: https://github.com/aws/copilot-cli/blob/0a64bf63801d18d00dc84da00fc1904a90076700/cf-custom-resources/lib/nlb-cert-validator.js#L748

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,324 @@
...
+  reason
+) {
+  return new Promise((resolve, reject) => {
+    const https = require("https");
+    const { URL } = require("url");
+
+    let responseBody = JSON.stringify({
+      Status: responseStatus,
+      Reason: reason,
+      PhysicalResourceId: physicalResourceId || context.logStreamName,
```
Lou1415926: Do you think we can:
```suggestion
      PhysicalResourceId: physicalResourceId,
```

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -17,8 +17,40 @@ CloudFrontDistribution:
             OriginProtocolPolicy: match-viewer # Will need to change with TLS termination and Alias usage
           DomainName: !GetAtt PublicLoadBalancer.DNSName
           Id: !Sub 'copilot-${AppName}-${EnvironmentName}-origin'
-      # ViewerCertificate: # Used for Aliases
-      #   AcmCertificateArn: # Imported certificate goes here
-      #   CloudFrontDefaultCertificate: true # If importing certificate, set to false
-      #   MinimumProtocolVersion: TLSv1 # Default for now, may need to change later
-      #   SsLSupportMethod: sni-only # Changing this can incur costs to the user
+{{- if not .PublicHTTPConfig.ImportedCertARNs}}
```
Lou1415926, line 20: In a follow up PR, this will also take the `cdn.cert` field in manifest into consideration, is this understanding correct?

Lou1415926 (commented):
LGTM! Just a few questions and nits

CaptainCarpensir (commented):
Read through the lambda function, but as javascript is new to me I'm not comfortable reviewing that. O/w template changes look good!

internal/pkg/template/templates/environment/partials/cdn-resources.yml
CaptainCarpensir: I imagine there will be an `{{- else}}` case at the end of this to cover the viewer certificate for when the user imports certificates

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -17,8 +17,40 @@ CloudFrontDistribution:
...
+      S3Bucket: {{$cr.Bucket}}
+      S3Key: {{$cr.Key}}
+    {{- end}}
+    Handler: "index.certificateReplicateHandler"
+    Timeout: 900
+    MemorySize: 512
+    Role: !GetAtt 'CustomResourceRole.Arn'
+    Runtime: nodejs12.x
+    
+CertificateReplicator:
```
CaptainCarpensir, line 44: nit
```suggestion
ReplicatedCertificate:
```

internal/pkg/template/templates/environment/partials/cdn-resources.yml
iamhopaul123: Yeah it will be the next pr

cf-custom-resources/lib/cert-replicator.js
iamhopaul123: I migrated this from cert validator function but i guess we don't actually need domain validation options here because they are not related to this lambda.

cf-custom-resources/lib/cert-replicator.js
iamhopaul123: yeah that's right

cf-custom-resources/lib/cert-replicator.js
efekarakus: Same questions! I think it's fine if we don't delete if the certificate is indeed used by other services 👍 But why are we checking domain validation options? I would expect that to be always populated 🤔

cf-custom-resources/lib/cert-replicator.js
iamhopaul123: this is to make the error msg slightly more helpful. because at least they'll be easier to find out where to look for the lambda logs, instead of just an empty string.

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,324 @@
...
+      if (!ok || inUseByResources.length) {
+        // Deleting resources can be quite slow - so just sleep 30 seconds between checks.
+        await sleep(30000);
+      } else {
+        break;
+      }
+    }
+    if (inUseByResources.length) {
+      throw new Error(
+        `Certificate still in use after checking for ${maxAttempts} attempts.`
```
efekarakus:
```suggestion
        `Certificate still in use by ${inUseByResources.join()} after checking for ${maxAttempts} attempts .`
```

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,324 @@
...
+    props.TargetRegion,
+    props.EnvRegion,
+    props.CertificateArn,
+  ];
+
+  const envRegionAcm = acmClient(envRegion);
+  const targetRegionAcm = acmClient(targetRegion);
+
+  try {
+    let response = {};
```
efekarakus: nit: can we move this inside the "Update" case?
```js
const response = await replicateCertificate(...)
```

cf-custom-resources/lib/cert-replicator.js
iamhopaul123: It could also be thrown by `deleteCertificate` right 🤔

cf-custom-resources/lib/cert-replicator.js
efekarakus: yeah, what happens if it's a `Create` event and `replicateCertificate` fails. Then we will call `report` with an `undefined` physical ID.
 I think that's why we need:
 ```js
 PhysicalResourceId: physicalResourceId || context.logStreamName,
 ```

internal/pkg/template/templates/environment/partials/cdn-resources.yml
iamhopaul123: Technically this is still a custom resource. It is just we set the physical resource ARN to be the same as the replicated cert. I'd still prefer to use this name also to align with the other custom resource action.

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,324 @@
...
+        response = await replicateCertificate(
+          event.RequestId,
+          props.AppName,
+          props.EnvName,
+          certArn,
+          envRegionAcm,
+          targetRegionAcm
+        );
+        responseData.Arn = physicalResourceId = response.CertificateArn;
+        await validateCertificate(response.CertificateArn, targetRegionAcm);
```
efekarakus, line 219: I'm trying to figure out if setting the physical resource ID before validating the certificate is the right thing to do:

1. It's an `Update` event.
2.  ✅  `replicateCertificate` is successfully and we update the `physicalResourceId` successfully 
3. ❌ `validateCertificate` then fails for some reason.
4. We go into the `catch` block with the **new** physical id.

In this scenario, _I think_ on rollback we will delete the new cert and the rollback will complete 👍 Can you validate if this is right?

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -17,8 +17,40 @@ CloudFrontDistribution:
...
+  Metadata:
+    'aws:copilot:description': 'Replicate certificate to us-east-1'
```
efekarakus, line 46: Can we create these resources on the condition that the `AWS::Region` isn't us-east-1?

cf-custom-resources/lib/cert-replicator.js
iamhopaul123: Yeah that's exactly the reason why we set the `physicalResourceId` early. Otherwise even if `validateCertificate` fails, the new cert won't be removed by CFN

cf-custom-resources/lib/cert-replicator.js
Lou1415926: I guess it'd just be clearer to reader which calls are throwing `ResourceNotFoundException` and what errors we are handling. This huge catch block read a little confusing to me in the first glance 😂 But it could just be me.

cf-custom-resources/lib/cert-replicator.js
Lou1415926: The log stream name is already in the `reason` though! I'm just not super sure if we should set `PhysicalResourceId` to a log stream name but judging from how we handle `Delete` it seems to be okay.

internal/pkg/template/templates/environment/partials/cdn-resources.yml
CaptainCarpensir: I think it makes more sense to have a resource which represents "the certificate used to validate cloudfront" rather than potentially having a certificate which is created in the specific case it's needed. The drawback here is that we may create more resources than we strictly need, but I think the simplifications to the code are worth it, especially that it helps allow the user to understand what we're using the stack for better.

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,320 @@
...
+        },
+      ],
+    })
+    .promise();
+};
+
+/**
+ * Deletes a certificate from AWS Certificate Manager (ACM) by its ARN.
+ * Specifically, if it is the last certificate attaching to the listener, it will also remove the CNAME records
+ * for validation in all the root, app, and env hosted zones in parallel.
```
Lou1415926: nit: it seems like we are not doing these here
```suggestion
```

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,320 @@
...
+      })
+      .promise();
+  } catch (err) {
+    if (err.name !== "ResourceNotFoundException") {
+      throw err;
+    }
+  }
+};
+
+const validateCertificate = async function (certificateARN, acm) {
```
Lou1415926, line 178: How are the validation records upserted? Or do we not need to because the original cert's validation records would be the same as this one, meaning that this replicated cert has already been validated before it was created?

cf-custom-resources/test/cert-replicator-test.js
```diff
@@ -0,0 +1,528 @@
...
+      });
+  });
+
+  test("Create operation fails if cannot describe old certificate", () => {
+    const describeCertificateFake = sinon.fake.rejects(new Error("some error"));
+
+    AWS.mock("ACM", "describeCertificate", describeCertificateFake);
+
+    const request = nock(ResponseURL)
+      .put("/", (body) => {
```
Lou1415926, line 118: Do you think we also want to match the `PhysicalResourceID` here? That it's a log stream name

cf-custom-resources/test/cert-replicator-test.js
```diff
@@ -0,0 +1,528 @@
...
+          })
+        );
+        sinon.assert.calledWith(
+          requestCertificateFake,
+          sinon.match({
+            DomainName: testDomainName,
+            SubjectAlternativeNames: testSANs,
+            ValidationMethod: "DNS",
+            Tags: testCopilotTags,
+          })
```
Lou1415926, line 195: nice tests!

cf-custom-resources/test/cert-replicator-test.js
```diff
@@ -0,0 +1,528 @@
...
+    const waitForCertificateValidationFake = sinon.stub();
+    waitForCertificateValidationFake.resolves();
+
+    AWS.mock("ACM", "requestCertificate", requestCertificateFake);
+    AWS.mock("ACM", "describeCertificate", describeCertificateFake);
+    AWS.mock("ACM", "waitFor", waitForCertificateValidationFake);
+
+    const request = nock(ResponseURL)
+      .put("/", (body) => {
+        return body.Status === "SUCCESS";
```
Lou1415926: Similarly, what do you think of matching the `PhysicalResourceID` here as well, that it is indeed the requested cert ARN?

cf-custom-resources/test/cert-replicator-test.js
```diff
@@ -0,0 +1,528 @@
...
+      .event({
+        RequestType: "Delete",
+        RequestId: testRequestId,
+        PhysicalResourceId: testCertificateArn,
+        ResourceProperties: {
+          AppName: testAppName,
+          EnvName: testEnvName,
+          TargetRegion: "us-east-1",
+          EnvRegion: "us-west-2",
+          CertificateArn: testCertificateArn,
```
Lou1415926, line 311: Do you think we want to use different mock ARN for `PhysicalResourceId` and `CertfificateArn`? Just to test that we are making the calls with the correct ARNs. I think your implementation is using the correct ARNs but maybe just to be future proof!

Lou1415926 (approved):

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,320 @@
...
+  try {
+    await Promise.race([exports.deadlineExpired(), handler()]);
+    await report(event, context, "SUCCESS", physicalResourceId, responseData);
+  } catch (err) {
+    console.log(`Caught error ${err}.`);
+    await report(
+      event,
+      context,
+      "FAILED",
+      physicalResourceId,
```
efekarakus, line 241: so if it's a `CREATE` event and  `replicateCertificate` fails then the `physicalResourceId` will be `undefined`. What happens then?
Do we need to keep:
```
let physicalResourceID = event.PhysicalResourceId || context.logStreamName; 
```

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,320 @@
...
+ * Specifically, if it is the last certificate attaching to the listener, it will also remove the CNAME records
+ * for validation in all the root, app, and env hosted zones in parallel.
+ * If the certificate does not exist, the function will return normally.
+ *
+ * @param {string} arn The certificate ARN
+ * @param {string} targetRegionAcm acm client in target region
+ */
+const deleteCertificate = async function (arn, acm) {
+  try {
+    let inUseByResources;
```
efekarakus: nit:
```suggestion
    let inUseByResources = [];
```

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -0,0 +1,320 @@
...
+
+    for (let attempt = 0; attempt < maxAttempts; attempt++) {
+      const { Certificate } = await acm
+        .describeCertificate({
+          CertificateArn: arn,
+        })
+        .promise();
+
+      inUseByResources = Certificate.InUseBy || [];
+      if (!inUseByResources.length) {
```
efekarakus: nit: just to make it clear
```suggestion
      if (inUseByResources.length === 0) {
```

cf-custom-resources/test/cert-replicator-test.js
efekarakus: 🥺 yes please, it be great to have a testcase that ensures this invariant: https://github.com/aws/copilot-cli/pull/3837/files#r936022666

cf-custom-resources/lib/cert-replicator.js
iamhopaul123: Yeah that's true.

cf-custom-resources/test/cert-replicator-test.js
iamhopaul123: I updated my PR and we don't use log stream name for physical resource ID now

cf-custom-resources/lib/cert-replicator.js
iamhopaul123: I guess it'll be fine? Just checked the other lambda i'll keep it for consistency

efekarakus (approved):

PR #3842, opened 2022-08-02
chore: allow domain lambda to support a cloudfront distribution

CaptainCarpensir:
Allows the CloudFront DNS to be passed into the custom domain lambda, and then validates the case where the domain lambda would create the A records based off default values/service stacks (is not compatible with CloudFront, CloudFront needs to have all aliases specified)

Part of #3701 milestone 1 to get CloudFront working with custom domains


internal/pkg/cli/env_deploy.go
```diff
@@ -123,6 +136,15 @@ func (o *deployEnvOpts) Execute() error {
 	if err != nil {
 		return err
 	}
+	if mft.CDNConfig.CDNEnabled() && mft.HTTPConfig.Public.Certificates == nil && o.targetApp.Domain != "" {
```
iamhopaul123, line 139: Note that we might need a dynamic manifest struct for env manifest (just like the one we do for the workload manifest) so as to resolve dynamic values by making API calls. maybe add a TODO?

internal/pkg/describe/env.go
```diff
@@ -317,6 +317,39 @@ func (d *EnvDescriber) filterDeployedJobs() ([]*config.Workload, error) {
 	return deployedJobs, nil
 }
 
+// ValidateCFServiceDomainAliases returns error if an environment using cdn is deployed without specifying http.alias for all services
+func (d *EnvDescriber) ValidateCFServiceDomainAliases() error {
+
+	stackDescr, err := d.cfn.Describe()
+	if err != nil {
+		return err
```
iamhopaul123: Could you wrap this error msg from a public function?

internal/pkg/describe/env.go
```diff
@@ -317,6 +317,39 @@ func (d *EnvDescriber) filterDeployedJobs() ([]*config.Workload, error) {
...
+	}
+
+	services, err := d.deployStore.ListDeployedServices(d.app, d.env.Name)
+	if err != nil {
+		return fmt.Errorf("list services: %w", err)
+	}
+
+	if jsonOutput, ok := stackDescr.Parameters[envParamAliases]; ok {
```
iamhopaul123:
```suggestion
	jsonOutput, ok := stackDescr.Parameters[envParamAliases]
	if !ok {
	   return fmt.error...
	}
```

internal/pkg/describe/env.go
```diff
@@ -317,6 +317,39 @@ func (d *EnvDescriber) filterDeployedJobs() ([]*config.Workload, error) {
...
+		var aliases map[string][]string
+		err := json.Unmarshal([]byte(jsonOutput), &aliases)
+
+		if err != nil {
+			return fmt.Errorf("failed to unmarshal \"%s\": %w", jsonOutput, err)
```
iamhopaul123:
```suggestion
			return fmt.Errorf("unmarshal \"%s\": %w", jsonOutput, err)
```

internal/pkg/describe/env.go
```diff
@@ -317,6 +317,39 @@ func (d *EnvDescriber) filterDeployedJobs() ([]*config.Workload, error) {
...
 
+// ValidateCFServiceDomainAliases returns error if an environment using cdn is deployed without specifying http.alias for all services
+func (d *EnvDescriber) ValidateCFServiceDomainAliases() error {
+
+	stackDescr, err := d.cfn.Describe()
+	if err != nil {
+		return err
+	}
+
+	services, err := d.deployStore.ListDeployedServices(d.app, d.env.Name)
```
iamhopaul123: Instead of calling this API we might need to grab from the `ALBWorkloads` param from the env parameters. Because `ListDeployedServices` includes services of the other types than just LBWS.

internal/pkg/describe/env.go
```diff
@@ -317,6 +317,39 @@ func (d *EnvDescriber) filterDeployedJobs() ([]*config.Workload, error) {
...
+		var aliases map[string][]string
+		err := json.Unmarshal([]byte(jsonOutput), &aliases)
+
+		if err != nil {
+			return fmt.Errorf("failed to unmarshal \"%s\": %w", jsonOutput, err)
+		}
+
+		for _, service := range services {
+			if _, ok := aliases[service]; !ok {
+				return fmt.Errorf("all services deployed in an environment with CloudFront enabled must have http.alias specified")
```
iamhopaul123: This rule should only be applied to Load Balanced Web Service.

internal/pkg/describe/env.go
```diff
@@ -317,6 +317,39 @@ func (d *EnvDescriber) filterDeployedJobs() ([]*config.Workload, error) {
...
+			}
+		}
+	} else {
+		return fmt.Errorf("all services deployed in an environment with CloudFront enabled must have http.alias specified")
```
iamhopaul123: If it is not ok to get the parameter we might want to throw a diffferent error? Like cannot find xxx in the env stack parameter set.

internal/pkg/describe/env.go
```diff
@@ -317,6 +317,39 @@ func (d *EnvDescriber) filterDeployedJobs() ([]*config.Workload, error) {
...
+
+	jsonOutput, ok := stackDescr.Parameters[cfnstack.EnvParamAliasesKey]
+	if !ok {
+		return fmt.Errorf("cannot find %s in env stack parameter set", cfnstack.EnvParamAliasesKey)
+	}
+
+	var aliases map[string][]string
+	err = json.Unmarshal([]byte(jsonOutput), &aliases)
+	if err != nil {
+		return fmt.Errorf("unmarshal \"%s\": %w", jsonOutput, err)
```
iamhopaul123:
```suggestion
		return fmt.Errorf("unmarshal %q: %w", jsonOutput, err)
```

internal/pkg/cli/env_deploy.go
```diff
@@ -123,6 +136,15 @@ func (o *deployEnvOpts) Execute() error {
...
+		describer, err := o.newEnvDescriber()
+		if err != nil {
+			return err
+		}
+		if err := describer.ValidateCFServiceDomainAliases(); err != nil {
```
iamhopaul123, line 144: we need to wrap this error too.

internal/pkg/describe/env.go
```diff
@@ -317,6 +317,39 @@ func (d *EnvDescriber) filterDeployedJobs() ([]*config.Workload, error) {
...
+	}
+	for _, service := range services {
+		if _, ok := aliases[service]; !ok {
+			return fmt.Errorf("all lb web services deployed in an environment with CloudFront enabled must have http.alias specified")
```
iamhopaul123: Maybe we can give a more helpful error message telling users which ones don't have `http.alias` specified?

internal/pkg/cli/env_deploy.go
CaptainCarpensir: I don't grasp what this means exactly, can you write a TODO comment I can add there since rn I don't have the time to research what that means?

iamhopaul123: Oh yeah just a

```
// TODO: Put into a new DynamicEnvironmentManifest struct in "manifest" package
```

internal/pkg/describe/errors.go
```diff
@@ -15,3 +20,13 @@ type ErrManifestNotFoundInTemplate struct {
 func (err *ErrManifestNotFoundInTemplate) Error() string {
 	return fmt.Sprintf("manifest metadata not found in template of stack %s-%s-%s", err.app, err.env, err.name)
 }
+
+type errLbWebSvcsOnCFWithoutAlias struct {
```
iamhopaul123:
```suggestion
type errLBWebSvcsOnCFWithoutAlias struct {
```
nit

internal/pkg/describe/errors.go
```diff
@@ -15,3 +20,13 @@ type ErrManifestNotFoundInTemplate struct {
...
+	services []string
+}
+
+// Error implements the error interface.
+func (err *errLbWebSvcsOnCFWithoutAlias) Error() string {
+	return fmt.Sprintf("%v %v must have http.alias specified when CloudFront is enabled", english.PluralWord(len(err.services), "service", "services"),
```
iamhopaul123: nice!

iamhopaul123:
```suggestion
	return fmt.Sprintf("%v %v must have %q specified when CloudFront is enabled", english.PluralWord(len(err.services), "service", "services"),
```
maybe it'd be better to quote "http.alias"

iamhopaul123 (approved):

internal/pkg/cli/env_deploy.go
```diff
@@ -123,6 +136,16 @@ func (o *deployEnvOpts) Execute() error {
 	if err != nil {
 		return err
 	}
+	// TODO: Put into a new DynamicEnvironmentManifest struct in "manifest" package
```
dannyrandall: is this getting resolved in your next PR? (if not and it's just like a "future would-be-nice TODO, let's get rid of it)

internal/pkg/cli/env_deploy_test.go
```diff
@@ -215,6 +225,7 @@ func TestDeployEnvOpts_Execute(t *testing.T) {
 				m.deployer.EXPECT().UploadArtifacts().Return(map[string]string{
 					"mockResource": "mockURL",
 				}, nil)
+				m.describer.EXPECT().ValidateCFServiceDomainAliases().Return(nil).Times(0)
```
dannyrandall: i'm _pretty sure_ omitting this does the same thing as `.Times(0)` (the test will fail if an unexpected function is called)

internal/pkg/cli/env_deploy_test.go
```diff
@@ -170,6 +172,14 @@ func TestDeployEnvOpts_Execute(t *testing.T) {
 			},
 			wantedErr: errors.New(`unmarshal environment manifest for "mockEnv"`),
 		},
+		"fail to validate cdn": {
+			setUpMocks: func(m *deployEnvExecuteMocks) {
+				m.ws.EXPECT().ReadEnvironmentManifest(gomock.Any()).Return([]byte("name: mockEnv\ntype: Environment\ncdn: true\n"), nil)
+				m.interpolator.EXPECT().Interpolate(gomock.Any()).Return("name: mockEnv\ntype: Environment\ncdn: true\n", nil)
+				m.describer.EXPECT().ValidateCFServiceDomainAliases().Return(fmt.Errorf("all services deployed in an environment with CloudFront enabled must have http.alias specified"))
```
dannyrandall: nit: since you're just testing a fake error anyways, might be easier to have something shorter
```suggestion
				m.describer.EXPECT().ValidateCFServiceDomainAliases().Return(errors.New("mock error"))
```

internal/pkg/describe/env.go
```diff
@@ -317,6 +317,46 @@ func (d *EnvDescriber) filterDeployedJobs() ([]*config.Workload, error) {
 	return deployedJobs, nil
 }
 
+// ValidateCFServiceDomainAliases returns error if an environment using cdn is deployed without specifying http.alias for all services
+func (d *EnvDescriber) ValidateCFServiceDomainAliases() error {
+
```
dannyrandall:
```suggestion
```

internal/pkg/describe/env.go
```diff
@@ -317,6 +317,46 @@ func (d *EnvDescriber) filterDeployedJobs() ([]*config.Workload, error) {
 	return deployedJobs, nil
 }
 
+// ValidateCFServiceDomainAliases returns error if an environment using cdn is deployed without specifying http.alias for all services
```
dannyrandall:
```suggestion
// ValidateCFServiceDomainAliases returns error if an environment using cdn is deployed without specifying http.alias for all load-balanced web services
```

internal/pkg/describe/env.go
```diff
@@ -317,6 +317,46 @@ func (d *EnvDescriber) filterDeployedJobs() ([]*config.Workload, error) {
...
+	if !ok {
+		return fmt.Errorf("cannot find %s in env stack parameter set", cfnstack.EnvParamAliasesKey)
+	}
+
+	var aliases map[string][]string
+	err = json.Unmarshal([]byte(jsonOutput), &aliases)
+	if err != nil {
+		return fmt.Errorf("unmarshal %q: %w", jsonOutput, err)
+	}
+	lbSvcsWithoutAlias := make([]string, 0)
```
dannyrandall:
```suggestion
	var lbSvcsWithoutAlias []string
```

internal/pkg/cli/env_deploy.go
CaptainCarpensir: @iamhopaul123 can tell you more about it, I don't know what this means to do, and won't be doing myself

dannyrandall (approved):
looks great!

internal/pkg/cli/env_deploy.go
efekarakus, line 139: nit: can we move this conditional to a method to make it more readable?
```go
func (o *deployEnvOpts) isManagedCDNEnabled() bool { 
    return mft.CDNConfig.CDNEnabled() && mft.HTTPConfig.Public.Certificates == nil && o.targetApp.Domain != ""
}
```

internal/pkg/cli/env_deploy.go
```diff
@@ -123,6 +136,15 @@ func (o *deployEnvOpts) Execute() error {
 	if err != nil {
 		return err
 	}
+	if mft.CDNConfig.CDNEnabled() && mft.HTTPConfig.Public.Certificates == nil && o.targetApp.Domain != "" {
+		describer, err := o.newEnvDescriber()
+		if err != nil {
+			return fmt.Errorf("describe env: %w", err)
```
efekarakus, line 142: nit: 
```go
return fmt.Errorf("new env describer: %w", err)
```

internal/pkg/cli/env_deploy.go
```diff
@@ -123,6 +136,15 @@ func (o *deployEnvOpts) Execute() error {
...
+		}
+		if err := describer.ValidateCFServiceDomainAliases(); err != nil {
```
efekarakus, line 144: What do you think of renaming this method like this?
I think it would also be super helpful to drop a comment why this is happening as it's not immediately obvious.
```go
if err := env.ValidateWebServicesHaveAliases(); err != nil {
   // With managed domain, if the customer isn't using `alias` the A-records are inserted in the service stack as each service domain is unique.
   // However, when clients enable CloudFront, they would need to update all their existing records to now point to the distribution. 
   // Hence, we force users to use `alias` and let the records be written in the environment stack instead.
   return err
}
```

internal/pkg/describe/errors.go
```diff
@@ -15,3 +20,14 @@ type ErrManifestNotFoundInTemplate struct {
...
 }
+
+type errLBWebSvcsOnCFWithoutAlias struct {
+	services   []string
+	aliasField string
+}
+
+// Error implements the error interface.
+func (err *errLBWebSvcsOnCFWithoutAlias) Error() string {
+	return fmt.Sprintf("%v %v must have %q specified when CloudFront is enabled", english.PluralWord(len(err.services), "service", "services"),
```
efekarakus, line 31: nit: using "%s" instead of "%v"
```go

	return fmt.Sprintf("%s %s must have %q specified when CloudFront is enabled", english.PluralWord(len(err.services), "service", "services"),
		english.WordSeries(template.QuoteSliceFunc(err.services), "and"), err.aliasField)
```

internal/pkg/template/templates/environment/partials/custom-resources.yml
```diff
@@ -44,5 +44,10 @@ CustomDomainAction:
...
     DomainName: !Ref AppDNSName
-    LoadBalancerDNS: !GetAtt PublicLoadBalancer.DNSName
-    LoadBalancerHostedZone: !GetAtt PublicLoadBalancer.CanonicalHostedZoneID
\ No newline at end of file
+    {{- if .CDNConfig}}
+    PublicAccessDNS: !GetAtt CloudFrontDistribution.DomainName
+    PublicAccessHostedZone: Z2FDTNDATAQYW2 # See https://go.aws/3cPhvlX
+    {{- else}}
+    PublicAccessDNS: !GetAtt PublicLoadBalancer.DNSName
+    PublicAccessHostedZone: !GetAtt PublicLoadBalancer.CanonicalHostedZoneID
```
efekarakus, line 52: This will re-trigger the CustomDomainAction for all existing customers that already have the action enabled. Are we certain that when we try to insert the records again it won't fail and the actions are idempotent?

CaptainCarpensir: As far as I know, the only changes the lambda function may have on subsequent calls with the same input have to do with whether the resource does not respond in time. Otherwise, it will write the same A-record so it shouldn't ever cause an issue for the users, nor should it cause down-time.

PR #3843, opened 2022-08-02
chore: package addons `Fn::Transform`

dannyrandall:
This adds support for packaging the `Location` parameter on a [`Fn::Transform`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-transform.html) that uses the [`AWS::Include`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/create-reusable-transform-function-snippets-and-add-to-your-template-with-aws-include-transform.html) macro.

This also fixes a small bug in `s3.ParseURL()` where we don't correctly parse bucket names with `.`'s in it.

### DIFF
--- a/internal/pkg/addon/package.go
+++ b/internal/pkg/addon/package.go
@@ -193,9 +193,14 @@ var resourcePackageConfig = map[string][]packagePropertyConfig{
 }
 
 func (t *cfnTemplate) pkg(a *Addons) error {
-	resources := mappingNode(&t.Resources)
+	var err error
+	t.Resources, err = a.packageTransforms(t.Resources)
+	if err != nil {
+		return fmt.Errorf("package global artifacts: %w", err)
+	}
 
-	for name, node := range resources {
+	// package resources
+	for name, node := range mappingNode(&t.Resources) {
 		resType := yamlMapGet(node, "Type").Value
 		confs, ok := resourcePackageConfig[resType]
 		if !ok {
@@ -211,6 +216,51 @@ func (t *cfnTemplate) pkg(a *Addons) error {
 	}
 
 	return nil
+}
+
+func (a *Addons) packageTransforms(node yaml.Node) (yaml.Node, error) {
+	if node.Kind != yaml.MappingNode {
+		return node, nil
+	}
+
+	m := mappingNode(&node)
+	for key, val := range m {
+		switch {
+		case key == "Fn::Transform":
+			name := yamlMapGet(val, "Name")
+			if name.Value != "AWS::Include" {
+				continue
+			}
+
+			loc := yamlMapGet(yamlMapGet(val, "Parameters"), "Location")
+			if !isFilePath(loc.Value) {
+				continue
+			}
+
+			obj, err := a.uploadAddonAsset(loc.Value, false)
+			if err != nil {
+				return node, fmt.Errorf("upload asset: %w", err)
+			}
+
+			loc.Value = s3.Location(obj.Bucket, obj.Key)
+		case val.Kind == yaml.MappingNode:
+			vv, err := a.packageTransforms(*val)
+			if err != nil {
+				return node, err
+			}
+			m[key] = &vv
+		case val.Kind == yaml.SequenceNode:
+			for i := range val.Content {
+				vv, err := a.packageTransforms(*val.Content[i])
+				if err != nil {
+					return node, err
+				}
+				val.Content[i] = &vv
+			}
+		}
+	}
+
+	return node, nil
 }
 
 // yamlMapGet parses node as a yaml map and searches key. If found,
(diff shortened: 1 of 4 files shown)


internal/pkg/addon/package.go
```diff
@@ -193,9 +191,13 @@ var resourcePackageConfig = map[string][]packagePropertyConfig{
 }
 
 func (t *cfnTemplate) pkg(a *Addons) error {
-	resources := mappingNode(&t.Resources)
+	err := a.packageTransforms(t.Metadata, t.Parameters, t.Mappings, t.Conditions, t.Transform, t.Resources, t.Outputs)
```
efekarakus: > We can use the AWS::Include in any section of the template, except Parameters. 
https://medium.com/swlh/using-the-cloudformation-aws-include-macro-9e3056cf75b0

This is such a funky transformation, I wonder how many people use it

internal/pkg/addon/package.go
```diff
@@ -213,6 +215,60 @@ func (t *cfnTemplate) pkg(a *Addons) error {
 	return nil
 }
 
+// packageTransforms searches each node in nodes for the CFN
+// intrinsic function "Fn::Transform" with the "AWS::Include" macro. If it
+// dectects one, and the "Location" parameter is set to a local path, it'll
+// upload those files to S3. If node is a yaml map or sequence, it will
+// recursivly traverse those nodes.
+func (a *Addons) packageTransforms(nodes ...yaml.Node) error {
```
efekarakus: nit: What do you think of renaming it to?
```suggestion
func (a *Addons) packageIncludeTransforms(roots ...yaml.Node) error {
```

internal/pkg/addon/package.go
```diff
@@ -213,6 +215,60 @@ func (t *cfnTemplate) pkg(a *Addons) error {
 	return nil
 }
 
+// packageTransforms searches each node in nodes for the CFN
+// intrinsic function "Fn::Transform" with the "AWS::Include" macro. If it
+// dectects one, and the "Location" parameter is set to a local path, it'll
```
efekarakus:
```suggestion
// detects one, and the "Location" parameter is set to a local path, it'll
```

internal/pkg/addon/package.go
```diff
@@ -213,6 +215,60 @@ func (t *cfnTemplate) pkg(a *Addons) error {
...
+// upload those files to S3. If node is a yaml map or sequence, it will
+// recursivly traverse those nodes.
```
efekarakus:
```suggestion
// recursively traverse those nodes.
```
doing @kohidave proud with the typos

internal/pkg/addon/package.go
```diff
@@ -213,6 +215,60 @@ func (t *cfnTemplate) pkg(a *Addons) error {
...
+func (a *Addons) packageTransforms(nodes ...yaml.Node) error {
+	forNode := func(node yaml.Node) error {
```
efekarakus: nit: Should we take as input a pointer to make it clear that we're mutating the internals
```suggestion
	pkg := func(node *yaml.Node) error {
```

internal/pkg/aws/s3/s3.go
```diff
@@ -153,8 +153,20 @@ func ParseURL(url string) (bucket string, key string, err error) {
...
+	// <bucket>.s3-<region>.amazonaws.com and <bucket>.s3.<region>.amazonaws.com
+	split := strings.Split(parsedURL[0], ".")
+	bucketEndIdx := len(split) - 1
+	for ; bucketEndIdx >= 0; bucketEndIdx-- {
+		if strings.HasPrefix(split[bucketEndIdx], "s3") {
+			break
+		}
+	}
+
+	return strings.Join(split[:bucketEndIdx], "."), parsedURL[1], nil
```
efekarakus, line 169: this definitely threw me off 😂 but great test case to make it clear what's happening

efekarakus (approved):
Looks great, just small things

internal/pkg/addon/package.go
dannyrandall: 😇 i'm just impressed that you catch them

internal/pkg/addon/package.go
dannyrandall: what's extra weird to me is how the `aws cloudformation package` doesn't package an `AWS::Include` if it's the top level transform 🤔 
like:
```
Transform:
  Name: AWS::Include
  Parameters:
    Location: ./tags
Resources:
# blah
Outputs:
# bleh
```
just leaves `Transform.Parameters.Location` as is :joy:

internal/pkg/aws/s3/s3.go
paragbhingre: I like this super amazing logic ❤️
But how about this? 
```suggestion
	bucket, key = strings.Split(parsedURL[0], ".s3")[0], parsedURL[1]
```

It's basically the same logic as before but just adding `.s3` instead of `.` gives us the result that we want. 
I don't why I can't suggest code block for the removed code up there?

paragbhingre: nvm if customers put `.s3` in their name itself then logic fails. So what you have is good 💯

paragbhingre (approved):
Ship it 🎉

PR #3848, opened 2022-08-03
chore(cli): replace app init spinner in favor of progress tracker

efekarakus:

### DIFF
--- a/internal/pkg/deploy/cloudformation/app_test.go
+++ b/internal/pkg/deploy/cloudformation/app_test.go
@@ -34,10 +34,11 @@ func TestCloudFormation_DeployApp(t *testing.T) {
 		region       string
 		want         error
 	}{
-		"Infrastructure Roles Stack Fails": {
+		"should return an error if infrastructure roles stack fails": {
 			mockStack: func(ctrl *gomock.Controller) cfnClient {
 				m := mocks.NewMockcfnClient(ctrl)
-				m.EXPECT().CreateAndWait(gomock.Any()).Return(errors.New("error creating stack"))
+				m.EXPECT().Create(gomock.Any()).Return("", errors.New("error creating stack"))
+				m.EXPECT().ErrorEvents(gomock.Any()).Return(nil, nil) // No additional error descriptions.
 				return m
 			},
 			mockStackSet: func(t *testing.T, ctrl *gomock.Controller) stackSetClient {
@@ -45,11 +46,11 @@ func TestCloudFormation_DeployApp(t *testing.T) {
 			},
 			want: errors.New("error creating stack"),
 		},
-		"fail to get admin role arn": {
+		"should return a wrapped error if region is invalid when populating the admin role arn": {
 			region: "bad-region",
 			mockStack: func(ctrl *gomock.Controller) cfnClient {
 				m := mocks.NewMockcfnClient(ctrl)
-				m.EXPECT().CreateAndWait(gomock.Any()).Return(&cloudformation.ErrStackAlreadyExists{})
+				m.EXPECT().Create(gomock.Any()).Return("", &cloudformation.ErrStackAlreadyExists{})
 				return m
 			},
 			mockStackSet: func(t *testing.T, ctrl *gomock.Controller) stackSetClient {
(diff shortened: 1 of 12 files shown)


internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -172,14 +172,38 @@ func (cf CloudFormation) errorEvents(stackName string) ([]string, error) {
...
 }
 
-func (cf CloudFormation) newRenderWorkloadInput(w progress.FileWriter, stack *cloudformation.Stack) *renderStackChangesInput {
-	in := &renderStackChangesInput{
+func (cf CloudFormation) newCreateChangeSetInput(w progress.FileWriter, stack *cloudformation.Stack) *executeAndRenderChangeSetInput {
+	in := &executeAndRenderChangeSetInput{
+		stackName:        stack.Name,
+		stackDescription: fmt.Sprintf("Creating the infrastructure for stack %s", stack.Name),
+	}
+	in.createChangeSet = func() (changeSetID string, err error) {
```
dannyrandall: nit
```suggestion
	in.createChangeSet = func() (string, error) {
```

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -172,14 +172,38 @@ func (cf CloudFormation) errorEvents(stackName string) ([]string, error) {
...
+		var errAlreadyExists *cloudformation.ErrStackAlreadyExists
+		changeSetID, err = cf.cfnClient.Create(stack)
+		switch {
+		case err != nil && !errors.As(err, &errAlreadyExists):
+			spinner.Stop(log.Serrorf("%s\n", label))
+			return "", cf.handleStackError(stack.Name, err)
+		default:
+			spinner.Stop(log.Ssuccessf("%s\n", label))
+			return changeSetID, err
+		}
```
dannyrandall: nit (i  `switch` but there's only one condition 🙄)
```suggestion
		if err != nil && !errors.As(err, &errAlreadyExists) {
		    spinner.Stop(log.Serrorf("%s\n", label))
			return "", cf.handleStackError(stack.Name, err)
		}
		
		spinner.Stop(log.Ssuccessf("%s\n", label))
		return changeSetID, err
```

internal/pkg/template/templates/app/app.yml
```diff
@@ -153,6 +161,8 @@ Resources:
       Name: !Sub ${AppName}.${AppDomainName}
 
   AppDomainDelegationRecordSet:
+    Metadata:
+      'aws:copilot:description': 'Add NS records to delegate responsibility to the {{.Name}}.{{.Domain}} subdomain'
```
dannyrandall, line 165:
```suggestion
      'aws:copilot:description': 'Add NS records to delegate responsibility to the {{.Name}}.{{.Domain}} hosted zone'
```

internal/pkg/deploy/cloudformation/app_test.go
```diff
@@ -99,6 +100,7 @@ func TestCloudFormation_DeployApp(t *testing.T) {
 				cfnClient:   tc.mockStack(ctrl),
 				appStackSet: tc.mockStackSet(t, ctrl),
 				region:      tc.region,
+				console:     new(discardFile),
```
dannyrandall, line 103: 😔

dannyrandall (approved):
looks great!😍 just a couple of nits you can ignore if you want😊

internal/pkg/deploy/cloudformation/cloudformation.go
efekarakus: derp, thanks! fixed

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -219,11 +241,14 @@ func (cf CloudFormation) newRenderWorkloadInput(w progress.FileWriter, stack *cl
 	return in
 }
 
-func (cf CloudFormation) renderStackChanges(in *renderStackChangesInput) error {
+func (cf CloudFormation) executeAndRenderChangeSet(in *executeAndRenderChangeSetInput) error {
```
paragbhingre, line 244:
```suggestion
func (cf CloudFormation) createAndRenderChangeSet(in *executeAndRenderChangeSetInput) error {
```

Is it okay if we keep the name of this method as `createAndRenderChangeSet` as `execute` sounds more of an execution of something and not creation of the stack. `createAndRenderChangeSet` name is also consistent with `createAndRenderEnvironment`. Not a major one so feel free to ignore if doesn't feel relevant.

paragbhingre (approved):
Yayyy progress tracker for app init 🎉

internal/pkg/deploy/cloudformation/cloudformation.go
efekarakus: >Is it okay if we keep the name of this method as createAndRenderChangeSet as execute sounds more of an execution of something and not creation of the stack.

That's exactly right! we're executing the change set in this method that's why I renamed it. The `createChangeSet` function is a bit of misnomer I'll update in the follow-up PR, it actually either:
- creates and executes a 'Create' changeset
- or creates and executes an 'Update' changeset

PR #3849, opened 2022-08-03
feat: Add ELB access log support for Load Balanced Web Service Type

### LINKED ISSUE #1530: Feature Request: Add ELB logging configuration when Load Balanced Web Service Type
Hi, I  assessed the copilot to use copilot for my future project.
In production use,  log of the Elastic Load Balancing is crucial for trobleshooting.
Could you tell me whether if you plan to add ELB logging configuration features? If not I would like to request this feature.

paragbhingre:
Resolves #1530

With this feature we will allow customers to enable access logs for their ELBs. 
Customers will have multiple options how they can enable the access logs:
1. Customer can specify `access_logs: true` as mentioned below and copilot will create S3 bucket and enable access logs on behalf of customers.  
```
name: qa
type: Environment
http:
  public:
    access_logs: true 
```
 2. Customer can specify advanced access logs config with the bucket name and bucket prefix where they want LB to write the access logs. 
 ```
name: qa
type: Environment
http:
  public:
    access_logs:
      bucket_name: my-bucket
      bucket_prefix: access-logs
```

3. Customers can also defined partial access logs config depends on their requirement for example if customers want to just provide bucket name or bucket prefix then they can and copilot will take care of the other resources.
 ```
name: qa
type: Environment
http:
  public:
    access_logs:
      bucket_prefix: access-logs
```

In the above example copilot will create the bucket on behalf of customer and enable access logs feature to let LB write logs inside prefix mentioned by the customer.  


internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -356,11 +361,16 @@ func (e *EnvStackConfig) cdnConfig() *template.CDNConfig {
 	return nil // no-op - return &template.CDNConfig{} when feature is ready
 }
 
-func (e *EnvStackConfig) publicHTTPConfig() template.HTTPConfig {
+func (e *EnvStackConfig) publicHTTPConfig() (template.HTTPConfig, error) {
+	ELBAccessLogsConfig, err := convertELBAccessLogsConfig(e.in.Mft)
```
efekarakus: Typically exported variable names start with a capital letter, so in this scenario it would be like this :)
```suggestion
	elbAccessLogsConfig, err := convertELBAccessLogsConfig(e.in.Mft)
```

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
```diff
@@ -122,7 +170,53 @@ network:
...
+					CustomResourcesURLs: map[string]string{
+						"CertificateValidationFunction": "https://mockbucket.s3-us-west-2.amazonaws.com/dns-cert-validator",
+						"DNSDelegationFunction":         "https://mockbucket.s3-us-west-2.amazonaws.com/dns-delegation",
+						"CustomDomainFunction":          "https://mockbucket.s3-us-west-2.amazonaws.com/custom-domain",
+					},
+					Mft:    &mft,
+					RawMft: []byte(rawMft),
+				}
+			}(),
+			wantedFileName: "template-with-ELB-access-logs-with-only-bucket-prefix-field.yml",
```
efekarakus: I really wish there was a way of just comparing the resources we care about and not the entire template ya know. It'd make our tests a lot less brittle

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
```diff
@@ -33,6 +33,7 @@ type: Environment
 cdn: true
 http:
   public:
+    access_logs: true
```
efekarakus: can we remove this from the container insights test case? ideally each test case would enable a single feature

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -44,6 +44,11 @@ const (
 	capacityProviderFargate     = "FARGATE"
 )
 
+// Default values for ELB access logs config.
+const (
+	defaultBucketPrefix = "ELBAccessLogs"
+)
```
efekarakus: What if we delete this and don't have a prefix so that the logs are automatically just placed under `bucket/AWSLogs/`?

Is there a default the AWS console sets?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -354,6 +359,33 @@ type networkLoadBalancerConfig struct {
...
+	}
+	if elbAccessLogsArgs.BucketName != nil {
+		bucketName = aws.StringValue(elbAccessLogsArgs.BucketName)
+	}
+	if elbAccessLogsArgs.BucketPrefix != nil {
+		bucketPrefix = aws.StringValue(elbAccessLogsArgs.BucketPrefix)
+	}
+
+	return &template.ELBAccessLogs{
+		BucketName:   bucketName,
```
efekarakus: nit: can be simplified to
```suggestion
	if elbAccessLogsArgs.BucketPrefix != nil {
		bucketPrefix = aws.StringValue(elbAccessLogsArgs.BucketPrefix)
	}
	return &template.ELBAccessLogs{
		BucketName:   aws.StringValue(elbAccessLogsArgs.BucketName),
```

internal/pkg/manifest/env.go
```diff
@@ -415,6 +416,70 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
...
+// If there is no access logs settings, then returns nil and false.
+func (cfg *EnvironmentConfig) ELBAccessLogs() (*ELBAccessLogsArgs, bool) {
+	if isEmpty := cfg.HTTPConfig.Public.ELBAccessLogs.isEmpty(); !isEmpty {
+		if cfg.HTTPConfig.Public.ELBAccessLogs.EnableAccessLogs != nil {
+			return nil, true
+		} else {
+			return &cfg.HTTPConfig.Public.ELBAccessLogs.ELBAccessLogsArgs, true
+		}
+	}
+	return nil, false
```
efekarakus: nit: we can flip the if-conditions to reduce nesting:
```go
accessLogs := cfg.HTTPConfig.Public.ELBAccessLogs
if accessLogs.isEmpty() {
    return nil, false
}
if accessLogs.Enabled {
    return nil, true
}
return accessLogs.Advanced, true 
```

internal/pkg/template/templates/environment/partials/s3-elb-access-logs.yml
```diff
@@ -0,0 +1,39 @@
+ELBAccessLogsBucketPolicy:
```
efekarakus: nit: Would you mind renaming the file to just `elb-access-logs.yml` since S3 is the only possible output location

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -56,8 +56,93 @@ Conditions:
   ManagedAliases: !And
     - !Condition DelegateDNS
     - !Not [!Equals [ !Ref Aliases, "" ]]
+{{- if and .PublicHTTPConfig.ELBAccessLogs (not .PublicHTTPConfig.ELBAccessLogs.BucketName) }} #.PublicHTTPConfig.ELBAccessLogs.CreateBucket
```
efekarakus: What do you think of adding a method instead:
```suggestion
{{- if and .PublicHTTPConfig.ELBAccessLogs.ShouldCreateBucket }} 
```

and in `template/env.go`
```go
func (elb *ELBAccessLogs) ShouldCreateBucket() bool {
   if elb == nil { return false }
   if len(elb.BucketName) > 0 { return false }
   return true
}
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -56,8 +56,93 @@ Conditions:
...
+Mappings:
+  RegionalConfigs:
+    us-east-1:
+      ELBAccountId: '127311923021'
+      ArnPrefix: 'arn:aws:'
```
efekarakus: Ugh this massive mappings makes me so sad as it will be difficult to maintain, what do you think of moving the `RegionalConfigs` section into a partial called `mappings-regional-configs.yml` and include it here?

Dropping a comment to the table would also be helpful: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#access-logging-bucket-permissions

efekarakus: Also it looks like we don't need the `ArnPrefix` key

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -56,8 +56,93 @@ Conditions:
...
+    us-east-2:
+      ELBAccountId: '033677994240'
```
efekarakus: nit: in CFN they use a different casing
```suggestion
      ElbAccountId: '033677994240'
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -278,6 +363,15 @@ Resources:
     Condition: CreateALB
     Type: AWS::ElasticLoadBalancingV2::LoadBalancer
     Properties:
+      {{- if .PublicHTTPConfig.ELBAccessLogs }}
+      LoadBalancerAttributes:
+        - Key: 'access_logs.s3.enabled'
+          Value: true
+        - Key: 'access_logs.s3.prefix'
+          Value: {{ .PublicHTTPConfig.ELBAccessLogs.BucketPrefix }}
```
efekarakus, line 293: Should this be:
```suggestion
        {{- if  .PublicHTTPConfig.ELBAccessLogs.BucketPrefix }}
        - Key: 'access_logs.s3.prefix'
          Value: {{ .PublicHTTPConfig.ELBAccessLogs.BucketPrefix }}
        {{- end}}
```
So that a prefix is there only if it's not empty 💭

internal/pkg/template/templates/environment/partials/s3-elb-access-logs.yml
```diff
@@ -0,0 +1,39 @@
...
+                - '/*'
+          Principal:
+            AWS: !Join [ "", [  "arn:aws:iam::", !FindInMap [ RegionalConfigs, !Ref 'AWS::Region', ELBAccountId ], ":root" ] ]
+ELBAccessLogsBucket:
+  Type: AWS::S3::Bucket
+  Condition: CreateALB
+  Properties:
+    {{- if .BucketName}}
+    BucketName: {{ .BucketName }}
+    {{- end }}
```
efekarakus: We can remove this as we should only create the bucket if the user doesn't use the `bucket_name` field

internal/pkg/template/templates/environment/partials/s3-elb-access-logs.yml
```diff
@@ -0,0 +1,39 @@
...
+          Resource:
+            - !Join
+              - ''
+              - - 'arn:'
+                - !Ref AWS::Partition
+                - ':s3:::'
+                - !Ref ELBAccessLogsBucket
+                - '/'
+                - {{ .BucketPrefix }}
+                - '/AWSLogs/'
```
efekarakus: If the `.BucketPrefix` is empty, I think this will result in `//AWSLogs/`  should this be something like:
```
{{- if .BucketPrefix }}
- '/'
- {{.BucketPrefix }}
{{- end}}
- '/AWSLogs/'

```

internal/pkg/template/templates/environment/partials/s3-elb-access-logs.yml
```diff
@@ -0,0 +1,39 @@
...
+                - !Ref AWS::AccountId
+                - '/*'
+          Principal:
+            AWS: !Join [ "", [  "arn:aws:iam::", !FindInMap [ RegionalConfigs, !Ref 'AWS::Region', ELBAccountId ], ":root" ] ]
```
efekarakus: Can we replace "arn:aws:iam" with `!Sub "arn:${AWS::Partition}:iam` (I don't remember the exact syntax) so that it uses the appropriate partition for the region

efekarakus (commented):
Very nice!

internal/pkg/deploy/cloudformation/stack/env.go
paragbhingre: What? 🥺 I thought I renamed it in small laters in my nits fix. I will change it now.

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
paragbhingre: Hmm, I was already creating 2 more tests with their own output templates so I thought I might just use existing one in order to reduce the code. I also saw new `cdn: true` changes got added to it. But if you suggest then I will change it to the different test case.

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: Yes you are right, logs will be placed under `bucket/AWSLogs/`
And AWS console does not set any default, if you do not provide prefix then console also puts logs into `bucket/AWSLogs/`

internal/pkg/template/templates/environment/cf.yml
paragbhingre: So prefix will be there in either case as we are setting it to default even if customers doesn't provide it, similar to how if customers doesn't provide bucket name but only provides prefix then we create the bucket and attach their prefix to it.

internal/pkg/template/templates/environment/partials/s3-elb-access-logs.yml
paragbhingre: We can do as you mentioned but in our implementation there won't be a case where bucket prefix is not defined so i kept it this way but happy to change it.

internal/pkg/template/templates/environment/partials/s3-elb-access-logs.yml
paragbhingre: Yeah right make sense, I had this block before, when we were still discussing if we want to provide an option to customers to create a bucket or not?

internal/pkg/template/templates/environment/cf.yml
paragbhingre: Sure, I had these 2 actions in mind but somehow it skipped through my mind. I will change it.

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
```diff
@@ -70,7 +70,84 @@ observability:
 			}(),
 			wantedFileName: "template-with-imported-certs-observability.yml",
 		},
-
+		"generate template with default access logs": {
```
Lou1415926, line 77: What do you think of just creating a test case named "generate template with fully configured environment manifest", and dump all the configurations into that file?

Right now we are comparing the whole template with each possible value of every manifest field. Our test files is blooming up rapidly 😂  I think we should have confidence in our unit test - instead of testing every thing in this integration test, we could just test some use cases.

internal/pkg/manifest/env.go
```diff
@@ -415,6 +416,70 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
 type PublicHTTPConfig struct {
 	SecurityGroupConfig ALBSecurityGroupsConfig `yaml:"security_groups,omitempty"`
 	Certificates        []string                `yaml:"certificates,omitempty"`
+	ELBAccessLogs       ELBAccessLogsArgsOrBool `yaml:"access_logs,omitempty"`
```
Lou1415926, line 421: Would we want to add this to `http.private` as well? Or there is no need for now.

internal/pkg/manifest/env.go
```diff
@@ -415,6 +416,70 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
...
+}
+
+// ELBAccessLogsArgsOrBool is a custom type which supports unmarshaling yaml which
+// can either be of type bool or type ELBAccessLogsArgs.
+type ELBAccessLogsArgsOrBool struct {
+	EnableAccessLogs  *bool
+	ELBAccessLogsArgs ELBAccessLogsArgs
```
Lou1415926: nit:
```suggestion
	Enabled  *bool
	AdvancedConfig ELBAccessLogsArgs
```
Hopefully it reads better this way! I think the `"AccessLogs"` in `EnableAccessLogs` and `ELBAccessLogsArgs` are sort of repetitive information because those keywords are already in the struct name.

internal/pkg/manifest/env.go
```diff
@@ -415,6 +416,70 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
...
+	}
+
+	if err := value.Decode(&al.EnableAccessLogs); err != nil {
+		return errUnmarshalELBAccessLogs
+	}
+	return nil
+}
+
+// AccessLogsConfig holds the access logs configuration.
+// These options are specifiable under the "access_logs" field.
```
Lou1415926: nit:
```suggestion
```

I feel like these struct comments are to describe what the struct does and usually a struct should be agnostic to its parents. Therefore, usually a struct comment should be agnostic to the parents as well 💭

internal/pkg/manifest/env.go
```diff
@@ -415,6 +416,70 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
...
+
+func (al *ELBAccessLogsArgs) isEmpty() bool {
+	return al.BucketName == nil && al.BucketPrefix == nil
+}
+
+func (al *ELBAccessLogsArgsOrBool) isEmpty() bool {
+	if al.EnableAccessLogs != nil {
+		return false
+	}
+	return al.ELBAccessLogsArgs.isEmpty()
```
Lou1415926: nit:
```suggestion
	return al.EnabledAccessLogs == nil && al.ELBAccessLogsArgs.isEmpty()
```

internal/pkg/manifest/env.go
```diff
@@ -450,7 +515,7 @@ func (i Ingress) IsEmpty() bool {
 
 // IsEmpty returns true if there is no customization to the public ALB.
 func (cfg PublicHTTPConfig) IsEmpty() bool {
-	return len(cfg.Certificates) == 0 && cfg.SecurityGroupConfig.IsEmpty()
+	return len(cfg.Certificates) == 0 && cfg.SecurityGroupConfig.IsEmpty() && cfg.ELBAccessLogs.isEmpty()
```
Lou1415926, line 516: nice! thanks for updating `PublicHTTPConfig`!

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -278,6 +284,15 @@ Resources:
...
+        - Key: 'access_logs.s3.bucket'
+          Value: {{- if .PublicHTTPConfig.ELBAccessLogs.BucketName }} {{ .PublicHTTPConfig.ELBAccessLogs.BucketName }}{{- else }} !Ref ELBAccessLogsBucket {{- end }}
```
Lou1415926, line 296: `ELBAccessLogsBucket` is created only when we `ShouldCreateBucket`, which returns `true` only when `ELBAccessLogs.BucketName` is not empty. My worry is that we need to go through through a relatively longer chain of logic for us to validate that it works correctly. Do you think we could simplify the logic a little?

For example, do we need `ShouldCreateBucket` - could
```
{{- if .PublicHTTPConfig.ELBAccessLogs.ShouldCreateBucket }}
```
be replaced by
```
{{- if and .PublicHTTPConfig.ELBAccessLogs .PublicHTTPConfig.ELBAccessLogs.BucketName}}
```

internal/pkg/manifest/env.go
```diff
@@ -415,6 +416,70 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
...
+func (al *ELBAccessLogsArgsOrBool) isEmpty() bool {
+	if al.EnableAccessLogs != nil {
+		return false
+	}
+	return al.ELBAccessLogsArgs.isEmpty()
+}
+
+// AccessLogs returns the access logs config if the user has set any values.
+// If there is no access logs settings, then returns nil and false.
+func (cfg *EnvironmentConfig) ELBAccessLogs() (*ELBAccessLogsArgs, bool) {
```
Lou1415926, line 472: what do you think of
```
func (cfg *EnvironmentConfig) ELBAccessLogs() (*template.ELBAccessLogs, bool) 
```

instead? I feel like this function and the one in `transformer.go` is sort of duplicated

Lou1415926 (commented):
Looks great - love the boolOrMap design!

internal/pkg/manifest/env.go
paragbhingre: I see. Noted.

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
paragbhingre: Yeah I think this point is valid. What I will do is instead of putting everything in one test case I will just create one test case separately for access logs and put all the possible configs related to the access logs in it instead of creating separate test for each use case. And I will make sure I will take care of the other edge cases in my unit test locally and through `validate_env.test`

internal/pkg/template/templates/environment/cf.yml
paragbhingre: As you mentioned I had the exact same logic [here](https://github.com/aws/copilot-cli/pull/3849/commits/98f346b8e7496c77e251d6e30b417448e81b210b#diff-e03bdf3cbbc34b11b5dca5b60e3916cbb5fb2b13cb17633661db00ce6c0f2561R59). But Efe was mentioning that we can have [separate method](https://github.com/aws/copilot-cli/pull/3849#discussion_r938325056) as we are checking this same condition twice while creating the bucket and also while importing the `regionalmaps` in the cf.yml file. Let me know what do you think about it.

internal/pkg/manifest/env.go
paragbhingre: Yes we are only planning on releasing this for public load balancers at the moment.

internal/pkg/deploy/cloudformation/stack/transformers.go
efekarakus: Should we remove it then? 😝 so that we don't set a bucket prefix either by default

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
efekarakus: Was this addressed? It looks like we still have 3 separate testdata files

internal/pkg/manifest/env.go
```diff
@@ -415,6 +416,66 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
...
+type ELBAccessLogsArgs struct {
+	BucketName   *string `yaml:"bucket_name,omitempty"`
+	BucketPrefix *string `yaml:"bucket_prefix,omitempty"`
+}
+
+func (al *ELBAccessLogsArgs) isEmpty() bool {
+	return al.BucketName == nil && al.BucketPrefix == nil
+}
+
+func (al *ELBAccessLogsArgsOrBool) isEmpty() bool {
```
efekarakus: nit: can we move this to l.428 under the `ELBAccessLogsArgsOrBool` type declaration so that the methods are grouped together?

internal/pkg/manifest/env.go
efekarakus: I'm a fan of keeping it as is and not return `template` types from the manifest pkg.
Ideally the `manifest` configuration can be transformed into multiple output types: CFN go templates, CDK code, Terraform (whatever we want) so it should return its own types.

All these convert functions would move to the `cfntemplate` pkg, and in unit tests we would ensure that `template.Parse` is called with the right arguments:

template-dependencies

internal/pkg/template/env.go
```diff
@@ -121,6 +123,24 @@ type HTTPConfig struct {
...
+
+// ShouldCreateBucket returns true if copilot should create bucket on behalf of customer.
+func (elb *ELBAccessLogs) ShouldCreateBucket() bool {
+	if elb == nil {
+		return false
+	}
+	if len(elb.BucketName) > 0 {
+		return false
+	}
+	return true
```
```suggestion
	return len(elb.BucketName) == 0
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -56,8 +56,14 @@ Conditions:
   ManagedAliases: !And
     - !Condition DelegateDNS
     - !Not [!Equals [ !Ref Aliases, "" ]]
+{{- if .PublicHTTPConfig.ELBAccessLogs.ShouldCreateBucket }}
+{{ include "mappings-regional-configs" . }}
+{{- end }}
```
efekarakus, line 61: aww yea, v nice! 👏

internal/pkg/template/templates/environment/cf.yml
efekarakus: Can we remove our default prefix plz 🥺 I feel like the console experience makes sense because they place logs already under `/AWSLogs/aws-account-id/elasticloadbalancing/` which is pretty clear enough

efekarakus (commented):
Looks great! Just some feedback around removing the default prefix, once that's addressed :shipit:

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
paragbhingre: I was going to remove them at the end because I was using them for the testing purpose until I address all other feedback comments. 
For now I have removed 2 tests related to access logs and kept only one which tests `advancedAccessLogs` and added default access logs as a part of one of the existing tests.

For now I will keep it like this and later we can create only one test including all the fields of the env manifest as Wanxian mentioned. But right now doing so is a bit heavy work for me as a part of this feature.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -354,6 +354,22 @@ type networkLoadBalancerConfig struct {
 	appDNSName           *string
 }
 
+func convertELBAccessLogsConfig(mft *manifest.Environment) (*template.ELBAccessLogs, error) {
+	elbAccessLogsArgs, isELBAccessLogsSet := mft.ELBAccessLogs()
+	if !isELBAccessLogsSet { //if ELB access logs configs are not defined.
```
efekarakus: nit: the code is pretty clear ✨ , I think we can remove these implementation comments

efekarakus (approved):
😍

internal/pkg/template/templates/environment/cf.yml
iamhopaul123: Ditto. I feel `{{- if and .PublicHTTPConfig.ELBAccessLogs .PublicHTTPConfig.ELBAccessLogs.BucketName}}` is probably ok to repeat., since it is quite straight forward.

internal/pkg/template/templates/environment/partials/elb-access-logs.yml
```diff
@@ -0,0 +1,38 @@
+ELBAccessLogsBucketPolicy:
```
iamhopaul123, line 1: How does this work? It seems like we don't refer to `ELBAccessLogsBucketPolicy` anywhere. We just need this policy defined here and then ELB will be able to write to the bucket?

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-imported-certs-observability.yml
```diff
@@ -418,6 +421,13 @@ Resources:
     Condition: CreateALB
     Type: AWS::ElasticLoadBalancingV2::LoadBalancer
     Properties:
+      LoadBalancerAttributes:
+        - Key: 'access_logs.s3.enabled'
+          Value: true
+        - Key: 'access_logs.s3.prefix'
+          Value: "accesslogsbucketprefix"
+        - Key: 'access_logs.s3.bucket'
+          Value: "accesslogsbucket"
```
Lou1415926, line 430: I like how it's easier this way to review the changes the code has on the template!

internal/pkg/manifest/env.go
```diff
@@ -417,6 +418,66 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
...
+type ELBAccessLogsArgs struct {
+	BucketName   *string `yaml:"bucket_name,omitempty"`
+	BucketPrefix *string `yaml:"bucket_prefix,omitempty"`
+}
+
+func (al *ELBAccessLogsArgs) isEmpty() bool {
+	return al.BucketName == nil && al.BucketPrefix == nil
+}
+
+// AccessLogs returns the access logs config if the user has set any values.
```
Lou1415926:
```suggestion
// ELBAccessLogs returns the access logs config if the user has set any values.
```

(PS: If you are using GoLand, you can just go to "Inspections" and turn on the warning so that it complains when the comment doesn't start with the exported name; if you are using VSCode, you will probably need to configure the linters directly somehow)

internal/pkg/manifest/env.go
```diff
@@ -417,6 +418,66 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
...
+// If there is no access logs settings, then returns nil and false.
+func (cfg *EnvironmentConfig) ELBAccessLogs() (*ELBAccessLogsArgs, bool) {
+	accessLogs := cfg.HTTPConfig.Public.ELBAccessLogs
+	if accessLogs.isEmpty() {
+		return nil, false
+	}
+	if accessLogs.Enabled != nil {
+		return nil, true
+	}
+	return &accessLogs.AdvancedConfig, true
```
Lou1415926: What about
```yaml
http:
  public:
    access_logs: false
```

internal/pkg/manifest/env.go
```diff
@@ -417,6 +418,66 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
...
+}
+
+func (al *ELBAccessLogsArgsOrBool) isEmpty() bool {
+	return al.Enabled == nil && al.AdvancedConfig.isEmpty()
+}
+
+// UnmarshalYAML overrides the default YAML unmarshaling logic for the ELBAccessLogsArgsOrBool
+// struct, allowing it to perform more complex unmarshaling behavior.
+// This method implements the yaml.Unmarshaler (v3) interface.
+func (al *ELBAccessLogsArgsOrBool) UnmarshalYAML(value *yaml.Node) error {
```
Lou1415926, line 438: We can add  / modify a unit test [here](https://github.com/aws/copilot-cli/blob/365f751c7100251a87643766cb84e91df98b4ddd/internal/pkg/manifest/env_test.go#L321) to test this unmarshal logic!

internal/pkg/manifest/env.go
```diff
@@ -417,6 +418,66 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
...
+	BucketPrefix *string `yaml:"bucket_prefix,omitempty"`
+}
+
+func (al *ELBAccessLogsArgs) isEmpty() bool {
+	return al.BucketName == nil && al.BucketPrefix == nil
+}
+
+// AccessLogs returns the access logs config if the user has set any values.
+// If there is no access logs settings, then returns nil and false.
+func (cfg *EnvironmentConfig) ELBAccessLogs() (*ELBAccessLogsArgs, bool) {
```
Lou1415926, line 472: Since this method is exported, I think it warrants a unit test!

internal/pkg/template/env.go
```diff
@@ -121,6 +123,21 @@ type HTTPConfig struct {
...
+	BucketName   string
+	BucketPrefix string
+}
+
+// ShouldCreateBucket returns true if copilot should create bucket on behalf of customer.
+func (elb *ELBAccessLogs) ShouldCreateBucket() bool {
+	if elb == nil {
+		return false
+	}
+	return len(elb.BucketName) == 0
```
Lou1415926: nit: 
```suggestion
	return elb.BucketName == ""
```

internal/pkg/manifest/env.go
paragbhingre: Added unit tests, thanks for pointing me to the tests :)

internal/pkg/template/templates/environment/partials/elb-access-logs.yml
paragbhingre: No no, if you see [here](https://github.com/aws/copilot-cli/pull/3849/files/8b5ae7abb677ddcaf6c8067697ae395c9b9ebf2b#diff-3622335dad872c1b6514d7b8e98203d3c30f9ab68350d949b14ac3b9a6774dfeR6) we are setting `Bucket` field on the policy which tells the policy that it needs to be applied on that particular bucket. Does that make sense?

internal/pkg/manifest/env.go
paragbhingre: That's a great catch. I had thought that if they want to remove the access logs then they will just remove the configuration instead of setting this flag as false. But I have changed the logic to catch false and remove the access logs feature.

internal/pkg/manifest/env.go
paragbhingre: Got it.

internal/pkg/template/templates/environment/partials/elb-access-logs.yml
iamhopaul123: I see `AWS::S3::BucketPolicy` is a special one. Thanks!

internal/pkg/manifest/env_test.go
```diff
@@ -873,3 +924,63 @@ func TestEnvironmentCDNConfig_CDNEnabled(t *testing.T) {
 		})
 	}
 }
+
+func TestEnvironmentAccessLogs_ELBAccessLogs(t *testing.T) {
```
Lou1415926: nit: so that it starts with the receiver type's name, and then the method name
```suggestion
func TestEnvironmentConfig_ELBAccessLogs(t *testing.T) {
```

internal/pkg/manifest/env_test.go
```diff
@@ -873,3 +924,63 @@ func TestEnvironmentCDNConfig_CDNEnabled(t *testing.T) {
...
+				},
+			},
+			wantedFlag:    true,
+			wantedConfigs: nil,
+		},
+		"disabled via bool": {
+			in: EnvironmentConfig{
+				HTTPConfig: EnvironmentHTTPConfig{
+					Public: PublicHTTPConfig{
+						ELBAccessLogs: ELBAccessLogsArgsOrBool{},
```
Lou1415926: nit: is this test intended to be the following?
```suggestion
						ELBAccessLogs: ELBAccessLogsArgsOrBool{
						     Enabled: aws.Bool(false)
						},
```

Lou1415926 (approved):
LGTM!! Thanks for adding the tests! Feel free to remove the label when you're ready!

internal/pkg/manifest/env_test.go
paragbhingre: Actually it is for both if they give empty config or sets it to false. Both the configs will disable access logs.  But I will set to false for better understanding.

internal/pkg/manifest/env_test.go
paragbhingre: Ohh got it.

PR #3851, opened 2022-08-04
chore: add cdn certificate to cloudfront and enable feature

CaptainCarpensir:
This adds `cdn.certificate` which allows the user to use certificates imported in `us-east-1` with CloudFront.
There's a few more edge cases needed for the feature to be release ready namely to work with or invalidate `hosted_zone` field in svc manifest with CloudFront.

@iamhopaul123 will be adding the capability for CloudFront to read aliases from the stack parameters to allow a CloudFront distribution creation with imported certs or `--domain` to be able to occur entirely within Copilot

Nearly completes milestone 1 of #3701 and requires #3842 to be merged before merging this


internal/pkg/cli/deploy/svc.go
```diff
@@ -1390,12 +1396,20 @@ func (d *lbWebSvcDeployer) validateALBRuntime() error {
 	}
 	if hasImportedCerts {
 		aliases, err := d.lbMft.RoutingRule.Alias.ToStringSlice()
+		cdnCert := d.environmentConfig.CDNConfig.CDNConfig.Certificate
+		defaultValidator := d.newAliasCertValidator(nil)
+		cfValidator := d.newAliasCertValidator(aws.String("us-east-1")) // make us-east-1 a constant somewhere
```
dannyrandall: let's make it a constant somewhere :)

internal/pkg/cli/deploy/svc.go
```diff
@@ -1390,12 +1396,20 @@ func (d *lbWebSvcDeployer) validateALBRuntime() error {
...
 		if err != nil {
 			return fmt.Errorf("convert aliases to string slice: %w", err)
 		}
```
dannyrandall:
```suggestion
		aliases, err := d.lbMft.RoutingRule.Alias.ToStringSlice()
		if err != nil {
			return fmt.Errorf("convert aliases to string slice: %w", err)
		}
		
		cdnCert := d.environmentConfig.CDNConfig.CDNConfig.Certificate
		defaultValidator := d.newAliasCertValidator(nil)
		cfValidator := d.newAliasCertValidator(aws.String("us-east-1")) // make us-east-1 a constant somewhere
```

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -127,7 +127,7 @@ Resources:
 {{- end}}
       Tags:
         - Key: Name
-          Value: !Sub 'copilot-${AppName}-${EnvironmentName}-lb'
+          Value: !Sub 'copilot-${AppName}-${EnvironmentName}-lb-http'
```
dannyrandall, line 130: ❤️

internal/pkg/template/templates/environment/partials/environment-manager-role.yml
```diff
@@ -32,6 +32,11 @@ EnvironmentManagerRole:
 {{- range $arn := .PrivateHTTPConfig.ImportedCertARNs}}
           - "{{$arn}}"
 {{- end}}
+{{- if .CDNConfig}}
+{{- if .CDNConfig.ViewerCertificate}}
```
paragbhingre:
```suggestion
{{- if and .CDNConfig .CDNConfig.ViewerCertificate}}
```

You can do this instead of two if statements

internal/pkg/cli/deploy/svc.go
iamhopaul123: we can make it a constant here for now.

internal/pkg/cli/deploy/svc.go
```diff
@@ -327,10 +327,10 @@ type customResourcesFunc func(fs template.Reader) ([]*customresource.CustomResou
 type lbWebSvcDeployer struct {
 	*svcDeployer
 	appVersionGetter       versionGetter
-	aliasCertValidator     aliasCertValidator
 	publicCIDRBlocksGetter publicCIDRBlocksGetter
 	lbMft                  *manifest.LoadBalancedWebService
 	customResources        customResourcesFunc
+	newAliasCertValidator  func(region *string) aliasCertValidator
```
iamhopaul123:
```suggestion
	newAliasCertValidator  func(optionalRegion *string) aliasCertValidator
```

internal/pkg/cli/deploy/svc.go
```diff
@@ -1390,12 +1396,20 @@ func (d *lbWebSvcDeployer) validateALBRuntime() error {
 	}
 	if hasImportedCerts {
 		aliases, err := d.lbMft.RoutingRule.Alias.ToStringSlice()
+		cdnCert := d.environmentConfig.CDNConfig.CDNConfig.Certificate
+		defaultValidator := d.newAliasCertValidator(nil)
```
iamhopaul123: How about `albCertValidator` and `cfCertValidator`

internal/pkg/cli/deploy/svc.go
```diff
@@ -1390,12 +1396,20 @@ func (d *lbWebSvcDeployer) validateALBRuntime() error {
...
+		cfValidator := d.newAliasCertValidator(aws.String("us-east-1")) // make us-east-1 a constant somewhere
 		if err != nil {
 			return fmt.Errorf("convert aliases to string slice: %w", err)
 		}
-		if err := d.aliasCertValidator.ValidateCertAliases(aliases, d.environmentConfig.HTTPConfig.Public.Certificates); err != nil {
+		if err := defaultValidator.ValidateCertAliases(aliases, d.environmentConfig.HTTPConfig.Public.Certificates); err != nil {
 			return fmt.Errorf("validate aliases against the imported certificate for env %s: %w", d.env.Name, err)
```
iamhopaul123:
```suggestion
			return fmt.Errorf("validate aliases against the imported public ALB certificate for env %s: %w", d.env.Name, err)
```

internal/pkg/cli/deploy/svc.go
```diff
@@ -1390,12 +1396,20 @@ func (d *lbWebSvcDeployer) validateALBRuntime() error {
...
 		}
+		if cdnCert != nil {
+			if err := cfValidator.ValidateCertAliases(aliases, []string{*cdnCert}); err != nil {
+				return fmt.Errorf("validate aliases against the cdn imported certificate for env %s: %w", d.env.Name, err)
```
iamhopaul123:
```suggestion
				return fmt.Errorf("validate aliases against the imported CDN certificate for env %s: %w", d.env.Name, err)
```

internal/pkg/manifest/validate_env.go
```diff
@@ -326,6 +338,11 @@ func (i RestrictiveIngress) validate() error {
 
 // validate is a no-op for AdvancedCDNConfig.
```
paragbhingre: You might want to change this description of the validate method.

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -353,7 +354,15 @@ func (e *BootstrapEnvStackConfig) ToEnv(stack *cloudformation.Stack) (*config.En
...
 
 func (e *EnvStackConfig) cdnConfig() *template.CDNConfig {
-	return nil // no-op - return &template.CDNConfig{} when feature is ready
+	if e.in.Mft == nil {
+		return nil
+	}
+	if !e.in.Mft.CDNConfig.CDNEnabled() {
+		return nil
+	}
+	return &template.CDNConfig{
```
iamhopaul123, line 363: woohoo it is happening

internal/pkg/manifest/validate_env.go
```diff
@@ -19,7 +20,7 @@ var (
 // Validate returns nil if Environment is configured correctly.
 func (e Environment) Validate() error {
 	if err := e.EnvironmentConfig.validate(); err != nil {
-		return fmt.Errorf(`validate "network": %w`, err)
+		return fmt.Errorf(`validate environment: %w`, err)
```
iamhopaul123: We don't even need to wrap this error i think. because otherwise the error msg will be

```
validate environment manifest for "test": validate environment: some error"
```

internal/pkg/manifest/validate_env.go
```diff
@@ -41,6 +42,17 @@ func (e EnvironmentConfig) validate() error {
 	if e.IsIngressRestrictedToCDN() && !e.CDNConfig.CDNEnabled() {
 		return errors.New("CDN must be enabled to limit security group ingress to CloudFront")
 	}
+	if e.CDNConfig.CDNEnabled() {
+		if e.HTTPConfig.Public.Certificates == nil {
+			if e.CDNConfig.CDNConfig.Certificate != nil {
```
iamhopaul123: We should probably rename the second CDNConfig in `environmentCDNConfig` to just `Config`, since the struct name should have the namespace info.

internal/pkg/manifest/validate_env.go
```diff
@@ -41,6 +42,17 @@ func (e EnvironmentConfig) validate() error {
...
+				return errors.New("must import ALB certificates to set \"cdn.certificate\"")
```
iamhopaul123: We might want to provide more context to users saying they have to specify together. Can we make use of `errFieldMustBeSpecified`?

internal/pkg/manifest/validate_env.go
```diff
@@ -41,6 +42,17 @@ func (e EnvironmentConfig) validate() error {
...
+			}
+		} else {
+			if e.CDNConfig.CDNConfig.Certificate == nil {
+				return errors.New("must set \"cdn.certificate\" to import ALB certificates with CDN enabled")
```
iamhopaul123: same for this one

iamhopaul123:
Also we need to write to the alternate name for CF distribution right? Will that be in the following PR?

CaptainCarpensir:
> Also we need to write to the alternate name for CF distribution right? Will that be in the following PR?

Ah I completely forgot about this! At this point yeah it should be, I don't think it will take much time, I'm just worried that we're enabling the feature when it's still a little too incomplete if that's the case. After this there are two loose ends to wrap up:
- alternate domain name writing for cloudfront
- validation of some kind for services using the `hosted_zone` feature

iamhopaul123:
I don't mind validation to go after this PR but yeah I agree the alternate name setting we might merge that one prior to this feat enabling PR.

CaptainCarpensir:
Alright, this PR isn't very big on its own so I'll add that in here then re-enable for reviews 👍

internal/pkg/manifest/validate_env.go
```diff
@@ -41,6 +42,17 @@ func (e EnvironmentConfig) validate() error {
...
+			}
+		}
+	}
```
paragbhingre: This condition has multiple nested ifs, I was thinking what we can do with it. I tried doing few permutations with the if logic and it passes your tests in validate_env but fails few other tests for some weird reasons. 
We can at least take this `e.CDNConfig.CDNConfig.Certificate` out in a variable and make it more readable.

paragbhingre (commented):
Looks good🤩 , just few nits.

internal/pkg/manifest/validate_env.go
paragbhingre: +100

internal/pkg/manifest/errors.go
```diff
@@ -56,16 +56,21 @@ func (e *ErrUnknownProvider) Is(target error) bool {
...
 func (e *errFieldMustBeSpecified) Error() string {
 	errMsg := fmt.Sprintf(`%q must be specified`, e.missingField)
 	if len(e.conditionalFields) == 0 {
 		return errMsg
 	}
-	return fmt.Sprintf(`%s if %s %s specified`, errMsg, english.WordSeries(quoteStringSlice(e.conditionalFields), "or"),
+	conjunction := "or"
+	if e.allMustBeSpecified {
+		conjunction = "and"
+	}
```
dannyrandall, line 72: idk why but this makes me laugh

internal/pkg/manifest/validate_env_test.go
```diff
@@ -695,14 +719,27 @@ func TestCDNConfiguration_validate(t *testing.T) {
...
 			},
+			wantedErrorMsgPrefix: "parse cdn certificate:",
 		},
 	}
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
 			gotErr := tc.in.validate()
-			if tc.wantedError != nil {
+			if tc.wantedErrorMsgPrefix != "" {
+				require.Error(t, gotErr)
```
dannyrandall, line 739: this is covered by the check below
```suggestion
```

dannyrandall (approved):

internal/pkg/deploy/cloudformation/stack/env.go
iamhopaul123, line 363: We should probably revert this change.

iamhopaul123 (approved):

PR #3853, opened 2022-08-05
test(regression): add regression test for multi-svc-app

Lou1415926:
This PR sets up regression tests to validate that an application that was functional in an older Copilot version would be still functional in a later version. Right now, there is only one test suite: `"multi-svc-app"` which has a similar but simpler set up as the existing `"multi-svc-app"` test suite in e2e. We will want to add more, for example, regression tests on applications with domain, with imported cert, and with alias.

This PR only contains the test codes, and the make target to dry run the test on local machines. It does not add the test to pipeline. This will be done in a separate PR.


scripts/dryrun-creds.sh
```diff
@@ -1,19 +1,26 @@
...
 
 cmd = subprocess.run(['aws', 'configure', 'list-profiles'], stdout=subprocess.PIPE)
 profiles = cmd.stdout.decode('utf-8').split('\n')
 
-for e2eprofile in ['e2etestenv', 'e2eprodenv']:
-  if e2eprofile in profiles:
+if sys.argv[1] == "e2e":
+  requiredProfiles = ['e2etestenv', 'e2eprodenv']
+elif sys.argv[1] == "regression":
+  requiredProfiles = ['regression']
```
efekarakus, line 14: nit: should we exit with an error if an arg is not passed
```python
else:
  sys.exit('no arg passed to script, must be one of: "e2e" or "regression"')
```

Makefile
```diff
@@ -125,11 +125,24 @@ e2e-dryrun: build # Sample command "make e2e-dryrun test=multi-env-app" to run t
...
 	@echo "Setup credentials"
-	./scripts/e2e-dryrun-creds.sh
+	./scripts/dryrun-creds.sh e2e
 	@echo "Run the $(test) test"
 	cd e2e/$(test) && DRYRUN=true ginkgo -v -r
 	cd -
 
+# Examples:
+# REGRESSION_TEST_FROM_PATH=/usr/local/bin/copilot make regression-dryrun test=multi-svc-app
+# REGRESSION_TEST_FROM_PATH=/usr/local/bin/copilot-v1.18.0 REGRESSION_TEST_FROM_PATH=/usr/local/bin/copilot-v1.19.0 make regression-dryrun test=multi-svc-app
```
efekarakus: 😂 
```suggestion
# REGRESSION_TEST_FROM_PATH=/usr/local/bin/copilot-v1.18.0 REGRESSION_TEST_TO_PATH=/usr/local/bin/copilot-v1.19.0 make regression-dryrun test=multi-svc-app
```

regression/client/cli.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (cli *CLI) Run(commands ...string) (string, error) {
+	return cli.exec(exec.Command(cli.path, commands...))
+}
+
+func (cli *CLI) exec(command *exec.Cmd) (string, error) {
+	// Turn off colors
+	command.Env = append(os.Environ(), "COLOR=false", "CI=true")
+	command.Dir = cli.workingDir
```
efekarakus: nit: workingDir never seems to be set should we remove it for now?

regression/client/outputs.go
```diff
@@ -0,0 +1,92 @@
...
+	Value       string `json:"value"`
+}
+
+type svcShowResourceInfo struct {
+	Type       string `json:"type"`
+	PhysicalID string `json:"physicalID"`
+}
+
+// ToSvcShowOutput unmarshal a JSON string to a svcShowOutput struct.
+func ToSvcShowOutput(jsonInput string) (*svcShowOutput, error) {
```
efekarakus: nit: Should the `*Output` types be exported?

regression/multi-svc-app/multi_svc_app_suite_test.go
```diff
@@ -0,0 +1,56 @@
...
+var _ = BeforeSuite(func() {
+	var err error
+
+	toCLI, err = client.NewCLI(os.Getenv("REGRESSION_TEST_TO_PATH"))
+	Expect(err).NotTo(HaveOccurred())
+
+	fromCLI, err = client.NewCLI(os.Getenv("REGRESSION_TEST_FROM_PATH"))
+	Expect(err).NotTo(HaveOccurred())
+
+	appName = fmt.Sprintf("regression-%d", time.Now().Unix())
```
efekarakus: I don't remember our app name char limit, but should we add an indication that this is for multisvcapp:
```suggestion
	appName = fmt.Sprintf("regression-mutisvcapp-%d", time.Now().Unix())
```

scripts/dryrun-creds.sh
```diff
@@ -1,19 +1,26 @@
...
+
+for profile in requiredProfiles:
+  print(profile)
```
iamhopaul123: Do we need this?

regression/multi-svc-app/multi_svc_app_suite_test.go
```diff
@@ -0,0 +1,56 @@
...
+	for _, svcName := range []string{"front-end", "www", "back-end"} {
+		err := os.Rename(fmt.Sprintf("%s/main.go", svcName), fmt.Sprintf("%s/swap/main.go", svcName))
+		Expect(err).NotTo(HaveOccurred())
+		err = os.Rename(fmt.Sprintf("%s/swap/main.tmp", svcName), fmt.Sprintf("%s/main.go", svcName))
+		Expect(err).NotTo(HaveOccurred())
+	}
+	err = os.Rename("query/entrypoint.sh", "query/swap/entrypoint.sh")
+	Expect(err).NotTo(HaveOccurred())
+	err = os.Rename("query/swap/entrypoint.tmp", "query/entrypoint.sh")
+	Expect(err).NotTo(HaveOccurred())
```
iamhopaul123: Why do we need this 🤔 Are all files gone after running the test with the container?

Lou1415926: you mean why do we swap them back? Because if you do `make regresssion-dryrun` on your machine and it doesn't swap back it sort up messes up your local files. Of course you can manually swap them back - not a great experience though.

regression/multi-svc-app/multi_svc_app_suite_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+var _ = AfterSuite(func() {
+	_, err := toCLI.Run("app", "delete", "--yes")
+	Expect(err).NotTo(HaveOccurred())
+
+	for _, svcName := range []string{"front-end", "www", "back-end"} {
+		err := os.Rename(fmt.Sprintf("%s/main.go", svcName), fmt.Sprintf("%s/swap/main.go", svcName))
+		Expect(err).NotTo(HaveOccurred())
+		err = os.Rename(fmt.Sprintf("%s/swap/main.tmp", svcName), fmt.Sprintf("%s/main.go", svcName))
+		Expect(err).NotTo(HaveOccurred())
```
efekarakus: Should these be best effort? For example what if the deployment fails with the old CLI then the cleanup step will also fail I think because `main.tmp` won't exist

regression/multi-svc-app/multi_svc_app_test.go
```diff
@@ -0,0 +1,250 @@
...
+	"github.com/aws/copilot-cli/regression/client"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("regression", func() {
+	var cli *client.CLI
+	var expectedWorkloadResponse = make(map[string]string)
+
+	deployWorkloadSpecs := func() {
```
efekarakus, line 22: nit: what do you think of extracting this function outside of `Describe` and taking the `cli` and `expectedWorkloadResponse` as params to remove the global vars? (just a nit tho, it's pretty clear in the test)
```go
func deployWorkloadsSpec(cli *client.CLI, wantedResponses map[string]string)
```

regression/multi-svc-app/multi_svc_app_test.go
```diff
@@ -0,0 +1,250 @@
...
+		It("service discovery should be enabled and working", func() {
+			// The front-end service is set up to have a path called
+			// "/front-end/service-discovery-test" - this route
+			// calls a function which makes a call via the service
+			// discovery endpoint, "back-end.local". If that back-end
+			// call succeeds, the back-end returns a response
+			// "back-end-service-discovery". This should be forwarded
+			// back to us via the front-end api.
+			// [test] -- http req -> [front-end] -- service-discovery -> [back-end]
+			By("Being able to call the service discovery endpoint from front-end")
```
efekarakus, line 82: :TIL: https://onsi.github.io/ginkgo/#documenting-complex-specs-by

regression/multi-svc-app/multi_svc_app_test.go
```diff
@@ -0,0 +1,250 @@
...
+		When("Deploying workloads", deployWorkloadSpecs)
+	})
+
+	When("Updating application code", func() {
+		It("should succeed", func() {
+			for _, svcName := range []string{"front-end", "www", "back-end"} {
+				err := os.Rename(fmt.Sprintf("%s/main.go", svcName), fmt.Sprintf("%s/swap/main.tmp", svcName))
+				Expect(err).NotTo(HaveOccurred())
+				err = os.Rename(fmt.Sprintf("%s/swap/main.go", svcName), fmt.Sprintf("%s/main.go", svcName))
+				Expect(err).NotTo(HaveOccurred())
```
efekarakus: nit: should we move these to a separate func? 
```go
func swapFiles(from, to string) error 
```

efekarakus (commented):
Sweet! looks great 😍

regression/multi-svc-app/multi_svc_app_test.go
Lou1415926: Would the below be an example?
```go
func deployWorkloadsSpec(cli *client.CLI, wantedResponses map[string]string) {}

var _ = Describe("regression", func() {
       When("Using the old CLI", func() {
              var cli *client.CLI
              var wantedResponses map[string]string
              BeforeEach(func() {
                     cli = fromCLI
                     wantedResponses = map[string]string{}
              })

              When("Deploying workloads", func() {
                     deployWorkloadsSpec(cli, wantedResponses)
              })
       })
})
```

This code snippet probably wouldn't work because ginkgo wouldn't be able to construct the spec tree during its ["Tree Construction Phase"](https://onsi.github.io/ginkgo/#mental-model-how-ginkgo-traverses-the-spec-hierarchy). 

It's possible to confine `cli` and `wantedResponse` into each `When("old")` block and `When("new")` block:
```go
var _ = Describe("regression", func() {

       When("Using the old CLI", func() {
              var cli *client.CLI
              BeforeEach(func() {
                     cli = fromCLI
              })

              When("Deploying workloads", func() {
                     It("should succeed", func() {
                            cli.Run()
                     })
              })
       })

       When("Using the new CLI", func() {
              var cli *client.CLI
              BeforeEach(func() {
                     cli = toCLI
              })

              When("Deploying workloads", func() {
                     It("should succeed", func() {
                            cli.Run()
                     })
              })
       })
})
```
However we wouldn't be able to reuse the specs like we are doing now with `deployWorkloadSpecs`, because:
```go
func deployWorkloadsSpec() {
       It("should succeed", func() {
              cli.Run() // Error: cli variable is not accessible
       }
}

var _ = Describe("regression", func() {
       When("Using the old CLI", func() {
              var cli *client.CLI
              BeforeEach(func() {
                     cli = fromCLI
              })

              When("Deploying workloads", deployWorkloadsSpec)
       })
})

// This doesn't work either.
var _ = Describe("regression", func() {
       When("Using the old CLI", func() {
              var cli *client.CLI = fromCLI // This code would be invoked at "tree phase" at which time `fromCLI` is `nil` yet.
              When("Deploying workloads", deployWorkloadsSpec) 
       })
})
```

Because of ginkgo's "Tree Construction Phase" and "Run Phase", we have several constraints:
1. The value of `cli` and `wantedResponses` should be assigned in `BeforeEach` (or `It`, just any "set up node" or "subject node"; not "container node") so that the value assignment actually happens during ginkgo's "Run Phase". This means it's required that their declaration are somewhere accessible by `BeforeEach`.
2. If we want to reuse the specs, then `cli` and `wantedResponses` need to be accessible by `deployWorkloadSpecs` as well, in addition to `BeforeEach`.

`deployWorkloadSpecs` is special in that it's supposed to be invoked during "tree phase", not "run phase". Therefore, it can't take any parameters whose actual value needs to be assigned during "run phase". Given all these, and given that we'd like to order the specs with this hierarchy, I've found that the smallest scope of `cli`, `wantedResponses` are probably inside of `Describe`.

Sorry for the lengthy response 🙇🏼‍♀️ ! The "Tree phase" and "Run phase" gave me some headache when I was writing the test suite, so I thought it'd probably be beneficial for me to share some learnings here :meow_wine

efekarakus: oh that's really interesting, thanks for sharing!

efekarakus (approved):

iamhopaul123 (approved):

PR #3856, opened 2022-08-05
chore: show URI after workload deploy and fix backward compatibility

iamhopaul123:
Due to the dynamicworkload type, workload deploy temporarily can't show any URL after deployment. Also currently svc show won't show http URL correctly for env lastly deployed by v1.20 customers.

### DIFF
--- a/internal/pkg/describe/uri.go
+++ b/internal/pkg/describe/uri.go
@@ -230,7 +230,8 @@ func (d *uriDescriber) envDNSName(path string) (accessURI, error) {
 	if err != nil {
 		return accessURI{}, fmt.Errorf("get stack outputs for environment %s: %w", d.env, err)
 	}
-	// Backward compatibility concern since envOutputPublicALBAccessible didn't exist before.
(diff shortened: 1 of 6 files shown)


internal/pkg/describe/uri.go
```diff
@@ -230,7 +230,8 @@ func (d *uriDescriber) envDNSName(path string) (accessURI, error) {
 	if err != nil {
 		return accessURI{}, fmt.Errorf("get stack outputs for environment %s: %w", d.env, err)
 	}
-	if accessible, ok := envOutputs[envOutputPublicALBAccessible]; ok && accessible == "true" {
+	// Backward compatibility concern since envOutputPublicALBAccessible didn't exist before.
```
Lou1415926: nit: what do you think of
```go
if d.envDescriber.Version() < "v1.18.0" {}
```
to (perhaps, idk if it does) make it easier to understand for future?

Lou1415926 (approved):

dannyrandall (approved):

PR #3857, opened 2022-08-05
chore: fix small issues with cloudfront

CaptainCarpensir:
Summary of changes:
- Addresses feedback from @efekarakus on PR #3842 
- Fixes issue where wrong error was given when no aliases are specified when importing certs with cloudfront
- Adds manifest validation for users trying to use a certificate for cloudfront in the wrong region

### DIFF
--- a/internal/pkg/cli/deploy/svc.go
+++ b/internal/pkg/cli/deploy/svc.go
@@ -23,6 +23,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ssm"
 
 	"github.com/aws/copilot-cli/internal/pkg/aws/acm"
+	"github.com/aws/copilot-cli/internal/pkg/aws/cloudfront"
 	"github.com/aws/copilot-cli/internal/pkg/aws/identity"
 	"github.com/aws/copilot-cli/internal/pkg/template/artifactpath"
 
@@ -1401,7 +1402,7 @@ func (d *lbWebSvcDeployer) validateALBRuntime() error {
 
 		cdnCert := d.environmentConfig.CDNConfig.Config.Certificate
 		albCertValidator := d.newAliasCertValidator(nil)
-		cfCertValidator := d.newAliasCertValidator(aws.String(manifest.EnvCloudFrontCertRegion))
+		cfCertValidator := d.newAliasCertValidator(aws.String(cloudfront.CertRegion))
 		if err := albCertValidator.ValidateCertAliases(aliases, d.environmentConfig.HTTPConfig.Public.Certificates); err != nil {
 			return fmt.Errorf("validate aliases against the imported public ALB certificate for env %s: %w", d.env.Name, err)
 		}
--- a/internal/pkg/cli/deploy/svc.go
+++ b/internal/pkg/cli/deploy/svc.go
@@ -1401,7 +1401,7 @@ func (d *lbWebSvcDeployer) validateALBRuntime() error {
 
 		cdnCert := d.environmentConfig.CDNConfig.Config.Certificate
 		albCertValidator := d.newAliasCertValidator(nil)
-		cfCertValidator := d.newAliasCertValidator(aws.String(envCloudFrontCertRegion))
+		cfCertValidator := d.newAliasCertValidator(aws.String(manifest.EnvCloudFrontCertRegion))
 		if err := albCertValidator.ValidateCertAliases(aliases, d.environmentConfig.HTTPConfig.Public.Certificates); err != nil {
 			return fmt.Errorf("validate aliases against the imported public ALB certificate for env %s: %w", d.env.Name, err)
 		}
--- a/internal/pkg/cli/env_deploy.go
+++ b/internal/pkg/cli/env_deploy.go
@@ -143,7 +143,7 @@ func (o *deployEnvOpts) Execute() error {
 	if o.isManagedCDNEnabled(mft) {
 		describer, err := o.newEnvDescriber()
 		if err != nil {
-			return fmt.Errorf("new env describer: %w", err)
+			return err
 		}
 		// With managed domain, if the customer isn't using `alias` the A-records are inserted in the service stack as each service domain is unique.
 		// However, when clients enable CloudFront, they would need to update all their existing records to now point to the distribution.
--- a/internal/pkg/cli/env_deploy.go
+++ b/internal/pkg/cli/env_deploy.go
@@ -127,7 +127,7 @@ func (o *deployEnvOpts) Ask() error {
 }
 
 func (o *deployEnvOpts) isManagedCDNEnabled(mft *manifest.Environment) bool {
-	return mft.CDNConfig.CDNEnabled() && mft.HTTPConfig.Public.Certificates == nil && o.targetApp.Domain != ""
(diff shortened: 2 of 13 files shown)


internal/pkg/cli/env_deploy.go
```diff
@@ -136,11 +140,14 @@ func (o *deployEnvOpts) Execute() error {
 	if err != nil {
 		return err
 	}
-	if mft.CDNConfig.CDNEnabled() && mft.HTTPConfig.Public.Certificates == nil && o.targetApp.Domain != "" {
+	if o.isManagedCDNEnabled(mft) {
 		describer, err := o.newEnvDescriber()
 		if err != nil {
-			return fmt.Errorf("describe env: %w", err)
+			return fmt.Errorf("new env describer: %w", err)
```
iamhopaul123: it's ok we don't need to wrap this error (private func)

internal/pkg/describe/env.go
```diff
@@ -336,7 +336,9 @@ func (d *EnvDescriber) ValidateCFServiceDomainAliases() error {
 	}
 
 	var aliases map[string][]string
-	err = json.Unmarshal([]byte(jsonOutput), &aliases)
+	if jsonOutput != "" {
+		err = json.Unmarshal([]byte(jsonOutput), &aliases)
+	}
 	if err != nil {
```
iamhopaul123: Should this be part of the prev if block?

internal/pkg/manifest/env.go
```diff
@@ -17,6 +17,10 @@ import (
 // EnvironmentManifestType identifies that the type of manifest is environment manifest.
 const EnvironmentManifestType = "Environment"
 
+const (
+	EnvCloudFrontCertRegion = "us-east-1"
```
iamhopaul123: Comment for public var pls

internal/pkg/manifest/errors.go
```diff
@@ -119,6 +119,19 @@ func (e *errAtLeastOneFieldMustBeSpecified) Error() string {
 	return errMsg
 }
 
+type errInvalidCloudFrontRegion struct{}
+
+func (e *errInvalidCloudFrontRegion) Error() string {
+	return fmt.Sprintf(`cdn certificate must belong to region %s`, EnvCloudFrontCertRegion)
```
iamhopaul123:
```suggestion
	return fmt.Sprintf(`cdn certificate must be in region %s`, EnvCloudFrontCertRegion)
```

internal/pkg/manifest/validate_env.go
```diff
@@ -39,6 +39,9 @@ func (e EnvironmentConfig) validate() error {
 	if err := e.Network.VPC.SecurityGroupConfig.validate(); err != nil {
 		return fmt.Errorf(`validate "security_group": %w`, err)
 	}
+	if err := e.CDNConfig.validate(); err != nil {
+		return fmt.Errorf(`validate "cdn config": %w`, err)
```
iamhopaul123: the users won't know what "cdn config" means. They can only see `cdn` in the env manifest doc.
```suggestion
		return fmt.Errorf(`validate "cdn": %w`, err)
```

internal/pkg/cli/env_deploy.go
```diff
@@ -126,6 +126,10 @@ func (o *deployEnvOpts) Ask() error {
 	return o.validateOrAskEnvName()
 }
 
+func (o *deployEnvOpts) isManagedCDNEnabled(mft *manifest.Environment) bool {
+	return mft.CDNConfig.CDNEnabled() && mft.HTTPConfig.Public.Certificates == nil && o.targetApp.Domain != ""
```
iamhopaul123: nit: i feel like the public method should just be `Enabled()` since it belongs to `CDNConfig`

iamhopaul123 (approved):

internal/pkg/cli/env_deploy.go
CaptainCarpensir: This causes a conflict with the variable `Enabled *bool`, not very big of an issue, but I need to think of what to rename that variable to for this to work first.

internal/pkg/cli/deploy/svc.go
```diff
@@ -1401,7 +1401,7 @@ func (d *lbWebSvcDeployer) validateALBRuntime() error {
 
 		cdnCert := d.environmentConfig.CDNConfig.Config.Certificate
 		albCertValidator := d.newAliasCertValidator(nil)
-		cfCertValidator := d.newAliasCertValidator(aws.String(envCloudFrontCertRegion))
+		cfCertValidator := d.newAliasCertValidator(aws.String(manifest.EnvCloudFrontCertRegion))
```
efekarakus: It's just a bit awkward to read the CloudFront cert region from the `manifest` pkg ya know? the `manifest` pkg should be responsible for fields and functionality about `manifest.yml` files whereas this seems to be a concern related to CloudFront.

What do you think of creating just a `aws/cloudfront` pkg with a single constant in it 😂 ?

```go
package cloudfront

const (
  // CertRegion is the only AWS region accepted by CloudFront while attaching certificates to a distribution.
  CertRegion = "us-east-1"
)
```

internal/pkg/cli/env_deploy.go
efekarakus: One option is also to move the method to the `mft` level :) The pattern that @paragbhingre  has been following is like this:
```go
func (cfg *environmentVPCConfig) CDN() (AdvancedCDNConfig, bool) {
	if cfg.CDNConfig.isEmpty() {
		return nil, false
	}
	return cfg.CDNConfig.Advanced, true
}
```
the clients can then do:
```go
_, cdnEnabled := mft.CDN()
```

internal/pkg/describe/env.go
```diff
@@ -336,10 +336,13 @@ func (d *EnvDescriber) ValidateCFServiceDomainAliases() error {
 	}
 
 	var aliases map[string][]string
-	err = json.Unmarshal([]byte(jsonOutput), &aliases)
-	if err != nil {
-		return fmt.Errorf("unmarshal %q: %w", jsonOutput, err)
+	if jsonOutput != "" {
```
efekarakus, line 339: Can we add a testcase for this 🥺

internal/pkg/manifest/validate_env.go
```diff
@@ -347,9 +350,13 @@ func (i RestrictiveIngress) validate() error {
 // validate returns nil if advancedCDNConfig is configured correctly.
 func (cfg advancedCDNConfig) validate() error {
 	if cfg.Certificate != nil {
```
efekarakus: nit: can we flip the if-statement to create [a guard-clause](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) to reduce nesting:

```go
if cfg.Certificate == nil { return nil }
// existing logic
```

efekarakus (commented):
Awesome, thank you Aiden!

internal/pkg/cli/deploy/svc.go
CaptainCarpensir: Yeah I moved it because of circular dependencies, but as funny as it is this actually seems like a good solution lol 👍

efekarakus (approved):

PR #3878, opened 2022-08-10
chore: error out import hosted zone when CDN is enabled

iamhopaul123:

### DIFF
--- a/internal/pkg/cli/deploy/errors.go
+++ b/internal/pkg/cli/deploy/errors.go
@@ -31,11 +31,11 @@ func (e *errSvcWithALBAliasHostedZoneWithCDNEnabled) Error() string {
 // Implements main.actionRecommender interface.
 func (e *errSvcWithALBAliasHostedZoneWithCDNEnabled) RecommendActions() string {
 	msgs := []string{
-		"If you already have an Load Balanced Web Service deployed, you can switch to CDN by:",
-		fmt.Sprintf("1. manually update the A-record value to be the CDN distribution domain name. Please make sure your A-record has %q set to %s.",
+		"If you already have a Load Balanced Web Service deployed, you can switch to CDN by:",
+		fmt.Sprintf(" 1. Update the A-record value to be the CDN distribution domain name. Please make sure your A-record has %q set to %s.",
 			"Evaluate target health", color.HighlightCode("True")),
-		fmt.Sprintf("2. Remove the %s setting from the service manifest.", color.HighlightCode(`"http.alias.hosted_zone"`)),
-		fmt.Sprintf("3. Redeploy the service via %s.", color.HighlightCode("copilot svc deploy")),
+		fmt.Sprintf(" 2. Remove the %s setting from the service manifest.", color.HighlightCode(`"http.alias.hosted_zone"`)),
+		fmt.Sprintf(" 3. Redeploy the service via %s.", color.HighlightCode("copilot svc deploy")),
 	}
 	return strings.Join(msgs, "\n")
 }
--- a/internal/pkg/cli/deploy/errors.go
+++ b/internal/pkg/cli/deploy/errors.go
@@ -31,11 +31,11 @@ func (e *errSvcWithALBAliasHostedZoneWithCDNEnabled) Error() string {
 // Implements main.actionRecommender interface.
 func (e *errSvcWithALBAliasHostedZoneWithCDNEnabled) RecommendActions() string {
 	msgs := []string{
-		"If you already have an Load Balanced Web Service deployed, you can switch to CDN by running:",
-		fmt.Sprintf("1. manually update the A-record value to be the CDN distribution domain name with %q set to %s.",
+		"If you already have an Load Balanced Web Service deployed, you can switch to CDN by:",
+		fmt.Sprintf("1. manually update the A-record value to be the CDN distribution domain name. Please make sure your A-record has %q set to %s.",
 			"Evaluate target health", color.HighlightCode("True")),
-		"2. Remove the alias hosted zones setting in the service manifest.",
-		fmt.Sprintf("3. Reploy the service %s.", color.HighlightCode("copilot svc deploy")),
+		fmt.Sprintf("2. Remove the %s setting from the service manifest.", color.HighlightCode(`"http.alias.hosted_zone"`)),
+		fmt.Sprintf("3. Redeploy the service via %s.", color.HighlightCode("copilot svc deploy")),
 	}
 	return strings.Join(msgs, "\n")
 }
--- a/internal/pkg/cli/deploy/errors.go
+++ b/internal/pkg/cli/deploy/errors.go
@@ -3,7 +3,12 @@
 
 package deploy
 
-import "fmt"
+import (
+	"fmt"
+	"strings"
+
+	"github.com/aws/copilot-cli/internal/pkg/term/color"
+)
 
 type errSvcWithNoALBAliasDeployingToEnvWithImportedCerts struct {
 	name    string
@@ -12,4 +17,25 @@ type errSvcWithNoALBAliasDeployingToEnvWithImportedCerts struct {
 
 func (e *errSvcWithNoALBAliasDeployingToEnvWithImportedCerts) Error() string {
 	return fmt.Sprintf("cannot deploy service %s without http.alias to environment %s with certificate imported", e.name, e.envName)
+}
+
+type errSvcWithALBAliasHostedZoneWithCDNEnabled struct {
+	envName string
+}
+
+func (e *errSvcWithALBAliasHostedZoneWithCDNEnabled) Error() string {
+	return fmt.Sprintf("cannot specify alias hosted zones when cdn is enabled in environment %q", e.envName)
+}
+
+// RecommendActions returns recommended actions to be taken after the error.
+// Implements main.actionRecommender interface.
+func (e *errSvcWithALBAliasHostedZoneWithCDNEnabled) RecommendActions() string {
+	msgs := []string{
+		"If you already have an Load Balanced Web Service deployed, you can switch to CDN by running:",
+		fmt.Sprintf("1. manually update the A-record value to be the CDN distribution domain name with %q set to %s.",
+			"Evaluate target health", color.HighlightCode("True")),
+		"2. Remove the alias hosted zones setting in the service manifest.",
+		fmt.Sprintf("3. Reploy the service %s.", color.HighlightCode("copilot svc deploy")),
+	}
+	return strings.Join(msgs, "\n")
 }
--- a/internal/pkg/cli/deploy/svc.go
+++ b/internal/pkg/cli/deploy/svc.go
@@ -153,10 +153,6 @@ type aliasCertValidator interface {
 	ValidateCertAliases(aliases []string, certs []string) error
 }
 
-type configDescriber interface {
-	Manifest() ([]byte, error)
-}
-
 type workloadDeployer struct {
 	name          string
 	app           *config.Application
--- a/internal/pkg/cli/deploy/svc.go
+++ b/internal/pkg/cli/deploy/svc.go
@@ -176,30 +176,13 @@ type workloadDeployer struct {
 	endpointGetter     endpointGetter
 	spinner            spinner
 	templateFS         template.Reader
-	envConfigDescriber configDescriber
 
 	// Cached variables.
 	defaultSess              *session.Session
 	defaultSessWithEnvRegion *session.Session
 	envSess                  *session.Session
 	store                    *config.Store
-	environmentConfig        *manifest.Environment
-}
-
-func (d *workloadDeployer) cachedEnvironmentConfig() (*manifest.Environment, error) {
-	if d.environmentConfig != nil {
-		return d.environmentConfig, nil
-	}
-	mft, err := d.envConfigDescriber.Manifest()
-	if err != nil {
-		return nil, fmt.Errorf("read the manifest used to deploy environment %s: %w", d.env.Name, err)
-	}
-	env, err := manifest.UnmarshalEnvironment(mft)
-	if err != nil {
-		return nil, fmt.Errorf("unmarshal the manifest used to deploy environment %s: %w", d.env.Name, err)
-	}
-	d.environmentConfig = env
-	return d.environmentConfig, nil
+	envConfig                *manifest.Environment
 }
 
 // WorkloadDeployerInput is the input to for workloadDeployer constructor.
@@ -265,6 +248,16 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 	if err != nil {
 		return nil, fmt.Errorf("initiate env describer: %w", err)
 	}
+
+	mft, err := envDescriber.Manifest()
+	if err != nil {
+		return nil, fmt.Errorf("read the manifest used to deploy environment %s: %w", in.Env.Name, err)
+	}
+	envConfig, err := manifest.UnmarshalEnvironment(mft)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshal the manifest used to deploy environment %s: %w", in.Env.Name, err)
+	}
+
 	return &workloadDeployer{
 		name:               in.Name,
 		app:                in.App,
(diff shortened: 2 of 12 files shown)


internal/pkg/manifest/env.go
```diff
@@ -93,7 +93,7 @@ type EnvironmentConfig struct {
 	Network       environmentNetworkConfig `yaml:"network,omitempty,flow"`
 	Observability environmentObservability `yaml:"observability,omitempty,flow"`
 	HTTPConfig    EnvironmentHTTPConfig    `yaml:"http,omitempty,flow"`
-	CDNConfig     environmentCDNConfig     `yaml:"cdn,omitempty,flow"`
+	CDNConfig     CDNConfig                `yaml:"cdn,omitempty,flow"`
```
CaptainCarpensir: What's the logic behind renaming this? Is the resource no longer solely associated with the environment? I think it would be completely fine to just capitalize `EnvironmentCDNConfig` to export the field in this case.

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
```diff
@@ -38,8 +38,13 @@ LoadBalancerDNSAlias{{$hostedZoneID}}:
           HostedZoneId: !GetAtt EnvControllerAction.InternalLoadBalancerHostedZone
           DNSName: !GetAtt EnvControllerAction.InternalLoadBalancerDNSName
           {{- else}}
+            {{- if $.CDNEnabled}}
```
CaptainCarpensir: This change means that when the hosted zone is updated with the CloudFront resources, the stack Resource is not longer accurately named and described. I'd suggest we rename `LoadBalancerDNSAlias{{$hostedZoneID}}` to `PublicAccessDNSAlias{{$hostedZoneID}}` and the description accordingly to fit with the other changes like those to the Load Balancer Security Group.

internal/pkg/manifest/env.go
iamhopaul123: It's because in `template` pkg integ test we need this. And the reason why I remove `environment` as the prefix is because CDN only exists in env manifest. So we probably don't need to namespace it.

iamhopaul123:
For backward compatibility concern we have to keep the resource `AWS::Route53::RecordSetGroup` logical name as `LoadBalancerDNSAlias{{$hostedZoneID}}` because otherwise if they already have this resource in the stack, changing resource name/adding a new resource with the same alias target will fail (since route53 only allow one alias target at a time and the resource itself is not updatable).

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -31,7 +31,6 @@ type BackendService struct {
 	*ecsWkld
 	manifest     *manifest.BackendService
 	httpsEnabled bool
-	certImported bool
```
Lou1415926, line 34: ohhh this is great news

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -35,6 +35,7 @@ type LoadBalancedWebService struct {
 	manifest               *manifest.LoadBalancedWebService
 	httpsEnabled           bool
 	certImported           bool
```
Lou1415926: is `certImported` no longer needed in backend service but still needed for LBWS?

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
Lou1415926: Should we or is it possible for us to create another resource for this record, instead of fitting it into this resource that was designed for Load Balancer DNS?

internal/pkg/template/templates/workloads/partials/cf/env-controller.yml
```diff
@@ -7,6 +7,11 @@ EnvControllerAction:
     Workload: !Ref WorkloadName
 {{- if .Aliases}}
     Aliases: {{ fmtSlice (quoteSlice .Aliases) }}
+{{- end}}
+{{- if $.CDNEnabled}}
+    # This will not be processed by env controller, but trigger env controller to
```
Lou1415926: Can this cause "No updates to be performed" error from env stack?

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
iamhopaul123: the thing is they could use the same alias. For example: `example.com` is used for the service and the target is LB DNS name. We can't create another resource with the same `example.com` pointing to the CF without deleting this resource first.

internal/pkg/template/templates/workloads/partials/cf/env-controller.yml
efekarakus: Should we just not use the env controller's outputs here? like don't force refresh but instead if `$.CDNEnabled` then in `LoadBalancerDNSAlias` resource we do:
```
Fn::ImportValue:
          !Sub "${App}-${Env}-CloudFrontDomainName"
```

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
iamhopaul123: `certImported` is used [here](https://github.com/aws/copilot-cli/blob/e9368805ed73a3232808862a52684e0458954910/internal/pkg/deploy/cloudformation/stack/lb_web_svc.go#L180-L182)

internal/pkg/template/templates/workloads/partials/cf/env-controller.yml
iamhopaul123: It will not if there's no env param diff.

iamhopaul123: Oh that's a good idea. Let me try that.

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
```diff
@@ -24,11 +24,11 @@ LoadBalancerDNSAlias:
 {{- range $hostedZoneID, $aliases := .HostedZoneAliases}}
 LoadBalancerDNSAlias{{$hostedZoneID}}:
   Metadata:
-    'aws:copilot:description': 'Alias records for the application load balancer in hosted zone {{$hostedZoneID}}'
+    'aws:copilot:description': 'Alias records for the service in hosted zone {{$hostedZoneID}}'
```
efekarakus: :wutface: I think I'm misunderstanding something if there is a `hosted_zone` field specified in the manifest then why are we inserting into the CloudFront hosted zone?

Should the A record be inserted outside of this range: `{{- range $hostedZoneID, $aliases := .HostedZoneAliases}}`? can you explain it to me

iamhopaul123: > why are we inserting into the CloudFront hosted zone

I don't understand the question...we are not inserting into the CF hosted zone. When adding an A-record we need to specify hosted zone for the resource (aka, `PublicLoadBalancerHostedZone` for ALB and each ALB it is different and `Z2FDTNDATAQYW2` for CF the same universally)

>Should the A record be inserted outside of this range: {{- range $hostedZoneID, $aliases := .HostedZoneAliases}}? 

why? An A-record corresponds to an alias.

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
Lou1415926: oh interesting...that does seem to be more fitting in `cli/deploy`. We can refactor it later.

iamhopaul123: Yeah that's what I thought too. Basically we need a place to validate the svc manifest against the env it is trying to deploy to. Maybe we should have dynamic manifest validation? Right now the validation starts from the static manifest.

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
efekarakus: nvm boss I misread 😅 it totally makes sense

internal/pkg/template/templates/workloads/partials/cf/env-controller.yml
iamhopaul123: <strike>Ok i've tried to switch between having CDN and no CDN and I think it works!</strike>

Updated: this is not gonna work for switching back from using CDN -> not using CDN. Because the CloudFront will not be deletable since the output is used in another stack 🤦. We still need the env controller to decouple them.

iamhopaul123: We'll just error out, because it is almost impossible without downtime. And eventually we'll migrate all the A-record resource within the workload stack to the env stack lambda.

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -175,9 +177,13 @@ func (s *LoadBalancedWebService) Template() (string, error) {
 	if err != nil {
 		return "", err
 	}
+	// TODO: need to be refactored because we can error out earlier.
```
efekarakus: nit: can you add the package where the refactor should go to? it sounds like it can go to `manifest` validation 💭

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -175,9 +177,13 @@ func (s *LoadBalancedWebService) Template() (string, error) {
...
 	if len(aliasesFor) != 0 && !s.certImported {
 		return "", fmt.Errorf("cannot specify alias hosted zones when env certificates are managed by Copilot")
 	}
+	if len(aliasesFor) != 0 && s.cdnEnabled {
+		return "", fmt.Errorf("cannot specify alias hosted zones when cdn is enabled in environment %q", s.env)
```
efekarakus: Should we add a `RecommendedAction` to this error? Telling the user to manage the A-record on their own to point to switch at the appropriate time between CDN and ALB.

efekarakus (commented):
LGTM, just a suggestion

internal/pkg/template/templates/workloads/partials/cf/env-controller.yml
Lou1415926: +1 to PH and just to share some thoughts with fellow reviewers 👀 . There are currently two types of aliases being managed in service stack:
1. Default alias
2. Aliases that are validated by imported certificates AND have `"alias.hosted_zone"` field (i.e. user wants us to upsert the A records for them)

We've decided to error out for case 1, I would +1 to erroring out for case 2 (exactly what https://github.com/aws/copilot-cli/pull/3878#discussion_r943942248 is saying) as well 🤔  I think it's not a bad idea to stop building on top of the service stack aliases if we are not sure about the implementation. I also believe both use cases 1 and 2 are on the rarer side.

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
Lou1415926: This `cdnEnabled` field seems to be here only for validation purpose (just like `certImported`). We either want to make it clear in the comment that both fields need to be refactored, or just move this `cdn` validation to maybe `cli/deploy` or `manifest` in this PR

internal/pkg/cli/deploy/errors.go
```diff
@@ -13,3 +18,24 @@ type errSvcWithNoALBAliasDeployingToEnvWithImportedCerts struct {
...
+
+// RecommendActions returns recommended actions to be taken after the error.
+// Implements main.actionRecommender interface.
+func (e *errSvcWithALBAliasHostedZoneWithCDNEnabled) RecommendActions() string {
+	msgs := []string{
+		"If you already have an Load Balanced Web Service deployed, you can switch to CDN by running:",
+		fmt.Sprintf("1. manually update the A-record value to be the CDN distribution domain name with %q set to %s.",
+			"Evaluate target health", color.HighlightCode("True")),
+		"2. Remove the alias hosted zones setting in the service manifest.",
+		fmt.Sprintf("3. Reploy the service %s.", color.HighlightCode("copilot svc deploy")),
```
Lou1415926:
```suggestion
	msgs := []string{
		"If you already have an Load Balanced Web Service deployed, you can switch to CDN by:",
		fmt.Sprintf("1. manually update the A-record value to be the CDN distribution domain name. Please make sure that your A-record has %q set to %s.",
			"Evaluate target health", color.HighlightCode("True")),
		fmt.Sprintf("2. Remove the %s setting from the service manifest.", code.HightlightCode(`"http.alias.hosted_zone"`)),
		fmt.Sprintf("3. Redeploy the service via %s.", color.HighlightCode("copilot svc deploy")),
```

internal/pkg/cli/deploy/svc.go
```diff
@@ -265,6 +244,16 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 	if err != nil {
 		return nil, fmt.Errorf("initiate env describer: %w", err)
 	}
+
```
Lou1415926, line 247: I guess the pros of the old way is to read and unmarshal the manifest only when it's needed (e.g. when the image wasn't able to build, it'd error out before we need the manifest - the read/unmarshal would be a waste). But this LGTM me too - less code, less function calls

internal/pkg/cli/deploy/svc.go
```diff
@@ -1369,26 +1346,15 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
...
-	}
-	return len(env.HTTPConfig.Private.Certificates) != 0, nil
-}
 func (d *lbWebSvcDeployer) validateALBRuntime() error {
-	hasImportedCerts, err := d.envHasImportedCertificates()
-	if err != nil {
-		return err
-	}
-
+	hasImportedCerts := len(d.envConfig.HTTPConfig.Public.Certificates) != 0
```
Lou1415926, line 1357: noice

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -177,24 +173,14 @@ func (s *LoadBalancedWebService) Template() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	// TODO: need to be refactored because we can error out earlier.
```
Lou1415926: Thanks you for refactoring this!!

Lou1415926 (approved):

internal/pkg/cli/deploy/svc.go
iamhopaul123: exactly. i removed this because i noticed recently added code use the cache var directly instead of calling that function, which is dangerous because if the func is never called before then they'll get the panic.

internal/pkg/cli/deploy/errors.go
```diff
@@ -13,3 +18,24 @@ type errSvcWithNoALBAliasDeployingToEnvWithImportedCerts struct {
...
+
+func (e *errSvcWithALBAliasHostedZoneWithCDNEnabled) Error() string {
+	return fmt.Sprintf("cannot specify alias hosted zones when cdn is enabled in environment %q", e.envName)
+}
+
+// RecommendActions returns recommended actions to be taken after the error.
+// Implements main.actionRecommender interface.
+func (e *errSvcWithALBAliasHostedZoneWithCDNEnabled) RecommendActions() string {
+	msgs := []string{
+		"If you already have an Load Balanced Web Service deployed, you can switch to CDN by:",
```
efekarakus:
```suggestion
		"If you already have a Load Balanced Web Service deployed, you can switch to CDN by:",
```

internal/pkg/cli/deploy/errors.go
```diff
@@ -13,3 +18,24 @@ type errSvcWithNoALBAliasDeployingToEnvWithImportedCerts struct {
...
+		fmt.Sprintf("1. manually update the A-record value to be the CDN distribution domain name. Please make sure your A-record has %q set to %s.",
```
efekarakus:
```suggestion
		fmt.Sprintf("  1. Update the A-record value to be the CDN distribution domain name. Please make sure your A-record has %q set to %s.",
```

internal/pkg/cli/deploy/errors.go
```diff
@@ -13,3 +18,24 @@ type errSvcWithNoALBAliasDeployingToEnvWithImportedCerts struct {
...
+			"Evaluate target health", color.HighlightCode("True")),
+		fmt.Sprintf("2. Remove the %s setting from the service manifest.", color.HighlightCode(`"http.alias.hosted_zone"`)),
+		fmt.Sprintf("3. Redeploy the service via %s.", color.HighlightCode("copilot svc deploy")),
```
efekarakus:
```suggestion
		fmt.Sprintf("  2. Remove the %s setting from the service manifest.", color.HighlightCode(`"http.alias.hosted_zone"`)),
		fmt.Sprintf("  3. Redeploy the service via %s.", color.HighlightCode("copilot svc deploy")),
```

efekarakus (approved):

internal/pkg/cli/deploy/errors.go
iamhopaul123: 🤦

PR #3880, opened 2022-08-10
docs: add command page for `env deploy`

Lou1415926:
As well as adding a version number beside the title.

Screen Shot 2022-08-10 at 4 54 00 PM

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -69,24 +69,24 @@ nav:
         - app upgrade: docs/commands/app-upgrade.en.md
         - app delete: docs/commands/app-delete.en.md
         - env init: docs/commands/env-init.en.md
-        - env deploy: docs/commands/env-deploy.en.md
         - env package: docs/commands/env-package.en.md
         - env delete: docs/commands/env-delete.en.md
         - job init: docs/commands/job-init.en.md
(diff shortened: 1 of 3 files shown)


mkdocs.yml
```diff
@@ -69,6 +69,7 @@ nav:
         - app upgrade: docs/commands/app-upgrade.en.md
         - app delete: docs/commands/app-delete.en.md
         - env init: docs/commands/env-init.en.md
+        - env deploy: docs/commands/env-deploy.en.md
```
efekarakus: 🤔 should our `deploy` commands be under `Release`?

efekarakus (approved):

huanjani (approved):
Yay for version! Looks good. Adding DNM label in case you want to move deploy(s) to Release.

mkdocs.yml
Lou1415926: I can see arguments for both! It feels to me that people run `deploy` commands in an earlier stage when they are setting up the infrastructure - to test out the infra, to know about Copilot, etc. Hence `Build`.  After this stage they use some CICD (GitHub actions, CodePipelines... etc.) to "release" their application changes, hence `Release`.

PR #3881, opened 2022-08-11
chore: generate cdn alternate domain names

dannyrandall:


cf-custom-resources/test/unique-aliases-test.js
```diff
@@ -0,0 +1,122 @@
...
+    "svc1": ["svc1.com"],
+    "svc2": ["svc2.com"]
+  }, ["svc1.com", "svc2.com"]);
+
+  aliasTest("two services with a common alias", {
+    "svc1": ["svc1.com", "example.com"],
+    "svc2": ["svc2.com", "example.com"]
+  }, ["example.com", "svc1.com", "svc2.com"]);
+
+  aliasTest("two services with a common alias", {
```
CaptainCarpensir: This test is a duplicate of the one above it

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -1,10 +1,53 @@
...
+    Aliases: !Ref Aliases
+
 CloudFrontDistribution:
   Metadata:
     'aws:copilot:description': 'A CloudFront distribution for global content delivery'
   Condition: CreateALB
   Type: AWS::CloudFront::Distribution
   Properties:
     DistributionConfig:
+      Aliases: !GetAtt UniqueAliasesAction.UniqueAliases
```
CaptainCarpensir: Does the distribution have a configuration error when there are Aliases but no SSL/TLS certificate? I can test this and get back to you.

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -1,10 +1,53 @@
...
+              - lambda.amazonaws.com
+          Action:
+            - sts:AssumeRole
+    Path: /
+    ManagedPolicyArns:
+      - !Sub arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+
+UniqueAliasesFunction:
+  Type: AWS::Lambda::Function
+  Condition: CreateALB
```
CaptainCarpensir, line 23: Perfect condition 😄

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -1,10 +1,53 @@
+UniqueAliasesFunctionRole:
```
CaptainCarpensir: These three added resources are missing in the integ test in `template-with-imported-certs-observability.yml`

internal/pkg/template/templates/environment/partials/cdn-resources.yml
CaptainCarpensir: Yes, this will error out if the user has working HTTPS and then tries to remove it. This happens because they will have no SSL/TLS Certificate, but the `Aliases` parameter of the env stack isn't removed until they redeploy the service.

This leaves the user in deadlock because the env can't be redeployed until all `Aliases` are removed, but the services can't be redeployed while the imported certificates still require those `Aliases`. This makes the only option to delete the services in order to redeploy the environment.

You can just move the `Aliases` field into the same conditions as `ViewerCertificate` as they're mutually dependent 👍

.gitignore
```diff
@@ -7,4 +7,5 @@ node_modules/
 .DS_Store
 site/node_modules/
 site/themes/docsy
-internal/pkg/template/templates/custom-resources/
\ No newline at end of file
+internal/pkg/template/templates/custom-resources/
+copilot/
```
efekarakus, line 11: ? why :P

cf-custom-resources/lib/unique-aliases.js
```diff
@@ -0,0 +1,141 @@
...
+ *  "svc2": ["example.com"]
+ * }
+ * 
+ * This function returns a list of unique aliases found in that list.
+ * From the previous example, UniqueAliases would be:
+ * ["svc1.com", "example.com"]
+ */
+exports.handler = async function (event, context) {
+  const responseData = {};
+  const physicalResourceId = event.PhysicalResourceId || `unique-aliases-${event.LogicalResourceId}`;
```
efekarakus: nit: maybe just since it will render as `unique-aliases-UniqueAliasesAction`
```suggestion
  const physicalResourceId = event.PhysicalResourceId || event.LogicalResourceId;
```

cf-custom-resources/lib/unique-aliases.js
```diff
@@ -0,0 +1,141 @@
...
+
+  console.log(`Event: ${JSON.stringify(event)}`);
```
efekarakus: can we remove the console log statements to not charge customers too much for CloudWatch  (except the exception one)

cf-custom-resources/lib/unique-aliases.js
```diff
@@ -0,0 +1,141 @@
...
+        break;
+      case "Delete":
+        // Do nothing on delete, since this isn't a "real" resource.
+        break;
+      default:
+        throw new Error(`Unsupported request type ${event.RequestType}`);
+    }
+
+    console.log(`Response data for ${event.RequestType}: ${JSON.stringify(responseData)}`);
+    await report(event, context, "SUCCESS", physicalResourceId, responseData);
```
efekarakus: Can we ensure that the CR will fail within the lambda timeout frame with`Promise.race`, ideally we should have it for all of our custom-resources :
https://github.com/aws/copilot-cli/blob/2c58169c98b3ed4cb238126be9f6ef2df978ea79/cf-custom-resources/lib/cert-replicator.js#L233-L234

> Set a timer in your function to respond to AWS CloudFormation with an error when a function is about to time out. A timer can help prevent delays for custom resources.
> https://aws.amazon.com/premiumsupport/knowledge-center/best-practices-custom-cf-lambda/

cf-custom-resources/lib/unique-aliases.js
```diff
@@ -0,0 +1,141 @@
...
+  } catch (err) {
+    console.error(`caught error: ${err}`);
```
efekarakus: I never get javascript, does this print a nice error message or is it `[Object object]`? I wonder if we need `JSON.stringify` 🤔

cf-custom-resources/lib/unique-aliases.js
```diff
@@ -0,0 +1,141 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
efekarakus: I wonder if we should rename the function to `unique-json-values` since it's pretty generic and not just applicable to aliases. The return key can be also `UniqueValues` 💭

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -1,10 +1,53 @@
...
+          Principal:
+            Service:
+              - lambda.amazonaws.com
+          Action:
+            - sts:AssumeRole
+    Path: /
+    ManagedPolicyArns:
+      - !Sub arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+
+UniqueAliasesFunction:
```
efekarakus: What do you think of having the logical IDs like this:
The logical IDs can be:
```
UniqueJSONValuesFunctionRole
UniqueJSONValuesFunction
UniqueAliasesAction # this one remains the same
```

internal/pkg/template/templates/environment/partials/cdn-resources.yml
efekarakus: Should there be a condition here as well?

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -1,10 +1,53 @@
...
+  Type: AWS::Lambda::Function
+  Condition: CreateALB
+  Properties:
+    {{- with $cr := index .CustomResources "UniqueAliasesFunction" }}
+    Code:
+      S3Bucket: {{$cr.Bucket}}
+      S3Key: {{$cr.Key}}
+    {{- end}}
+    Handler: "index.handler"
+    Timeout: 600
```
efekarakus, line 31: note: The `Promise.race` has to be within the 10 min window

cf-custom-resources/test/unique-aliases-test.js
dannyrandall: whoops. i just really wanted to make sure it works lol. good catch.

.gitignore
dannyrandall: bc i have a service based on `Dockerfile.site` lol

cf-custom-resources/lib/unique-aliases.js
dannyrandall: rip :sob:

cf-custom-resources/lib/unique-aliases.js
dannyrandall: it appears to work without `JSON.stringify()` 😊 (based on seeing the logs from the tests that hit this case)

cf-custom-resources/lib/unique-aliases.js
```diff
@@ -0,0 +1,138 @@
...
+      .end(responseBody, "utf8");
+  });
+};
+
+const logGroupName = function (context) {
+  return defaultLogGroup || context.logGroupName;
+};
+
+const logStreamName = function (context) {
+  return defaultLogStream || context.logStreamName;
```
Lou1415926: TIL: we can probably use [`clientContext`](https://github.com/vandium-io/lambda-tester/blob/9ebc71a61206749c467cc71658a633785f10dfa2/lib/index.js#L171) to inject the context from unit texts. If so, we should be able to remove all of these codes that are only for test purpose from the application file.

cf-custom-resources/lib/unique-aliases.js
```diff
@@ -0,0 +1,138 @@
...
+  try {
+    switch (event.RequestType) {
+      case "Create":
+      case "Update":
+        const aliasesForService = JSON.parse(event.ResourceProperties.Aliases || "{}");
+        const unique = new Set(Object.values(aliasesForService).flat());
+        responseData.UniqueAliases = Array.from(unique).sort();
+        break;
+      case "Delete":
+        // Do nothing on delete, since this isn't a "real" resource.
```
Lou1415926: I'm trying to think if under any scenarios we'd want "Delete" to have a non-empty responseData... It looks like we should be fine.

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -240,24 +240,12 @@ exports.certificateReplicateHandler = async function (event, context) {
...
- * Update parameter by adding workload to the parameter values.
- *
- * @param {string} requestType type of the request.
- * @param {string} workload name of the workload.
- * @param {string} paramValue value of the parameter.
- *
- * @returns {string} The updated parameter.
- * @returns {bool} whether the parameter is modified.
- */
-
```
iamhopaul123, line 260: :homer-disappearing:

cf-custom-resources/lib/cert-replicator.js
```diff
@@ -240,24 +240,12 @@ exports.certificateReplicateHandler = async function (event, context) {
...
       null,
-      `${err.message} (Log: ${defaultLogGroup || context.logGroupName}/${
-        defaultLogStream || context.logStreamName
+      `${err.message} (Log: ${defaultLogGroup || context.logGroupName}/${defaultLogStream || context.logStreamName
       })`
     );
   }
 };
 
-/**
```
Lou1415926, line 250: so random 🤣

cf-custom-resources/lib/unique-aliases.js
```diff
@@ -0,0 +1,151 @@
...
+ *  "svc2": ["example.com"]
+ * }
+ * 
+ * This function returns a list of unique aliases found in that list.
+ * From the previous example, UniqueAliases would be:
+ * ["svc1.com", "example.com"]
+ */
+exports.handler = async function (event, context) {
+  const responseData = {};
+  const physicalResourceId = event.PhysicalResourceId || event.LogicalResourceId;
```
Lou1415926: It seems like we never need to change `physicalResourceId`.

What do you think of just `physicalResourceId = event.LogicalResourceId`

internal/pkg/template/templates/environment/partials/cdn-resources.yml
iamhopaul123: Does it work when we don't need to specify any alias?

cf-custom-resources/lib/unique-aliases.js
dannyrandall: yeah that's _probably_ fine... the response has to have a matching `PhysicalResourceId`, so this just felt safe to make sure we always match. Probably fine to just always use `LogicalResourceId` though.

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -1,10 +1,58 @@
...
+
+UniqueAliasesAction:
+  Metadata:
+    'aws:copilot:description': 'An action to get unique custom aliases for services in your environment'
+  Condition: CreateALB
+  Type: Custom::UniqueJSONValuesFunction
+  Properties:
+    ServiceToken: !GetAtt UniqueJSONValuesFunction.Arn
+    Aliases: !Ref Aliases
+{{- end}}
```
iamhopaul123:
```suggestion
{{- end}}{{/* endif or .CDNConfig.ViewerCertificate .DelegateDNS */}}
```

internal/pkg/template/templates/environment/partials/cdn-resources.yml
dannyrandall: ok i changed the condition to only set `Aliases` if one of these is true:
- they created the app with `--domain`
- they imported certs to the cdn

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -1,10 +1,58 @@
+{{- if or .CDNConfig.ViewerCertificate .DelegateDNS}}
```
iamhopaul123: Can we rename `ViewerCertificate` to be `ImportedCertificate`? since `ViewerCertificate` is too CF specific. Plus without `.CDNConfig.ViewerCertificate` doesn't even mean we don't have `ViewerCertificate` for CF. Instead it just means we don't have any imported cert for CF/CDN.

cf-custom-resources/lib/unique-json-values.js
```diff
@@ -0,0 +1,117 @@
...
+  const responseData = {};
+  const physicalResourceId = event.PhysicalResourceId || event.LogicalResourceId;
+
+  const handler = async function () {
+    switch (event.RequestType) {
+      case "Create":
+      case "Update":
+        const aliasesForService = JSON.parse(event.ResourceProperties.Aliases || "{}");
+        const unique = new Set(Object.values(aliasesForService).flat());
+        responseData.UniqueValues = Array.from(unique).sort();
```
iamhopaul123, line 100: Nice i like the sort

iamhopaul123 (approved):

internal/pkg/template/templates/environment/partials/cdn-resources.yml
dannyrandall: i'm wondering if that could cause issues, based on https://github.com/aws/copilot-cli/pull/3881#discussion_r943692647. If they import a cert to their public lb without setting the cdn cert, we would be still be setting the CF `Aliases` field without any certificates for those aliases

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -47,7 +50,9 @@ CloudFrontDistribution:
   Type: AWS::CloudFront::Distribution
   Properties:
     DistributionConfig:
-      Aliases: !GetAtt UniqueAliasesAction.UniqueAliases
+      {{- if or .CDNConfig.ViewerCertificate .DelegateDNS}}
+      Aliases: !GetAtt UniqueAliasesAction.UniqueValues
+      {{- end}}
```
Lou1415926: Several questions come up in my mind 🤔 :
1. Is an empty value like `"[]"` (not `"AWS::NoValue"`) valid for `Aliases` ?
2. Is `Aliases` required if `ViewerCertificate` is used? (I guess it's the other way around: `ViewerCertificate` is required if `Aliases` is used)

CaptainCarpensir: 1. Yes it's valid, it just doesn't fill the resource
2. Yes, iff `ViewerCertificate` does `Aliases` need to be specified (and by definition vise versa)

internal/pkg/template/templates/environment/partials/cdn-resources.yml
iamhopaul123: > If they import a cert to their public lb without setting the cdn cert

We'll error out in this condition i think. I don't understand why renaming will cause any issue 🤔

dannyrandall: i misunderstood what you were asking 😊 i renamed it!

CaptainCarpensir (approved):
Awesome!

PR #3882, opened 2022-08-11
chore(stackset): split functions to return the opID for streaming

efekarakus:
We need to split the `stackset` methods to no longer create operation + wait operation to complete. This will allow us to start the progress tracker once an operation starts.

### DIFF
--- a/internal/pkg/aws/cloudformation/stackset/errors.go
+++ b/internal/pkg/aws/cloudformation/stackset/errors.go
@@ -4,6 +4,7 @@
 package stackset
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -13,12 +14,50 @@ import (
 // ErrStackSetOutOfDate occurs when we try to read and then update a StackSet but between reading it
 // and actually updating it, someone else either started or completed an update.
 type ErrStackSetOutOfDate struct {
-	stackSetName string
-	parentErr    error
+	name      string
+	parentErr error
 }
 
 func (e *ErrStackSetOutOfDate) Error() string {
-	return fmt.Sprintf("stack set %s update was out of date (feel free to try again): %v", e.stackSetName, e.parentErr)
+	return fmt.Sprintf("stack set %q update was out of date (feel free to try again): %v", e.name, e.parentErr)
+}
+
+// ErrStackSetNotFound occurs when a stack set with the given name does not exist.
+type ErrStackSetNotFound struct {
+	name string
+}
+
+// Error implements the error interface.
+func (e *ErrStackSetNotFound) Error() string {
+	return fmt.Sprintf("stack set %q not found", e.name)
+}
+
+func (e *ErrStackSetNotFound) isEmpty() bool {
+	return true
+}
+
+// ErrStackSetInstancesNotFound occurs when a stack set operation should be applied to instances but they don't exist.
+type ErrStackSetInstancesNotFound struct {
+	name string
+}
+
+// Error implements the error interface.
+func (e *ErrStackSetInstancesNotFound) Error() string {
+	return fmt.Sprintf("stack set %q has no instances", e.name)
+}
+
+func (e *ErrStackSetInstancesNotFound) isEmpty() bool {
+	return true
+}
+
+// IsEmptyStackSetErr returns true if the error occurred because the stack set does not exist or does not contain any instances.
+func IsEmptyStackSetErr(err error) bool {
+	type isEmpty interface {
+		isEmpty() bool
+	}
+
+	var emptyErr isEmpty
+	return errors.As(err, &emptyErr)
 }
 
 // isAlreadyExistingStackSet returns true if the underlying error is a stack already exists error.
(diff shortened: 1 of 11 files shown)


internal/pkg/aws/cloudformation/stackset/stackset.go
```diff
@@ -28,9 +28,9 @@ type api interface {
 }
 
 const (
-	opStatusSucceeded = "SUCCEEDED"
-	opStatusStopped   = "STOPPED"
-	opStatusFailed    = "FAILED"
+	opStatusSucceeded = cloudformation.StackSetOperationStatusSucceeded
+	opStatusStopped   = cloudformation.StackSetOperationStatusStopped
+	opStatusFailed    = cloudformation.StackSetOperationStatusFailed
```
Lou1415926: nit: do these var creation belong in `stackset/opstatus.go` ?

internal/pkg/aws/cloudformation/stackset/errors.go
```diff
@@ -13,12 +14,50 @@ import (
...
+func (e *ErrStackSetInstancesNotFound) Error() string {
+	return fmt.Sprintf("stack set %q has no instances", e.name)
+}
+
+func (e *ErrStackSetInstancesNotFound) isEmpty() bool {
+	return true
+}
+
+// IsEmptyStackSetErr returns true if the error occurred because the stack set does not exist or does not contain any instances.
+func IsEmptyStackSetErr(err error) bool {
```
Lou1415926: Would you consider this as a good alternative: defining `IsEmptyStackSetErr` in `cloudformation` package (which is the only place this seems to get used) and export `isEmpty`?

efekarakus: That is fine too! I also realized that i should have named it `IsEmptyErr` so that it reads `stackset.IsEmptyErr` instead of stuttering.

I like your suggestion, if we're going to have both `stackset` and `aws/cloudformation` implement the `isEmpty` interface then it makes sense to move it to `deploy/cloudformation`.

Side note: this pattern is similar to how `os.IsTimeout` works in the stdlib:  https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/os/error.go;l=114

And there are errors like `SyscallError` that implements the internal `timeout` interface: https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/os/error.go;l=61

but also errors in the sub package of `fs`: https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/io/fs/fs.go;l=255

I'll updated the code to `cloudformation.IsEmptyErr` 👍

Lou1415926 (approved):

internal/pkg/aws/cloudformation/stackset/opstatus_test.go
```diff
@@ -0,0 +1,113 @@
...
+		"false when queued": {
+			status: cloudformation.StackSetOperationStatusQueued,
+		},
+		"false when running": {
+			status: cloudformation.StackSetOperationStatusRunning,
+		},
+		"false when stopping": {
+			status: cloudformation.StackSetOperationStatusStopping,
+		},
+		"true when succeeded": {
```
huanjani:
```suggestion
		"false when succeeded": {
```

internal/pkg/aws/cloudformation/stackset/opstatus_test.go
```diff
@@ -0,0 +1,113 @@
...
+			status: cloudformation.StackSetOperationStatusSucceeded,
+		},
+		"false when stopped": {
+			status: cloudformation.StackSetOperationStatusStopped,
+			wanted: true,
+		},
+		"false when failed": {
```
huanjani:
```suggestion
		"true when failed": {
```

internal/pkg/aws/cloudformation/stackset/opstatus_test.go
```diff
@@ -0,0 +1,113 @@
...
+		"false when running": {
+			status: cloudformation.StackSetOperationStatusRunning,
+		},
+		"false when stopping": {
+			status: cloudformation.StackSetOperationStatusStopping,
+		},
+		"true when succeeded": {
+			status: cloudformation.StackSetOperationStatusSucceeded,
+		},
+		"false when stopped": {
```
huanjani:
```suggestion
		"true when stopped": {
```

internal/pkg/aws/cloudformation/stackset/stackset.go
```diff
@@ -95,18 +109,29 @@ func (ss *StackSet) UpdateAndWait(name, template string, opts ...CreateOrUpdateO
...
+	}
+
+	if len(summaries) == 0 {
+		return "", &ErrStackSetInstancesNotFound{
+			name: name,
 		}
-		return err
 	}
 
 	// We want to delete all the stack instances, so we create a set of account ids and regions.
```
huanjani: nit:
```suggestion
	// We want to delete all the stack instances, so we create a set of account IDs and regions.
```

internal/pkg/aws/cloudformation/stackset/opstatus_test.go
efekarakus: Done! thanks for catching these

huanjani (approved):

PR #3885, opened 2022-08-11
test(e2e): add import certificate

paragbhingre:


e2e/import-certs/copilot/frontend/manifest.yml
```diff
@@ -0,0 +1,17 @@
...
+image:
+  build:
+    dockerfile: src/Dockerfile
+    context: src
+  port: 80
+
+http:
+  path: '/'
+  alias: "test.copilot-e2e-tests.ecs.aws.dev"
+  hosted_zone: "Z0512687YJJFTKMRQUQR"
```
iamhopaul123, line 12: ooh this is nice!

iamhopaul123 (approved):
Awesome!

e2e/import-certs/copilot/frontend/manifest.yml
```diff
@@ -0,0 +1,17 @@
+# We're creating the manifest ahead of time as we want to test the service with a count > 1.
```
rclinard-amzn: out of curiosity: is this actually the case? since `count : 1` is still at the bottom, I'm not sure if that's modified elsewhere

rclinard-amzn (commented):
Looks good, just a small question!

e2e/import-certs/copilot/frontend/manifest.yml
paragbhingre: Ohh that I was trying before but then decided to stay with only 1 service and forgot to delete this. I will remove this comment.

rclinard-amzn (approved):
Great work! Let's see if the test works now 👍

Issue #1612, opened 2020-11-01
Ability to import docker-compose defintion and convert them as Copilot app and services


sundarnarasiman:
Docker compose is commonly used for local development and testing. We need an ability for Copilot to import Docker Compose files and convert them as Copilot native app and svc objects, in a guide way. This will be a big boost for developers.

mikelhamer:
Pretty please

efekarakus:
Hi @mikelhamer and everyone!

Would it make sense to you if we took each service defined in the Compose file and then create a separate Copilot service with `copilot svc init`? In this scenario there would be no sidecars for any of your Copilot services.  

For images such as `postgres` would you expect it to translated into an Aurora Serverless PostgreSQL database addon or instead get an ECS service with postgres?

Thanks!

mikelhamer:
> Would it make sense to you if we took each service defined in the Compose file and then create a separate Copilot service with `copilot svc init`? In this scenario there would be no sidecars for any of your Copilot services.

Yes that would be awesome. I don't even really understand what a sidecar is at this point so I don't have an issue with that (yet) lol.

> For images such as `postgres` would you expect it to translated into an Aurora Serverless PostgreSQL database addon or instead get an ECS service with postgres?

I would expect it to just make an ECS service with postgres. The option to turn it into an RDS cluster could be cool, but I'd rather see that effort go towards expanding the 'storage' functionality :P

This team is awesome!

srsridharan:
> Hi @mikelhamer and everyone!
> 
> Would it make sense to you if we took each service defined in the Compose file and then create a separate Copilot service with `copilot svc init`? In this scenario there would be no sidecars for any of your Copilot services.

Yes that would work well. 


> For images such as `postgres` would you expect it to translated into an Aurora Serverless PostgreSQL database addon or instead get an ECS service with postgres?

1.  ECS service with postgres  to keep it simple.  In addition see (2) below for alternative needs that arise. 
2. if our dev ops team has setup / maintains the data ware house (e.g. using redshift, or snowflake) would it be possible to specify/map it from the docker file?

Also could we specify the VPC/subnet into which it should be launched? 

Ideally for me the goal is to move easily from dev to deployment without needing to take up a lot of dev-ops support.

Thanks for the amazing work on this.

pharindoko:
@efekarakus  any update on this topic ?

Lou1415926:
[...]

PR #3886, opened 2022-08-11
chore: Docker Compose import milestone 1 - library setup and basic service conversion

rclinard-amzn:
Note: This is targeting the `experiment/docker-compose-import` branch. It is not targeting the mainline release branch. 
Also note that most of the added code is unit tests - there's <250 lines of actual functional code.

This PR implements the first milestone of the Docker Compose import experiment. We set up the library and handle the Compose concepts that mostly mirror 1:1 with Copilot fields.

In the next PR, I will add detection for unsupported Compose fields that this experiment won't support.

### DIFF
--- a/internal/pkg/docker/dockercompose/svc.go
+++ b/internal/pkg/docker/dockercompose/svc.go
@@ -66,18 +66,23 @@ func convertTaskConfig(service *types.ServiceConfig) (manifest.TaskConfig, error
 		}
 	}
 
+	taskCfg := manifest.TaskConfig{
+		Platform: manifest.PlatformArgsOrString{
+			PlatformString: (*manifest.PlatformString)(nilIfEmpty(service.Platform)),
+		},
+		EnvFile: envFile,
+	}
+
 	envVars, err := convertMappingWithEquals(service.Environment)
 	if err != nil {
 		return manifest.TaskConfig{}, fmt.Errorf("convert environment variables: %w", err)
 	}
 
-	return manifest.TaskConfig{
-		Platform: manifest.PlatformArgsOrString{
-			PlatformString: (*manifest.PlatformString)(nilIfEmpty(service.Platform)),
-		},
-		Variables: envVars,
-		EnvFile:   envFile,
-	}, nil
+	if len(envVars) != 0 {
+		taskCfg.Variables = envVars
+	}
+
+	return taskCfg, nil
 }
 
 // convertHealthCheckConfig trivially converts a Compose container health check into its Copilot variant.
--- a/internal/pkg/docker/dockercompose/svc.go
+++ b/internal/pkg/docker/dockercompose/svc.go
@@ -41,13 +41,6 @@ func convertBackendService(service *compose.ServiceConfig, port uint16) (*manife
 
 	if service.HealthCheck != nil {
 		svcCfg.ImageConfig.HealthCheck = convertHealthCheckConfig(service.HealthCheck)
-		for ext := range service.HealthCheck.Extensions {
-			ignored = append(ignored, "healthcheck."+ext)
-		}
-	}
-
-	for ext := range service.Extensions {
-		ignored = append(ignored, ext)
 	}
 
 	return svcCfg, ignored, nil
--- a/internal/pkg/docker/dockercompose/svc.go
+++ b/internal/pkg/docker/dockercompose/svc.go
@@ -1,3 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
 package dockercompose
 
 import (
--- a/internal/pkg/docker/dockercompose/svc.go
+++ b/internal/pkg/docker/dockercompose/svc.go
@@ -51,49 +51,6 @@ func convertBackendService(service *types.ServiceConfig, port uint16) (*manifest
 	return svcCfg, ignored, nil
 }
 
-func convertLBWS(service *types.ServiceConfig, port uint16) (*manifest.LoadBalancedWebServiceConfig, IgnoredKeys, error) {
-	image, ignored, err := convertImageConfig(service.Build, service.Labels, service.Image)
-	if err != nil {
-		return nil, ignored, fmt.Errorf("convert image config: %w", err)
-	}
-
-	taskCfg, err := convertTaskConfig(service)
-	if err != nil {
-		return nil, ignored, fmt.Errorf("convert task config: %w", err)
-	}
-
-	svcCfg := &manifest.LoadBalancedWebServiceConfig{
-		ImageConfig: manifest.ImageWithPortAndHealthcheck{
-			ImageWithPort: manifest.ImageWithPort{
-				Image: image,
-				Port:  &port,
-			},
-		},
-		ImageOverride: manifest.ImageOverride{
-			Command: manifest.CommandOverride{
-				StringSlice: service.Command,
-			},
-			EntryPoint: manifest.EntryPointOverride{
-				StringSlice: service.Entrypoint,
-			},
-		},
-		TaskConfig: taskCfg,
-	}
-
-	if service.HealthCheck != nil {
-		svcCfg.ImageConfig.HealthCheck = convertHealthCheckConfig(service.HealthCheck)
-		for ext := range service.HealthCheck.Extensions {
-			ignored = append(ignored, "healthcheck"+ext)
-		}
-	}
-
-	for ext := range service.Extensions {
-		ignored = append(ignored, ext)
-	}
-
-	return svcCfg, ignored, nil
-}
-
 // convertTaskConfig converts environment variables, env files, and platform strings.
 func convertTaskConfig(service *types.ServiceConfig) (manifest.TaskConfig, error) {
 	var envFile *string
--- a/internal/pkg/docker/dockercompose/svc.go
+++ b/internal/pkg/docker/dockercompose/svc.go
@@ -39,6 +39,13 @@ func convertBackendService(service *types.ServiceConfig, port uint16) (*manifest
 
 	if service.HealthCheck != nil {
 		svcCfg.ImageConfig.HealthCheck = convertHealthCheckConfig(service.HealthCheck)
+		for ext := range service.HealthCheck.Extensions {
+			ignored = append(ignored, "healthcheck."+ext)
+		}
+	}
+
+	for ext := range service.Extensions {
+		ignored = append(ignored, ext)
 	}
 
 	return svcCfg, ignored, nil
@@ -75,6 +82,13 @@ func convertLBWS(service *types.ServiceConfig, port uint16) (*manifest.LoadBalan
 
 	if service.HealthCheck != nil {
 		svcCfg.ImageConfig.HealthCheck = convertHealthCheckConfig(service.HealthCheck)
+		for ext := range service.HealthCheck.Extensions {
+			ignored = append(ignored, "healthcheck"+ext)
+		}
+	}
+
+	for ext := range service.Extensions {
+		ignored = append(ignored, ext)
 	}
 
 	return svcCfg, ignored, nil
--- a/internal/pkg/docker/dockercompose/svc.go
+++ b/internal/pkg/docker/dockercompose/svc.go
@@ -79,13 +79,11 @@ func convertTaskConfig(service *compose.ServiceConfig) (manifest.TaskConfig, err
 // convertHealthCheckConfig trivially converts a Compose container health check into its Copilot variant.
 func convertHealthCheckConfig(healthcheck *compose.HealthCheckConfig) manifest.ContainerHealthCheck {
 	retries := 3
-
 	if healthcheck.Retries != nil {
 		retries = int(*healthcheck.Retries)
 	}
 
 	cmd := healthcheck.Test
-
 	if healthcheck.Disable {
 		cmd = []string{"NONE"}
 	}
--- a/internal/pkg/docker/dockercompose/svc.go
+++ b/internal/pkg/docker/dockercompose/svc.go
@@ -1,0 +1,119 @@
+package dockercompose
+
+import (
+	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/copilot-cli/internal/pkg/manifest"
+	"github.com/compose-spec/compose-go/types"
+	"time"
+)
+
+func convertBackendService(service *types.ServiceConfig, port uint16) (*manifest.BackendServiceConfig, IgnoredKeys, error) {
+	image, ignored, err := convertImageConfig(service.Build, service.Labels, service.Image)
+	if err != nil {
+		return nil, ignored, fmt.Errorf("convert image config: %w", err)
+	}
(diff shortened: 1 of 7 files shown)


rclinard-amzn:
One issue that needs to be addressed after this PR - how do we handle LBWS conversion without duplicating tons of code? The LBWS manifest is almost exactly the same as the Backend Service manifest, but they use two different data structures in our codebase because the Port field is optional on Backend services. Alternatively, we can just not add the same level of unit tests for the LBWS conversion function since the logic should otherwise be the same.

internal/pkg/docker/dockercompose/svc.go
```diff
@@ -0,0 +1,104 @@
...
+package dockercompose
+
+import (
+	"fmt"
+	"github.com/aws/copilot-cli/internal/pkg/manifest"
+	"github.com/compose-spec/compose-go/types"
+	"time"
+)
+
+func convertBackendService(service *types.ServiceConfig, port uint16) (*manifest.BackendServiceConfig, IgnoredKeys, error) {
```
huanjani: Just curious how the returned `IgnoredKeys` will be handled.

internal/pkg/docker/dockercompose/svc_image.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+// convertImageConfig takes in a Compose build config & image field and returns a Copilot image manifest.
+func convertImageConfig(build *types.BuildConfig, labels map[string]string, imageLoc string) (manifest.Image, IgnoredKeys, error) {
+	image := manifest.Image{
+		DockerLabels: labels,
+	}
+
+	// note: Compose allows both build & image to be specified, but image takes precedence.
+	//       in Copilot, however, build & image are mutually exclusive.
```
huanjani: nit:
```suggestion
	// In Copilot, however, build and image are mutually exclusive.
```

CaptainCarpensir:
> One issue that needs to be addressed after this PR - how do we handle LBWS conversion without duplicating tons of code? The LBWS manifest is almost exactly the same as the Backend Service manifest, but they use two different data structures in our codebase because the Port field is optional on Backend services. Alternatively, we can just not add the same level of unit tests for the LBWS conversion function since the logic should otherwise be the same.

Unit tests are not good to skip even in these cases because they're partly needed to alert people when things break. Because they may be similar code it may be attractive to skip writing them, but they can save you lots of time debugging, and help you find bugs before they're sent to production. It's not just to test the logic of code, but also to make sure that the code is what you think it is, typos can break a surprising amount of stuff 😄

internal/pkg/docker/dockercompose/svc.go
rclinard-amzn: I intend for it to be passed through up to the point where the user is interacting with the software, so that an appropriate warning message can be displayed to them, without disrupting conversion. The fatal keys, however, generate an error instead.

rclinard-amzn:
That's a good point! I think that once the Backend Service handling is fleshed out in a PR or two, the LBWS handling can be copy pasted from that, instead of having to copy paste every time I make a logic change to Backend Service handling. That should amortize the cost of copy-pasting and avoid introducing minor deviations along the way!

huanjani:
> One issue that needs to be addressed after this PR - how do we handle LBWS conversion without duplicating tons of code? The LBWS manifest is almost exactly the same as the Backend Service manifest, but they use two different data structures in our codebase because the Port field is optional on Backend services. Alternatively, we can just not add the same level of unit tests for the LBWS conversion function since the logic should otherwise be the same.

We do plan to refactor that package to reduce the duplicated code. One day...!

internal/pkg/docker/dockercompose/svc.go
huanjani: 👍🏼 
Just a note: let's build with conversion extensability in mind; as Docker Compose or we add/change/(deprecate) fields, it should be easy enough to adapt in the conversion code.

huanjani: Does the `compose-go` library handle `port` separately? Where is port coming from? Is that from `expose`?

rclinard-amzn: There is some complex logic that will be needed to derive the proper value for the `port` that will happen in Milestone 3 - since our model of a service having a single port doesn't exactly line up with how Compose works with its port declarations. It is being passed in here since that logic will also decide whether the service becomes a Backend Service or a Load Balanced Web Service, so it needs to be called outside of `convertBackendService` anyways.

internal/pkg/docker/dockercompose/svc_image.go
```diff
@@ -0,0 +1,140 @@
...
+		if v == nil {
+			badArgs = append(badArgs, k)
+		} else {
+			args[k] = *v
+		}
+	}
+
+	if badArgs != nil {
+		return nil, fmt.Errorf("some entries are missing values and require user input, "+
+			"this is not supported in Copilot: %v", badArgs)
```
huanjani: maybe?: 
```suggestion
		return nil, fmt.Errorf("%s '%v' %s missing %s and %s user input", english.PluralWord(badArgs, "entry", "entries"), badArgs, english.PluralWord(badArgs, "is", "are"), english.PluralWord(badArgs, "a value", "values"), english.PluralWord(badArgs, "requires", "require"))
```

rclinard-amzn: Looks like a nice change! I'll keep the "not supported" part though I think.

internal/pkg/docker/dockercompose/unsupported.go
```diff
@@ -0,0 +1,79 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package dockercompose
+
+// NOTE: This file is currently unused, it will be used in the next PR. Don't mind the comments for now!
+
+type IgnoredKeys []string
```
iamhopaul123, line 11: Comment for public type pls 🙏  Also what do we use it for?

internal/pkg/docker/dockercompose/svc.go
```diff
@@ -0,0 +1,104 @@
...
+	image, ignored, err := convertImageConfig(service.Build, service.Labels, service.Image)
+	if err != nil {
+		return nil, ignored, fmt.Errorf("convert image config: %w", err)
```
iamhopaul123:
```suggestion
		return nil, nil, fmt.Errorf("convert image config: %w", err)
```
If it fails at some point we probably want to fail entirely and in that scenario we won't use `ignored` anymore.

internal/pkg/docker/dockercompose/svc_image.go
```diff
@@ -0,0 +1,144 @@
...
+			ignoredKeys = append(ignoredKeys, "build."+ext)
+		}
+	}
+
+	if build.Labels != nil {
+		ignoredKeys = append(ignoredKeys, "build.labels")
+	}
+
+	if len(ignoredKeys) == 0 {
+		return nil, nil
```
iamhopaul123: Why do we return `nil` 🤔

internal/pkg/docker/dockercompose/svc.go
dannyrandall: should this be an exported function? i'm trying to understand how a future command would use this package...

internal/pkg/docker/dockercompose/svc_image.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/dustin/go-humanize/english"
+)
+
+// convertImageConfig takes in a Compose build config & image field and returns a Copilot image manifest.
+func convertImageConfig(build *types.BuildConfig, labels map[string]string, imageLoc string) (manifest.Image, IgnoredKeys, error) {
+	image := manifest.Image{
+		DockerLabels: labels,
+	}
+
+	// note: Compose allows both build & image to be specified, but image takes precedence.
```
dannyrandall: nit
```suggestion
	// Compose allows both build & image to be specified, but image takes precedence.
```

internal/pkg/docker/dockercompose/svc_image.go
```diff
@@ -0,0 +1,144 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package dockercompose
+
+import (
+	"errors"
+	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/copilot-cli/internal/pkg/manifest"
+	"github.com/compose-spec/compose-go/types"
```
dannyrandall: while in general i avoid this...i wonder if renaming this import would help with readability💭 something like
```suggestion
	compose "github.com/compose-spec/compose-go/types"
```
and then references would look like `compose.BuildConfig` against `manifest.Image`. idk, just a thought

internal/pkg/docker/dockercompose/unsupported.go
dannyrandall: does it need to be exported...? (probably related to my other comment about trying to understand how this will all be used)

internal/pkg/docker/dockercompose/svc_image.go
```diff
@@ -0,0 +1,144 @@
...
+
+// unsupportedBuildKeys checks for unsupported keys in the given Compose build config.
+func unsupportedBuildKeys(build *types.BuildConfig) (IgnoredKeys, error) {
+	if build.SSH != nil || build.Secrets != nil {
+		return nil, errors.New("`build.ssh` and `build.secrets` are not supported yet, see " +
+			"https://github.com/aws/copilot-cli/issues/2090 for details")
+	}
+
+	if build.ExtraHosts != nil {
+		return nil, errors.New("key `build.extra_hosts` is not supported yet, this might break your app")
```
dannyrandall, line 96: :joy:

internal/pkg/docker/dockercompose/svc_image.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+	}
+
+	if build.Labels != nil {
+		ignoredKeys = append(ignoredKeys, "build.labels")
+	}
+
+	if len(ignoredKeys) == 0 {
+		return nil, nil
+	}
```
dannyrandall: i think we can probably just remove the `if` block...

internal/pkg/docker/dockercompose/svc_image.go
```diff
@@ -0,0 +1,144 @@
...
+func convertMappingWithEquals(inArgs types.MappingWithEquals) (map[string]string, error) {
+	args := map[string]string{}
+	var badArgs []string
+
+	for k, v := range inArgs {
+		if v == nil {
+			badArgs = append(badArgs, k)
+		} else {
+			args[k] = *v
+		}
```
dannyrandall: nit
```suggestion
		if v == nil {
			badArgs = append(badArgs, k)
			continue
		}
		args[k] = *v
```

dannyrandall (commented):
logic looks pretty good! just a few questions & nits!

internal/pkg/docker/dockercompose/svc.go
rclinard-amzn: This will actually be called by another function that will be exported instead! This is an internal helper for those functions, since it doesn't handle actually parsing the Compose yaml.

internal/pkg/docker/dockercompose/svc_image.go
rclinard-amzn: It made testing more convenient (since `nil` is the empty value for maps) but I found a different way to do it that allows us to remove this.

internal/pkg/docker/dockercompose/unsupported.go
rclinard-amzn: Comment added! And this does need to be exported since it will eventually be threaded through all the way to the CLI output.

internal/pkg/docker/dockercompose/svc_image.go
rclinard-amzn: Makes sense... I did this and it did improve readability, so I'll push it, thanks for the suggestion!

rclinard-amzn:
Thanks for the great feedback, that should all be addressed now!

internal/pkg/docker/dockercompose/svc_image.go
```diff
@@ -0,0 +1,143 @@
...
+		BuildArgs: buildArgs,
+	}
+
+	return image, ignored, nil
+}
+
+func nilIfEmpty(s string) *string {
+	if s == "" {
+		return nil
+	} else {
```
huanjani, line 61: nit:
```suggestion
	}
```

internal/pkg/docker/dockercompose/svc_image.go
```diff
@@ -0,0 +1,143 @@
...
+		return &s
+	}
```
huanjani: nit:
```suggestion
	}
	return &s
```

internal/pkg/docker/dockercompose/svc_image.go
```diff
@@ -0,0 +1,143 @@
...
+	for k, v := range inArgs {
+		if v == nil {
+			badArgs = append(badArgs, k)
+			continue
+		}
+		args[k] = *v
+	}
+
+	if len(badArgs) != 0 {
+		return nil, fmt.Errorf("%s '%v' %s missing %s and %s user input, this is unsupported in Copilot",
```
huanjani: What do you think about this (plus deleting "requires"/"require"? We get across the idea that this is not allowed in Copilot (even if it is in Docker Compose), but is less redundant.
```suggestion
		return nil, fmt.Errorf("%s '%v' %s missing %s; this is unsupported in Copilot",
```

huanjani (approved):
LGTM!

internal/pkg/docker/dockercompose/svc_image.go
rclinard-amzn: Sure, that makes sense - I wanted to error on the side of not being too vague, but I think that can certainly be removed. Thanks!

internal/pkg/docker/dockercompose/svc.go
```diff
@@ -0,0 +1,100 @@
...
+// convertHealthCheckConfig trivially converts a Compose container health check into its Copilot variant.
+func convertHealthCheckConfig(healthcheck *compose.HealthCheckConfig) manifest.ContainerHealthCheck {
+	retries := 3
+
+	if healthcheck.Retries != nil {
+		retries = int(*healthcheck.Retries)
+	}
+
+	cmd := healthcheck.Test
+
```
dannyrandall:
```suggestion
```

internal/pkg/docker/dockercompose/svc.go
```diff
@@ -0,0 +1,100 @@
...
+		taskCfg.Variables = envVars
+	}
+
+	return taskCfg, nil
+}
+
+// convertHealthCheckConfig trivially converts a Compose container health check into its Copilot variant.
+func convertHealthCheckConfig(healthcheck *compose.HealthCheckConfig) manifest.ContainerHealthCheck {
+	retries := 3
+
```
dannyrandall:
```suggestion
```

dannyrandall (approved):
lgtm!

PR #3891, opened 2022-08-12
chore(render): add MultiRenderer

efekarakus:


internal/pkg/term/progress/render.go
```diff
@@ -78,3 +78,44 @@ func EraseAndRender(out FileWriteFlusher, r Renderer, prevNumLines int) (int, er
...
 }
+
+// MultiRenderer returns a Renderer that's the concatenation of the input renderers.
+// The renderers are rendered sequentially, and the MultiRenderer is only Done once all renderers are Done.
+func MultiRenderer(renderers ...DynamicRenderer) DynamicRenderer {
+	mr := &multiRenderer{
+		renderers: renderers,
+		done:      make(chan struct{}),
+	}
+	go mr.listen()
```
dannyrandall, line 89: hahah go mr. listen! 🥳

dannyrandall (approved):

internal/pkg/term/progress/render.go
efekarakus: I chuckled at that too

internal/pkg/term/progress/component.go
```diff
@@ -20,7 +20,7 @@ type noopComponent struct{}
 
 // Render does not do anything.
 // It returns 0 and nil for the error.
-func (c *noopComponent) Render(out io.Writer) (numLines int, err error) {
+func (c *noopComponent) Render(_ io.Writer) (numLines int, err error) {
```
iamhopaul123, line 23: Why not just
```suggestion
func (c *noopComponent) Render(io.Writer) (numLines int, err error) {
```

iamhopaul123 (approved):

internal/pkg/term/progress/component.go
efekarakus: oh no reason, I didn't think of it 😂

Issue #3858, opened 2022-08-06
Add ability to select container for logs command


Fodoj:
At the moment, `copilot svc logs` command only streams logs from the main application container. If there are any sidecars, its now showing those logs. Feature proposal is to do something like `copilot svc logs -n web -e preprod -c $sidecarName` to show logs of $sidecarName container when needed.

paragbhingre:
@Fodoj Thank you for your request. This feature makes sense to me.

Lou1415926:
Hello @Fodoj !

From what I see now, `copilot svc logs` should show the logs from both the main container as well as the sidecars. For example
```
Found only one deployed service web in environment test
copilot/mysidecar/1111111 This is from a sidecar
copilot/mysidecar/2222222 This is from a sidecar
copilot/web/1111111 2022/08/12 20:20:16 Get Succeeded
copilot/web/2222222 2022/08/12 20:20:16 Get Succeeded
```
where `mysidecars` is the sidecar container's name, and `web` is the service.

Is this consistent with what you are seeing right now?

Fodoj:
It does indeed show logs from all containers, which can be a bit noisy if primary container logs a lot, and sidecar logs way less - and I want to see only sidecar logs to investigate some issue there :)

Lou1415926:
Sounds good! Yes just wanted to clarify 👍🏼. For example, running  `copilot svc logs -n web -e test --container mysidecar ` should show the logs from the `mysidecar` only.

efekarakus:
This feature is now released in [v1.22](https://github.com/aws/copilot-cli/releases/tag/v1.22.0) 🚀 !

For the blog post: https://aws.github.io/copilot-cli/blogs/release-v122/
Release notes: https://github.com/aws/copilot-cli/releases/tag/v1.22.0

PR #3895, opened 2022-08-12
feat: display only log streams from a specific container in `svc logs`

Lou1415926:
Addresses #3858 


internal/pkg/cli/flag.go
```diff
@@ -240,6 +241,7 @@ Defaults to all logs. Only one of end-time / follow may be used.`
 	tasksLogsFlagDescription               = "Optional. Only return logs from specific task IDs."
 	includeStateMachineLogsFlagDescription = "Optional. Include logs from the state machine executions."
 	logGroupFlagDescription                = "Optional. Only return logs from specific log group."
+	containerLogFlagDescription            = "Optional. Return only logs from a container."
```
dannyrandall:
```suggestion
	containerLogFlagDescription            = "Optional. Return only logs from a specific container."
```

internal/pkg/logging/workload.go
```diff
@@ -200,20 +205,25 @@ func (s *WorkloadClient) WriteLogEvents(opts WriteLogEventsOpts) error {
 	}
 }
 
-func (s *WorkloadClient) logStreams(taskIDs []string, includeStateMachineLogs bool) (logStreamPrefixes []string) {
+func (s *WorkloadClient) logStreams(taskIDs []string, includeStateMachineLogs bool, container string) []string {
 	// By default, we only want logs from copilot task log streams.
-	// This filters out log streams not starting with `copilot/`
-	logStreamPrefixes = []string{fmt.Sprintf(fmtWkldLogStreamPrefix, "")}
-	// includeStateMachineLogs is mutually exclusive with specific task IDs and only used for jobs. Therefore, we
+	// This filters out log streams not starting with `copilot/`, or `copilot/mysidecar` if a container is appointed.
```
dannyrandall:
```suggestion
	// This filters out log streams not starting with `copilot/`, or `copilot/mysidecar` if container is set.
```

internal/pkg/logging/workload.go
```diff
@@ -200,20 +205,25 @@ func (s *WorkloadClient) WriteLogEvents(opts WriteLogEventsOpts) error {
...
+	logStreamPrefixes := []string{fmt.Sprintf("%s/%s", wkldLogStreamPrefix, container)}
+	// includeStateMachineLogs is mutually exclusive with specific task IDs and only used for jobgs. Therefore, we
```
dannyrandall: lol

internal/pkg/logging/workload.go
```diff
@@ -200,20 +205,25 @@ func (s *WorkloadClient) WriteLogEvents(opts WriteLogEventsOpts) error {
...
 		return append(logStreamPrefixes, stateMachineLogStreamPrefix)
 	}
 	if len(taskIDs) != 0 {
 		logStreamPrefixes = []string{}
 		for _, taskID := range taskIDs {
-			logStreamPrefixes = append(logStreamPrefixes, fmt.Sprintf("%s/%s", s.logStreamNamePrefix, taskID))
+			prefix := wkldLogStreamPrefix
+			if container == "" {
+				container = s.name
+			}
```
dannyrandall: could we move this above the for loop?

internal/pkg/logging/workload.go
```diff
@@ -200,20 +205,25 @@ func (s *WorkloadClient) WriteLogEvents(opts WriteLogEventsOpts) error {
 	}
 }
 
-func (s *WorkloadClient) logStreams(taskIDs []string, includeStateMachineLogs bool) (logStreamPrefixes []string) {
+func (s *WorkloadClient) logStreams(taskIDs []string, includeStateMachineLogs bool, container string) []string {
```
dannyrandall, line 209: just a thought (i know this was mostly here before, and it'll probably be changed re: the above `TODO`)
```
switch {
case includeStateMachineLogs:
  return []string{fmt.Sprintf("%s/%s", wkldLogStreamPrefix, container), stateMachineLogStreamPrefix}
case len(taskIDs) == 0:
  return []string{fmt.Sprintf("%s/%s", wkldLogStreamPrefix, container)}
}

var logStreamPrefixes []string
if container == "" {
  container = s.name
}
for _, taskID := range taskIDs {
  // etc
}
```

dannyrandall (approved):
great work!

internal/pkg/logging/workload.go
Lou1415926: Done! Ya it's likely going to change after the refactor, but it's still nicer this way. I bet with the original way it's written I'll be very confused again after I'm back 😂  At least this way it'll take less time for me to regain my understanding 😂

internal/pkg/logging/workload.go
```diff
@@ -173,10 +177,11 @@ func (s *WorkloadClient) WriteLogEvents(opts WriteLogEventsOpts) error {
 	if opts.IncludeStateMachineLogs {
 		logStreamLimit *= 2
 	}
-	// TODO: there should be a separate logging client for ECS services
+	// TODO(lou1415926): there should be a separate logging client for ECS services
 	// and App Runner services. This `if` check is only ever true for ECS services.
+	// Refactor so that there are separate client for rdws, job and other services.
 	if s.logStreamNamePrefix != "" {
```
rclinard-amzn: Isn't `s.logStreamNamePrefix` only used as a marker value now after this PR? I think it could probably be replaced by some sort of boolean since its content (other than the fact of it being non-empty) is ignored now.

rclinard-amzn (commented):
Looks great! Just one small clarification comment.

internal/pkg/logging/workload.go
iamhopaul123: +1 maybe will be included in this pending refactoring? https://github.com/aws/copilot-cli/pull/3895/files#diff-5b02f52c7202c54f55f56d2003862894a82400445cfa061e6069ddfeb0c2f8a0R180

Lou1415926: Yes I expect that variable to be refactored out as part of the `TODO` refactor. But I changed it to bool and renamed it anyway for readability before I finish the refactor/

rclinard-amzn (approved):
Great work!

PR #3901, opened 2022-08-15
chore: remove redundant info from recommendation

iamhopaul123:
Evaluate target health is not needed because if users update the A record value on their own then CFN won't be able to delete the record anymore, since the value itself is different from when it was created by CFN.


dannyrandall (approved):

internal/pkg/cli/deploy/errors.go
```diff
@@ -32,8 +32,7 @@ func (e *errSvcWithALBAliasHostedZoneWithCDNEnabled) Error() string {
 func (e *errSvcWithALBAliasHostedZoneWithCDNEnabled) RecommendActions() string {
 	msgs := []string{
 		"If you already have a Load Balanced Web Service deployed, you can switch to CDN by:",
-		fmt.Sprintf(" 1. Update the A-record value to be the CDN distribution domain name. Please make sure your A-record has %q set to %s.",
-			"Evaluate target health", color.HighlightCode("True")),
+		" 1. Update the A-record value to be the CDN distribution domain name.",
```
huanjani: 🤔 I wonder if we should change to either "you can switch to CDN by:" ... "Updating" "Removing" "Redeploying" OR "you can switch to CDN:" ... "Update" "Remove" "Redeploy"

Do you see what I mean?

efekarakus (approved):

PR #3902, opened 2022-08-15
chore(render): add ListeningStackSetRenderer

efekarakus:


internal/pkg/aws/cloudformation/stackstatus.go
```diff
@@ -52,7 +52,8 @@ func (ss StackStatus) UpsertInProgress() bool {
 	return ss == cloudformation.StackStatusCreateInProgress || ss == cloudformation.StackStatusUpdateInProgress
 }
 
-func (ss StackStatus) Success() bool {
+// IsSuccessful returns true if the resource mutated successfully.
+func (ss StackStatus) IsSuccessful() bool {
```
iamhopaul123: Why do we use `IsSuccessful` instead of just `Successful` 🤔

internal/pkg/term/progress/status.go
```diff
@@ -7,24 +7,49 @@ import (
...
+	IsSuccessful() bool
+	IsFailure() bool
+	InProgress() bool
+	fmt.Stringer
+}
+
+// notStartedResult represents an unbegun status that implements the result interface.
+type notStartedResult struct{}
+
+func (r notStartedResult) IsSuccessful() bool {
```
iamhopaul123: nit: comment pls

internal/pkg/aws/cloudformation/stackset/status.go
```diff
@@ -28,6 +31,16 @@ func (s OpStatus) IsCompleted() bool {
 	return s.IsSuccessful() || s.IsFailure()
 }
 
+// InProgress returns true if the operation is not in a final state yet.
```
iamhopaul123: Is it true? How about not started or failed.
```suggestion
// InProgress returns true if the operation is in progress.
```
Maybe something like that

internal/pkg/term/progress/cloudformation.go
```diff
@@ -239,6 +265,46 @@ func (c *stackComponent) addResourceRenderer(ev stream.StackEvent, description s
...
+			value:  ev.Operation.Status,
+			reason: ev.Operation.Reason,
+		})
+		updateComponentTimer(&c.mu, c.statuses, c.stopWatch)
+	}
+	close(c.done)
+}
+
+// Render renders the stack set status updates to out.
+func (c *stackSetComponent) Render(out io.Writer) (numLines int, err error) {
```
iamhopaul123: Maybe put that into the comment?
```suggestion
func (c *stackSetComponent) Render(out io.Writer) (int, error) {
```

iamhopaul123 (commented):
LGTM overall just some nits 🥺

internal/pkg/aws/cloudformation/stackset/status.go
efekarakus: Updated!

internal/pkg/term/progress/cloudformation.go
efekarakus: Done!

internal/pkg/term/progress/status.go
efekarakus: Done! But recently I found out that the stdlib doesn't do this if the type isn't public for example:
https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/io/multi.go;l=17

Personally, I think it's okay if there is a comment on the type highlighting the interfaces that it implements 💭

internal/pkg/aws/cloudformation/stackstatus.go
efekarakus: No reason, I just picked one - I think both are pretty readable 🤷 renamed the method to `IsSuccess` to keep it a bit shorter

internal/pkg/term/progress/cloudformation.go
```diff
@@ -17,11 +17,16 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-// StackSubscriber is the interface to subscribe channels to a CloudFormation stack stream event.
+// StackSubscriber is the interface to subscribe to a CloudFormation stack event stream.
 type StackSubscriber interface {
 	Subscribe() <-chan stream.StackEvent
 }
 
+// StackSetSubscriber is the interface to subscribe channels to a CloudFormation stack set operation event stream.
```
rclinard-amzn: nit for consistency (alternatively, the other comment can be edited instead):
```suggestion
// StackSetSubscriber is the interface to subscribe channels to a CloudFormation stack set event stream.
```

internal/pkg/term/progress/cloudformation_test.go
```diff
@@ -111,14 +115,128 @@ world
...
+				value: stackset.OpStatus("SUCCEEDED"),
+			},
+		}, r.statuses)
+		_, hasStarted := r.stopWatch.elapsed()
+		require.True(t, hasStarted, "the stopwatch should have started")
+	})
+}
+
+func TestStackSetComponent_Render(t *testing.T) {
+	t.Run("renders a sack set operation that succeeded", func(t *testing.T) {
```
rclinard-amzn: nit:
```suggestion
	t.Run("renders a stack set operation that succeeded", func(t *testing.T) {
```

rclinard-amzn (approved):
Looks great to me! Only two tiny nits

internal/pkg/term/progress/cloudformation.go
efekarakus: Done!

internal/pkg/term/progress/cloudformation_test.go
efekarakus: Done! 😂

iamhopaul123 (approved):

iamhopaul123:
@mergifyio refresh

PR #3904, opened 2022-08-16
test: fix regression test job failure

dannyrandall:
The `query` job doesn't run ~50% of the time in the regression test suite. This is because we are creating the job with the schedule of `@every 4m`, with the assumption that it will run once on start, and every 4 minutes thereafter. However, after testing, this is not the case - it waits 4 minutes to run the first time.

In the regression test suite, the `query` job is the last one we deploy. Then, we wait a max of 4 minutes to see if the job has run. Since it waits 4 minutes to run after being deployed, it often doesn't get run before we decide it failed. 

This doesn't occur in the e2e test suite because the job is the _second_ (of 4) workloads we deploy. The deployments of the other two services after the job take long enough that the job always has time to execute.

---

### DIFF
--- a/regression/multi-svc-app/multi_svc_app_test.go
+++ b/regression/multi-svc-app/multi_svc_app_test.go
@@ -183,7 +183,7 @@ var _ = Describe("regression", func() {
 				Expect(err).NotTo(HaveOccurred())
 				_, err = cli.Run("job", "init",
 					"--name", "query",
-					"--schedule", "@every 4m",
+					"--schedule", "@every 1m",
 					"--dockerfile", fmt.Sprintf("./%s/Dockerfile", "query"))
 				Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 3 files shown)


regression/multi-svc-app/multi_svc_app_test.go
```diff
@@ -183,7 +183,7 @@ var _ = Describe("regression", func() {
 				Expect(err).NotTo(HaveOccurred())
 				_, err = cli.Run("job", "init",
 					"--name", "query",
-					"--schedule", "@every 4m",
+					"--schedule", "@every 1m",
```
efekarakus, line 186: Nice find! I thought we were calling `job run` in the test :doge:

efekarakus (approved):

iamhopaul123 (approved):
This is awesome! Great find!

PR #3917, opened 2022-08-17
chore: Docker Compose import, milestone 2

rclinard-amzn:
This implements Milestone 2 of the Docker Compose import experiment, allowing us to detect unsupported keys in Docker Compose files and report them to the user. We make a distinction between ignored keys, which do not have a substantial effect on the function of the app, and fatally-unsupported keys, which will very likely break critical parts of the app's function.


internal/pkg/docker/dockercompose/decompose.go
```diff
@@ -0,0 +1,113 @@
...
+// DecomposeService parses a Compose YAML file and then converts a single service to a Copilot manifest.
+func DecomposeService(content []byte, svcName string, workingDir string) (*manifest.BackendServiceConfig, IgnoredKeys, error) {
+	config, err := loader.ParseYAML(content)
+	if err != nil {
+		return nil, nil, fmt.Errorf("parse compose yaml: %w", err)
+	}
+
+	service, err := serviceConfig(config, svcName)
+	if err != nil {
+		return nil, nil, fmt.Errorf("get service config: %w", err)
```
iamhopaul123: we usually don't need to wrap errors from functions within the same package

internal/pkg/docker/dockercompose/decompose.go
```diff
@@ -0,0 +1,113 @@
...
+		},
+	})
+	if err != nil {
+		return nil, nil, fmt.Errorf("load Compose project: %w", err)
+	}
+
+	svcConfig, err := project.GetService(svcName)
+	if err != nil {
+		// I don't think this is possible due to the previous check,
+		// but I'm not confident enough in that assessment to make this a panic.
```
iamhopaul123: We could ditch this possibly. This seems more like something for reviewers to see.

internal/pkg/docker/dockercompose/decompose.go
```diff
@@ -0,0 +1,113 @@
...
+	ignored = append(ignored, svcIgnored...)
+	sort.Strings(ignored)
+
+	return backendSvc, ignored, nil
+}
+
+// serviceConfig extracts the map corresponding to a Compose service from the parsed Compose config map.
+func serviceConfig(config map[string]interface{}, svcName string) (map[string]interface{}, error) {
+	svcs, ok := config["services"]
+	if !ok || svcs == nil {
```
iamhopaul123, line 68: If svcs is a map then i would recommend
```suggestion
	if !ok || len(svcs) == 0 {
```

internal/pkg/docker/dockercompose/decompose.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	"fmt"
+	"github.com/aws/copilot-cli/internal/pkg/manifest"
+	"github.com/compose-spec/compose-go/loader"
+	compose "github.com/compose-spec/compose-go/types"
+	"sort"
+)
+
+// DecomposeService parses a Compose YAML file and then converts a single service to a Copilot manifest.
+func DecomposeService(content []byte, svcName string, workingDir string) (*manifest.BackendServiceConfig, IgnoredKeys, error) {
```
iamhopaul123: Why is this `*manifest.BackendServiceConfig`? How about load balanced web service type?

iamhopaul123 (commented):
Very nice unit tests!

internal/pkg/docker/dockercompose/decompose.go
rclinard-amzn: This is mostly just a temporary thing - since the code for Backend Service and Load Balanced Web Service is *almost* exactly the same (and I intend to use both), I'm just handling Backend Service now just to avoid lots of manual error-prone copy pasting of any updates to the code dealing with backend service / LBWS.

Though, this is a good point - since this is likely going to be the public-facing method of the conversion, it should be able to return an LBWS as well no matter what.

I'm still thinking about what the best way to return one or the other from this function, though, and the signature of this function could certainly be updated in this PR. Maybe this signature would be the best? Or is the number of return values starting to get excessive here?

```golang
func DecomposeService(content []byte, svcName string, workingDir string) (*manifest.BackendServiceConfig, *manifest.LoadBalancedWebServiceConfig, IgnoredKeys, error) {
```

internal/pkg/docker/dockercompose/decompose.go
rclinard-amzn: Unfortunately, I have to revert this, since `svcs` is of type `interface{}` instead of being a map.

internal/pkg/docker/dockercompose/decompose.go
rclinard-amzn: I looked more closely and figured out that this is indeed impossible 👍

internal/pkg/docker/dockercompose/decompose.go
iamhopaul123: gotcha that makes sense.

internal/pkg/docker/dockercompose/decompose.go
rclinard-amzn: Makes sense, this should be addressed now! There is still one remaining case of in-package error wrapping in the `dockercompose` package but it's at the point where one large function calls another large function so I think it is useful in that case.

rclinard-amzn:
Thanks for the feedback! Aside from the question, it should be addressed now with these latest changes 👍

internal/pkg/docker/dockercompose/decompose.go
iamhopaul123: Maybe we can use go generic to avoid duplicated code blocks. One example would be how we deal with [`AdvancedScalingConfig`](https://github.com/aws/copilot-cli/blob/7980a515b98156dfbce4c3070edc101061b00a0e/internal/pkg/manifest/svc.go#L206).

coderbot16: I didn't even think of generics! I'll take a look 👀

rclinard-amzn: Unfortunately, it seems like this won't allow us to eliminate the main classes of boilerplate because we can't access the shared struct fields: https://stackoverflow.com/a/71531151

iamhopaul123: Actually why do we need to return a concrete type like `*manifest.BackendServiceConfig`? If we'll output a manifest anyway can we just return an interface with a write function that output locally? Something like `encoding.BinaryMarshaler`?

rclinard-amzn: That's a fair point! I'd have to investigate how to handle the testing for that. My main concern would be that it might go against the general principle of returning as concrete of a type as possible.

huanjani: Hm, yeah...especially since we don't yet know how/where this will get used, a write function makes more sense than using one of our concrete types.

rclinard-amzn:
I've decided to approach the remaining comment by having multiple return types for backend service & load balanced web service. Ideally, we will return the marshalled manifest content as a `[]byte`, but unfortunately our manifest writing code does not currently work for more advanced manifests and is only intended to work with the initial generated manifests. I will work on a PR to fix up our manifest marshalling to ensure that it passes a "roundtrip" test and then update the code here for that, but for now to keep the Docker Compose PR series moving along I implemented this current temporary solution. Hopefully that makes sense!

huanjani (approved):
Looks good; this temporary solution works to keep the project moving!

internal/pkg/docker/dockercompose/decompose.go
```diff
@@ -0,0 +1,123 @@
...
+import (
+	"fmt"
+	"github.com/aws/copilot-cli/internal/pkg/manifest"
+	"github.com/compose-spec/compose-go/loader"
+	compose "github.com/compose-spec/compose-go/types"
+	"sort"
+)
+
+// DecomposeService parses a Compose YAML file and then converts a single service to a Copilot manifest.
+func DecomposeService(content []byte, svcName string, workingDir string) (*manifest.LoadBalancedWebService, *manifest.BackendService, IgnoredKeys, error) {
```
dannyrandall: 4 return values is a lot - maybe combine the two manifest types into a struct to return instead?

internal/pkg/docker/dockercompose/decompose.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil, nil, nil, fmt.Errorf("convert Compose service to Copilot manifest: %w", err)
+	}
+
+	ignored = append(ignored, svcIgnored...)
+	sort.Strings(ignored)
+
+	return lbws, bs, ignored, nil
+}
+
+func getServices(config map[string]interface{}) (map[string]map[string]interface{}, error) {
```
dannyrandall: can use `map[string]any` instead!

internal/pkg/docker/dockercompose/decompose.go
```diff
@@ -0,0 +1,123 @@
...
+	if !ok || svcs == nil {
+		return nil, fmt.Errorf("compose file has no services")
+	}
+
+	services, ok := svcs.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("\"services\" top-level element was not a map, was: %v", svcs)
+	}
+
+	typedServices := make(map[string]map[string]interface{})
```
dannyrandall: i wonder if this would make sense to be it's own type? something like
```
type services map[string]map[string]any
```

internal/pkg/docker/dockercompose/svc.go
```diff
@@ -5,45 +5,85 @@ package dockercompose
...
+					Image: image,
+					Port:  port,
+				},
+				HealthCheck: hc,
+			},
+			ImageOverride: imgOverride,
+			TaskConfig:    taskCfg,
+		}
+		return nil, &bs, ignored, nil
+	}
```
dannyrandall: this doesn't need to be in an else block

internal/pkg/docker/dockercompose/unsupported.go
```diff
@@ -3,17 +3,15 @@
...
 // IgnoredKeys stores a list of keys in the Compose YAML that couldn't be processed,
 // but are likely to not be significant enough to cause the converted application to
 // fail. It's expected that this will eventually be displayed to the user.
 type IgnoredKeys []string
 
-/*
 // ignoredServiceKeys lists out the keys on Compose services that are ignored in conversion.
 //
 // note: build keys are handled separately in convertBuildConfig
+// note: using a map as a set here
```
dannyrandall: this is a fairly common use case, probably don't need a comment for it
```suggestion
```

i like the way `map[string]bool` reads more - ie `if ignoredServiceKey["key"]` vs `if _, ok := ignoredServiceKey["key"]; ok `, but whichever you prefer is fine

dannyrandall (commented):
love the test cases!

rclinard-amzn:
Thanks for the comments, that should all be fixed now! 👍

dannyrandall (approved):
lgtm!

PR #3918, opened 2022-08-17
feat: add --previous-stopped-task flag to svc logs

### LINKED ISSUE #3272: Proposal: `logs -p` or `logs --open`
Currently, if I want to check for the logs of exited tasks, I need to go to AWS console, navigate to ECS, find the cluster, then the service, than switch to stopped tasks, then search for the logs there.

It would be helpful to have a `copilot svg logs -p` where `-p` stands for "previous task" (like k8s), that dumps the logs of the most recent stopped task.

Besides that, an `copilot svc logs --open` that open the cloudwatch logs console with filter applied would be helpful too, just some extra steps (like login in).

paragbhingre:
Fixes #3272 


internal/pkg/cli/flag.go
```diff
@@ -29,23 +29,24 @@ const (
...
+	dockerFileFlag          = "dockerfile"
+	dockerFileContextFlag   = "build-context"
+	imageTagFlag            = "tag"
+	resourceTagsFlag        = "resource-tags"
+	stackOutputDirFlag      = "output-dir"
+	uploadAssetsFlag        = "upload-assets"
+	limitFlag               = "limit"
+	lastFlag                = "last"
+	followFlag              = "follow"
+	previousStoppedTaskFlag = "previous-stopped-task"
```
efekarakus: can we rename the flag to just `-p, --previous` (this way we can reuse it maybe across commands)

internal/pkg/cli/flag.go
```diff
@@ -230,8 +231,9 @@ Uploaded asset locations are filled in the template configuration.`
 unless any time filtering flags are set.`
 	lastFlagDescription = `Optional. The number of executions of the scheduled job for which
 logs should be shown.`
-	followFlagDescription = "Optional. Specifies if the logs should be streamed."
-	sinceFlagDescription  = `Optional. Only return logs newer than a relative duration like 5s, 2m, or 3h.
+	followFlagDescription              = "Optional. Specifies if the logs should be streamed."
+	previousStoppedTaskFlagDescription = "Optional. Only return logs for the recently stopped task."
```
efekarakus: nit:
```suggestion
	previousStoppedTaskFlagDescription = "Optional. Print logs for the last stopped task if one exists."
```

internal/pkg/cli/svc_logs.go
```diff
@@ -84,6 +90,10 @@ func newSvcLogOpts(vars wkldLogsVars) (*svcLogsOpts, error) {
 			configStore: configStore,
 			deployStore: deployStore,
 			sel:         selector.NewDeploySelect(prompt.New(), configStore, deployStore),
+			newSvcDescriber: func(s *session.Session) serviceDescriber {
+				return ecs.New(s)
+			},
+			sessProvider: sessProvider,
```
efekarakus: To make the unit tests not depend on the `~/.aws/config` file and the `sessProvider`, we can slightly alter this:
```go
opts.newSvcDescriber = func(envCfg *config.Environment) serviceDescriber {
	sess, err := sessProvider.FromRole(envCfg.ManagerRoleARN, envCfg.Region)
	if err != nil {
		return fmt.Errorf("get session %s: %w", o.envName, err)
	}
	return ecs.New(sess)
}
```

internal/pkg/cli/svc_logs.go
```diff
@@ -193,6 +209,31 @@ func (o *svcLogsOpts) Execute() error {
...
+	svcDesc, err := o.newSvcDescriber(sess).DescribeService(o.appName, o.envName, o.name)
+	if err != nil {
+		return fmt.Errorf("describe serivce %s: %w", o.name, err)
+	}
+	if len(svcDesc.StoppedTasks) > 0 {
+		taskID, err := awsecs.TaskID(aws.StringValue(svcDesc.StoppedTasks[0].TaskArn))
+		if err != nil {
+			return err
+		}
+		o.taskIDs = append(o.taskIDs, taskID)
```
efekarakus: What do you think isntead of appending we just set to the single value?
```suggestion
		o.taskIDs = []string{ taskID }
```

efekarakus: What do you think of returning an error in `Validate` if the user specificies both `--tasks` and `--previous`?

internal/pkg/cli/svc_logs.go
```diff
@@ -193,6 +209,31 @@ func (o *svcLogsOpts) Execute() error {
...
+	} else {
+		return fmt.Errorf("previously stopped task for the service %s: no logs available", o.name)
```
efekarakus: I wonder if we shouldn't error and do a `log.Warningln("No previously stopped tasks found")` and print nothing to `stdout` 🤔

efekarakus (commented):
Looks really good! I liked how we are leveraging the existing `logging.WriteLogEventsOpts.TaskIDs` field for this feature 👍

internal/pkg/cli/flag.go
paragbhingre: I was just wondering when we use shortcut flags like -p? 
Looks like we use them only when they are mandatory by the command and this flag is not mandatory anywhere that is why I did not create shortcut. Is my understanding correct?

efekarakus: Your understanding is correct! More specifically:
> Required flags must have a short name, only provide a short name to an optional flag **if you think it will be frequently used.**
> https://github.com/aws/copilot-cli/blob/mainline/STYLE_GUIDE.md#flags

I feel like this is a really neat flag 🌈  that would be used frequently to troubleshoot so `-p` is warranted

internal/pkg/cli/svc_logs.go
paragbhingre: Yeah that sounds good to me.

internal/pkg/cli/svc_logs.go
paragbhingre: I also have an alternative. So what we can do is set `newSvcDescriber` shown in the below snipped, inside [this](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/cli/svc_logs.go#L89-L116) block on L102 so that we can reutilize the sess we created for `NewWorkloadClient`. 

```
opts.newSvcDescriber = func() serviceDescriber {
			return ecs.New(sess)
		}
```

What do you think about that?

internal/pkg/cli/flag.go
paragbhingre: Got it.

internal/pkg/cli/svc_logs.go
```diff
@@ -99,6 +104,9 @@ func newSvcLogOpts(vars wkldLogsVars) (*svcLogsOpts, error) {
 		if err != nil {
 			return err
 		}
+		opts.newSvcDescriber = func() serviceDescriber {
+			return ecs.New(sess)
+		}
```
efekarakus: oh! instead of this can we rename `initLogsSvc` to `initRuntimeClients` and do this inside:
```go
opts.ecs = ecs.New(sess)
```

internal/pkg/cli/svc_logs.go
```diff
@@ -155,6 +164,13 @@ func (o *svcLogsOpts) Validate() error {
 		return fmt.Errorf("--limit %d is out-of-bounds, value must be between %d and %d", o.limit, cwGetLogEventsLimitMin, cwGetLogEventsLimitMax)
 	}
 
+	if o.previous {
+		err := o.validatePrevious()
+		if err != nil {
```
efekarakus: nit:
```suggestion
		if err := o.validatePrevious(); err != nil {
```

internal/pkg/cli/svc_logs.go
```diff
@@ -193,6 +217,23 @@ func (o *svcLogsOpts) Execute() error {
 	return nil
 }
 
+func (o *svcLogsOpts) setPreviousStoppedTaskID() (string, error) {
+	svcDesc, err := o.newSvcDescriber().DescribeService(o.appName, o.envName, o.name)
+	if err != nil {
+		return "", fmt.Errorf("describe serivce %s: %w", o.name, err)
```
efekarakus:
```suggestion
		return "", fmt.Errorf("describe service %s: %w", o.name, err)
```

internal/pkg/cli/svc_logs.go
```diff
@@ -193,6 +217,23 @@ func (o *svcLogsOpts) Execute() error {
...
+	}
+	if len(svcDesc.StoppedTasks) > 0 {
+		taskID, err := awsecs.TaskID(aws.StringValue(svcDesc.StoppedTasks[0].TaskArn))
```
efekarakus, line 234: Did we validate if the tasks returned are sorted by timestamp? if so can we add a comment here highlighting why it's okay to grab the first item

internal/pkg/cli/svc_logs.go
```diff
@@ -179,6 +195,14 @@ func (o *svcLogsOpts) Execute() error {
 	if o.limit != 0 {
 		limit = aws.Int64(int64(o.limit))
 	}
+	if o.previous {
+		taskID, err := o.setPreviousStoppedTaskID()
+		if taskID != "" {
+			o.taskIDs = []string{taskID}
```
efekarakus: It's odd that `setPreviousStoppedTaskID` doesn't set but returns the last stopped task ID, maybe:
```suggestion
		taskID, err := o.latestStoppedTaskID()
		if taskID != "" {
			o.taskIDs = []string{taskID}
```

internal/pkg/cli/svc_logs.go
```diff
@@ -179,6 +195,14 @@ func (o *svcLogsOpts) Execute() error {
...
+		} else {
+			return err
+		}
+	}
 	err := o.logsSvc.WriteLogEvents(logging.WriteLogEventsOpts{
```
efekarakus, line 211: because `taskIDs` is `nil`, is it going to print all _running_ task logs instead of not printing?

paragbhingre: if `taskIDs` are `nil` then we return with the error in the else block so we exit with the warning.

internal/pkg/cli/svc_logs.go
paragbhingre: I could do that but this `initLogsSvc` is being used in job_logs as well and there it is just setting `opts.logsSvc`. Is it okay to still change the name to `initRuntimeClients`?

internal/pkg/cli/svc_logs.go
paragbhingre: I read through AWS SDK documentation, and nowhere they mentioned that they sort the list of tasks by the stopped time but for me I was getting it in sorted manner consistently but after you mentioned this I stopped almost 10 tasks and then checked and sometimes it is not sorted so for safer side I have added sorting logic based on stopped time. Now it will be sorted no matter how we get it from `listTasks` api.

internal/pkg/cli/flag.go
```diff
@@ -230,8 +232,9 @@ Uploaded asset locations are filled in the template configuration.`
 unless any time filtering flags are set.`
 	lastFlagDescription = `Optional. The number of executions of the scheduled job for which
 logs should be shown.`
-	followFlagDescription = "Optional. Specifies if the logs should be streamed."
-	sinceFlagDescription  = `Optional. Only return logs newer than a relative duration like 5s, 2m, or 3h.
+	followFlagDescription   = "Optional. Specifies if the logs should be streamed."
+	previousFlagDescription = "Optional. Print logs for the last stopped task if one exists."
```
iamhopaul123:
```suggestion
	previousFlagDescription = "Optional. Print logs for the last stopped task if exists."
```

internal/pkg/cli/flag.go
```diff
@@ -135,6 +136,7 @@ const (
 	envFlagShort      = "e"
 	typeFlagShort     = "t"
 	workloadFlagShort = "w"
+	previousFlagShort = "p"
```
iamhopaul123, line 139: We only give short hand flags for required fields.
```suggestion
```

internal/pkg/cli/svc_logs.go
```diff
@@ -229,6 +273,13 @@ func (o *svcLogsOpts) validateAndAskSvcEnvName() error {
 	return nil
 }
 
+func (o *svcLogsOpts) validatePrevious() error {
+	if o.previous && len(o.taskIDs) != 0 {
+		return fmt.Errorf("validate service logs: cannot specify both --%s and --%s", previousFlag, tasksFlag)
```
iamhopaul123:
```suggestion
		return fmt.Errorf("cannot specify both --%s and --%s", previousFlag, tasksFlag)
```

internal/pkg/cli/svc_logs.go
```diff
@@ -179,6 +195,14 @@ func (o *svcLogsOpts) Execute() error {
 	if o.limit != 0 {
 		limit = aws.Int64(int64(o.limit))
 	}
+	if o.previous {
+		taskID, err := o.latestStoppedTaskID()
+		if taskID != "" {
+			o.taskIDs = []string{taskID}
+		} else {
+			return err
+		}
```
iamhopaul123: I feel we should log here.
```suggestion
		taskID, err := o.latestStoppedTaskID()
		if err != nil {
			return err
		}
		if taskID == "" {
			log warning
			// return early because we don't need to print any logs.
			return nil 
		}
		o.taskIDs = []string{taskID}
```

internal/pkg/cli/svc_logs.go
```diff
@@ -193,6 +217,26 @@ func (o *svcLogsOpts) Execute() error {
...
 }
 
+func (o *svcLogsOpts) latestStoppedTaskID() (string, error) {
+	svcDesc, err := o.newSvcDescriber().DescribeService(o.appName, o.envName, o.name)
+	if err != nil {
+		return "", fmt.Errorf("describe service %s: %w", o.name, err)
+	}
+	if len(svcDesc.StoppedTasks) > 0 {
+		sort.Slice(svcDesc.StoppedTasks, func(i, j int) bool {
+			return svcDesc.StoppedTasks[i].StoppingAt.After(aws.TimeValue(svcDesc.StoppedTasks[j].StoppingAt))
```
iamhopaul123, line 232: nice!

internal/pkg/cli/svc_logs.go
```diff
@@ -193,6 +217,26 @@ func (o *svcLogsOpts) Execute() error {
...
+		})
+		taskID, err := awsecs.TaskID(aws.StringValue(svcDesc.StoppedTasks[0].TaskArn))
+		if err != nil {
+			return "", err
+		}
+		return taskID, nil
+	} else {
```
iamhopaul123: We don't need else because the previous `if` block always returns.

internal/pkg/cli/flag.go
paragbhingre: I had the same understanding, but @efekarakus helped me understand that we need short flags for frequently used functionalities as well. Referring our conversation [here](https://github.com/aws/copilot-cli/pull/3918#discussion_r948471582)

internal/pkg/cli/svc_logs.go
paragbhingre: I have done something similar in the latest commit that you have mentioned here, let me know if that works? Else happy to change as you've mentioned.

efekarakus (approved):
boom 🚀

iamhopaul123 (approved):

PR #3921, opened 2022-08-18
chore: enable progress tracker for all stack set operations

efekarakus:

### DIFF
--- a/internal/pkg/deploy/cloudformation/app.go
+++ b/internal/pkg/deploy/cloudformation/app.go
@@ -399,7 +399,7 @@ func (cf CloudFormation) AddEnvToApp(opts *AddEnvToAppOpts) error {
 		return fmt.Errorf("adding %s environment resources to application: %w", opts.EnvName, err)
 	}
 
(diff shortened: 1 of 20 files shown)


internal/pkg/deploy/cloudformation/app.go
```diff
@@ -279,7 +284,7 @@ func (cf CloudFormation) addWorkloadToApp(app *config.Application, wlName string
 		Accounts: previouslyDeployedConfig.Accounts,
 		App:      appConfig.Name,
 	}
-	if err := cf.deployAppConfig(appConfig, &newDeploymentConfig); err != nil {
+	if err := cf.deployAppConfig(appConfig, &newDeploymentConfig, shouldAddNewWl); err != nil {
```
dannyrandall, line 287: isn't this always true?

(i'm not sure i'd even want switch it to just `true`, but just )

dannyrandall: yeah i like it the way you have it 😊

dannyrandall (approved):
lgtm!

iamhopaul123 (approved):
:shipit:

PR #3926, opened 2022-08-19
test: codecov

dannyrandall:
check it out: https://app.codecov.io/gh/dannyrandall/copilot-cli


iamhopaul123 (approved):
This is cool!

.gitignore
```diff
@@ -8,4 +8,5 @@ node_modules/
 site/node_modules/
 site/themes/docsy
 internal/pkg/template/templates/custom-resources/
-regression/*/copilot/
\ No newline at end of file
+regression/*/copilot/
```
efekarakus, line 11: nit: do we need to add coverage.out here as well? 🤷 just asking because you have one for the custom resources

README.md
```diff
@@ -1,7 +1,9 @@
 ##  <img align="left" alt="AWS Copilot CLI" src="./site/content/assets/images/copilot-logo-48-light.svg" width="85" /> AWS Copilot CLI
 ###### _Build, Release and Operate Containerized Applications on AWS._ 
 
-![latest version](https://img.shields.io/github/v/release/aws/copilot-cli) [![Join the chat at https://gitter.im/aws/copilot-cli](https://badges.gitter.im/aws/copilot-cli.svg)](https://gitter.im/aws/copilot-cli?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+![latest version](https://img.shields.io/github/v/release/aws/copilot-cli)
+[![Join the chat at https://gitter.im/aws/copilot-cli](https://badges.gitter.im/aws/copilot-cli.svg)](https://gitter.im/aws/copilot-cli?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+[![codecov](https://codecov.io/gh/aws/copilot-cli/branch/mainline/graph/badge.svg)](https://codecov.io/gh/aws/copilot-cli)
```
efekarakus: :doge: I can't tell if this a good sell or not 😂 

Is there a way of modifying the `test-cov` job and setting a treshold such that we get a slack message when we are below a certain percentage? I don't want to block merging PRs but it would be good to know if we are below our current threshold in the future by getting a notification

.gitignore
dannyrandall: it's already there :blush:

README.md
rclinard-amzn: I agree with the second part but if it's too much work to add that to this PR then we can probably put it in a follow-up PR! Up to @dannyrandall though 😄

dannyrandall: ok so here's what i'm thinking! we'll upload reports as part of `ci/test` and on commits to mainline. then, when it runs on PRs, it'll comment on the PR with the changes like [here](https://github.com/aws/copilot-cli/pull/3926#issuecomment-1227716386).

rclinard-amzn: Sounds good! Will this comment be made every time there's commits? That could become spammy.

dannyrandall: i _think_ what it does is just edit the first comment it makes if it needs to update anything

huanjani (approved):
Bring on the finger-wagging!

PR #3933, opened 2022-08-23
chore: Docker Compose Import milestone 3 - port detection

rclinard-amzn:
This implements the third milestone of the Docker Compose import experiment, adding support for parsing the `expose` and `ports` keys from Compose files as well as reusing our `EXPOSE` detection from any provided `Dockerfile`s. This is the first revision of the code that is actually useful at all for converting Compose services to Copilot!

### DIFF
--- a/internal/pkg/docker/dockercompose/decompose_test.go
+++ b/internal/pkg/docker/dockercompose/decompose_test.go
@@ -262,6 +262,17 @@ func TestDecomposeService_ExposedPorts(t *testing.T) {
 		},
 		TaskConfig: taskCfg,
 	}
+	nginxLbws8096 := manifest.LoadBalancedWebServiceConfig{
+		ImageConfig: manifest.ImageWithPortAndHealthcheck{
+			ImageWithPort: manifest.ImageWithPort{
+				Image: manifest.Image{
+					Location: aws.String("nginx"),
+				},
+				Port: aws.Uint16(8096),
+			},
+		},
+		TaskConfig: taskCfg,
+	}
 
 	testCases := []decomposeTest{
 		{
@@ -291,18 +302,8 @@ func TestDecomposeService_ExposedPorts(t *testing.T) {
 			},
 		},
 		{
-			svcName: "expose-and-ports",
-			wantLbws: &manifest.LoadBalancedWebServiceConfig{
-				ImageConfig: manifest.ImageWithPortAndHealthcheck{
-					ImageWithPort: manifest.ImageWithPort{
-						Image: manifest.Image{
-							Location: aws.String("nginx"),
-						},
-						Port: aws.Uint16(8096),
-					},
-				},
-				TaskConfig: taskCfg,
-			},
+			svcName:  "expose-and-ports",
+			wantLbws: &nginxLbws8096,
 		},
 		{
 			svcName:   "remap-ports",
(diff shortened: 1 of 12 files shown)


internal/pkg/docker/dockercompose/decompose_test.go
```diff
@@ -88,31 +122,31 @@ func TestDecomposeService(t *testing.T) {
 			filename: "nginx-golang-postgres.yml",
 			svcName:  "db",
 
-			wantError: errors.New("\"services.db\" relies on fatally-unsupported Compose keys: [expose secrets volumes]"),
+			wantError: errors.New("\"services.db\" relies on fatally-unsupported Compose keys: [secrets volumes]"),
```
huanjani: nit: maybe we can separate the keys with a strings.Split(keys, ", ")?

internal/pkg/docker/dockercompose/svc.go
```diff
@@ -96,10 +101,68 @@ type exposedPort struct {
...
+			return nil, fmt.Errorf("could not parse exposed port: %w", err)
+		}
+
+		return &exposedPort{
+			port:   uint16(port),
+			public: false,
+		}, nil
+	} else if len(service.Expose) > 1 {
+		return nil, fmt.Errorf("cannot expose more than one port in Copilot, but %v ports are exposed: %v", len(service.Expose), service.Expose)
+	}
```
dannyrandall:
```suggestion
	switch {
	case len(service.Ports) > 1:
		// blah
	case len(service.Ports) == 1:
	case len(service.Expose) > 1:
	case len(service.Expose) == 1:
	}
```

internal/pkg/docker/dockercompose/testdata/buildtest/Dockerfile-unparseable
```diff
@@ -0,0 +1 @@
+EXPOSE "FACT: According to all known laws of aviation, there is no way that a bee should be able to fly."
```
dannyrandall, line 1: lol

dannyrandall (approved):
just a small nit!

internal/pkg/docker/dockercompose/svc.go
```diff
@@ -96,10 +101,68 @@ type exposedPort struct {
...
+		binding := service.Ports[0]
+		port := uint16(binding.Target)
+
+		if binding.Published != "" && strconv.Itoa(int(port)) != binding.Published {
+			return nil, fmt.Errorf("cannot publish the container port %v under a different public port %v in Copilot", port, binding.Published)
+		}
+
+		return &exposedPort{
+			port:   port,
+			public: true,
```
huanjani: Just wondering if we need to check if `binding.Published` isn't empty before knowing this is public...?

internal/pkg/docker/dockercompose/svc.go
```diff
@@ -96,10 +101,68 @@ type exposedPort struct {
...
+	if service.Build.Context == "" {
+		return nil, errors.New("service is missing an image location or Dockerfile path")
+	}
+
+	dockerfilePath := service.Build.Dockerfile
+	if dockerfilePath == "" {
+		dockerfilePath = "Dockerfile"
+	}
+
+	dockerfilePath = filepath.Join(workingDir, service.Build.Context, dockerfilePath)
```
huanjani, line 149: This is yet another grey area resulting from not knowing what form/location this tool will take, but does this rely on the user running the conversion in the same place as they'll eventually launch their Copilot services?

internal/pkg/docker/dockercompose/svc.go
rclinard-amzn: makes sense, good idea! This will be my first switch statement in golang 😄

dannyrandall: i think they are prettier than `if`/`else if` ☺️ but it's just an opinion

internal/pkg/docker/dockercompose/svc.go
rclinard-amzn: So this is actually a weird edge case in the spec! Here's what [the spec](https://docs.docker.com/compose/compose-file/#ports) says about it... so I think it should be fair to make it the same as the container port here, but I do need to add this check! Good catch 👍 

> Either specify both ports (HOST:CONTAINER), or just the container port. In the latter case, the Compose implementation SHOULD automatically allocate any unassigned host port.

rclinard-amzn:
Thanks for the feedback, I applied the style changes and also improved the robustness of exposed port handling!

huanjani (approved):
LGTM!

PR #3935, opened 2022-08-23
chore: Docker Compose import experiment - work around automatic handling of env files in compose-go

rclinard-amzn:
The `compose-go` library will automatically load any `env_file`s and merge their content into the `environment` keys, which is equivalent to our `variables` section in Copilot manifests. What this means is that we can no longer differentiate environment variables passed individually as overrides from environment variables sourced from env files. This feature is useful to simplify a Compose implementation that just deploys the application straight away, but this breaks our importing use case.

While ideally this can be addressed in a PR to the compose-go library perhaps to add an option to disable this behavior, for now this works and allows us to keep on iterating on this experiment.

### DIFF
--- a/internal/pkg/docker/dockercompose/decompose.go
+++ b/internal/pkg/docker/dockercompose/decompose.go
@@ -145,13 +145,13 @@ func isolateEnvFiles(service map[string]interface{}) ([]string, error) {
 		return []string{envFileStr.EnvFile}, nil
 	}
 
-	envFileOnly := struct {
+	envFileList := struct {
 		EnvFile []string `mapstructure:"env_file"`
 	}{}
-	err = mapstructure.Decode(service, &envFileOnly)
+	err = mapstructure.Decode(service, &envFileList)
 
 	if err == nil {
-		return envFileOnly.EnvFile, nil
+		return envFileList.EnvFile, nil
 	}
 
 	return nil, err
--- a/internal/pkg/docker/dockercompose/decompose.go
+++ b/internal/pkg/docker/dockercompose/decompose.go
@@ -37,6 +37,7 @@ func DecomposeService(content []byte, svcName string, workingDir string) (*Conve
 
 	// workaround for compose-go automatically loading env files from disk and merging their
 	// content into the "environment" field (the equivalent of the Copilot variables key)
+	// this separates them from the YAML before compose-go has a chance to parse the YAML.
 	// TODO (rclinard-amzn): make a PR to compose-go instead of using this workaround
 	envFiles, err := isolateEnvFiles(service)
 	if err != nil {
--- a/internal/pkg/docker/dockercompose/decompose.go
+++ b/internal/pkg/docker/dockercompose/decompose.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"github.com/compose-spec/compose-go/loader"
 	compose "github.com/compose-spec/compose-go/types"
-	"github.com/mitchellh/mapstructure"
 	"sort"
 )
 
@@ -136,29 +135,35 @@ func unsupportedServiceKeys(service map[string]any, svcName string) (IgnoredKeys
 }
 
 // isolateEnvFiles manually handles the env_file key so that compose-go does not try to read the env files from the disk.
-func isolateEnvFiles(service map[string]interface{}) ([]string, error) {
-	envFileStr := struct {
-		EnvFile string `mapstructure:"env_file"`
-	}{}
-	err := mapstructure.Decode(service, &envFileStr)
-
-	if err == nil && envFileStr.EnvFile != "" {
-		return []string{envFileStr.EnvFile}, nil
+func isolateEnvFiles(service map[string]any) ([]string, error) {
+	envFile, ok := service["env_file"]
+	if !ok || envFile == nil {
+		return nil, nil
 	}
 
-	envFileList := struct {
-		EnvFile []string `mapstructure:"env_file"`
-	}{}
-	err = mapstructure.Decode(service, &envFileList)
-
-	if err == nil {
-		return envFileList.EnvFile, nil
+	if envFileStr, ok := envFile.(string); ok {
+		return []string{envFileStr}, nil
 	}
 
-	return nil, err
+	envFileList, ok := envFile.([]any)
+	if !ok {
+		return nil, fmt.Errorf("expected string or string array for env_file key, but got %v (%T)", envFile, envFile)
+	}
+
+	envFiles := make([]string, len(envFileList))
+	for i, ent := range envFileList {
+		ef, ok := ent.(string)
+		if !ok {
+			return nil, fmt.Errorf("expected string or string array for env_file key, but got %v where %v is not a string", envFile, ent)
+		}
+
+		envFiles[i] = ef
+	}
+
+	return envFiles, nil
 }
 
-func removeEnvFiles(services map[string]map[string]interface{}) {
+func removeEnvFiles(services composeServices) {
 	for _, svc := range services {
 		delete(svc, "env_file")
 	}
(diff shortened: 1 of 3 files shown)


internal/pkg/docker/dockercompose/decompose_test.go
```diff
@@ -193,6 +199,7 @@ func TestDecomposeService(t *testing.T) {
 					Platform: manifest.PlatformArgsOrString{
 						PlatformString: (*manifest.PlatformString)(aws.String("linux/arm64")),
 					},
+					EnvFile: aws.String("/file-that-does-not-exist.env"),
```
paragbhingre, line 202: Why do we need this?

internal/pkg/docker/dockercompose/decompose.go
```diff
@@ -34,6 +35,15 @@ func DecomposeService(content []byte, svcName string, workingDir string) (*Conve
...
+	// workaround for compose-go automatically loading env files from disk and merging their
+	// content into the "environment" field (the equivalent of the Copilot variables key)
+	// TODO (rclinard-amzn): make a PR to compose-go instead of using this workaround
+	envFiles, err := isolateEnvFiles(service)
+	if err != nil {
+		return nil, nil, fmt.Errorf("isolate env file(s): %w", err)
+	}
+	removeEnvFiles(services)
+
 	project, err := loader.Load(compose.ConfigDetails{
```
paragbhingre, line 48: Do compose-go try to read the env files from the disk here? Sorry I am not sure where that is happening and how we are stopping that here by `isolateEnvFiles`? 
Or we are just overriding it on the [L65](https://github.com/aws/copilot-cli/pull/3935/files#diff-16da4d8aa47f3d1ec247adb442017f524fba30d0f9373942e2635709c223fc65R65)?

internal/pkg/docker/dockercompose/decompose_test.go
```diff
@@ -143,6 +143,12 @@ func TestDecomposeService(t *testing.T) {
 				},
 			},
 		},
+		"multiple env files": {
```
huanjani, line 146: 🤔 Since we're already breaking down the env files into vars, couldn't we loop through multiple env files and combine their contents (filtering out duplicates)? We'd have to handle clashes/overrides.

internal/pkg/docker/dockercompose/decompose.go
```diff
@@ -122,3 +133,32 @@ func unsupportedServiceKeys(service map[string]any, svcName string) (IgnoredKeys
...
+	envFileStr := struct {
+		EnvFile string `mapstructure:"env_file"`
+	}{}
+	err := mapstructure.Decode(service, &envFileStr)
+
+	if err == nil && envFileStr.EnvFile != "" {
+		return []string{envFileStr.EnvFile}, nil
+	}
+
+	envFileOnly := struct {
```
huanjani: would this var name make it a little clearer that the above is for
```
env_file: .env
```
while this is for
```
env_file:
  - ./a.env
  - ./b.env
```
?
```suggestion
	envFileList := struct {
```

huanjani (approved):
Clever way to handle the mismatch between the library and our needs! Adding the DNM label in case you want to address the teeny tiny nit.

internal/pkg/docker/dockercompose/decompose.go
rclinard-amzn: `compose-go` does try to read the env files from disk and merge them. So we detect the env files for the current service, remove them from all services in the YAML (since compose-go tries to load them for all services), ask compose-go to parse the YAML, and then put them back into the parsed Compose service. Hopefully that makes sense!

internal/pkg/docker/dockercompose/decompose.go
rclinard-amzn: Good idea, done!

internal/pkg/docker/dockercompose/decompose.go
```diff
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"github.com/compose-spec/compose-go/loader"
 	compose "github.com/compose-spec/compose-go/types"
+	"github.com/mitchellh/mapstructure"
```
dannyrandall: can you run `go mod tidy`? i think there should be a `go.mod` update...

internal/pkg/docker/dockercompose/decompose.go
```diff
@@ -122,3 +134,32 @@ func unsupportedServiceKeys(service map[string]any, svcName string) (IgnoredKeys
...
+	if err == nil && envFileStr.EnvFile != "" {
+		return []string{envFileStr.EnvFile}, nil
+	}
+
+	envFileList := struct {
+		EnvFile []string `mapstructure:"env_file"`
+	}{}
+	err = mapstructure.Decode(service, &envFileList)
+
+	if err == nil {
```
dannyrandall:
```suggestion
	if err := mapstructure.Decode(service, &envFileList); err == nil {
```

internal/pkg/docker/dockercompose/decompose.go
```diff
@@ -122,3 +134,32 @@ func unsupportedServiceKeys(service map[string]any, svcName string) (IgnoredKeys
 
 	return ignored, nil
 }
+
+// isolateEnvFiles manually handles the env_file key so that compose-go does not try to read the env files from the disk.
+func isolateEnvFiles(service map[string]interface{}) ([]string, error) {
+	envFileStr := struct {
+		EnvFile string `mapstructure:"env_file"`
+	}{}
+	err := mapstructure.Decode(service, &envFileStr)
```
dannyrandall: whats the difference between this and just doing something like:
```
envFile, ok := service["env_file"].(string)
```
i'd prefer not to add the dep if this is all it's doing...

internal/pkg/docker/dockercompose/decompose.go
rclinard-amzn: `mapstructure` is actually a dependency of `compose-go` so in a way we get it for "free" without having to add a new dependency in this PR!

dannyrandall: right, but it should still show up in `go.mod` when you explicitly use it 🤔 (also re: my later comment, even though it's in the dep tree, i'd rather not make us explicitly depend on it if it's only doing something simple)

rclinard-amzn:
I removed the `mapstructure` dependency, so Danny's comments should be resolved now!

internal/pkg/docker/dockercompose/decompose.go
rclinard-amzn: The situation is a little bit more complex for arrays but the required code is quite a bit less complex than I thought. Thanks for suggesting that I give it another look!

internal/pkg/docker/dockercompose/decompose.go
```diff
@@ -126,3 +137,38 @@ func unsupportedServiceKeys(service map[string]any, svcName string) (IgnoredKeys
...
+
+// isolateEnvFiles manually handles the env_file key so that compose-go does not try to read the env files from the disk.
+func isolateEnvFiles(service map[string]any) ([]string, error) {
+	envFile, ok := service["env_file"]
+	if !ok || envFile == nil {
+		return nil, nil
+	}
+
+	if envFileStr, ok := envFile.(string); ok {
+		return []string{envFileStr}, nil
```
dannyrandall, line 149: not really a suggestion, bc it's probably better the way you have it bc of the nested-ness, but wanted to tell you about the type switch
```
switch v := envFile.(type) {
case string:
  // can use v as a string
case []any:
default:
}
```

dannyrandall (approved):

internal/pkg/docker/dockercompose/decompose.go
rclinard-amzn: I didn't consider that feature, thanks for reminding me! I think you're probably right though about the nested-ness, though, that is one thing I was aiming to avoid here 👍

PR #3942, opened 2022-08-24
fix: add permission to enable `--generate-cmd`

huanjani:
#3095 changed the session for `--generate-cmd` from default to EnvManagerRole; that role needs the `states:DescribeStateMachine` permission.


dannyrandall (approved):

Lou1415926:
what do you think about adding a version check like this one: https://github.com/aws/copilot-cli/blob/0572e47c1339a9240fdcb61429ba2559fc67d561/internal/pkg/cli/job_run.go#L201

so that we remind people to upgrade their environment template so that the env manager role has the permission when `--generate-cmd` is executed against a job?

huanjani:
> what do you think about adding a version check like this one:
> 
> https://github.com/aws/copilot-cli/blob/0572e47c1339a9240fdcb61429ba2559fc67d561/internal/pkg/cli/job_run.go#L201
> 
> so that we remind people to upgrade their environment template so that the env manager role has the permission when `--generate-cmd` is executed against a job?

Should we do this for everyone who runs `task run` or just those who use the relatively rare `--generate-cmd` flag? Never hurts to upgrade env version, but also maybe annoying for people just running a quick task?

Lou1415926:
> Should we do this for everyone who runs task run or just those who use the relatively rare --generate-cmd flag? Never hurts to upgrade env version, but also maybe annoying for people just running a quick task?

I think just for those who use `--generate-cmd`! I agree that people that just run a task (and hence don't need the new permissions) shouldn't be forced to upgrade their env.

I'm happy to merge this as is, and have the version check as a follow-up because like you said, it's relatively rarely used :cool-crying

huanjani:
Okay, I can add the version check as a follow-up so this can move. Thanks!

Lou1415926 (approved):

PR #3949, opened 2022-08-25
fix: preserve existing service discovery endpoint

Lou1415926:
### The bug
Back in v1.9.0, we introduced “multiple environment per VPC”. This feature is shipped with upgrading the service discovery endpoint, from app.local to env.app.local. Since during env upgrade, we preserve all the old parameters, app.local has always been preserved for old environments.

However, env deploy didn’t account for this. It will try to use env.app.local for all environments, old or new. When env deploy is executed against an environment that was created pre-v1.9.0, it results in an attempt to update app.local to env.app.local. This will be breaking for environments that already have services in it, because the service discovery endpoint is exported and imported by the service stack, thus not updatable.

### The fix
The fix simply preserves the value of the  he existing `ServiceDiscoveryEndpoint` parameter. It also accounts for the scenario where `ServiceDiscoveryEndpoint` is not yet in the parameter (pre-v1.9.0 environment that has never been upgraded), and use `app.local` for those environment.


internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -243,22 +247,23 @@ func (e *EnvStackConfig) transformParameters(
 	// Remove or transform each of the current parameters.
 	var params []*cloudformation.Parameter
 	for k, p := range curr {
-		if transformed := transform(p, old[k]); transformed != nil {
-			params = append(params, transformed)
+		var currP = p
```
dannyrandall:
```suggestion
		currP := p
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -243,22 +247,23 @@ func (e *EnvStackConfig) transformParameters(
...
+// transformEnvControllerParameters transforms an env-controller managed parameter.
+// If the parameter exists in the old template, it uses its previous value. Otherwise, it returns its new default value.
+func transformEnvControllerParameters(new *cloudformation.Parameter, old *cloudformation.Parameter) *cloudformation.Parameter {
+	if new == nil {
+		return nil
 	}
-
 	var (
 		isEnvControllerManaged = make(map[string]struct{})
 		exists                 = struct{}{}
```
dannyrandall: i know you didn't change this but 🥺 
```suggestion
```
and just change below to `isEnvControllerManaged[f] = struct{}{}`

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -220,20 +220,24 @@ func (e *EnvStackConfig) StackName() string {
 	return NameForEnv(e.in.App.Name, e.in.Name)
 }
 
+type transformParameterFunc func(new *cloudformation.Parameter, old *cloudformation.Parameter) *cloudformation.Parameter
```
dannyrandall: same on all function signatures
```suggestion
type transformParameterFunc func(new, old *cloudformation.Parameter) *cloudformation.Parameter
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -243,31 +244,57 @@ func (e *EnvStackConfig) transformParameters(
...
 
-// transformEnvControllerParameters transforms a parameter such that it uses its previous value if:
-// 1. The parameter exists in the old template.
-// 2. The parameter is env-controller managed.
-// Otherwise, it returns the parameter untouched.
-func transformEnvControllerParameters(new cloudformation.Parameter, old *cloudformation.Parameter) *cloudformation.Parameter {
-	if old == nil { // The ParamKey doesn't exist in the old stack, use the new value.
-		return &new
+// transformEnvControllerParameters transforms an env-controller managed parameter.
+// If the parameter exists in the old template, it uses its previous value. Otherwise, it returns its new default value.
```
dannyrandall:
```suggestion
// If the parameter exists in the old template, it uses its previous value. Otherwise, it returns its new value.
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -243,31 +244,57 @@ func (e *EnvStackConfig) transformParameters(
...
+	}
+	if old == nil { // The EnvController-managed parameter doesn't exist in the old stack. Use the new value.
+		return new
+	}
+	return &cloudformation.Parameter{
+		ParameterKey: new.ParameterKey,
+
+		// Ideally, we would set `UsePreviousValue: true` unfortunately CodePipeline template config does not support it.
+		// https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/continuous-delivery-codepipeline-cfn-artifacts.html#w2ab1c21c15c15
+		ParameterValue: old.ParameterValue,
```
dannyrandall: isn't this the same as just returning old?

Lou1415926: Ideally the code here should be
```
ParameterKey: new.ParameterKey
UsePreviousValue: true
```
With this "ideal" code, there is no assumption on `old`'s `ParameterKey` name. This `transform...` function doesn't assume that `new.ParameterKey == old.ParameterKey`.

We can't `UsePreviousValue` because of the said reason in comment, and instead use `ParameterValue`. If we start returning `old` here, we are effectively adding the assumption that `new.ParameterKey == old.ParameterKey`.

dannyrandall: i see - that assumption doesn't exist. but it _is_ always true that `new.ParameterKey == old.ParameterKey`, right? (based on `transformParameters`)

dannyrandall: plus, what would `UsePreviousValue` do if `old.ParameterKey != new.ParameterKey`? i think using that field would imply that they're equal... otherwise there might not be a previous value to use

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -220,20 +220,21 @@ func (e *EnvStackConfig) StackName() string {
...
-// If `old` is `nil`, the parameter does not exist in the old template.
-// `transform` should return `nil` if caller intends to delete the parameter.
-func (e *EnvStackConfig) transformParameters(
-	currParams []*cloudformation.Parameter,
-	oldParams []*cloudformation.Parameter,
-	transform func(new cloudformation.Parameter, old *cloudformation.Parameter) *cloudformation.Parameter) ([]*cloudformation.Parameter, error) {
+// The parameter`transformFunc` are functions that transform a parameter, given its value in the new template and the old template.
+// Each transform functions should keep the following in mind:
+// 1. It should return `nil` if the parameter should be removed.
+// 2. The transform functions are applied in a convolutional manner.
```
huanjani, line 230: Sorry-- what does this mean?

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -243,31 +244,57 @@ func (e *EnvStackConfig) transformParameters(
...
+	if new == nil {
+		return nil
+	}
+	if aws.StringValue(new.ParameterKey) != EnvParamServiceDiscoveryEndpoint {
+		return new
+	}
+	if old == nil {
+		return &cloudformation.Parameter{
+			ParameterKey:   new.ParameterKey,
+			ParameterValue: aws.String(fmt.Sprintf(`%s.local`, e.in.App.Name)),
```
huanjani, line 294: wait, isn't the new default `env.app.local`?

Lou1415926: Yes but we want to use the old default in this case. Otherwise we'd be trying to update `app.local` to `env.app.local` which is the bug in the first place

internal/pkg/deploy/cloudformation/stack/env.go
Lou1415926: that the second filter is applied on the result of the first filter, and the third filter is applied on the result of the second filter, etc...
any suggestion on better wordings?

internal/pkg/deploy/cloudformation/stack/env.go
Lou1415926: you are right! updated the code :meow_blob:

internal/pkg/deploy/cloudformation/stack/env.go
huanjani: Ah, gotcha gotcha-- I missed the second part of "the fix."

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -243,39 +244,59 @@ func (e *EnvStackConfig) transformParameters(
...
 
-// transformEnvControllerParameters transforms a parameter such that it uses its previous value if:
-// 1. The parameter exists in the old template.
-// 2. The parameter is env-controller managed.
-// Otherwise, it returns the parameter untouched.
-func transformEnvControllerParameters(new cloudformation.Parameter, old *cloudformation.Parameter) *cloudformation.Parameter {
-	if old == nil { // The ParamKey doesn't exist in the old stack, use the new value.
-		return &new
+// transformEnvControllerParameters transforms an env-controller managed parameter.
+// If the parameter exists in the old template, it uses its previous value. Otherwise, it returns its new default value.
```
dannyrandall: maybe just document the assumption that `new.ParamKey == old.ParamKey`

dannyrandall (approved):
great fix 👏!

internal/pkg/deploy/cloudformation/stack/env.go
huanjani: 🤔 I dunno...maybe "sequential"?

huanjani (approved):
Looks good! Adding label in case you want to address nits on comments.

PR #3955, opened 2022-08-26
chore: Docker Compose import experiment milestone 4 - Named Volume Import

rclinard-amzn:
This PR implements Milestone 4 of the Docker Compose import experiment, adding partial support for the `volumes` key on services. This gets us extremely close to successfully importing one of our Compose file examples and this is the last of the really important but also straightforward feature milestones - everything from here on out is smaller things that need extra effort to support in Copilot but are also somewhat commonly used.

There are a lot of checks to make sure that volumes aren't being used to share data between services since this isn't something that Copilot currently accommodates with EFS. Additionally, bind mounts are currently completely unsupported.

Finally, some light support has been added for the mechanics of `tmpfs` volumes in Compose.

### DIFF
--- a/internal/pkg/docker/dockercompose/decompose_test.go
+++ b/internal/pkg/docker/dockercompose/decompose_test.go
@@ -429,3 +429,101 @@ func TestDecomposeService_ExposedPorts(t *testing.T) {
 
 	runDecomposeTests(t, actualTestCases)
 }
+
+func TestDecomposeService_Volumes(t *testing.T) {
+	efsTrue := manifest.EFSConfigOrBool{Enabled: aws.Bool(true)}
+
+	testCases := map[string]struct {
+		wantError         error
+		wantErrorContains string
+		wantStorage       manifest.Storage
+	}{
+		"simple-named": {
+			wantStorage: manifest.Storage{
+				Volumes: map[string]*manifest.Volume{
+					"nvol-1": {
+						EFS: efsTrue,
+						MountPointOpts: manifest.MountPointOpts{
+							ContainerPath: aws.String("/test/path/please/ignore"),
+							ReadOnly:      aws.Bool(false),
+						},
+					},
+				},
+			},
+		},
+		"bind-file": {
+			wantError: errors.New("convert Compose service to Copilot manifest: volume type \"bind\" is not supported yet"),
+		},
+		"bind-directory": {
+			wantError: errors.New("convert Compose service to Copilot manifest: volume type \"bind\" is not supported yet"),
+		},
+		"sharing1": {
+			wantError: errors.New("convert Compose service to Copilot manifest: named volume nvol-shared-1 is shared with service [sharing2], this is not supported in Copilot"),
+		},
+		"sharing2": {
+			wantError: errors.New("convert Compose service to Copilot manifest: named volume nvol-shared-1 is shared with service [sharing1], this is not supported in Copilot"),
+		},
+		"sharing-within": {
+			wantError: errors.New("convert Compose service to Copilot manifest: cannot mount named volume nvol-shared-2 a second time at /test2, it is already mounted at /test"),
+		},
+		"mount-ro-simple": {
+			wantStorage: manifest.Storage{
+				Volumes: map[string]*manifest.Volume{
+					"nvol-2": {
+						EFS: efsTrue,
+						MountPointOpts: manifest.MountPointOpts{
+							ContainerPath: aws.String("/testdir"),
+							ReadOnly:      aws.Bool(true),
+						},
+					},
+				},
+			},
+		},
+		"mount-ro-selinux-simple": {
+			wantStorage: manifest.Storage{
+				Volumes: map[string]*manifest.Volume{
+					"nvol-3": {
+						EFS: efsTrue,
+						MountPointOpts: manifest.MountPointOpts{
+							ContainerPath: aws.String("/testdir"),
+							ReadOnly:      aws.Bool(true),
+						},
+					},
+				},
+			},
+		},
+		"uses-external-volume": {
+			wantError: errors.New("convert Compose service to Copilot manifest: named volume nvol-external is marked as external, this is unsupported"),
+		},
+	}
+
+	actualTestCases := map[string]decomposeTest{}
+
+	for name, tc := range testCases {
+		actualTestCases[name] = decomposeTest{
+			svcName:           name,
+			filename:          "volume-tests.yml",
+			wantError:         tc.wantError,
+			wantErrorContains: tc.wantErrorContains,
+			wantBs: &manifest.BackendServiceConfig{
+				ImageConfig: manifest.ImageWithHealthcheckAndOptionalPort{
+					ImageWithOptionalPort: manifest.ImageWithOptionalPort{
+						Image: manifest.Image{
+							Location: aws.String("nginx"),
+						},
+					},
+				},
+				TaskConfig: manifest.TaskConfig{
+					CPU:    aws.Int(256),
+					Memory: aws.Int(512),
+					Count: manifest.Count{
+						Value: aws.Int(1),
+					},
+					Storage: tc.wantStorage,
+				},
+			},
+		}
+	}
+
+	runDecomposeTests(t, actualTestCases)
+}
--- a/internal/pkg/docker/dockercompose/decompose_test.go
+++ b/internal/pkg/docker/dockercompose/decompose_test.go
@@ -128,7 +128,7 @@ func TestDecomposeService_General(t *testing.T) {
 			filename: "nginx-golang-postgres.yml",
 			svcName:  "proxy",
 
-			wantError: errors.New("\"services.proxy\" relies on fatally-unsupported Compose keys: volumes"),
+			wantError: errors.New("convert Compose service to Copilot manifest: volume type bind is not supported yet"),
 		},
 		"react-express-mongo frontend": {
 			filename: "react-express-mongo.yml",
--- a/internal/pkg/docker/dockercompose/decompose_test.go
+++ b/internal/pkg/docker/dockercompose/decompose_test.go
@@ -128,7 +128,7 @@ func TestDecomposeService_General(t *testing.T) {
 			filename: "nginx-golang-postgres.yml",
 			svcName:  "proxy",
 
-			wantError: errors.New("convert Compose service to Copilot manifest: volume type \"bind\" is not supported yet"),
+			wantError: errors.New(`convert Compose service to Copilot manifest: volume type "bind" is not supported yet`),
 		},
 		"react-express-mongo frontend": {
 			filename: "react-express-mongo.yml",
--- a/internal/pkg/docker/dockercompose/decompose_test.go
+++ b/internal/pkg/docker/dockercompose/decompose_test.go
@@ -74,25 +74,25 @@ func TestDecomposeService_General(t *testing.T) {
 			filename: "bad-services-compose.yml",
 			svcName:  "test",
 
-			wantError: errors.New("\"services\" top-level element was not a map, was: invalid"),
+			wantError: errors.New(`"services" top-level element was not a map, was: invalid`),
 		},
 		"wrong name": {
 			filename: "unsupported-keys.yml",
 			svcName:  "test",
 
-			wantError: errors.New("no service named \"test\" in this Compose file, valid services are: fatal1, fatal2, fatal3"),
+			wantError: errors.New(`no service named "test" in this Compose file, valid services are: fatal1, fatal2, fatal3`),
 		},
 		"invalid service not a map": {
 			filename: "bad-service-compose.yml",
 			svcName:  "bad",
 
-			wantError: errors.New("\"services.bad\" element was not a map"),
+			wantError: errors.New(`"services.bad" element was not a map`),
 		},
 		"unsupported keys fatal1": {
 			filename: "unsupported-keys.yml",
 			svcName:  "fatal1",
 
-			wantError: errors.New("\"services.fatal1\" relies on fatally-unsupported Compose keys: external_links, privileged"),
+			wantError: errors.New(`"services.fatal1" relies on fatally-unsupported Compose keys: external_links, privileged`),
 		},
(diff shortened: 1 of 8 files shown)


internal/pkg/docker/dockercompose/svc.go
```diff
@@ -52,6 +52,7 @@ func convertService(service *compose.ServiceConfig, workingDir string) (*Convert
 		return nil, nil, err
 	}
 	ignored = append(ignored, portIgnored...)
+	ignored = append(ignored, tcIgn...)
```
dannyrandall: nit: move this right after the `convertTaskConfig` err check to keep it close

internal/pkg/docker/dockercompose/volume.go
```diff
@@ -0,0 +1,145 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package dockercompose
+
+import (
+	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/copilot-cli/internal/pkg/manifest"
+	compose "github.com/compose-spec/compose-go/types"
+	"github.com/docker/go-units"
```
dannyrandall, line 11: `go mod tidy` pls 😊

internal/pkg/docker/dockercompose/volume.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+		}
+	}
+
+	var ephemeralBytes uint64 = 20 * units.GiB
+	var ignored IgnoredKeys
+
+	for idx, vol := range volumes {
+		if vol.Target == "" {
+			return nil, nil, fmt.Errorf("volume mounted from \"%v\" (type \"%v\") is missing a target mount point", vol.Source, vol.Type)
```
dannyrandall: `%q` will add quotes for you!

internal/pkg/docker/dockercompose/volume.go
```diff
@@ -0,0 +1,145 @@
...
+			},
+			MountPointOpts: mountOpts,
+		}
+	}
+
+	// math trick to round up to the next highest GiB if it's not an even size in GiB
+	ephemeralGiB := (ephemeralBytes + units.GiB - 1) / units.GiB
+	var storage manifest.Storage
+
+	// default: 20 GiB
```
dannyrandall: is this default a compose default? or copilot?

internal/pkg/docker/dockercompose/volume.go
```diff
@@ -0,0 +1,145 @@
...
+func (vc *volumeConverter) convertVolumes(volumes []compose.ServiceVolumeConfig, otherSvcs compose.Services) (*manifest.Storage, IgnoredKeys, error) {
+	for _, otherSvc := range otherSvcs {
+		for _, vol := range otherSvc.Volumes {
+			if vol.Type == "volume" {
+				vc.otherSvcVols[vol.Source] = append(vc.otherSvcVols[vol.Source], otherSvc.Name)
+			}
+		}
+	}
+
+	var ephemeralBytes uint64 = 20 * units.GiB
```
dannyrandall: maybe make this a const

internal/pkg/docker/dockercompose/volume.go
```diff
@@ -0,0 +1,145 @@
...
+	var ignored IgnoredKeys
+
+	if topLevel, ok := vc.topLevelVols[name]; ok {
+		var err error
+
+		switch {
+		case topLevel.External.External:
+			err = fmt.Errorf("named volume %s is marked as external, this is unsupported", name)
+		case topLevel.Driver != "":
+			err = fmt.Errorf("only the default driver is supported, but the volume %s tries to use a different driver", name)
```
dannyrandall:
```suggestion
		switch {
		case topLevel.External.External:
			return nil, nil, fmt.Errorf("named volume %s is marked as external, this is unsupported", name)
		case topLevel.Driver != "":
			return nil, nil, fmt.Errorf("only the default driver is supported, but the volume %s tries to use a different driver", name)
```

internal/pkg/docker/dockercompose/volume.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	if len(vc.copilotVols) != 0 {
+		storage.Volumes = vc.copilotVols
+	}
+
+	return &storage, ignored, nil
+}
+
+func (vc *volumeConverter) checkNamedVolume(vol compose.ServiceVolumeConfig) (*string, IgnoredKeys, error) {
```
huanjani: great checks!

internal/pkg/docker/dockercompose/volume.go
rclinard-amzn: I didn't know that, that's really neat!

internal/pkg/docker/dockercompose/volume.go
rclinard-amzn: It's a Copilot default - I added a comment to clarify!

rclinard-amzn:
Addressed! Thanks for the feedback, `%q` is a neat feature!

internal/pkg/docker/dockercompose/volume.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/aws/copilot-cli/internal/pkg/manifest"
+	compose "github.com/compose-spec/compose-go/types"
+	"github.com/docker/go-units"
+	"github.com/dustin/go-humanize/english"
+	"strings"
+)
+
+// The default size that Copilot reserves for a service's ephemeral on-disk storage.
+const defaultEphemeralGiB = 20
+const defaultEphemeralBytes uint64 = defaultEphemeralGiB * units.GiB
```
huanjani, line 18: nit: 
```suggestion
const (
  defaultEphemeralGiB = 20
  defaultEphemeralBytes uint64 = defaultEphemeralGiB * units.GiB
)  
```

internal/pkg/docker/dockercompose/volume.go
```diff
@@ -0,0 +1,143 @@
...
+func newVolumeConverter(topLevelVols compose.Volumes) volumeConverter {
+	return volumeConverter{
+		otherSvcVols:  map[string][]string{},
+		copilotVols:   map[string]*manifest.Volume{},
+		topLevelVols:  topLevelVols,
+		tmpfsVolNames: map[string]bool{},
+	}
+}
+
+// convertVolumes converts a list of Compose volumes into the Copilot storage equivalent.
```
huanjani: Q: Does this handle both short syntax and long syntax?

rclinard-amzn: Yes! Compose converts short syntax to long syntax in almost all places, and my test cases use both.

internal/pkg/docker/dockercompose/svc.go
```diff
@@ -22,16 +22,17 @@ type ConvertedService struct {
...
 
-func convertService(service *compose.ServiceConfig, workingDir string) (*ConvertedService, IgnoredKeys, error) {
+func convertService(service *compose.ServiceConfig, workingDir string, otherSvcs compose.Services, vols compose.Volumes) (*ConvertedService, IgnoredKeys, error) {
 	image, ignored, err := convertImageConfig(service.Build, service.Labels, service.Image)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	taskCfg, err := convertTaskConfig(service)
+	taskCfg, tcIgn, err := convertTaskConfig(service, otherSvcs, vols)
```
bvtujo: Can we rename this to something more descriptive, like `taskCfgIgnoredKeys`, for easier reading?

internal/pkg/docker/dockercompose/svc.go
```diff
@@ -22,16 +22,17 @@ type ConvertedService struct {
 	BackendSvc *manifest.BackendService
 }
 
-func convertService(service *compose.ServiceConfig, workingDir string) (*ConvertedService, IgnoredKeys, error) {
+func convertService(service *compose.ServiceConfig, workingDir string, otherSvcs compose.Services, vols compose.Volumes) (*ConvertedService, IgnoredKeys, error) {
 	image, ignored, err := convertImageConfig(service.Build, service.Labels, service.Image)
```
bvtujo: Can we rename this `ignored` var to something more descriptive, like `ignoredComposeKeys`?

internal/pkg/docker/dockercompose/svc.go
```diff
@@ -193,16 +194,22 @@ func toExposedPort(binding compose.ServicePortConfig) (*exposedPort, IgnoredKeys
 }
 
 // convertTaskConfig converts environment variables, env files, and platform strings.
```
bvtujo: Can this comment be made more descriptive? What are the top level volumes, in comparison to other volumes which might exist?

internal/pkg/docker/dockercompose/decompose_test.go
```diff
@@ -122,31 +122,31 @@ func TestDecomposeService_General(t *testing.T) {
...
 
-			wantError: errors.New("\"services.db\" relies on fatally-unsupported Compose keys: secrets, volumes"),
+			wantError: errors.New("\"services.db\" relies on fatally-unsupported Compose keys: secrets"),
 		},
 		"nginx-golang-postgres proxy": {
 			filename: "nginx-golang-postgres.yml",
 			svcName:  "proxy",
 
-			wantError: errors.New("\"services.proxy\" relies on fatally-unsupported Compose keys: volumes"),
+			wantError: errors.New("convert Compose service to Copilot manifest: volume type \"bind\" is not supported yet"),
```
dannyrandall: nit
```suggestion
			wantError: errors.New(`convert Compose service to Copilot manifest: volume type "bind" is not supported yet`),
```

internal/pkg/docker/dockercompose/volume.go
```diff
@@ -0,0 +1,143 @@
...
+			return nil, nil, fmt.Errorf("volume mounted from %q (type %q) is missing a target mount point", vol.Source, vol.Type)
+		}
+
+		mountOpts := manifest.MountPointOpts{
+			ContainerPath: aws.String(vol.Target),
+			ReadOnly:      aws.Bool(vol.ReadOnly),
+		}
+
+		if vol.Type == "tmpfs" {
+			ephemeralBytes += uint64(vol.Tmpfs.Size)
```
dannyrandall, line 61: i don't know much about volumes in compose so i might just not understand something - but it seems like the logic is to:
a. default to 20gb 
b. if tmpfs, add 20gb to whatever the sum of configured volumes is

and i'm curious about option b there - why are we adding 20 gb no matter what they specify?

internal/pkg/docker/dockercompose/volume.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	if len(vc.copilotVols) != 0 {
+		storage.Volumes = vc.copilotVols
+	}
+
+	return &storage, ignored, nil
+}
+
+func (vc *volumeConverter) checkNamedVolume(vol compose.ServiceVolumeConfig) (*string, IgnoredKeys, error) {
```
dannyrandall: why is the first return value a `*string`? if you don't need a pointer for a specific reason it should probably just be a `string`

bvtujo (approved):
Overall very thorough, great checks. I'd love an overview of exactly what IS supported, and some clear docs on the types of compose files that will pass this validation and conversion, but the code looks solid!

internal/pkg/docker/dockercompose/volume.go
rclinard-amzn: That analysis is correct! It's an approximation of Compose behavior because the only relevant meaning of a tmpfs mount that matters to Copilot is if the service needs it for additional temporary storage.

I chose this approach, since if for example the service mounted a 100mb tmpfs volume as extra space then we wouldn't want to reduce the total ephemeral storage to 100MB. But, if it mounts a 50 GB tmpfs volume, then certainly we'd need that amount of additional space. I think this approach is the most sensible sensible since in either case, this method will ensure that the needs of the service are met.

internal/pkg/docker/dockercompose/volume.go
rclinard-amzn: Good catch, fixed!

internal/pkg/docker/dockercompose/decompose_test.go
rclinard-amzn: Neat! Another feature I'm learning about 😄

internal/pkg/docker/dockercompose/svc.go
rclinard-amzn: Done! 

```
// topLevelVols is the top-level volume list in the Compose file shared between services, as opposed to the volumes
// key of the service config which describes bind mounts and mounts of top-level named volumes.
```

rclinard-amzn:
Thanks for the reviews! I think that feedback should be addressed, aside from the notes on documenting what is and is not supported. I've tried to partially address that but ultimately I'll probably need to include that in the doc at the end of the project!

internal/pkg/docker/dockercompose/decompose_test.go
```diff
@@ -122,31 +122,31 @@ func TestDecomposeService_General(t *testing.T) {
...
 
-			wantError: errors.New("\"services.proxy\" relies on fatally-unsupported Compose keys: volumes"),
+			wantError: errors.New(`convert Compose service to Copilot manifest: volume type "bind" is not supported yet`),
 		},
 		"react-express-mongo frontend": {
 			filename: "react-express-mongo.yml",
 			svcName:  "frontend",
 
-			wantError: errors.New("\"services.frontend\" relies on fatally-unsupported Compose keys: networks, volumes"),
+			wantError: errors.New("\"services.frontend\" relies on fatally-unsupported Compose keys: networks"),
```
dannyrandall: you can add the backticks to the other test cases as well :)

dannyrandall (approved):
great work! 🎉

internal/pkg/docker/dockercompose/decompose_test.go
rclinard-amzn: Done! No more `\"` in this package 😄

PR #3960, opened 2022-08-29
chore: remove svc logs features for patch

dannyrandall:
Revert before v1.22.0.


internal/pkg/cli/svc_logs.go
```diff
@@ -351,7 +351,7 @@ func buildSvcLogsCmd() *cobra.Command {
 	cmd.Flags().IntVar(&vars.limit, limitFlag, 0, limitFlagDescription)
 	cmd.Flags().StringSliceVar(&vars.taskIDs, tasksFlag, nil, tasksLogsFlagDescription)
 	cmd.Flags().StringVar(&vars.logGroup, logGroupFlag, "", logGroupFlagDescription)
-	cmd.Flags().BoolVarP(&vars.previous, previousFlag, previousFlagShort, false, previousFlagDescription)
-	cmd.Flags().StringVar(&vars.containerName, containerLogFlag, "", containerLogFlagDescription)
+	// cmd.Flags().BoolVarP(&vars.previous, previousFlag, previousFlagShort, false, previousFlagDescription)
```
Lou1415926, line 354: We can probably use this feature https://pkg.go.dev/github.com/spf13/pflag#readme-hidden-flags to hide some flags!

dannyrandall: i considered that - i'm good either way. hidden flags they can still use the feature if wanted, this way they can't :) do you have a preference?

huanjani: Sounds like @paragbhingre is going to update the feature further, so maybe continue with the revert? Parag, can you confirm?

Lou1415926: an alternative would be putting everything behind a "featureFlag"
```go
type svcLogsOpts struct {
   featureFlag bool
}

func (o *svcLogsOpts) Execute() error {
  if o.previous && o.featureFlag {}
  var container string
  if o.featureFlag {
    container = o.containerName
  }
  err := o.logsSvc.WriteLogEvents(logging.WriteLogEventsOpts{
		Follow:        o.follow,
		Limit:         limit,
		EndTime:       o.endTime,
		StartTime:     o.startTime,
		TaskIDs:       o.taskIDs,
		OnEvents:      eventsWriter,
		ContainerName: containerName,
	})
}
``` 

but I'm happy with either way! Just putting this out there!

dannyrandall: i'm in favor of just keeping it simple...

paragbhingre: I think what @dannyrandall has done is fine. I can cut a separate PR to undo Danny's changes after patch release and add display `taskId` feature that we discussed in the meeting.

huanjani (approved):

internal/pkg/cli/flag.go
```diff
@@ -137,7 +137,7 @@ const (
 	envFlagShort      = "e"
 	typeFlagShort     = "t"
 	workloadFlagShort = "w"
-	previousFlagShort = "p"
+	// previousFlagShort = "p"
```
paragbhingre, line 140: We can comment comment [previous flag](https://github.com/aws/copilot-cli/blob/d4c353e0ddbcd0e1c644780391b01b6024d1d5ac/internal/pkg/cli/flag.go#L41) too.

paragbhingre (commented):
Just want to confirm that are we reverting Wanxian's container feature as well?

internal/pkg/cli/flag.go
dannyrandall: it's used in https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/cli/svc_logs.go#L282, so we don't need to comment it out

dannyrandall:
@paragbhingre yes, we are! that's the `--container` feature.

internal/pkg/cli/flag.go
paragbhingre: Yeah that's just a validation but yeah we can keep it anyway.

paragbhingre (approved):

PR #3956, opened 2022-08-26
fix: ensure env manager role has permission to upload

Lou1415926:
Basically bringing back logic in #3562  and #3567 .

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the Apache 2.0 License.

PR #3966, opened 2022-08-30
chore: env manager patcher and service discovery namespace should work with bootstrap stack

Lou1415926:
Previously, we have two fixes:
1. #3956 which patches the S3 permissions to the environment manager role if it's needed.
2. #3949 which preserves service discovery endpoint namespace from a previously deployed env stack.

Fix 1 performs an environment template version check to infer whether the necessary permissions are present. This version check fails to take into account the `"bootstrap"` version, which is present when the environment is only bootstrapped (by running `env init` and not having run `env deploy`)

Fix 2 grabs the old parameter from a deployed environment stack. It fails to take into consideration that, while an env stack that's on `"bootstrap"` version is not considered a "deployed environment" in Copilot, it is a deployed stack in CFN.

### DIFF
--- a/internal/pkg/deploy/cloudformation/env.go
+++ b/internal/pkg/deploy/cloudformation/env.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/copilot-cli/internal/pkg/template"
+	"gopkg.in/yaml.v3"
 
 	"github.com/aws/aws-sdk-go/aws"
 	awscfn "github.com/aws/aws-sdk-go/service/cloudformation"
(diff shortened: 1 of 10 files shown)


internal/pkg/deploy/cloudformation/env.go
```diff
@@ -188,3 +196,18 @@ func (cf CloudFormation) cachedStack(stackName string) (*cloudformation.StackDes
 	cf.cachedDeployedStack = stackDescr
 	return cf.cachedDeployedStack, nil
 }
+
+// isInitialDeployment returns the environment stack's parameters.
```
huanjani: nit: 
```suggestion
// isInitialDeployment returns whether this is the first deployment of the environment stack.
```

huanjani (approved):
LGTM! I hadn't realized previously that we were using "boostrap" for version! 😄

dannyrandall (approved):

Issue #2986, opened 2021-11-03
Support IAM roles with permission boundaries


Lou1415926:
Permission boundaries is supported for IAM entities, so that only actions that are allowed by both the identity-based policies and the permissions boundaries can be performed.

Some organizations require IAM roles to be created with permission boundaries, for example, #2933 and #2969. Copilot should be able to support that.

PeterBengtson:
I second this request. Permission boundaries are vital to a least-privilege system in enterprise situations, and it is the only thing preventing me from recommending Copilot wholeheartedly for a migration involving 500 developers. 

The simplest implementation would probably be to add a `boundary_permission` key to the service manifests for the various architectures, overridable per environment. It is important that Roles are created with permission boundaries, as SCPs otherwise will prevent a Role from being created.

Lou1415926:
Hello @PeterBengtson. Thank you very much for your input! I have a few follow-up questions to understand your use case better:
1. Are the manifests editable by developers, or are they strictly edited by DevOps/Platform teams?
2. If the manifests are editable by developers - what prevents a developer to change the permission boundary policy in the manifest and get a lot more access?
3. Do you want all IAM roles (e.g. `EnvManagerRole`, `AppDelegationRole`, etc.) to be created with permission boundary, or just IAM roles created with addon?

PeterBengtson:
@Lou1415926 : 
1. Yes, the manifests are of course editable by developers. The Boundary Permission policy will restrict what teams _actually_ can do, no matter what the teams put in their own policies.
2. The permission boundary policies are presupplied by the Foundation Team (or a Platform Team). They are protected from modification by developers by an SCP which also prohibits any role from being created unless it is with the presupplied policy/policies. It also blocks the removal of the policy. The SCP in question is a best practice one from AWS; you can find it here: https://getstarted.awsworkshop.io/03-dev/07-reference/02-controlling-builder-team-access.html.
3. It would have to be all roles, as the SCP will block all new roles without the presupplied Boundary Permission policy from being created.

Here's the relevant section of the SCP:
```
       {
            "Sid": "AllowRoleWithPBs",
            "Effect": "Allow",
            "Action": [
                "iam:CreateRole",
                "iam:AttachRolePolicy",
                "iam:DetachRolePolicy",
                "iam:PutRolePolicy",
                "iam:DeleteRolePolicy"
            ],
            "NotResource": "arn:aws:iam::*:role/example-infra-*",
            "Condition": {
                "StringLike": {
[...]

PR #3969, opened 2022-09-01
feat: enable IAM permissions boundaries

huanjani:
Related: #2986.

This is the first PR in the IAM permissions boundary feature.

Some companies leverage AWS Organizations's [Service Control Policies](https://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_policies_scps.html) (SCP) feature, which can be used together with [IAM Permissions Boundaries](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_boundaries.html) (see diagram below) to [limit](https://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_policies_scps.html#scp-effects-on-permissions) the actions devs may perform and the resources they may affect. 
image

By allowing users to indicate their existing permissions boundary policies when running `app init`, Copilot can add them to IAM Roles when creating them. This PR conditionally adds a boundary policy to the IAM roles created at the app level. More IAM Role modifications to come.

### DIFF
--- a/internal/pkg/template/templates/app/app.yml
+++ b/internal/pkg/template/templates/app/app.yml
@@ -42,6 +42,9 @@ Resources:
               Service: cloudformation.amazonaws.com
             Action:
               - sts:AssumeRole
+{{- if .PermissionsBoundary }}
+      PermissionsBoundary: !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:policy/{{.PermissionsBoundary}}'
+{{- end }} 
       Path: /
       Policies:
         - PolicyName: AssumeRole-AWSCloudFormationStackSetExecutionRole
@@ -67,6 +70,9 @@ Resources:
               AWS: !GetAtt AdministrationRole.Arn
             Action:
               - sts:AssumeRole
+{{- if .PermissionsBoundary }}
+      PermissionsBoundary: !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:policy/{{.PermissionsBoundary}}'
+{{- end }}
       Path: /
       Policies:
       - PolicyName: ExecutionRolePolicy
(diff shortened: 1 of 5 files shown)


internal/pkg/template/templates/app/app.yml
```diff
@@ -42,6 +42,9 @@ Resources:
               Service: cloudformation.amazonaws.com
             Action:
               - sts:AssumeRole
+{{- if .PermissionsBoundary }}
+      PermissionsBoundary: !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:policy/{{.PermissionsBoundary}}'
```
Lou1415926, line 46: Just to make sure - we are assuming that the PermissionBoundary lives in the account that deploys the app, is that right?

Lou1415926 (approved):
LGTM! We probably will need to update some integration test too 🤔

internal/pkg/template/templates/app/app.yml
huanjani: For these IAM roles, yes. If they deploy cross-account environments, we'll refer to those account IDs.

dannyrandall: does that mean they'd need a copy of the permission boundary in every account they deploy in to? would it make sense to let them specify the full ARN and reference a permission boundary in a specific account? (i have no idea if that's possible, just checking)

dannyrandall (approved):
looks good, just a question!

internal/pkg/template/templates/app/app.yml
efekarakus: They'd duplicate the same managed policy in all the accounts! The policy name should be the same in the app and env accounts as well (that's why we can safely use the flag value for _all_ our iam roles)

iamhopaul123: We might need to upgrade the app template version as well 👀

huanjani: upgraded app template version. thanks!

Issue #3961, opened 2022-08-29
Showcase task id for `copilot svc logs --previous` command


paragbhingre:
When customers run `copilot svc logs --previous` then we should also be able to display the task id for which we are displaying the logs. 
It would be great if we also showcase AWS console link to the task logs.

PR #3983, opened 2022-09-07
feat: previously stopped task with the stopped task id

### LINKED ISSUE #3272: Proposal: `logs -p` or `logs --open`

paragbhingre:
Fixes #3272 and #3961, reverting #3960


internal/pkg/cli/svc_logs.go
```diff
@@ -209,6 +209,7 @@ func (o *svcLogsOpts) Execute() error {
 			return err
 		}
 		o.taskIDs = []string{taskID}
+		fmt.Printf("previously stopped task: %s\n", taskID)
```
huanjani: delete?

internal/pkg/cli/svc_logs_test.go
```diff
@@ -106,6 +108,14 @@ func TestSvcLogs_Validate(t *testing.T) {
 
 			wantedError: fmt.Errorf("--limit 10001 is out-of-bounds, value must be between 1 and 10000"),
 		},
+		"returns error if both previous and tasks flags are defined": {
```
huanjani, line 112: Do we also have any unit tests for success with taskID(s)?

paragbhingre: Yes, I had implemented that unit test as a part of original ask of --previous flag. You can check the unit test [here](https://github.com/aws/copilot-cli/blob/e1ae307bda11ab8d8e940f9fefd398ced094bb16/internal/pkg/cli/svc_logs_test.go#L374)

internal/pkg/cli/svc_logs.go
paragbhingre: Ohh so we discussed in the meeting that we want to show taskId too as a part of `--previous` functionality. That is why I am displaying taskId before we show logs for the that stopped task.

huanjani: Ah, gotcha. I was thrown off by the `fmt.Printf`...should we use a `log.` instead?

huanjani (approved):
Looks good. One nit, then feel free to remove label!

Lou1415926 (approved):
LGTM! and +1 to @huanjani about using `log`. Does this fix #3272 though? Did you mean to link a different issue?

paragbhingre:
> LGTM! and +1 to @huanjani about using `log`. Does this fix #3272 though? Did you mean to link a different issue?

We had reverted the changes for the fix #3272 in #3960 right? So this change adds those changes back along with printing task id. Does that make sense?

PR #3986, opened 2022-09-09
chore: retrieve RDWS type so that the loggings client is correctly configured

Lou1415926:


iamhopaul123 (approved):

internal/pkg/cli/svc_logs.go
```diff
@@ -119,14 +118,17 @@ func newSvcLogOpts(vars svcLogsVars) (*svcLogsOpts, error) {
...
+			return fmt.Errorf("get workload: %w", err)
+		}
 		newWorkloadLoggerOpts := &logging.NewWorkloadLoggerOpts{
 			App:  opts.appName,
 			Env:  opts.envName,
 			Name: opts.name,
 			Sess: sess,
 		}
-		if opts.targetSvcType != manifest.RequestDrivenWebServiceType {
+		if workload.Type != manifest.RequestDrivenWebServiceType {
```
efekarakus: Why do we fetch from SSM instead of using the workspace value? did it result in a bug?

Lou1415926: See if you like the updated implementation better!

Previously it was introducing a bug because `DeployedWorkload` in `selector` package did not always return `SvcTyp`. This resulted in `opts.targetSvcType` being `""`.

efekarakus (approved):
Nice, I like this so much more 😂 this fix should prevent future bugs

PR #3987, opened 2022-09-09
chore: validate env version for `task run --generate-cmd`

huanjani:
Follow-up to #3942. Fails earlier if user's env version doesn't support the `--generate-cmd` flag for the `task run` command due to missing permissions.


internal/pkg/cli/task_run.go
```diff
@@ -501,6 +516,30 @@ func (o *runTaskOpts) confirmSecretsAccess() error {
...
+		return err
+	}
+	version, err := envStack.Version()
+	if err != nil {
+		return fmt.Errorf("retrieve version of environment stack %q in application %q: %v", o.env, o.appName, err)
+	}
+	// The '--generate-cmd' flag was introduced in env v1.4.0. In env v1.8.0,
+	// EnvManagerRole took over, but "states:DescribeStateMachine" permissions
+	// weren't added until 1.12.2. Any version predating the feature's
+	// introduction or between v1.8.0 and 1.12.2 are invalid.
```
Lou1415926: Do you think we want to check this version only if people run `task run --generate-cmd app/env/job`? They

internal/pkg/cli/task_run.go
```diff
@@ -241,6 +245,17 @@ func newTaskRunOpts(vars runTaskVars) (*runTaskOpts, error) {
...
+	opts.envCompatibilityChecker = func() (versionCompatibilityChecker, error) {
+		envDescriber, err := describe.NewEnvDescriber(describe.NewEnvDescriberConfig{
+			App:         opts.appName,
+			Env:         opts.env,
+			ConfigStore: opts.store,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("new environment compatibility checker: %v", err)
+		}
+		return envDescriber, nil
```
Lou1415926: Does this initialize the client that we want? If they run `task run --generate app/env/job` - iirc - `opts.appName` and `opts.envName` would be empty. Do we need to pass the `app` and `env` in `app/env/job` instead to initialize the client?

internal/pkg/cli/task_run.go
```diff
@@ -501,6 +516,30 @@ func (o *runTaskOpts) confirmSecretsAccess() error {
...
+	}
+	version, err := envStack.Version()
+	if err != nil {
+		return fmt.Errorf("retrieve version of environment stack %q in application %q: %v", env, app, err)
+	}
+	// The '--generate-cmd' flag was introduced in env v1.4.0. In env v1.8.0,
+	// EnvManagerRole took over, but "states:DescribeStateMachine" permissions
+	// weren't added until 1.12.2. Any version predating the feature's
+	// introduction or between v1.8.0 and 1.12.2 are invalid.
+	if semver.Compare(version, "v1.4.0") < 0 || ((semver.Compare(version, "v1.7.0") > 0) && (semver.Compare(version, "v1.12.2") < 0)) {
```
Lou1415926:
```suggestion
	if semver.Compare(version, "v1.12.2") < 0 {
```

I feel like it's okay to simplify it like this. If the environment version is between 1.4.0 and 1.7.0 or before 1.4.0, people are still not able to run `task run --generate-cmd app/env/job` if they are on the latest Copilot CLI version, because the latest Copilot CLI version uses `EnvManagerRole` to fetch state machine information, and `EnvManagerRole` doesn't have the permission in any env version that predates v1.12.0 - doesn't matter if it's between 1.4.0 and 1.7.0 or not.

internal/pkg/cli/task_run.go
```diff
@@ -501,6 +516,30 @@ func (o *runTaskOpts) confirmSecretsAccess() error {
 	return nil
 }
 
+func (o *runTaskOpts) validateEnvCompatibility(app, env string) error {
```
Lou1415926: Given that the purpose of this function is only to validate the version when it's specifically job - 

```suggestion
func (o *runTaskOpts) validateEnvCompatibilityForGenerateJobCmd(app, env string) error {
```

Lou1415926 (commented):
Overall LGTM! Just two small questions!

internal/pkg/cli/task_run.go
huanjani: Don't we care about the people who _aren't_ on the latest Copilot CLI version? If they're using a version that doesn't use the `EnvManagerRole` and their env version is between 1.4.0 and 1.7.0, it should work and not error out. if they haven't touched anything, I don't think we should start erroring out on them. I think it's worth having the code be a little more hairy, no?

Lou1415926 (approved):
:shipit: !

internal/pkg/cli/task_run.go
```diff
@@ -195,11 +197,13 @@ func newTaskRunOpts(vars runTaskVars) (*runTaskOpts, error) {
 
 	prompter := prompt.New()
 	store := config.NewSSMStore(identity.New(defaultSess), ssm.New(defaultSess), aws.StringValue(defaultSess.Config.Region))
+	ws, err := workspace.New()
```
iamhopaul123: We miss the error handling here

internal/pkg/cli/task_run.go
```diff
@@ -501,6 +516,30 @@ func (o *runTaskOpts) confirmSecretsAccess() error {
...
+	}
+	version, err := envStack.Version()
+	if err != nil {
+		return fmt.Errorf("retrieve version of environment stack %q in application %q: %v", env, app, err)
+	}
+	// The '--generate-cmd' flag was introduced in env v1.4.0. In env v1.8.0,
+	// EnvManagerRole took over, but "states:DescribeStateMachine" permissions
+	// weren't added until 1.12.2. Any version predating the feature's
+	// introduction or between v1.8.0 and 1.12.2 are invalid.
+	if semver.Compare(version, "v1.12.2") < 0 {
```
iamhopaul123, line 526: Maybe make "v1.12.2" a constant and move the comment there as well?

internal/pkg/cli/task_run.go
```diff
@@ -501,6 +516,30 @@ func (o *runTaskOpts) confirmSecretsAccess() error {
...
+		return &errFeatureIncompatibleWithEnvironment{
+			ws:             o.ws,
```
iamhopaul123: It seems like `ws` here is to list all the local environments and then prompt users to upgrade them. However, `version` here is just one environment. One out-dated environment doesn't necessarily mean they need to upgrade the other environments 🤔 

Sorry this is more like a problem for `errFeatureIncompatibleWithEnvironment`. Since we'll use it I think it's worth bringing it up here.

internal/pkg/cli/task_run.go
huanjani: Good catch!

internal/pkg/cli/task_run.go
huanjani: I will simplify the comment here instead.

iamhopaul123 (approved):

Issue #3828, opened 2022-07-29
Still no FIFO SNS/SQS?


str11ngfello:
I saw a previous issue stating that SQS fifo changes were rolled back because SNS fifo wasn't ready. About a year ago now.
https://github.com/aws/copilot-cli/issues/2800

I've got heavy pubsub implemented with Copilot and, in fact, the pubsub feature is what drew me in. Now I need FIFO, lol. So I'm faced with:

- Create it all through addons 
- Fork Copilot and make the changes I need
- ?

Both are decent sized time investments for me and my small team that rely on Copilot for infrastructure creation.

Has anyone in the community got a recommendation for how they worked around this missing functionality in their project?

Lou1415926:
Hello @str11ngfello ! Sorry that we haven't been able to get to this feature yet, but thank you very much for resurfacing it! While some features are not currently on our short-term sprint board yet, requests like yours certainly help in prioritizing them!

I can try my best to suggest a workaround for you before we get to the feature. Do all of your SQS/SNS need to use fifo, or only some of them? Do you constantly need to update your services using `copilot svc deploy` or pipeline?

str11ngfello:
Hi @Lou1415926, yes all queues and topics will be FIFO. And, yes we deploy manually a couple times per day with 'copilot svc deploy'. We do not use code pipelines.

Thanks for the quick response!

Lou1415926:
Hello @str11ngfello ! I've experimented around myself - and to be honest, at this moment i can't find a great workaround that is easy and future-proof (such that when Copilot does support fifo, it guarantees a smooth transition😞.  My sincere apologies 🙇🏼‍♀️ ! 

> Other workarounds?

Whether a topic or a queue is FIFO or standard is currently immutable. This is why `aws-cli` or AWS console doesn't have a way yet to change a topic/queue type from standard to FIFO. CFN could do this it but under the hood it's really creating a new fifo topic/queue and delete the old standard one. Therefore, unfortunately I can't think of a third workaround yet that doesn’t go through copilot/CFN to convert the standard queue/topic to fifo.

>  Create it all through addons

[...]

PR #3994, opened 2022-09-12
feat: add support for FIFO SNS Topics 

paragbhingre:
This PR fix 1st part of #3828 


paragbhingre:
I am working on modifying the solution as we have decided to change the implementation. I will update the comment once the PR is ready to be reviewed

paragbhingre:
The PR is ready to be reviewed.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -795,10 +796,17 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
...
 	}
 	var publishers template.PublishOpts
+	var topicType string
 	// convert the topics to template Topics
 	for _, topic := range topics {
+		if topic.Type != nil {
+			topicType = aws.StringValue(topic.Type)
+		} else {
+			topicType = defaultTopicType
+		}
```
efekarakus: nit: Usually, in the presence of an `else` statement I try to find ways of getting rid of it as long as the operation is cheap:
```suggestion
		topicType := defaultTopicType
		if topic.Type != nil {
			topicType = aws.StringValue(topic.Type)
		}
```

internal/pkg/manifest/validate.go
```diff
@@ -1411,6 +1412,9 @@ func (p PublishConfig) validate() error {
 
 // validate returns nil if Topic is configured correctly.
 func (t Topic) validate() error {
+	if t.Type != nil && !contains(aws.StringValue(t.Type), validTopicsTypeValues) {
+		return fmt.Errorf(`the "Type" value must be one of %s`, english.WordSeries(validTopicsTypeValues, "or"))
```
```suggestion
		return fmt.Errorf(`"type" value %q is not allowed; must be one of %s`, aws.StringValue(t.Type), english.WordSeries(validTopicsTypeValues, "or"))
```

internal/pkg/manifest/validate.go
```diff
@@ -1634,9 +1638,9 @@ func validatePubSubName(name string) error {
 			missingField: "name",
 		}
 	}
-	// Name must contain letters, numbers, and can't use special characters besides underscores and hyphens.
+	// Name must contain letters, numbers, and can't use special characters besides underscores, and hyphens.
 	if !awsSNSTopicRegexp.MatchString(name) {
-		return fmt.Errorf(`"name" can only contain letters, numbers, underscores, and hypthens`)
+		return fmt.Errorf(`"name" can only contain letters, numbers, underscores, and hyphens`)
```
efekarakus, line 1643: 😂 ty for fixing this typo

internal/pkg/manifest/validate_test.go
```diff
@@ -565,6 +565,37 @@ func TestBackendService_validate(t *testing.T) {
...
+						Topics: []Topic{
+							{
+								Name: aws.String("mytopic"),
+								Type: aws.String("incorrectValue"),
+							},
+						},
+					},
+				},
+			},
+			wantedErrorMsgPrefix: `validate "publish": `,
```
efekarakus: Is there a way of writing a test case that validates the content of the new error message? maybe:
```suggestion
			wantedErrorMsgPrefix: `validate "publish": validate topics[0]: "type" value "incorrectValue" is not allowed`,
```

internal/pkg/manifest/workload.go
```diff
@@ -400,6 +400,7 @@ type PublishConfig struct {
 // Topic represents the configurable options for setting up a SNS Topic.
 type Topic struct {
 	Name *string `yaml:"name"`
+	Type *string `yaml:"type"`
```
efekarakus, line 403: What do you think of always setting a default value for `Type` here and removing the need of leaking "standard" to `stack/transformers.go`.

For example: https://go.dev/play/p/ADtA1hKgF7s

internal/pkg/template/templates/workloads/partials/cf/publish.yml
```diff
@@ -2,10 +2,17 @@
 {{- range $topic := .Publish.Topics}}
 {{logicalIDSafe $topic.Name}}SNSTopic:
   Metadata:
-    'aws:copilot:description': 'A SNS topic to broadcast {{$topic.Name}} events'
+    'aws:copilot:description':  {{- if eq $topic.Type "fifo" }} 'A FIFO SNS topic to broadcast {{$topic.Name}} events' {{- else}} 'A SNS topic to broadcast {{$topic.Name}} events'{{- end}}
```
efekarakus, line 5: Really nice!

efekarakus (commented):
Nice! I really appreciate that this PR increases our test coverage numbers

internal/pkg/template/templates/workloads/partials/cf/publish.yml
Lou1415926: +1 !

Lou1415926 (commented):
+1 to Efe's feedback but otherwise it LGTM!

internal/pkg/manifest/worker_svc.go
```diff
@@ -86,6 +86,19 @@ func (q *SQSQueueOrBool) IsEmpty() bool {
 	return q.Advanced.IsEmpty() && q.Enabled == nil
 }
 
+// UnmarshalYAML implements the yaml(v3) interface. Here it sets default value
+// of the type field if it is not set already.
+func (t *Topic) UnmarshalYAML(value *yaml.Node) error {
+	defaultTopicType := "standard"
```
efekarakus: nit: can we move this to a const at the top of the file?
```go
const defaultTopicType = "standard"

// func (t *Topic) UnmarshalYAML...
t.Type = aws.String(defaultTopicType)
```

efekarakus (approved):
Looks great! :shipit: just a tiny request, feel free to remove the label afterwards

internal/pkg/manifest/workload.go
paragbhingre: That's a great idea. I like it. I was contemplating where I should set the default value.

internal/pkg/manifest/validate.go
```diff
@@ -58,6 +58,7 @@ var (
 	httpProtocolVersions = []string{"GRPC", "HTTP1", "HTTP2"}
 
 	invalidTaskDefOverridePathRegexp = []string{`Family`, `ContainerDefinitions\[\d+\].Name`}
+	validTopicsTypeValues            = []string{"standard", "fifo"}
```
Lou1415926: tiny little nit: 

Since we constant-alize `standard`, we could as well do
```go
// in `manifest/worker_svc.go`
standardTopicType = "standard"
fifoTopicType = "fifo"
defaultTopicType = standardTopicType

// here
validatTopicsTypeValues = []string{standardTopicType, fifoTopicType}
```

internal/pkg/manifest/worker_svc.go
```diff
@@ -86,6 +87,18 @@ func (q *SQSQueueOrBool) IsEmpty() bool {
 	return q.Advanced.IsEmpty() && q.Enabled == nil
 }
 
+// UnmarshalYAML implements the yaml(v3) interface. Here it sets default value
+// of the type field if it is not set already.
+func (t *Topic) UnmarshalYAML(value *yaml.Node) error {
```
Lou1415926, line 94: Just to mention that we might want to be careful with setting default values in `UnmarshalYAML` in the future because it could easily mess up environment override, which takes advantage of `nil` pointers to determine if a value has override.

In this case we should be fine since `Topic` struct isn't stand-alone and is instead always in a list of `Topic`s - i.e. there is no `Topic`-level environment override.

Lou1415926 (approved):

internal/pkg/manifest/validate.go
paragbhingre: I don't think we need `defaultTopicType`yet in the validate. I will add other two constants though.

internal/pkg/manifest/worker_svc.go
paragbhingre: That's a great call out.

internal/pkg/manifest/validate.go
```diff
@@ -1411,6 +1412,9 @@ func (p PublishConfig) validate() error {
 
 // validate returns nil if Topic is configured correctly.
 func (t Topic) validate() error {
+	if t.Type != nil && !contains(aws.StringValue(t.Type), validTopicsTypeValues) {
```
iamhopaul123, line 1415:
```suggestion
	if !contains(aws.StringValue(t.Type), validTopicsTypeValues) {
```

internal/pkg/manifest/validate.go
```diff
@@ -1411,6 +1412,9 @@ func (p PublishConfig) validate() error {
...
+		return fmt.Errorf(`"type" value %q is not allowed; must be one of %s`, aws.StringValue(t.Type), english.WordSeries(validTopicsTypeValues, "or"))
```
iamhopaul123, line 1416: This sounds to be very like a general type of error. Can we make it a type? Like it happens when the value of a field is not in a valid list.

iamhopaul123 (commented):
Overall looks good! Awesome!

internal/pkg/manifest/validate.go
paragbhingre: Actually that check is required as we allow customers to not define type and we set default value to it. In that case type is `nil` and that is fine. But if it is not nil then we check if it has valid values or not. Does that make sense?

internal/pkg/manifest/validate.go
paragbhingre: Actually when the error occurs it looks something like `validate "topics[0]": "type" value "lifo" is not allowed; must be one of standard or fifo"` 
Does that look good to you?

paragbhingre: Just for the future we can build more generic type of error so that we can use it in multiple places. And I will cut the separate PR for it.

PR #3997, opened 2022-09-13
chore: Allow service manifests and other YAML manifests to be successfully marshalled without losing data

rclinard-amzn:
- Clarified the comment on Backend Service when it exposes no ports - previously, it suggested that no traffic (inbound or outbound) was possible, while in reality outbound traffic is still allowed.
- Added additional test cases for marshalling various components of service manifests
- Extended most test cases that test the marshalling to also run a round-trip marshal test - checking whether something can be marshalled and then immediately unmarshalled to get the same data back.
- Added the `omitempty` setting to most yaml tags so that nil/null values aren't placed in the resulting YAML
- Implemented `MarshalYAML` functions corresponding to every implementation of `UnmarshalYAML` so that marshalling & unmarshalling behavior is consistent
- Removed an unnecessary implementation of UnmarshalYAML on Image - it did validation that was already done elsewhere with Image.validate()
     - Works around a currently-unreported bug in go-yaml: if you have an inline struct that implements Unmarshaler and call yaml.Marshal on a struct containing that inline struct, the data of the inline struct will get silently dropped
- execute `template.New()` on demand instead of storing it as a field to allow easier unmarshalling of service manifests as YAML
- Remove some test cases for `MarshalBinary` implementations that only actually tested implementation details


internal/pkg/manifest/http_test.go
```diff
@@ -158,6 +159,13 @@ func TestAlias_UnmarshalYAML(t *testing.T) {
 				// check memberwise dereferenced pointer equality
 				require.Equal(t, tc.wantedStruct.StringSliceOrString, r.Alias.StringSliceOrString)
 				require.Equal(t, tc.wantedStruct.AdvancedAliases, r.Alias.AdvancedAliases)
+
+				roundtrip, err := yaml.Marshal(tc.wantedStruct)
+				require.NoError(t, err)
+				fmt.Printf("re-marshalled form:\n%s\n", string(roundtrip))
```
huanjani: Is this here intentionally?

internal/pkg/manifest/storage_test.go
```diff
@@ -103,6 +104,13 @@ efs:
 				require.Equal(t, tc.want.EFS.Advanced.AuthConfig, v.EFS.Advanced.AuthConfig)
 				require.Equal(t, tc.want.EFS.Advanced.UID, v.EFS.Advanced.UID)
 				require.Equal(t, tc.want.EFS.Advanced.GID, v.EFS.Advanced.GID)
+
+				roundtrip, err := yaml.Marshal(tc.want)
+				require.NoError(t, err)
+				fmt.Printf("re-marshalled form:\n%s\n", string(roundtrip))
```
huanjani: same as above. just wondering!

internal/pkg/manifest/http_test.go
rclinard-amzn: This is intentional but it's possible that this and the other occurrences could be changed... it's really helpful to have them though when the test fails and even when it doesn't it helps me catch a lot of stylistic issues in the yaml just from inspection!

One example of stylistic issues is `omitempty` being forgotten, it's harmless (so it won't cause a test failure) but results in big structs with all-null fields and such in the resulting yaml, and being able to catch that via visual inspection helps.

internal/pkg/addon/storage.go
```diff
@@ -41,13 +41,11 @@ var storageTemplateFunctions = map[string]interface{}{
...
 type DynamoDBTemplate struct {
 	DynamoDBProps
-
-	parser template.Parser
 }
 
 // MarshalBinary serializes the content of the template into binary.
 func (d *DynamoDBTemplate) MarshalBinary() ([]byte, error) {
-	content, err := d.parser.Parse(dynamoDbTemplatePath, *d, template.WithFuncs(storageTemplateFunctions))
+	content, err := template.New().Parse(dynamoDbTemplatePath, *d, template.WithFuncs(storageTemplateFunctions))
```
Lou1415926: > execute template.New() on demand instead of storing it as a field to allow easier unmarshalling of service manifests as YAML

I didn't quite get this one! Why does it make it easier to unmarshal?

In some cases we need it to be stored as a field (in `d.parser`) because we wants to be able to mock the `Parse` method in unit tests. There are some other cases in `cli` package where a client is stored as a field because it makes it easier/possible to mock. I think in this particular case here in `addons` package, this is part of the reason why it's stored.

internal/pkg/addon/storage_test.go
```diff
@@ -5,230 +5,12 @@
...
 	"fmt"
 	"testing"
 
-	"github.com/aws/copilot-cli/internal/pkg/template"
-	"github.com/aws/copilot-cli/internal/pkg/template/mocks"
-	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
 )
 
-func TestDynamoDBTemplate_MarshalBinary(t *testing.T) {
```
Lou1415926: Why do we not need these unit tests anymore? 😳

internal/pkg/manifest/env_test.go
```diff
@@ -522,6 +524,13 @@ http:
 			} else {
 				require.NoError(t, gotErr)
 				require.Equal(t, tc.wantedStruct, got)
+
+				roundtrip, err := yaml.Marshal(tc.wantedStruct)
```
Lou1415926, line 527: yay thank you for adding this to the unit tests!

internal/pkg/manifest/env_test.go
```diff
@@ -522,6 +524,13 @@ http:
...
+				require.NoError(t, err)
+				fmt.Printf("re-marshalled form:\n%s\n", string(roundtrip))
```
Lou1415926: We could in the future have unit test suites for `MarshalYAML` and compare the marshaled string there. For now I think this probably suffices for us to just visualize it.

internal/pkg/manifest/svc_test.go
```diff
@@ -225,6 +209,74 @@ environments:
 				require.Equal(t, wantedManifest, actualManifest)
 			},
 		},
+		"load balanced web service with taskdef overrides": {
```
Lou1415926, line 212: Thank you for separating it to a different unit test!

internal/pkg/manifest/workload_test.go
```diff
@@ -19,14 +18,43 @@ func TestImage_UnmarshalYAML(t *testing.T) {
 		inContent []byte
 
 		wantedError error
+		wantImage   Image
 	}{
-		"error if both build and location are set": {
```
Lou1415926, line 23: i might be missing something! Do we not need this test anymre?

internal/pkg/template/templates/workloads/services/backend/manifest.yml
```diff
@@ -8,7 +8,7 @@ type: {{.Type}}
 {{if .ImageConfig.Port}}
 # Your service is reachable at "http://{{.Name}}.${COPILOT_SERVICE_DISCOVERY_ENDPOINT}:{{.ImageConfig.Port}}" but is not public.
 {{- else}}
-# Your service does not allow any traffic.
+# Your service does not allow any incoming traffic.
```
Lou1415926, line 11: yay thanks!

iamhopaul123:
Same questions as @Lou1415926. Also could you explain why we need to add `MarshalYAML`?

internal/pkg/manifest/env_test.go
dannyrandall: could we just use `t.Logf()` instead so that the output isn't printed if the test passes? (if we keep this)

internal/pkg/addon/storage.go
rclinard-amzn: This was a typo - I meant to say marshalling! Basically, if you unmarshal and then marshal again, this un-exported field never gets set and then you get a crash. If it's only needed for mocking, then maybe we can keep the field but default to `template.New()` if it is unavailable?

internal/pkg/addon/storage_test.go
rclinard-amzn: These might have been the tests you were referring to in your previous comment - after scrutinizing their behavior, it looked like they were only testing whether `MarshalBinary` deferred to the `template.Parse()` implementation. Obviously mocking is sometimes needed, but a test that *only* tests for this one trivial implementation detail seemed like a code smell to me and that we should just rely on more of an integration test here.

If the Copilot team approaches this differently and tests like this should be kept, I can revert this in combination with the previous changes.

internal/pkg/manifest/env_test.go
rclinard-amzn: We can test that as well eventually - this catches most things by just checking for data loss in the round-trip.

internal/pkg/manifest/workload_test.go
rclinard-amzn: It's not necessary to check this when unmarshaling the YAML because we check it shortly afterwards in `Image.validate`! This test doesn't take that into account.

internal/pkg/manifest/env_test.go
rclinard-amzn: Good idea! I'll add that.

rclinard-amzn:
> Also could you explain why we need to add `MarshalYAML`?

A common pattern we have is:

```go
struct StructOrString {
    str *string
    struct *struct
}
```

Without MarshalYAML, this will get marshalled as something like

```yaml
structorstring:
    str: "test"
```

or:

```yaml
structorstring:
    struct:
        key: val
```

But, what we want is:

```yaml
structorstring: test
```

or:

```yaml
structorstring:
    key: val
```

`UnmarshalYAML` allows us to pick which field of `StructOrString` the YAML gets unmarshaled into. Similarly, `MarshalYAML` allows us to pick which field of `StructOrString` is written in place of the whole object.

internal/pkg/manifest/env_test.go
rclinard-amzn: Added!

internal/pkg/addon/storage_test.go
rclinard-amzn: These tests have been restored - see the latest changes :+1:

internal/pkg/addon/storage.go
rclinard-amzn: This has been implemented!

rclinard-amzn:
I've restored the removed tests & added the `parser` field back - but I've instead made it so that the `parser` field isn't required to be set, and `template.New()` will be used if it is not. Hopefully that resolves the concerns with this PR!

Lou1415926 (commented):
I think this PR LGTM! Thank you Ryan for the work 🪄 !

The only concern I have about merging it to the `mainline` branch is that the customized `MarshalYAML` logic is not used at all in mainline as of now. Once merged, it'd become dead code in the mainline branch until some day in the future when it becomes necessary - which we always try to avoid. 

What are the concerns that you have about merging it just into docker-compose branch? Let me know what you think! 🙇🏼‍♀️

rclinard-amzn:
Thanks for the feedback! Your comment is actually partially addressed in the follow up PR - my follow up PR does some substantial reworks to our service templates to incorporate the newly-added YAML marshaling code! For example, the healthcheck fields are now fully marshaled with `yaml.Marshal` and then stitched into our service manifests instead of being manually written out in the template. In addition, this code is being tested by our unit tests to make sure it is functional in this PR.

My main concern with merging this PR series into the docker compose branch is that this this PR series is fairly wide-reaching and touches a pretty core part of our codebase. It's very likely to cause conflicts in the future if it's not merged, and I want to make sure that we're testing this code separately since it changes quite a bit! I also want to make sure the docker compose work is focused on docker compose and not fixing dependencies in the rest of the Copilot codebase, since otherwise the diff between mainline and the docker compose branch could quickly become overwhelming!

I can still target this at the docker compose branch, but I think this PR series will be a nice cleanup of mainline to make sure that our `MarshalBinary` implementations degrade gracefully instead of silently dropping data if we ever tweak the default manifest values. Let me know what you think!

iamhopaul123:
Hello @rclinard-amzn. I see what you meant. Basically we want the same Marshal logic as how we customize YAML Unmarshal for some fields in the manifest. This makes sense to me.

In terms of whether we should put Marshal in the mainline branch or not. I think I'm inclined to put it into the docker compose branch for now because it won't be an issue for the mainline branch as we don't do Marshal logic there.

> since otherwise the diff between mainline and the docker compose branch could quickly become overwhelming!

I agree this is an issue so that we need to act fast and let Copilot customers try docker compose and see how they like it. And then merge those two branches. But I don't think having custom `Marshal` logic would help that much with this issue, as we can always quickly add them back for every one of them when we merge the branch. Before making the final decision on whether (because we'll see customers reaction) and how we can merge the branch, any code written to the mainline branch could end up being modified a lot in the future and add extra implementation complexity when we introduce new fields to the manifest.

efekarakus:
Closing the PR for now, we'll resurrect it as we need to introduce `MarshalYAML` for the manifest 😍

PR #3998, opened 2022-09-13
fix: custom resource upload should be namespaced by env

### LINKED ISSUE #3984: Pipeline for service fails if service is not already deployed to environment.
We have a pipeline running in a build account, it deploys to an environment in another account. If the service hasn't already been deployed to an account the cloudformation stage will fail with some variation of a S3 permissions error.

`Resource handler returned message: "Your access has been denied by S3, please make sure your request credentials have permission to GetObject for stackset-app-infrastruc-pipelinebuiltartifactbuc-xyz/manual/scripts/custom-resources/envcontrollerfunction/3ffcf03598029891816b7ce2d1ff14fdd8079af4406a0cfeff1d4aa0109dcd7d.zip. S3 Error Code: AccessDenied. S3 Error Message: Access Denied (Service: Lambda, Status Code: 403, Request ID: aae1a382-3a20-4345-80ca-ac160bc75175)" (RequestToken: ccbd59fc-caa1-dd84-e5a9-9b5e871070bb, HandlerErrorCode: AccessDenied)`

We are potentially going to have many apps, and many environments so at best this will be very inconvenient. Worse we restrict access to our prod account so manual deploys are not be possible.

iamhopaul123:
Fixes #3984

Screen Shot 2022-09-15 at 9 06 01 AM


internal/pkg/aws/s3/s3.go
```diff
@@ -56,28 +54,6 @@ func New(s *session.Session) *S3 {
 	}
 }
 
-// ZipAndUpload zips all files and uploads the zipped file to an S3 bucket under the specified key.
-// Per s3's recommendation https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html:
-// The bucket owner, in addition to the object owner, is granted full control.
-func (s *S3) ZipAndUpload(bucket, key string, files ...NamedBinary) (string, error) {
```
efekarakus, line 62: yay! thank you for deleting dead code

internal/pkg/cli/deploy/env.go
```diff
@@ -89,7 +89,7 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
 		env: in.Env,
 
 		templateFS:       template.New(),
-		s3:               s3.New(envRegionSession),
+		s3:               s3.New(envManagerSession),
```
efekarakus, line 92: I actually don't get why this issue is happening  if the object is uploaded with bucket ownership: 
https://github.com/aws/copilot-cli/blob/4b45d6682865598b368819023f15185b6d572b93/internal/pkg/aws/s3/s3.go#L217

and the bucket allows all actions from the env accounts:
https://github.com/aws/copilot-cli/blob/4b45d6682865598b368819023f15185b6d572b93/internal/pkg/template/templates/app/cf.yml#L74-L82

Then why is there an access denied error? What am I missing? I'd have guessed that as long as the objects are uploaded with BucketOwnership then it wouldn't matter if they're overriden.

efekarakus:
OK I finally get what is going on 

The buckets are created by default with [`BucketOwnership` set to `ObjectWriter`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-ownershipcontrolsrule.html).

Everything gets fixed if we change the `BucketOwnership` to `BucketOwnerEnforced` or `BucketOwnerPreferred` in the app template (we don't have to touch any of our `s3.go` code). 
In these scenarios, the access control to the data becomes based on the BucketPolicy instead of these confusing ACLs:
https://github.com/aws/copilot-cli/blob/35772b453f37f4b82a2a27c932a8d661f7d6ee15/internal/pkg/template/templates/app/cf.yml#L71-L82

My recommendation is instead of namespacing the custom resources, we should change the control ownership to be `BucketOwnerEnforced` and force the users that need the fix to run `copilot app upgrade`.

internal/pkg/deploy/upload/customresource/customresource.go
```diff
@@ -84,15 +83,15 @@ func (cr *CustomResource) init() error {
 	for _, file := range cr.files {
 		f, err := w.Create(file.name)
 		if err != nil {
-			return fmt.Errorf("create zip file %q for custom resource %q: %v", file.name, cr.FunctionName(), err)
+			return fmt.Errorf("create zip file %q for custom resource %q: %v", file.name, cr.name, err)
```
efekarakus: nit: can we use `.Name()` instead to rely on the public interface of the type rather than the internal data structure

efekarakus (commented):
Should we update the `TemplateVersion: 'v1.1.0'` field in the app template and `deploy.LatestAppTemplateVersion` to `v1.2.0` ?

internal/pkg/deploy/upload/customresource/customresource.go
iamhopaul123: > Should we update the TemplateVersion: 'v1.1.0' field in the app template and deploy.LatestAppTemplateVersion to v1.2.0 ?

No we don't because it is not released yet.

internal/pkg/aws/s3/s3.go
```diff
@@ -56,28 +54,6 @@ func New(s *session.Session) *S3 {
 	}
 }
 
-// ZipAndUpload zips all files and uploads the zipped file to an S3 bucket under the specified key.
```
Lou1415926, line 59: :rip

internal/pkg/cli/deploy/env.go
Lou1415926: thanks for fixing the session here!

internal/pkg/cli/deploy/svc.go
```diff
@@ -208,9 +207,6 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 	if err != nil {
 		return nil, fmt.Errorf("create default: %w", err)
 	}
-	if err != nil {
```
Lou1415926, line 211: :doge

Lou1415926 (approved):

efekarakus (approved):

PR #4002, opened 2022-09-15
chore: remove unused config for s3 upload

iamhopaul123:
`ObjectCannedACLBucketOwnerFullControl` is useless because ACL has already been disabled by https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/app/cf.yml#L94


efekarakus:
I was slightly worried about disabling it in case folks don't upgrade the app template first

iamhopaul123:
> I was slightly worried about disabling it in case folks don't upgrade the app template first

ooh that's a good call.

PR #4003, opened 2022-09-15
feat: add `http.redirect_to_https` field to service manifest

dannyrandall:
adds the ability to disable the ALB redirect from http -> https when using a custom domain.


internal/pkg/cli/deploy/svc.go
```diff
@@ -1360,6 +1360,9 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
 
 func (d *lbWebSvcDeployer) validateALBRuntime() error {
 	hasImportedCerts := len(d.envConfig.HTTPConfig.Public.Certificates) != 0
+	if d.lbMft.RoutingRule.Redirect != nil && *d.lbMft.RoutingRule.Redirect && d.app.Domain == "" && !hasImportedCerts {
```
paragbhingre: What is the difference between 1st two conditions?

`d.lbMft.RoutingRule.Redirect != nil && *d.lbMft.RoutingRule.Redirect`

internal/pkg/cli/deploy/svc.go
```diff
@@ -1360,6 +1360,9 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
...
+		return fmt.Errorf("cannot redirect to https without using a custom domain")
```
paragbhingre: How about?
```suggestion
		return fmt.Errorf(`validate "redirect": cannot redirect to https without using a custom domain`)
```

internal/pkg/cli/deploy/svc.go
dannyrandall: `RoutingRule.Redirect` is `*bool`, so it's basically saying "if redirect is configured and set to true". If we didn't have the first condition then the second would panic if `Redirect` was not configured.

internal/pkg/cli/deploy/svc_test.go
```diff
@@ -957,6 +993,58 @@ func TestWorkloadDeployer_DeployWorkload(t *testing.T) {
 				m.mockServiceDeployer.EXPECT().DeployService(gomock.Any(), "mockBucket", gomock.Any()).Return(nil)
 			},
 		},
+		"success with http redirect disabled and certs imported": {
```
paragbhingre, line 996: tests ❤️

internal/pkg/cli/deploy/svc.go
iamhopaul123:
```suggestion
	if aws.BoolValue(d.lbMft.RoutingRule.Redirect) && d.app.Domain == "" && !hasImportedCerts {
```

internal/pkg/cli/deploy/svc.go
```diff
@@ -1360,6 +1360,9 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
 
 func (d *lbWebSvcDeployer) validateALBRuntime() error {
 	hasImportedCerts := len(d.envConfig.HTTPConfig.Public.Certificates) != 0
+	if d.lbMft.RoutingRule.Redirect != nil && *d.lbMft.RoutingRule.Redirect && d.app.Domain == "" && !hasImportedCerts {
+		return fmt.Errorf(`validate "redirect": cannot redirect to https without using a custom domain`)
```
iamhopaul123:
```suggestion
		return fmt.Errorf(`cannot redirect http to https without having a domain associated with the app %q or importing any certificates in env %q`)
```

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -33,7 +33,8 @@ type loadBalancedWebSvcReadParser interface {
 type LoadBalancedWebService struct {
 	*ecsWkld
 	manifest               *manifest.LoadBalancedWebService
-	httpsEnabled           bool
+	customDomain           bool
+	httpsRedirect          bool
```
iamhopaul123: Why do we need this? Can't we use the manifest field? We only set `HTTPSRedirect` false if it is set `False` in the manifest.

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -66,7 +67,7 @@ type LoadBalancedWebServiceConfig struct {
 func NewLoadBalancedWebService(conf LoadBalancedWebServiceConfig,
 	opts ...LoadBalancedWebServiceOption) (*LoadBalancedWebService, error) {
 	parser := template.New()
-	var dnsDelegationEnabled, httpsEnabled bool
+	var dnsDelegationEnabled, customDomain bool
```
iamhopaul123: This is not customDomain, but strictly means https is enabled...

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -104,6 +105,7 @@ func NewLoadBalancedWebService(conf LoadBalancedWebServiceConfig,
 		},
 		manifest:             conf.Manifest,
 		httpsEnabled:         httpsEnabled,
+		httpRedirect:         httpsEnabled && (conf.Manifest.RoutingRule.Redirect == nil || *conf.Manifest.RoutingRule.Redirect),
```
iamhopaul123: maybe consider making `(conf.Manifest.RoutingRule.Redirect == nil || *conf.Manifest.RoutingRule.Redirect)` this a method like `redirectEnabled`

dannyrandall: changed the name of the field to `disable_redirect` which i think should help with some confusion!

internal/pkg/cli/deploy/svc.go
```diff
@@ -1360,6 +1360,9 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
 
 func (d *lbWebSvcDeployer) validateALBRuntime() error {
 	hasImportedCerts := len(d.envConfig.HTTPConfig.Public.Certificates) != 0
+	if aws.BoolValue(d.lbMft.RoutingRule.DisableRedirect) && d.app.Domain == "" && !hasImportedCerts {
+		return fmt.Errorf(`validate "redirect": cannot disable http to https redirect without having a domain associated with the app %q or importing any certificates in env %q`, d.app.Name, d.env.Name)
```
iamhopaul123:
```suggestion
		return fmt.Errorf(`cannot disable http to https redirect without having a domain associated with the app %q or importing any certificates in env %q`, d.app.Name, d.env.Name)
```

iamhopaul123 (approved):

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
```diff
@@ -85,12 +86,24 @@ HTTPListenerRuleWithDomain:
           Path: "/#{path}"
           Query: "#{query}"
           StatusCode: HTTP_301
+      {{- else}}
+      - TargetGroupArn: !Ref TargetGroup
+        Type: forward
+      {{- end}}
     Conditions:
-{{- if .Aliases }}
+      {{- if .AllowedSourceIps}}
```
efekarakus, line 94: oh is this a bug fix?

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -152,6 +152,7 @@ func (s *BackendService) Template() (string, error) {
 		Secrets:                  convertSecrets(s.manifest.BackendServiceConfig.Secrets),
 		Aliases:                  aliases,
 		HTTPSListener:            s.httpsEnabled,
+		HTTPRedirect:             s.httpsEnabled,
```
efekarakus, line 155: gotcha, so we're only supporting this field for now for LBWS as there is no need of a cdn for internal ALBs 👍

efekarakus (approved):

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
dannyrandall: "bug" fix, yeah :)

PR #4004, opened 2022-09-15
chore: add PB policy to IAM roles generated/updated by `task run`, `env package`, `env deploy`

huanjani:
Related: #3969, #2986.

This adds the permissions boundary policy stored in app config to CFN templates to form the PermissionsBoundary field for IAM roles generated for `env deploy`, `env package`, and `task run`. (Note that if a user is running their task in the default or other cluster and not in an app, then the permissions boundary policy will not apply. We could possibly pass the BP in with a flag as a future milestone.)

(Disregard most of the commit messages-- I forgot to switch to my new branch 🙈.)


internal/pkg/cli/env_init.go
```diff
@@ -285,7 +285,7 @@ func (o *initEnvOpts) Execute() error {
 	// If the call fails because the role already exists, nothing to do.
 	// If the call fails because the user doesn't have permissions, then the role must be created outside of Copilot.
 	_ = o.iam.CreateECSServiceLinkedRole()
-	
+
```
Lou1415926, line 288: I thought we'd accept 
```
$ copilot app init --permissions-boundary A
$ copilot env init --permissions-boundary B
```
and we'd use boundary policy B - instead of A - for all the roles within the environment.

Is this going to be in an upcoming PR, or did I misunderstand?

Lou1415926: Also I wonder what'd be the story for `copilot init`? Do we want to accept this flag there?

internal/pkg/cli/task_run.go
```diff
@@ -954,6 +954,15 @@ func (o *runTaskOpts) deploy() error {
 		deployOpts = []awscloudformation.StackOption{awscloudformation.WithRoleARN(o.targetEnvironment.ExecutionRoleARN)}
 	}
 
+	var boundaryPolicy string
+	if o.appName != "" {
+		app, err := o.store.GetApplication(o.appName)
```
Lou1415926, line 959: nit: what do you think of caching the `app config`  [here](https://github.com/aws/copilot-cli/blob/a6ae6784bf9a46cecf7697e4e33857ba5c3370a7/internal/pkg/cli/task_run.go#L1049) and try to reuse the cached version here?

internal/pkg/cli/env_package_test.go
```diff
@@ -280,6 +281,7 @@ func TestPackageEnvOpts_Execute(t *testing.T) {
 					},
 					fs:     fs,
 					envCfg: &config.Environment{Name: "test"},
+					appCfg: &config.Application{},
```
Lou1415926: nit: perhaps we could augment the unit test by having
```
appCfg: &config.Application{
  permissionsBoundary: "mockPermissionsBoundary",
},
```

and then have in line 262
```
deployer.EXPECT().GenerateCloudFormationTemplate(gomock.Any()).Return(&deploy.GenerateCloudFormationTemplateOutput{
					Template:   "template",
					Parameters: "parameters",
					PermissionsBoundary: "mockPermissionsBoundary",
				}, nil)
```

internal/pkg/cli/env_init.go
efekarakus: > and we'd use boundary policy B - instead of A - for all the roles within the environment.
>
> Is this going to be in an upcoming PR, or did I misunderstand?

We won't do this until we hear a customer request :) 

> Also I wonder what'd be the story for copilot init? Do we want to accept this flag there?

My opinion is that this feature is for L+ organizations and doesn't fit into `init` which is more for just getting started.

efekarakus (commented):
Looks great! just a +1 to Wanxian's feedbac

internal/pkg/cli/task_run.go
huanjani: Yeah, I thought of that, but then we won't have it if they use the selector, not the flag to indicate the app name.

internal/pkg/cli/env_package_test.go
huanjani: done.

Lou1415926 (approved):

efekarakus (approved):

PR #4014, opened 2022-09-20
feat: add support for SQS FIFO Queue

### LINKED ISSUE #2800: Allow worker pattern to create FIFO queue
Hey there, thanks for the great feature for worker pattern implemented in https://github.com/aws/copilot-cli/issues/2550.

It would be nice if we could also precise as parameter in the manifest if a FIFO queue is needed.

Something like:
```
subscribe:
  topics:
    - name: jobs
      service: api
      queue:
        fifo: true
```

Thanks!

### LINKED ISSUE #3828: Still no FIFO SNS/SQS?

paragbhingre:
This PR resolves #3828 , resolves #2800


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -25,8 +25,13 @@ import (
...
 const (
-	enabled  = "ENABLED"
-	disabled = "DISABLED"
+	enabled           = "ENABLED"
+	disabled          = "DISABLED"
+	standardQueueType = "standard"
+	fifoQueueType     = "fifo"
+	defaultQueueType  = standardQueueType
+	messageGroup      = "messageGroup"
+	perMessageGroupId = "perMessageGroupId"
```
efekarakus: nit: what do you think of starting the constants with the service name that it applies to? and separating the `const` blocks:
```suggestion
	efsEnabled           = "ENABLED"
	efsDisabled          = "DISABLED"
	
	sqsStandardQueueType = "standard"
	sqsFIFOQueueType     = "fifo"
	sqsDefaultQueueType  = standardQueueType
	sqsDedupeScopeMessageGroup      = "messageGroup"
	sqsFIFOThroughputLimitPerMessageGroupId = "perMessageGroupId"
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -835,16 +840,20 @@ func convertTopicSubscription(t manifest.TopicSubscription) (
...
-			Queue:        &template.SQSQueue{},
+			Name:    t.Name,
+			Service: t.Service,
+			Queue: &template.SQSQueue{
+				Type: aws.String(defaultQueueType),
+			},
 			FilterPolicy: filterPolicy,
 		}, nil
 	}
+	queueAdvancedConfig := convertQueue(t.Queue.Advanced)
```
efekarakus: can we revert and delete the variable? I feel like it's fine to extract a variable when the name is going to clarify some obscure method (which isn't the case here 💭 )

https://refactoring.com/catalog/inlineVariable.html

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -864,11 +873,39 @@ func convertQueue(q manifest.SQSQueue) *template.SQSQueue {
 	if q.IsEmpty() {
 		return nil
 	}
+
+	if strings.Compare(aws.StringValue(q.Type), standardQueueType) == 0 {
```
efekarakus:
```suggestion
	if aws.StringValue(q.Type) == standardQueueType {
```

paragbhingre:
Integration and e2e tests are coming in.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -864,11 +873,39 @@ func convertQueue(q manifest.SQSQueue) *template.SQSQueue {
 	if q.IsEmpty() {
 		return nil
 	}
+
```
efekarakus, line 873: A slightly different way of writing this function body to reduce nesting would be:
```go
// Step 1. set the common default across all if-statements.
out := &template.SQSQueue{ 
	Retention:  convertRetention(in.Retention),
	Delay:      convertDelay(in.Delay),
	Timeout:    convertTimeout(in.Timeout),
	DeadLetter: convertDeadLetter(in.DeadLetter),
	Type:       in.Type,
}

// Step 2. Handle cases that exist with the baseline.
if aws.StringValue(in.Type) == standardQueueType {
	return out
}

// Step 3. Augment for exception.
out.ContentBasedDeduplication = in.ContentBasedDeduplication
out.FIFOThroughputLimit = in.FifoThroughputLimit
out.DeduplicationScope = in.DeduplicationScope
if aws.BoolValue(q.HighThroughputFifo) {
	out.FIFOThroughputLimit = aws.String(perMessageGroupId)
	out.DeduplicationScope = aws.String(messageGroup)
}
return out
```

internal/pkg/manifest/validate.go
```diff
@@ -1455,7 +1457,20 @@ func (q SQSQueueOrBool) validate() error {
 	if q.IsEmpty() {
 		return nil
 	}
-	return q.Advanced.validate()
+
+	if q.Enabled != nil {
+		return nil
+	} else {
```
efekarakus: no need for an else-statement after a return

```
if q.Enabled != nil {
   return nil
}
if q.isFIFO() {
  return q.Advanced.validateFIFO()
}
if q.hasFIFOFieldsEnabled() {
   return fmt.Errorf(...)
}
```

internal/pkg/manifest/validate.go
```diff
@@ -1455,7 +1457,20 @@ func (q SQSQueueOrBool) validate() error {
...
+		if q.Advanced.Type != nil && strings.Compare(aws.StringValue(q.Advanced.Type), "fifo") == 0 {
+			return q.Advanced.validateFIFO()
+		} else {
+			if q.Advanced.FifoThroughputLimit != nil || q.Advanced.HighThroughputFifo != nil ||
```
efekarakus: Consider decomposing complex conditionals to a separate method to make the body easier to parse: 
```go
func (q SQSQueue) hasFIFOFieldsEnabled() bool {
   return q.Advanced.FifoThroughputLimit != nil || q.Advanced.HighThroughputFifo != nil // || ...
}
```


https://refactoring.com/catalog/decomposeConditional.html

internal/pkg/manifest/validate.go
```diff
@@ -1455,7 +1457,20 @@ func (q SQSQueueOrBool) validate() error {
 	if q.IsEmpty() {
 		return nil
 	}
-	return q.Advanced.validate()
+
+	if q.Enabled != nil {
+		return nil
+	} else {
+		if q.Advanced.Type != nil && strings.Compare(aws.StringValue(q.Advanced.Type), "fifo") == 0 {
```
efekarakus: Why are we using `strings.Compare` instead of `==` 😛 ?

internal/pkg/manifest/validate.go
```diff
@@ -1469,6 +1484,30 @@ func (q SQSQueue) validate() error {
...
 }
 
+// validateFIFO returns nil if FIFO SQSQueue is configured correctly.
+func (q SQSQueue) validateFIFO() error {
+	if q.IsEmpty() {
+		return nil
+	}
+
+	if q.HighThroughputFifo != nil && (q.FifoThroughputLimit != nil || q.DeduplicationScope != nil) {
+		return fmt.Errorf(`validate "high_throughput_fifo": cannot define fifo_throughput_limit and fifo_throughput_limit when high_throughput_fifo is defined`)
```
```suggestion
		return fmt.Errorf(`validate "high_throughput_fifo": cannot define "fifo_throughput_limit" and "fifo_throughput_limit" when "high_throughput_fifo" is defined`)
```

efekarakus: Should we accept the values of `messageGroup` and `perMessageGroupId` even if the user sets `high_throughput_fifo` to `true`?

internal/pkg/manifest/validate.go
```diff
@@ -1469,6 +1484,30 @@ func (q SQSQueue) validate() error {
...
+	}
+	if q.DeduplicationScope != nil && !contains(aws.StringValue(q.DeduplicationScope), validDeduplicationScopeValues) {
+		return fmt.Errorf(`validate "deduplication_scope": deduplication scope value must be one of %s`, english.WordSeries(validDeduplicationScopeValues, "or"))
+	}
+	if q.FifoThroughputLimit != nil && !contains(aws.StringValue(q.FifoThroughputLimit), validFIFOThroughputLimitValues) {
+		return fmt.Errorf(`validate "fifo_throughput_limit": fifo throughput limit value must be one of %s`, english.WordSeries(validFIFOThroughputLimitValues, "or"))
+	}
+	if q.FifoThroughputLimit != nil && strings.Compare(aws.StringValue(q.FifoThroughputLimit), "perMessageGroupId") == 0 && q.DeduplicationScope != nil && strings.Compare(aws.StringValue(q.DeduplicationScope), "queue") == 0 {
+		return fmt.Errorf(`when deduplication_scope scope is set to queue, fifo_throughput_limit must be set to perQueue`)
```
```suggestion
	if aws.StringValue(q.DeduplicationScope) == "queue" && aws.StringValue(q.FifoThroughputLimit) == "perMessageGroupId" {
		return errors.New(`"fifo_throughput_limit" must be set to "perQueue" when "deduplication_scope" is set to "queue"`)
```

internal/pkg/manifest/worker_svc.go
```diff
@@ -110,16 +111,22 @@ func (q *SQSQueueOrBool) UnmarshalYAML(value *yaml.Node) error {
...
-	DeadLetter DeadLetterQueue `yaml:"dead_letter"`
+	Retention                 *time.Duration  `yaml:"retention"`
+	Delay                     *time.Duration  `yaml:"delay"`
+	Timeout                   *time.Duration  `yaml:"timeout"`
+	DeadLetter                DeadLetterQueue `yaml:"dead_letter"`
+	ContentBasedDeduplication *bool           `yaml:"content_based_deduplication"`
+	DeduplicationScope        *string         `yaml:"deduplication_scope"`
+	FifoThroughputLimit       *string         `yaml:"fifo_throughput_limit"`
+	HighThroughputFifo        *bool           `yaml:"high_throughput_fifo"`
+	Type                      *string         `yaml:"type"`
```
efekarakus: Was there a reason why we didnt' want to group all fifo related groups together?

Here are some sample manifests:
```yaml
# option 1. standard queue
queue:
  delay: 10s
  
# option 2. fifo queue
queue:
  delay: 10s
  fifo: true

# Option 3. Advanced FIFO settings
queue:
  delay: 10s
  fifo:
     content_based_deduplication: true
     high_throughput: true
```

efekarakus: Can we remove any of these fields here from the user to reduce complexity? 🤔 

For example, can we get rid of `DeduplicationScope` and `FifoThroughputLimit` if we have `high_throughput: true`?
We know that these fields must be set to respectively `messageGroup` and `perMessageGroupId` so I wonder why anyone would need to tweak them if  we already have `high_throughput` 💭

internal/pkg/manifest/worker_svc.go
```diff
@@ -195,6 +202,29 @@ func (s *WorkerService) Subscriptions() []TopicSubscription {
 	return s.Subscribe.Topics
 }
 
+// RetrofitFIFOConfig appends ".fifo" suffix to fifo topics, also adds values to type field.
+func (s *WorkerService) RetrofitFIFOConfig() {
```
efekarakus: Could we do all of this retrofiting work inside `UnmarshalYAML` so that we don't have to call separate `RetrofitFIFOConfig` method

```go
func (sub *TopicSubscription) UnmarshalYAML() error {
  // Set automatically the .fifo suffix to the topic name.
}

func (q *SQSQueue) UnmarshalYAML() error {
  // Set the queue type based on the user values.
}
```

internal/pkg/manifest/workload.go
```diff
@@ -22,6 +22,13 @@ import (
 
 const (
 	defaultDockerfileName = "Dockerfile"
+	perMessageGroupID     = "perMessageGroupId"
```
efekarakus: same feedback with const prefixes here 🥺

internal/pkg/template/template_functions.go
```diff
@@ -65,6 +65,9 @@ func EnvVarSecretFunc(s string) string {
 	return StripNonAlphaNumFunc(s) + "Secret"
 }
 
+// Deref dereferences the pointer.
+func Deref(i *string) string { return *i }
```
efekarakus: Instead of adding `Deref` which only works for `*string` what do you think of adding a method to the `template.SQSQueue` type:
```go
func (q SQSQueue) IsFIFO() bool
```
which can be used in the template:
```go
{{- if .Susbcribe.Queue.IsFIFO }}
```

internal/pkg/template/templates/workloads/partials/cf/subscribe.yml
```diff
@@ -80,16 +80,31 @@ EventsQueue:
...
+    {{- if .Subscribe.Queue.DeduplicationScope }}
+    DeduplicationScope: {{.Subscribe.Queue.DeduplicationScope}}
+    {{- end }}
+    {{- if .Subscribe.Queue.ContentBasedDeduplication }}
+    ContentBasedDeduplication: {{.Subscribe.Queue.ContentBasedDeduplication}}
+    {{- end}}
   {{- end}}
 
 {{- if .Subscribe.Queue}}{{- if .Subscribe.Queue.DeadLetter}}
 DeadLetterQueue:
```
efekarakus, line 100: huh TIL I didn't know that the dead letter queue of a FIFO queue must also be FIFO:
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html

This is a really nice catch 👍

internal/pkg/template/templates/workloads/partials/cf/subscribe.yml
```diff
@@ -228,5 +261,7 @@ QueuePolicy:
           Resource: !GetAtt {{logicalIDSafe $topic.Service}}{{logicalIDSafe $topic.Name}}EventsQueue.Arn
           Condition:
             ArnEquals:
-              aws:SourceArn: !Join ['', [!Sub 'arn:${AWS::Partition}:sns:${AWS::Region}:${AWS::AccountId}:', !Ref AppName, '-', !Ref EnvName, '-{{$topic.Service}}-{{logicalIDSafe $topic.Name}}']]
-{{- end}}{{- end}}{{- end}}
\ No newline at end of file
+              aws:SourceArn: {{ if eq (deref $topic.Queue.Type) "fifo"}} !Join ['', [!Sub 'arn:${AWS::Partition}:sns:${AWS::Region}:${AWS::AccountId}:', !Ref AppName, '-', !Ref EnvName, '-{{$topic.Service}}-{{$topic.Name}}']] {{ else }} !Join ['', [!Sub 'arn:${AWS::Partition}:sns:${AWS::Region}:${AWS::AccountId}:', !Ref AppName, '-', !Ref EnvName, '-{{$topic.Service}}-{{logicalIDSafe $topic.Name}}']] {{ end }}
```
efekarakus: oh I see, we don't want to remove the "." in case it's fifo 💭

internal/pkg/manifest/validate.go
paragbhingre: We can accept the values, but then we need to again have one more condition to prioritize high_throughput_fifo over those 2 fields. So it's better error out saying you can either set high_throughput_fifo or those 2 fields.

internal/pkg/template/workload.go
```diff
@@ -424,10 +424,14 @@ type TopicSubscription struct {
...
-	Timeout    *int64
-	DeadLetter *DeadLetterQueue
+	Retention                 *int64
+	Delay                     *int64
+	Timeout                   *int64
+	DeadLetter                *DeadLetterQueue
+	FifoThroughputLimit       *string
+	ContentBasedDeduplication *bool
+	DeduplicationScope        *string
+	Type                      *string
```
Lou1415926: Could this be `Type string`, without the pointer type?

I think since `Type` is either `standard` or `fifo`, it's arbitrary what a `nil` type means here. From the implementation, I can see that `nil` type means `standard`, so we might as well just give it a `string` value of `standard` instead of letting it be a `string pointer` of `nil`.

This also has the perk of getting rid of `Deref` 😂

internal/pkg/template/templates/workloads/partials/cf/subscribe.yml
paragbhingre: To add to your point, I was wondering if we should be allowing customers to set fifo related configs to Deadletter queue as well? It will be a lot of configurations that we will be exposing to the customer at once. I think we can add those configs if customers ask for it. What do you think about it?

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: You are right, actually previously I was doing some operation over here after getting response from the convertQueue method that is why I had this parameter. But now I am doing that work on RetrofitFifoConfig method. So we don't need this parameter anymore.

internal/pkg/manifest/validate.go
Lou1415926: Is this about being case-insensitive?  Perhaps `strings.EqualFold` could do a better job than `compare`  ?

internal/pkg/manifest/validate.go
Lou1415926: perhaps we could reuse this typed error? https://github.com/aws/copilot-cli/blob/6132cb54fb1167b297bce0a218026760cfbae0a0/internal/pkg/manifest/errors.go#L94

internal/pkg/manifest/worker_svc.go
Lou1415926: For each subscription, to determine whether it's `fifo` we need two pieces of information:
1. The type of the subscription's dedicated `queue`
2. The type of the default `queue`

We don't need 2 if we have 1.

So `UnmarshalYAML` might not work here 🤔  Because inside of 
```
func (sub *TopicSubscription) UnmarshalYAML() error { // Set automatically the .fifo suffix to the topic name. }
```
we can't know 2.

I'm thinking an alternative option is that we can have 
```
func (cfg *SubscribeConfig) Topics() []*TopicSubscription {
    // range over `cfg.Topics`, appending `.fifo` to `cfg.Topics[idx].Name` given `cfg.Topics[idx].Queue` or `cfg.Queue`
}

```

and then call `Topics` in `stack/transfomer.go`

paragbhingre: The only reason why we can not call this `Topics` in `stack/transfomer.go` is because there is a [validateTopicsExist](https://github.com/aws/copilot-cli/blob/4b82a0943e24c1db4c4f90e6a74b41928cb334e3/internal/pkg/cli/deploy/svc.go#L1177) function that checks for valid topics and if we do not retrofit before that point then that call fails for fifo topics.

internal/pkg/manifest/validate.go
paragbhingre: I have added new func that checks weather `type` has correct value or not so we are good here. `==` will also works as we only allow small case `fifo` and `standard` as type values.

internal/pkg/manifest/worker_svc.go
Lou1415926: Then can we call it in `validateTopicsExist` 🤔  ` validateTopicsExist(d.wsMft.Subscriptions(), topicARNs, d.app.Name, d.env.Name`. We could maybe just modify the implementation of `Subscriptions`?

internal/pkg/template/workload.go
paragbhingre: our new implementation doesn't need deref at all so I have removed it from the code.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -29,6 +29,12 @@ const (
 	disabled = "DISABLED"
 )
 
+// SQS Queue configs
```
```suggestion
// SQS Queue field values.
```
https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -865,12 +870,49 @@ func convertQueue(q manifest.SQSQueue) *template.SQSQueue {
...
+
+	if q.FIFO.IsEmpty() {
+		return queue
+	}
+
+	if aws.BoolValue(q.FIFO.Enable) {
+		queue.FIFOQueueConfig = &template.FIFOQueueConfig{
+			Enable: q.FIFO.Enable,
+		}
+		return queue
```
efekarakus: Can we do like SNS and not have `Enable` as field in the `template` type:
```suggestion
	queue.FIFOQueueConfig = &template.FIFOQueueConfig{}
	if q.FIFO.IsEnabled() && q.FIFO.Advanced.IsEmpty() {
		return queue
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -865,12 +870,49 @@ func convertQueue(q manifest.SQSQueue) *template.SQSQueue {
...
+	}
+
+	if !q.FIFO.Advanced.IsEmpty() {
+		queue.FIFOQueueConfig = &template.FIFOQueueConfig{
+			Advanced: &template.FIFOAdvanceConfig{
+				ContentBasedDeduplication: nil,
+				DeduplicationScope:        nil,
+				FIFOThroughputLimit:       nil,
```
efekarakus: no need to define the fields with `nil` as that's the zero-value

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -865,12 +870,49 @@ func convertQueue(q manifest.SQSQueue) *template.SQSQueue {
...
-	return &template.SQSQueue{
+
+	queue := &template.SQSQueue{
 		Retention:  convertRetention(q.Retention),
 		Delay:      convertDelay(q.Delay),
 		Timeout:    convertTimeout(q.Timeout),
 		DeadLetter: convertDeadLetter(q.DeadLetter),
 	}
+
+	if q.FIFO.IsEmpty() {
```
efekarakus: nit: can we rename the parameter from `q` to `in` for "input" it's confusing reading about `queue` vs `q` 😂

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -865,12 +870,49 @@ func convertQueue(q manifest.SQSQueue) *template.SQSQueue {
...
+		}
+		if q.FIFO.Advanced.ContentBasedDeduplication != nil {
+			queue.FIFOQueueConfig.Advanced.ContentBasedDeduplication = q.FIFO.Advanced.ContentBasedDeduplication
+		}
+		if q.FIFO.Advanced.DeduplicationScope != nil {
+			queue.FIFOQueueConfig.Advanced.DeduplicationScope = q.FIFO.Advanced.DeduplicationScope
+		}
+		if q.FIFO.Advanced.FIFOThroughputLimit != nil {
+			queue.FIFOQueueConfig.Advanced.FIFOThroughputLimit = q.FIFO.Advanced.FIFOThroughputLimit
+		}
```
efekarakus: no need for the if-checks:

```go
queue.FIFOQueueConfig = &template.FIFOQueueConfig {
   ContentBasedDeduplication: q.FIFO.Advanced.ContentBasedDeduplication,
   DeduplicationScope: q.FIFO.Advanced.DeduplicationScope,
   FIFOThroughputLimit: q.FIFO.Advanced.FIFOThroughputLimit,
}
```

internal/pkg/manifest/validate.go
```diff
@@ -58,7 +58,8 @@ var (
 	httpProtocolVersions = []string{"GRPC", "HTTP1", "HTTP2"}
 
 	invalidTaskDefOverridePathRegexp = []string{`Family`, `ContainerDefinitions\[\d+\].Name`}
-	validTopicsTypeValues            = []string{standardTopicType, fifoTopicType}
+	validDeduplicationScopeValues    = []string{sqsFifoMessageGroup, sqsFifoQueue}
+	validFIFOThroughputLimitValues   = []string{sqsFifoPerMessageGroupID, sqsFifoPerQueue}
```
```suggestion
	validSQSDeduplicationScopeValues    = []string{sqsFifoMessageGroup, sqsFifoQueue}
	validSQSFIFOThroughputLimitValues   = []string{sqsFifoPerMessageGroupID, sqsFifoPerQueue}
```

internal/pkg/manifest/validate.go
```diff
@@ -1473,6 +1484,70 @@ func (q SQSQueue) validate() error {
 	return nil
 }
 
+// validate returns nil if FIFOAdvanceConfig is configured correctly.
+func (q FIFOAdvanceConfig) validate() error {
+	if q.IsEmpty() {
+		return nil
+	}
+
+	if err := q.validateHighThroughputFifo(); err != nil {
```
```suggestion
	if err := q.validateHighThroughputFIFO(); err != nil {
```

internal/pkg/manifest/validate.go
```diff
@@ -1473,6 +1484,70 @@ func (q SQSQueue) validate() error {
...
+// validateFIFO returns nil if FIFOAdvanceConfigOrBool is configured correctly.
+func (q FIFOAdvanceConfigOrBool) validate() error {
+	if q.Enable != nil {
+		return nil
+	}
+	return q.Advanced.validate()
+}
+
+func (q FIFOAdvanceConfig) validateHighThroughputFifo() error {
+	if q.HighThroughputFifo != nil {
```
efekarakus: nit: can we flip this if-statement to reduce nesting

internal/pkg/manifest/validate.go
```diff
@@ -1473,6 +1484,70 @@ func (q SQSQueue) validate() error {
...
+		return nil
+	}
+
+	if err := q.validateHighThroughputFifo(); err != nil {
+		return err
+	}
+	if err := q.validateDeduplicationScope(); err != nil {
+		return err
+	}
+	if err := q.validateFifoThroughputLimit(); err != nil {
```
```suggestion
	if err := q.validateFIFOThroughputLimit(); err != nil {
```

internal/pkg/manifest/worker_svc.go
```diff
@@ -207,9 +238,22 @@ func (s *WorkerService) EnvFile() string {
 // Subscriptions returns a list of TopicSubscriotion objects which represent the SNS topics the service
 // receives messages from.
 func (s *WorkerService) Subscriptions() []TopicSubscription {
+	s.RetrofitFIFOConfig()
```
efekarakus: can we make `RetrofitFIFOConfig`now private and rename to `retrofitFIFOTopicNames`? can we also mention in the comment of this method that this .fifo suffix happens.

internal/pkg/template/workload.go
```diff
@@ -436,10 +436,24 @@ type TopicSubscription struct {
...
+	Retention       *int64
+	Delay           *int64
+	Timeout         *int64
+	DeadLetter      *DeadLetterQueue
+	FIFOQueueConfig *FIFOQueueConfig
+}
+
+// FifoAdvanceConfigOrBool holds information needed to render a FIFO SQS Queue in a container definition.
+type FIFOQueueConfig struct {
+	Enable   *bool
```
Lou1415926: Similar to SNS, we could probably get rid of `Enable` because a `nil` `FIFOQueueConfig` could mean disabled, and `non-nil` meaning enabled.

```go
type FIFOQueueConfig struct {
         FIFOThroughputLimit       *string
	ContentBasedDeduplication *bool
	DeduplicationScope        *string
}
```

internal/pkg/template/templates/workloads/partials/cf/subscribe.yml
```diff
@@ -60,7 +60,7 @@ EventsKMSKey:
 
 EventsQueue:
   Metadata:
-    'aws:copilot:description': 'An events SQS queue to buffer messages'
+    'aws:copilot:description': {{ if and  .Subscribe .Subscribe.Queue .Subscribe.Queue.IsFIFO }}'An events SQS FIFO queue to buffer messages'{{ else}}'An events SQS queue to buffer messages'{{ end}}
```
Lou1415926, line 63: Is it safe to simplify this to `{{ if .Subscribe.Queue.IsFIFO }}` ?

internal/pkg/manifest/workload.go
```diff
@@ -24,6 +24,14 @@ const (
 	defaultDockerfileName = "Dockerfile"
 )
 
+// SQS Queue configs
+const (
+	sqsFifoPerMessageGroupID = "perMessageGroupId"
+	sqsFifoPerQueue          = "perQueue"
+	sqsFifoMessageGroup      = "messageGroup"
+	sqsFifoQueue             = "queue"
+)
```
efekarakus: I find it helpful to name "enum" field consts after what they represent, the [AWS sdk for example](https://docs.aws.amazon.com/sdk-for-go/api/service/cloudformation/#pkg-constants) does similarly

```suggestion
// SQS Queue field options.
const (
	sqsFIFOThroughputLimitPerMessageGroupID = "perMessageGroupId"
	sqsFIFOThroughputLimitPerQueue          = "perQueue"
	sqsDeduplicationScopeMessageGroup      = "messageGroup"
	sqsDeduplicationScopeQueue             = "queue"
)
```

internal/pkg/template/templates/workloads/partials/cf/subscribe.yml
efekarakus: >It will be a lot of configurations that we will be exposing to the customer at once. I think we can add those configs if customers ask for it. What do you think about it?

Sounds good!

internal/pkg/template/workload.go
```diff
@@ -436,10 +436,24 @@ type TopicSubscription struct {
...
+	Advanced *FIFOAdvanceConfig
+}
+
+// FifoAdvanceConfig holds information needed to render a advanced FIFO SQS Queue in a container definition.
+type FIFOAdvanceConfig struct {
+	FIFOThroughputLimit       *string
+	ContentBasedDeduplication *bool
+	DeduplicationScope        *string
```
efekarakus:
```suggestion
type FIFOQueueConfig struct {
	FIFOThroughputLimit       *string
	ContentBasedDeduplication *bool
	DeduplicationScope        *string
}
```

internal/pkg/manifest/worker_svc.go
```diff
@@ -15,13 +15,11 @@ import (
...
 	workerSvcManifestPath = "workloads/services/worker/manifest.yml"
-	standardTopicType     = "standard"
-	fifoTopicType         = "fifo"
-	defaultTopicType      = standardTopicType
 )
 
 var (
-	errUnmarshalQueueOpts = errors.New(`cannot unmarshal "queue" field into bool or map`)
+	errUnmarshalQueueOpts  = errors.New(`cannot unmarshal "queue" field into bool or map`)
+	errUnmarshalFifoConfig = errors.New("unable to unmarshal fifo field into boolean or compose-style map")
```
Lou1415926:
```suggestion
	errUnmarshalFifoConfig = errors.New(`unable to unmarshal "fifo" field into boolean or compose-style map`)
```
I guess this was because the SNS topic isn't merged yet?

internal/pkg/manifest/worker_svc.go
```diff
@@ -207,9 +238,22 @@ func (s *WorkerService) EnvFile() string {
 // Subscriptions returns a list of TopicSubscriotion objects which represent the SNS topics the service
 // receives messages from.
 func (s *WorkerService) Subscriptions() []TopicSubscription {
+	s.RetrofitFIFOConfig()
 	return s.Subscribe.Topics
 }
 
+// RetrofitFIFOConfig appends ".fifo" suffix to fifo topics.
+func (s *WorkerService) RetrofitFIFOConfig() {
```
Lou1415926: We probably no longer need to export this method now.

Also, I think we should be cautious about changing the value of `s.Subscribe.Topics[idx].Name` here. That is a user input, I feel like we should keep the user input as is. Besides, changing the value seems like a "side effect" of the `Subscriptions()` method. It is not obvious to the callers that calling `Subscription` would cause changes of the underlying data within `s *WorkerService`.

What do you think of:
```go
subs := make([]TopicSubscription, len(s.Subscribe.Topics))
for idx, topic := range s.Subscribe.Topics {
                 var topicSubscription TopicSubscription
                 // name = aws.StringValue(topic.Name) 
	         // if fifo { name = aws.String(aws.StringValue(topic.Name) + ".fifo") }
                 // topicSubscription.Name = name
                 subs = append(subs, topicSubscription)
}
return subs

```

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: Yes I was working on the same thing here when I got that comment in SNS PR. The change is coming in the latest commit. 

I already have this in place - 

```
         if !q.FIFO.IsEnabled() {
		return queue
	}

	if aws.BoolValue(q.FIFO.Enable) {
		queue.FIFOQueueConfig = &template.FIFOQueueConfig{}
		return queue
	}
       if !q.FIFO.Advanced.IsEmpty() {
                // advanced config ...
       }
```

Let me know if that works

internal/pkg/manifest/validate.go
```diff
@@ -1459,7 +1457,20 @@ func (q SQSQueueOrBool) validate() error {
 	if q.IsEmpty() {
 		return nil
 	}
-	return q.Advanced.validate()
+	if q.Enabled != nil {
+		return nil
+	}
```
Lou1415926: Do we need this block?

internal/pkg/manifest/validate.go
```diff
@@ -1459,7 +1457,20 @@ func (q SQSQueueOrBool) validate() error {
...
+	if err := q.Advanced.validate(); err != nil {
+		return err
+	}
+	if q.isFIFO() {
```
Lou1415926: Do we need this conditional?

I feel like `validate` is supposed to validate the manifest fields without having to know whether the queue is FIFO or not. 

I think methods like `isFIFO` has the assumption that the manifest is valid. Therefore, I'd argue that making a call to `isFIFO` in `validate` is a little backward logically 🤔

internal/pkg/manifest/validate.go
paragbhingre: Yes because this block checks for the following manifest config.

```
subscribe:
  topics:
     - name: mytopic
       service: myservice
       queue: true
```

We don't need to validate further if the queue is true.
Does that make sense?

internal/pkg/manifest/validate.go
paragbhingre: isFIFO simply checks if manifest has any fifo config or not? Is there is any fifo configs available then only we call validate on fifo configs. does that make sense? Or I shouldn't be checking if `fifo` field has been mentioned or not at all?

internal/pkg/manifest/validate.go
Lou1415926: I feel like ideally `validate` shouldn't have any assumption like that... (again, ideally) it probably shouldn't make assumptions such as  "if `q.Enabled != nil` then `q.Advanced` must have been empty so we can skip now". It should brainlessly validate as if it knows nothing about logics in other methods such as `UnmarshalYAML`.

internal/pkg/manifest/validate.go
Lou1415926: In another word, I think `validate` should just validate every fields out there. It should call `validate` on the fifo struct even if it's empty.

paragbhingre: Okay I get it

internal/pkg/manifest/worker_svc.go
paragbhingre: Sounds good to me

internal/pkg/template/templates/workloads/partials/cf/subscribe.yml
paragbhingre: It is not safe because `.Subscribe.Queue` can be nil and if we call  `.Subscribe.Queue.IsFIFO` then it will also fail.

internal/pkg/cli/deploy/svc.go
```diff
@@ -1171,17 +1171,17 @@ func (d *workerSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (*
 	for _, topic := range topics {
 		topicARNs = append(topicARNs, topic.ARN())
 	}
-	subs := d.wsMft.Subscriptions()
+	subs := d.wsMft.Subscribe.Subscriptions()
```
efekarakus: nit: reading `d.wsMft.Subscribe.Subscriptions()` is odd (`.Subscribe` should already be the subscriptions), can we move it back to be at the manifest level?

efekarakus (approved):
A tiny change and then feel free to remove the label

internal/pkg/manifest/validate.go
```diff
@@ -1459,7 +1457,10 @@ func (q SQSQueueOrBool) validate() error {
 	if q.IsEmpty() {
 		return nil
 	}
-	return q.Advanced.validate()
+	if err := q.Advanced.validate(); err != nil {
+		return err
+	}
+	return q.Advanced.FIFO.validate()
```
Lou1415926: Is there any reason why we prefer this line here over in `func (q SQSQueue) validate`?
```go
func (q SQSQueue) validate() error {
	if q.IsEmpty() {
		return nil
	}
	if err := q.DeadLetter.validate(); err != nil {
		return fmt.Errorf(`validate "dead_letter": %w`, err)
	}
	return q.FIFO.validate()
}
```

internal/pkg/manifest/validate.go
```diff
@@ -1473,6 +1474,71 @@ func (q SQSQueue) validate() error {
...
+	}
+	return nil
+}
+
+// validateFIFO returns nil if FIFOAdvanceConfigOrBool is configured correctly.
+func (q FIFOAdvanceConfigOrBool) validate() error {
+	if q.Enable != nil {
+		return nil
+	}
+	return q.Advanced.validate()
```
```suggestion
	if q.isEmpty() {
		return nil
	}
	return q.Advanced.validate()
```

nit: imo I feel like we don't want to skip `q.Advanced.validate()` if `q.Enable != nil`.. so far my understanding for `validate` is that it should just validate every fields without the assumption that some fields being non-empty implies that the other must be empty

Lou1415926 (commented):
LGTM just one final question!

internal/pkg/manifest/validate.go
paragbhingre: There is no preference as such, I will move it into `func (q SQSQueue) validate()`

internal/pkg/manifest/worker_svc.go
```diff
@@ -125,16 +111,66 @@ func (q *SQSQueueOrBool) UnmarshalYAML(value *yaml.Node) error {
...
+func (f *FIFOAdvanceConfigOrBool) IsEnabled() bool {
+	return aws.BoolValue(f.Enable) || !f.Advanced.IsEmpty()
+}
+
+// FifoAdvanceConfig represents the advanced fifo queue config.
+type FIFOAdvanceConfig struct {
+	ContentBasedDeduplication *bool   `yaml:"content_based_deduplication"`
+	DeduplicationScope        *string `yaml:"deduplication_scope"`
+	FIFOThroughputLimit       *string `yaml:"fifo_throughput_limit"`
+	HighThroughputFifo        *bool   `yaml:"high_throughput_fifo"`
```
efekarakus: can we rename this field to remove the "_fifo" suffix? it's redundant :) 
```suggestion
	HighThroughputFifo        *bool   `yaml:"high_throughput"`
```

internal/pkg/manifest/worker_svc.go
```diff
@@ -125,16 +111,66 @@ func (q *SQSQueueOrBool) UnmarshalYAML(value *yaml.Node) error {
...
+// IsEmpty returns true if the FifoAdvanceConfigOrBool struct has all nil values.
+func (f *FIFOAdvanceConfigOrBool) IsEnabled() bool {
+	return aws.BoolValue(f.Enable) || !f.Advanced.IsEmpty()
+}
+
+// FifoAdvanceConfig represents the advanced fifo queue config.
+type FIFOAdvanceConfig struct {
+	ContentBasedDeduplication *bool   `yaml:"content_based_deduplication"`
+	DeduplicationScope        *string `yaml:"deduplication_scope"`
+	FIFOThroughputLimit       *string `yaml:"fifo_throughput_limit"`
```
efekarakus: similarly for this one, if we can rename to:
```suggestion
	FIFOThroughputLimit       *string `yaml:"throughput_limit"`
```

Lou1415926 (approved):
:shipit:

PR #4017, opened 2022-09-20
feat: env manifest `cdn.terminate_tls`

dannyrandall:
To enable TLS termination on `cdn`, all currently deployed services must have `http.redirect_to_https` set to `false`


internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -73,6 +74,27 @@ func (e *ELBV2) ListenerRuleHostHeaders(ruleARN string) ([]string, error) {
 	return hostHeaders, nil
 }
 
+// ListenerRuleIsRedirect returns true if ruleARN has an action to redirect.
+func (e *ELBV2) ListenerRuleIsRedirect(ctx context.Context, ruleARN string) (bool, error) {
```
efekarakus: What do you think of?
```suggestion
func (elb *ELBV2) HasRedirectAction(ctx context.Context, ruleARN string) (bool, error) {
```

Alternatively, we can do something like this that's a bit more work but I think a nicer API:
```go
// DescribeRule returns information about a listener rule.
func (elb *ELBV2) DescribeRule(ruleARN string) (Rule, error)

// Rule represents a load balancer listener rule.
type Rule struct { ... }

// HasRedirectAction returns true if the rule has an action that redirects.
func (r Rule) HasRedirectAction() bool
```

internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -73,6 +74,27 @@ func (e *ELBV2) ListenerRuleHostHeaders(ruleARN string) ([]string, error) {
...
+	resp, err := e.client.DescribeRules(&elbv2.DescribeRulesInput{
+		RuleArns: aws.StringSlice([]string{ruleARN}),
+	})
+	switch {
+	case err != nil:
+		return false, fmt.Errorf("get listener rule for %s: %w", ruleARN, err)
+	case len(resp.Rules) == 0:
+		return false, fmt.Errorf("cannot find listener rule %s", ruleARN)
```
efekarakus: nit:
```suggestion
		return false, fmt.Errorf("listener rule %q not found", ruleARN)
```

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +141,115 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
-			return stack.NewEnvConfigFromExistingStack(in, lastForceUpdateID, oldParams)
+			return cfnstack.NewEnvConfigFromExistingStack(in, lastForceUpdateID, oldParams)
+		},
+		envDescriber:      envDescriber,
+		envManagerSession: envManagerSession,
+		newLBDescriber: func(sess *session.Session) lbDescriber {
+			return elbv2.New(sess)
+		},
+		newServiceStackDescriber: func(svc string, sess *session.Session) stackDescriber {
+			return stack.NewStackDescriber(cfnstack.NameForService(in.App.Name, in.Env.Name, svc), sess)
```
efekarakus: do we need to accept a session as a parameter? if I'm following correctly the session seems to be always the `envManagerSession`
```suggestion
		elb: elbv2.New(envManagerSession)
		newServiceStackDescriber: func(svc string) stackDescriber {
			return stack.NewStackDescriber(cfnstack.NameForService(in.App.Name, in.Env.Name, svc), envManagerSession)
```

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +141,115 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
 		},
 	}, nil
 }
 
+// Verify checks that the manifest configuration is valid to deploy alongside
+// the currently deployed resources.
+func (d *envDeployer) Verify(ctx context.Context, mft *manifest.Environment) error {
```
efekarakus: What do you think of renaming the method to `Validate` to be consistent with our other packages (like `manifest`)?
```suggestion
// Validate returns an error if the environment manifest is incompatible with services and application configurations.
func (d *envDeployer) Validate(ctx context.Context, mft *manifest.Environment) error {
```

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +141,115 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+	if mft.CDNEnabled() && mft.HTTPConfig.Public.Certificates == nil && d.app.Domain != "" {
```
efekarakus: nit: I'm a little worried that we're "leaking" the internal structure of our manifest to packages outside of `manifest` (like in `stack` and `deploy`) that would make it difficult to change the internal representation.
 
What do you think of adding a method instead?
```suggestion
	if mft.CDNEnabled() && !mft.HasImportedPublicALBCerts() && d.app.Domain != "" {
```

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +141,115 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+		// With managed domain, if the customer isn't using `alias` the A-records are inserted in the service stack as each service domain is unique.
+		// However, when clients enable CloudFront, they would need to update all their existing records to now point to the distribution.
+		// Hence, we force users to use `alias` and let the records be written in the environment stack instead.
+		if err := d.envDescriber.ValidateCFServiceDomainAliases(); err != nil {
+			return err
+		}
+	}
+
+	if mft.CDNEnabled() && aws.BoolValue(mft.CDNConfig.Config.TerminateTLS) {
```
efekarakus: samesies here, should we have more methods in the manifest pkg something like `mft.CDN().IsTLSTerminated()`?

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +141,115 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+		// ensure all services _are not_ doing http->https redirect
+		if err := d.verifyALBWorkloadsDontRedirect(ctx); err != nil {
+			return fmt.Errorf("can't enable TLS termination on CDN: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// verifyALBWorkloadsDontRedirect verifies that none of the ALB Workloads
```
```suggestion
// verifyALBWorkloadsDontRedirect verifies that none of the public ALB Workloads
```

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +141,115 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+		return err
+	}
+	if len(badServices) > 0 {
+		sort.Strings(badServices)
+		return fmt.Errorf("HTTP traffic redirects to HTTPS in %v %v.\nSet http.redirect_to_https to false for %v and redeploy %v.",
+			english.PluralWord(len(badServices), "service", "services"),
+			english.OxfordWordSeries(badServices, "and"),
+			english.PluralWord(len(badServices), "that service", "those services"),
+			english.PluralWord(len(badServices), "it", "them"),
+		)
```
efekarakus: Really nice error message 

instead of having a new line in the error mesage, what do you think of adding a custom error type that implements the following interface:
https://github.com/aws/copilot-cli/blob/6132cb54fb1167b297bce0a218026760cfbae0a0/cmd/copilot/main.go#L19-L21

```go
type errEnvHasPublicServicesWithRedirect struct  { /* ... */ }

func (e errEnvHasPublicServicesWithRedirect) Error() string {
   return fmt.Sprintf("services %q redirect HTTP traffic to HTTPS"
}

func (e errEnvHasPublicServicesWithRedirect) RecommendedAction() string {
   return fmt.Sprintf(`In the service manifests for %q, set the following field:
%s
and run %s`, naughtyServices, color.HighlightCodeBlock(`http:
  redirect_to_https: true
`), color.HighlightCode("copilot svc deploy") 
}
```

internal/pkg/manifest/validate_env_test.go
```diff
@@ -732,41 +732,57 @@ func TestSubnetsConfiguration_validate(t *testing.T) {
...
-			in: environmentCDNConfig{},
+			in: EnvironmentCDNConfig{},
 		},
 		"valid if bool specified": {
-			in: environmentCDNConfig{
+			in: EnvironmentCDNConfig{
 				Enabled: aws.Bool(false),
 			},
 		},
 		"valid if advanced config configured correctly": {
```
Lou1415926: nit: now that we have `terminate_tls`, we might want to  rename this unit test to something like "success with cert without tls"

internal/pkg/cli/deploy/env.go
```diff
@@ -44,6 +55,23 @@ type prefixListGetter interface {
...
+type lbDescriber interface {
+	ListenerRuleIsRedirect(context.Context, string) (bool, error)
+}
+
+type stackDescriber interface {
+	Resources() ([]*stack.Resource, error)
+}
+
+const (
+	svcStackResourceHTTPListenerRuleLogicalID = "HTTPListenerRuleWithDomain"
```
Lou1415926: Could this variable be better defined in `template` pkg and imported here?

internal/pkg/cli/deploy/env.go
Lou1415926: just a random thought  -  I think I did like how `isManagedCDNEnabled` characterized this conditional and how it helped with understanding the flow. What about ```if managedCDNEnabled :=  mft.CDNEnabled() && !mft.HasImportedPublicALBCerts() && d.app.Domain != ""; managedCDNEnabled {```

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +141,115 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+
+// lbServiceRedirects returns true if svc's HTTP listener rule redirects.
+func (d *envDeployer) lbServiceRedirects(ctx context.Context, svc string) (bool, error) {
+	stackDescriber := d.newServiceStackDescriber(svc, d.envManagerSession)
+	resources, err := stackDescriber.Resources()
+	if err != nil {
+		return false, fmt.Errorf("get stack resources: %w", err)
+	}
+
+	ruleARN := ""
```
Lou1415926: nit: 
```suggestion
	var ruleARN string
```

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +141,115 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+	for _, res := range resources {
+		if res.LogicalID == svcStackResourceHTTPListenerRuleLogicalID {
+			ruleARN = res.PhysicalID
```
Lou1415926, line 238:
```suggestion
			ruleARN = res.PhysicalID
			break
```

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +141,115 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+		}
+	}
+	if ruleARN == "" {
+		return false, fmt.Errorf("resource %q not present", svcStackResourceHTTPListenerRuleLogicalID)
+	}
+
+	lbDescriber := d.newLBDescriber(d.envManagerSession)
+	isRedirect, err := lbDescriber.ListenerRuleIsRedirect(ctx, ruleARN)
```
Lou1415926: qq:

If we go with the route of fetching the physical resource ID from the service stack, then is it safe for us to make the assumption that `HTTPListenerRuleWithDomain` must have the `redirect` action?

dannyrandall: I think so - here's my reasoning:

We enable HTTPS for LBWS if a [domain or public certs have been imported](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/deploy/cloudformation/stack/lb_web_svc.go#L72-L88). If HTTPS is enabled, then the service stack [uses](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/workloads/partials/cf/alb.yml#L82-L86) the https-listener, which _always_ creates [`HTTPListenerRuleWithDomain`](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/workloads/partials/cf/https-listener.yml#L74). Then, based on --------------

great question - thanks for making me think about it :) i think we should modify the check [here](https://github.com/aws/copilot-cli/pull/4017/files#diff-924fba6d7ac2c8e2b099ec2776e78f958c7b225d5484001289b0e62283df3db1R169) to ensure we only run this check when a domain or public cert is imported to the env. once i change that, does this logic make sense to you? we can be safe assuming that `HTTPListenerRuleWithDomain` will always be there because we know the environment is using HTTPS.

internal/pkg/aws/elbv2/elbv2.go
dannyrandall: changed the error when changing the API https://github.com/aws/copilot-cli/pull/4017#discussion_r976787597

internal/pkg/cli/deploy/env.go
dannyrandall: 🤦

internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -73,6 +75,33 @@ func (e *ELBV2) ListenerRuleHostHeaders(ruleARN string) ([]string, error) {
 	return hostHeaders, nil
 }
 
+type Rule elbv2.Rule
```
iamhopaul123, line 81: maybe some comment for the public type? Like why we need to define this instead of using `elbv2.Rule`

internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -73,6 +75,33 @@ func (e *ELBV2) ListenerRuleHostHeaders(ruleARN string) ([]string, error) {
...
+func (e *ELBV2) DescribeRule(ctx context.Context, ruleARN string) (Rule, error) {
+	resp, err := e.client.DescribeRules(&elbv2.DescribeRulesInput{
+		RuleArns: aws.StringSlice([]string{ruleARN}),
+	})
+	switch {
+	case err != nil:
+		return Rule{}, err
+	case len(resp.Rules) == 0:
+		return Rule{}, errors.New("not found")
+	}
```
iamhopaul123: We probably don't need a switch here
```suggestion
        if err != nil:
            return Rule{}, fmt.Errorf("describe listener rules: %w", err)
        if !len(resp.Rules):
            return Rule{}, fmt.Errorf("cannot find any rules related to %s", ruleARN)
```

internal/pkg/cli/deploy/env.go
```diff
@@ -83,6 +114,14 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
 	if err != nil {
 		return nil, fmt.Errorf("get env session: %w", err)
 	}
+	envDescriber, err := describe.NewEnvDescriber(describe.NewEnvDescriberConfig{
+		App:         in.App.Name,
+		Env:         in.Env.Name,
+		ConfigStore: in.ConfigStore,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("get env describer: %w", err)
```
```suggestion
		return nil, fmt.Errorf("initialize env describer: %w", err)
```

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +139,154 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+		lbDescriber:       elbv2.New(envManagerSession),
+		newServiceStackDescriber: func(svc string) stackDescriber {
+			return stack.NewStackDescriber(cfnstack.NameForService(in.App.Name, in.Env.Name, svc), envManagerSession)
 		},
 	}, nil
 }
 
+// Validate returns an error if the environment manifest is incompatible with services and application configurations.
+func (d *envDeployer) Validate(ctx context.Context, mft *manifest.Environment, output io.Writer) error {
+	isManagedCDNEnabled := mft.CDNEnabled() && !mft.HasImportedPublicALBCerts() && d.app.Domain != ""
```
iamhopaul123, line 153: nice move!

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +139,154 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+		envManagerSession: envManagerSession,
+		lbDescriber:       elbv2.New(envManagerSession),
+		newServiceStackDescriber: func(svc string) stackDescriber {
+			return stack.NewStackDescriber(cfnstack.NameForService(in.App.Name, in.Env.Name, svc), envManagerSession)
 		},
 	}, nil
 }
 
+// Validate returns an error if the environment manifest is incompatible with services and application configurations.
+func (d *envDeployer) Validate(ctx context.Context, mft *manifest.Environment, output io.Writer) error {
```
iamhopaul123: Maybe define this function as a private function "validate cdn" because it is all CDN related and call it in `Validate()`. And we can return early if cdn is not enabled to simplify the logic.

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +139,154 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+	if mft.CDNEnabled() && mft.CDNDoesTLSTermination() && (mft.HasImportedPublicALBCerts() || d.app.Domain != "") {
+		err := d.validateALBWorkloadsDontRedirect(ctx)
+		var redirErr *errEnvHasPublicServicesWithRedirect
+		switch {
+		case errors.As(err, &redirErr) && mft.ALBIngressRestrictedToCDN():
+			return err
+		case errors.As(err, &redirErr):
+			fmt.Fprintln(output, redirErr.Warning())
+		case err != nil:
+			return fmt.Errorf("can't enable TLS termination on CDN: %w", err)
```
iamhopaul123: Error from private function. We can be more verbose in `validateALBWorkloadsDontRedirect`.
```suggestion
			return err
```

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +139,154 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+			break
+		}
+	}
+	if ruleARN == "" {
+		return false, fmt.Errorf("http listener not found on service %q", svc)
+	}
+
+	rule, err := d.lbDescriber.DescribeRule(ctx, ruleARN)
+	if err != nil {
+		return false, fmt.Errorf("get listener rule %q: %w", ruleARN, err)
```
```suggestion
		return false, fmt.Errorf("describe listener rule %q: %w", ruleARN, err)
```

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +139,154 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+
+// lbServiceRedirects returns true if svc's HTTP listener rule redirects. We only check
+// HTTPListenerRuleWithDomain because HTTPListenerRule:
+// a) doesn't ever redirect
+// b) won't work with cloudfront anyways (can't point ALB default DNS to CF)
+func (d *envDeployer) lbServiceRedirects(ctx context.Context, svc string) (bool, error) {
+	stackDescriber := d.newServiceStackDescriber(svc)
+	resources, err := stackDescriber.Resources()
+	if err != nil {
+		return false, fmt.Errorf("get stack resources: %w", err)
```
iamhopaul123, line 232:
```suggestion
		return false, fmt.Errorf("get stack resources for service %q: %w", svc, err)
```

internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -73,6 +75,33 @@ func (e *ELBV2) ListenerRuleHostHeaders(ruleARN string) ([]string, error) {
 	return hostHeaders, nil
 }
 
+type Rule elbv2.Rule
+
+// DescribeRule returns the Rule with ruleARN.
+func (e *ELBV2) DescribeRule(ctx context.Context, ruleARN string) (Rule, error) {
+	resp, err := e.client.DescribeRules(&elbv2.DescribeRulesInput{
```
iamhopaul123: We can use the WithConext api

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +139,154 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+// in this environment have a redirect in their HTTPWithDomain listener.
+// If any services redirect, an error is returned.
+func (d *envDeployer) validateALBWorkloadsDontRedirect(ctx context.Context) error {
+	params, err := d.envDescriber.Params()
+	if err != nil {
+		return fmt.Errorf("get env params: %w", err)
+	}
+
+	services := strings.Split(params[cfnstack.EnvParamALBWorkloadsKey], ",")
+	g, ctx := errgroup.WithContext(ctx)
```
iamhopaul123: I feel like we can create the context within this function because this is where it get used.

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +139,154 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+
+			return nil
+		})
+	}
+
+	if err := g.Wait(); err != nil {
+		return err
+	}
+	if len(badServices) > 0 {
+		sort.Strings(badServices)
```
iamhopaul123, line 215: nice!

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +139,154 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+			fmt.Fprintln(output, redirErr.Warning())
+		case err != nil:
+			return fmt.Errorf("can't enable TLS termination on CDN: %w", err)
+		}
+	}
+
+	return nil
+}
+
+type errEnvHasPublicServicesWithRedirect struct {
```
iamhopaul123: Maybe move it to `errors.go`?

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +139,154 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+To fix this, set the following field in %s manifest:
+%s
+and run %s.`,
+		english.PluralWord(n, "Service", "Services"),
+		english.OxfordWordSeries(quoted, "and"),
+		english.PluralWord(n, "redirects", "redirect"),
+		color.Emphasize(english.PluralWord(n, "This service", "These services")+" will not be reachable through the CDN."),
+		english.PluralWord(n, "its", "each"),
+		color.HighlightCodeBlock("http:\n  redirect_to_https: true"),
+		color.HighlightCode("copilot svc deploy"),
```
iamhopaul123: Awesome!

internal/pkg/aws/elbv2/elbv2.go
dannyrandall: oh, good idea. it's so we can add functions to it - i'll a doc comment

internal/pkg/cli/deploy/env.go
dannyrandall: yeah i guess my main thought here was that we can add more Validation logic for the env against the deployed stuff to this function - but i'm happy to change it to be CDN specific for now. i just want one function to be called from `svc_deploy.go` and `svc_package.go` for this class of Validation in the future.

iamhopaul123: Yeah we could do both like
```
func (d *envDeployer) Validate {
   return d.validateCDN()
}

File: internal/pkg/cli/deploy/env.go (hunk 21 above)
Comment (dannyrandall) [reply internal/pkg/cli/deploy/env.go:232]: service context is already added [here](https://github.com/aws/copilot-cli/pull/4017/files#diff-924fba6d7ac2c8e2b099ec2776e78f958c7b225d5484001289b0e62283df3db1R237)

File: internal/pkg/cli/deploy/env.go (hunk 19 above)
Comment (dannyrandall) [reply internal/pkg/cli/deploy/env.go]: i feel like it's a little weird for the function `validateALBWorkloadsDontRedirect` to embed the context that "i'm doing this to figure out if we can enable CDN termination" - what do you think?

File: internal/pkg/aws/elbv2/elbv2.go (hunk 15 above)
Comment (dannyrandall) [reply internal/pkg/aws/elbv2/elbv2.go]: sad 😔 i'll do it to save 1 line

File: internal/pkg/cli/deploy/env.go (hunk 19 above)
Comment (iamhopaul123) [reply internal/pkg/cli/deploy/env.go]: Yeah then maybe omitting it would be better? I am not too sure either. Just feel like the wrapping isn't actually what the function does.

File: internal/pkg/aws/elbv2/elbv2.go (hunk 15 above)
Comment (iamhopaul123) [reply internal/pkg/aws/elbv2/elbv2.go]: lolll

File: internal/pkg/cli/deploy/env.go (hunk 23 above)
Comment (dannyrandall) [reply internal/pkg/cli/deploy/env.go]: true.............................................i _wish_ i had a good reason to keep it in `svc_deploy.go`, since that's where the whole chain "starts"......but i guess i don't ¯\\\_(ツ)\_/¯

File: internal/pkg/cli/deploy/env.go (hunk 19 above)
Comment (dannyrandall) [reply internal/pkg/cli/deploy/env.go]: hmmm, i do feel like it's useful information for the users to see though - otherwise the error would just be like "verify service 'svc1': http listener not found" and have no context _why_ `env deploy` even cared

File: internal/pkg/aws/elbv2/elbv2.go (hunk 14 above)
Comment (efekarakus) [reply internal/pkg/aws/elbv2/elbv2.go:81]: "to add some nice functionality to it." 🤣 

nit: I know this is too much work so I'm totally fine keeping it as is, but if we want to make the swap from SDKv1 to v2, then we probably don't even want to surface the underlying dependency

File: internal/pkg/aws/elbv2/elbv2.go [hunk 27]
```diff
@@ -73,6 +77,33 @@ func (e *ELBV2) ListenerRuleHostHeaders(ruleARN string) ([]string, error) {
...
+
+// DescribeRule returns the Rule with ruleARN.
+func (e *ELBV2) DescribeRule(ctx context.Context, ruleARN string) (Rule, error) {
+	resp, err := e.client.DescribeRulesWithContext(ctx, &elbv2.DescribeRulesInput{
+		RuleArns: aws.StringSlice([]string{ruleARN}),
+	})
+	if err != nil {
+		return Rule{}, err
+	} else if len(resp.Rules) == 0 {
+		return Rule{}, errors.New("not found")
```
efekarakus, line 91: nit:
```suggestion
		return Rule{}, fmt.Errorf("rule ARN %s is not found", ruleARN)
```

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +136,123 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
 			Env:             in.Env,
 		},
 		newStackSerializer: func(in *deploy.CreateEnvironmentInput, lastForceUpdateID string, oldParams []*awscfn.Parameter) stackSerializer {
-			return stack.NewEnvConfigFromExistingStack(in, lastForceUpdateID, oldParams)
+			return cfnstack.NewEnvConfigFromExistingStack(in, lastForceUpdateID, oldParams)
+		},
+		envDescriber:      envDescriber,
+		envManagerSession: envManagerSession,
```
efekarakus: can we delete this field 🥺

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +136,123 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+	}
+
+	if mft.CDNEnabled() && mft.CDNDoesTLSTermination() && (mft.HasImportedPublicALBCerts() || d.app.Domain != "") {
+		err := d.validateALBWorkloadsDontRedirect()
+		var redirErr *errEnvHasPublicServicesWithRedirect
+		switch {
+		case errors.As(err, &redirErr) && mft.ALBIngressRestrictedToCDN():
+			return err
+		case errors.As(err, &redirErr):
+			fmt.Fprintln(output, redirErr.Warning())
```
efekarakus: instead of `fmt` can we use the `log` pkg?

internal/pkg/cli/deploy/env.go
```diff
@@ -100,11 +136,123 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+		case err != nil:
+			return fmt.Errorf("can't enable TLS termination on CDN: %w", err)
```
efekarakus: nit: can we drop the `can't`? how does the error read if we have just `enable TLS termination on CDN: %w` (this is from Uber's go guideline:
> When adding context to returned errors, keep the context succinct by avoiding phrases like "failed to", which state the obvious and pile up as the error percolates up through the stack:
> https://github.com/uber-go/guide/blob/master/style.md#errors

I'll leave this to your judgement!

internal/pkg/cli/deploy/errors.go
```diff
@@ -38,3 +40,38 @@ func (e *errSvcWithALBAliasHostedZoneWithCDNEnabled) RecommendActions() string {
 	}
 	return strings.Join(msgs, "\n")
 }
+
+type errEnvHasPublicServicesWithRedirect struct {
+	services []string
+}
+
+func (e *errEnvHasPublicServicesWithRedirect) Error() string {
```
efekarakus, line 48: I still don't know if it's common for an `Error()` to return a multi-line string that's colored. 
Do you know of any multi-line string in the `stdlib`?

My recommendation is to keep `Error` to be a succinct single line message like:
```
Error() {
   "{len(services) redirect HTTP to HTTPS"
}
```
and have this nice explanation in `RecommendedActions`.

internal/pkg/cli/deploy/errors.go
```diff
@@ -38,3 +40,38 @@ func (e *errSvcWithALBAliasHostedZoneWithCDNEnabled) RecommendActions() string {
...
+		english.OxfordWordSeries(quoted, "and"),
+		english.PluralWord(n, "redirects", "redirect"),
+		color.Emphasize(english.PluralWord(n, "This service", "These services")+" will not be reachable through the CDN."),
+		english.PluralWord(n, "its", "each"),
+		color.HighlightCodeBlock("http:\n  redirect_to_https: true"),
+		color.HighlightCode("copilot svc deploy"),
+	)
+}
+
+func (e *errEnvHasPublicServicesWithRedirect) Warning() string {
```
efekarakus: nit: can this be private?

internal/pkg/cli/env_deploy.go
```diff
@@ -140,18 +124,6 @@ func (o *deployEnvOpts) Execute() error {
 	if err != nil {
 		return err
 	}
-	if o.isManagedCDNEnabled(mft) {
```
efekarakus, line 143: very nice

internal/pkg/cli/env_package.go
```diff
@@ -154,7 +154,9 @@ func (o *packageEnvOpts) Execute() error {
 	if err != nil {
 		return err
 	}
-
+	if err := deployer.Validate(mft, os.Stderr); err != nil {
```
efekarakus: nit: can we use https://github.com/aws/copilot-cli/blob/b6b1c869b0bfafcb9d011aa78a5391e0581854b1/internal/pkg/term/log/log.go#L16
instead of `os.Stderr`. I don't fully remember but these is some funky stuff with colors and Windows https://github.com/mattn/go-colorable/blob/11a925cff3d38c293ddc8c05a16b504e3e2c63be/colorable_windows.go#L102-L118

internal/pkg/manifest/env.go
```diff
@@ -184,24 +184,25 @@ func (cfg *EnvironmentConfig) EnvSecurityGroup() (*securityGroupConfig, bool) {
 	return nil, false
 }
 
-type environmentCDNConfig struct {
+type EnvironmentCDNConfig struct {
```
efekarakus, line 188: doc comments 🥺 for here and the other methods (HasImportedPublicALBCerts, HasImportedPublicALBCerts, HasImportedPublicALBCerts)

efekarakus (commented):
Looks great, just a bunch of tiny feedback.

internal/pkg/aws/elbv2/elbv2.go
dannyrandall: i think i'll leave it for now since the ruleARN context is added for both error cases [here](https://github.com/aws/copilot-cli/pull/4017/files#diff-924fba6d7ac2c8e2b099ec2776e78f958c7b225d5484001289b0e62283df3db1R248)

internal/pkg/cli/deploy/errors.go
dannyrandall: that makes sense - see how it looks now!

efekarakus: Looks great! ty.

nit: doc-comments are missing

efekarakus (approved):

internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -73,6 +77,33 @@ func (e *ELBV2) ListenerRuleHostHeaders(ruleARN string) ([]string, error) {
...
+type Rule elbv2.Rule
+
+// DescribeRule returns the Rule with ruleARN.
+func (e *ELBV2) DescribeRule(ctx context.Context, ruleARN string) (Rule, error) {
+	resp, err := e.client.DescribeRulesWithContext(ctx, &elbv2.DescribeRulesInput{
+		RuleArns: aws.StringSlice([]string{ruleARN}),
+	})
+	if err != nil {
+		return Rule{}, err
+	} else if len(resp.Rules) == 0 {
```
iamhopaul123, line 90: nit: we don't need the else

iamhopaul123 (approved):
Looks good thank you!

PR #4018, opened 2022-09-21
feat: add support for SNS FIFO

paragbhingre:
This PR fix 1st part of https://github.com/aws/copilot-cli/issues/3828

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -797,24 +797,24 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 	var publishers template.PublishOpts
 	// convert the topics to template Topics
 	for _, topic := range topics {
-		var fifo template.Fifo
-		if !topic.Fifo.IsEmpty() {
-			if aws.BoolValue(topic.Fifo.Enable) {
-				fifo.Enable = aws.BoolValue(topic.Fifo.Enable)
+		var fifoConfig *template.FIFOTopicConfig
+		if !topic.FIFO.IsEmpty() {
+			fifoConfig = &template.FIFOTopicConfig{
+				Enable: topic.FIFO.Enable,
 			}
-			if !topic.Fifo.Advanced.IsEmpty() {
-				fifo.ContentBasedDeduplication = aws.BoolValue(topic.Fifo.Advanced.ContentBasedDeduplication)
+			if !topic.FIFO.Advanced.IsEmpty() {
+				fifoConfig.ContentBasedDeduplication = topic.FIFO.Advanced.ContentBasedDeduplication
 			}
 		}
 		publishers.Topics = append(publishers.Topics, &template.Topic{
-			Name:      topic.Name,
-			Fifo:      fifo,
-			AccountID: accountID,
-			Partition: partition.ID(),
-			Region:    region,
-			App:       app,
-			Env:       env,
-			Svc:       svc,
+			Name:            topic.Name,
+			FIFOTopicConfig: fifoConfig,
+			AccountID:       accountID,
+			Partition:       partition.ID(),
+			Region:          region,
+			App:             app,
+			Env:             env,
+			Svc:             svc,
 		})
 	}
 
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -798,10 +798,8 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 	// convert the topics to template Topics
 	for _, topic := range topics {
 		var fifoConfig *template.FIFOTopicConfig
-		if !topic.FIFO.IsEmpty() {
-			if topic.FIFO.IsEnabled() {
-				fifoConfig = &template.FIFOTopicConfig{}
-			}
+		if topic.FIFO.IsEnabled() {
+			fifoConfig = &template.FIFOTopicConfig{}
 			if !topic.FIFO.Advanced.IsEmpty() {
 				fifoConfig = &template.FIFOTopicConfig{
 					ContentBasedDeduplication: topic.FIFO.Advanced.ContentBasedDeduplication,
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -799,11 +799,13 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 	for _, topic := range topics {
 		var fifoConfig *template.FIFOTopicConfig
 		if !topic.FIFO.IsEmpty() {
-			fifoConfig = &template.FIFOTopicConfig{
-				Enable: topic.FIFO.Enable,
+			if topic.FIFO.IsEnabled() {
+				fifoConfig = &template.FIFOTopicConfig{}
 			}
 			if !topic.FIFO.Advanced.IsEmpty() {
-				fifoConfig.ContentBasedDeduplication = topic.FIFO.Advanced.ContentBasedDeduplication
+				fifoConfig = &template.FIFOTopicConfig{
+					ContentBasedDeduplication: topic.FIFO.Advanced.ContentBasedDeduplication,
+				}
 			}
 		}
 		publishers.Topics = append(publishers.Topics, &template.Topic{
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -797,9 +797,18 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 	var publishers template.PublishOpts
 	// convert the topics to template Topics
 	for _, topic := range topics {
+		var fifo template.Fifo
+		if !topic.Fifo.IsEmpty() {
+			if aws.BoolValue(topic.Fifo.Enable) {
+				fifo.Enable = aws.BoolValue(topic.Fifo.Enable)
+			}
+			if !topic.Fifo.Advanced.IsEmpty() {
+				fifo.ContentBasedDeduplication = aws.BoolValue(topic.Fifo.Advanced.ContentBasedDeduplication)
+			}
+		}
 		publishers.Topics = append(publishers.Topics, &template.Topic{
 			Name:      topic.Name,
-			Type:      aws.StringValue(topic.Type),
+			Fifo:      fifo,
 			AccountID: accountID,
 			Partition: partition.ID(),
 			Region:    region,
--- a/internal/pkg/manifest/validate.go
+++ b/internal/pkg/manifest/validate.go
@@ -1414,31 +1414,22 @@ func (t Topic) validate() error {
 	if err := validatePubSubName(aws.StringValue(t.Name)); err != nil {
 		return err
 	}
-	if err := t.Fifo.validate(); err != nil {
+	if err := t.FIFO.validate(); err != nil {
 		return err
 	}
 	return nil
 }
 
 // validate returns nil if FifoAdvanceConfigOrBool is configured correctly.
-func (f FifoAdvanceConfigOrBool) validate() error {
+func (f FIFOAdvanceConfigOrBool) validate() error {
 	if f.IsEmpty() {
 		return nil
 	}
-	if f.Enable != nil {
-		return nil
-	}
-	if err := f.Advanced.validate(); err != nil {
-		return err
-	}
-	return nil
+	return f.Advanced.validate()
 }
 
 // validate returns nil if FifoAdvanceConfig is configured correctly.
-func (a FifoAdvanceConfig) validate() error {
-	if a.IsEmpty() {
-		return fmt.Errorf("content_based_deduplication is not defined correctly")
-	}
+func (a FIFOAdvanceConfig) validate() error {
 	return nil
 }
 
--- a/internal/pkg/manifest/validate.go
+++ b/internal/pkg/manifest/validate.go
@@ -1414,20 +1414,23 @@ func (t Topic) validate() error {
 	if err := validatePubSubName(aws.StringValue(t.Name)); err != nil {
 		return err
 	}
-	if t.Fifo.IsEmpty() {
-		return nil
-	}
-	if t.Fifo.Enable != nil {
-		return nil
-	}
-	if err := t.Fifo.Advanced.validate(); err != nil {
+	if err := t.Fifo.validate(); err != nil {
 		return err
 	}
 	return nil
 }
 
-// validate returns true if Fifo configuration is empty.
-func (f Fifo) validate() error {
+// validate returns nil if FifoAdvanceConfigOrBool is configured correctly.
+func (f FifoAdvanceConfigOrBool) validate() error {
+	if f.IsEmpty() {
+		return nil
+	}
+	if f.Enable != nil {
+		return nil
+	}
+	if err := f.Advanced.validate(); err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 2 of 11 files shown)


internal/pkg/manifest/workload.go
```diff
@@ -44,6 +44,7 @@ var (
 	errUnmarshalSecurityGroupOpts = errors.New(`unable to unmarshal "security_groups" field into slice of strings or compose-style map`)
 	errUnmarshalPlacementOpts     = errors.New("unable to unmarshal placement field into string or compose-style map")
 	errUnmarshalSubnetsOpts       = errors.New("unable to unmarshal subnets field into string slice or compose-style map")
+	errUnmarshalFifoConfig        = errors.New("unable to unmarshal fifo field into boolean or compose-style map")
```
Lou1415926:
```suggestion
	errUnmarshalFifoConfig        = errors.New(`unable to unmarshal "fifo" field into boolean or compose-style map`)
```

internal/pkg/manifest/workload.go
```diff
@@ -399,8 +400,48 @@ type PublishConfig struct {
 
 // Topic represents the configurable options for setting up a SNS Topic.
 type Topic struct {
-	Name *string `yaml:"name"`
-	Type *string `yaml:"type"`
+	Name *string                 `yaml:"name"`
+	Fifo FifoAdvanceConfigOrBool `yaml:"fifo"`
```
Lou1415926: nit: we may probably want to write it such that letters in `fifo` have consistent case (all upper or all lower) because "fifo" itself is an acronym. So maybe `FIFO FIFOAdvancedConfigOrBool` here.

internal/pkg/manifest/validate.go
```diff
@@ -1412,10 +1411,35 @@ func (p PublishConfig) validate() error {
...
+	}
+	return nil
+}
+
+// validate returns nil if FifoAdvanceConfig is configured correctly.
+func (a FifoAdvanceConfig) validate() error {
+	if a.IsEmpty() {
+		return fmt.Errorf("content_based_deduplication is not defined correctly")
+	}
+	return nil
```
Lou1415926: what do you think of 
```suggestion
        if f.IsEmpty() {
                return nil
        }
	return f.Advanced.validate()
}

// validate returns nil if FifoAdvanceConfig is configured correctly.
func (a FifoAdvanceConfig) validate() error {
	return nil
}
```

My concern is that, with the current implementation, `func (a FifoAdvanceConfig) validate() error` can only be invoked when `!f.IsEmpty()` and when `f.Enabled == nil` - otherwise there would be validation bug. This introduces implicit tangling between the caller and callee and could be error-prone in the future.

internal/pkg/template/workload.go
```diff
@@ -389,7 +389,7 @@ type PublishOpts struct {
 // Topic holds information needed to render a SNSTopic in a container definition.
 type Topic struct {
 	Name *string
-	Type string
+	Fifo Fifo
```
Lou1415926: What do you think of
```suggestion
type Topic struct {
	FIFO *FIFO
}

type FIFO struct {
         ContentBasedDeduplication bool
}	
```

and in CFN template we use
```
{{ if $topic.Fifo  }}
```
instead of `{{ if or $topic.Fifo.Enable $topic.Fifo.ContentBasedDeduplication }}` to check if FIFO is enabled.

Basically, if FIFO is a `nil` pointer, then it is not enabled. O/w, it is enable.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -797,9 +797,18 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 	var publishers template.PublishOpts
 	// convert the topics to template Topics
 	for _, topic := range topics {
+		var fifo template.Fifo
```
efekarakus: can we rename all references to `Fifo` to be either `fifo` or `FIFO`? 🥺 

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -797,9 +797,18 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
...
+		if !topic.Fifo.IsEmpty() {
+			if aws.BoolValue(topic.Fifo.Enable) {
+				fifo.Enable = aws.BoolValue(topic.Fifo.Enable)
+			}
```
efekarakus: can be simplified to:
```suggestion
	fifo.Enable = aws.BoolValue(topic.Fifo.Enable)
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -797,9 +797,18 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
...
+			if !topic.Fifo.Advanced.IsEmpty() {
+				fifo.ContentBasedDeduplication = aws.BoolValue(topic.Fifo.Advanced.ContentBasedDeduplication)
```
efekarakus: What do you think of adding more methods to the manifest similar to this feedback: https://github.com/aws/copilot-cli/pull/4017#discussion_r976809646

For example, a sample API might be:
```go
// FIFO returns a first-in first-out SNS topic and true if the underlying topic representation has FIFO enabled.
func (t Topic) FIFO() (FIFOTopic, bool)

type FIFOTopic Topic
func (t FIFOTopic) HasContentBasedDeduplication() bool
```

internal/pkg/manifest/validate.go
```diff
@@ -1412,10 +1411,35 @@ func (p PublishConfig) validate() error {
...
+	if err := f.Advanced.validate(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// validate returns nil if FifoAdvanceConfig is configured correctly.
+func (a FifoAdvanceConfig) validate() error {
+	if a.IsEmpty() {
+		return fmt.Errorf("content_based_deduplication is not defined correctly")
```
```suggestion
		return fmt.Errorf(`"content_based_deduplication" is not set`)
```

internal/pkg/manifest/validate.go
```diff
@@ -1412,10 +1411,35 @@ func (p PublishConfig) validate() error {
...
+func (f FifoAdvanceConfigOrBool) validate() error {
+	if f.IsEmpty() {
+		return nil
 	}
-	return validatePubSubName(aws.StringValue(t.Name))
+	if f.Enable != nil {
+		return nil
+	}
+	if err := f.Advanced.validate(); err != nil {
+		return err
```
```suggestion
		return fmt.Errorf(`validate "fifo": %v`, err)
```

internal/pkg/template/templates/workloads/partials/cf/publish.yml
```diff
@@ -2,19 +2,21 @@
 {{- range $topic := .Publish.Topics}}
 {{logicalIDSafe $topic.Name}}SNSTopic:
   Metadata:
-    'aws:copilot:description':  {{- if eq $topic.Type "fifo" }} 'A FIFO SNS topic to broadcast {{$topic.Name}} events' {{- else}} 'A SNS topic to broadcast {{$topic.Name}} events'{{- end}}
+    'aws:copilot:description': {{ if or $topic.Fifo.Enable $topic.Fifo.ContentBasedDeduplication }} 'A FIFO SNS topic to broadcast {{$topic.Name}} events' {{ else}} 'A SNS topic to broadcast {{$topic.Name}} events'{{ end}}
```
efekarakus: similar here, I think a method would simplify this logc that's in the template `$topic.IsFIFO()`

internal/pkg/template/workload.go
efekarakus: Perhaps something like this:
```go
type Topic struct {
  // other fields...
  FIFOConfig *FIFOConfig
}
```
where non-nil would indicate that fifo is enabled

efekarakus (commented):
Thanks for quickly addressing the manifest changes!

internal/pkg/template/workload.go
paragbhingre: Great catch, I am doing the exact same thing for the SQS FIFO right now, then only I thought of doing the same of SNS FIFO. I will change it.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -797,15 +797,24 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 	var publishers template.PublishOpts
 	// convert the topics to template Topics
 	for _, topic := range topics {
+		var fifoConfig *template.FIFOTopicConfig
+		if !topic.FIFO.IsEmpty() {
+			fifoConfig = &template.FIFOTopicConfig{
+				Enable: topic.FIFO.Enable,
+			}
```
efekarakus: should we keep `fifoConfig` as `nil` if the `topic.FIFO.Enable == false`? Otherwise as far as I can tell we ignore the config:
```yaml
fifo: false
```
And still enable FIFO in CloudFormation.

```go
if topic.FIFO.IsEnabled() {
   fifoConfig = &template.FIFOTopicConfig{}
   if !topic.FIFO.Advanced.IsEmpty() { 
      // Content dedupe stuff...
   }
}
```

internal/pkg/manifest/validate.go
```diff
@@ -1412,10 +1411,26 @@ func (p PublishConfig) validate() error {
...
 func (t Topic) validate() error {
-	if t.Type != nil && !contains(aws.StringValue(t.Type), validTopicsTypeValues) {
-		return fmt.Errorf(`"type" value %q is not allowed; must be one of %s`, aws.StringValue(t.Type), english.WordSeries(validTopicsTypeValues, "or"))
+	if err := validatePubSubName(aws.StringValue(t.Name)); err != nil {
+		return err
+	}
+	if err := t.FIFO.validate(); err != nil {
+		return err
+	}
+	return nil
```
```suggestion
	return t.FIFO.validate()
```

internal/pkg/template/workload.go
```diff
@@ -399,6 +399,12 @@ type Topic struct {
 	Svc       string
 }
 
+// Fifo holds configuration needed if the topic is FIFO.
+type FIFOTopicConfig struct {
+	Enable                    *bool
```
efekarakus: (same feedback as above) We shouldn't need this field

internal/pkg/template/workload.go
```diff
@@ -692,5 +698,8 @@ func contains(list []string, s string) bool {
 
 // ARN determines the arn for a topic using the SNSTopic name and account information
 func (t Topic) ARN() string {
+	if t.FIFOTopicConfig != nil {
+		return fmt.Sprintf(snsARNPattern, t.Partition, t.Region, t.AccountID, t.App, t.Env, t.Svc, aws.StringValue(t.Name)+".fifo")
```
efekarakus, line 725: TIL that it's okay to have a dot in the ARN

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: Makes sense to me. The only problem I can see is we are putting advanced config inside of `topic.FIFO.IsEnabled()`. This can cause a problem when customers have advanced config and this code is unreachable in that context because `topic.FIFO.IsEnabled()` will be false. 

How about?
```
if !topic.FIFO.IsEmpty() {
			if topic.FIFO.IsEnabled() {
				fifoConfig = &template.FIFOTopicConfig{}
			}
			if !topic.FIFO.Advanced.IsEmpty() {
				fifoConfig = &template.FIFOTopicConfig{
					ContentBasedDeduplication: topic.FIFO.Advanced.ContentBasedDeduplication,
				}
			}
		}
```

internal/pkg/template/templates/workloads/partials/cf/publish.yml
```diff
@@ -2,17 +2,18 @@
 {{- range $topic := .Publish.Topics}}
 {{logicalIDSafe $topic.Name}}SNSTopic:
   Metadata:
-    'aws:copilot:description':  {{- if eq $topic.Type "fifo" }} 'A FIFO SNS topic to broadcast {{$topic.Name}} events' {{- else}} 'A SNS topic to broadcast {{$topic.Name}} events'{{- end}}
+    'aws:copilot:description': {{ if or $topic.FIFOTopicConfig }} 'A FIFO SNS topic to broadcast {{$topic.Name}} events' {{ else}} 'A SNS topic to broadcast {{$topic.Name}} events'{{ end}}
```
Lou1415926: we probably don't need the `or`:
```suggestion
    'aws:copilot:description': {{ if $topic.FIFOTopicConfig }} 'A FIFO SNS topic to broadcast {{$topic.Name}} events' {{ else}} 'A SNS topic to broadcast {{$topic.Name}} events'{{ end}}
```

internal/pkg/template/workload.go
```diff
@@ -399,6 +399,12 @@ type Topic struct {
 	Svc       string
 }
 
+// Fifo holds configuration needed if the topic is FIFO.
+type FIFOTopicConfig struct {
+	Enable                    *bool
+	ContentBasedDeduplication *bool
```
Lou1415926, line 418: I wonder what'd happen in this case:
1. First deploy the CFN template with `ContentBasedDeduplication: true` field
2. Then deploy the CFN template without the field at all

Do they set `ContentBasedDeduplication` to `false`, or keep it as `true`?

paragbhingre: If `ContentBasedDeduplication` is not there in the template then CFN sets it to false by default

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -797,15 +797,26 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 	var publishers template.PublishOpts
 	// convert the topics to template Topics
 	for _, topic := range topics {
+		var fifoConfig *template.FIFOTopicConfig
+		if !topic.FIFO.IsEmpty() {
+			if topic.FIFO.IsEnabled() {
```
efekarakus: I feel like the difference between `IsEmpty` and `IsEnabled` is a little subtle, what if the implementation of `IsEnabled` was like this:
```go
func (cfg FIFOAdvanceConfigOrBool) IsEnabled() bool {
   return aws.BoolValue(cfg.Enabled) || !cfg.Advanced.IsEmpty()
}
```
and then we can replace the `IsEmpty` check here with `IsEnabled`

efekarakus (approved):
Looks good just a tiny request, then we can remove the label

internal/pkg/template/workload.go
Lou1415926: Then what about we have 
```
ContentBasedDeduplication bool
```

and then in template we have
```
ContentBasedDeduplication: {{ .ContentBasedDeduplication }}
```

If CFN always sets it to `false` if it's not there (instead of keeping the existing config) then this should have the same effect as the current implementation but the logic would be simpler.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -798,12 +798,12 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 	// convert the topics to template Topics
 	for _, topic := range topics {
 		var fifoConfig *template.FIFOTopicConfig
-		if !topic.FIFO.IsEmpty() {
-			fifoConfig = &template.FIFOTopicConfig{
-				Enable: topic.FIFO.Enable,
-			}
+		if topic.FIFO.IsEnabled() {
+			fifoConfig = &template.FIFOTopicConfig{}
```
Lou1415926, line 808: nit: perhaps we could
```suggestion
		if topic.FIFO.IsEnabled() {
			fifoConfig = &template.FIFOTopicConfig{
			         ContentBasedDeduplication: topic.FIFO.Advanced.ContentBasedDeduplication,
			}
		}
```

Lou1415926 (approved):
LGTM! just a nit. Adding the DNM in case you'd like to address!

PR #4019, opened 2022-09-21
chore: add `--permissions-boundary` flag and update docs

huanjani:
Related: #3969, #2986.

### DIFF
--- a/internal/pkg/cli/app_init.go
+++ b/internal/pkg/cli/app_init.go
@@ -330,6 +330,7 @@ An application is a collection of containerized services that operate together.`
 		}),
 	}
 	cmd.Flags().StringVar(&vars.domainName, domainNameFlag, "", domainNameFlagDescription)
+	cmd.Flags().StringVar(&vars.permissionsBoundary, permissionsBoundaryFlag, "", permissionsBoundaryFlagDescription)
 	cmd.Flags().StringToStringVar(&vars.resourceTags, resourceTagsFlag, nil, resourceTagsFlagDescription)
 	return cmd
 }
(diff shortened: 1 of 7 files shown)


internal/pkg/cli/app_init.go
```diff
@@ -330,6 +330,7 @@ An application is a collection of containerized services that operate together.`
 		}),
 	}
 	cmd.Flags().StringVar(&vars.domainName, domainNameFlag, "", domainNameFlagDescription)
+	cmd.Flags().StringVar(&vars.permissionsBoundary, permissionsBoundaryFlag, "", permissionsBoundaryFlagDescription)
```
efekarakus, line 335: Wanna add an example as well in the help menu? :)

efekarakus (approved):

internal/pkg/cli/app_init.go
```diff
@@ -315,6 +315,8 @@ An application is a collection of containerized services that operate together.`
   /code $ copilot app init test
   Create a new application with an existing domain name in Amazon Route53.
   /code $ copilot app init --domain example.com
+  Create a new application with an existing IAM policy for roles with a permissions boundary.
```
Lou1415926: I probably don't know what I'm talking about but what do you think of this alternative
```suggestion
  Create a new application with an existing IAM policy as the permissions boundary for roles.
```
My only concern is that people parsing it as "...with an existing IAM policy for roles" i.e. the roles' permissions are granted by the IAM policy, instead of "limited/bound".

Lou1415926 (approved):
LGTM! Adding a label just in case you prefer nit my suggestion !

internal/pkg/cli/app_init.go
huanjani: Your version is much better! Thanks!

PR #4021, opened 2022-09-21
feat: enable e2e encryption by setting TargetGroup protocol to HTTPS

### LINKED ISSUE #1582: TargetGroup for Port 443 uses HTTP Protocol instead of HTTPS
What changes do I need to make to the manifest to create a targetgroup that listens on port 443 protocol HTTPS instead of on port 443 protocol HTTP?

I created my app using `copilot app init --domain my.example.com`

Here's my manifest:

```yml
# The manifest for the "php-apache" service.
# Read the full specification for the "Load Balanced Web Service" type at:
#  https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/

# Your service name will be used in naming your resources like log groups, ECS services, etc.
name: php-apache
# The "architecture" of the service you're running.

image:
  # Docker build arguments. You can specify additional overrides here. Supported: dockerfile, context, args.
  build: apache/Dockerfile
  # Port exposed through your container to route traffic to it.
  port: 443

http:
  # Requests to this path will be forwarded to your service.
  # To match all requests you can use the "/" path.
  path: '/'
  # You can specify a custom health check path. The default is "/"
  healthcheck: '/'
  # You can enable sticky sessions.
  # stickiness: true

# Number of CPU units for the task.
cpu: 1024
memory: 4096
# Number of tasks that should be running in your service.
count: 2

# Optional fields for more advanced use-cases.
#
#variables:                    # Pass environment variables as key value pairs.
#  LOG_LEVEL: info
#
#secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store.
#  GITHUB_TOKEN: GITHUB_TOKEN  # The key is the name of the environment variable, the value is the name of the SSM parameter.

#environments:
#  test:

### LINKED ISSUE #2981: Enable HTTPS connection between ALB and ECS Task ENI
We're deploying a Load-balanced web service. It consists of a Rails app container with an Apache Modproxy based reverse proxy container running as a sidecar.

In our current deployment, Apache is configured to server HTTPS traffic. To deploy with Copilot, we currently need to disable this and serve HTTP traffic instead. This is likely to raise questions from our auditors about whether we truly still have end-to-end encryption. It also means that we can't easily use the same image in an environment where there isn't something else to terminate the SSL connection.

Ideally, we could indicate in our `manifest.yml` that a service is listening on port 443 and it would be assumed to be expecting HTTPS. Alternatively, an explicit flag would be fine.

As I understand it, we can use any self-signed certificate because the ALB won't verify it. We would use Copilot's existing secrets feature to supply the certificate and key to our Apache sidecar. Alternatively, we could generate a certificate in the container on startup.

We do not need HTTP in addition to HTTPS. It would be sufficient for the ALB to redirect HTTP -> HTTPS.

To make this feature easy for anyone to incorporate into their project, you could provide a Docker image that acts as a reverse proxy that generates its own certificate. It might be documented as follows:

Copilot terminates your HTTPS connection at the ALB. The connection from the ALB to your container is HTTP (unencrypted). This is adequate security for most purposes. If you have audit requirements that require end-to-end encryption, you may achieve this by adding an HTTPS enabled reverse-proxy sidecar. Update your manifest to include:

```yaml
http:
    targetContainer: 'reverse_proxy'

sidecars:
  reverse_proxy:
    port: 443
    protocol: HTTPS
    image: "copilot-reverse-proxy-sidecar:latest"
```

efekarakus:
When the target container's port is set to 443 for Load Balanced Web Services or Backend Services with internal ALB then we set the TargetGroup's protocols to HTTPS allowing for e2e encryption even within the VPC.

Resolves #1582, resolves #2981

### DIFF
--- a/internal/pkg/template/workload.go
+++ b/internal/pkg/template/workload.go
@@ -203,6 +203,17 @@ type LogConfigOpts struct {
 	Secrets        map[string]Secret
 }
 
+// HTTPTargetContainer represents the target group of a load balancer that points to a container.
+type HTTPTargetContainer struct {
+	Name string // Name of the container.
+	Port string // Port of the container.
+}
+
+// IsHTTPS returns true if the target container's port is 443.
+func (tg HTTPTargetContainer) IsHTTPS() bool {
+	return tg.Port == "443"
+}
+
 // HTTPHealthCheckOpts holds configuration that's needed for HTTP Health Check.
 type HTTPHealthCheckOpts struct {
 	HealthCheckPath     string
@@ -524,6 +535,7 @@ type WorkloadOpts struct {
 	// Additional options for service templates.
 	WorkloadType            string
 	HealthCheck             *ContainerHealthCheck
+	HTTPTargetContainer     HTTPTargetContainer
 	HTTPHealthCheck         HTTPHealthCheckOpts
 	DeregistrationDelay     *int64
 	AllowedSourceIps        []string
--- a/internal/pkg/deploy/cloudformation/stack/backend_svc.go
+++ b/internal/pkg/deploy/cloudformation/stack/backend_svc.go
@@ -141,6 +141,7 @@ func (s *BackendService) Template() (string, error) {
 		allowedSourceIPs = append(allowedSourceIPs, string(ipNet))
 	}
 
+	targetContainerName, targetContainerPort := s.httpLoadBalancerTarget()
 	content, err := s.parser.ParseBackendService(template.WorkloadOpts{
 		AppName:            s.app,
 		EnvName:            s.env,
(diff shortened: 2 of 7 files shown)


efekarakus:
The actual change is incredibly tiny, however I cleaned up massively our `_Template()` tests which should make it hopefully easier to add new test cases as we add fields to the manifest

internal/pkg/template/workload.go
```diff
@@ -203,6 +203,17 @@ type LogConfigOpts struct {
 	Secrets        map[string]Secret
 }
 
+// HTTPTargetContainer represents the target group of a load balancer that points to a container.
+type HTTPTargetContainer struct {
+	Name string // Name of the container.
```
Lou1415926, line 208: qq: where is `Name` used?

efekarakus: nowhere atm, I just did it for consistency but I can remove it since it's not used 😅

Lou1415926 (approved):

huanjani (approved):
LGTM!

PR #4027, opened 2022-09-23
docs: add manifest fields for TLS connections from ALB to Fargate tasks

efekarakus:

### DIFF
--- a/internal/pkg/template/workload.go
+++ b/internal/pkg/template/workload.go
@@ -205,7 +205,6 @@ type LogConfigOpts struct {
 
 // HTTPTargetContainer represents the target group of a load balancer that points to a container.
 type HTTPTargetContainer struct {
-	Name string // Name of the container.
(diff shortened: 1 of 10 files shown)


internal/pkg/template/workload.go
```diff
@@ -205,7 +205,6 @@ type LogConfigOpts struct {
 
 // HTTPTargetContainer represents the target group of a load balancer that points to a container.
 type HTTPTargetContainer struct {
-	Name string // Name of the container.
```
iamhopaul123, line 208: Maybe remove `HTTPTargetContainer` and keep `HTTPTargetPort` only?

efekarakus: I'll keep it as is for now as I've validated this template 😝

dannyrandall (approved):

iamhopaul123 (approved):

PR #4033, opened 2022-09-26
chore: fix for same topic names for standard and fifo use case

paragbhingre:

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
@@ -31,6 +31,8 @@ command: "shine on you"
 publish:
   topics:
     - name: customers
+    - name: mytopic
+      fifo: true
 
 network:
   vpc:
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -804,9 +804,9 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 	// convert the topics to template Topics
 	for _, topic := range topics {
 		var fifoConfig *template.FIFOTopicConfig
-		topicName := aws.StringValue(topic.Name)
+		//topicName := aws.StringValue(topic.Name)
 		if topic.FIFO.IsEnabled() {
-			topicName += ".fifo"
+			//topicName += ".fifo"
 			fifoConfig = &template.FIFOTopicConfig{}
 			if !topic.FIFO.Advanced.IsEmpty() {
 				fifoConfig = &template.FIFOTopicConfig{
@@ -815,7 +815,7 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 			}
 		}
 		publishers.Topics = append(publishers.Topics, &template.Topic{
-			Name:            aws.String(topicName),
+			Name:            topic.Name,
 			FIFOTopicConfig: fifoConfig,
 			AccountID:       accountID,
 			Partition:       partition.ID(),
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -804,9 +804,7 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 	// convert the topics to template Topics
 	for _, topic := range topics {
 		var fifoConfig *template.FIFOTopicConfig
-		//topicName := aws.StringValue(topic.Name)
 		if topic.FIFO.IsEnabled() {
-			//topicName += ".fifo"
 			fifoConfig = &template.FIFOTopicConfig{}
 			if !topic.FIFO.Advanced.IsEmpty() {
 				fifoConfig = &template.FIFOTopicConfig{
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -804,7 +804,9 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 	// convert the topics to template Topics
 	for _, topic := range topics {
 		var fifoConfig *template.FIFOTopicConfig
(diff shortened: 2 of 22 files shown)


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -804,7 +804,9 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 	// convert the topics to template Topics
 	for _, topic := range topics {
 		var fifoConfig *template.FIFOTopicConfig
+		topicName := aws.StringValue(topic.Name)
 		if topic.FIFO.IsEnabled() {
+			topicName += ".fifo"
```
efekarakus: So we only had to fix for SNS but not SQS right because we have:
https://github.com/aws/copilot-cli/blob/dd8dc3356ef47130587fb6092e7a29a716899f45/internal/pkg/manifest/worker_svc.go#L245-L258

Can we not do a similar thing for SNS? like updating the existing `Publish` method to handle this logic. Otherwise, we're being inconsistent with where this augmentation happens

paragbhingre: I totally agree to your point. I have changed the implementation, do let me know how do you feel about new implementation.

paragbhingre: I think I will need to make the similar change for the Job as well. I have done it for all the service types. I will do it for the job as well in a bit.

internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
```diff
@@ -31,6 +31,8 @@ command: "shine on you"
 publish:
   topics:
     - name: customers
+    - name: mytopic
```
Lou1415926, line 34: nit: if we add this in the environment override section, e.g. for the `test` environment override, then we wouldn't need to touch `rdws-prod.stack.yml`. It's fine keeping it this way - just a small idea of making the PR smaller.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -804,9 +804,9 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 	// convert the topics to template Topics
 	for _, topic := range topics {
 		var fifoConfig *template.FIFOTopicConfig
-		topicName := aws.StringValue(topic.Name)
+		//topicName := aws.StringValue(topic.Name)
```
Lou1415926: 👀

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -804,7 +804,9 @@ func convertPublish(topics []manifest.Topic, accountID, region, app, env, svc st
 	// convert the topics to template Topics
 	for _, topic := range topics {
 		var fifoConfig *template.FIFOTopicConfig
+		//topicName := aws.StringValue(topic.Name)
```
efekarakus: Can be deleted now 😂

internal/pkg/manifest/backend_svc.go
```diff
@@ -100,7 +100,15 @@ func (s *BackendService) Port() (port uint16, ok bool) {
 
 // Publish returns the list of topics where notifications can be published.
 func (s *BackendService) Publish() []Topic {
-	return s.BackendServiceConfig.PublishConfig.Topics
+	var pubs []Topic
+	for _, topic := range s.BackendServiceConfig.PublishConfig.Topics {
```
efekarakus: nit: instead of replicating the same body across all service types, I think in this scenario the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle is valid.

We could move this method to 

```
// publishedTopics augments each  topic to append the .fifo suffix if the topic is FIFO, and returns their configuration.
func (cfg PublishConfig) publishedTopics() []Topic {
  ...
}
```

internal/pkg/manifest/backend_svc.go
```diff
@@ -100,7 +100,15 @@ func (s *BackendService) Port() (port uint16, ok bool) {
...
+		tempTopic := topic
+		if topic.FIFO.IsEnabled() {
+			tempTopic.Name = aws.String(aws.StringValue(topic.Name) + ".fifo")
+		}
+		pubs = append(pubs, tempTopic)
```
Lou1415926: nit:
```suggestion
		if topic.FIFO.IsEnabled() {
			topic.Name = aws.String(aws.StringValue(topic.Name) + ".fifo")
		}
		pubs = append(pubs, topic)
```
Because `topic` is passed by value here in the `range` over. Changing `topic.Name` won't affect `s.BackendServiceConfig.PublishConfig.Topics[idx].Name`.

internal/pkg/template/workload.go
```diff
@@ -721,8 +721,5 @@ func contains(list []string, s string) bool {
 
 // ARN determines the arn for a topic using the SNSTopic name and account information
 func (t Topic) ARN() string {
-	if t.FIFOTopicConfig != nil {
-		return fmt.Sprintf(snsARNPattern, t.Partition, t.Region, t.AccountID, t.App, t.Env, t.Svc, aws.StringValue(t.Name)+".fifo")
-	}
```
efekarakus, line 725: nice!

Lou1415926 (commented):
LGTM just some small nits!!

internal/pkg/manifest/workload.go
```diff
@@ -819,3 +819,14 @@ func placementStringP(p PlacementString) *PlacementString {
 	placement := p
 	return &placement
 }
+
+func (cfg PublishConfig) publishedTopics() []Topic {
+	var pubs []Topic
+	for _, topic := range cfg.Topics {
```
efekarakus: nit: since the topic length is known ahead of time we can speed things up and be a bit more mem efficient slightly like this:
```
pubs := make([]Topic, len(cfg.Topics)
for i, topic := range cfg.Topics {
   // if... { } 
   pubs[i] = topic[i]
}
return pubs
```

efekarakus (approved):

Lou1415926 (approved):

PR #4035, opened 2022-09-26
chore: validate permissions boundary policy name so as to fail faster if invalid

huanjani:


paragbhingre (approved):

internal/pkg/aws/iam/iam.go
```diff
@@ -98,6 +99,21 @@ func (c *IAM) CreateECSServiceLinkedRole() error {
 	return nil
 }
 
+// ListPermBoundPolicyNames returns a list of permissions boundary policy names.
+func (c *IAM) ListPermBoundPolicyNames() ([]*string, error) {
+	policies, err := c.client.ListPolicies(&iam.ListPoliciesInput{
```
iamhopaul123: We might need to handle pagination here: https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#ListPoliciesOutput

internal/pkg/aws/iam/iam.go
```diff
@@ -98,6 +99,21 @@ func (c *IAM) CreateECSServiceLinkedRole() error {
...
+		PolicyUsageFilter: aws.String("PermissionsBoundary"),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("list IAM policies: %w", err)
+	}
+	var policyNames = make([]*string, len(policies.Policies))
```
iamhopaul123: Why not `[]string` 🤔

internal/pkg/aws/iam/iam.go
huanjani: Yeah. I was hoping the `PolicyUsageFilter: aws.String("PermissionsBoundary")` filter would reduce the number to a reasonable amount, but I just tested it and you're right. Will add!

internal/pkg/aws/iam/iam.go
huanjani: Changed.

internal/pkg/aws/iam/iam.go
iamhopaul123: yeah basically we need to do pagination for every list API we call as long as the response is paginated and we return the full list of it (instead of grabbing first xyz items)

internal/pkg/aws/iam/iam.go
```diff
@@ -98,6 +99,32 @@ func (c *IAM) CreateECSServiceLinkedRole() error {
...
 }
 
+// ListPermBoundPolicyNames returns a list of permissions boundary policy names.
+func (c *IAM) ListPermBoundPolicyNames() ([]string, error) {
+	var policies []*iam.Policy
+	var marker *string
+	for {
+		output, err := c.client.ListPolicies(&iam.ListPoliciesInput{
+			Marker:            marker,
+			Scope:             aws.String("Local"),
```
iamhopaul123, line 109: How is this config related to "returns a list of permissions boundary policy names."?

internal/pkg/cli/app_init.go
```diff
@@ -234,6 +242,19 @@ func (o *initAppOpts) validateAppName(name string) error {
...
+	IAMPolicies, err := o.iam.ListPermBoundPolicyNames()
+	if err != nil {
+		return fmt.Errorf("list permissions boundary policies: %w", err)
+	}
+	for _, policy := range IAMPolicies {
+		if policy == policyName {
+			return nil
+		}
+	}
+	return fmt.Errorf("IAM policy '%s' not found in this account", policyName)
```
iamhopaul123:
```suggestion
	return fmt.Errorf("IAM policy %q not found in this account", policyName)
```

internal/pkg/aws/iam/iam.go
huanjani: changing the wording

iamhopaul123 (approved):

PR #4037, opened 2022-09-27
chore: add permission boundary policy name to progress tracker if present

huanjani:

### DIFF
--- a/internal/pkg/template/templates/app/app.yml
+++ b/internal/pkg/template/templates/app/app.yml
@@ -30,11 +30,7 @@ Conditions:
 Resources:
   AdministrationRole:
     Metadata:
-      {{- if .PermissionsBoundary}}
-      'aws:copilot:description': 'A StackSet admin role with permissions boundary {{.PermissionsBoundary}} assumed by CloudFormation to manage regional stacks'
-      {{- else}}
-      'aws:copilot:description': 'A StackSet admin role assumed by CloudFormation to manage regional stacks'
-      {{- end}}
+      'aws:copilot:description': 'A StackSet admin role {{- if .PermissionsBoundary}} with permissions boundary {{.PermissionsBoundary}} {{- end}} assumed by CloudFormation to manage regional stacks'
     Type: AWS::IAM::Role
     Properties:
       RoleName: !Ref AdminRoleName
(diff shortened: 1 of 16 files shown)


internal/pkg/template/templates/app/app.yml
```diff
@@ -30,7 +30,11 @@ Conditions:
 Resources:
   AdministrationRole:
     Metadata:
+      {{- if .PermissionsBoundary}}
+      'aws:copilot:description': 'A StackSet admin role with permissions boundary {{.PermissionsBoundary}} assumed by CloudFormation to manage regional stacks'
```
paragbhingre: Just a one-liner alternative as there are many places where similar code is there. 
```suggestion
      'aws:copilot:description':  {{ if .PermissionsBoundary}} 'A StackSet admin role with permissions boundary {{.PermissionsBoundary}} assumed by CloudFormation to manage regional stacks' {{ else}} 'A StackSet admin role assumed by CloudFormation to manage regional stacks' {{ end }}
```

paragbhingre (commented):
Looks good, just one alternative.

internal/pkg/template/templates/environment/partials/cfn-execution-role.yml
```diff
@@ -2,7 +2,11 @@
 # See #1533.
 CloudformationExecutionRole:
   Metadata:
+    {{- if .PermissionsBoundary}}
+    'aws:copilot:description': 'An IAM Role with permissions boundary {{.PermissionsBoundary}} for AWS CloudFormation to manage resources'
+    {{- else}}
```
Lou1415926: nit: alternatively

```suggestion
    'aws:copilot:description': 'An IAM Role {{if .PermissionsBoundary -}} with permissions boundary {{.PermissionsBoundary}} {{- end}} for AWS CloudFormation to manage resources'
```

Lou1415926 (approved):
DNM label in case you want to address!

internal/pkg/template/templates/app/app.yml
iamhopaul123: Can we possibly make it a function that takes a string and output a processed string when `{{ if .PermissionsBoundary}}`? The benefit is that we don't need to repeat ourselves whenever we add new roles or want to modify the description.

huanjani: Yeah, I'll think about it and possibly do it in a different PR!

internal/pkg/template/templates/app/app.yml
```diff
@@ -30,7 +30,7 @@ Conditions:
 Resources:
   AdministrationRole:
     Metadata:
-      'aws:copilot:description': 'A StackSet admin role assumed by CloudFormation to manage regional stacks'
+      'aws:copilot:description': 'A StackSet admin role {{- if .PermissionsBoundary}} with permissions boundary {{.PermissionsBoundary}} {{- end}} assumed by CloudFormation to manage regional stacks'
```
iamhopaul123, line 33: This seems to repeat a lot...and also for upcoming new roles we need this as well. Maybe consider making it a template function?

iamhopaul123: oh nvm...

iamhopaul123 (approved):

PR #4043, opened 2022-09-28
test(e2e): FIFO SNS and SQS

paragbhingre:


e2e/worker/worker_test.go
```diff
@@ -78,6 +78,12 @@ var _ = Describe("Worker Service E2E Test", func() {
 publish:
   topics:
   - name: events
+  - name: mytopic
```
iamhopaul123, line 81: does the e2e test work with this duplication...
```suggestion
```

iamhopaul123 (commented):
How do we make sure FIFO really work (like the messages are passed through at strict FIFO pattern) without updating the service code 🤔

e2e/worker/worker_test.go
```diff
@@ -160,14 +166,43 @@ publish:
...
-				TopicSubscriptions: []string{fmt.Sprintf("%s:%s", workerServiceName, "processed-msg-count")},
+				Name:            counterServiceName,
+				SvcType:         "Worker Service",
+				Dockerfile:      "./counter/Dockerfile",
+				NoSubscriptions: true,
 			})
 			Expect(err).NotTo(HaveOccurred())
 		})
 
+		It("adds a topic and queues to push to the service", func() {
```
Lou1415926, line 177: An alternative would be to write add a `manifest.yml` file under `copilot/worker/manifest.yml`, like what we do with `multi-svc-app`.

Lou1415926 (commented):
same question as PH's: do we want to test that the messages are indeed FIFO?

Lou1415926:
Closing this for now. We can reopen it in the future if it's deemed necessary to include FIFO in e2e tests.

PR #4046, opened 2022-09-29
feat: add service connect mft

iamhopaul123:


internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/simple-params.json
```diff
@@ -10,7 +10,9 @@
 		"TaskCount": "1",
 		"LogRetention": "30",
 		"ContainerPort": "8080",
-		"EnvFileARN": ""
+		"EnvFileARN": "",
+    "TargetContainer": "simple-backend",
+    "TargetPort": "8080"
```
efekarakus: are my eyes deceiving me 🧓 or is this not formatted properly?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -496,6 +496,18 @@ func convertExecuteCommand(e *manifest.ExecuteCommand) *template.ExecuteCommandO
 	return &template.ExecuteCommandOpts{}
 }
 
+func convertServiceConnect(s manifest.ServiceConnect) *template.ServiceConnect {
+	if s.EnableServiceConnect != nil && *s.EnableServiceConnect {
```
```suggestion
	if aws.BoolValue(s.EnableServiceConnect) {
```

internal/pkg/manifest/validate.go
```diff
@@ -193,6 +195,19 @@ func (b BackendService) validate() error {
...
+		mainContainerName: aws.StringValue(b.Name),
+		mainContainerPort: b.ImageConfig.Port,
+		targetContainer:   b.RoutingRule.GetTargetContainer(),
+		sidecarConfig:     b.Sidecars,
+	}); err != nil {
+		return fmt.Errorf("validate HTTP load balancer target: %w", err)
+	}
+	if b.BackendServiceConfig.Network.Connect.enabled() {
+		if b.RoutingRule.GetTargetContainer() == nil && b.ImageConfig.Port == nil {
+			return fmt.Errorf(`cannot enable "network.connect" when no port exposed`)
```
efekarakus, line 208:
```suggestion
			return fmt.Errorf(`cannot enable "network.connect" when no "image.port" is exposed`)
```

internal/pkg/manifest/workload.go
```diff
@@ -473,6 +475,49 @@ func (c *NetworkConfig) requiredEnvFeatures() []string {
 	return nil
 }
 
+// ServiceConnect represents ECS Service Connect.
+type ServiceConnect struct {
```
efekarakus: What do you think of enabling service connect always by default such that:
1. If LBWS, it's `true` by default. If they want to turn it off, then they can write:
   ```yaml
   network:
      connect: false
   ```
2. For Backend Service, it will be set to `true` if the target container has an exposed port.

Basically, we'll always enable it from now on and people that want to opt-out will have to write `connect: false` in the manifest?

internal/pkg/manifest/workload.go
```diff
@@ -473,6 +475,49 @@ func (c *NetworkConfig) requiredEnvFeatures() []string {
...
+		s.EnableServiceConnect = nil
+		return nil
+	}
+	if err := value.Decode(&s.EnableServiceConnect); err != nil {
+		return errUnmarshalServiceConnectOpts
+	}
+	return nil
+}
+
+// ServiceConnectArgs represents the configuration for ECS Service Connect.
```
efekarakus: the comment between `ServiceConnect` and `ServiceConnectArgs` is pretty much the same 😅 can we rework them please?

internal/pkg/manifest/workload.go
```diff
@@ -473,6 +475,49 @@ func (c *NetworkConfig) requiredEnvFeatures() []string {
...
+type ServiceConnectArgs struct {
+	Alias *string
+}
+
+// IsEmpty returns if ServiceConnectArgs is empty.
```
efekarakus:
```suggestion
// IsEmpty returns true if there are no advanced service connect configurations set.
```

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
```diff
@@ -36,15 +36,15 @@ CapacityProviderStrategy:
 {{- if .ServiceConnect }}
 ServiceConnectConfiguration:
   Enabled: True
-  Namespace: {{.ServiceConnect.Namespace}}
+  Namespace: {{.ServiceDiscoveryEndpoint}}
```
efekarakus, line 39: oh it's okay if this value is the same for both SD and SC?

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
```diff
@@ -43,8 +43,8 @@
 {{- end}}
 {{- end}} {{/* end if eq .WorkloadType "Load Balanced Web Service"*/}}
 {{- if eq .WorkloadType "Backend Service"}}
-{{- if .ServiceConnect }} # remove when release
-  PortMappings: !If [ExposePort, [{ContainerPort: !Ref ContainerPort, Name: !Ref WorkloadName}], !Ref "AWS::NoValue"]
+{{- if and .ServiceConnect (eq .HTTPTargetContainer.Name .WorkloadName)}} # remove when release
+  PortMappings: !If [ExposePort, [{ContainerPort: !Ref ContainerPort, Name: target}], !Ref "AWS::NoValue"]
```
efekarakus, line 49: this `Name` field is unique within the task definition across containers right? like it would make no sense to have two `'target'` port mapping in different container definitions?

internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/simple-params.json
iamhopaul123: lol it's not but json doesn't care. i'll fix it

internal/pkg/manifest/validate.go
iamhopaul123: It's actually not "image.port". It is a valid one if no "image.port" but they route traffic to one of the sidecar container and set it to be the target container.

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
iamhopaul123: Yeah this is what they worked on for backward compatiblity.

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
iamhopaul123: Yeah that's true. They can't have two containers with the same name. and each port mapping should have a unique name.

internal/pkg/manifest/workload.go
iamhopaul123: ok addressed!

internal/pkg/manifest/backend_svc.go
```diff
@@ -98,6 +98,22 @@ func (s *BackendService) Port() (port uint16, ok bool) {
...
+		return *s.Network.Connect.EnableServiceConnect
+	}
+	if !s.Network.Connect.ServiceConnectArgs.isEmpty() {
+		return true
+	}
+	// Try our best to enable Service Connect by default.
+	if s.BackendServiceConfig.ImageConfig.Port != nil ||
+		s.BackendServiceConfig.RoutingRule.TargetContainer != nil {
+		return true
+	}
```
efekarakus, line 114: Ah I get why now you moved this method to be part of `*BackendService` instead of `ServiceConnect` because you need access to the other fields in the manifest like target container.

What do you think of changing the method name to be more generic, for example:
```go
// IsDiscoverable returns true if the service publishes a port that can be reached by other services in the VPC.
func (svc *BackendService) IsDiscoverable() bool {
   // This logic.
}

func (svc *LoadBalancedWebService) IsDiscoverable() bool {
   // returns true if service connect is enabled.
}

// IsDiscoverable is  false for workers as queue processing services don't accept network traffic.
func (svc *WorkerService) IsDiscoverable() bool { return false }

// IsDiscoverable is false as ephemeral tasks cannot be reached consistently.
func (job *ScheduledJob) IsDiscoverable() bool { return false }
```

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -169,6 +169,14 @@ func (s *LoadBalancedWebService) Port() (port uint16, ok bool) {
 	return aws.Uint16Value(s.ImageConfig.Port), true
 }
 
+// ServiceConnectEnabled returns if ServiceConnect is enabled or not.
+func (s *LoadBalancedWebService) ServiceConnectEnabled() bool {
+	if s.Network.Connect.EnableServiceConnect != nil && !*s.Network.Connect.EnableServiceConnect {
+		return false
+	}
+	return true
```
efekarakus: I think this is equivalent:
```suggestion
	return s.Network.Connect.EnableServiceConnect == nil || *s.Network.Connect.EnableServiceConnect
```

efekarakus (approved):
Thanks boss looks great, just a refactor request and then :shipit: 

Feel free to remove the label once addressed

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -241,18 +246,18 @@ func (s *BackendService) Parameters() ([]*cloudformation.Parameter, error) {
...
 			ParameterValue: aws.String(s.rc.EnvFileARN),
 		},
+		{
+			ParameterKey:   aws.String(WorkloadTargetContainerParamKey),
+			ParameterValue: targetContainer,
+		},
+		{
+			ParameterKey:   aws.String(WorkloadTargetPortParamKey),
+			ParameterValue: targetPort,
+		},
```
dannyrandall: haha https://github.com/aws/copilot-cli/pull/4047/files#diff-5820762069c47f3a63a9e5716ef26ede183e4c2de7b645729a26bd5f6a35829fR244-R251

internal/pkg/manifest/backend_svc.go
iamhopaul123: Hmm `ServiceConnectEnabled` might still be better since it isn't really whether the service is discoverable within the VPC because they could use SD.

internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-nlb-manifest.yml
```diff
@@ -10,6 +10,8 @@ image:
   build: ./Dockerfile
   port: 80
 http: false
+network:
+  connect: false
```
dannyrandall, line 14: maybe we should add a test case where it's true?

internal/pkg/deploy/cloudformation/stack/worker_svc.go
```diff
@@ -5,9 +5,10 @@ package stack
 
 import (
 	"fmt"
-	"github.com/aws/copilot-cli/internal/pkg/config"
 	"strings"
 
+	"github.com/aws/copilot-cli/internal/pkg/config"
+
```
dannyrandall: why is there a space here 🤔

internal/pkg/manifest/backend_svc.go
```diff
@@ -98,6 +98,22 @@ func (s *BackendService) Port() (port uint16, ok bool) {
 	return aws.Uint16Value(value), true
 }
 
+// ServiceConnectEnabled returns if ServiceConnect is enabled or not.
```
dannyrandall, line 101:
```suggestion
// ServiceConnectEnabled returns if ServiceConnect is enabled or not. Unless explicitly disabled in the manifest or if no server is configured we default to true.
```

internal/pkg/manifest/validate.go
```diff
@@ -1609,14 +1638,17 @@ func validateTargetContainer(opts validateTargetContainerOpts) error {
...
 		return nil
 	}
 	sidecar, ok := opts.sidecarConfig[targetContainer]
 	if !ok {
-		return fmt.Errorf("target container %s doesn't exist", targetContainer)
+		return fmt.Errorf("target container %q doesn't exist", targetContainer)
 	}
 	if sidecar.Port == nil {
-		return fmt.Errorf("target container %s doesn't expose any port", targetContainer)
+		return fmt.Errorf("target container %q doesn't expose any port", targetContainer)
```
dannyrandall: probably more accurate as we only support 1 (same above)
```suggestion
		return fmt.Errorf("target container %q doesn't expose a port", targetContainer)
```

internal/pkg/manifest/validate_test.go
```diff
@@ -580,6 +580,37 @@ func TestBackendService_validate(t *testing.T) {
 			},
 			wantedErrorMsgPrefix: `validate "publish": `,
 		},
+		"error if target container not found": {
```
dannyrandall, line 583: yay! coverage 📈

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
```diff
@@ -30,8 +30,8 @@
   PortMappings:
     - ContainerPort: !Ref ContainerPort
     {{- if .ServiceConnect }} # remove when release
-    {{- if eq .HTTPTargetContainer.Container .WorkloadName}}
-      Name: TargetPort
+    {{- if eq .HTTPTargetContainer.Name .WorkloadName}}
+      Name: target
```
dannyrandall: could we combine these conditions with `and` like below

iamhopaul123: it's fine. we'll get rid of the sc check condition when we release this feature.

internal/pkg/manifest/backend_svc.go
efekarakus: Ah I see what you mean, perhaps you're right and right now it's not worthwhile to introduce more abstraction.
So let's keep it as is 👍 , nit: maybe just rename to `IsServiceConnectEnabled`?

efekarakus (commented):
We talked offline but for visibility: we need to add a feature flag here otherwise it'll start deploying all new services with the updated template

internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/https-path-alias-manifest.yml
```diff
@@ -10,6 +10,9 @@ http:
     - name: "*.foobar.com"
   hosted_zone: mockHostedZone1
 
+network:
+  connect: false
+
```
Lou1415926, line 15: Do we not want to keep some integration test for when service connect is enabled (i.e. it's not opted out)? Or are we looking to add those integration test in the future?

internal/pkg/manifest/workload.go
```diff
@@ -473,6 +475,41 @@ func (c *NetworkConfig) requiredEnvFeatures() []string {
 	return nil
 }
 
+// ServiceConnect represents ECS Service Connect configuration.
+type ServiceConnect struct {
```
Lou1415926: nit: what do you think of  `ServiceConnectArgsOrBool` for the struct name?

internal/pkg/manifest/workload.go
```diff
@@ -47,13 +47,14 @@ var (
...
 	ErrAppRunnerInvalidPlatformWindows = errors.New("Windows is not supported for App Runner services")
 
-	errUnmarshalBuildOpts         = errors.New("unable to unmarshal build field into string or compose-style map")
-	errUnmarshalPlatformOpts      = errors.New("unable to unmarshal platform field into string or compose-style map")
-	errUnmarshalSecurityGroupOpts = errors.New(`unable to unmarshal "security_groups" field into slice of strings or compose-style map`)
-	errUnmarshalPlacementOpts     = errors.New("unable to unmarshal placement field into string or compose-style map")
-	errUnmarshalSubnetsOpts       = errors.New("unable to unmarshal subnets field into string slice or compose-style map")
-	errUnmarshalCountOpts         = errors.New(`unable to unmarshal "count" field to an integer or autoscaling configuration`)
-	errUnmarshalRangeOpts         = errors.New(`unable to unmarshal "range" field`)
+	errUnmarshalBuildOpts          = errors.New("unable to unmarshal build field into string or compose-style map")
```
Lou1415926, line 50: We should make these "unable to unmarshal X into Y or Z" into a customized error struct some time 😂

internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
```diff
@@ -48,11 +48,9 @@
 {{- if $sidecar.Port}}
   PortMappings:
     - ContainerPort: {{$sidecar.Port}}
-    {{- if $.ServiceConnect }} # remove when release
-    {{- if eq $.HTTPTargetContainer.Container $sidecar.Name}}
-    {{- if eq $.HTTPTargetContainer.Port $sidecar.Port}}
-      Name: TargetPort
-    {{- end}}
+    {{- if and $.ServiceConnect $.SCFeatureFlag}} # remove when release
+    {{- if eq $.HTTPTargetContainer.Name $sidecar.Name}}
```
Lou1415926, line 53: Should we add a comment here stating the assumption that currently holds true: "multiple exposed port is not supported yet; therefore, if the target container is the sidecar, then the target port must be the one and only one port that the sidecar exposes."

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
```diff
@@ -29,9 +29,9 @@
 {{- if eq .WorkloadType "Load Balanced Web Service"}}
   PortMappings:
     - ContainerPort: !Ref ContainerPort
-    {{- if .ServiceConnect }} # remove when release
-    {{- if eq .HTTPTargetContainer.Container .WorkloadName}}
-      Name: TargetPort
+    {{- if and .ServiceConnect .SCFeatureFlag}} # remove when release
+    {{- if eq .HTTPTargetContainer.Name .WorkloadName}}
+      Name: target
```
Lou1415926, line 35: Is this field safe to update in the future? For example, is it okay to rename it from `target` to `target1`?

internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/https-path-alias-manifest.yml
iamhopaul123: I think we have all three cases...connect false, connect true and no config.

internal/pkg/manifest/workload.go
iamhopaul123: I thought about it but tbh i'd rather keep as is 🤯

iamhopaul123: Because adding `ArgsOrBool` really adds no value. It might make more sense to differentiate if there's already a `ServiceConnect` struct in this package.

internal/pkg/manifest/workload.go
iamhopaul123: Agreed. I can do a follow up pr to address this.

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
iamhopaul123: yeah i think so. it's only gonna be referenced within the stack. and right now we don't make any assumption based on that elsewhere.

internal/pkg/manifest/workload.go
Lou1415926: I liked that for all the other `XorY` namings because they makes it clear that the struct is meant to have values in one or the other. It's just a nit though no biggeie/

iamhopaul123: I personally don't like this format because I think this has the same problem if we name a list var to be `var subnetList []string`.

iamhopaul123: but eventually across the team we need a naming consensus for this...

efekarakus:
Adding the label for "feat:" PRs until we release a patch release for v1.22.1

dannyrandall (approved):

PR #4047, opened 2022-09-29
fix: target_container incorrectly using main container port/protocol

### LINKED ISSUE #4041: target_container option not working as expected
Website Manifest
```
name: website

http:
  path: '/'
  target_container: proxy

image:
  build: copilot/website/Dockerfile
  port: 80

cpu: 256
memory: 512
platform: linux/x86_64
count: 1

sidecars:
  proxy:
    port: 8443
    image: 123456789012.dkr.ecr.us-east-2.amazonaws.com/application-name/proxy
```

When I use the above manifest to setup an sidecar proxy, the Target Group gets setup with the original container port instead of the target container port.  Here's a look at the section of the CFT that matters:
```
      LoadBalancers:
        - ContainerName: !Ref TargetContainer <--- Here the target container is used
          ContainerPort: !Ref TargetPort <--- Here the target port is used
          TargetGroupArn: !Ref TargetGroup
      ServiceRegistries:
        - RegistryArn: !GetAtt DiscoveryService.Arn
          Port: !Ref ContainerPort <--- The original container port.  Should probably be the target port.
  TargetGroup:
    Metadata:
      'aws:copilot:description': "A target group to connect the load balancer to your service"
    Type: AWS::ElasticLoadBalancingV2::TargetGroup
    Properties:
      HealthCheckPath: / # Default is '/'.
      Port: !Ref ContainerPort <--- This should be the target port
      Protocol: HTTP
```

dannyrandall:
Fixes #4041 

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/backend_svc.go
+++ b/internal/pkg/deploy/cloudformation/stack/backend_svc.go
@@ -241,18 +241,18 @@ func (s *BackendService) Parameters() ([]*cloudformation.Parameter, error) {
 			ParameterKey:   aws.String(WorkloadEnvFileARNParamKey),
 			ParameterValue: aws.String(s.rc.EnvFileARN),
 		},
+		{
+			ParameterKey:   aws.String(WorkloadTargetContainerParamKey),
+			ParameterValue: targetContainer,
+		},
+		{
+			ParameterKey:   aws.String(WorkloadTargetPortParamKey),
+			ParameterValue: targetPort,
+		},
 	}...)
 
 	if !s.manifest.RoutingRule.IsEmpty() {
 		params = append(params, []*cloudformation.Parameter{
-			{
-				ParameterKey:   aws.String(WorkloadTargetContainerParamKey),
-				ParameterValue: targetContainer,
-			},
-			{
-				ParameterKey:   aws.String(WorkloadTargetPortParamKey),
-				ParameterValue: targetPort,
-			},
 			{
 				ParameterKey:   aws.String(WorkloadRulePathParamKey),
 				ParameterValue: s.manifest.RoutingRule.Path,
--- a/internal/pkg/template/templates/workloads/partials/cf/alb.yml
+++ b/internal/pkg/template/templates/workloads/partials/cf/alb.yml
@@ -26,7 +26,7 @@ TargetGroup:
     {{- if .HTTPTargetContainer.IsHTTPS }}
     HealthCheckProtocol: HTTPS
     {{- end}}
-    Port: !Ref ContainerPort
+    Port: !Ref TargetPort
     {{- if .HTTPTargetContainer.IsHTTPS }}
     Protocol: HTTPS
     {{- else }}
--- a/internal/pkg/template/templates/workloads/partials/cf/alb.yml
+++ b/internal/pkg/template/templates/workloads/partials/cf/alb.yml
@@ -23,8 +23,12 @@ TargetGroup:
     {{- if .HTTPHealthCheck.Timeout}}
     HealthCheckTimeoutSeconds: {{.HTTPHealthCheck.Timeout}}
     {{- end}}
-    {{- if .HTTPTargetContainer.IsHTTPS }}
+    {{/* if health check port is configured to 443, or if health check port not specified and target container is https */}}
+    {{- if or (and .HTTPHealthCheck.Port .HTTPHealthCheck.IsHTTPS) (and (not .HTTPHealthCheck.Port) .HTTPTargetContainer.IsHTTPS)}}
     HealthCheckProtocol: HTTPS
+    {{/* for backwards compat, only specify health check protocol as http if target container is https but specified health check port is not */}}
+    {{- else if and .HTTPHealthCheck.Port (not .HTTPHealthCheck.IsHTTPS) .HTTPTargetContainer.IsHTTPS}}
+    HealthCheckProtocol: HTTP
     {{- end}}
     Port: !Ref TargetPort
     {{- if .HTTPTargetContainer.IsHTTPS }}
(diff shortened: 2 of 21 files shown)


iamhopaul123 (approved):

internal/pkg/template/templates/workloads/partials/cf/alb.yml
```diff
@@ -26,7 +26,7 @@ TargetGroup:
     {{- if .HTTPTargetContainer.IsHTTPS }}
     HealthCheckProtocol: HTTPS
     {{- end}}
-    Port: !Ref ContainerPort
+    Port: !Ref TargetPort
```
efekarakus, line 29: Is the health check port already pointing to the target port? 

This `Port` field isn't useful 😂 https://stackoverflow.com/questions/42715647/whats-the-target-group-port-for-when-using-application-load-balancer-ec2-con/42823808#42823808 but it's good to be consistent

efekarakus (commented):
Looks great, just a question

internal/pkg/template/templates/workloads/services/backend/cf.yml
```diff
@@ -102,7 +102,7 @@ Resources:
       {{- end}}
     Properties:
       {{- "\n"}}{{ include "service-base-properties" . | indent 6 }}
-      ServiceRegistries: !If [ExposePort, [{RegistryArn: !GetAtt DiscoveryService.Arn, Port: !Ref ContainerPort}], !Ref "AWS::NoValue"]
+      ServiceRegistries: !If [ExposePort, [{RegistryArn: !GetAtt DiscoveryService.Arn, Port: !Ref TargetPort}], !Ref "AWS::NoValue"]
```
iamhopaul123, line 105: Actually this is not working. `ExposePort` is for main container, and we should check targetPort instead.

internal/pkg/template/templates/workloads/partials/cf/alb.yml
dannyrandall: You're right, it's not useful😂 But! This field does show up in the target group console so this should make it more obvious which port we are using.

dannyrandall: also to answer your first question - health check port defaults to traffic port if not set manually.

internal/pkg/template/templates/workloads/services/backend/cf.yml
dannyrandall: good catch. changed backend service params to always include target port, and changed the `ExposePort` condition to use target port.

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -241,18 +241,18 @@ func (s *BackendService) Parameters() ([]*cloudformation.Parameter, error) {
...
 			ParameterValue: aws.String(s.rc.EnvFileARN),
 		},
+		{
+			ParameterKey:   aws.String(WorkloadTargetContainerParamKey),
+			ParameterValue: targetContainer,
+		},
+		{
+			ParameterKey:   aws.String(WorkloadTargetPortParamKey),
+			ParameterValue: targetPort,
+		},
```
iamhopaul123, line 251: 🤝

internal/pkg/template/workload.go
```diff
@@ -235,6 +235,10 @@ type HTTPHealthCheckOpts struct {
 	DeregistrationDelay *int64
 }
 
+func (h HTTPHealthCheckOpts) IsHTTPS() bool {
```
iamhopaul123, line 240: Comment pls

internal/pkg/template/templates/workloads/partials/cf/alb.yml
```diff
@@ -23,10 +23,14 @@ TargetGroup:
     {{- if .HTTPHealthCheck.Timeout}}
     HealthCheckTimeoutSeconds: {{.HTTPHealthCheck.Timeout}}
     {{- end}}
-    {{- if .HTTPTargetContainer.IsHTTPS }}
+    {{/* if health check port is configured to 443, or if health check port not specified and target container is https */}}
+    {{- if or (and .HTTPHealthCheck.Port .HTTPHealthCheck.IsHTTPS) (and (not .HTTPHealthCheck.Port) .HTTPTargetContainer.IsHTTPS)}}
```
iamhopaul123: Would it be better (more readable) if we move this logic to the function itself 🤔 Like can we combine

```
	HTTPTargetContainer     HTTPTargetContainer
	HTTPHealthCheck         HTTPHealthCheckOpts
```

to 

```
type HTTPTarget struct {
    HTTPTargetContainer
    HTTPHealthCheckOpts
}
```

efekarakus: 👍 or we can add a template function too, something like:
```yaml
HealthCheckProtocol: {{healthCheckProtocol .HTTPHealthCheck .HTTPTargetContainer}}
```
That returns either "HTTP" or "HTTPS"

internal/pkg/template/workload.go
```diff
@@ -593,6 +599,24 @@ type WorkloadOpts struct {
...
+// HealthCheckProtocol returns the protocol for the Load Balancer health check,
+// or an empty string if it shouldn't be configured, defaulting to the
+// target protocol. (which is what happens, even if it isn't documented as such :))
+// https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-targetgroup.html#cfn-elasticloadbalancingv2-targetgroup-healthcheckprotocol
+func (w WorkloadOpts) HealthCheckProtocol() string {
+	switch {
+	case w.HTTPHealthCheck.Port == "443":
+		return "HTTPS"
+	case w.HTTPTargetContainer.IsHTTPS() && w.HTTPHealthCheck.Port == "":
+		return "HTTPS"
```
efekarakus, line 611: do we need this case then, or can it be ""? since the TG protocol will be set to HTTPS

efekarakus (approved):

internal/pkg/template/workload.go
dannyrandall: i think we should explicitly configure it since the behavior isn't documented correctly

PR #4048, opened 2022-09-29
test: add integ test for permissions boundary

huanjani:
This integ test ensures that as we add CFN templates and/or IAM roles to CFN templates, we also conditionally attach a permission boundary if present in the app.

### DIFF
--- a/internal/pkg/template/permissions_boundary_integration_test.go
+++ b/internal/pkg/template/permissions_boundary_integration_test.go
@@ -9,28 +9,23 @@ import (
 	"bytes"
 	"fmt"
 	"github.com/stretchr/testify/require"
+	"io/fs"
 	"os"
-	"os/exec"
-	"strings"
+	"path/filepath"
 	"testing"
 )
 
 func TestPermissions_Boundary(t *testing.T) {
 	t.Run("every CloudFormation template must contain conditional permissions boundary field for all IAM roles", func(t *testing.T) {
-		cmd := exec.Command("find", "templates", "-name", "*yml")
-		output, err := cmd.Output()
-		require.NoError(t, err, "should return output of 'find' command")
-		
-		files := strings.Fields(string(output))
-		for _, file := range files {
-			contents, err := os.ReadFile(file)
-			require.NoError(t, err, "should read file")
+		err := filepath.WalkDir("templates", func(path string, di fs.DirEntry, err error) error {
+			contents, _ := os.ReadFile(path)
 			IAMRoles := bytes.Count(contents, []byte("AWS::IAM::Role"))
 			permissionsBoundaryFields := bytes.Count(contents, []byte("PermissionsBoundary:"))
-			
-			msg := fmt.Sprintf("number of IAM roles (%d) does not equal number of permissions boundary fields (%d) in file '%s'", IAMRoles, permissionsBoundaryFields, file)
-			require.True(t, IAMRoles == permissionsBoundaryFields, msg)
 
-		}
+			msg := fmt.Sprintf("number of IAM roles (%d) does not equal number of permissions boundary fields (%d) in file '%s'", IAMRoles, permissionsBoundaryFields, path)
+			require.True(t, IAMRoles == permissionsBoundaryFields, msg)
+			return nil
+		})
+		require.NoError(t, err, "should walk templates dir for template files")
 	})
 }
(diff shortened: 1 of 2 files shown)


internal/pkg/template/permissions_boundary_integration_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+func TestPermissions_Boundary(t *testing.T) {
+	t.Run("every CloudFormation template must contain conditional permissions boundary field for all IAM roles", func(t *testing.T) {
+		cmd := exec.Command("find", "templates", "-name", "*yml")
+		output, err := cmd.Output()
```
efekarakus: Is there a reason why we preffered using `find`? would this work on different platforms like Windows 🤔 
Can we leverage `os.ReadDir` and walk through the files that way?

internal/pkg/template/permissions_boundary_integration_test.go
```diff
@@ -0,0 +1,37 @@
...
+		for _, file := range files {
+			contents, err := os.ReadFile(file)
+			require.NoError(t, err, "should read file")
+			IAMRoles := bytes.Count(contents, []byte("AWS::IAM::Role"))
+			permissionsBoundaryFields := bytes.Count(contents, []byte("PermissionsBoundary:"))
+				
+			totalIAMRoles += IAMRoles
+			totalPermissionsBoundaryFields += permissionsBoundaryFields
+		}
+		require.True(t, totalIAMRoles == totalPermissionsBoundaryFields, "number of IAM roles does not equal number of permissions boundary fields")
```
efekarakus: I think with the current implementation we won't be able to tell which file doesn't have the `PermissionsBoundary:` set.

Can we fail early instead in the for-loop writing exactly which file (the filepath) has the permissions boundary missing? 
For example, try removing one in the templates and see the error message you get from the test to see if that would help the next engineer 😂 

We could even potentially figure out which IAM role doesn't have PermissionsBoundary set by analyzing the line numbers but it's prob over engineering

internal/pkg/template/permissions_boundary_integration_test.go
huanjani: I switched it to `filepath.WalkDir`...what do you think?

internal/pkg/template/permissions_boundary_integration_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"io/fs"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestPermissions_Boundary(t *testing.T) {
+	t.Run("every CloudFormation template must contain conditional permissions boundary field for all IAM roles", func(t *testing.T) {
+		err := filepath.WalkDir("templates", func(path string, di fs.DirEntry, err error) error {
+			contents, _ := os.ReadFile(path)
```
efekarakus: nit:
```suggestion
			contents, err := os.ReadFile(path)
			require.NoError(err, "read file at %s", path)
```

internal/pkg/template/permissions_boundary_integration_test.go
```diff
@@ -0,0 +1,31 @@
...
+			IAMRoles := bytes.Count(contents, []byte("AWS::IAM::Role"))
+			permissionsBoundaryFields := bytes.Count(contents, []byte("PermissionsBoundary:"))
+
+			msg := fmt.Sprintf("number of IAM roles (%d) does not equal number of permissions boundary fields (%d) in file '%s'", IAMRoles, permissionsBoundaryFields, path)
+			require.True(t, IAMRoles == permissionsBoundaryFields, msg)
```
efekarakus: nit:
```suggestion
			require.Equal(t, roleCount, fieldCount, "number of IAM roles does not equal number of permissions boundary fields in file %s", path)
```

efekarakus (approved):
Looks great! just some minor suggestions to leverage the `testify` lib feel free to remove the label after

internal/pkg/template/permissions_boundary_integration_test.go
huanjani: Ah, I had taken that `err` out because `ReadFile` was returning errors when trying to read directories, like `read templates: is a directory`

huanjani: added `IsDir()`

bvtujo (approved):
`WalkDir` is a nifty function!

PR #4056, opened 2022-10-03
fix(FIFO): worker service manifest with fifo subscriptions

paragbhingre:
This PR fix the following scenario for FIFO SNS and SQS.
Initiate a new worker service and select FIFO topic for the subscriptions from the listed items and manifest gets populated with for example - 
```
subscribe:
  topics:
    - name: orders.fifo
      service: api
```
Whereas in this scenario manifest should be - 
```
subscribe:
  topics:
    - name: orders
      service: api
  queue:
    fifo: true
```


internal/pkg/manifest/worker_svc.go
```diff
@@ -206,6 +208,14 @@ func NewWorkerService(props WorkerServiceProps) *WorkerService {
 		svc.WorkerServiceConfig.TaskConfig.CPU = aws.Int(MinWindowsTaskCPU)
 		svc.WorkerServiceConfig.TaskConfig.Memory = aws.Int(MinWindowsTaskMemory)
 	}
+	if len(props.Topics) > 0 {
+		for idx, topic := range props.Topics {
+			if strings.Contains(aws.StringValue(topic.Name), ".fifo") {
```
iamhopaul123: Maybe use https://pkg.go.dev/strings#HasSuffix

internal/pkg/manifest/worker_svc.go
```diff
@@ -206,6 +208,14 @@ func NewWorkerService(props WorkerServiceProps) *WorkerService {
...
+				props.Topics[idx].Name = aws.String(strings.TrimSuffix(aws.StringValue(topic.Name), ".fifo"))
+				svc.WorkerServiceConfig.Subscribe.Queue.FIFO.Enable = aws.Bool(true)
+			}
+		}
+	}
```
iamhopaul123: Can we move this chunk into a separate private function? So that i'll be clearer that we are checking the FIFO.

```go
svc.WorkerServiceConfig.Subscribe.Queue.FIFO.Enable = aws.Bool(fifoEnabled(props.Topics))
```

something like that.

Lou1415926: It seems like by this for-loop we are saying as long as there is at least one fifo topic that the service subscribe to - even if the service could subscribe also to some standard topics - we create the default queue as fifo queue. Is this what we want?

internal/pkg/template/templates/workloads/services/worker/manifest.yml
```diff
@@ -42,6 +42,10 @@ subscribe:
     - name: {{$topic.Name}}
       service: {{$topic.Service}}
 {{- end}}
+  {{- if and .Subscribe.Queue .Subscribe.Queue.FIFO .Subscribe.Queue.FIFO.Enable }}
```
iamhopaul123: We might just need
```suggestion
  {{- if .Subscribe.Queue.FIFO.Enable }}
```

iamhopaul123 (commented):
Can we add any test cases making sure we write correct manifest with FIFO?

internal/pkg/manifest/worker_svc.go
paragbhingre: Yes, so when customer selects subscriptions from the wizard while initiating the worker service - 
1. There is no option for them to specify topic level sqs queue.
2. Only default sqs queue used to be created after selecting subscriptions.
3. Now that FIFO has been introduced, if customer selects FIFO topics then we will create default FIFO sqs queue as we can not create both default Standard and FIFO queues at the same time. 

So customers has to select either Standard or FIFO topic subscriptions. And we can convey the same through some warning may be.

internal/pkg/template/templates/workloads/services/worker/manifest.yml
paragbhingre: I think we need this because there is no check before this point for the `Queue` and `FIFO` configs. If they turned out to be nil then we will face nil pointer exception.

iamhopaul123: Can they turn `nil` 🤔  https://github.com/aws/copilot-cli/blob/9a6d5d9101c8e947ceada726a4bec72e1d72422b/internal/pkg/manifest/worker_svc.go#L62

internal/pkg/manifest/worker_svc.go
Lou1415926: Would it be a bad idea if we populate the manifest such that:
1. If all subscribed topics are fifo, then we write `queue.fifo: true`
2. If all subscribed topics are standard, then we don't write `queue.fifo: true` (create default standard queue)
3. If some are fifo, then we keep the default queue as "standard, but populate the manifest so that each FIFO topics has their topic-sepecifc queue that is fifo.

efekarakus: If the user selects FIFO and standard topics, why can't we generate a manifest that has standard queues for the standard topics?


Something like this in pseudo-code:
```
standardTopics <- filter(props.Topics, isStandard)
fifoTopics <- filter(props.Topics, isFifo)

if len(fifoTopics) == 0:
   return # nothing to do

enableEventsQueueWithFifo()
for standardTopic in standardTopics:
   enableStandardQueue(standardTopic)
```

paragbhingre: I think option 1 and 2 are valid. But 3rd doesn't seem consistent because we do not create topic-specific queues for standard ever reason being we create default standard queue. Moreover customer is now aquatinted with the behavior that when you subscribe topics through wizard are having only one default queue and not topic-specific ones.

paragbhingre: > If the user selects FIFO and standard topics, why can't we generate a manifest that has standard queues for the standard topics?
> 
> Something like this in pseudo-code:
> 
> ```
> standardTopics <- filter(props.Topics, isStandard)
> fifoTopics <- filter(props.Topics, isFifo)
> 
> if len(fifoTopics) == 0:
>    return # nothing to do
> 
> enableEventsQueueWithFifo()
> for standardTopic in standardTopics:
>    enableStandardQueue(standardTopic)
> ```

so you mean `enableEventsQueueWithFifo()` will create fifo queue to which all the fifo topics will be subscribed to?

Lou1415926: To clarify, I meant:
1. The manifest that we generate would be:
```yaml
subscribe:
  topics:
     - name: myFIFOTopicA
     - name: myFIFOTopicB
   queue:
      fifo: true
```
2. The manifest generated would be
```yaml
subscribe:
  topics:
     - name: myStandardTopicA
     - name: myStandardTopicB
   queue:
```
3. The manifest would be
```yaml
subscribe:
  topics:
     - name: myFIFOTopicA
        queue:
            fifo: true
     - name: myStandardTopicA
   queue:
```

internal/pkg/template/templates/workloads/services/worker/manifest.yml
paragbhingre: But in CFN we use struct from `template/workflow` like [this](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/workload.go#L435). So it can be nil.

internal/pkg/manifest/worker_svc.go
paragbhingre: I understood your point, but I am just hesitant because of inconsistency between standard and FIFO implementation. 
What if customers want FIFO to be the default queue and standard being the topic-specific queue. That will still be impossible through the CLI.

Lou1415926: People can still modify the manifest after `svc init`. 

The purpose of generating a default manifest by `svc init` is to give users a good start, not to produce a manifest that they are stuck with forever - so I think it's okay for us to make a decision between: A. write in the manifest that the default queue is standard + topic-specific fifo queue; B. default fifo + topic-specific standard. 

If we generate a manifest that goes with A, and the user wants b, they can modify the manfiest; and vice versa.

internal/pkg/manifest/worker_svc.go
```diff
@@ -206,12 +208,39 @@ func NewWorkerService(props WorkerServiceProps) *WorkerService {
...
+			svc.WorkerServiceConfig.Subscribe.Queue.FIFO.Enable = aws.Bool(true)
+		}
+	}
 	svc.WorkerServiceConfig.Subscribe.Topics = props.Topics
 	svc.WorkerServiceConfig.Platform = props.Platform
 	svc.parser = template.New()
 	return svc
 }
 
+func FIFOEnabled(topics []TopicSubscription) *bool {
```
efekarakus: Can this be private? 🥺

efekarakus: Right now this function sounds like it doesn't have any side-effects (like changing the underlying `TopicSubscriptions`), but it indeed does mutate the data.

What do you think of a structure like this:
```go
func setSubscriptionQueueDefaults(topics []TopicSubscription, eventsQueue *SQSQueue) {
  //...
  if !isFIFOEnabled {
     return
  }
  eventsQueue.FIFO.Enable = aws.Bool(true)
  for idx, topic := range topics {
	if strings.HasSuffix(aws.StringValue(topic.Name), ".fifo") {
		topics[idx].Name = aws.String(strings.TrimSuffix(aws.StringValue(topic.Name), ".fifo"))
	} else {
		topics[idx].Queue.Enabled = aws.Bool(true)
	}
  }
}
```
and then in the constructor we can do:
```go
if len(props.Topic) > 0 {
   setSubscriptionQueueDefaults(props.Topics, &svc.WorkerServiceConfig.Subscribe.Queue)
}
```

efekarakus (commented):
Looks great! just one tiny refactor request

efekarakus (approved):

internal/pkg/template/templates/workloads/services/worker/manifest.yml
iamhopaul123: im not convinced. Your url is for cfn template generation and here we are talking about manifest, which is marshaling the manifest struct.

dannyrandall:
just a clarifying question around `subscribe` - what happens if users subscribe to some fifo topics and some not-fifo topics?

internal/pkg/template/templates/workloads/services/worker/manifest.yml
paragbhingre: Ohh you are right, I got confused with the templates. Thanks for catching this.

internal/pkg/manifest/worker_svc.go
```diff
@@ -206,12 +208,36 @@ func NewWorkerService(props WorkerServiceProps) *WorkerService {
...
 	svc.WorkerServiceConfig.Subscribe.Topics = props.Topics
 	svc.WorkerServiceConfig.Platform = props.Platform
 	svc.parser = template.New()
 	return svc
 }
 
+func setSubscriptionQueueDefaults(topics []TopicSubscription, eventsQueue *SQSQueue) {
+	var isFIFOEnabled bool
+	for _, topic := range topics {
+		if strings.HasSuffix(aws.StringValue(topic.Name), ".fifo") {
```
dannyrandall: might be nice to add a function to `TopicSubscription` like `IsFIFO() bool` and use that instead? or maybe we add a bool to that struct instead of relying on the Name so we don't have to rely on the topic "name" that we have to then trim off later? (L234)

paragbhingre: Good question Danny, So in case of some fifo topics and some not-fifo topics we have an opinion that we will create FIFO default events queue to which all the FIFO topics will be subscribing to. And all the other non-fifo topics will be having their own topic specific standard queue. 

Basically `isFIFOEnabled` variable tells us if we need to have a default FIFO Events queue or not. So in that we will still need to remove `.fifo` from the name in order for our manifest not to have `.fifo` named topics. 
does that make sense?

iamhopaul123 (approved):

internal/pkg/manifest/worker_svc.go
dannyrandall: > we have an opinion that we will create FIFO default events queue to which all the FIFO topics will be subscribing to

gotcha

> So in that we will still need to remove .fifo from the name in order for our manifest not to have .fifo named topics

i guess it just feels ...off that the name suffix and the type of topic are so mixed, but i get that's kinda just how it is and looks like already is assumed in other functions, so i think it's ok in these scenario as well

dannyrandall (approved):

PR #4061, opened 2022-10-05
fix: allow cdn certificate import without http cert import

dannyrandall:
Also fixes a bug where CDN fails to create if there are any internal ALB services deployed.
Also fixes a bug where on `svc/env package` we don't escape param values

### DIFF
--- a/cf-custom-resources/test/unique-json-values-test.js
+++ b/cf-custom-resources/test/unique-json-values-test.js
@@ -75,8 +75,8 @@ describe("Unique Aliases", () => {
       });
   });
 
-  const aliasTest = (name, input, expectedOutput) => {
-    const tt = (name, reqType, input, expectedOutput) => {
+  const aliasTest = (name, props, expectedOutput) => {
+    const tt = (name, reqType, props, expectedOutput) => {
       test(name, () => {
         const request = nock(responseURL)
           .put("/", (body) => {
@@ -95,9 +95,7 @@ describe("Unique Aliases", () => {
             ResponseURL: responseURL,
             RequestType: reqType,
             RequestId: testRequestId,
-            ResourceProperties: {
-              Aliases: JSON.stringify(input), // aliases get passed as a string
-            },
+            ResourceProperties: props,
             LogicalResourceId: "mockID",
           })
           .expectResolve(() => {
@@ -106,28 +104,80 @@ describe("Unique Aliases", () => {
       });
     };
 
-    tt(`Create/${name}`, "Create", input, expectedOutput);
-    tt(`Update/${name}`, "Update", input, expectedOutput);
+    tt(`Create/${name}`, "Create", props, expectedOutput);
+    tt(`Update/${name}`, "Update", props, expectedOutput);
   };
 
-  aliasTest("no aliases", {}, []);
+  aliasTest("no aliases", {
+    Aliases: "",
+    FilterFor: ""
+  }, []);
 
   aliasTest("one service", {
-    "svc1": ["svc1.com", "example.com"],
+    Aliases: JSON.stringify({
+      "svc1": ["svc1.com", "example.com"]
+    }),
+    FilterFor: "svc1",
   }, ["example.com", "svc1.com"]);
 
+  aliasTest("one service excluded", {
+    Aliases: JSON.stringify({
+      "svc1": ["svc1.com", "example.com"]
+    }),
+    FilterFor: "svc2",
+  }, []);
+
+  aliasTest("one service empty filter for", {
+    Aliases: JSON.stringify({
+      "svc1": ["svc1.com", "example.com"]
+    }),
+    FilterFor: "",
+  }, []);
+
   aliasTest("two services no common aliases", {
-    "svc1": ["svc1.com"],
-    "svc2": ["svc2.com"]
+    Aliases: JSON.stringify({
+      "svc1": ["svc1.com"],
+      "svc2": ["svc2.com"]
+    }),
+    FilterFor: "svc1,svc2",
   }, ["svc1.com", "svc2.com"]);
 
   aliasTest("two services, one with multiple common aliases", {
-    "svc1": ["svc1.com"],
-    "svc2": ["svc2.com", "example.com"]
+    Aliases: JSON.stringify({
+      "svc1": ["svc1.com"],
+      "svc2": ["svc2.com", "example.com"]
+    }),
+    FilterFor: "svc1,svc2",
   }, ["example.com", "svc1.com", "svc2.com"]);
 
   aliasTest("two services with a common alias", {
-    "svc1": ["svc1.com", "example.com"],
-    "svc2": ["svc2.com", "example.com"]
+    Aliases: JSON.stringify({
+      "svc1": ["svc1.com", "example.com"],
+      "svc2": ["svc2.com", "example.com"]
+    }),
+    FilterFor: "svc1,svc2",
   }, ["example.com", "svc1.com", "svc2.com"]);
+
+  aliasTest("three services with a common alias one service filtered out", {
+    Aliases: JSON.stringify({
+      "svc1": ["svc1.com", "example.com"],
+      "svc2": ["svc2.com", "example.com", "example2.com"],
+      "svc3": ["svc3.com", "example.com", "example2.com"]
+    }),
+    FilterFor: "svc2,svc1",
+  }, ["example.com", "example2.com", "svc1.com", "svc2.com"]);
+
+  aliasTest("bunch of services with single alias, some filtered out, out of order", {
+    Aliases: JSON.stringify({
+      "lbws3": ["three.lbws.com"],
+      "backend1": ["one.backend.internal"],
+      "lbws1": ["one.lbws.com"],
+      "lbws4": ["four.lbws.com"],
+      "backend2": ["two.backend.internal"],
+      "lbws2": ["two.lbws.com"],
+      "lbws6": ["lbws.com"],
+      "lbws5": ["lbws.com"],
+    }),
+    FilterFor: "lbws2,lbws3,lbws4,lbws1,lbws5,lbws6",
+  }, ["four.lbws.com", "lbws.com", "one.lbws.com", "three.lbws.com", "two.lbws.com"]);
 });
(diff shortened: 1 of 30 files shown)


cf-custom-resources/test/unique-json-values-test.js
```diff
@@ -106,28 +104,80 @@ describe("Unique Aliases", () => {
...
+      "lbws3": ["three.lbws.com"],
+      "backend1": ["one.backend.internal"],
+      "lbws1": ["one.lbws.com"],
+      "lbws4": ["four.lbws.com"],
+      "backend2": ["two.backend.internal"],
+      "lbws2": ["two.lbws.com"],
+      "lbws6": ["lbws.com"],
+      "lbws5": ["lbws.com"],
+    }),
+    FilterFor: "lbws2,lbws3,lbws4,lbws1,lbws5,lbws6",
```
efekarakus, line 181: nit: can u add one more where `FilterFor` has a key like "unknown" but it's not in `Aliases`

efekarakus: ps: really nice tests 🥹

internal/pkg/cli/deploy/svc.go
```diff
@@ -1360,12 +1360,13 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
 }
 
 func (d *lbWebSvcDeployer) validateALBRuntime() error {
-	hasImportedCerts := len(d.envConfig.HTTPConfig.Public.Certificates) != 0
-	if d.lbMft.RoutingRule.RedirectToHTTPS != nil && d.app.Domain == "" && !hasImportedCerts {
+	hasALBCerts := len(d.envConfig.HTTPConfig.Public.Certificates) != 0
+	hasCDNCerts := d.envConfig.CDNConfig.Config.Certificate != nil
```
efekarakus, line 1364: I kinda liked the `hasImportedCerts` field, what do you think of introducing it back with:
```suggestion
	hasCDNCerts := d.envConfig.CDNConfig.Config.Certificate != nil
	hasImportedCerts := hasALBCerts || hasCDNCerts
```
This way we can retain the conditionals like before for the if-statements

efekarakus (commented):
Nice! the code lgtm, just some super tiny feedback.

I wonder if we can also fix the `env package --output-dir` bug where the aliases json string isn't escaped

internal/pkg/deploy/cloudformation/stack/workload.go
```diff
@@ -186,19 +182,30 @@ func serializeTemplateConfig(parser template.Parser, stack templateConfigurer) (
...
+		config.Tags[aws.StringValue(tag.Key)] = tag.Value
+	}
+
+	str, err := json.MarshalIndent(config, "", "  ")
 	if err != nil {
 		return "", err
 	}
-	return doc.String(), nil
+
+	return string(str), nil
```
efekarakus: nit: should we wrap the err since it's an external call?
```go
out, err := json.MarshalIndent()
if err != nil {
   return "", fmt.Errorf("marshal stack parameters to JSON: %v", err)
}
return string(out), nil
```

efekarakus (approved):
sweet! Thank you Danny!

internal/pkg/deploy/cloudformation/stack/workload.go
dannyrandall: oh yeah, good idea

internal/pkg/manifest/validate_env.go
```diff
@@ -49,11 +49,8 @@ func (e EnvironmentConfig) validate() error {
 	if e.CDNEnabled() {
 		cdnCert := e.CDNConfig.Config.Certificate
 		if e.HTTPConfig.Public.Certificates == nil {
-			if cdnCert != nil {
-				return &errFieldMustBeSpecified{
-					missingField:      "http.public.certificates",
-					conditionalFields: []string{"cdn.certificate"},
-				}
+			if cdnCert != nil && !aws.BoolValue(e.CDNConfig.Config.TerminateTLS) {
+				return errors.New("cdn.terminate_tls must be true if cdn.certificate is set without http.public.certificates")
```
Lou1415926: nit: perhaps some `%q` to add the quotation marks?

internal/pkg/cli/deploy/env.go
```diff
@@ -160,7 +160,7 @@ func (d *envDeployer) validateCDN(mft *manifest.Environment) error {
 		}
 	}
 
-	if mft.CDNEnabled() && mft.CDNDoesTLSTermination() && (mft.HasImportedPublicALBCerts() || d.app.Domain != "") {
+	if mft.CDNEnabled() && mft.CDNDoesTLSTermination() {
```
Lou1415926, line 163: qq: Why do we no longer need the second part?

dannyrandall: good question! we want to allow people to use CDN TLS termination without importing public alb certs or having a copilot managed domain. So they could have just the CDN cert and use aliases from that.

As part of this change, we also had to remove the error [here](https://github.com/aws/copilot-cli/pull/4061/files#diff-924fba6d7ac2c8e2b099ec2776e78f958c7b225d5484001289b0e62283df3db1R241-R242)

Lou1415926: I think - iirc -  the redirection rule is created only when HTTPS is enabled => only when public ALB cert is created / app.domain != nil. Wasn't that why we needed that second part back then (and maybe now as well)?

iamhopaul123: So basically it is possible for service to not enable HTTPS at all (native domain or import cert) and in that scenario the function that check if HTTP redirect is enabled or not will not throw any error for HTTP service.

internal/pkg/cli/deploy/errors.go
```diff
@@ -46,7 +46,10 @@ type errEnvHasPublicServicesWithRedirect struct {
 }
 
 func (e *errEnvHasPublicServicesWithRedirect) Error() string {
-	return fmt.Sprintf("%v services redirect HTTP to HTTPS", len(e.services))
+	return fmt.Sprintf("%v %s HTTP to HTTPS",
```
iamhopaul123, line 49: Do we need to include the service names  in the error msg? Also btw is this included in any unit test?

internal/pkg/manifest/validate_env.go
```diff
@@ -49,11 +49,8 @@ func (e EnvironmentConfig) validate() error {
 	if e.CDNEnabled() {
 		cdnCert := e.CDNConfig.Config.Certificate
 		if e.HTTPConfig.Public.Certificates == nil {
-			if cdnCert != nil {
-				return &errFieldMustBeSpecified{
-					missingField:      "http.public.certificates",
-					conditionalFields: []string{"cdn.certificate"},
-				}
+			if cdnCert != nil && !aws.BoolValue(e.CDNConfig.Config.TerminateTLS) {
+				return errors.New(`"cdn.terminate_tls" must be true if "cdn.certificate" is set without "http.public.certificates"`)
```
iamhopaul123, line 53: What if they use default Copilot alias and specify `cdn.certificate` for that? But I guess in that scenario they can just remove `cdn.certificate` and it should still work.

internal/pkg/deploy/cloudformation/stack/testdata/workloads/windows-svc-test.params.json
```diff
@@ -1,26 +1,26 @@
...
     "ContainerPort": "80",
     "DNSDelegated": "false",
-    "TargetContainer": "windows-fe",
-    "TargetPort": "80",
     "EnvFileARN": "",
-    "RulePath": "/",
+    "EnvName": "test",
     "HTTPSEnabled": "false",
-    "Stickiness": "false"
+    "LogRetention": "30",
```
iamhopaul123, line 11: Are we just moving things around for those test files?

internal/pkg/cli/deploy/env.go
Lou1415926: I guess my point is just that if https is not enabled at all, there is no need to validate if any redirect is there. But yeah it doesn't hurt to check.

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -44,6 +44,7 @@ UniqueAliasesAction:
   Properties:
     ServiceToken: !GetAtt UniqueJSONValuesFunction.Arn
     Aliases: !Ref Aliases
+    FilterFor: !Ref ALBWorkloads
```
iamhopaul123, line 47: Why do we need this?

internal/pkg/cli/deploy/env.go
iamhopaul123: That is true. I think we should return early and don't call `validateALBWorkloadsDontRedirect` at all.

dannyrandall: so the problem with returning early is just because someone has updated their environment manifest to not have public alb certs, we can't assume that they haven't deployed https services and are now _moving_ the cert to the cdn. in that case we still want to tell them to disable the redirect on those services.

additionally, as part of this PR, customers can deploy "HTTPS" services (meaning we create two listener rules) without importing any http public certs and only relying on the cdn cert. then as people make other manifest changes and deploy, it's still relevant to check and warn them if some services are redirecting.

(happy to explain more, typed this on my phone so sorry😅)

internal/pkg/template/templates/environment/partials/cdn-resources.yml
dannyrandall: discussed offline - needed to remove `InternalALBWorkloads` from `Aliases`

internal/pkg/deploy/cloudformation/stack/testdata/workloads/windows-svc-test.params.json
dannyrandall: yeah, since we removed the parameters template and now use `json.Marshal`, the parameters are alphabetized - from [json.Marshal](https://pkg.go.dev/encoding/json#Marshal):
> The map keys are sorted and used as JSON object keys

internal/pkg/cli/deploy/errors.go
dannyrandall: I don't think so - service name(s) are printed in `RecommendActions()`. there is a test [here](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/cli/deploy/env_test.go#L561-L617).

internal/pkg/manifest/validate_env.go
dannyrandall: i think in that case yeah, they should just remove `cdn.certificate` since we manage a cert for them.

iamhopaul123 (approved):

PR #4062, opened 2022-10-05
feat: Set ReadonlyRootFilesystem to true by default in manifest

### LINKED ISSUE #4010: Set `ReadonlyRootFilesystem` to `true` by default in manifest files
## Ask

To comply with Security Hub's recommendation of [[ECS.5] ECS containers should be limited to read-only access to root filesystems](https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-standards-fsbp-controls.html#fsbp-ecs-5). Copilot should consider generating default manifest files with the following field already populated:
```yaml
name: my-service

storage:
  readonly_root_fs: true
```
This way existing deployed service won't break, but newer services will comply with ECS.5 by enabling [`ReadonlyRootFilesystem: true`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdefinitions.html#cfn-ecs-taskdefinition-containerdefinition-readonlyrootfilesystem) in the CloudFormation template.

KollaAdithya:
This PR resolves #4010


internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
```diff
@@ -48,6 +48,11 @@
     StartPeriod: {{.HealthCheck.StartPeriod}}
     Timeout: {{.HealthCheck.Timeout}}
 {{- end}}
+{{- if .Storage}}
+{{- if .Storage.ReadonlyRootFS}}
```
paragbhingre: nit: you can just club 2 conditions together.
```suggestion
{{- if and .Storage .Storage.ReadonlyRootFS}}
```

internal/pkg/manifest/validate_test.go
```diff
@@ -3242,6 +3246,18 @@ func TestValidateWindows(t *testing.T) {
 			in:          validateWindowsOpts{},
 			wantedError: nil,
 		},
+		"error if readonlyfs is true": {
+			in: validateWindowsOpts{
+				readOnlyFS: aws.Bool(true),
+			},
+			wantedError: errors.New(`'ReadOnlyFS' can not be set to true when deploying a Windows container`),
+		},
+		"should return nil readonlyfs is false or nil": {
```
paragbhingre:
```suggestion
		"should return nil if readonlyfs is false or nil": {
```

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
```diff
@@ -48,6 +48,11 @@
...
+  ReadonlyRootFilesystem: {{.Storage.ReadonlyRootFS}}
```
paragbhingre, line 64: Is `ReadonlyRootFS` has default value set to true? Because the [issue](https://github.com/aws/copilot-cli/issues/4010) says we should by default have following config in the manifest. 
```
storage:
  readonly_root_fs: true
```

paragbhingre (commented):
Overall looks good, just a couple nits and a question.

internal/pkg/manifest/validate.go
```diff
@@ -1740,6 +1745,9 @@ func isValidSubSvcName(name string) bool {
 }
 
 func validateWindows(opts validateWindowsOpts) error {
+	if aws.BoolValue(opts.readOnlyFS) {
+		return errors.New(`'ReadOnlyFS' can not be set to true when deploying a Windows container`)
```
efekarakus: nit: technically, our field name in the manifest is `readonly_fs` :) The [%q fmt verb](https://pkg.go.dev/fmt#hdr-Printing) adds double quotes automatically for the string.
```suggestion
		return fmt.Errorf(`%q can not be set to 'true' when deploying a Windows container`, "readonly_fs")
```

internal/pkg/manifest/validate_test.go
```diff
@@ -211,17 +211,18 @@ func TestLoadBalancedWebService_validate(t *testing.T) {
...
 					TaskConfig: TaskConfig{
 						Platform: PlatformArgsOrString{PlatformString: (*PlatformString)(aws.String("windows/amd64"))},
-						Storage: Storage{Volumes: map[string]*Volume{
-							"foo": {
-								EFS: EFSConfigOrBool{
-									Enabled: aws.Bool(true),
-								},
-								MountPointOpts: MountPointOpts{
-									ContainerPath: aws.String("mockPath"),
+						Storage: Storage{ReadonlyRootFS: aws.Bool(false),
```
efekarakus: nit: can we move the value to the next line 🥺 it took me a bit to figure out what changed lol
```suggestion
						Storage: Storage{
							ReadonlyRootFS: aws.Bool(false),
```

internal/pkg/manifest/validate_test.go
```diff
@@ -869,17 +871,18 @@ func TestWorkerService_validate(t *testing.T) {
...
```
efekarakus: We don't need to set the `ReadonlyRootFS` field in these test cases (removing the field, gives it a default value of `nil` which should be equivalent to `aws.Bool(false)`). So I think it's safe to remove them from here.

internal/pkg/manifest/validate_test.go
efekarakus: The test checks for `aws.Bool(false)` but not when the value is `nil`. Maybe we can rename it `should return nil if readonly_fs is false`. 

We can rename the above test case from `should return nil efs not specified` to `should return nil when no fields are specified`

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
efekarakus: This cloudformation template looks good to me! (only enable the feature if `.Storage.ReadonlyRootFS` is true).

But 👍 to Parag's feedback. We would need to add the field in the manifest in these files:
https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/workloads/services/lb-web/manifest.yml

So that the default generated manifest has the field set to true

efekarakus:
Looks great ❤️, thanks Adi!!!

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
KollaAdithya: working on  it

Lou1415926 (approved):
Thank you Adi! This looks great - I am stealing one approval because it's so good ❤️

internal/pkg/manifest/testdata/backend-svc-customhealthcheck.yml
```diff
@@ -39,4 +39,6 @@ exec: true     # Enable running commands in your container.
 #  test:
 #    count: 2               # Number of tasks to run for the "test" environment.
 #    deployment:            # The deployment strategy for the "test" environment.
-#       rolling: 'recreate' # Stops existing tasks before new ones are started for faster deployments.
\ No newline at end of file
+#       rolling: 'recreate' # Stops existing tasks before new ones are started for faster deployments.
+storage:
+  readonly_fs: true
```
efekarakus: Can we move the `storage` setting under  `exec: true`? Ideally the `#environments` comment is the last thing in their manifest file 🥺

internal/pkg/template/templates/workloads/services/backend/manifest.yml
```diff
@@ -57,4 +57,6 @@ exec: true     # Enable running commands in your container.
 #  test:
 #    count: 2               # Number of tasks to run for the "test" environment.
 #    deployment:            # The deployment strategy for the "test" environment.
-#       rolling: 'recreate' # Stops existing tasks before new ones are started for faster deployments.
\ No newline at end of file
+#       rolling: 'recreate' # Stops existing tasks before new ones are started for faster deployments.
+storage:
+  readonly_fs: true
```
efekarakus: Sweet, can we also add a comment next to it (aligned like the other fields), saying something like:
```yaml
storage:
   readonly_fs: true       # Limit to read-only access to mounted root filesystems by default.
```

efekarakus: Do we need a check to write this value if `{{- if .Platform.PlatformString}}` is not Windows?

efekarakus (commented):
Looks great! two small things 🙏

internal/pkg/template/templates/workloads/services/backend/manifest.yml
KollaAdithya: I will populate readonly_fs field to true in the manifest file only when the platform is not Windows. If it is Windows I will not at all populate this field. Does this sounds good?

efekarakus (approved):
Awesome! My apologies, for the super late review. It looks great!

PR #4068, opened 2022-10-06
fix: Env manifest UX fix for http config 

paragbhingre:
This PR fix multiple problems - 
1. This PR fix the env manifest fields that were organized in incorrect manner.
2. Also resolves #4032 
3. `copilot env init` with private subnets create env manifest with following field
```
http:
  private:
```
This should only be created when there is at least one parameter inside private is enabled. This PR fix this nit as well. 


Lou1415926:
Just making sure I'm understanding correctly (also for other reviewers' reference!)
We currently have:
```yaml
http:
  public:
    security_groups:
      ingress: # This doesn't make sense because `ingress` is for one sg, not sgs
        restrict_to: # This is redundant because `ingress` implies "restrict_to".
          cdn: true
  private:
    security_groups:
      ingress: # This doesn't make sense because `ingress` is for one sg, not sgs
        from_vpc: true
```

After the fix, we want to have:
```yaml
http:
  public:
    ingress:
      cdn: true
  private:
    ingress:
      from_vpc: true
```
while still accepting the old manifest.

paragbhingre:
> Just making sure I'm understanding correctly (also for other reviewers' reference!) We currently have:
> 
> ```yaml
> http:
>   public:
>     security_groups:
>       ingress: # This doesn't make sense because `ingress` is for one sg, not sgs
>         restrict_to: # This is redundant because `ingress` implies "restrict_to".
>           cdn: true
>   private:
>     security_groups:
>       ingress: # This doesn't make sense because `ingress` is for one sg, not sgs
>         from_vpc: true
> ```
> 
> After the fix, we want to have:
> 
> ```yaml
> http:
>   public:
>     ingress:
>       cdn: true
>   private:
>     ingress:
>       from_vpc: true
> ```
> 
> while still accepting the old manifest.

Yes you are right @Lou1415926

internal/pkg/manifest/env.go
```diff
@@ -494,25 +498,20 @@ func (cfg *EnvironmentConfig) ELBAccessLogs() (*ELBAccessLogsArgs, bool) {
 	return &accessLogs.AdvancedConfig, true
 }
 
-// ALBSecurityGroupsConfig represents security group configuration settings for an ALB.
-type ALBSecurityGroupsConfig struct {
-	Ingress Ingress `yaml:"ingress"`
+// DeprecatedALBSecurityGroupsConfig represents security group configuration settings for an ALB.
+type DeprecatedALBSecurityGroupsConfig struct {
+	Ingress DeprecatedIngress `yaml:"ingress"` // Deprecated. This field not available inside PublicHTTPConfig.Ingress and privateHTTPConfig.Ingress field.
```
Lou1415926:
```suggestion
	Ingress DeprecatedIngress `yaml:"ingress"` // Deprecated. This field now available inside PublicHTTPConfig.Ingress and privateHTTPConfig.Ingress field.
```

internal/pkg/manifest/env.go
```diff
@@ -521,13 +520,24 @@ type RestrictiveIngress struct {
 	CDNIngress *bool `yaml:"cdn"`
 }
 
+// RelaxedIngress represents ingress field which relaxes default
+// behavior of not allowing any public ingress.
```
Lou1415926: nit:
```suggestion
// RelaxedIngress contains ingress fields that relax the default ingress.
```

internal/pkg/manifest/env.go
```diff
@@ -537,20 +547,13 @@ func (cfg PublicHTTPConfig) IsEmpty() bool {
 }
 
 type privateHTTPConfig struct {
-	InternalALBSubnets   []string             `yaml:"subnets,omitempty"`
-	Certificates         []string             `yaml:"certificates,omitempty"`
-	SecurityGroupsConfig securityGroupsConfig `yaml:"security_groups,omitempty"`
+	InternalALBSubnets   []string                          `yaml:"subnets,omitempty"`
+	Certificates         []string                          `yaml:"certificates,omitempty"`
+	SecurityGroupsConfig DeprecatedALBSecurityGroupsConfig `yaml:"security_groups,omitempty"`
```
Lou1415926: nit:
```suggestion
	SecurityGroupsConfig DeprecatedALBSecurityGroupsConfig `yaml:"security_groups,omitempty"` // Deprecated.
```
Adding the comment (like you do with the `publicHTTPConfig` struct) helps the IDE to show the field differently!

internal/pkg/manifest/env.go
```diff
@@ -97,9 +97,12 @@ type EnvironmentConfig struct {
...
 
-// IsIngressRestrictedToCDN returns whether or not an environment has its
+// ALBIngressRestrictedToCDN returns whether an environment has its
 // Public Load Balancer ingress restricted to a Content Delivery Network.
-func (mft *EnvironmentConfig) IsIngressRestrictedToCDN() bool {
+func (mft *EnvironmentConfig) ALBIngressRestrictedToCDN() bool {
+	if !mft.HTTPConfig.Public.Ingress.IsEmpty() {
+		return aws.BoolValue(mft.HTTPConfig.Public.Ingress.CDNIngress)
+	}
 	return aws.BoolValue(mft.HTTPConfig.Public.SecurityGroupConfig.Ingress.RestrictiveIngress.CDNIngress)
```
Lou1415926: We'll probably need some comments here:
```suggestion
        // Check the fixed manifest first. This would be `http.public.ingress.cdn`.
        // For more information, see https://github.com/aws/copilot-cli/pull/4068#issuecomment-1275080333
	if !mft.HTTPConfig.Public.Ingress.IsEmpty() {
		return aws.BoolValue(mft.HTTPConfig.Public.Ingress.CDNIngress)
	}
	// Fall through to the old manifest: `http.public.security_groups.ingress.cdn`.
	return aws.BoolValue(mft.HTTPConfig.Public.SecurityGroupConfig.Ingress.RestrictiveIngress.CDNIngress)
```

internal/pkg/manifest/validate_env.go
```diff
@@ -290,6 +290,12 @@ func (cfg EnvironmentHTTPConfig) validate() error {
 
 // validate returns nil if PublicHTTPConfig is configured correctly.
 func (cfg PublicHTTPConfig) validate() error {
+	if !cfg.SecurityGroupConfig.Ingress.IsEmpty() && !cfg.Ingress.IsEmpty() {
```
Lou1415926: To elaborate a bit for fellow reviewer's thoughts  :

`cfg.SecurityGroupConfig.Ingress` is the existing field that has been there in multiple releases, `cfg.Ingress` is the new field that fixes it (in one of the upcoming release).

Since it's a new field - if it's non-empty, it must mean that the user has modified their manifest, adding configurations under `http.public.ingress`. Erroring out in this case is not "breaking".

On the other hand, manifests with `http.public.security_groups.ingress` field configured will keep working - given that they have not been modified to use `http.public.ingress`.

internal/pkg/manifest/validate_env.go
```diff
@@ -318,12 +327,21 @@ func (al ELBAccessLogsArgs) validate() error {
 }
 
 // validate returns nil if ALBSecurityGroupsConfig is configured correctly.
-func (cfg ALBSecurityGroupsConfig) validate() error {
+func (cfg DeprecatedALBSecurityGroupsConfig) validate() error {
+	if cfg.Ingress.IsEmpty() {
+		return nil
+	}
```
Lou1415926: Sharing my thoughts:

This if-statement is skipping validating `Ingress` if it's empty. I think we should validate every single field no matter what. Whether `cfg.Ingress.IsEmpty` makes it a valid input should be `cfg.Ingress.validate()`'s call.

In terms of validation result, this doesn't make any difference here. I just wanted to share my thoughts on how I think about `validate`

internal/pkg/template/env.go
```diff
@@ -123,10 +123,11 @@ type EnvOpts struct {
...
 type HTTPConfig struct {
-	CIDRPrefixListIDs []string
-	ImportedCertARNs  []string
-	CustomALBSubnets  []string
-	ELBAccessLogs     *ELBAccessLogs
+	CIDRPrefixListIDs           []string
+	ImportedCertARNs            []string
+	CustomALBSubnets            []string
+	ELBAccessLogs               *ELBAccessLogs
+	EnableInternalALBVPCIngress *bool
```
Lou1415926: I know you are looking to remove this field - just making sure!

internal/pkg/template/env.go
```diff
@@ -152,7 +153,7 @@ type CDNConfig struct {
 type VPCConfig struct {
 	Imported            *ImportVPC // If not-nil, use the imported VPC resources instead of the Managed VPC.
 	Managed             ManagedVPC
-	AllowVPCIngress     bool
+	AllowVPCIngress     *bool
```
Lou1415926: Why do we need a pointer here?

internal/pkg/manifest/errors.go
```diff
@@ -98,6 +98,29 @@ func (e *errFieldMutualExclusive) Error() string {
...
+	return fmt.Sprintf(`
+It looks like you have specified mutually exclusive fields for Load Balancer configuration.
+The "%s.http.security_group.ingress" has been deprecated and we encourage you to use "%s.http.ingress".
+Following is the example manifest fields that we recommend you to use for your environment.
+
+http:
+  %s:
+    ingress:
+      cdn: true
+`, privateOrPublicField, privateOrPublicField, privateOrPublicField)
```
Lou1415926: What do you think of a more customized recommended actions?

Something I can think of is we can probably detect what they have in `http.public.security_groups.ingress` and just tell them how to migrate to `http.public.ingress` exactly (instead of giving a general example). 

For example, if we see `http.public.security_groups.ingress.cdn: true` and `http.public.ingress.anewfield: ture` , we could probably just straight up produce the following snippet:
```yaml
http:
  public:
     ingress:
        cdn: true # Migrated from http.public.security_groups.ingress.cdn.
        anewfield: true
```

internal/pkg/manifest/env_test.go
```diff
@@ -509,6 +504,36 @@ http:
 				},
 			},
 		},
+		"unmarshal with new http fields": {
```
Lou1415926, line 507: Let's also add a use case that unmarshal an "invalid" manifest that has both old ingress and new ingress and see how it unmarshals.
```yaml
http:
  public:
     ingresS:
       cdn: true
     security_groups:
        ingress:
           cdn: true
```

internal/pkg/cli/env_init.go
```diff
@@ -823,7 +830,7 @@ func (o *initEnvOpts) writeManifest() (string, error) {
 		VPCConfig:                   o.adjustVPCConfig(),
 		ImportCertARNs:              o.importCerts,
 		InternalALBSubnets:          o.internalALBSubnets,
-		EnableInternalALBVPCIngress: o.allowVPCIngress,
+		EnableInternalALBVPCIngress: o.setAllowVPCIngress(),
```
Lou1415926: Why do we need this change?

internal/pkg/config/env.go
```diff
@@ -34,15 +34,15 @@ type CustomizeEnv struct {
 	VPCConfig                   *AdjustVPC `json:"adjustVPC,omitempty"`
 	ImportCertARNs              []string   `json:"importCertARNs,omitempty"`
 	InternalALBSubnets          []string   `json:"internalALBSubnets,omitempty"`
-	EnableInternalALBVPCIngress bool       `json:"enableInternalALBVPCIngress,omitempty"`
+	EnableInternalALBVPCIngress *bool      `json:"enableInternalALBVPCIngress,omitempty"`
```
Lou1415926: Same question here. Why do we need a pointer?

internal/pkg/cli/env_init.go
```diff
@@ -672,6 +672,13 @@ func (o *initEnvOpts) adjustVPCConfig() *config.AdjustVPC {
 	}
 }
 
+func (o *initEnvOpts) setAllowVPCIngress() *bool {
+	if o.allowVPCIngress {
+		return aws.Bool(o.allowVPCIngress)
```
efekarakus: I feel like this is a bit strange 🤔 If i'm following this the same as setting:
```suggestion
		return aws.Bool(true)
```
I agree with Wanxian's feedback below, I can't tell what the pointer is allowing us to do instead of just a `bool`

internal/pkg/config/env.go
efekarakus: Is it because you want to differentiate between `nil` and `&false`? I feel like if this field is `false` then it still means CustomizeEnv is empty

internal/pkg/cli/deploy/env.go
```diff
@@ -280,7 +280,7 @@ func (d *envDeployer) cidrPrefixLists(in *DeployEnvironmentInput) ([]string, err
 	var cidrPrefixListIDs []string
 
 	// Check if ingress is allowed from cloudfront
-	if in.Manifest == nil || !in.Manifest.IsIngressRestrictedToCDN() {
+	if in.Manifest == nil || !in.Manifest.ALBIngressRestrictedToCDN() {
```
efekarakus: nit:
```suggestion
	if in.Manifest == nil || !in.Manifest.IsPublicLBIngressRestrictedToCDN() {
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -423,11 +424,18 @@ func (e *EnvStackConfig) vpcConfig() (template.VPCConfig, error) {
 	return template.VPCConfig{
 		Imported:            e.importVPC(),
 		Managed:             e.managedVPC(),
-		AllowVPCIngress:     aws.BoolValue(e.in.Mft.HTTPConfig.Private.SecurityGroupsConfig.Ingress.VPCIngress),
+		AllowVPCIngress:     e.isAllowVPCIngress(),
 		SecurityGroupConfig: securityGroupConfig,
 	}, nil
 }
 
+func (e *EnvStackConfig) isAllowVPCIngress() *bool {
```
efekarakus: can this be instead a method in the `manifest` package similar to `IsPublicLBIngressRestrictedToCDN`:
`IsPrivateLBIngressAllowingVPC` ?

internal/pkg/manifest/env.go
```diff
@@ -422,17 +425,18 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
...
 
 // PublicHTTPConfig represents the configuration settings for an environment public ALB.
 type PublicHTTPConfig struct {
-	SecurityGroupConfig ALBSecurityGroupsConfig `yaml:"security_groups,omitempty"`
-	Certificates        []string                `yaml:"certificates,omitempty"`
-	ELBAccessLogs       ELBAccessLogsArgsOrBool `yaml:"access_logs,omitempty"`
+	SecurityGroupConfig DeprecatedALBSecurityGroupsConfig `yaml:"security_groups,omitempty"` // Deprecated. This configuration is now available inside Ingress field.
+	Certificates        []string                          `yaml:"certificates,omitempty"`
+	ELBAccessLogs       ELBAccessLogsArgsOrBool           `yaml:"access_logs,omitempty"`
+	Ingress             RestrictiveIngress
```
efekarakus: I am confused 😝 is this field unmarshaled properly used?
If I look at the [proposed manifest schema](https://github.com/aws/copilot-cli/pull/4068#issuecomment-1275080333), I would have expected:
```
Ingress RestrictiveIngress  `yaml:"ingress"`
```
Am I missing something?

internal/pkg/manifest/env.go
```diff
@@ -494,25 +498,20 @@ func (cfg *EnvironmentConfig) ELBAccessLogs() (*ELBAccessLogsArgs, bool) {
 	return &accessLogs.AdvancedConfig, true
 }
 
-// ALBSecurityGroupsConfig represents security group configuration settings for an ALB.
-type ALBSecurityGroupsConfig struct {
-	Ingress Ingress `yaml:"ingress"`
+// DeprecatedALBSecurityGroupsConfig represents security group configuration settings for an ALB.
```
efekarakus: What do you think of moving these deprecated types under a new file: `deprecated.go`?

internal/pkg/manifest/env.go
efekarakus: If we can let's not reuse the word "relax" or "restrict" in the comments to add some more clarity:
```suggestion
// RelaxedIngress contains ingress configuration to add to a security group.
```

internal/pkg/manifest/errors.go
```diff
@@ -98,6 +98,29 @@ func (e *errFieldMutualExclusive) Error() string {
...
+type errSpecifiedBothIngressFields struct {
+	firstField  string
+	secondField string
+}
+
+func (e *errSpecifiedBothIngressFields) Error() string {
+	return fmt.Sprintf(`must specify one, not both, of "%s" and "%s"`, e.firstField, e.secondField)
+}
+
+func (e *errSpecifiedBothIngressFields) RecommendActions() string {
```
efekarakus, line 111: nit: doc comments

internal/pkg/manifest/env.go
```diff
@@ -422,17 +425,18 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
...
 	}
 	cfg.Public.Certificates = env.ImportCertARNs
 }
 
 // PublicHTTPConfig represents the configuration settings for an environment public ALB.
 type PublicHTTPConfig struct {
-	SecurityGroupConfig ALBSecurityGroupsConfig `yaml:"security_groups,omitempty"`
-	Certificates        []string                `yaml:"certificates,omitempty"`
-	ELBAccessLogs       ELBAccessLogsArgsOrBool `yaml:"access_logs,omitempty"`
+	SecurityGroupConfig DeprecatedALBSecurityGroupsConfig `yaml:"security_groups,omitempty"` // Deprecated. This configuration is now available inside Ingress field.
```
efekarakus: Should we rename the field as well to start with `Deprecated` so that it's easy to follow when it's being accessed:
```suggestion
	DeprecatedSG DeprecatedALBSecurityGroupsConfig `yaml:"security_groups,omitempty"` // Deprecated. This configuration is now available inside Ingress field.
```

internal/pkg/template/env.go
```diff
@@ -123,10 +123,11 @@ type EnvOpts struct {
 
 // HTTPConfig represents configuration for a Load Balancer.
 type HTTPConfig struct {
```
efekarakus: Why do we need these changes 🤔 I don't see any CloudFormation template modifications am I misunderstanding something?

internal/pkg/template/templates/environment/manifest.yml
```diff
@@ -64,6 +64,10 @@ http:
     {{- if $privateHTTP.InternalALBSubnets}}
     subnets: {{$privateHTTP.InternalALBSubnets}}
     {{- end}}
+    {{- if and $privateHTTP.Ingress $privateHTTP.Ingress.VPCIngress}}
+    ingress:
+      from_vpc: {{ $privateHTTP.Ingress.VPCIngress }}
```
efekarakus, line 69:
```suggestion
      from_vpc: true
```

internal/pkg/config/env.go
paragbhingre: The reason why made this change because if we do not keep `EnableInternalALBVPCIngress` as pointer then this will always be `false` and in that case if customer imports vpc config with private subnet without `--internal-alb-allow-vpc-ingress` flag then also manifest has following config specified. 

```
# Configure the load balancers in your environment, once created.
http:
  private:
    ingress:
      from_vpc: false

``` 

In order to get rid of this part I made changes related to pointer. Does that make sense?

But right now I have removed pointer logic and kept the existing logic as it is.

internal/pkg/manifest/env.go
paragbhingre: Yes I got confused here with the multiple fields and thought that we are going to handle this ingress field internally through `security_groups` itself. But I am changing this one.

internal/pkg/cli/env_init.go
paragbhingre: Related - https://github.com/aws/copilot-cli/pull/4068/commits/0a965e81a3eab7d32bc354d4a18d1816426ba977#r997427724

internal/pkg/template/templates/environment/manifest.yml
paragbhingre: `{{ $privateHTTP.Ingress.VPCIngress }}` is a pointer so it check for if it exist in `if` condition and if it set then the same value gets assigned to `from_vpc` field. 
The reason behind keeping `from_vpc` as pointer because we are accepting this field from manifest where it is easier to validate the field if it pointer in order to figure out if it set or not through its `nil` value. In this case we are accepting this value from `--internal-alb-allow-vpc-ingress` as well which is not a pointer and its value is set to false when it is not being used. But if we make `from_vpc: true` in the manifest.yml then whenever `--internal-alb-allow-vpc-ingress`  is even false then also its value will be set to `$privateHTTP.Ingress.VPCIngress` pointer which is not nil anymore and it will set `from_vpc: true` which is wrong. 
Does that make sense?

internal/pkg/template/env.go
paragbhingre: Removed these changes.

internal/pkg/manifest/env.go
paragbhingre: good idea.

internal/pkg/template/env.go
paragbhingre: Removed in the new implementation. Removed all the unnecessary logic regarding  EnableInternalALBVPCIngress and related to that.

internal/pkg/config/env.go
paragbhingre: But right now I have removed pointer logic and kept the existing logic as it is.

efekarakus: Could we just not write to the manifest if the field is false? 🤔

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -423,7 +423,7 @@ func (e *EnvStackConfig) vpcConfig() (template.VPCConfig, error) {
 	return template.VPCConfig{
 		Imported:            e.importVPC(),
 		Managed:             e.managedVPC(),
-		AllowVPCIngress:     aws.BoolValue(e.in.Mft.HTTPConfig.Private.SecurityGroupsConfig.Ingress.VPCIngress),
+		AllowVPCIngress:     aws.BoolValue(e.in.Mft.HTTPConfig.Private.Ingress.VPCIngress) || aws.BoolValue(e.in.Mft.HTTPConfig.Private.SecurityGroupsConfig.DeprecatedIngress.VPCIngress),
```
efekarakus: nit: what do you think of adding a method to `manifest.privateHTTPConfig` that returns true or false for this condition?
```go
package manifest

// HasVPCIngress returns true if the private ALB allows ingress from within the VPC.
func (cfg privateHTTPConfig) HasVPCIngress() bool
```

internal/pkg/manifest/deprecated.go
```diff
@@ -0,0 +1,25 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
```
efekarakus, line 3: nit: can we yoink the unit tests from `manifest/env_test.go` to `manifest/deprecated_test.go` for these types?

internal/pkg/manifest/env.go
```diff
@@ -422,17 +428,18 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
...
 
 // PublicHTTPConfig represents the configuration settings for an environment public ALB.
 type PublicHTTPConfig struct {
-	SecurityGroupConfig ALBSecurityGroupsConfig `yaml:"security_groups,omitempty"`
-	Certificates        []string                `yaml:"certificates,omitempty"`
-	ELBAccessLogs       ELBAccessLogsArgsOrBool `yaml:"access_logs,omitempty"`
+	DeprecatedSG  DeprecatedALBSecurityGroupsConfig `yaml:"security_groups,omitempty"` // Deprecated. This configuration is now available inside Ingress field.
+	Certificates  []string                          `yaml:"certificates,omitempty"`
+	ELBAccessLogs ELBAccessLogsArgsOrBool           `yaml:"access_logs,omitempty"`
+	Ingress       RestrictiveIngress
```
efekarakus: Do we need the yaml tag here?
```suggestion
	Ingress       RestrictiveIngress `yaml:"ingress,omitempty"`
```

internal/pkg/manifest/env.go
```diff
@@ -494,63 +501,40 @@ func (cfg *EnvironmentConfig) ELBAccessLogs() (*ELBAccessLogsArgs, bool) {
...
+	return len(cfg.Certificates) == 0 && cfg.DeprecatedSG.IsEmpty() && cfg.ELBAccessLogs.isEmpty()
 }
 
 type privateHTTPConfig struct {
-	InternalALBSubnets   []string             `yaml:"subnets,omitempty"`
-	Certificates         []string             `yaml:"certificates,omitempty"`
-	SecurityGroupsConfig securityGroupsConfig `yaml:"security_groups,omitempty"`
+	InternalALBSubnets   []string                          `yaml:"subnets,omitempty"`
+	Certificates         []string                          `yaml:"certificates,omitempty"`
+	SecurityGroupsConfig DeprecatedALBSecurityGroupsConfig `yaml:"security_groups,omitempty"` // Deprecated. This field is now available in Ingress.
```
efekarakus: nit: for consistency with public http config:
```suggestion
	DeprecatedSG DeprecatedALBSecurityGroupsConfig `yaml:"security_groups,omitempty"` // Deprecated. This field is now available in Ingress.
```

internal/pkg/manifest/env.go
```diff
@@ -494,63 +501,40 @@ func (cfg *EnvironmentConfig) ELBAccessLogs() (*ELBAccessLogsArgs, bool) {
...
 
-// IsEmpty returns true if there are no specified fields for ingress.
-func (i Ingress) IsEmpty() bool {
-	return i.VPCIngress == nil && i.RestrictiveIngress.IsEmpty()
-}
-
 // IsEmpty returns true if there is no customization to the public ALB.
 func (cfg PublicHTTPConfig) IsEmpty() bool {
-	return len(cfg.Certificates) == 0 && cfg.SecurityGroupConfig.IsEmpty() && cfg.ELBAccessLogs.isEmpty()
+	return len(cfg.Certificates) == 0 && cfg.DeprecatedSG.IsEmpty() && cfg.ELBAccessLogs.isEmpty()
```
efekarakus: Do we not need
```suggestion
	return len(cfg.Certificates) == 0 && cfg.DeprecatedSG.IsEmpty() && cfg.ELBAccessLogs.isEmpty() && cfg.Ingress.IsEmpty()
```

internal/pkg/config/env.go
paragbhingre: I was not sure how right it is to not write that field to the manifest when it is `false`. But now I have made changes in the new commit, let me know if that works.

internal/pkg/manifest/errors.go
paragbhingre: As per the discussion keeping the same error message.

internal/pkg/template/templates/environment/manifest.yml
efekarakus: Gotcha, I think I get it.

So I wonder if that means that we have to fix this line:
https://cs.github.com/aws/copilot-cli/blob/df59724e6c4d7bd5a7844795d1b738abb2a25f3c/internal/pkg/manifest/env.go#L425


```go
if env.EnableInternalALBVPCIngress {
     cfg.Private.Ingress.VPCIngress = aws.Bool(true)
}
```
This way the field will be non-nil only if the SSM configuration is set to true, and only then it will be rendered in the manifest.

internal/pkg/manifest/env.go
paragbhingre: I just figured that out and was sending out the commit before you see it 😁, but you were quicker 😎

internal/pkg/manifest/deprecated.go
paragbhingre: There are no specific tests related to these types. There are couple for Unmarshal `EnvironmentConfig` but not for the specific type.

internal/pkg/cli/deploy/env_test.go
```diff
@@ -293,8 +293,8 @@ func TestEnvDeployer_DeployEnvironment(t *testing.T) {
 				EnvironmentConfig: manifest.EnvironmentConfig{
 					HTTPConfig: manifest.EnvironmentHTTPConfig{
 						Public: manifest.PublicHTTPConfig{
-							SecurityGroupConfig: manifest.ALBSecurityGroupsConfig{
-								Ingress: manifest.Ingress{
+							DeprecatedSG: manifest.DeprecatedALBSecurityGroupsConfig{
```
Lou1415926, line 296: What do you think of using the new fields instead of the deprecated ones here?

internal/pkg/manifest/env.go
```diff
@@ -494,63 +503,45 @@ func (cfg *EnvironmentConfig) ELBAccessLogs() (*ELBAccessLogsArgs, bool) {
...
-
 // RestrictiveIngress represents ingress fields which restrict
 // default behavior of allowing all public ingress.
 type RestrictiveIngress struct {
 	CDNIngress *bool `yaml:"cdn"`
 }
 
+// RelaxedIngress contains ingress configuration to add to a security group.
+type RelaxedIngress struct {
+	VPCIngress *bool `yaml:"from_vpc"`
```
Lou1415926: what do we think about renaming `from_vpc` to `vpc`?

internal/pkg/manifest/errors.go
```diff
@@ -98,6 +98,30 @@ func (e *errFieldMutualExclusive) Error() string {
...
+
+func (e *errSpecifiedBothIngressFields) Error() string {
+	return fmt.Sprintf(`must specify one, not both, of "%s" and "%s"`, e.firstField, e.secondField)
+}
+
+// RecommendActions returns recommended actions to be taken after the error.
+func (e *errSpecifiedBothIngressFields) RecommendActions() string {
+	privateOrPublicField := strings.Split(e.firstField, ".")[0]
+	return fmt.Sprintf(`
+It looks like you have specified mutually exclusive fields for Load Balancer configuration.
```
Lou1415926: I think we can be more specific here. For example, for a customer that has `http.public.security_groups.ingress.cdn` and `http.ingress.source_ips`, we can say something like 

> "It looks like you specified ingress under both `http.public.security_groups.ingress` and `http.public.ingress`. 
> After Copilot v1.X, we have deprecated `http.public.security_groups.ingress` in favor of `http.public.ingress`.  This means that `http.public.security_groups.ingress.cdn`  is removed in favor of `http.public.ingress.cdn`:
> <manifest example>

Just an idea to better communicate the solution to customers so that there is no :panic

Lou1415926 (commented):
Do you think this sort of manifest UX fix is worth testing in a regression test? Like we can test a manifest that looks like:
```
http:
   public: 
        ingress:
             from_cdn: true
```
works the same before & after?

internal/pkg/cli/deploy/env_test.go
paragbhingre: I thought of that but I felt that keeping old fields gives us more confidence that existing things are not breaking in any of those use cases. Because we don't want to break anything existing.

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
```diff
@@ -318,3 +314,127 @@ network:
...
+						"UniqueJSONValuesFunction":      "https://mockbucket.s3-us-west-2.amazonaws.com/unique-json-values",
+					},
+					Mft:    &mft,
+					RawMft: []byte(rawMft),
+				}
+			}(),
+			wantedFileName: "template-with-old-and-new-http-fields.yml",
+		},
+	}
+	for name, tc := range testCases {
```
Lou1415926, line 417: yayy thank you for the tests! I have one nit thought 💭 

I can see how this test suite can be used to ensure that the old/new manifest produce the same CFN template, effectively speaking. My nit thought is more from a semantic perspective - that these test does not convey the purpose that "we want to ensure the CFN template produced by the old manifest is the same as the new manifest". 

Imagine a new contributor coming to the codebase and see these tests, it wouldn't be obvious to them that *these two tests are meant to be viewed together*, and that their wanted template *has to point to the same template in order for them to serve their purpose*. 

Not knowing this, the contributor may modify one of these test cases (e.g. adding a new manifest field, changing the wanted template to a different one). The purpose of these tests will be lost.

I have a code snippet that shows an alternative way to structure the test, hopefully that will better convey the intention of the test:

```go
func TestEnvStack_Regression(t *testing.T) {
	t.Run("produce the same template after migrating environment manifest to the new field hierarchy", func(t *testing.T) {
		originalManifest := `name: test
type: Environment
http:
  public:
    security_groups:
      ingress:
        restrict_to:
          cdn: true
  private:
    security_groups:
      ingress:
        from_vpc: true`
		// NOTE: in newManifest, `http.public.security_groups.ingress.restrict_to.cdn` is migrated to `http.public.ingress.cdn`.
		// and `http.private.security_groups.ingress.from_vpc` is migrated to `http.private.ingress.vpc`.
		newManifest := `name: test
type: Environment
http:
  public:
    ingress:
      cdn: true
  private:
    ingress:
      vpc: true`

		// Get the template content given the original manifest.
		originalStack := stack.NewEnvStackConfig(&deploy.CreateEnvironmentInput{
			RawMft: []byte(originalManifest),
			// ...
		})
		originalTmpl, err := originalStack.Template()
		require.NoError(t, err, "should serialize the template given the original environment manifest")
		originalObj := make(map[any]any)
		require.NoError(t, yaml.Unmarshal([]byte(originalTmpl), originalObj))

		// Get the template content given the manifest after migrating to the new UX.
		newStack := stack.NewEnvStackConfig(&deploy.CreateEnvironmentInput{
			RawMft: []byte(newManifest),
			// ...
		})
		newTmpl, err := newStack.Template()
		require.NoError(t, err, "should serialize the template given a migrated environment manifest")
		newObj := make(map[any]any)
		require.NoError(t, yaml.Unmarshal([]byte(newTmpl), newObj))

		// Check that the template content does not change after migrating to new manifest hierarchy.
		delete(originalObj, "Metadata")
		delete(newObj, "Metadata")
		require.Equal(t, originalObj, newObj)

	})
}
```

paragbhingre: Sounds like a good idea to me.

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-old-and-new-http-fields.yml
```diff
@@ -0,0 +1,1104 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
Lou1415926: do we still need this file?

internal/pkg/manifest/env_test.go
```diff
@@ -509,6 +504,103 @@ http:
...
+			wantedStruct: &Environment{
+				Workload: Workload{
+					Name: aws.String("prod"),
+					Type: aws.String("Environment"),
+				},
+				EnvironmentConfig: EnvironmentConfig{
+					HTTPConfig: EnvironmentHTTPConfig{
+						Public: PublicHTTPConfig{
+							Certificates: []string{"cert-1", "cert-2"},
+							DeprecatedSG: DeprecatedALBSecurityGroupsConfig{DeprecatedIngress: DeprecatedIngress{RestrictiveIngress: RestrictiveIngress{CDNIngress: aws.Bool(true)}}},
```
Lou1415926: nit: (it's difficult to indent using gh editor but let's not squash them into one line 🥺 )
```suggestion
							DeprecatedSG: DeprecatedALBSecurityGroupsConfig{
							  DeprecatedIngress: DeprecatedIngress{
							    RestrictiveIngress: RestrictiveIngress{
							      CDNIngress: aws.Bool(true)
							      }
							    }
							   },
```

internal/pkg/manifest/errors.go
```diff
@@ -98,6 +98,45 @@ func (e *errFieldMutualExclusive) Error() string {
...
+func (e *errSpecifiedBothIngressFields) Error() string {
+	return fmt.Sprintf(`must specify one, not both, of "%s" and "%s"`, e.firstField, e.secondField)
+}
+
+// RecommendActions returns recommended actions to be taken after the error.
+func (e *errSpecifiedBothIngressFields) RecommendActions() string {
+	privateOrPublicField := strings.Split(e.firstField, ".")[0]
+	if privateOrPublicField == "public" {
+		return `
+It looks like you specified ingress under both "http.public.security_groups.ingress"" and "http.public.ingress"".
```
Lou1415926: <strike> This suggestion doesn't seem to work with `from_vpc` :(. </strike>

Lou1415926 (commented):
lgtm!

internal/pkg/manifest/env.go
```diff
@@ -423,17 +429,20 @@ func (cfg *EnvironmentHTTPConfig) loadLBConfig(env *config.CustomizeEnv) {
 	if env.ImportVPC != nil && len(env.ImportVPC.PublicSubnetIDs) == 0 {
 		cfg.Private.InternalALBSubnets = env.InternalALBSubnets
 		cfg.Private.Certificates = env.ImportCertARNs
-		cfg.Private.SecurityGroupsConfig.Ingress.VPCIngress = aws.Bool(env.EnableInternalALBVPCIngress)
+		if env.EnableInternalALBVPCIngress {
+			cfg.Private.Ingress.VPCIngress = aws.Bool(true)
+		}
```
Lou1415926: nit: maybe we can add a comment to explain this line, so that we don't accidentally remove it in the future. 
```suggestion
		if env.EnableInternalALBVPCIngress {
			cfg.Private.Ingress.VPCIngress = aws.Bool(true) // NOTE: Do not load the configuration unless it's positive, so that the default manifest does not contain the unnecessary line `http.private.ingress.vpc: false`.
		}
```

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
```diff
@@ -318,3 +314,125 @@ network:
 		})
 	}
 }
+
+func TestEnvStack_Regression(t *testing.T) {
+	testCases := map[string]struct {
+		originalManifest *deploy.CreateEnvironmentInput
+		newManifest      *deploy.CreateEnvironmentInput
+	}{
+		"produce the same template after migrating environment manifest to the new field hierarchy": {
```
efekarakus: This is such a good integ test ❤️ !

Just a tiny nit: can we make the test case name more specific?
```suggestion
		"should produce the same template after migrating load balancer ingress fields": {
```

internal/pkg/manifest/env.go
efekarakus: A slightly different way is capturing the comment in a new unit test case.

You can add a unit test that says something like:
```
"should only enable private LB VPC ingress if the field is set in SSM": {
  // We don't want to render the manifest with the value set to false, so only assign to non-nil if the value is true.
  ...
}
```

internal/pkg/manifest/errors.go
```diff
@@ -98,6 +98,45 @@ func (e *errFieldMutualExclusive) Error() string {
...
+After Copilot v1.23.0, we have deprecated "http.public.security_groups.ingress" in favor of "http.public.ingress". 
+This means that "http.public.security_groups.ingress.cdn"" is removed in favor of "http.public.ingress.cdn".
+With the new release manifest configuration for cdn looks like:
```
efekarakus: nit looks like there is accidentally extra double quotes in the string
```suggestion
		return `
It looks like you specified ingress under both "http.public.security_groups.ingress" and "http.public.ingress".
After Copilot v1.23.0, we have deprecated "http.public.security_groups.ingress" in favor of "http.public.ingress". 
This means that "http.public.security_groups.ingress.cdn" is removed in favor of "http.public.ingress.cdn".
With the new release manifest configuration for cdn looks like:
```

efekarakus: same for the return statement

efekarakus (approved):
Looks good, just tiny nits. Feel free to remove the label afterwards!

Lou1415926 (approved):
:shipit: !!

PR #4075, opened 2022-10-10
feat: add aurora serverless v2 in storage

### LINKED ISSUE #3493: Supporting Aurora Serverless v2
Since now Aurora Serverless v2 is generally available (https://aws.amazon.com/blogs/aws/amazon-aurora-serverless-v2-is-generally-available-instant-scaling-for-demanding-workloads/), are there any plans to support this improved version of Aurora Serverless as a built-in supported storage type?

hkford:
Added `--serverless-version` flag in `storage init` command.

```text
Aurora Serverless Flags
      --engine string               The database engine used in the cluster.
                                    Must be either "MySQL" or "PostgreSQL".
      --initial-db string           The initial database to create in the cluster.
      --parameter-group string      Optional. The name of the parameter group to associate with the cluster.
      --serverless-version string   Aurora Serverless version.
                                    Must be of either "V1" or "V2"
```

When you pass `--serverless-version V1` the `storage init` command generates an addon template of Aurora Serverless V1 cluster the same as before. When you pass `--serverless-version V2` the `storage init` command generates an addon template of Aurora Serverless V2 cluster, which is [recently released](https://aws.amazon.com/jp/about-aws/whats-new/2022/10/amazon-aurora-serverless-v2-supports-aws-cloudformation/).

Fixes #3493 


internal/pkg/addon/storage.go
```diff
@@ -15,11 +15,19 @@ import (
...
+	rdsTemplatePath       = "addons/aurora/cf.yml"
+	rdsV2TemplatePath     = "addons/aurora/serverlessv2.yml"
+	rdsRDWSTemplatePath   = "addons/aurora/rdws/cf.yml"
+	rdsRDWSV2TemplatePath = "addons/aurora/rdws/serverlessv2.yml"
+	rdsRDWSParamsPath     = "addons/aurora/rdws/addons.parameters.yml"
+)
+
+const (
+	// Aurora Serverless versions.
+	ServerlessVersionV1 = "V1"
```
huanjani: Maybe we should lowercase the 'v' (here, flag description, prompt, examples, etc.), based on https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/aurora-serverless-v2.html and Go/Semantic Versioning conventions. The flag should accept both upper and lower case input, though.

internal/pkg/cli/flag.go
```diff
@@ -278,6 +279,8 @@ Must be of the format '<keyName>:<dataType>'.`
 	storageNoLSIFlagDescription     = `Optional. Don't ask about configuring alternate sort keys.`
 	storageLSIConfigFlagDescription = `Optional. Attribute to use as an alternate sort key. May be specified up to 5 times.
 Must be of the format '<keyName>:<dataType>'.`
+	storageRDSServerlessVersionFlagDescription = `Aurora Serverless version.
+Must be of either "V1" or "V2"`
```
huanjani: nit: 
```suggestion
Must be either "V1" or "V2".`
```

hkford: Fixed

internal/pkg/addon/storage.go
hkford: I lowercased the `serverless-version` flag to `v1` or `v2`.

> The flag should accept both upper and lower case input, though.

I think the flag should accept only lower case for now. The database engine flag currently accepts only "MySQL" or "PostgreSQL", not "mysql" nor "postgresql" and `serverless-version` flag should do the same.

Lou1415926 (commented):
Thank you @hkford so much for adding this ❤️ !!

I wonder if we should just always generate the template for v2. That is, when people run
```
copilot storage init --storage-type Aurora
```
we will generate a template that creates the aurora serverless v2 cluster.

hkford:
@Lou1415926 
I removed the prompt to ask Aurora Serverless version and treat the default version as v2.
Unless specifying `--serverless-version v1` Copilot generates a v1 cluster template.

```shell
  Create an RDS Aurora Serverless v2 cluster using PostgreSQL as the database engine.
  `$ copilot storage init -n my-cluster -t Aurora -w frontend --engine PostgreSQL --initial-db testdb`
  Create an RDS Aurora Serverless v1 cluster using MySQL as the database engine.
  `$ copilot storage init -n my-cluster -t Aurora --serverless-version v1 -w frontend --engine MySQL --initial-db testdb`
```

internal/pkg/addon/storage.go
```diff
@@ -15,11 +15,19 @@ import (
...
+	rdsV2TemplatePath     = "addons/aurora/serverlessv2.yml"
+	rdsRDWSTemplatePath   = "addons/aurora/rdws/cf.yml"
+	rdsRDWSV2TemplatePath = "addons/aurora/rdws/serverlessv2.yml"
+	rdsRDWSParamsPath     = "addons/aurora/rdws/addons.parameters.yml"
+)
+
+const (
+	// Aurora Serverless versions.
+	ServerlessVersionV1 = "v1"
+	ServerlessVersionV2 = "v2"
```
efekarakus: nit: do these need to be public?
```suggestion
	auroraServerlessVersionV1 = "v1"
	auraraServerlessVersionV2 = "v2"
```

internal/pkg/addon/storage.go
```diff
@@ -147,12 +159,13 @@ func NewDDBTemplate(input *DynamoDBProps) *DynamoDBTemplate {
...
 type RDSProps struct {
-	WorkloadType   string   // The type of the workload associated with the RDS addon.
-	ClusterName    string   // The name of the cluster.
-	Engine         string   // The engine type of the RDS Aurora Serverless cluster.
-	InitialDBName  string   // The name of the initial database created inside the cluster.
-	ParameterGroup string   // The parameter group to use for the cluster.
-	Envs           []string // The copilot environments found inside the current app.
+	WorkloadType      string   // The type of the workload associated with the RDS addon.
+	ClusterName       string   // The name of the cluster.
+	ServerlessVersion string   // The version of Aurora Serverless
```
efekarakus: I wonder if this is a hint that we need two types of constructors? and can replace the [existing `RDSTemplate` constructor](https://github.com/hkford/copilot-cli/blob/8a2ab8eb528dac097901ffc8ae2967772f535610/internal/pkg/addon/storage.go#L172):

```go
type RDSProps struct {
   WorkloadType
   // Other existing fields...
   Envs []string
   
   serverlessVersion string
}

func NewServerlessV1Template(rds RDSProps) *RDSTemplate {
  rds.serverlessVersion = auroraServerlessVersionV1
  // existing logic...
}

func NewServerlessV2Template(rds RDSProps) *RDSTemplate {
   rds.serverlessVersion = auroraServerlessVersionV2
}
```

internal/pkg/cli/storage_init.go
```diff
@@ -126,18 +126,27 @@ var attributeTypes = []string{
...
-	storageInitRDSDBEnginePrompt      = "Which database engine would you like to use?"
+	storageInitRDSInitialDBNamePrompt     = "What would you like to name the initial database in your cluster?"
+	storageInitRDSDBEnginePrompt          = "Which database engine would you like to use?"
 )
 
 // RDS Aurora Serverless specific constants and variables.
 const (
+	serverlessVersionV1            = "v1"
+	serverlessVersionV2            = "v2"
+	defaultServerlessVersionString = "v2"
```
efekarakus: nit:
```suggestion
	defaultServerlessVersion = serverlessVersionV2
```

internal/pkg/cli/storage_init.go
```diff
@@ -544,6 +563,16 @@ func (o *initStorageOpts) askDynamoLSIConfig() error {
 	}
 }
 
+func (o *initStorageOpts) askAuroraServerlessVersion() error {
+	if o.rdsServerlessVersion != "" {
+		return nil
+	}
+	// If the user has not specified aurora serverless version we treat aurora serverless version as v2.
+	defaultServerlessVersion := defaultServerlessVersionString
+	o.rdsServerlessVersion = defaultServerlessVersion
```
efekarakus: nit:
```suggestion
	o.rdsServerlessVersion = defaultServerlessVersion
```

internal/pkg/template/templates/addons/aurora/rdws/serverlessv2.yml
```diff
@@ -0,0 +1,183 @@
...
+  #       character_set_client: 'utf8'
+  {{- else}}
+  {{logicalIDSafe .ClusterName}}DBClusterParameterGroup:
+    Metadata:
+      'aws:copilot:description': 'A DB parameter group for engine configuration values'
+    Type: 'AWS::RDS::DBClusterParameterGroup'
+    Properties:
+      Description: !Ref 'AWS::StackName'
+      {{- if eq .Engine "MySQL"}}
+      Family: 'aurora-mysql8.0'
```
efekarakus, line 110: Just to make sure was the addon tested in an an account with `copilot deploy` for both MySQL and Postgres?

internal/pkg/template/templates/addons/aurora/rdws/serverlessv2.yml
```diff
@@ -0,0 +1,183 @@
...
+      {{- end}}
+      DBClusterParameterGroupName: {{- if .ParameterGroup}} {{.ParameterGroup}} {{- else}} !Ref {{logicalIDSafe .ClusterName}}DBClusterParameterGroup {{- end}}
+      DBSubnetGroupName: !Ref {{logicalIDSafe .ClusterName}}DBSubnetGroup
+      VpcSecurityGroupIds:
+        - !Ref {{logicalIDSafe .ClusterName}}DBClusterSecurityGroup
+      ServerlessV2ScalingConfiguration:
+        # Replace "All" below with "!Ref Env" to set different autoscaling limits per environment.
+        MinCapacity: !FindInMap [{{logicalIDSafe .ClusterName}}EnvScalingConfigurationMap, All, DBMinCapacity]
+        MaxCapacity: !FindInMap [{{logicalIDSafe .ClusterName}}EnvScalingConfigurationMap, All, DBMaxCapacity]
+  {{logicalIDSafe .ClusterName}}DBWriterInstance:
```
efekarakus, line 144: Oh very interesting, can you explain to be why an explicit writer instance is needed?
We didn't set one for aurora serverless v1 so I'm curious why v2 is different?

efekarakus (commented):
Awesome, thank you so much @hkford ! Apologies on the late review

internal/pkg/addon/storage.go
Lou1415926: nit:
```suggestion
	AuroraServerlessVersionV1 = "v1"
	AuroraServerlessVersionV2 = "v2"
```

internal/pkg/addon/storage_test.go
```diff
@@ -159,6 +164,42 @@ func TestRDSTemplate_MarshalBinary(t *testing.T) {
 			},
 			wantedBinary: []byte("mysql"),
 		},
+		"renders postgresql v2 content": {
```
Lou1415926, line 167: Yayy thank you for all the unit tests  !

internal/pkg/cli/validate.go
```diff
@@ -438,6 +439,19 @@ func validateDBNameCharacters(name string) error {
...
+		return errValueNotAString
+	}
+	for _, valid := range serverlessVersions {
+		if version == valid {
+			return nil
+		}
+	}
+	return fmt.Errorf(fmtErrInvalidServerlessVersion, version, prettify(serverlessVersions))
+}
+
```
Lou1415926: I think in this case we can move this validation function to `storage_init.go`, similar to https://github.com/aws/copilot-cli/blob/08983f7fd662621f6cb0c59964122ab65ee2c941/internal/pkg/cli/storage_init.go#L333 


```suggestion
func (o *initStorageOpts) validateServerlessVersion() error {
	for _, valid := range serverlessVersions {
		if o.serverlessVersion == valid {
			return nil
		}
	}
	return fmt.Errorf(fmtErrInvalidServerlessVersion, version, prettify(serverlessVersions))
}

```

The validation functions in this `validate.go` file accept `val interface{}` because these validation functions are passed to the prompt handlers which require the function signature to be `(interface{}) error`. 

In our case, since we no longer prompt for versions, we won't need to adhere to this signature!

internal/pkg/cli/storage_init.go
```diff
@@ -848,6 +889,7 @@ Resource names are injected into your containers as environment variables for ea
 	cmd.Flags().BoolVar(&vars.noLSI, storageNoLSIFlag, false, storageNoLSIFlagDescription)
 	cmd.Flags().BoolVar(&vars.noSort, storageNoSortFlag, false, storageNoSortFlagDescription)
 
+	cmd.Flags().StringVar(&vars.rdsServerlessVersion, storageRDSServerlessVersionFlag, "", storageRDSServerlessVersionFlagDescription)
```
Lou1415926: instead of using `askAuroraServerlessVersion` to set the default, what do you think of setting it here:
```suggestion
	cmd.Flags().StringVar(&vars.rdsServerlessVersion, storageRDSServerlessVersionFlag, defaultServerlessVersion, storageRDSServerlessVersionFlagDescription)
```

internal/pkg/cli/flag.go
```diff
@@ -278,6 +279,8 @@ Must be of the format '<keyName>:<dataType>'.`
 	storageNoLSIFlagDescription     = `Optional. Don't ask about configuring alternate sort keys.`
 	storageLSIConfigFlagDescription = `Optional. Attribute to use as an alternate sort key. May be specified up to 5 times.
 Must be of the format '<keyName>:<dataType>'.`
+	storageRDSServerlessVersionFlagDescription = `Optional. Aurora Serverless version. Must be either "v1" or "v2".
+(default v2)`
```
Lou1415926: If we do decide to set the default in `cmd.Flags().StringVar(&vars.rdsServerlessVersion, storageRDSServerlessVersionFlag, "", storageRDSServerlessVersionFlagDescription)` (in `storage_init.go`), then the `cobra` will automatically append `(default "v2")` to the end of the flag description I think!

 ```suggestion
	storageRDSServerlessVersionFlagDescription = `Optional. Aurora Serverless version. Must be either "v1" or "v2".`
```

File: internal/pkg/template/templates/addons/aurora/rdws/serverlessv2.yml [hunk 13]
```diff
@@ -0,0 +1,183 @@
...
+Mappings:
+  {{logicalIDSafe .ClusterName}}EnvScalingConfigurationMap: {{range $env := .Envs}}
+    {{$env}}:
+      {{- if eq $.Engine "MySQL"}}
+      "DBMinCapacity": 0.5 # AllowedValues: from 0.5 through 128
+      "DBMaxCapacity": 8   # AllowedValues: from 0.5 through 128
+      {{- else}}
+      "DBMinCapacity": 0.5 # AllowedValues: from 0.5 through 128
+      "DBMaxCapacity": 8   # AllowedValues: from 0.5 through 128
+      {{end -}}
```
Lou1415926: nit: seems like we don't need to differentiate based on engine anymore (previously, the min-max was different for mysql vs. postgresql)

```suggestion
      "DBMinCapacity": 0.5 # AllowedValues: from 0.5 through 128
      "DBMaxCapacity": 8   # AllowedValues: from 0.5 through 128
```

internal/pkg/template/templates/addons/aurora/rdws/serverlessv2.yml
Lou1415926: Same question! When creating `AWS::RDS::DBCluster`, does RDS automatically create a writer instance for that cluster?

Lou1415926 (commented):
Thank you very much ! This is awesome. I just have some small nits and one question (same as Efe's)!

internal/pkg/template/templates/addons/aurora/rdws/serverlessv2.yml
Lou1415926: We had an offline discussion and I'm sharing it here for visibility:

> Does RDS automatically create a writer instance for that cluster?

For v2, it doesn't. If we have only the `DBCluster` resource in the template, only a cluster is created - no writer instance.

> We didn't set one for aurora serverless v1 so I'm curious why v2 is different?

v1 contains one and only one read/write instance. When we create a `DBCluster`, that one and only one read/write instance also gets automatically created. This is different from v2 and is also the reason why we didn't need to explicitly create an instance before.

internal/pkg/addon/storage.go
hkford: `addon.AuroraServerlessVersion` is used in `newRDSTemplate()` of `cli/storage_init.go`. This function calls `addon.NewServerlessVxTemplate` using switch statement, same as `addon.RDSEngineTypeMySQL` or `addon.RDSEngineTypePostgreSQL`.

internal/pkg/addon/storage.go
hkford: I agree with you. I made two types of constructors.

internal/pkg/cli/storage_init.go
hkford: Thanks. I fixed it.

internal/pkg/cli/storage_init.go
hkford: Fixed it.

internal/pkg/addon/storage.go
hkford: Thanks. I fixed it.

internal/pkg/cli/validate.go
hkford: Thank you for your suggestion. I moved `validateServerlessVersion` to `storage_init.go`.

internal/pkg/cli/storage_init.go
hkford: I fixed it.

internal/pkg/cli/flag.go
hkford: I set the default value in `cmd.Flags().StringVar(&vars.rdsServerlessVersion, storageRDSServerlessVersionFlag, defaultServerlessVersion, storageRDSServerlessVersionFlagDescription)` and `cobra` appended `(default "v2")` as expected.

internal/pkg/template/templates/addons/aurora/rdws/serverlessv2.yml
hkford: I fixed it.

hkford:
@efekarakus @Lou1415926 
I updated my PR based on your comments. Would you review my code again please?

internal/pkg/addon/storage.go
```diff
@@ -15,11 +15,19 @@ import (
...
+	rdsV2TemplatePath     = "addons/aurora/serverlessv2.yml"
+	rdsRDWSTemplatePath   = "addons/aurora/rdws/cf.yml"
+	rdsRDWSV2TemplatePath = "addons/aurora/rdws/serverlessv2.yml"
+	rdsRDWSParamsPath     = "addons/aurora/rdws/addons.parameters.yml"
+)
+
+const (
+	// Aurora Serverless versions.
+	AuroraServerlessVersionV1 = "v1"
+	AuroraServerlessVersionV2 = "v2"
```
efekarakus: nit: can these be private?

internal/pkg/addon/storage.go
```diff
@@ -147,16 +159,28 @@ func NewDDBTemplate(input *DynamoDBProps) *DynamoDBTemplate {
...
 type RDSProps struct {
-	WorkloadType   string   // The type of the workload associated with the RDS addon.
-	ClusterName    string   // The name of the cluster.
-	Engine         string   // The engine type of the RDS Aurora Serverless cluster.
-	InitialDBName  string   // The name of the initial database created inside the cluster.
-	ParameterGroup string   // The parameter group to use for the cluster.
-	Envs           []string // The copilot environments found inside the current app.
+	WorkloadType            string   // The type of the workload associated with the RDS addon.
+	ClusterName             string   // The name of the cluster.
+	AuroraServerlessVersion string   // The version of Aurora Serverless
```
efekarakus: nit: can be private, and can we add [a period](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences) at the end of the comment plz 🥺 
```suggestion
	auroraServerlessVersion string   // The version of Aurora Serverless.
```

internal/pkg/cli/storage_init.go
```diff
@@ -737,14 +763,33 @@ func (o *initStorageOpts) newRDSTemplate() (*addon.RDSTemplate, error) {
...
+			AuroraServerlessVersion: version,
+			Engine:                  engine,
+			InitialDBName:           o.rdsInitialDBName,
+			ParameterGroup:          o.rdsParameterGroup,
+			Envs:                    envs,
+			WorkloadType:            o.workloadType,
+		}), nil
+	default:
+		return nil, errors.New("unknown serverless version")
+	}
```
efekarakus: To reduce duplication:
```go
	props := addon.RDSProps{
		ClusterName:             o.storageName,
		Engine:                  engine,
		InitialDBName:           o.rdsInitialDBName,
		ParameterGroup:          o.rdsParameterGroup,
		Envs:                    envs,
		WorkloadType:            o.workloadType,
	}
	switch v:= o.auroraServerlessVersion; v {
	case auroraServerlessVersionV1:
		return addon.NewServerlessV1Template(props), nil
	case auroraServerlessVersionV2:
		return addon.NewServerlessV2Template(props), nil
	default:
		return nil, fmt.Errorf("unknown Aurora serverless version %q", v)
	}
```
We shouldn't need to set the `AuroraServerlessVersion` as that's done by the constructor in the `addon` pkg

efekarakus (commented):
Looks great, just some tiny comments and then we can :shipit: ! thank you @hkford

internal/pkg/cli/storage_init_test.go
```diff
@@ -173,6 +174,30 @@ func TestStorageInitOpts_Validate(t *testing.T) {
 
 			wantedErr: errors.New("invalid engine type mysql: must be one of \"MySQL\", \"PostgreSQL\""),
 		},
+		"successfully validates aurora serverless version v1": {
+			mockWs:              func(m *mocks.MockwsAddonManager) {},
+			mockStore:           func(m *mocks.Mockstore) {},
+			inAppName:           "bowie",
+			inStorageType:       rdsStorageType,
+			inServerlessVersion: auroraServerlessVersionV1,
+			wantedErr:           nil,
```
Lou1415926: nit: since `nil` is the zero value for `wantedErr`, it would by default be `nil` -
```suggestion
```

internal/pkg/cli/storage_init_test.go
```diff
@@ -173,6 +174,30 @@ func TestStorageInitOpts_Validate(t *testing.T) {
...
+		},
+		"successfully validates aurora serverless version v2": {
+			mockWs:              func(m *mocks.MockwsAddonManager) {},
+			mockStore:           func(m *mocks.Mockstore) {},
+			inAppName:           "bowie",
+			inStorageType:       rdsStorageType,
+			inServerlessVersion: auroraServerlessVersionV2,
+			wantedErr:           nil,
```
Lou1415926: samesies!
```suggestion
```

internal/pkg/cli/storage_init_test.go
```diff
@@ -173,6 +174,30 @@ func TestStorageInitOpts_Validate(t *testing.T) {
...
+		},
+		"invalid aurora serverless version": {
+			mockWs:              func(m *mocks.MockwsAddonManager) {},
+			mockStore:           func(m *mocks.Mockstore) {},
+			inAppName:           "bowie",
+			inStorageType:       rdsStorageType,
+			inServerlessVersion: "weird-serverless-version",
```
Lou1415926, line 196: weird version!

Lou1415926 (approved):
@hkford Thank you very much! This looks great. I don't have much to add in addition to what Efe said. I will approve and add the do-not-merge label for you to address Efe's feedback!

internal/pkg/addon/storage.go
hkford: changed into private field and added a period.

internal/pkg/cli/storage_init.go
hkford: Yes I noticed. I reduced duplication.

internal/pkg/cli/storage_init_test.go
hkford: I removed `wantedErr: nil` in two test cases and left other test cases alone. Should I remove other `wantedErr: nil` or do so in another PR?

efekarakus (approved):

PR #4076, opened 2022-10-10
chore: bump env template patch version

dannyrandall:


efekarakus (approved):

huanjani:
Do we need to update integ tests, too?

dannyrandall:
i think they passed... do the e2e tests depend on the version?

dannyrandall:
as for why they pass - we pass a fixed env version in so it's always a fixed value in the generated cf template https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/deploy/cloudformation/stack/lb_web_service_integration_test.go#L105

huanjani (approved):

PR #4077, opened 2022-10-10
chore: add sc to svc show

iamhopaul123:
### Sample Output
```
Service Endpoint

  Endpoint                              Environment  Type
  --------                              -----------  ----
  my-svc                                test         Service Connect
  api                                   prod         Service Connect
  http://my-svc.test.my-app.local:5000  test         Service Discovery
  http://my-svc.prod.my-app.local:5000  prod         Service Discovery
```
Note that service discovery section for human output is now gone.

### DIFF
--- a/internal/pkg/aws/ecs/service.go
+++ b/internal/pkg/aws/ecs/service.go
@@ -80,14 +80,13 @@ func (s *Service) ServiceConnectAliases() []string {
 	if scConfig == nil || !aws.BoolValue(scConfig.Enabled) {
 		return nil
 	}
-	ns := aws.StringValue(scConfig.Namespace)
 	var aliases []string
 	for _, service := range scConfig.Services {
 		defaultName := aws.StringValue(service.PortName)
 		if aws.StringValue(service.DiscoveryName) != "" {
 			defaultName = aws.StringValue(service.DiscoveryName)
 		}
-		defaultAlias := fmt.Sprintf("%s.%s", defaultName, ns)
+		defaultAlias := fmt.Sprintf("%s.%s", defaultName, aws.StringValue(scConfig.Namespace))
 		if len(service.ClientAliases) == 0 {
 			aliases = append(aliases, defaultAlias)
 			continue
--- a/internal/pkg/aws/ecs/service.go
+++ b/internal/pkg/aws/ecs/service.go
@@ -70,6 +70,44 @@ func (s *Service) ServiceStatus() ServiceStatus {
 	}
 }
 
+// ServiceConnectAliases returns the ECS Service Connect client aliases for a service.
+func (s *Service) ServiceConnectAliases() []string {
+	if len(s.Deployments) == 0 {
+		return nil
+	}
+	lastDeployment := s.Deployments[0]
+	scConfig := lastDeployment.ServiceConnectConfiguration
+	if scConfig == nil || !aws.BoolValue(scConfig.Enabled) {
+		return nil
+	}
+	ns := aws.StringValue(scConfig.Namespace)
+	var aliases []string
+	for _, service := range scConfig.Services {
+		defaultName := aws.StringValue(service.PortName)
+		if aws.StringValue(service.DiscoveryName) != "" {
+			defaultName = aws.StringValue(service.DiscoveryName)
+		}
+		defaultAlias := fmt.Sprintf("%s.%s", defaultName, ns)
+		if len(service.ClientAliases) == 0 {
+			aliases = append(aliases, defaultAlias)
+			continue
+		}
+		for _, clientAlias := range service.ClientAliases {
+			alias := defaultAlias
+			if aws.StringValue(clientAlias.DnsName) != "" {
+				alias = aws.StringValue(clientAlias.DnsName)
+			}
+			aliases = append(aliases, fmt.Sprintf("%s:%v", alias, aws.Int64Value(clientAlias.Port)))
+		}
+	}
+	return aliases
+}
+
+// LastUpdatedAt returns the last updated time of the ECS service.
+func (s *Service) LastUpdatedAt() time.Time {
+	return aws.TimeValue(s.Deployments[0].UpdatedAt)
+}
+
 // TargetGroups returns the ARNs of target groups attached to the service.
 func (s *Service) TargetGroups() []string {
 	var targetGroupARNs []string
--- a/internal/pkg/describe/backend_service.go
+++ b/internal/pkg/describe/backend_service.go
@@ -104,7 +104,8 @@ func (d *BackendServiceDescriber) Describe() (HumanJSONStringer, error) {
 
 	var routes []*WebServiceRoute
 	var configs []*ECSServiceConfig
-	var services []*ServiceDiscovery
+	var sds serviceDiscoveries
+	var scs serviceConnects
 	var envVars []*containerEnvVar
 	var secrets []*secret
 	for _, env := range environments {
@@ -137,11 +138,18 @@ func (d *BackendServiceDescriber) Describe() (HumanJSONStringer, error) {
 				return nil, err
 			}
 			port = svcParams[cfnstack.WorkloadContainerPortParamKey]
-			services = appendServiceDiscovery(services, serviceDiscovery{
+			sds = appendServiceDiscovery(sds, serviceDiscovery{
 				Service:  d.svc,
 				Port:     port,
 				Endpoint: endpoint,
 			}, env)
+			scDNSNames, err := svcDescr.ServiceConnectDNSNames()
+			if err != nil {
+				return nil, fmt.Errorf("retrieve service connect DNS names: %w", err)
+			}
+			for _, dnsName := range scDNSNames {
+				scs = appendServiceConnect(scs, dnsName, env)
+			}
 		}
 		containerPlatform, err := svcDescr.Platform()
 		if err != nil {
@@ -185,17 +193,20 @@ func (d *BackendServiceDescriber) Describe() (HumanJSONStringer, error) {
 	}
 
 	return &backendSvcDesc{
-		Service:          d.svc,
-		Type:             manifest.BackendServiceType,
-		App:              d.app,
-		Configurations:   configs,
-		Routes:           routes,
-		ServiceDiscovery: services,
-		Variables:        envVars,
-		Secrets:          secrets,
-		Resources:        resources,
+		ecsSvcDesc: ecsSvcDesc{
+			Service:          d.svc,
+			Type:             manifest.BackendServiceType,
+			App:              d.app,
+			Configurations:   configs,
+			Routes:           routes,
+			ServiceDiscovery: sds,
+			ServiceConnect:   scs,
+			Variables:        envVars,
+			Secrets:          secrets,
+			Resources:        resources,
 
-		environments: environments,
+			environments: environments,
+		},
 	}, nil
 }
 
@@ -211,12 +222,17 @@ func (d *BackendServiceDescriber) Manifest(env string) ([]byte, error) {
 
 // backendSvcDesc contains serialized parameters for a backend service.
 type backendSvcDesc struct {
+	ecsSvcDesc
+}
+
+type ecsSvcDesc struct {
 	Service          string               `json:"service"`
 	Type             string               `json:"type"`
 	App              string               `json:"application"`
 	Configurations   ecsConfigurations    `json:"configurations"`
 	Routes           []*WebServiceRoute   `json:"routes"`
 	ServiceDiscovery serviceDiscoveries   `json:"serviceDiscovery"`
+	ServiceConnect   serviceConnects      `json:"serviceConnect,omitempty"`
 	Variables        containerEnvVars     `json:"variables"`
 	Secrets          secrets              `json:"secrets,omitempty"`
 	Resources        deployedSvcResources `json:"resources,omitempty"`
@@ -255,9 +271,15 @@ func (w *backendSvcDesc) HumanString() string {
 			fmt.Fprintf(writer, "  %s\t%s\n", route.Environment, route.URL)
 		}
 	}
-	fmt.Fprint(writer, color.Bold.Sprint("\nService Discovery\n\n"))
-	writer.Flush()
-	w.ServiceDiscovery.humanString(writer)
+	if len(w.ServiceConnect) > 0 || len(w.ServiceDiscovery) > 0 {
+		fmt.Fprint(writer, color.Bold.Sprint("\nService Endpoint\n\n"))
+		writer.Flush()
+		endpoints := serviceEndpoints{
+			discoveries: w.ServiceDiscovery,
+			connects:    w.ServiceConnect,
+		}
+		endpoints.humanString(writer)
+	}
 	fmt.Fprint(writer, color.Bold.Sprint("\nVariables\n\n"))
 	writer.Flush()
 	w.Variables.humanString(writer)
--- a/internal/pkg/describe/backend_service.go
+++ b/internal/pkg/describe/backend_service.go
@@ -104,8 +104,8 @@ func (d *BackendServiceDescriber) Describe() (HumanJSONStringer, error) {
 
 	var routes []*WebServiceRoute
 	var configs []*ECSServiceConfig
-	var sdEndpoints serviceDiscoveries
-	var scEndpoints serviceConnects
+	sdEndpoints := make(serviceDiscoveries)
+	scEndpoints := make(serviceConnects)
 	var envVars []*containerEnvVar
 	var secrets []*secret
 	for _, env := range environments {
@@ -213,21 +213,6 @@ func (d *BackendServiceDescriber) Manifest(env string) ([]byte, error) {
 // backendSvcDesc contains serialized parameters for a backend service.
 type backendSvcDesc struct {
 	ecsSvcDesc
-}
-
-type ecsSvcDesc struct {
-	Service          string               `json:"service"`
-	Type             string               `json:"type"`
-	App              string               `json:"application"`
-	Configurations   ecsConfigurations    `json:"configurations"`
-	Routes           []*WebServiceRoute   `json:"routes"`
-	ServiceDiscovery serviceDiscoveries   `json:"serviceDiscovery"`
-	ServiceConnect   serviceConnects      `json:"serviceConnect,omitempty"`
-	Variables        containerEnvVars     `json:"variables"`
-	Secrets          secrets              `json:"secrets,omitempty"`
-	Resources        deployedSvcResources `json:"resources,omitempty"`
-
-	environments []string `json:"-"`
 }
 
 // JSONString returns the stringified backendService struct with json format.
--- a/internal/pkg/describe/backend_service.go
+++ b/internal/pkg/describe/backend_service.go
@@ -134,10 +134,10 @@ func (d *BackendServiceDescriber) Describe() (HumanJSONStringer, error) {
 		port := blankContainerPort
 		if isReachableWithinVPC(svcParams) {
 			port = svcParams[cfnstack.WorkloadTargetPortParamKey]
-			if err := sdEndpoints.collectServiceDiscoveryEndpoints(envDescr, d.svc, env, port); err != nil {
+			if err := sdEndpoints.collectEndpoints(envDescr, d.svc, env, port); err != nil {
 				return nil, err
 			}
-			if err := scEndpoints.collectServiceConnectEndpoints(svcDescr, env); err != nil {
+			if err := scEndpoints.collectEndpoints(svcDescr, env); err != nil {
 				return nil, err
 			}
 		}
--- a/internal/pkg/describe/backend_service.go
+++ b/internal/pkg/describe/backend_service.go
@@ -262,7 +262,7 @@ func (w *backendSvcDesc) HumanString() string {
 		}
 	}
 	if len(w.ServiceConnect) > 0 || len(w.ServiceDiscovery) > 0 {
-		fmt.Fprint(writer, color.Bold.Sprint("\nService Endpoint\n\n"))
+		fmt.Fprint(writer, color.Bold.Sprint("\nInternal Service Endpoint\n\n"))
 		writer.Flush()
 		endpoints := serviceEndpoints{
 			discoveries: w.ServiceDiscovery,
--- a/internal/pkg/describe/backend_service.go
+++ b/internal/pkg/describe/backend_service.go
@@ -104,8 +104,8 @@ func (d *BackendServiceDescriber) Describe() (HumanJSONStringer, error) {
 
 	var routes []*WebServiceRoute
 	var configs []*ECSServiceConfig
-	var sds serviceDiscoveries
-	var scs serviceConnects
+	var sdEndpoints serviceDiscoveries
+	var scEndpoints serviceConnects
 	var envVars []*containerEnvVar
 	var secrets []*secret
 	for _, env := range environments {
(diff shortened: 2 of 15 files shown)


internal/pkg/aws/ecs/service.go
```diff
@@ -70,6 +70,44 @@ func (s *Service) ServiceStatus() ServiceStatus {
...
+	}
+	lastDeployment := s.Deployments[0]
+	scConfig := lastDeployment.ServiceConnectConfiguration
+	if scConfig == nil || !aws.BoolValue(scConfig.Enabled) {
+		return nil
+	}
+	ns := aws.StringValue(scConfig.Namespace)
+	var aliases []string
+	for _, service := range scConfig.Services {
+		defaultName := aws.StringValue(service.PortName)
```
efekarakus, line 85: Why are we setting the default to be the port name 🤔 isn't a `DiscoveryName` always going to be present?

internal/pkg/aws/ecs/service.go
```diff
@@ -70,6 +70,44 @@ func (s *Service) ServiceStatus() ServiceStatus {
...
+			defaultName = aws.StringValue(service.DiscoveryName)
+		}
+		defaultAlias := fmt.Sprintf("%s.%s", defaultName, ns)
+		if len(service.ClientAliases) == 0 {
+			aliases = append(aliases, defaultAlias)
+			continue
+		}
+		for _, clientAlias := range service.ClientAliases {
+			alias := defaultAlias
+			if aws.StringValue(clientAlias.DnsName) != "" {
```
efekarakus, line 96: I see so it's possible that users specify alternative ports but no custom dns name lol

internal/pkg/describe/backend_service.go
```diff
@@ -104,7 +104,8 @@ func (d *BackendServiceDescriber) Describe() (HumanJSONStringer, error) {
 
 	var routes []*WebServiceRoute
 	var configs []*ECSServiceConfig
-	var services []*ServiceDiscovery
+	var sds serviceDiscoveries
+	var scs serviceConnects
```
efekarakus: nit:
```suggestion
	var sdEndpoints serviceDiscoveries
	var scEndpoints serviceConnects
```

internal/pkg/describe/backend_service.go
```diff
@@ -137,11 +138,18 @@ func (d *BackendServiceDescriber) Describe() (HumanJSONStringer, error) {
 				return nil, err
 			}
 			port = svcParams[cfnstack.WorkloadContainerPortParamKey]
-			services = appendServiceDiscovery(services, serviceDiscovery{
+			sds = appendServiceDiscovery(sds, serviceDiscovery{
```
efekarakus: I wonder if we can make this method a bit more readable by adding some smaller readable methods:
```go
if d.isReachableWithinVPC(params) {
   sdEndpoints, err = collectServiceDiscoveryEndpoints(envDescr, d.svc, port)
   if err != nil { ... }
   scEndpoints, err = collectServiceConnectEndpoints(svcDescr)
   if err != nil { ... }
}
```

efekarakus: I wonder if `svc show` works correctly with service discovery now with @dannyrandall target container fix?
like I wonder if we should update these functions to use `WorkloadTargetPortParamKey` instead of `WorkloadContainerPortParamKey`

internal/pkg/describe/backend_service.go
```diff
@@ -255,9 +271,15 @@ func (w *backendSvcDesc) HumanString() string {
 			fmt.Fprintf(writer, "  %s\t%s\n", route.Environment, route.URL)
 		}
 	}
-	fmt.Fprint(writer, color.Bold.Sprint("\nService Discovery\n\n"))
-	writer.Flush()
-	w.ServiceDiscovery.humanString(writer)
+	if len(w.ServiceConnect) > 0 || len(w.ServiceDiscovery) > 0 {
+		fmt.Fprint(writer, color.Bold.Sprint("\nService Endpoint\n\n"))
```
efekarakus:
```suggestion
		fmt.Fprint(writer, color.Bold.Sprint("\nInternal Service Endpoints\n\n"))
```

internal/pkg/describe/lb_web_service.go
```diff
@@ -289,36 +300,79 @@ type WebServiceRoute struct {
 	URL         string `json:"url"`
 }
 
-// ServiceDiscovery contains serialized service discovery info for an service.
+// ServiceDiscovery contains serialized service discovery info for an ECS service.
 type ServiceDiscovery struct {
 	Environment []string `json:"environment"`
-	Namespace   string   `json:"namespace"`
+	// Namespace is not an accurate name but unfortunately we can't take it back.
+	Namespace string `json:"namespace"`
```
efekarakus: I am honestly so confused on what we are doing here 😝 
1. Does this mean this "namespace" is applicable for  all the `.Environments`?

I guess this structure exists because of our old service discovery setup with `{app}.local` 😭 .

I feel like our life would be so much simpler if we have:
```go
// A generic endpoint data type that can be used for both SC and SD.
type InternalEndpoint struct {
   Endpoint string 
   Environment string
}

// sdMarshaler is a collection of Service Discovery endpoints.
type sdMarshaler []*InternalEndpoint

// scMarshaler is a collection of Service Connect endpoints.
type scMarshaler []*InternalEndpoint


// MarshalJSON writes a collection of service discovery endpoints in JSON format.
func (sdMarshaler) MarshalJSON() ([]byte, error) {
    // for backwards compatibility define the old type here.
    type out struct {
        Namespace string
        Environments []string
    }
    //...
}
func (scMarshlaer) MarshalJSON() ([]byte, error)

// WriteHumanString writes a collection of service discovery endpoints grouped by environments.
func (sdMarshaler) WriteHumanString(w io.Writer) error { ... }
func (scMarshaler) WriteHumanString(w io.Writer) error { ... }

// collectEndpointsFor returns the collection of endpoints for a given environment.
func (sdMarshaler) collectEndpointsFor(env string) []string { ... }
```

internal/pkg/describe/lb_web_service.go
```diff
@@ -289,36 +300,79 @@ type WebServiceRoute struct {
...
+	connects    serviceConnects
+}
+
+func (s serviceEndpoints) humanString(w io.Writer) {
+	headers := []string{"Environment", "Endpoint", "Type"}
 	fmt.Fprintf(w, "  %s\n", strings.Join(headers, "\t"))
 	fmt.Fprintf(w, "  %s\n", strings.Join(underline(headers), "\t"))
-	for _, sd := range s {
-		fmt.Fprintf(w, "  %s\t%s\n", strings.Join(sd.Environment, ", "), sd.Namespace)
+	scEnvToEndpoints := make(map[string]map[string]struct{})
```
efekarakus: I feel like our data structures are pretty complicated 💭  something is wrong with how we're modeling data, and we have to duplicate the same marshaling logic for both SC and SD.

Let me know what you think of the above comment for an alternative API.

internal/pkg/describe/lb_web_service.go
```diff
@@ -411,3 +471,20 @@ func appendServiceDiscovery(sds []*ServiceDiscovery, sd serviceDiscovery, env st
 	}
 	return sds
 }
+
+func appendServiceConnect(scs serviceConnects, dnsName string, env string) serviceConnects {
+	exist := false
+	for _, s := range scs {
+		if s.DNSName == dnsName {
```
efekarakus: The short variable names make it a bit difficult to figure out what is being represented here due to all the acronyms.
Can we expand on our variable names?

efekarakus (commented):
I haven't reviewed the entire PR, take a look at the recommendations for the `describe` package and let  me know what you think 🥺 

I think we should follow Kent Becks' recommendation here:
> "First make the change easy, then make the easy change"
> https://www.adamtal.me/2019/05/first-make-the-change-easy-then-make-the-easy-change

internal/pkg/aws/ecs/service.go
iamhopaul123: Port name is the default name and the only required field, regardless of how Copilot set it. DiscoveryName is optional.

internal/pkg/aws/ecs/service.go
iamhopaul123: yeah. port is the only required field in each `ClientAliases`.

internal/pkg/describe/lb_web_service.go
iamhopaul123: Namespace is wrong...it is a mistake we made before and we can't take it back because otherwise the json output will change. Namespace ideally should be only `env.app.local` whereas here the namespace is actually the full endpoint name (at least this is what it is now)

iamhopaul123: TBH I'd rather switch to
```golang
type InternalEndpoint struct {
   Endpoint string 
   Environment string
}
```
for both sc and sd. change `Namespace` to `Endpoint` which makes more sense and we can argue bug fixing is not breaking change.

internal/pkg/describe/lb_web_service.go
iamhopaul123: this is just a map of string and set...let me try if i can get rid of it after adopting the new apis

internal/pkg/describe/backend_service.go
iamhopaul123: <strike>It actually doesn't really matter...i wonder if we should ditch the port. @dannyrandall can you weigh in on this?</strike>

I'll just update to the targetport to be safe.

dannyrandall: You're right, it doesn't really affect port/service reachability (any exposed ports are reachable using the SD url). I think we should probably use `WorkloadTargetPortParamKey` so that SD creates a SRV record, just because we have been creating an SRV record in the past.

internal/pkg/ecs/ecs.go
```diff
@@ -112,17 +112,26 @@ func (c Client) DescribeService(app, env, svc string) (*ServiceDesc, error) {
...
+	service, err := c.ecsClient.Service(clusterName, serviceName)
 	if err != nil {
-		return time.Time{}, fmt.Errorf("get ECS service %s: %w", serviceName, err)
+		return nil, fmt.Errorf("get ECS service %s: %w", serviceName, err)
+	}
+	return service, nil
+}
+
+// LastUpdatedAt returns the last updated time of the ECS service.
+func (c Client) LastUpdatedAt(app, env, svc string) (time.Time, error) {
```
dannyrandall, line 129: i wonder if we can just remove this function and replace calls to `LastUpdatedAt()` with
```
svc, err := client.Service()
// err
svc.LastUpdatedAt()
```
just bc `(*ecs.Service).LastUpdatedAt()` doesn't return an error anyways, so we're not really shortening code we call this 🤔

internal/pkg/describe/backend_service.go
```diff
@@ -131,17 +132,14 @@ func (d *BackendServiceDescriber) Describe() (HumanJSONStringer, error) {
 			return nil, err
 		}
 		port := blankContainerPort
-		if svcParams[cfnstack.WorkloadContainerPortParamKey] != cfnstack.NoExposedContainerPort {
-			endpoint, err := envDescr.ServiceDiscoveryEndpoint()
-			if err != nil {
+		if isReachableWithinVPC(svcParams) {
+			port = svcParams[cfnstack.WorkloadTargetPortParamKey]
```
efekarakus, line 136: nice! this flow is so much easier to read :)

internal/pkg/describe/lb_web_service.go
```diff
@@ -289,36 +291,80 @@ type WebServiceRoute struct {
...
 	Environment []string `json:"environment"`
-	Namespace   string   `json:"namespace"`
+	Endpoint    string   `json:"endpoint"`
+}
+
+type internalEndpoints []*InternalEndpoint
+
+type serviceDiscoveries internalEndpoints
+
+func (sds *serviceDiscoveries) collectServiceDiscoveryEndpoints(descr envDescriber, svc, env, port string) error {
```
efekarakus: nit: maybe just `collectEndpoints` for both methods since the type already has `serviceDiscoveries` in it

internal/pkg/describe/lb_web_service.go
```diff
@@ -289,36 +291,80 @@ type WebServiceRoute struct {
 	URL         string `json:"url"`
 }
 
-// ServiceDiscovery contains serialized service discovery info for an service.
-type ServiceDiscovery struct {
+// InternalEndpoint contains serialized internal endpoint (for both service connect and service discovery) info for an ECS service.
+type InternalEndpoint struct {
```
efekarakus: nit: does the type need to be exported?

internal/pkg/describe/lb_web_service.go
dannyrandall: what about just `func collectServiceDiscoveryEndpoints() ([]*InternalEndpoint, error)`? same for sc. i'm just not clear on the benefit of having each of them be their own types 🤔

efekarakus (approved):
Nice! looks so much better than before, thanks boss!
Some tiny feedback

internal/pkg/describe/lb_web_service.go
```diff
@@ -289,36 +291,80 @@ type WebServiceRoute struct {
...
+	return nil
 }
 
-type serviceDiscoveries []*ServiceDiscovery
+func (ies *internalEndpoints) appendInternalEndpoints(endpoint string, env string) {
+	exist := false
+	for _, ie := range *ies {
+		if ie.Endpoint == endpoint {
+			ie.Environment = append(ie.Environment, env)
+			exist = true
```
dannyrandall: just return here, remove `exist`

internal/pkg/aws/ecs/service.go
```diff
@@ -70,6 +70,44 @@ func (s *Service) ServiceStatus() ServiceStatus {
...
+		return nil
+	}
+	ns := aws.StringValue(scConfig.Namespace)
+	var aliases []string
+	for _, service := range scConfig.Services {
+		defaultName := aws.StringValue(service.PortName)
+		if aws.StringValue(service.DiscoveryName) != "" {
+			defaultName = aws.StringValue(service.DiscoveryName)
+		}
+		defaultAlias := fmt.Sprintf("%s.%s", defaultName, ns)
```
dannyrandall: nit
```suggestion
		defaultAlias := fmt.Sprintf("%s.%s", defaultName, aws.StringValue(scConfig.Namespace)
```
(or even just `scConfig.Namespace`, let go do the deref)

internal/pkg/ecs/ecs.go
iamhopaul123: It's probably fine because we have some logic here and it can keep the same signature as RDWS too.

internal/pkg/describe/lb_web_service.go
iamhopaul123: Let me modify the method name and i think it'll make more sense.

internal/pkg/aws/ecs/service.go
iamhopaul123: good catch!

internal/pkg/describe/lb_web_service.go
dannyrandall: i'm still not clear on why we should have separate types for `serviceDiscoveries` and `serviceConnects`....

internal/pkg/describe/lb_web_service.go
```diff
@@ -289,36 +291,76 @@ type WebServiceRoute struct {
...
+	sd := serviceDiscovery{
+		Service:  svc,
+		Port:     port,
+		Endpoint: endpoint,
+	}
+	(*internalEndpoints)(sds).appendInternalEndpoints(sd.String(), env)
+	return nil
+}
+
+func (ies *internalEndpoints) appendInternalEndpoints(endpoint string, env string) {
```
dannyrandall: if it's already on the type, maybe just `append()`?

internal/pkg/describe/lb_web_service.go
```diff
@@ -289,36 +291,76 @@ type WebServiceRoute struct {
...
-// ServiceDiscovery contains serialized service discovery info for an service.
-type ServiceDiscovery struct {
+type internalEndpoint struct {
 	Environment []string `json:"environment"`
-	Namespace   string   `json:"namespace"`
+	Endpoint    string   `json:"endpoint"`
 }
 
-type serviceDiscoveries []*ServiceDiscovery
+type internalEndpoints []*internalEndpoint
```
dannyrandall: i think this could be `[]internalEndpoint` if L318 was changed to:
```
for i := range ies {
  if ies[i].Endpoint == endpoint {
    ies[i].Environment = append(ies[i].Environment, env)
  }
}
```

internal/pkg/describe/lb_web_service.go
```diff
@@ -289,36 +291,76 @@ type WebServiceRoute struct {
...
+		if ie.Endpoint == endpoint {
+			ie.Environment = append(ie.Environment, env)
+			return
+		}
+	}
+	*ies = append(*ies, &internalEndpoint{
+		Environment: []string{env},
+		Endpoint:    endpoint,
+	})
+	return
```
dannyrandall: remove :)

internal/pkg/describe/lb_web_service.go
iamhopaul123: Because we have multiple env so we actually keep adding things to `sds *serviceDiscoveries`. using `collectServiceDiscoveryEndpoints() ([]*InternalEndpoint, error)` will create extra space every time. if we use `collectServiceDiscoveryEndpoints(*[]*InternalEndpoint) (error)` its kinda weird.

iamhopaul123: One alternative is `collectServiceDiscoveryEndpoints(sds *serviceDiscoveries, descr envDescriber, svc, env, port string) error` but its essentially the same.

dannyrandall: oh i see. it basically takes care of the `append` from extra environments, instead of
```
var sdEndpoints []internalEndpoints
endpoints := getSDEndpoints()
append(sdEndpoints, endpoints...)
```

that makes sense to me! thanks 😊

dannyrandall (approved):
great work! 🎉

internal/pkg/describe/lb_web_service.go
iamhopaul123: Discussed offline and it ended up if we use `[]internalEndpoint` then it'll be `(*ies)[i]` to refer to struct. Using `[]* internalEndpoint` is probably more readable here.

internal/pkg/aws/ecs/service.go
```diff
@@ -70,6 +70,43 @@ func (s *Service) ServiceStatus() ServiceStatus {
...
+	scConfig := lastDeployment.ServiceConnectConfiguration
+	if scConfig == nil || !aws.BoolValue(scConfig.Enabled) {
+		return nil
+	}
+	var aliases []string
+	for _, service := range scConfig.Services {
+		defaultName := aws.StringValue(service.PortName)
+		if aws.StringValue(service.DiscoveryName) != "" {
+			defaultName = aws.StringValue(service.DiscoveryName)
+		}
```
Lou1415926, line 88: i see - this is so that the `aws` package is agnostic of how we write our CFN template. 

Seems like `if aws.StringValue(service.DiscoveryName) != ""` will always be true because we always write this field: https://github.com/aws/copilot-cli/blob/0d0436a7f074d9c3715bad37e667a71c722a1c64/internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml#L49

Do you think we want a comment about this?

internal/pkg/describe/uri.go
```diff
@@ -192,8 +192,8 @@ func (d *BackendServiceDescriber) URI(envName string) (URI, error) {
 	if err != nil {
 		return URI{}, fmt.Errorf("get stack parameters for environment %s: %w", envName, err)
 	}
-	port := svcStackParams[stack.WorkloadContainerPortParamKey]
-	if port == stack.NoExposedContainerPort {
+	port := svcStackParams[stack.WorkloadTargetPortParamKey]
```
Lou1415926: nit: since `port` is only used in L208, it could probably be refactored as `Port: svcStackParams[stack.WorkloadTargetPortParamKey]`?

internal/pkg/describe/lb_web_service.go
```diff
@@ -289,36 +291,75 @@ type WebServiceRoute struct {
...
 }
 
-type serviceDiscoveries []*ServiceDiscovery
+type internalEndpoints []*internalEndpoint
+
+type serviceDiscoveries internalEndpoints
 
-func (s serviceDiscoveries) humanString(w io.Writer) {
-	headers := []string{"Environment", "Namespace"}
+func (sds *serviceDiscoveries) collectEndpoints(descr envDescriber, svc, env, port string) error {
```
Lou1415926: nit: these can probably be moved to `service.go` since they aren't lbws specific.

internal/pkg/describe/lb_web_service.go
Lou1415926: Could `internalEndpoints` be a `map[Endpoint]Environments` where `Endpoint string` and `Environments []stirng`? Would this increase readability?

internal/pkg/aws/ecs/service.go
iamhopaul123: Maybe not because in CFN template there's a comment, and here it should be agnostic about how Copilot will use it.

internal/pkg/describe/lb_web_service.go
iamhopaul123: I like this idea. i think readability is about the same but it'll be more efficient

internal/pkg/describe/backend_service_test.go
```diff
@@ -595,7 +584,7 @@ Resources
...
...p\",\"configurations\":[{\"environment\":\"test\",\"port\":\"80\",\"cpu\":\"256\",\"memory\":\"512\",\"platform\":\"LINUX/X86_64\",\"tasks\":\"1\"},{\"environment\":\"prod\",\"port\":\"5000\",\"cpu\":\"512\",\"memory\":\"1024\",\"platform\":\"LINUX/ARM64\",\"tasks\":\"3\"}],\"routes\":null,\"serviceDiscovery\":[{\"environment\":[\"prod\"],\"endpoint\":\"http://my-svc.prod.my-app.local:5000\"},{\"environment\":[\"test\"],\"endpoint\":\"http://my-svc.test.my-app.local:5000\"}],\"variables\":[{\"environment\":\"prod\",\"name\":\"COPILOT_ENVIRONMENT_NAME\",\"value\":\"prod\",\"container\":\"container\"},{\"environment\":\"test\",\"name\":\"COPILOT_ENVIRONMENT_NAME\",\"value\":\"test\",\"container\":\"container\"}],\"secrets\":[{\"name\":\"GITHUB_WEBHOOK_SECRET\",\"container\":\"container\",\"environment\":\"test\",\"valueFrom\":\"GH_WEBHOOK_SECRET\"},{\"name\":\"SOME_OTHER_SECRET\",\"container\":\"container\",\"environment\":\"prod\",\"valueFrom\":\"SHHHHH\"}],\"resources\":{\"prod\":[{\"type\":\"AWS::EC2::SecurityGroupIngress\",\"physicalID\":\"ContainerSecurityGroupIngressFromPublicALB\"}],\"test\":[{\"type\":\"AWS::EC2::SecurityGroup\",\"physicalID\":\"sg-0758ed6b233743530\"}]}}\n",
```
Lou1415926, line 587: i can't tell what changed :open_eye_crying_laughing

internal/pkg/describe/service.go
```diff
@@ -406,3 +423,151 @@ func (e containerEnvVars) humanString(w io.Writer) {
...
+func underline(headings []string) []string {
+	var lines []string
+	for _, heading := range headings {
+		line := strings.Repeat("-", len(heading))
+		lines = append(lines, line)
+	}
+	return lines
+}
+
+type internalEndpoint struct {
```
Lou1415926: nit: what do you think of moving the struct inside of the `marshalJSON` since it's the only place where it's used?

internal/pkg/describe/service.go
```diff
@@ -406,3 +423,151 @@ func (e containerEnvVars) humanString(w io.Writer) {
...
+			Environment: (*e)[endpoint],
+			Endpoint:    endpoint,
+		})
+	}
+	sort.Slice(internalEndpoints, func(i, j int) bool { return internalEndpoints[i].Endpoint < internalEndpoints[j].Endpoint })
+	return json.Marshal(&internalEndpoints)
+}
+
+func (e *endpointToEnvs) add(endpoint string, env string) {
+	(*e)[endpoint] = append((*e)[endpoint], env)
```
Lou1415926: I think go's map is passed by reference so we could probably just:
```go
func (e endpointToEnvs) add(endpoint string, env string) {
	e[endpoint] = append(e[endpoint], env)
```

Lou1415926 (approved):
LGTM just some nits!!

internal/pkg/describe/backend_service_test.go
iamhopaul123: Order lol

PR #4085, opened 2022-10-13
chore(service-connect): update uri and fix deployment

iamhopaul123:


internal/pkg/describe/uri.go
```diff
@@ -198,6 +199,16 @@ func (d *BackendServiceDescriber) URI(envName string) (URI, error) {
 			AccessType: URIAccessTypeNone,
 		}, nil
 	}
+	scDNSNames, err := svcDescr.ServiceConnectDNSNames()
+	if err != nil {
+		return URI{}, fmt.Errorf("retrieve service connect DNS names: %w", err)
+	}
+	if len(scDNSNames) != 0 {
+		return URI{
```
Lou1415926, line 207: is it possible that the service has both sc and sd? should we fetch uris for both?

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
```diff
@@ -54,6 +55,9 @@ ServiceConnectConfiguration:
           {{- else}}
           DnsName: !Ref WorkloadName
           {{- end}}
+  {{- else}}
+  Enabled: False
```
Lou1415926, line 59: What was the issue with the original snippet? When sc is not enabled, how does CFN/ECS handle it differently between removing the entire `ServiceConnectConfiguration`  vs. setting `Enabled: False`?

internal/pkg/describe/uri.go
iamhopaul123: Yes it is possible. I *think* it's probably fine to show SC instead of SD in this case? Because if they have SC enabled they can just use SC. Like internal ALB when it is enabled we don't show SD as well.

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
iamhopaul123: If we don't set `ServiceConnectConfiguration` CFN keeps using the old template 🤦 If you want to disable this feature you have to explicitly specify in the template.

Lou1415926 (approved):

internal/pkg/describe/uri.go
```diff
@@ -198,6 +199,16 @@ func (d *BackendServiceDescriber) URI(envName string) (URI, error) {
 			AccessType: URIAccessTypeNone,
 		}, nil
 	}
+	scDNSNames, err := svcDescr.ServiceConnectDNSNames()
+	if err != nil {
+		return URI{}, fmt.Errorf("retrieve service connect DNS names: %w", err)
+	}
+	if len(scDNSNames) != 0 {
```
dannyrandall: nit
```suggestion
	if len(scDNSNames) > 0 {
```

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
```diff
@@ -45,7 +45,7 @@
 {{- end}} {{/* end if eq .WorkloadType "Load Balanced Web Service"*/}}
 {{- if eq .WorkloadType "Backend Service"}}
 {{- if eq .HTTPTargetContainer.Name .WorkloadName}} 
-{{- if and .ServiceConnect .SCFeatureFlag}} # remove when release
+{{- if .SCFeatureFlag}} # remove when release
   PortMappings: !If [ExposePort, [{ContainerPort: !Ref ContainerPort, Name: target}], !Ref "AWS::NoValue"]
```
dannyrandall, line 49: so after we'll just always set `Name`, even if they disable sc? i assume that's fine?

iamhopaul123: Yeah i think so too.

dannyrandall (approved):

PR #4089, opened 2022-10-14
chore: update recommended actions for missing credentials

genbit:
This removes link to AWS Go SDK and uses only link to AWS Copilot docs

### DIFF
--- a/internal/pkg/aws/sessions/errors.go
+++ b/internal/pkg/aws/sessions/errors.go
@@ -39,16 +39,11 @@ func (e *errCredRetrieval) Error() string {
 // RecommendActions returns recommended actions to be taken after the error.
 // Implements main.actionRecommender interface.
 func (e *errCredRetrieval) RecommendActions() string {
-	notice := "It looks like your credential settings are misconfigured or missing"
+	notice := "It looks like AWS credentials are misconfigured or missing"
 	if e.profile != "" {
-		notice = fmt.Sprintf("It looks like your profile [%s] is misconfigured or missing", e.profile)
+		notice = fmt.Sprintf("It looks like AWS profile [%s] is misconfigured or missing", e.profile)
 	}
 	return fmt.Sprintf(`%s:
-https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html#specifying-credentials
-- We recommend including your credentials in the shared credentials file.
-- Alternatively, you can also set credentials through 
-	* Environment Variables
-	* EC2 Instance Metadata (credentials only)
 More information: https://aws.github.io/copilot-cli/docs/credentials/`, notice)
 }
 


internal/pkg/aws/sessions/errors.go
```diff
@@ -39,16 +39,11 @@ func (e *errCredRetrieval) Error() string {
 // RecommendActions returns recommended actions to be taken after the error.
 // Implements main.actionRecommender interface.
 func (e *errCredRetrieval) RecommendActions() string {
-	notice := "It looks like your credential settings are misconfigured or missing"
+	notice := "It looks like AWS credentials are misconfigured or missing"
```
huanjani, line 42:
```suggestion
	notice := "It looks like your AWS credentials are misconfigured or missing"
```

internal/pkg/aws/sessions/errors.go
```diff
@@ -39,16 +39,11 @@ func (e *errCredRetrieval) Error() string {
...
 	if e.profile != "" {
-		notice = fmt.Sprintf("It looks like your profile [%s] is misconfigured or missing", e.profile)
+		notice = fmt.Sprintf("It looks like AWS profile [%s] is misconfigured or missing", e.profile)
```
huanjani, line 44:
```suggestion
		notice = fmt.Sprintf("It looks like your AWS profile [%s] is misconfigured or missing", e.profile)
```

kohidave:
Are you doing this for Hacktoberfest22 cred 🤔

Just kidding, thank you!!!

internal/pkg/aws/sessions/errors.go
```diff
@@ -39,16 +39,11 @@ func (e *errCredRetrieval) Error() string {
...
 	}
 	return fmt.Sprintf(`%s:
-https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html#specifying-credentials
-- We recommend including your credentials in the shared credentials file.
-- Alternatively, you can also set credentials through 
-	* Environment Variables
-	* EC2 Instance Metadata (credentials only)
```
iamhopaul123, line 51: Why do we remove these 🤔

iamhopaul123 (approved):
LGTM! Just a question.

dannyrandall (approved):
+1 to @huanjani's fb but otherwise looks great!! 🥳

internal/pkg/aws/sessions/errors.go
genbit: Documentation link has ll details and for someone who is using copilot for first time and doesnt have relevant experience of dealing with  aws  from terminal — those don’t mean match without context. In fact could give a wrong impression that one need to run copilot from within EC2 instance. 

If the goal for those messages to provide ideas on how someone who think they configured credentials and now need to debug missing credentials error - we should think more about this use case and how to help

iamhopaul123: Gotcha makes sense to me.

internal/pkg/aws/sessions/errors.go
genbit: curious, why your is needed?

internal/pkg/aws/sessions/errors.go
genbit: curious, why your is needed?

internal/pkg/aws/sessions/errors.go
huanjani: So the user knows it's their issue to fix rather than something internal.

internal/pkg/aws/sessions/errors.go
huanjani: This one not so much bc you name the profile. So we could take it or leave it. But it emphasizes that it's the user's profile.

PR #4097, opened 2022-10-18
chore: add `read_only_fs` as a comment in the default manifest 

Lou1415926:
This PR comments out the `read_only_fs` field that was originally added by default to the manifest. Setting `read_only_fs: true` prevents containers from `RUN`ning command such as `mkdir`.

We still recommend enabling this field. So we add it in the manifest as a comment for visibility.


internal/pkg/template/templates/workloads/services/backend/manifest.yml
```diff
@@ -43,10 +43,11 @@ count: {{.Count.Value}}       # Number of tasks that should be running in your s
...
 exec: true     # Enable running commands in your container.
 {{- end}}
-
-storage:
+  
 {{- if not .TaskConfig.IsWindows}}
-  readonly_fs: true       # Limit to read-only access to mounted root filesystems by default.
+
+# storage:
+  # readonly_fs: true       # Limit to read-only access to mounted root filesystems.
```
paragbhingre, line 50: So by default `readonly_fs` won't be enabled?  But isn't that contradicting with the original [ask](https://github.com/aws/copilot-cli/issues/4010)?

Lou1415926: Adding that field by default in manifest adds to confusion for users with `RUN some write command` in their Dockerfile. Worse, they'd have to wait for the circuit breaker to get triggered in order to know the deployment fails.

This is a middle ground between "adding it by default for strict security" and "not adding it at all for smooth experience"

paragbhingre: > Worse, they'd have to wait for the circuit breaker to get triggered in order to know the deployment fails.

Totally agree to this.

huanjani (approved):

efekarakus (approved):

PR #4100, opened 2022-10-20
chore: add yaml Union type

dannyrandall:
This PR contains a new generic `Union` type in `package manifest` that can hold one of two underlying types. It also changes the existing type `HealthCheckArgsOrString` to use type `Union[string, HTTPHealthCheckArgs]`. Obviously it will make more sense once we convert _all_ of the yaml union types to use `Union`, but changing it in just this one place should give us chance to test it a bit and make sure we like the implementation.

Some benefits of this change:
- No more custom unmarshal logic for each existing union type
- Can use a single transformer for merging union types

Downsides:
- Struct field access might look a little weird: `type.Field.A` or `type.Field.B.Field`

One bit of weirdness I noticed is that `NewUnionB()` will infer type B, but for some reason the same doesn't happen in `NewUnionA()`. That's why `NewUnionA` has both types but `NewUnionB` has a single type. (unnecessary types are a lint error)

Another callout so it's not missed is that I removed the error when unmarshalling into `HealthCheckArgsOrString` results in an empty `HTTPHealthCheckArgs` struct. This means that previously, we would error with this yaml:
```
http:
  healthcheck:
    random_key: value
```
but now we won't error, the Union value will just be zero or nil (see Union.Unmarshal()). This brings sub fields in line with how we process top level yaml fields - for example, we don't error when unmarshalling this:
```
random_key: value
http:
  path: "/"
```


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -325,29 +325,36 @@ func convertAutoscaling(a manifest.AdvancedCount) (*template.AutoscalingOpts, er
...
-	if hc.HealthCheckArgs.Path != nil {
-		opts.HealthCheckPath = *hc.HealthCheckArgs.Path
-	} else if hc.HealthCheckPath != nil {
-		opts.HealthCheckPath = *hc.HealthCheckPath
+
+	if hc.IsZero() {
+		return opts
+	}
+	if hc.IsA() {
+		opts.HealthCheckPath = hc.A
```
iamhopaul123: I wonder in Copilot since all the union type consists of a simpler one and a more complicated one: like `string` and some struct, maybe we can make `A` and `B` more informative.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -325,29 +325,36 @@ func convertAutoscaling(a manifest.AdvancedCount) (*template.AutoscalingOpts, er
...
 	}
-	if hc.HealthCheckArgs.Timeout != nil {
-		opts.Timeout = aws.Int64(int64(hc.HealthCheckArgs.Timeout.Seconds()))
+	if hc.B.Timeout != nil {
+		opts.Timeout = aws.Int64(int64(hc.B.Timeout.Seconds()))
 	}
-	if hc.HealthCheckArgs.GracePeriod != nil {
-		opts.GracePeriod = int64(hc.HealthCheckArgs.GracePeriod.Seconds())
+	if hc.B.GracePeriod != nil {
+		opts.GracePeriod = int64(hc.B.GracePeriod.Seconds())
```
iamhopaul123: I wonder if we can simplify the logic and use `B` in literal.
```golang
opts := template.HTTPHealthCheckOpts{
		HealthCheckPath:    manifest.DefaultHealthCheckPath,
		HealthyThreshold:   hc.B.HealthyThreshold,
		UnhealthyThreshold: hc.B.UnhealthyThreshold,
		GracePeriod:        manifest.DefaultHealthCheckGracePeriod,
......
	}
if hc.IsA() {
  opts.HealthCheckPath = hc.A
}
return opts
```

internal/pkg/manifest/union.go
```diff
@@ -0,0 +1,163 @@
...
+// Union will only ever hold one of the underlying types, never both.
+//
+// Use NewUnionA() and NewUnionB() to create a Union with an underlying
+// type already set. See Unmarshal() for details on how yaml is decoded
+// into Union.
+//
+// Union is exported to enable type embedding.
+type Union[A, B any] struct {
+	// isA is true if the underlying type of Union is A.
+	isA bool
```
iamhopaul123: Something like `Short` vs `AdvancedConfig` or anything equivalent.  At any rate it is `manifest.Union` so it is possible that this type is not an absolutely general `Union`.

internal/pkg/manifest/union.go
```diff
@@ -0,0 +1,163 @@
...
+
+// IsB returns true if the underlying value of t is type B.
+func (t *Union[_, _]) IsB() bool {
+	return t.isB
+}
+
+// Value returns either the underlying value of t, which is either type
+// A or B, depending on which was filled by yaml.Unmarshal or created
+// with NewUnionA/B(). If neither have been/were set, Value returns nil.
+func (t *Union[A, B]) Value() any {
```
iamhopaul123: Does this need to be a public func?

internal/pkg/manifest/union.go
```diff
@@ -0,0 +1,163 @@
...
+	if aErr != nil {
+		return aErr
+	}
+	return bErr
+}
+
+// isZeroerAndZero returns true if v is a yaml.Zeroer
+// _and_ is zero, as determined by v.IsZero().
+func isZeroerAndZero(v any) bool {
+	z, ok := v.(yaml.IsZeroer)
```
iamhopaul123, line 102: How does this work if hc arg doesn't implement `IsZero`? Or did I miss anything 🤔  https://github.com/aws/copilot-cli/pull/4100/files#diff-4465c235880a9abf21ad9b3cab6077493fa6f767be82b60324def2bbbc3b23e2L368

iamhopaul123 (commented):
Great innovation! Could remove a lot of code duplication 🎉

internal/pkg/deploy/cloudformation/stack/transformers.go
dannyrandall: do you think it would make sense in all/enough cases to change `A` -> `Simple` and `B` -> `Complex`, or something like that? that could make it a little better for reading 🤔

iamhopaul123: Yeah i think simple and complex sounds good enough to me.

dannyrandall: I just checked all of our structs - pretty much all are `[builtinType, struct]`, but a few are a bit more complex:
```
SubnetListOrArgs -> []string | SubnetArgs
SecurityGroupsIDsOrConfig -> []string | SecurityGroupsConfig
```
I'll add a test to confirm those types work (I may already have a test, just want to make sure :)) I think I'm still ok with [`Simple`/`Advanced`](https://github.com/aws/copilot-cli/pull/4100#discussion_r1001983070) in those scenarios, what do you think?

internal/pkg/manifest/union.go
dannyrandall: discussed in https://github.com/aws/copilot-cli/pull/4100#discussion_r1001974462!

internal/pkg/deploy/cloudformation/stack/transformers.go
iamhopaul123: :shipit:

internal/pkg/manifest/union.go
dannyrandall: good catch! i think we can actually just remove it and move it's logic to `MarshalYAML()`

internal/pkg/manifest/union.go
dannyrandall: that link isn't working for me so i'm not sure what you're referencing 😢. if `v` doesn't implement `IsZero()` this func returns false. So basically when unmarshalling, if your type _does not_ implement `IsZero()`, and there was no error from `value.Decode(typeWithoutIsZero)`, then `Union` will assume it was that type.

If your type _does_ implement `IsZero()`, and there was no error from `value.Decode(typeWithZero)`, `Union` will _not_ set that type, and will try the other one or just be nil.

dannyrandall: An example could be:
```
type NoIsZero struct {
  Key string `yaml:"key"`
}

type HasIsZero struct {
  Key string `yaml:"key"`
}

func (z *HasIsZero) IsZero() {
  return z.Key == ""
}
```

input yaml:
```
obj:
  key: ""
```
---
Say you unmarshal into: 
```
type Obj struct {
  Union[HasIsZero, NoIsZero]
}
```
The Union's type will be NoIsZero. However, if the type was `Union[NoIsZero, HasIsZero]`, that type's Union would be NoIsZero, because there was no error unmarshalling into that type.

Does that help?

internal/pkg/deploy/cloudformation/stack/transformers.go
dannyrandall: the only type we can move up to the declaration is `SuccessCodes` - the others all need some conversion and we don't want to set them if they are unset in the Advanced config. I'll move up that up though!

internal/pkg/manifest/union.go
dannyrandall: discussed offline, changing `HTTPHealthCheckArgs.isEmpty()` to `HTTPHealthCheckArgs.IsZero()`

internal/pkg/deploy/cloudformation/stack/transformers.go
huanjani: Ooh I like "advanced" more than "complex"...how do you feel about "basic" over "simple"?

internal/pkg/manifest/svc.go
```diff
@@ -365,57 +365,23 @@ type HTTPHealthCheckArgs struct {
 	GracePeriod        *time.Duration `yaml:"grace_period"`
 }
 
-func (h *HTTPHealthCheckArgs) isEmpty() bool {
+func (h *HTTPHealthCheckArgs) IsZero() bool {
```
efekarakus, line 369: Can we add a doc comment here saying that it implements the [`yaml.IsZeroer`](https://pkg.go.dev/gopkg.in/yaml.v3#IsZeroer)  interface? 🥺

internal/pkg/deploy/cloudformation/stack/transformers.go
dannyrandall: i don't care _too_ much either way - i feel like Basic sorta implies a subset (but still a struct) of Advanced though? as opposed to a single field? maybe that's just me though 🙄

internal/pkg/manifest/initial_manifest_integration_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build localintegration
```
efekarakus, line 1: (I know this isn't related to your change)

Would you mind renaming the file from `initial_manifest_integration_test.go` to maybe `marshal_manifest_integration_test.go`?

(sidenote: It'd be dreamy if we can actually move all of our integ tests to a separate directory to make our source files a bit easier to parse and ensure we don't refer to internal state.)

internal/pkg/manifest/svc.go
```diff
@@ -365,57 +365,23 @@ type HTTPHealthCheckArgs struct {
...
-	return nil
-}
-
-// IsEmpty returns true if there are no health check configuration set.
-func (hc *HealthCheckArgsOrString) IsEmpty() bool {
-	if hc.HealthCheckPath != nil {
-		return false
-	}
-	return hc.HealthCheckArgs.isEmpty()
+	Union[string, HTTPHealthCheckArgs]
```
efekarakus, line 377: Instead of embeddding, can we just write the following?
```go 
type HealthCheckArgsOrString Union[string, HTTPHealthCheckArgs]
```


I'm thinking now with your configuration we can simplify a lot the names of our types, what do you think of the following?
```go
type AdvancedHTTPHealthCheckConfig struct {}
type HTTPHealthCheckConfig Union[string, AdvancedHTTPHealthCheckConfig]
```

internal/pkg/manifest/union.go
```diff
@@ -0,0 +1,156 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package manifest
+
+import (
+	"gopkg.in/yaml.v3"
+)
+
+// Union is a type used for yaml keys that may be of type Simple or Advanced.
```
efekarakus: Let me know if I'm being pedantic (requiring precision without illuminating anything 😂). A `Union` to me sounds like an _inclusive or_ where it's valid to set both `Simple` or `Advanced` or both.
Whereas this type is an `xor` operation.

What do you think about the following names:
```go
type XOrConfig[Basic, Advanced any] struct { }
type EitherConfig[Basic, Advanced any] struct { }
type LeveledConfig[Basic, Advanced any] struct { }
type ProgressiveConfig[Basic, Advanced any] struct { }
```

internal/pkg/manifest/union.go
```diff
@@ -0,0 +1,156 @@
...
+// Union will only ever hold one of the underlying types, never both.
+//
+// Use NewUnionSimple() and NewUnionAdvanced() to create a Union with an underlying
+// type already set. See Unmarshal() for details on how yaml is decoded
+// into Union.
```
efekarakus: nit: not needed

internal/pkg/manifest/union.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// Simple holds the value of Union if IsSimple() is true.
+	// If IsSimple() is false, this is the zero value of type Simple.
+	//
+	// Simple is exported to support mergo. It should not be set
+	// directly. Use NewUnionSimple() to create a Union with Simple set.
+	Simple Simple
+
+	// isAdvanced is true if the underlying type of Union is Advanced.
+	isAdvanced bool
```
efekarakus, line 26: isn't `isAdvanced = !isSimple`? do we need this field

internal/pkg/manifest/union.go
```diff
@@ -0,0 +1,156 @@
...
+	// If IsAdvanced() is false, this is the zero value of type Advanced.
+	//
+	// Advanced is exported to support mergo. It should not be set
+	// directly. Use NewUnionAdvanced() to create a Union with Advanced set.
+	Advanced Advanced
+}
+
+// NewUnionSimple creates a new Union[Simple, Advanced] with the underlying
+// type set to Simple, holding val.
+func NewUnionSimple[Simple, Advanced any](val Simple) Union[Simple, Advanced] {
```
efekarakus: nit: do you think it's nicer to read without the `New` prefix?
```go
func BasicLeveledConfig[Basic, Advanced any](val Simple) BasicLeveledConfig[Simple, Advanced] {
```

efekarakus (commented):
I haven't read everything yet! I am publishing my comments for now to hear your thoughts 💭

internal/pkg/manifest/initial_manifest_integration_test.go
dannyrandall: renamed for now, it would be a great idea to have all of our integ tests (in other packages too!) in a separate package from the areas they test!

internal/pkg/manifest/svc.go
dannyrandall: so i've been thinking about this option, but the main downside i've found is the inability to define functions on the new types. here's my synopsis on the different options (that i can think of!):

1. Embed in a struct (current)
```
type stringOrAdvanced struct {
  Union[string, advancedType]
}

func (s *stringOrAdvanced) Foo() string {
  return s.Union.Simple()
}
```
The main downside here is that you have to initialized the `stringOrAdvanced` struct every time you declare this type (see test files in this PR).

2. Create named type
```
type stringOrAdvanced *Union[string, advancedType]

// not allowed (compile err: invalid receiver type stringOrAdvanced (pointer or interface type) 
func (s *stringOrAdvanced) Hi() string {}
```
So the clear downside is that you can't define other functions, but it is nicer when declaring the type inline, though we'd probably have to change the `NewUnionXxx` functions to return a pointer instead to make that possible. We can't use `type stringOrAdvanced *Union[string, advancedType]` because then `stringOrAdvanced` no longer has an `UnmarshalYAML()` function on it...so it uses the default. I'm also not entirely sure if there is a potential for nil because of the pointer - it's kinda weird.

3. Type alias
```
type advancedType struct{}
type stringOrAdvanced = Union[string, advancedType]

// not allowed (compile err: cannot define new methods on instantiated type Union[string, advancedType])
func (s *stringOrAdvanced) Hi() {}
```
Same downside of the named type of not being able to declare a function, but I think avoids the potential nil oddity.

A helpful read: https://github.com/golang/proposal/blob/master/design/18130-type-alias.md

---
So those are the tradeoffs. If we wanted to do option 2 or 3, we could easily add functions onto their parent structs (i.e., `HealthCheckPath()`) to get around the function issue. Either way, to your point, we can remove the `ThisOrThat` naming convention from our types.

If you can think of any other options let me know and I can test them out :)

internal/pkg/manifest/union.go
dannyrandall: clever clever 😏

dannyrandall: actually jk. we need it😛. there are three states:
1. Underlying type is `Simple`
2. Underlying type is `Advanced`
3. Underlying type is neither

we could represent with a single field (int enum or *bool) instead, though i think i prefer this option 🤔

internal/pkg/manifest/union.go
dannyrandall: i think a union type implies that it's not valid to be both at once 🥺: 
> In contrast with a [record](https://en.wikipedia.org/wiki/Record_(computer_science)) (or structure), which could be defined to contain a float and an integer; in a union, there is only one value at any given time.

(https://en.wikipedia.org/wiki/Union_type)

internal/pkg/manifest/svc.go
efekarakus: Gotcha! okay so the maindownside for anyone else wondering is that if you write:
```go
type stringOrAdvanced Union[string, advancedType]
```
The method set of `Union` isn't embeded in `stringOrAdvanced`. Makes sense 👍 let's keep it embedded like you have!

internal/pkg/deploy/cloudformation/stack/transformers.go
dannyrandall: i changed the name to Basic😊. It _does_ support `[structA, structB]` so I think that it makes more sense that way

internal/pkg/manifest/union.go
```diff
@@ -0,0 +1,152 @@
...
+func BasicToUnion[Basic, Advanced any](val Basic) Union[Basic, Advanced] {
+	return Union[Basic, Advanced]{
+		isBasic: true,
+		Basic:   val,
+	}
+}
+
+// NewUnionAdvanced creates a new Union[Basic, Advanced] with the underlying
+// type set to Advanced, holding val.
+func NewUnionAdvanced[Basic, Advanced any](val Advanced) Union[Basic, Advanced] {
```
efekarakus: 🥺 
```suggestion
func AdvancedToUnion[Basic, Advanced any](val Advanced) Union[Basic, Advanced] {
```

internal/pkg/manifest/union.go
```diff
@@ -0,0 +1,152 @@
...
+	case t.IsAdvanced():
+		return t.Advanced, nil
+	}
+	return nil, nil
+}
+
+// IsZero returns true if the set value of t implements
+// yaml.IsZeroer and IsZero. It also returns true if
+// neither value for t is set.
+func (t Union[_, _]) IsZero() bool {
```
efekarakus, line 123: nit: can these be pointer receivers as well so that all receivers are consistent?

internal/pkg/manifest/union.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	return true
+}
+
+// validate calls t.validate() on the set value of t. If the
+// current value doesn't have a validate() function, it returns nil.
+func (t Union[_, _]) validate() error {
+	// type declarations inside generic functions not currently supported,
+	// so we use an inline validate() interface
+	if t.isBasic {
```
efekarakus: nit: can we use the methods in the if-statements so that it's easier to refactor the internals in case it's needed
```suggestion
	if t.IsBasic() {
```
ps: I kinda love calling something `something.IsBasic()`

efekarakus (approved):
LGTM! Can't wait to refactor all of our remaining types and rename them ✨ 

Some tiny comments, feel free to remove the "do-not-merge" label afterwards

iamhopaul123 (approved):
:yay_fox:

internal/pkg/manifest/union.go
dannyrandall: the three funcs that have value receivers (`MarshalYAML()`, `IsZero()`, and `validate()`) all get called after the type has been cast to an interface, and since `Union` is embedded, not `*Union`, they have to be on the value receiver.

Issue #3553, opened 2022-05-11
Pipeline fails in CodeBuild POST_BUILD with `COMMAND_EXECUTION_ERROR Message: Error while executing command: for env in $pl_envs;`


ryancody:
Hello, I'm using Copilot in windows for a dotnet project.  I have a pipeline that that fails in CodeBuild on the POST_BUILD phase with the following error:

<img width="677" alt="image" src="https://user-images.githubusercontent.com/8892867/167770808-b5731c7b-f72d-4c1d-b847-255341e393e5.png">

Build logs shows this error:
`[Container] 2022/05/11 04:20:06 Phase context status code: COMMAND_EXECUTION_ERROR Message: Error while executing command: for env in $pl_envs; do
  tag=$(sed 's/:/-/g' <<<"${CODEBUILD_BUILD_ID##*:}-${env}" | rev | cut -c 1-128 | rev)
  for svc in $svcs; do
  ./copilot-linux svc package -n $svc -e $env --output-dir './infrastructure' --tag $tag --upload-assets;
  if [ $? -ne 0 ]; then
    echo "Cloudformation stack and config files were not generated. Please check build logs to see if there was a manifest validation error." 1>&2;
    exit 1;
  fi
  done;
  for job in $jobs; do
  ./copilot-linux job package -n $job -e $env --output-dir './infrastructure' --tag $tag --upload-assets;
  if [ $? -ne 0 ]; then
    echo "Cloudformation stack and config files were not generated. Please check build logs to see if there was a manifest validation error." 1>&2;
    exit 1;
  fi
  done;
done;
. Reason: exit status 1`

It refers to this line in buildspec.yml:
<img width="538" alt="image" src="https://user-images.githubusercontent.com/8892867/167771186-2ac7fd36-6f9a-4a35-bf40-a233752e2c74.png">

huanjani:
Hi, @ryancody!
Would you mind copy-pasting your manifest here?
Thanks!

ryancody:
```
# The manifest for the "projectname-master" pipeline.
# This YAML file defines your pipeline: the source repository it tracks and the order of the environments to deploy to.
# For more info: https://aws.github.io/copilot-cli/docs/manifest/pipeline/

# The name of the pipeline.
name: projectname-master

# The version of the schema used in this template.
version: 1

# This section defines your source, changes to which trigger your pipeline.
source:
  # The name of the provider that is used to store the source artifacts.
[...]

PR #4102, opened 2022-10-21
chore(pipelines): remove here string from buildspec

huanjani:
Fixes second issue in #3553.

Here strings (`<<<`) don't work with `sh`. This change makes the pipeline buildspec a bit more universal.

### DIFF
--- a/internal/pkg/template/templates/cicd/buildspec.yml
+++ b/internal/pkg/template/templates/cicd/buildspec.yml
@@ -49,7 +49,7 @@ phases:
       # Check if the `svc package` commanded exited with a non-zero status. If so, echo error msg and exit.
       - >
         for env in $pl_envs; do
-          tag=$(sed 's/:/-/g' <<<"${CODEBUILD_BUILD_ID##*:}-${env}" | rev | cut -c 1-128 | rev)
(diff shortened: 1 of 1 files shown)


iamhopaul123 (approved):

internal/pkg/template/templates/cicd/buildspec.yml
```diff
@@ -49,7 +49,7 @@ phases:
       # Check if the `svc package` commanded exited with a non-zero status. If so, echo error msg and exit.
       - >
         for env in $pl_envs; do
-          tag=$(sed 's/:/-/g' <<<"${CODEBUILD_BUILD_ID##*:}-${env}" | rev | cut -c 1-128 | rev)
+          tag=$(echo ${CODEBUILD_BUILD_ID##*:}-$env | sed 's/:/-/g' | rev | cut -c 1-128 | rev)
```
efekarakus, line 52: Just to make sure was this tested in manually? do the tags in ECR look the same as before by default?

huanjani: yup yup!

efekarakus (approved):

PR #4103, opened 2022-10-21
feat: allow additional source ips to the Public ALB's SG

### LINKED ISSUE #2557: Allow additional security groups to the load balancer

paragbhingre:
Resolves #2557 


internal/pkg/manifest/env.go
```diff
@@ -507,7 +512,8 @@ func (cfg *EnvironmentConfig) ELBAccessLogs() (*ELBAccessLogsArgs, bool) {
 // RestrictiveIngress represents ingress fields which restrict
 // default behavior of allowing all public ingress.
 type RestrictiveIngress struct {
-	CDNIngress *bool `yaml:"cdn"`
+	CDNIngress *bool   `yaml:"cdn"`
+	SourceIPs  []IPNet `yaml:"source_ips"`
```
iamhopaul123, line 517: :+1:

iamhopaul123 (approved):
:shipit:

internal/pkg/cli/deploy/env.go
```diff
@@ -383,6 +394,7 @@ func (d *envDeployer) buildStackInput(in *DeployEnvironmentInput) (*deploy.Creat
 		return nil, err
 	}
 	cidrPrefixListIDs, err := d.cidrPrefixLists(in)
+	publicALBSourceIPs := d.publicALBSourceIPList(in)
```
huanjani: This is splitting up the line above and line below, which should be adjacent. 

Can this just go below, in line 413 as
`PublicALBSourceIPs: d.publicALBSourceIPList(in)`?

internal/pkg/cli/deploy/env.go
```diff
@@ -293,6 +293,17 @@ func (d *envDeployer) cidrPrefixLists(in *DeployEnvironmentInput) ([]string, err
...
 
+func (d *envDeployer) publicALBSourceIPList(in *DeployEnvironmentInput) []string {
+	if in.Manifest == nil || len(in.Manifest.GetPublicALBSourceIPs()) == 0 {
+		return nil
+	}
+	var publicALBSourceIPList []string
+	for _, sourceIP := range in.Manifest.GetPublicALBSourceIPs() {
+		publicALBSourceIPList = append(publicALBSourceIPList, string(sourceIP))
+	}
+	return publicALBSourceIPList
```
efekarakus: nit:
```suggestion
	ips := make([]string, len(in.Manifest.GetPublicALBSourceIPs()))
	for i, sourceIP := range in.Manifest.GetPublicALBSourceIPs() {
		ips[i] = string(sourceIP)
	}
	return ips
```
I follow [this rule from the naming guideline](https://a-nickels-worth.blogspot.com/2016/04/a-guide-to-naming-variables.html#orgheadline12) for a past sr. principal. In this case `ips` carry the same amount of info as `publicALBSourceIPList` because it's in the field of vision. Might as well make it shorter.

internal/pkg/cli/deploy/env.go
```diff
@@ -293,6 +293,17 @@ func (d *envDeployer) cidrPrefixLists(in *DeployEnvironmentInput) ([]string, err
 	return cidrPrefixListIDs, nil
 }
 
+func (d *envDeployer) publicALBSourceIPList(in *DeployEnvironmentInput) []string {
```
```suggestion
func (d *envDeployer) publicALBSourceIPs(in *DeployEnvironmentInput) []string {
```

internal/pkg/cli/deploy/env.go
```diff
@@ -398,6 +410,7 @@ func (d *envDeployer) buildStackInput(in *DeployEnvironmentInput) (*deploy.Creat
 		ArtifactBucketARN:    s3.FormatARN(partition.ID(), resources.S3Bucket),
 		ArtifactBucketKeyARN: resources.KMSKeyARN,
 		CIDRPrefixListIDs:    cidrPrefixListIDs,
+		PublicALBSourceIPs:   publicALBSourceIPs,
```
efekarakus: nit: delete the variable 
```suggestion
		PublicALBSourceIPs:   d.publicALBSourceIPs(in),
```
https://a-nickels-worth.blogspot.com/2016/04/a-guide-to-naming-variables.html#orgheadline13

internal/pkg/template/env.go
```diff
@@ -131,10 +131,11 @@ func (e *EnvOpts) HasImportedCerts() bool {
...
 type HTTPConfig struct {
-	CIDRPrefixListIDs []string
-	ImportedCertARNs  []string
-	CustomALBSubnets  []string
-	ELBAccessLogs     *ELBAccessLogs
+	CIDRPrefixListIDs  []string
+	ImportedCertARNs   []string
+	CustomALBSubnets   []string
+	ELBAccessLogs      *ELBAccessLogs
+	PublicALBSourceIPs []string
```
efekarakus: Hmm my understanding is that this type is for any type of LB. Whereas, the new field is strictly for public ALBs.

What if we create a new type here:
```go
   HTTPConfig
   PublicALBSourceIPs []string
}
func (cfg *PublicHTTPConfig) HasCustomIngress() bool {
   return len(cfg.PublicALBSourceIPs) > 0 && len(CIDRPrefixListIDs) >0 
}
```

efekarakus (commented):
Looks good! some minor comments

internal/pkg/template/env.go
```diff
@@ -109,8 +109,8 @@ type EnvOpts struct {
 	ArtifactBucketKeyARN string
 
 	VPCConfig         VPCConfig
-	PublicHTTPConfig  HTTPConfig
-	PrivateHTTPConfig HTTPConfig
+	PublicHTTPConfig  PublicHTTPConfig
+	PrivateHTTPConfig PrivateHTTPConfig
```
efekarakus, line 113: Awesome, ty so much for also refactoring `PrivateHTTPConfig`!

efekarakus (approved):

PR #4111, opened 2022-10-25
chore: arps manifest updates, cfn template

dannyrandall:

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/env.go
+++ b/internal/pkg/deploy/cloudformation/stack/env.go
@@ -42,6 +42,7 @@ const (
 	envParamInternalALBWorkloadsKey        = "InternalALBWorkloads"
 	envParamEFSWorkloadsKey                = "EFSWorkloads"
 	envParamNATWorkloadsKey                = "NATWorkloads"
+	envParamAppRunnerPrivateWorkloadsKey   = "AppRunnerPrivateWorkloads"
 	envParamCreateHTTPSListenerKey         = "CreateHTTPSListener"
 	envParamCreateInternalHTTPSListenerKey = "CreateInternalHTTPSListener"
 	EnvParamServiceDiscoveryEndpoint       = "ServiceDiscoveryEndpoint"
@@ -192,6 +193,10 @@ func (e *EnvStackConfig) Parameters() ([]*cloudformation.Parameter, error) {
 		},
 		{
 			ParameterKey:   aws.String(envParamNATWorkloadsKey),
+			ParameterValue: aws.String(""),
+		},
+		{
+			ParameterKey:   aws.String(envParamAppRunnerPrivateWorkloadsKey),
 			ParameterValue: aws.String(""),
 		},
 	}
--- a/internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-basic-manifest.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-basic-manifest.yml
@@ -1029,7 +1029,7 @@ Resources:
       SecurityGroupIds:
         - !Ref AppRunnerVpcEndpointSecurityGroup
       # ServiceName: !Sub 'com.amazonaws.${AWS::Region}.apprunner'
-      ServiceName: !Sub "com.amazonaws.${AWS::Region}.apprunner.requests.beta" # TODO switch to non-beta on release
+      ServiceName: !Sub "com.amazonaws.${AWS::Region}.apprunner.requests.gamma" # TODO switch to non-beta on release
       SubnetIds:
         - !Ref PrivateSubnet1
         - !Ref PrivateSubnet2
--- a/internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-basic-manifest.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-basic-manifest.yml
@@ -999,15 +999,15 @@ Resources:
       PublicAccessHostedZone: !GetAtt PublicLoadBalancer.CanonicalHostedZoneID
   AppRunnerVpcEndpointSecurityGroup:
     Metadata:
-      'aws:copilot:description': 'A security group to services in this environment talk to App Runner private services'
+      'aws:copilot:description': 'A security group for App Runner private services'
     Type: AWS::EC2::SecurityGroup
     Condition: CreateAppRunnerVPCEndpoint
     Properties:
-      GroupDescription: !Join ['', [!Ref AppName, '-', !Ref EnvironmentName, AppRunnerVpcEndpointSecurityGroup]]
+      GroupDescription: demo-test-AppRunnerVpcEndpointSecurityGroup
       VpcId: !Ref VPC
       Tags:
         - Key: Name
-          Value: !Sub 'copilot-${AppName}-${EnvironmentName}-app-runner-vpc-endpoint'
+          Value: copilot-demo-test-app-runner-vpc-endpoint
   
   AppRunnerVpcEndpointSecurityGroupIngressFromEnvironment:
     Type: AWS::EC2::SecurityGroupIngress
@@ -1020,7 +1020,7 @@ Resources:
   
   AppRunnerVpcEndpoint:
     Metadata:
-      'aws:copilot:description': 'VPC Endpoint in private subnets to App Runner for private services'
+      'aws:copilot:description': 'VPC Endpoint to connect environment to App Runner for private services'
     Type: AWS::EC2::VPCEndpoint
     Condition: CreateAppRunnerVPCEndpoint
     Properties:
--- a/internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-basic-manifest.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-basic-manifest.yml
@@ -18,6 +18,8 @@ Parameters:
     Type: String
   NATWorkloads:
     Type: String
+  AppRunnerPrivateWorkloads:
+    Type: String
   ToolsAccountPrincipalARN:
     Type: String
   AppDNSName:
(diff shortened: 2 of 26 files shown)


internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-basic-manifest.yml
```diff
@@ -993,6 +997,42 @@ Resources:
       DomainName: !Ref AppDNSName
       PublicAccessDNS: !GetAtt PublicLoadBalancer.DNSName
       PublicAccessHostedZone: !GetAtt PublicLoadBalancer.CanonicalHostedZoneID
+  AppRunnerVpcEndpointSecurityGroup:
+    Metadata:
+      'aws:copilot:description': 'A security group to services in this environment talk to App Runner private services'
```
huanjani:
```suggestion
      'aws:copilot:description': 'A security group to allow services in this environment talk to App Runner private services'
```

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-basic-manifest.yml
```diff
@@ -993,6 +997,42 @@ Resources:
       DomainName: !Ref AppDNSName
       PublicAccessDNS: !GetAtt PublicLoadBalancer.DNSName
       PublicAccessHostedZone: !GetAtt PublicLoadBalancer.CanonicalHostedZoneID
+  AppRunnerVpcEndpointSecurityGroup:
```
huanjani, line 1000: I wonder if we could/should make this and the other VPC endpoint resources more generic so they could potentially be used for other use cases requiring VPC endpoints...?

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-basic-manifest.yml
```diff
@@ -993,6 +997,42 @@ Resources:
...
+    Condition: CreateAppRunnerVPCEndpoint
+    Properties:
+      Description: Ingress from services in the environment
+      GroupId: !Ref AppRunnerVpcEndpointSecurityGroup
+      IpProtocol: -1
+      SourceSecurityGroupId: !Ref EnvironmentSecurityGroup
+  
+  AppRunnerVpcEndpoint:
+    Metadata:
+      'aws:copilot:description': 'VPC Endpoint in private subnets to App Runner for private services'
```
huanjani: nit: Maybe it's just me, but can we add a verb in this description? 😆

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -42,6 +42,7 @@ const (
 	envParamInternalALBWorkloadsKey        = "InternalALBWorkloads"
 	envParamEFSWorkloadsKey                = "EFSWorkloads"
 	envParamNATWorkloadsKey                = "NATWorkloads"
+	envParamAppRunnerPrivateWorkloadsKey   = "AppRunnerPrivateWorkloads"
```
iamhopaul123, line 45: Should we just use `service`?

internal/pkg/template/templates/environment/partials/ar-vpc-connector.yml
```diff
@@ -0,0 +1,48 @@
...
+  Properties:
+    GroupDescription: !Join ['', [!Ref AppName, '-', !Ref EnvironmentName, AppRunnerVpcEndpointSecurityGroup]]
+    {{- if .VPCConfig.Imported}}
+    VpcId: {{.VPCConfig.Imported.ID}}
+    {{- else}}
+    VpcId: !Ref VPC
+    {{- end}}
+    Tags:
+      - Key: Name
+        Value: !Sub 'copilot-${AppName}-${EnvironmentName}-app-runner-vpc-endpoint'
```
efekarakus: Would you mind checking the char limit for `GroupDescription` and `Tag` values.
I wonder if we should use go templates here and do something like:
```
{{truncate "copilot-.App-.Env-apprunner-vpc-endpoint" 255}}
```

internal/pkg/template/templates/environment/partials/ar-vpc-connector.yml
```diff
@@ -0,0 +1,48 @@
+AppRunnerVpcEndpointSecurityGroup:
+  Metadata:
+    'aws:copilot:description': 'A security group to services in this environment talk to App Runner private services'
```
efekarakus: Is the sentence structure a bit odd here? (also nit: can we stay within 80 chars)

```
A security group for App Runner private services
```

internal/pkg/template/templates/environment/partials/ar-vpc-connector.yml
```diff
@@ -0,0 +1,48 @@
...
+  Condition: CreateAppRunnerVPCEndpoint
+  Properties:
+    Description: Ingress from services in the environment
+    GroupId: !Ref AppRunnerVpcEndpointSecurityGroup
+    IpProtocol: -1
+    SourceSecurityGroupId: !Ref EnvironmentSecurityGroup
+
+# TODO should we remove condition on security group, so that it's always created?
+# TODO does ingress from another app runner private service work? It comes through it's VPC connector.
+# re: above: maybe need to add a IngressRule to AppRunnerVpcEndpointSecurityGroup in vpc-connector.yml to enable connection
```
efekarakus: can we remove these? sounds like you found the answer :)

internal/pkg/template/env.go
```diff
@@ -18,35 +18,38 @@ const (
...
-	EFSFeatureName:         "v1.3.0",
-	NATFeatureName:         "v1.3.0",
-	InternalALBFeatureName: "v1.10.0",
-	AliasesFeatureName:     "v1.4.0",
+	ALBFeatureName:                     "v1.0.0",
+	EFSFeatureName:                     "v1.3.0",
+	NATFeatureName:                     "v1.3.0",
+	InternalALBFeatureName:             "v1.10.0",
+	AliasesFeatureName:                 "v1.4.0",
+	AppRunnerPrivateServiceFeatureName: "v1.23.0",
```
efekarakus, line 47: Should we bump the `deploy.LatestEnvTemplateVersion` to `v1.13`? I honestly can't remember why the environment template can't just be equal to the CLI version 🤔 maybe @Lou1415926 remembers. That would get rid of one more place we have to update when we make env template changes.

internal/pkg/template/templates/workloads/partials/cf/instancerole.yml
```diff
@@ -19,7 +19,10 @@ InstanceRole:
       Statement:
         - Effect: Allow
           Principal:
-            Service: tasks.apprunner.amazonaws.com
+            Service:
+              - tasks.apprunner.amazonaws.com
+              - !Sub '${AWS::Region}.beta.credentialvendingserviceproxy.bullet.aws.internal' # TODO remove on release
```
efekarakus, line 24: nit: my suggestion is to keep the internal endpoints + VPC connector change in a separate commit that you can always cherry-pick while locally testing but keeping the PR without these changes

internal/pkg/manifest/rd_web_svc.go
```diff
@@ -81,8 +81,17 @@ func (c *rdwsVpcConfig) isEmpty() bool {
...
 // RequestDrivenWebServiceHttpConfig represents options for configuring http.
 type RequestDrivenWebServiceHttpConfig struct {
-	HealthCheckConfiguration HealthCheckArgsOrString `yaml:"healthcheck"`
-	Alias                    *string                 `yaml:"alias"`
+	HealthCheckConfiguration HealthCheckArgsOrString  `yaml:"healthcheck"`
+	Alias                    *string                  `yaml:"alias"`
+	Private                  Union[bool, VPCEndpoint] `yaml:"private"`
+}
+
+type VPCEndpoint struct {
```
efekarakus, line 90: doc comments plz for the exported names 🥺

efekarakus (commented):
Looks really good! just tiny suggestions

iamhopaul123 (commented):
LGTM overall just one nit

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-basic-manifest.yml
dannyrandall: see what you think of https://github.com/aws/copilot-cli/pull/4111#discussion_r1005012535 😊

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-basic-manifest.yml
dannyrandall: i think we can cross that bridge if/when it comes 😌

internal/pkg/deploy/cloudformation/stack/env.go
dannyrandall: i originally had `service`, but went with Workloads bc of this... https://github.com/dannyrandall/copilot-cli/blob/30866c0a085d43bc0157286da90574b055f773d3/cf-custom-resources/lib/env-controller.js#L113-L116

i'm happy to change it and add logic for the "Service" suffix if you think it's worth it though.

iamhopaul123: it's ok that sounds good to me.

internal/pkg/template/templates/workloads/partials/cf/instancerole.yml
dannyrandall: i'll leave it for now just since i'm merging into a branch - but good idea for future 😊

efekarakus (approved):

iamhopaul123 (approved):

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
```diff
@@ -133,12 +133,18 @@ Resources:
...
       {{- end }}
-      {{- if requiresVPCConnector .}}
       NetworkConfiguration:
         EgressConfiguration:
+          {{- if requiresVPCConnector .}}
           EgressType: VPC
           VpcConnectorArn: !Ref VpcConnector
-      {{- end }}
+          {{- else }}
+          EgressType: DEFAULT
```
Lou1415926, line 142: Is this one of those field that you can't remove after adding it? What does it mean if a template has `EgressType: DEFAULT`, and then it's deployed again without that field?

A more realistic example: say Copilot adds `EgressType: Default` into the template, and then the deployment fails, triggering a rollback removing the field. Will the CFN handler complain?

internal/pkg/template/env.go
Lou1415926: I think env template version was first introduced because of `env upgrade`. 

Retroactively thinking, I assume the concern was that a major change in env template != a major change in Copilot, hence the versioning wasn't kept in sync.

PR #4113, opened 2022-10-26
fix: Check to see if an application exists globally before creating it 

### LINKED ISSUE #815: Check to see if an application exists globally before creating it
# Problem 

Right now application configuration is setup in a particular region - whichever region the default profile is configured with. 

Application names, however, have to be unique within the account, across regions - mostly to reduce the confusion of working on two projects named the same thing, in different regions.

In order to enforce this restriction right now - we create the application admin role as the name of the application. 

https://github.com/aws/copilot-cli/blob/ebd3abd80defc3b170f46b6a4b60306fac4c2b3d/templates/app/versions/v1.0.1/app.yml#L31-L34

So for example, if you run `copilot app init my-proj` in `us-east-1`, then run `copilot app init my-proj` in `us-west-2` - the second init should fail.

That currently happens - but the error you get is pretty gross:

```
Error: execute app init init: describe change set copilot-d2d6af69-9a90-4269-bbaa-204f6e7638b9 for stack codenames-infrastructure-roles: ChangeSetNotFound: ChangeSet [ecscli-d2d6af69-9a90-4269-bbaa-204f6e7638b9] does not exist
	status code: 404, request id: ca14ed0a-aa2d-4d7b-bc54-1a20626d8b06
```

This happens because CloudFormation can't create two roles with the same name since roles are global.

# Fix

To fix this, we should check to see if the admin role exists already (which is of the form `{app-name}-adminrole`). 

In the project deploy code:

https://github.com/aws/copilot-cli/blob/29349c6467e49d6e8c7ca0bc0fa0482910728805/internal/pkg/deploy/cloudformation/app.go#L26-L38

We can add a check which calls fetches `StackSetAdminRoleARN()` from the stack object, calls IAM (via [GetRole](https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#IAM.GetRole)) to see if it exists, and if it does returns an error.

KollaAdithya:
This PR resolves #815
For example, if you run `copilot app init my-proj` in `us-east-1`, then run `copilot app init my-proj` in `us-west-2`  the second init will fail.
This happens because CloudFormation creates an adminrole for the application of the form `{app-name}-adminrole`. CloudFormation can't create two roles with the same name since roles are global.

Currently, CloudFormation errors out saying
failed to create: [AdministrationRole]
`stack {app-name}-infrastructure-roles did not complete successfully and exited with status ROLLBACK_COMPLETE`

This PR will do the following
1. Check to see if the application name  exists in SSM
2. If it doesn't, check to see if the role exists
3. If the role exists, but the application doesn't exist shows an error: `application named  {app-name} already exists in another region`
Case 1:
If you run `copilot app init test-project` in `us-east-1` and If run `copilot app init test-project `in `us-west-2` Then an error message shows `application named test-project already exists in another region`
Case 2: 
If there is already a role existing in the account with same name as `applicationname-adminrole` then we will error out `IAM admin role  {applicationname}-adminrole already exists in this account`


internal/pkg/cli/app_init.go
```diff
@@ -232,6 +234,14 @@ func (o *initAppOpts) validateAppName(name string) error {
 	if err != nil {
 		var noSuchAppErr *config.ErrNoSuchApplication
 		if errors.As(err, &noSuchAppErr) {
+			roleName := fmt.Sprintf("%s-adminrole", name)
+			tags, err := o.iamRoleManager.ListRoleTags(roleName)
+			if err != nil {
+				return nil
+			}
+			if _, hasTag := tags[deploy.AppTagKey]; hasTag {
+				return fmt.Errorf("application named %s already exists in another region", name)
```
paragbhingre: If we have region of the another application handy by any chance then can we also show that to the customer?

internal/pkg/cli/app_init.go
```diff
@@ -80,6 +81,7 @@ func newInitAppOpts(vars initAppVars) (*initAppOpts, error) {
 		prompt:           prompt.New(),
 		prog:             termprogress.NewSpinner(log.DiagnosticWriter),
 		iam:              iam.New(sess),
+		iamRoleManager:   iam.New(sess),
```
paragbhingre: Can we reuse the existing session instead of creating new one? for example -
```
                 iam := iam.New(sess),
                 iam: iam,
		 iamRoleManager:  iam,
```

paragbhingre (commented):
Overall looks good 😎 I just have a question and a suggestion.

paragbhingre:
> 1. Check to see if the application name exists in SSM 

Are we actually checking anywhere that application does exist in the SSM?

KollaAdithya:
>There already a function `func (s *Store) GetApplication(applicationName string) (*Application, error)` that fetches the params in ssm by application name.
https://github.com/KollaAdithya/copilot-cli/blob/mainline/internal/pkg/config/app.go#L77

paragbhingre:
> > There already a function `func (s *Store) GetApplication(applicationName string) (*Application, error)` that fetches the params in ssm by application name.
> > https://github.com/KollaAdithya/copilot-cli/blob/mainline/internal/pkg/config/app.go#L77

So just trying to understand the flow here. So currently we create admin-role and then add application to SSM? Is that why we are specifically checking if SSM doesn't have the application and admin-role exist before erroring out?

KollaAdithya:
> > > There already a function `func (s *Store) GetApplication(applicationName string) (*Application, error)` that fetches the params in ssm by application name.
> > > https://github.com/KollaAdithya/copilot-cli/blob/mainline/internal/pkg/config/app.go#L77
> 
> So just trying to understand the flow here. So currently we create admin-role and then add application to SSM? Is that why we are specifically checking if SSM doesn't have the application and admin-role exist before erroring out?

Yes, when you are trying to deploy an application we create `appname-adminrole` and add application to Parameter Store. That is the reason we are checking SSM. So if the app name does not exist in SSM but `appname-adminrole` already exists means there is already an application with same name in other region. In this case we are erroring out that there is already an application with this name in another region.

internal/pkg/cli/app_init.go
```diff
@@ -232,6 +234,14 @@ func (o *initAppOpts) validateAppName(name string) error {
 	if err != nil {
 		var noSuchAppErr *config.ErrNoSuchApplication
 		if errors.As(err, &noSuchAppErr) {
+			roleName := fmt.Sprintf("%s-adminrole", name)
+			tags, err := o.iamRoleManager.ListRoleTags(roleName)
+			if err != nil {
+				return nil
```
Lou1415926: nit: we can probably add a comment here stating why we swallow the error.

internal/pkg/cli/app_init.go
Lou1415926: Great move checking the Copilot tag!

Should we still error out even if they don't have the Copilot tag?  As long as the role exists (regardless whether it's created by Copi or not), Copilot won't be able to  create the application.

KollaAdithya: As per conversation with @paragbhingre and @Lou1415926. marking this as resolved

internal/pkg/cli/app_init.go
```diff
@@ -232,6 +235,15 @@ func (o *initAppOpts) validateAppName(name string) error {
 	if err != nil {
 		var noSuchAppErr *config.ErrNoSuchApplication
 		if errors.As(err, &noSuchAppErr) {
+			roleName := fmt.Sprintf("%s-adminrole", name)
+			tags, err := o.iamRoleManager.ListRoleTags(roleName)
+			// return nil when both adminrole and application does not exists then it is a valid app name
```
Lou1415926: Is "RoleNotFound" the only error that could be returned? What do you think of the following comment?
```go
// NOTE: This is a best-effort attempt to check if the app exists in other regions.
// The error either indicates that the role does not exist, or not. 
// In the first case, it means that this is a valid app name, hence we don't error out.
// In the second case, since this is a best-effort, we don't need to surface the error either.
```

internal/pkg/cli/app_init.go
```diff
@@ -232,6 +235,15 @@ func (o *initAppOpts) validateAppName(name string) error {
...
+			if err != nil {
+				return nil
+			}
+			if tags != nil {
+				return fmt.Errorf("application named %s already exists in another region", name)
```
Lou1415926: I think the error message should be different, depending whether there is an `copilot-application` tag. If there is such a tag, then the existing error message makes sense. Otherwise, it'd mean that there isn't an application - just the role - in which case the error message wouldn't make sense.

What do you think of the following? Does it overcomplicate it?

Example interface:
```go

type errAppAlreadyExistsInAccount struct
func (e *errAppAlreadyExistsInAccount) Error() string
func (e *errAppAlreadyExistsInAccount) RecommendedActions() string 
// I think our recommended actions here would be something like
// 1. If they just want to add a workspace, reusing that existing application, we will ask them to switch to that region, and run `copilot app init`.
// 2. If they think the existing application should be deleted, they need to switch to that region, and run `copilot app delete`.


type errStackSetAdminRoleExistsInAccount struct
func (e *errStackSetAdminRoleExistsInAccount) Error() string
func (e *errStackSetAdminRoleExistsInAccount) RecommendedActions() string 
// 1. We explain why that IAM role prevent application to create
// 2. We ask them to create the application with a different name, so that the IAM role name does not collide

// Here
if _, ok := tags[deploy.AppTag]; ok {
   return &errAppAlreadyExistsInAccount{}
}
return &errAppAlreadyExistsInAccount{}
```

internal/pkg/cli/app_init.go
```diff
@@ -53,11 +53,40 @@ type initAppOpts struct {
 	prompt               prompter
 	prog                 progress
 	iam                  policyLister
+	iamRoleManager       roleManager
 	isSessionFromEnvVars func() (bool, error)
 
 	cachedHostedZoneID string
 }
 
+type errAppAlreadyExistsInAccount struct {
```
Lou1415926: Can we move these errors to near the bottom of the file, just above `func buildAppInitCommand() *cobra.Command {`, i.e. around L325?

internal/pkg/cli/app_init.go
```diff
@@ -53,11 +53,40 @@ type initAppOpts struct {
...
+	appName string
+}
+
+type errStackSetAdminRoleExistsInAccount struct {
+	roleName string
+}
+
+func (e *errAppAlreadyExistsInAccount) Error() string {
+	return fmt.Sprintf("application named %s already exists in other region", e.appName)
```
Lou1415926:
```suggestion
	return fmt.Sprintf("application named %s already exists in another region", e.appName)
```

internal/pkg/cli/app_init.go
```diff
@@ -53,11 +53,40 @@ type initAppOpts struct {
...
+}
+
+func (e *errAppAlreadyExistsInAccount) RecommendActions() string {
+	return fmt.Sprintf(`
+- If you want to add a workspace, reusing that existing application %s, please switch to that region, and run %s.
```
Lou1415926:
```suggestion
- If you want to create a new workspace reusing the existing application %s, please switch to that region, and run %s.
```

internal/pkg/cli/app_init.go
```diff
@@ -53,11 +53,40 @@ type initAppOpts struct {
...
+- If you'd like to delete the application and all of its resources, please switch to that region, and run %s.`, e.appName, color.HighlightCode("copilot init"), color.HighlightCode("copilot app delete"))
```
Lou1415926:
```suggestion
- If you'd like to recreate the application and all of its resources, please switch to that region, and run %s.`, e.appName, color.HighlightCode("copilot init"), color.HighlightCode("copilot app delete"))
```

internal/pkg/cli/app_init.go
```diff
@@ -53,11 +53,40 @@ type initAppOpts struct {
...
+}
+
+func (e *errStackSetAdminRoleExistsInAccount) Error() string {
+	return fmt.Sprintf("IAM admin role %s already exists in this account", e.roleName)
+}
+
+func (e *errStackSetAdminRoleExistsInAccount) RecommendActions() string {
+	return fmt.Sprintf(`
+- We create a IAM admin role named %s as applicationname-adminrole in order to have unique application name across regions.
```
Lou1415926:
```suggestion
- Copilot will create an IAM admin role named %s to manage the stack set of the application %s. You have an existing role with the exact same name in your account, which will collide with the role that Copilot creates.
```

internal/pkg/cli/app_init.go
```diff
@@ -232,6 +263,22 @@ func (o *initAppOpts) validateAppName(name string) error {
...
+			if err != nil {
+				return nil
+			}
+			_, hasTag := tags[deploy.AppTagKey]
+			if hasTag {
+				return &errAppAlreadyExistsInAccount{appName: name}
+			}
+			if !hasTag {
+				return &errStackSetAdminRoleExistsInAccount{roleName: roleName}
+			}
```
Lou1415926: I think we can get rid of the `if !hasTag` since it'd always be true when it reaches here
```suggestion
			return &errStackSetAdminRoleExistsInAccount{roleName: roleName}
		
```

Lou1415926 (commented):
Looks good! Just some nits around the recommendations!

internal/pkg/cli/app_init.go
```diff
@@ -53,11 +53,21 @@ type initAppOpts struct {
 	prompt               prompter
 	prog                 progress
 	iam                  policyLister
+	iamRoleManager       roleManager
 	isSessionFromEnvVars func() (bool, error)
 
 	cachedHostedZoneID string
 }
 
+type errAppAlreadyExistsInAccount struct {
```
Lou1415926: Can we move the definition of the error struct as well? To just above their method definition (`Error` and `RecommendedActions`)

internal/pkg/cli/app_init.go
```diff
@@ -323,6 +350,27 @@ func (o *initAppOpts) askSelectExistingAppName(existingApps []*config.Applicatio
 	return nil
 }
 
+func (e *errAppAlreadyExistsInAccount) Error() string {
+	return fmt.Sprintf("application named %s already exists in another region", e.appName)
+}
+
+func (e *errAppAlreadyExistsInAccount) RecommendActions() string {
+	return fmt.Sprintf(`
+- If you want to create a new workspace reusing the existing application %s, please switch to that region, and run %s
```
Lou1415926:
```suggestion
- If you want to create a new workspace reusing the existing application %s, please switch to that region, and run %s.
```

internal/pkg/cli/app_init.go
```diff
@@ -323,6 +350,27 @@ func (o *initAppOpts) askSelectExistingAppName(existingApps []*config.Applicatio
...
+- If you'd like to recreate the application and all of its resources, please switch to that region, and run %s.`, e.appName, color.HighlightCode("copilot init"), color.HighlightCode("copilot app delete"))
```
Lou1415926:
```suggestion
- If you'd like to recreate the application and all of its resources, please switch to that region, and run %s.`, e.appName, color.HighlightCode("copilot app init"), color.HighlightCode("copilot app delete"))
```

internal/pkg/cli/app_init.go
```diff
@@ -323,6 +347,34 @@ func (o *initAppOpts) askSelectExistingAppName(existingApps []*config.Applicatio
...
+	return fmt.Sprintf("application named %s already exists in another region", e.appName)
+}
+
+func (e *errAppAlreadyExistsInAccount) RecommendActions() string {
+	return fmt.Sprintf(`- If you want to create a new workspace reusing the existing application %s, please switch to the region where you created the application, and run %s.
+- If you'd like to recreate the application and all of its resources, please switch to the region where you created the application, and run %s.`, e.appName, color.HighlightCode("copilot app init"), color.HighlightCode("copilot app delete"))
+}
+
+func (e *errStackSetAdminRoleExistsInAccount) Error() string {
+	return fmt.Sprintf("IAM admin role %s already exists in this account", e.roleName)
```
dannyrandall:
```suggestion
	return fmt.Sprintf("IAM admin role %q already exists in this account", e.roleName)
```

internal/pkg/cli/app_init.go
```diff
@@ -323,6 +347,34 @@ func (o *initAppOpts) askSelectExistingAppName(existingApps []*config.Applicatio
...
+	appName string
+}
+
+type errStackSetAdminRoleExistsInAccount struct {
+	appName  string
+	roleName string
+}
+
+func (e *errAppAlreadyExistsInAccount) Error() string {
+	return fmt.Sprintf("application named %s already exists in another region", e.appName)
```
dannyrandall:
```suggestion
	return fmt.Sprintf("application named %q already exists in another region", e.appName)
```

internal/pkg/cli/app_init.go
```diff
@@ -323,6 +347,34 @@ func (o *initAppOpts) askSelectExistingAppName(existingApps []*config.Applicatio
...
+}
+
+func (e *errStackSetAdminRoleExistsInAccount) Error() string {
+	return fmt.Sprintf("IAM admin role %s already exists in this account", e.roleName)
+}
+
+func (e *errStackSetAdminRoleExistsInAccount) RecommendActions() string {
+	return fmt.Sprintf(`- Copilot will create an IAM admin role named %s to manage the stack set of the application %s. 
+- You have an existing role with the exact same name in your account, which will collide with the role that Copilot creates.
+- Please create the application with a different name, so that the IAM role name does not collide.`, e.roleName, e.appName)
```
dannyrandall: i'm not sure the `-`'s make sense here, seems like just a single paragraph 🤔

internal/pkg/cli/app_init.go
```diff
@@ -79,7 +81,8 @@ func newInitAppOpts(vars initAppVars) (*initAppOpts, error) {
 		cfn:              cloudformation.New(sess, cloudformation.WithProgressTracker(os.Stderr)),
 		prompt:           prompt.New(),
 		prog:             termprogress.NewSpinner(log.DiagnosticWriter),
-		iam:              iam.New(sess),
+		iam:              iam,
+		iamRoleManager:   iam,
```
dannyrandall, line 85: why do we need `iamRoleManager` and can't just use `iam`?

internal/pkg/cli/app_init.go
```diff
@@ -232,7 +238,22 @@ func (o *initAppOpts) validateAppName(name string) error {
 	if err != nil {
 		var noSuchAppErr *config.ErrNoSuchApplication
 		if errors.As(err, &noSuchAppErr) {
-			return nil
+			roleName := fmt.Sprintf("%s-adminrole", name)
```
dannyrandall: since there's a lot of logic in the `if err != nil` case, i wonder if it makes more sense to handle the `err == nil` case _first_ to reduce the amount of nested code here 🤔

internal/pkg/cli/app_init.go
```diff
@@ -323,6 +347,34 @@ func (o *initAppOpts) askSelectExistingAppName(existingApps []*config.Applicatio
...
+	appName  string
+	roleName string
+}
+
+func (e *errAppAlreadyExistsInAccount) Error() string {
+	return fmt.Sprintf("application named %s already exists in another region", e.appName)
+}
+
+func (e *errAppAlreadyExistsInAccount) RecommendActions() string {
+	return fmt.Sprintf(`- If you want to create a new workspace reusing the existing application %s, please switch to the region where you created the application, and run %s.
```
iamhopaul123: Same for this one. Maybe we can ditch "-"

internal/pkg/cli/app_init.go
KollaAdithya: If I am correct, `iamRoleManager` has this method `ListRoleTags(string)` which fetches the tags of that IAM role.
https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/cli/interfaces.go#L594

dannyrandall: ah gotcha, different interfaces :) makes sense!

dannyrandall (approved):
lgtm!

internal/pkg/cli/app_init.go
```diff
@@ -136,6 +139,9 @@ https://aws.github.io/copilot-cli/docs/credentials/`)
 			log.Infoln(fmt.Sprintf(
 				"Your workspace is registered to application %s.",
 				color.HighlightUserInput(summary.Application)))
+			if err := o.validateAppName(summary.Application); err != nil {
```
Lou1415926: Great!! Thanks for remember this place!

internal/pkg/cli/app_init.go
```diff
@@ -302,6 +323,9 @@ func (o *initAppOpts) askAppName(formatMsg string) error {
 	if err != nil {
 		return fmt.Errorf("prompt get application name: %w", err)
 	}
+	if err := o.validateAppName(appName); err != nil {
```
Lou1415926, line 344: nit: there are two `validateAppName`, one is a method of `initAppOpts`, the other is a function with no receiver. Can we rename the latter one to `validateAppNameString` or `validateAppNaming`?

Lou1415926 (approved):
The tests LGTM but why is the coverage dropping 😭

PR #4117, opened 2022-10-27
chore: arps describe

dannyrandall:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -58,3 +58,5 @@ require (
 	google.golang.org/protobuf v1.28.0 // indirect
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
 )
+
+replace github.com/aws/aws-sdk-go => /Users/dnrnd/Downloads/aws-sdk-go
--- a/go.mod
+++ b/go.mod
@@ -58,5 +58,3 @@ require (
 	google.golang.org/protobuf v1.28.0 // indirect
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
 )
-
-replace github.com/aws/aws-sdk-go => /Users/dnrnd/Downloads/aws-sdk-go
--- a/internal/pkg/aws/apprunner/apprunner.go
+++ b/internal/pkg/aws/apprunner/apprunner.go
@@ -5,6 +5,7 @@
 package apprunner
 
 import (
+	"context"
 	"fmt"
 	"regexp"
 	"sort"
@@ -13,6 +14,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/apprunner"
 )
@@ -42,6 +44,7 @@ type api interface {
 	ResumeService(input *apprunner.ResumeServiceInput) (*apprunner.ResumeServiceOutput, error)
 	StartDeployment(input *apprunner.StartDeploymentInput) (*apprunner.StartDeploymentOutput, error)
 	DescribeObservabilityConfiguration(input *apprunner.DescribeObservabilityConfigurationInput) (*apprunner.DescribeObservabilityConfigurationOutput, error)
+	DescribeVpcIngressConnectionWithContext(ctx aws.Context, input *apprunner.DescribeVpcIngressConnectionInput, opts ...request.Option) (*apprunner.DescribeVpcIngressConnectionOutput, error)
 }
 
 // AppRunner wraps an AWS AppRunner client.
@@ -51,6 +54,8 @@ type AppRunner struct {
 
 // New returns a Service configured against the input session.
 func New(s *session.Session) *AppRunner {
+	s.Config.Region = aws.String("us-east-1")
+	s.Config.Endpoint = aws.String("https://fusion.gamma.us-east-1.bullet.aws.dev/")
 	return &AppRunner{
 		client: apprunner.New(s),
 	}
@@ -211,6 +216,18 @@ func (a *AppRunner) WaitForOperation(operationId, svcARN string) error {
 		}
 		time.Sleep(3 * time.Second)
 	}
+}
+
+// PrivateURL returns the url associated with a VPC Ingress Connection.
+func (a *AppRunner) PrivateURL(ctx context.Context, vicArn string) (string, error) {
+	resp, err := a.client.DescribeVpcIngressConnectionWithContext(ctx, &apprunner.DescribeVpcIngressConnectionInput{
+		VpcIngressConnectionArn: aws.String(vicArn),
+	})
+	if err != nil {
+		return "", fmt.Errorf("describe vpc ingress connection %q: %w", vicArn, err)
+	}
+
+	return aws.StringValue(resp.VpcIngressConnection.DomainName), nil
 }
 
 // ParseServiceName returns the service name.
--- a/internal/pkg/aws/apprunner/apprunner.go
+++ b/internal/pkg/aws/apprunner/apprunner.go
@@ -215,12 +215,12 @@ func (a *AppRunner) WaitForOperation(operationId, svcARN string) error {
 }
 
 // PrivateURL returns the url associated with a VPC Ingress Connection.
-func (a *AppRunner) PrivateURL(vicArn string) (string, error) {
+func (a *AppRunner) PrivateURL(vicARN string) (string, error) {
 	resp, err := a.client.DescribeVpcIngressConnection(&apprunner.DescribeVpcIngressConnectionInput{
-		VpcIngressConnectionArn: aws.String(vicArn),
+		VpcIngressConnectionArn: aws.String(vicARN),
 	})
 	if err != nil {
-		return "", fmt.Errorf("describe vpc ingress connection %q: %w", vicArn, err)
+		return "", fmt.Errorf("describe vpc ingress connection %q: %w", vicARN, err)
 	}
 
 	return aws.StringValue(resp.VpcIngressConnection.DomainName), nil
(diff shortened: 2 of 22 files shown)


dannyrandall:
Build will fail for this PR due to https://github.com/aws/copilot-cli/pull/4117/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R62. Merging into a non-mainline branch, so tests will pass before merging into mainline!

go.mod
```diff
@@ -58,3 +58,5 @@ require (
 	google.golang.org/protobuf v1.28.0 // indirect
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
 )
+
+replace github.com/aws/aws-sdk-go => /Users/dnrnd/Downloads/aws-sdk-go
```
iamhopaul123: Is this mostly to avoid committing the custom SDK? It fails all the CI for testing tho.

dannyrandall: yeah i was trying to avoid that - i figure i'll have to make the tests pass when it's merged into mainline with the updated SDK anyways

dannyrandall: the tests pass locally 😊

internal/pkg/describe/lb_web_service.go
```diff
@@ -287,6 +287,7 @@ func underline(headings []string) []string {
 type WebServiceRoute struct {
 	Environment string `json:"environment"`
 	URL         string `json:"url"`
+	Public      *bool  `json:"public,omitempty"`
```
iamhopaul123: Do we have three different states for this field? If not it can probably be `bool`?

iamhopaul123 (commented):
LGTM! <strike>TBH I'd rather committing the new pkg as a code reviewer...</strike> Actually it's ok. As long as it'll be only for this PR.

iamhopaul123 (approved):

internal/pkg/describe/service.go
```diff
@@ -294,17 +316,39 @@ func (d *appRunnerServiceDescriber) Service() (*apprunner.Service, error) {
...
 func (d *appRunnerServiceDescriber) ServiceURL() (string, error) {
-	service, err := d.Service()
+	vicArn, err := d.vpcIngressConnectionARN()
 	if err != nil {
 		return "", err
 	}
 
-	return formatAppRunnerUrl(service.ServiceURL), nil
+	if vicArn != nil {
+		url, err := d.apprunnerClient.PrivateURL(context.Background(), *vicArn)
```
iamhopaul123: This doesn't look to be the right place for creating a context. My suggestion is still we use ctx when it is *required* it'll be less painful when we switch to sdk v2.

internal/pkg/aws/apprunner/apprunner.go
```diff
@@ -51,6 +54,8 @@ type AppRunner struct {
 
 // New returns a Service configured against the input session.
 func New(s *session.Session) *AppRunner {
+	s.Config.Region = aws.String("us-east-1")
+	s.Config.Endpoint = aws.String("https://fusion.gamma.us-east-1.bullet.aws.dev/")
```
Lou1415926: Does not feel super right exposing these endpoints : ( . What do you think of `cherry-pick` or `rebase` locally when you run the tests?

internal/pkg/describe/service.go
```diff
@@ -280,6 +286,22 @@ func (d *appRunnerServiceDescriber) ServiceARN() (string, error) {
 	return "", fmt.Errorf("no App Runner Service in service stack")
 }
 
+func (d *appRunnerServiceDescriber) vpcIngressConnectionARN() (*string, error) {
```
Lou1415926: nit: why do we return pointer? does `""` mean different thing than `nil` here?

internal/pkg/describe/service.go
```diff
@@ -280,6 +286,22 @@ func (d *appRunnerServiceDescriber) ServiceARN() (string, error) {
...
+	serviceStackResources, err := d.ServiceStackResources()
```
Lou1415926, line 292: Ahh perhaps we should cache some information to reduce the amount of API calls. I see `IsPrivate` getting called a lot

internal/pkg/describe/lb_web_service.go
dannyrandall: We do have a third - since LBWS uses the same object. I didn't want to add a new field to lbws since it's inherently "public"

internal/pkg/describe/service.go
dannyrandall: `ServiceStackResources` caches the list of resources :)

internal/pkg/aws/apprunner/apprunner.go
```diff
@@ -213,6 +218,18 @@ func (a *AppRunner) WaitForOperation(operationId, svcARN string) error {
 	}
 }
 
+// PrivateURL returns the url associated with a VPC Ingress Connection.
+func (a *AppRunner) PrivateURL(ctx context.Context, vicArn string) (string, error) {
```
efekarakus: nit: https://github.com/golang/go/wiki/CodeReviewComments#initialisms
```suggestion
func (a *AppRunner) PrivateURL(ctx context.Context, vicARN string) (string, error) {
```

internal/pkg/describe/lb_web_service.go
efekarakus: What if this was an `enum`? where the type instead of a `bool` is a `string` and the field name is `"ingress"`?

I'm thinking of a scenario where maybe the url is accessible within an environment vs. a shared VPC vs. public

internal/pkg/describe/rd_web_service.go
```diff
@@ -238,11 +246,11 @@ func (w *rdWebSvcDesc) HumanString() string {
 	}
 	fmt.Fprint(writer, color.Bold.Sprint("\nRoutes\n\n"))
 	writer.Flush()
-	headers := []string{"Environment", "URL"}
+	headers := []string{"Environment", "Public", "URL"}
```
efekarakus: Samesies here, what do you think of "Ingress" as the column name with the values ("public" or "private (environment)") for example

internal/pkg/describe/service.go
efekarakus: I think it's possible that it's nil if there is no `VpcIngressConnection` resource, but to address Wanxian's feedback. I wonder if this method signature is a bit more explicit?
```suggestion
// vpcIngressConnectionARN returns an errResourceNotFound error if the vpc ingress does not exist.
func (d *appRunnerServiceDescriber) vpcIngressConnectionARN() (string, error) {
```
We can return a non-pointer, but instead return a custom error `errResourceNotFound` if it doesn't exist.
Then the caller has to check if this is the error returned and skip or whatever

internal/pkg/describe/service.go
dannyrandall: discussed offline, i'll just use the non-context func for now

internal/pkg/aws/apprunner/apprunner.go
dannyrandall: going back and forth between cfn and go is hard 😓

efekarakus:
oh can we remove the output now from the CFN stack, now that you can display the URL in `svc show`?

go.mod
dannyrandall: actually just going to remove the go.mod from the PR, should make some tests run at least. this package won't build and neither will the bin though.

internal/pkg/describe/errors.go
```diff
@@ -31,3 +31,9 @@ func (err *errLBWebSvcsOnCFWithoutAlias) Error() string {
 	return fmt.Sprintf("%s %s must have %q specified when CloudFront is enabled", english.PluralWord(len(err.services), "service", "services"),
 		english.WordSeries(template.QuoteSliceFunc(err.services), "and"), err.aliasField)
 }
+
+type errVPCIngressConnectionNotFound struct{}
+
+func (e *errVPCIngressConnectionNotFound) Error() string {
+	return "no vpc ingress connection found"
+}
```
efekarakus: nit: 
```suggestion
var errVPCIngressConnectionNotFound = errors.New("no vpc ingress connection found")
```

internal/pkg/describe/service.go
```diff
@@ -294,17 +318,46 @@ func (d *appRunnerServiceDescriber) Service() (*apprunner.Service, error) {
 	return service, nil
 }
 
+// IsPrivate returns true if the service is configured as non-public.
+func (d *appRunnerServiceDescriber) IsPrivate() (bool, error) {
+	_, err := d.vpcIngressConnectionARN()
+	if err != nil {
+		var errNotFound *errVPCIngressConnectionNotFound
+		if errors.As(err, &errNotFound) {
```
efekarakus: nit: with the sentinel error we can use `if errors.Is(err, errVPCIngressConnectionNotFound)`  (same in the `ServiceURL` method)

efekarakus (approved):
Just two tiny refactors, the staticcheck error in `cli/svc_deploy.go` l.263 is pretty odd

dannyrandall:
yeah that is a weird error...i don't see the same thing when running `golangci-lint --enable-all ./...` so i'm guessing it's just something weird bc the build is failing

internal/pkg/describe/service.go
```diff
@@ -318,23 +320,28 @@ func (d *appRunnerServiceDescriber) Service() (*apprunner.Service, error) {
...
-	return vicArn != nil, nil
+	return true, nil
 }
 
 // ServiceURL retrieves the app runner service URL.
 func (d *appRunnerServiceDescriber) ServiceURL() (string, error) {
-	vicArn, err := d.vpcIngressConnectionARN()
-	if err != nil {
+	vicARN, err := d.vpcIngressConnectionARN()
+	isErrNotFound := errors.Is(err, errVPCIngressConnectionNotFound)
```
Lou1415926:
```suggestion
	isVICNotFound := errors.Is(err, errVPCIngressConnectionNotFound)
```
hopefully this makes it clearer in L343 that vicFound => private => private URL 🤔 ?

Lou1415926 (approved):
LGTM!!! Adding label in case you'd like to address the nit but otherwise feel free to remove it anytime!

PR #4118, opened 2022-10-27
chore: arps `svc init` prompt

dannyrandall:

### DIFF
--- a/internal/pkg/cli/flag.go
+++ b/internal/pkg/cli/flag.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/aws/copilot-cli/internal/pkg/manifest"
+	"github.com/dustin/go-humanize/english"
 )
 
 // Long flag names.
@@ -201,6 +202,9 @@ Mutually exclusive with the -%s ,--%s and --%s flags.`, nameFlagShort, nameFlag,
 
 	repoURLFlagDescription = fmt.Sprintf(`The repository URL to trigger your pipeline.
 Supported providers are: %s.`, strings.Join(manifest.PipelineProviders, ", "))
+
+	ingressTypeFlagDescription = fmt.Sprintf(`Required for a Request-Driven Web Service. Allowed source of traffic to your service.
+Must be one of %s`, english.OxfordWordSeries(rdwsIngressOptions, "or"))
 )
 
 const (
@@ -265,9 +269,6 @@ Defaults to all logs. Only one of end-time / follow may be used.`
 	localPipelineFlagDescription     = "Only show pipelines in the workspace."
 	deleteSecretFlagDescription      = "Deletes AWS Secrets Manager secret associated with a pipeline source repository."
 	svcPortFlagDescription           = "The port on which your service listens."
-
-	ingressTypeFlagDescription = `Required for a Request-Driven Web Service. Allowed source of traffic to your service.
-Must be "environment" or "internet."`
 
 	noSubscriptionFlagDescription  = "Optional. Turn off selection for adding subscriptions for worker services."
 	subscribeTopicsFlagDescription = `Optional. SNS Topics to subscribe to from other services in your application.
--- a/internal/pkg/cli/flag.go
+++ b/internal/pkg/cli/flag.go
@@ -64,6 +64,8 @@ const (
 	noSubscriptionFlag  = "no-subscribe"
 	subscribeTopicsFlag = "subscribe-topics"
 
+	ingressTypeFlag = "ingress-type"
+
 	storageTypeFlag              = "storage-type"
 	storagePartitionKeyFlag      = "partition-key"
 	storageSortKeyFlag           = "sort-key"
@@ -263,6 +265,8 @@ Defaults to all logs. Only one of end-time / follow may be used.`
 	localPipelineFlagDescription     = "Only show pipelines in the workspace."
 	deleteSecretFlagDescription      = "Deletes AWS Secrets Manager secret associated with a pipeline source repository."
 	svcPortFlagDescription           = "The port on which your service listens."
+
+	ingressTypeFlagDescription = "Required for Request-Driven Web Service. Who can send traffic to your service."
 
 	noSubscriptionFlagDescription  = "Optional. Turn off selection for adding subscriptions for worker services."
 	subscribeTopicsFlagDescription = `Optional. SNS Topics to subscribe to from other services in your application.
--- a/internal/pkg/cli/flag.go
+++ b/internal/pkg/cli/flag.go
@@ -266,7 +266,7 @@ Defaults to all logs. Only one of end-time / follow may be used.`
 	deleteSecretFlagDescription      = "Deletes AWS Secrets Manager secret associated with a pipeline source repository."
 	svcPortFlagDescription           = "The port on which your service listens."
 
-	ingressTypeFlagDescription = "Required for Request-Driven Web Service. Who can send traffic to your service."
+	ingressTypeFlagDescription = "Required for a Request-Driven Web Service. Allowed source of traffic to your service."
 
 	noSubscriptionFlagDescription  = "Optional. Turn off selection for adding subscriptions for worker services."
 	subscribeTopicsFlagDescription = `Optional. SNS Topics to subscribe to from other services in your application.
--- a/internal/pkg/cli/flag.go
+++ b/internal/pkg/cli/flag.go
@@ -266,7 +266,8 @@ Defaults to all logs. Only one of end-time / follow may be used.`
 	deleteSecretFlagDescription      = "Deletes AWS Secrets Manager secret associated with a pipeline source repository."
 	svcPortFlagDescription           = "The port on which your service listens."
 
-	ingressTypeFlagDescription = "Required for a Request-Driven Web Service. Allowed source of traffic to your service."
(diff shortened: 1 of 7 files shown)


internal/pkg/cli/svc_init.go
```diff
@@ -712,6 +747,7 @@ This command is also run as part of "copilot init".`,
 	cmd.Flags().Uint16Var(&vars.port, svcPortFlag, 0, svcPortFlagDescription)
 	cmd.Flags().StringArrayVar(&vars.subscriptions, subscribeTopicsFlag, []string{}, subscribeTopicsFlagDescription)
 	cmd.Flags().BoolVar(&vars.noSubscribe, noSubscriptionFlag, false, noSubscriptionFlagDescription)
+	cmd.Flags().StringVar(&vars.ingressType, ingressTypeFlag, "", ingressTypeFlagDescription)
```
Lou1415926, line 767: What do you think of setting the default to "Environment"? There are a bunch of script users and this change will break their script.

internal/pkg/cli/flag.go
```diff
@@ -264,6 +266,8 @@ Defaults to all logs. Only one of end-time / follow may be used.`
 	deleteSecretFlagDescription      = "Deletes AWS Secrets Manager secret associated with a pipeline source repository."
 	svcPortFlagDescription           = "The port on which your service listens."
 
+	ingressTypeFlagDescription = "Required for Request-Driven Web Service. Who can send traffic to your service."
```
efekarakus: nit: like discussed in the bug bash can we list the available options here 🥺

internal/pkg/cli/svc_init.go
```diff
@@ -253,6 +262,9 @@ func (o *initSvcOpts) Ask() error {
 	if err := o.validateSvc(); err != nil {
 		return err
 	}
+	if err := o.askIngressType(); err != nil {
```
efekarakus, line 276: I feel like the #dream setup would be separate commands `initLBWSvcOpts`, `initBackendSvcOpts`, `initWorkerSvcOpts`, `initRDWSvcOpts` that all implement the same `Validate(), Ask(), Execute()` interface and this command just does the switchin based on the type

efekarakus: 2. Do we not need to update `Validate` to ensure the customer's input is one of the valid values?

internal/pkg/template/templates/workloads/services/rd-web/manifest.yml
```diff
@@ -19,24 +19,40 @@ image:
 {{- end}}
   # Port exposed through your container to route traffic to it.
   port: {{.ImageConfig.Port}}
-
+{{if and .Private.IsBasic .Private.Basic}}
```
efekarakus: wut 💭 this is strange to read, why not:
```suggestion
{{if  .Private.IsBasic }}
```

internal/pkg/cli/flag.go
huanjani: This wording isn't quite right, either, but may be slightly clearer?
```suggestion
	ingressTypeFlagDescription = "Required for a Request-Driven Web Service. Allowed source of traffic to your service."
```

internal/pkg/cli/svc_init.go
```diff
@@ -82,9 +82,17 @@ You should set this to the port which your Dockerfile uses to communicate with t
 	svcInitPublisherHelpPrompt = `A publisher is an existing SNS Topic to which a service publishes messages. 
 These messages can be consumed by the Worker Service.`
 
+	svcInitIngressTypePrompt     = "Would you like to accept traffic from your environment or the internet?"
+	svcInitIngressTypeHelpPrompt = "Environment will configure your service as private.\nInternet will configure your service as public."
```
huanjani:
```suggestion
	svcInitIngressTypeHelpPrompt = "'Environment' will configure your service as private.\n'Internet' will configure your service as public."
```

internal/pkg/cli/svc_init.go
huanjani: Normally I'd say the most private option should be the default, but should "Internet" be the default in this case because it's currently the only option?

internal/pkg/cli/svc_init_test.go
```diff
@@ -354,6 +354,44 @@ type: Request-Driven Web Service`), nil)
 			},
 			wantedErr: fmt.Errorf("service name iamoverfortycharacterlongandaninvalidrdwsname is invalid: value must not exceed 40 characters"),
 		},
+		"rdws prompt for ingress type": {
```
huanjani, line 388: Do we need a test with ingress type passed in via flag?

internal/pkg/cli/svc_init.go
dannyrandall: if we have default, are you thinking to remove the prompt..?

internal/pkg/cli/flag.go
```diff
@@ -264,6 +266,9 @@ Defaults to all logs. Only one of end-time / follow may be used.`
 	deleteSecretFlagDescription      = "Deletes AWS Secrets Manager secret associated with a pipeline source repository."
 	svcPortFlagDescription           = "The port on which your service listens."
 
+	ingressTypeFlagDescription = `Required for a Request-Driven Web Service. Allowed source of traffic to your service.
+Must be "environment" or "internet."`
```
efekarakus: nit:
```suggestion
	ingressTypeFlagDescription = fmt.Sprintf(`Required for a Request-Driven Web Service. Allowed source of traffic to your service.
Must be one of %s`, english.WordSeries(rdwsIngressOptions, "or"))
```

internal/pkg/cli/svc_init.go
```diff
@@ -389,6 +407,38 @@ func (o *initSvcOpts) askSvcName() error {
...
+}
+
+func (o *initSvcOpts) validateIngressType() error {
+	if o.wkldType != manifest.RequestDrivenWebServiceType {
+		return nil
+	}
+	if strings.EqualFold(o.ingressType, "internet") || strings.EqualFold(o.ingressType, "environment") {
+		return nil
+	}
+	return fmt.Errorf("invalid ingress type %q", o.ingressType)
```
efekarakus: Can we list the valid options in the error message to help users figure out what should be provided?

internal/pkg/cli/svc_init.go
```diff
@@ -253,6 +264,12 @@ func (o *initSvcOpts) Ask() error {
 	if err := o.validateSvc(); err != nil {
 		return err
 	}
+	if err := o.askIngressType(); err != nil {
+		return err
+	}
+	if err := o.validateIngressType(); err != nil {
+		return err
+	}
```
efekarakus: should this call be in the `Validate()` method rather than `Ask`? I think after `askIngressType` it's guaranteed that the user selected a valid option from the prompt, but in `Validate` we check the flag value

internal/pkg/template/templates/workloads/services/rd-web/manifest.yml
```diff
@@ -19,24 +19,40 @@ image:
 {{- end}}
   # Port exposed through your container to route traffic to it.
   port: {{.ImageConfig.Port}}
-
+{{if .Private.IsBasic}}
```
efekarakus, line 22: Was this on purpose?
```suggestion
{{- if .Private.IsBasic}}
```

internal/pkg/template/templates/workloads/services/rd-web/manifest.yml
```diff
@@ -19,24 +19,40 @@ image:
...
 #     interval: 10s
 #     timeout: 5s
+{{- end}}
 
 # Number of CPU units for the task.
 cpu: {{.InstanceConfig.CPU}}
 # Amount of memory in MiB used by the task.
 memory: {{.InstanceConfig.Memory}}
-
+{{if .Network.VPC.Placement.PlacementString}}
```
efekarakus, line 45: same here for the spacing

internal/pkg/template/templates/workloads/services/rd-web/manifest.yml
dannyrandall: yeah 🥺 otherwise it looks like
```
  port: 8080
http:
  private: true
```
and i wanted to keep the "section space" between `port`/`http`

internal/pkg/cli/svc_init.go
```diff
@@ -436,7 +438,7 @@ func (o *initSvcOpts) validateIngressType() error {
 	if strings.EqualFold(o.ingressType, "internet") || strings.EqualFold(o.ingressType, "environment") {
 		return nil
 	}
-	return fmt.Errorf("invalid ingress type %q", o.ingressType)
+	return fmt.Errorf("invalid ingress type %q. must be one of %s", o.ingressType, english.OxfordWordSeries(rdwsIngressOptions, "or"))
```
efekarakus: nit:
```suggestion
	return fmt.Errorf("invalid ingress type %q: must be one of %s", o.ingressType, english.OxfordWordSeries(rdwsIngressOptions, "or"))
```

efekarakus (approved):

internal/pkg/cli/svc_init.go
dannyrandall: discussed offline, we'll leave the prompt for now. script users will have to add the flag.

Lou1415926 (approved):
let's go!!

PR #4121, opened 2022-10-27
chore: aurora v2 use correct port for db cluster

iamhopaul123:
> The serverless engine mode only applies for Aurora Serverless v1 DB clusters.

https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBCluster.html

```
Port
The port number on which the DB instances in the DB cluster accept connections.

Default:

When EngineMode is provisioned, 3306 (for both Aurora MySQL and Aurora PostgreSQL)

When EngineMode is serverless:

3306 when Engine is aurora or aurora-mysql

5432 when Engine is aurora-postgresql
```
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html#cfn-rds-dbcluster-port

### DIFF
--- a/internal/pkg/template/templates/addons/aurora/rdws/serverlessv2.yml
+++ b/internal/pkg/template/templates/addons/aurora/rdws/serverlessv2.yml
@@ -135,6 +135,7 @@ Resources:
       {{- end}}
       DBClusterParameterGroupName: {{- if .ParameterGroup}} {{.ParameterGroup}} {{- else}} !Ref {{logicalIDSafe .ClusterName}}DBClusterParameterGroup {{- end}}
       DBSubnetGroupName: !Ref {{logicalIDSafe .ClusterName}}DBSubnetGroup
+      Port: {{if eq .Engine "MySQL"}}3306{{else}}5432{{end}}
       VpcSecurityGroupIds:
         - !Ref {{logicalIDSafe .ClusterName}}DBClusterSecurityGroup
       ServerlessV2ScalingConfiguration:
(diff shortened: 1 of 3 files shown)


internal/pkg/template/templates/addons/aurora/rdws/serverlessv2.yml
```diff
@@ -135,6 +135,7 @@ Resources:
       {{- end}}
       DBClusterParameterGroupName: {{- if .ParameterGroup}} {{.ParameterGroup}} {{- else}} !Ref {{logicalIDSafe .ClusterName}}DBClusterParameterGroup {{- end}}
       DBSubnetGroupName: !Ref {{logicalIDSafe .ClusterName}}DBSubnetGroup
+      Port: {{if eq .Engine "MySQL"}}3306{{else}}5432{{end}}
```
efekarakus, line 138: wut but from the commit description it looks like RDS does this by default? what am I misunderstanding?

iamhopaul123: Commit msg is a lie...

iamhopaul123: Basically we have two ways of fixing the problem: change the db cluster port or change the sg port. The first one makes more sense because 5432 is the default port for postgre, instead of making it always to be 3306.

efekarakus: > Instead of making it always to be 3306.

I see what you're saying that right now regardless of the engine being postgres or mysql, it defaults to 3306.

Whereas it should have defaulted to 5432 if it was postgres. Since it doesn't do that, you have to explicitly assign it.

Am I understanding correctly?

efekarakus (approved):

internal/pkg/template/templates/addons/aurora/rdws/serverlessv2.yml
iamhopaul123: Yeah precisely. It should work regardless of how they change the default port on the service side.

huanjani (approved):

PR #4136, opened 2022-10-31
chore: don't create VPC Endpoint if endpoint specified

dannyrandall:

### DIFF
--- a/internal/pkg/template/workload.go
+++ b/internal/pkg/template/workload.go
@@ -735,7 +735,7 @@ func envControllerParameters(o WorkloadOpts) []string {
 		}
 	}
 	if o.WorkloadType == "Request-Driven Web Service" {
-		if o.Private {
+		if o.Private && o.AppRunnerVPCEndpoint == nil {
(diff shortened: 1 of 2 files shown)


efekarakus (approved):

internal/pkg/template/workload.go
```diff
@@ -735,7 +735,7 @@ func envControllerParameters(o WorkloadOpts) []string {
 		}
 	}
 	if o.WorkloadType == "Request-Driven Web Service" {
-		if o.Private {
+		if o.Private && o.AppRunnerVPCEndpoint == nil {
```
efekarakus, line 738: Do we not have any unit tests for the template functions? I feel like it would be nice to have a unit test for `envControllerParameters` (although the fn is private, it's really "public" as it's a template function)

Lou1415926 (approved):

internal/pkg/template/workload.go
dannyrandall: added!

PR #4139, opened 2022-10-31
chore: don't create app runner vpc endpoint if imported vpc

dannyrandall:
plus a few test fixes!

### DIFF
--- a/internal/pkg/template/templates/environment/cf.yml
+++ b/internal/pkg/template/templates/environment/cf.yml
@@ -581,7 +581,9 @@ Resources:
 {{include "lambdas" . | indent 2}}
 {{include "custom-resources" . | indent 2}}
 {{- end}}
+{{- if not .VPCConfig.Imported}}
 {{include "ar-vpc-connector" . | indent 2}}
+{{- end}}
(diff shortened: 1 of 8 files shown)


huanjani (approved):

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -581,7 +581,9 @@ Resources:
 {{include "lambdas" . | indent 2}}
 {{include "custom-resources" . | indent 2}}
 {{- end}}
+{{- if not .VPCConfig.Imported}}
```
Lou1415926, line 584: Is this impossible to support or do we just need time to decide whether to support it in the future?

Lou1415926 (commented):
LGTM just a small q!

internal/pkg/template/templates/environment/cf.yml
dannyrandall: i think it's similar to the NAT gateway handling - we _could_ support it, but since we don't know if customers have already deployed something in the imported VPC, we'll avoid creating one and have them be in charge of those environment resources that cost $.

Lou1415926 (approved):

PR #4147, opened 2022-11-01
chore: read environment addons from the workspace

Lou1415926:
This PR implements methods to read environment addons from the workspace in `copilot/environments/addons` folder.

Related: #4219 


internal/pkg/addon/addons.go
```diff
@@ -31,8 +31,8 @@ var (
 )
 
 type workspaceReader interface {
-	ReadAddonsDir(svcName string) ([]string, error)
-	ReadAddon(svcName, fileName string) ([]byte, error)
+	ReadWorkloadAddonsDir(svcName string) ([]string, error)
+	ReadWorkloadAddon(svcName, fileName string) ([]byte, error)
```
efekarakus: Instead of `ReadWorkloadAddonsDir` vs `ReadEnvAddonsDir`, what do you think of the following API?

```go
package workspace

// ListAddonsFiles returns a list of file paths to all the addons templates defined under the dir.
func (ws *Workspace) ListAddonsFiles(dirPath string) ([]string, error)

// ReadAddonsFile reads the addons template file under fpath.
func (ws *Workspace) ReadAddonsFile(fpath string) ([]byte, error)

// EnvAddonsPath returns the addons/ directory file path for environments.
func (ws *Workspace) EnvAddonsPath() string

// WorkloadAddonsPath returns the addons/ directory file path for a given workload.
func (ws *Workspace) WorkloadAddonsPath(name string) string
```
The idea is to have a generic `ListAddonsFiles` that can be given as input the `EnvAddonsPath` or `WorkloadAddonsPath`.

efekarakus (commented):
Thanks for the great test cases 

Let me know what you think of the suggestion below!

internal/pkg/addon/addons.go
Lou1415926: Addressed! 

I actually had:
```
WorkloadAddonsPath(name string) (string, error)
WorkloadAddonFilePath(wkldName, fName string) (string, error)
EnvAddonsPath() string
EnvAddonFilePath() string
ListFiles(dirPath string) ([]string, error)
ReadFile(fPath string) ([]byte, error)
```
So that we are still letting the `workspace` package to determine how an addon file is arranged under an `addons/` directory. lmk how you like it!

internal/pkg/addon/addons.go
```diff
@@ -81,7 +90,7 @@ func (s *Stack) Template() (string, error) {
 	return s.encode(s.template)
 }
 
-// Template returns Stack's CloudFormation parameters as a yaml string.
+// Parameters returns Stack's CloudFormation parameters as a yaml string.
```
dannyrandall, line 93: lol

internal/pkg/workspace/workspace.go
```diff
@@ -427,9 +458,16 @@ func (ws *Workspace) ReadAddonsDir(svcName string) ([]string, error) {
...
 }
 
-// ReadAddon returns the contents of a file under the service's "addons/" directory.
-func (ws *Workspace) ReadAddon(svc, fname string) ([]byte, error) {
-	return ws.read(svc, addonsDirName, fname)
+// ReadFile returns the content of a file.
+func (ws *Workspace) ReadFile(fPath string) ([]byte, error) {
+	exist, err := ws.fs.Exists(fPath)
+	if err != nil {
+		return nil, fmt.Errorf("check if manifest file %s exists: %w", fPath, err)
```
dannyrandall: should probably update this error since it could be reading a manifest file or an addon file now

dannyrandall (approved):
looks great, just a nit!

internal/pkg/workspace/workspace.go
```diff
@@ -409,15 +409,46 @@ func (ws *Workspace) DeleteWorkspaceFile() error {
 	return ws.fs.Remove(filepath.Join(CopilotDirName, SummaryFileName))
 }
 
-// ReadAddonsDir returns a list of file names under a service's "addons/" directory.
-func (ws *Workspace) ReadAddonsDir(svcName string) ([]string, error) {
+// EnvAddonsPath returns the addons/ directory file path for environments.
+func (ws *Workspace) EnvAddonsPath() (string, error) {
 	copilotPath, err := ws.copilotDirPath()
```
efekarakus, line 414: I wonder if we could have gotten-rid of so many of our errors if we had a different approach to `New` in our package. For example:
```go
package workspace

// Workspace is a created copilot/ directory in a git repository.
type Workspace struct{...}

// Use returns an existing workspace and replaces the old "New" function. 
// It searches a copilot/ directory from the current wd, up to 5 levels above.
func Use() (*Workspace, error) { ... }

// Create is no longer a method and instead is a function that creates a new *Workspace in the current working directory for appName.
func Create(appName string) (*Workspace, error)
```

With the above API, I think we can completely ditch the `copilotDirPath` error because `*Workspace` will have the value stored in a field.

Let me know what you think of the above, if it sounds good we can do it as a follow-up to this PR?

efekarakus (approved):

internal/pkg/workspace/workspace.go
Lou1415926: Makes sense to me!

We can use `Use()` to do what've been done by `copilotDirPath` so that the rest of the operations can proceed with the assumption that we are in a worksapce.

PR #4150, opened 2022-11-01
chore: private services env compatibility checker, only enable if enabled 🤦 

dannyrandall:


huanjani (dismissed):

internal/pkg/manifest/rd_web_svc.go
```diff
@@ -86,6 +86,11 @@ type RequestDrivenWebServiceHttpConfig struct {
 	Private                  Union[bool, VPCEndpoint] `yaml:"private"`
 }
 
+// PrivateEnabled returns true if r is configured for private ingress.
+func (r *RequestDrivenWebServiceHttpConfig) PrivateEnabled() bool {
+	return r.Private.Basic || !r.Private.Advanced.IsZero()
```
Lou1415926: ohh unrelated to this PR - does having `bool` for `r.Private.Basic` play nicely with environment override?
```
http:
   private: true

environments:
  test:
     # did not specify http.private
```
What is the result of this manifest for `test` env's `http.private`?

internal/pkg/manifest/rd_web_svc.go
```diff
@@ -186,6 +191,9 @@ func (s RequestDrivenWebService) applyEnv(envName string) (workloadManifest, err
 
 func (s *RequestDrivenWebService) requiredEnvironmentFeatures() []string {
 	var features []string
+	if s.PrivateEnabled() {
```
Lou1415926, line 190: ahhh this is annoying :homer-disappearing

We need a way to remind ourselves to update this place. On top of my head I'm thinking of this dumb tests...


```go
// In template pkg (unlikely to forget it here because of the `TestEnv_AvailableEnvFeatures` tests
var LBWSEnvironmentFeatures = []string{natgateway, efs, alb..}
var RDWSEnvironmentFeatures = []string{privateservice, natgateway, ...}
```
and here
```go
// In manifest/rd_web_svc_Test.go
func requiredEnvironmentFeatures(t *testing.T) {
   t.Run("all features enabled", func(t *testing.T) {
         svc := RequestDrivenWebService{
...
}
         features, err := svc.requiredEnvironmentFeatures()
         require.EqualElem(features, template.RDWSEnvironmentFeatures)
   })
} 
```

internal/pkg/manifest/rd_web_svc.go
dannyrandall: great question! this was a rabbit hole for me :) 

first issue: i forgot to add the new union type (`[bool, VPCEndpoint]`) to the list of transformers. (done)

second issue: i assumed that mergo would merge the bool values correctly, but then i started writing tests and found https://github.com/imdario/mergo/issues/89. Apparently mergo doesn't override dst with src if it's a zero value. (you already figured this out, with the `basicTransformer`😊 - but it only works for pointers) i fixed this for the union type by checking if the Union type thinks the value was set. Some examples of how it works (now!) - there's also some tests in `loader_test.go`: 
```
# 1 - will override
environments:
  test:
    http:
      private: true
# 2 - will override
environments:
  test:
    http:
      private: false
# 3 - will not override
environments:
  test:
    http:
      alias: "example.com"
# 4 - will not override
environments:
  test:
    cpu: 1024
```

internal/pkg/manifest/rd_web_svc.go
dannyrandall: i just added a comment - since even with the test way we're just hoping someone notices they need to update a variable ¯\_(ツ)_/¯ lmk if you'd prefer tests though!

internal/pkg/manifest/transform.go
```diff
@@ -305,13 +313,23 @@ func (t unionTransformer[Basic, Advanced]) Transformer(typ reflect.Type) func(ds
 			var zero Advanced
 			dstStruct.isAdvanced, dstStruct.Advanced = false, zero
 			dstStruct.isBasic = true
+
+			switch any(srcStruct.Basic).(type) {
+			case string, bool, int, int8, int16, int32, int64, uint8, uint16, uint32, uint64, float32, float64, complex64, complex128:
```
iamhopaul123: do we actually need `basicKinds`? Or we don't want array/slice.

internal/pkg/manifest/transform.go
```diff
@@ -305,13 +313,23 @@ func (t unionTransformer[Basic, Advanced]) Transformer(typ reflect.Type) func(ds
...
+				dstStruct.Basic = srcStruct.Basic
+			}
 		} else if srcStruct.IsAdvanced() {
 			var zero Basic
 			dstStruct.isBasic, dstStruct.Basic = false, zero
 			dstStruct.isAdvanced = true
+
+			switch any(srcStruct.Advanced).(type) {
+			case string, bool, int, int8, int16, int32, int64, uint8, uint16, uint32, uint64, float32, float64, complex64, complex128:
```
iamhopaul123: technically this doesn't look like an advanced type

internal/pkg/manifest/rd_web_svc.go
Lou1415926: I think this is good enough for this PR. Even if we do add tests, we will need to do it in a separate PR - was just throwing out ideas 💭

internal/pkg/manifest/transform.go
dannyrandall: i think there is a potential use case for "advanced" to be a string. for example, we could have a `Union[bool, string]` or `Union[int, string]`. Although we call it "advanced," it doesn't _have to_ be a struct.

internal/pkg/manifest/loader_test.go
```diff
@@ -90,3 +90,139 @@ func TestDynamicWorkloadManifest_Load(t *testing.T) {
...
+			},
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			mft, err := UnmarshalWorkload([]byte(tc.manifest))
+			require.NoError(t, err)
+
+			mft, err = mft.ApplyEnv("test")
```
efekarakus: is this the right file to put the test or should it be in `applyenv_test.go`?

internal/pkg/deploy/cloudformation/stack/rd_web_svc.go
```diff
@@ -141,7 +141,7 @@ func (s *RequestDrivenWebService) Template() (string, error) {
 			Tracing: strings.ToUpper(aws.StringValue(s.manifest.Observability.Tracing)),
 		},
 		PermissionsBoundary:  s.permBound,
-		Private:              !s.manifest.Private.IsZero(),
```
efekarakus: Would this work if we address this comment:
https://github.com/aws/copilot-cli/pull/4164#discussion_r1022136746

Is it worthwhile to cut that PR first before we merge this PR or #4164 given that we now have 2 places that would benefit from this change?

internal/pkg/manifest/rd_web_svc.go
```diff
@@ -186,6 +191,9 @@ func (s RequestDrivenWebService) applyEnv(envName string) (workloadManifest, err
 
 func (s *RequestDrivenWebService) requiredEnvironmentFeatures() []string {
```
efekarakus, line 188: This has nothing to do with your PR but it's pretty strange to me that this method: https://github.com/aws/copilot-cli/blob/5781c535bc78e5e3a544b855d930256562e52b63/internal/pkg/manifest/loader.go#L53

belongs to `*DynamicWorkloadManifest` 🤔 it doesn't seem relevant to a "dynamic" aspect. 
I'd have expected it to be part of the individual Manifest types.

internal/pkg/template/env.go
```diff
@@ -44,7 +48,7 @@ var leastVersionForFeature = map[string]string{
 	NATFeatureName:                     "v1.3.0",
 	InternalALBFeatureName:             "v1.10.0",
 	AliasesFeatureName:                 "v1.4.0",
-	AppRunnerPrivateServiceFeatureName: "v1.23.0",
+	AppRunnerPrivateServiceFeatureName: "v1.13.0",
```
efekarakus, line 53: can we add a comment to this variable to highlight if these versions are environment template versions or copilot versions

internal/pkg/manifest/transform.go
```diff
@@ -27,6 +27,7 @@ var defaultTransformers = []mergo.Transformers{
 	placementArgOrStringTransformer{},
 	subnetListOrArgsTransformer{},
 	unionTransformer[string, HTTPHealthCheckArgs]{},
+	unionTransformer[bool, VPCEndpoint]{},
```
efekarakus: hmm do we have to update this slice every time we use a `Union` type in the manifest? This worries me a little bit.

My brain is struggling retaining all the places that we need to make a change when adding a field now to the manifest:
1. Need to add validate
2. Need to add transformer? if not, why not?
3. Need to potentially update `requiredEnvironmentFeatures` if using environment controlled parameters.

I think I'm just worried that we don't have mechanisms to catch 2. and 3. like this PR resulting in bugs.
Either we need to think of a way of simplifying the process, or we need mechanisms catching gaps like local integ tests.

internal/pkg/manifest/transform.go
dannyrandall: removed! not needed after #4189

internal/pkg/manifest/transform.go
dannyrandall: resolved in separate PR! #4189

PR #4164, opened 2022-11-08
feat: Add log retention on VPC flow logs

### LINKED ISSUE #4162: Add log retention on VPC flow logs
Please add ability to add log retention in VPC logs.

KollaAdithya:
This PR resolves #4162  the retention for a log group of VPC FlowLog  i.e, The number of days to retain the log events in the specified log group.

If the flowlogs is `on`
```yaml
network:
  vpc:
    flow_logs: on
```

The default value for retention of log events in log group is set to 14 days[2 weeks].

If the customer want to have a specific retention rate then Environment Manifest can be modified to
```yaml
network:
  vpc:
    flow_logs:
      retention: 30
```
Here are the possible values for retention in days: [https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-loggroup.html](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-loggroup.html)


internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-defaultvpc-flowlogs.yml
```diff
@@ -1003,14 +1003,20 @@ Resources:
       DomainName: !Ref AppDNSName
       PublicAccessDNS: !GetAtt PublicLoadBalancer.DNSName
       PublicAccessHostedZone: !GetAtt PublicLoadBalancer.CanonicalHostedZoneID
+  LogGroup:
```
efekarakus: Can we rename the logical ID to be more specific to flow logs:
```suggestion
  VpcFlowLogGroup:
```

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-defaultvpc-flowlogs.yml
```diff
@@ -1003,14 +1003,20 @@ Resources:
...
+    Type: AWS::Logs::LogGroup
+    Metadata:
+      'aws:copilot:description': 'A  log group to create a flow log for the VPC'
```
efekarakus: nit: there are extra spaces
```suggestion
      'aws:copilot:description': 'A CloudWatch log group to store VPC flow log data'
```

Lou1415926 (commented):
I wonder if it's worthy doing something similar to [workload's log retention](https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#logging-retention): , that we give users the option to write:
```yaml
network:
  vpc:
    flow_log:
      retention: 21   
```
to specify their preferred retention, and if it is`networ.vpc.flow_log: true` we give it a default of `14` .

Just throwing this random thought here - perhaps no one needs to configure the retention and that a default of 14 is enough. What do yall think?

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -585,14 +585,20 @@ Resources:
 {{include "ar-vpc-connector" . | indent 2}}
 {{- end}}
 {{- if .VPCConfig.FlowLogs }}
+  LogGroup:
+    Type: AWS::Logs::LogGroup
+    Metadata:
+      'aws:copilot:description': 'A  log group to create a flow log for the VPC'
```
paragbhingre:
```suggestion
      'aws:copilot:description': 'A log group to create a flow log for the VPC'
```

paragbhingre: Another suggestion could be -
"A CloudWatch log group to hold your VPC flow logs"

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -585,14 +585,20 @@ Resources:
...
+    Properties:
+      RetentionInDays: 14
```
paragbhingre: I get that we are setting its default value to 14. But how will customers set their specific value of RetentionInDays?
Or have we just decided to set the default value to 14 and not let customers modify it?

paragbhingre (commented):
Looks good, just a nit and a question.

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-defaultvpc-flowlogs.yml
```diff
@@ -1003,14 +1003,20 @@ Resources:
...
+      RetentionInDays: 14
   FlowLog:
     Metadata:
       'aws:copilot:description': 'A flow log for the VPC to capture information about the IP traffic'
     Type: AWS::EC2::FlowLog
     Properties:
       DeliverLogsPermissionArn: !GetAtt FlowLogRole.Arn
       LogDestinationType: cloud-watch-logs
-      LogGroupName: !Join ['-', [!Ref AppName, !Ref EnvironmentName, FlowLogGroup]]
+      LogGroupName: !Ref LogGroup
```
efekarakus: I think this will be backwards incompatible for users that have already enabled flow logs using the environment manifest. The update requires a replacement: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-flowlog.html#cfn-ec2-flowlog-loggroupname

Can you test the following:
*  What happens after your change to the previous log group? is it deleted or retained. 
   * If retained, I think this is okay.
   * If not, then I wonder if there is a way of using the same log group name as before but changing it's retention.

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-defaultvpc-flowlogs.yml
```diff
@@ -1003,14 +1003,20 @@ Resources:
       DomainName: !Ref AppDNSName
       PublicAccessDNS: !GetAtt PublicLoadBalancer.DNSName
       PublicAccessHostedZone: !GetAtt PublicLoadBalancer.CanonicalHostedZoneID
+  LogGroup:
+    Type: AWS::Logs::LogGroup
+    Metadata:
+      'aws:copilot:description': 'A  log group to create a flow log for the VPC'
+    Properties:
+      RetentionInDays: 14
```
efekarakus: Should we assign a name here to make it easier to find the flow log group? What happens if the previous name is assigned here, I assume the call will fail

internal/pkg/template/templates/environment/cf.yml
gautam-nutalapati: +1 for customizable log retention period.
One use-case I can think of is log retention requirements due to compliance reasons.

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-defaultvpc-flowlogs.yml
KollaAdithya: 1. After changing previous log group name, it is still retained.
2. A new log group is created after the change to previous log group name.

Screen Shot 2022-11-09 at 3 08 17 PM

KollaAdithya: Now the log group name is going to be same as before `AppName-EnvironmentName-FlowLogGroup`

efekarakus: If it's the same name and a user had previously created the log group with:
```yaml
network:
 vpc:
   flow_logs: on
```
What happens?

 I assume CFN would fail saying the log group already exists 🤔 Would you mind testing that? 
 
 If it doesn't fail, perfect then retention was added successfully I assume?
 If it fails, we can just name the new log group `AppName-EnvironmentName-FlowLogs`

KollaAdithya: Oh yes , CFN fails saying that the log group already exists. So I created a new log group with the suggested name `AppName-EnvironmentName-FlowLogs`.

With this new log group creation we will have both log groups 
`AppName-EnvironmentName-FlowLogGroup` 
`AppName-EnvironmentName-FlowLogs`

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -433,12 +433,16 @@ func (e *EnvStackConfig) vpcConfig() (template.VPCConfig, error) {
 	if err != nil {
 		return template.VPCConfig{}, err
 	}
+	vpcFlowLogsArgs, err := convertFlowLogsConfig(e.in.Mft)
```
efekarakus: nit: I really like this guide for naming variables: https://a-nickels-worth.blogspot.com/2016/04/a-guide-to-naming-variables.html#orgheadline12. Take a look at the entire article if you have time.

Since this variable is declared close to where it's used I think it's ok if we have a shorter name:
```suggestion
	flowLogs, err := convertFlowLogsConfig(e.in.Mft)
```

internal/pkg/deploy/cloudformation/stack/env_integration_test.go
```diff
@@ -270,7 +270,8 @@ network:
       private:
         - id: 'subnet-33333'
         - id: 'subnet-44444'
-    flow_logs: on`
+    flow_logs:
+     retention: 60`
```
efekarakus: :rip: I kinda wish we made the unit more explicit.
Like accepting `60d` as input or `60`.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -381,6 +381,19 @@ func convertELBAccessLogsConfig(mft *manifest.Environment) (*template.ELBAccessL
 	}, nil
 }
 
+func convertFlowLogsConfig(mft *manifest.Environment) (*template.VPCFlowLogs, error) {
+	vpcFlowLogsArgs := mft.EnvironmentConfig.Network.VPC.FlowLogs.AdvancedFlowLogConfig
+	if !mft.VPCFlowLogsEnabled() {
+		return nil, nil
+	}
+	if vpcFlowLogsArgs.Retention == nil {
+		return &template.VPCFlowLogs{}, nil
```
efekarakus: What do you think about setting the default value here?
```suggestion
		return &template.VPCFlowLogs{
			RetentionInDays: 14,
		}, nil
```
This way you don't need an if-else statement in the Go template.

internal/pkg/manifest/env.go
```diff
@@ -119,11 +120,11 @@ type environmentNetworkConfig struct {
...
-	ID                  *string              `yaml:"id,omitempty"`
-	CIDR                *IPNet               `yaml:"cidr,omitempty"`
-	Subnets             subnetsConfiguration `yaml:"subnets,omitempty"`
-	SecurityGroupConfig securityGroupConfig  `yaml:"security_group,omitempty"`
-	Flowlogs            *bool                `yaml:"flow_logs,omitempty"`
+	ID                  *string               `yaml:"id,omitempty"`
+	CIDR                *IPNet                `yaml:"cidr,omitempty"`
+	Subnets             subnetsConfiguration  `yaml:"subnets,omitempty"`
+	SecurityGroupConfig securityGroupConfig   `yaml:"security_group,omitempty"`
+	FlowLogs            VPCFlowLogsArgsorBool `yaml:"flow_logs,omitempty"`
```
efekarakus: Can we leverage the [`Union` type](https://github.com/aws/copilot-cli/blob/5845a2abfc326641843bb81df286485e2c645665/internal/pkg/manifest/union.go#L14) that @dannyrandall introduced here?

Here are some example usage: https://github.com/aws/copilot-cli/blob/5845a2abfc326641843bb81df286485e2c645665/internal/pkg/manifest/rd_web_svc.go#L86

https://github.com/aws/copilot-cli/blob/5845a2abfc326641843bb81df286485e2c645665/internal/pkg/manifest/svc.go#L376-L386

internal/pkg/template/env.go
```diff
@@ -215,6 +215,10 @@ type SecurityGroupRule struct {
 	ToPort     int
 }
 
+type VPCFlowLogs struct {
```
efekarakus, line 219: https://github.com/golang/go/wiki/CodeReviewComments#doc-comments 🥺

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -382,15 +382,17 @@ func convertELBAccessLogsConfig(mft *manifest.Environment) (*template.ELBAccessL
...
-		return &template.VPCFlowLogs{}, nil
+	if vpcFlowLogs.Advanced.Retention == nil {
+		return &template.VPCFlowLogs{
+			Retention: aws.Int(14),
+		}, nil
 	}
 	return &template.VPCFlowLogs{
-		Retention: vpcFlowLogsArgs.Retention,
+		Retention: vpcFlowLogs.Advanced.Retention,
 	}, nil
```
```go
retentionInDays := aws.Int(14)
if vpcFlowLogs.Advanced.Retention != nil {
   retentionInDays = vpcFlowLogs.Advanced.Retention
}
return &template.VPCFlowLogs{
	Retention: retentionInDays,
}, nil
```

internal/pkg/manifest/env.go
```diff
@@ -120,11 +119,11 @@ type environmentNetworkConfig struct {
...
-	ID                  *string               `yaml:"id,omitempty"`
-	CIDR                *IPNet                `yaml:"cidr,omitempty"`
-	Subnets             subnetsConfiguration  `yaml:"subnets,omitempty"`
-	SecurityGroupConfig securityGroupConfig   `yaml:"security_group,omitempty"`
-	FlowLogs            VPCFlowLogsArgsorBool `yaml:"flow_logs,omitempty"`
+	ID                  *string                      `yaml:"id,omitempty"`
+	CIDR                *IPNet                       `yaml:"cidr,omitempty"`
+	Subnets             subnetsConfiguration         `yaml:"subnets,omitempty"`
+	SecurityGroupConfig securityGroupConfig          `yaml:"security_group,omitempty"`
+	FlowLogs            Union[bool, VPCFlowLogsArgs] `yaml:"flow_logs,omitempty"`
```
efekarakus: For `UnmarshalYAML` to work as expected @dannyrandall , do we need to implement the `IsZero` method for custom types like `VPCFLowLogsArgs`?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -382,15 +382,17 @@ func convertELBAccessLogsConfig(mft *manifest.Environment) (*template.ELBAccessL
 }
 
 func convertFlowLogsConfig(mft *manifest.Environment) (*template.VPCFlowLogs, error) {
-	vpcFlowLogsArgs := mft.EnvironmentConfig.Network.VPC.FlowLogs.AdvancedFlowLogConfig
-	if !mft.VPCFlowLogsEnabled() {
+	vpcFlowLogs := mft.EnvironmentConfig.Network.VPC.FlowLogs
+	if !vpcFlowLogs.Basic && vpcFlowLogs.Advanced.Retention == nil {
```
efekarakus: @dannyrandall I wodner if it's worthwhile updating the `Union` type's `IsZero` method? so that it returns true if the zero-value of the type is also true:
```go
	if t.IsBasic() {
		var zero Basic // NEW
		return isZeroerAndZero(t.Basic) ||  t.Basic == zero /* the conditional is new */
	}
	if t.IsAdvanced() {
		var zero Advanced
		return isZeroerAndZero(t.Advanced) || t.Advanced == zero
	}
	return true
```

this way we can replace this if statement with:
```go
if !vpcFlowLogs.IsZero() {
```

internal/pkg/template/env.go
```diff
@@ -215,6 +215,7 @@ type SecurityGroupRule struct {
 	ToPort     int
 }
 
+// VPCFlowLogs holds the fields to configure a FlowLog
```
efekarakus: nit: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences 
> Comments should begin with the name of the thing being described and end in a period
```suggestion
// VPCFlowLogs holds the fields to configure logging IP traffic using VPC flow logs.
```
😅

efekarakus (commented):
Sweet! Looks great! 

I have some clarification questions to @dannyrandall regarding the Union type and based on his feedback, we can then ship it!

internal/pkg/manifest/env.go
dannyrandall: chatted with @KollaAdithya offline, but probably not? reasoning:
If the manifest looks like this:
```
vpc:
  cidr: <config>
```
Then the yaml unmarshaler will never call `FlowLogs.UnmarshalYAML` because it's key isn't present. `FlowLogs` will just be a zero `Union` value, with `IsBasic() == false` and `IsAdvanced() == false`.

However, if it looks like
```
vpc:
  flow_logs:
    random_key: random_value
```
Then `IsAdvanced()` will be **true**, since there's no error unmarshaling that into `VPCFlowLogsArgs`. This is the weird case that you'd need `IsZero()` to handle.

dannyrandall: actually, thinking about this more - we should probably have `IsZero` to keep `IsAdvanced == false` in that second case.

in general, it'd be nice (i think) if we could have an error if a non-existent struct field is set 🤔

dannyrandall: ah, i remember why i didn't include that feature https://github.com/go-yaml/yaml/issues/460

internal/pkg/deploy/cloudformation/stack/transformers.go
dannyrandall: sadly this doesn't work right now - because the type constraint is `any`. i think if we changed it to `comparable` it would work - and i think that's probably fine since we don't want interface types anyways 🤔 

i'll make this change in a separate PR @KollaAdithya!

efekarakus: Sorry I just marked this as unresolved, I'm willing to give it a :shipit: with a do-not-merge label but I would love to see the `comparable` PR merged first so that this code is simplified before merging

internal/pkg/manifest/env.go
```diff
@@ -119,11 +119,11 @@ type environmentNetworkConfig struct {
...
-	ID                  *string              `yaml:"id,omitempty"`
-	CIDR                *IPNet               `yaml:"cidr,omitempty"`
-	Subnets             subnetsConfiguration `yaml:"subnets,omitempty"`
-	SecurityGroupConfig securityGroupConfig  `yaml:"security_group,omitempty"`
-	Flowlogs            *bool                `yaml:"flow_logs,omitempty"`
+	ID                  *string                      `yaml:"id,omitempty"`
+	CIDR                *IPNet                       `yaml:"cidr,omitempty"`
+	Subnets             subnetsConfiguration         `yaml:"subnets,omitempty"`
+	SecurityGroupConfig securityGroupConfig          `yaml:"security_group,omitempty"`
+	FlowLogs            Union[bool, VPCFlowLogsArgs] `yaml:"flow_logs,omitempty"`
```
efekarakus: I saw this change in @dannyrandall's PR:
https://github.com/aws/copilot-cli/pull/4150/files#diff-d5ce3a84f02de238dd26290c8a7fcf956134998ff644184f5d13a7e692eda1bbR30

I think we'd need one as well in `transform.go` but I'm not fully certain. I would love finding a way such taht using the `Union` type doesn't result in having to modify `transform.go`

efekarakus (approved):

iamhopaul123 (approved):

PR #4166, opened 2022-11-10
chore: implement new api for the `workspace` package

Lou1415926:
This PR implements a new API for the `workspace` package. With the new API, a workspace client that gets returned to the caller always have the `copilotDirAbs` field in it. This means that the method calls can assume non-empty value of that field.
This addresses the proposal made in https://github.com/aws/copilot-cli/pull/4147#discussion_r1017019403

### DIFF
--- a/internal/pkg/addon/addons.go
+++ b/internal/pkg/addon/addons.go
@@ -49,8 +49,7 @@ type Stack struct {
 // files found there. If no addons are found, Parse returns a nil
 // Stack and ErrAddonsNotFound.
 func Parse(workloadName string, ws workspaceReader) (*Stack, error) {
-	path := ws.WorkloadAddonsPath(workloadName)
-	fNames, err := ws.ListFiles(path)
+	fNames, err := ws.ListFiles(ws.WorkloadAddonsPath(workloadName))
 	if err != nil {
 		return nil, &ErrAddonsNotFound{
 			WlName:    workloadName,
--- a/internal/pkg/addon/addons.go
+++ b/internal/pkg/addon/addons.go
@@ -31,8 +31,8 @@ var (
 )
 
 type workspaceReader interface {
-	WorkloadAddonsPath(name string) (string, error)
-	WorkloadAddonFilePath(wkldName, fName string) (string, error)
+	WorkloadAddonsPath(name string) string
+	WorkloadAddonFilePath(wkldName, fName string) string
 	ListFiles(dirPath string) ([]string, error)
 	ReadFile(fPath string) ([]byte, error)
 }
@@ -49,13 +49,7 @@ type Stack struct {
 // files found there. If no addons are found, Parse returns a nil
 // Stack and ErrAddonsNotFound.
 func Parse(workloadName string, ws workspaceReader) (*Stack, error) {
-	path, err := ws.WorkloadAddonsPath(workloadName)
-	if err != nil {
-		return nil, &ErrAddonsNotFound{
-			WlName:    workloadName,
-			ParentErr: err,
-		}
-	}
+	path := ws.WorkloadAddonsPath(workloadName)
 	fNames, err := ws.ListFiles(path)
 	if err != nil {
 		return nil, &ErrAddonsNotFound{
@@ -128,10 +122,7 @@ func parseTemplate(fnames []string, workloadName string, ws workspaceReader) (*c
 
 	mergedTemplate := newCFNTemplate("merged")
 	for _, fname := range templateFiles {
-		path, err := ws.WorkloadAddonFilePath(workloadName, fname)
-		if err != nil {
-			return nil, fmt.Errorf("get addon %s file path under %s: %w", fname, workloadName, err)
-		}
+		path := ws.WorkloadAddonFilePath(workloadName, fname)
 		out, err := ws.ReadFile(path)
 		if err != nil {
 			return nil, fmt.Errorf("read addon %s under %s: %w", fname, workloadName, err)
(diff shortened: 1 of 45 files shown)


internal/pkg/workspace/workspace.go
```diff
@@ -72,65 +72,75 @@ type Workspace struct {
...
-// New returns a workspace, used for reading and writing to user's local workspace.
-func New() (*Workspace, error) {
-	fs := afero.NewOsFs()
-	fsUtils := &afero.Afero{Fs: fs}
-	logger := log.Infof
-
-	workingDirAbs, err := os.Getwd()
+// Use returns an existing workspace, searching for a copilot/ directory from the current wd,
+// up to 5 levels above. It returns ErrWorkspaceNotFound if no copilot/ directory is found.
+func Use(fs *afero.Afero, workingDirAbs string) (*Workspace, error) {
```
efekarakus: I didn't read the PR 😂 I thought I could leave this early feedback to see what you think:
1. Is it possible to use the [`afero.Fs` interface](https://pkg.go.dev/github.com/spf13/afero#Fs) as the input instead of the concrete type? or are we depending on some more specific functionality

2. Can we make the "workingDirAbs" optional? could the constructor by default use `os.Getwd` unless the user specified `workspace.Use(&afero.Afero{Fs: afero.NewOsFs()}, workspace.WithWorkingDir("some/other/path"))`
Would that simplify the life of the `cli` pkg?

Lou1415926: Addressed both!

The whole reason why we needed to pass `workingDir` as a param was for the unit tests - specifically, the ones that test that Copilot is able to find a workspace in a parent directory.  I fixed the unit tests such that it doesn't need to do so. This way, we don't need it as a param anymore, not even as an optional param.

internal/pkg/cli/env_package.go
```diff
@@ -92,7 +93,7 @@ func newPackageEnvOpts(vars packageEnvVars) (*packageEnvOpts, error) {
 		ws:           ws,
 		sel:          selector.NewLocalEnvironmentSelector(prompt.New(), cfgStore, ws),
 		caller:       identity.New(defaultSess),
-		fs:           &afero.Afero{Fs: afero.NewOsFs()},
+		fs:           fs,
```
Lou1415926, line 96: This are a couple of refactors like this that are unrelated to the main purpose of this PR.
 
The `fs` field of `packageEnvOpts` needs to just implement `afero.Fs`, and `afero.NewFs()` already does this. We don't need the other methods in `*afero.Afero`.

internal/pkg/addon/addons.go
```diff
@@ -49,13 +49,7 @@ type Stack struct {
...
 // Stack and ErrAddonsNotFound.
 func Parse(workloadName string, ws workspaceReader) (*Stack, error) {
-	path, err := ws.WorkloadAddonsPath(workloadName)
-	if err != nil {
-		return nil, &ErrAddonsNotFound{
-			WlName:    workloadName,
-			ParentErr: err,
-		}
-	}
+	path := ws.WorkloadAddonsPath(workloadName)
```
efekarakus: nit: now that we defined the error out of existence 💅 should we just inline these temp variables?
```suggestion
	ws.ListFiles(ws.WorkloadAddonsPath(workloadName))
```

internal/pkg/cli/deploy/svc.go
```diff
@@ -195,13 +195,9 @@ type WorkloadDeployerInput struct {
...
 // newWorkloadDeployer is the constructor for workloadDeployer.
 func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
-	ws, err := workspace.New()
+	ws, err := workspace.Use(afero.NewOsFs())
 	if err != nil {
-		return nil, fmt.Errorf("new workspace: %w", err)
-	}
-	workspacePath, err := ws.Path()
-	if err != nil {
-		return nil, fmt.Errorf("get workspace path: %w", err)
```
efekarakus, line 204: 💅

internal/pkg/deploy/cloudformation/stack/lb_grpc_web_service_integration_test.go
```diff
@@ -60,9 +62,15 @@ func TestGrpcLoadBalancedWebService_Template(t *testing.T) {
 		v, ok := content.(*manifest.LoadBalancedWebService)
 		require.True(t, ok)
 
-		ws, err := workspace.New()
+		// Create in-memory mock file system.
+		wd, err := os.Getwd()
+		require.NoError(t, err)
+		fs := afero.NewMemMapFs()
+		_ = fs.MkdirAll(fmt.Sprintf("%s/copilot", wd), 0755)
+		_ = afero.WriteFile(fs, fmt.Sprintf("%s/copilot/.workspace", wd), []byte(fmt.Sprintf("---\napplication: %s", "DavidsApp")), 0644)
```
efekarakus, line 70: Awesome improvement! 💅 but terrible app name

internal/pkg/workspace/workspace.go
```diff
@@ -519,68 +498,49 @@ func (ws *Workspace) writeSummary(appName string) error {
...
 	existingWorkspace, _ := ws.copilotDirPath()
 	if existingWorkspace != "" {
-		return nil
+		return existingWorkspace, nil
+	}
+	if err := ws.fs.Mkdir(CopilotDirName, 0755); err != nil {
+		return "", err
 	}
-	return ws.fs.Mkdir(CopilotDirName, 0755)
+	return filepath.Join(ws.workingDirAbs, CopilotDirName), nil
```
efekarakus, line 518: is this different than before?

efekarakus (approved):
Amazing work!! 💅

iamhopaul123 (approved):

internal/pkg/deploy/cloudformation/stack/lb_grpc_web_service_integration_test.go
Lou1415926: :sus

internal/pkg/workspace/workspace.go
Lou1415926: Do you mean if `filepath.Join(ws.workingDirAbs, CopilotDirName)` really always the copilot directory path? (correct me if im wrong!)

If so, I think it's so (that it is the copilot dir path)!

`ws.fs.Mkdir(CopilotDirName, 0755)` is intermediately  `ws.fs.Mkdir("copilot", 0755)`, which will create a `copilot` dir under the current work dir - since `"copilot"` is not an absolute path. Since `ws.workingDirAbs` always gets its value from `os.Getwd()`, its value should be correct - so it checks out 🤔 !

efekarakus: oh I get it now I am so dumb, I don't know what I was thinking. 👍 totally makes sense

For some reason, I kept reading the old code and thinking that `ws.fs.Mkdir()` returned a `(string, error)` rather than just error 🤦

PR #4168, opened 2022-11-10
feat: `env deploy --no-rollback` flag

dannyrandall:
should help with some debugging.
- https://gitter.im/aws/copilot-cli?at=636cb648473cf96648dbcbb9
- https://gitter.im/aws/copilot-cli?at=636c024ca34b51121122d168


internal/pkg/cli/env_deploy.go
```diff
@@ -155,6 +158,16 @@ necessary by a service deployment.
 
 In this case, you can run %s to push a modified template, even if there are no immediate changes.
 `, color.HighlightCode("copilot env deploy --force"))
+		}
+		if o.disableRollback {
+			stackName := stack.NameForEnv(o.targetApp.Name, o.targetEnv.Name)
+			rollbackCmd := fmt.Sprintf("aws cloudformation rollback-stack --stack-name %s --role-arn %s", stackName, o.targetEnv.ExecutionRoleARN)
```
efekarakus, line 165: Really nice to include the role ARN! 💖

efekarakus: nit: This is nice so you don't have to change it, I'm just thinking out loud here. 

I'd have expected this log to be in either the (1) `RecommendedActions` method, _or_, (2) we implement a custom error type that implements the `RecommendActions` method.

I don't think (1) fits since we are not exiting `Execute` successfully, so I think (2) is the better route 💭 

So perhaps something like:
```go
err := fmt.Errorf("deploy environment %s: %w", o.name, err)
if o.disableRollback{ 
  return &errWithDisabledRollback{
     parentErr: err,
     stackName: stack.NameForEnv(o.targetApp.Name, o.targetEnv.Name),
     cmd: fmt.Sprintf("aws cloudformation rollback-stack --stack-name %s --role-arn %s", stackName, o.targetEnv.ExecutionRoleARN)
  }
}
return err
```

efekarakus (approved):

iamhopaul123 (commented):
This is awesome! One quick question: shall we give a different hint to them if their `env deploy` fails because they did `env deploy --no-rollback` before and they haven't manually rollback the deployment?

dannyrandall:
This is what the error looks like right now - I think it's probably good enough, what do you think? The second command is the specific case you're talking about:
```
➜ cl env deploy --no-rollback
✔ Proposing infrastructure changes for the app-env environment.
- Creating the infrastructure for the app-env environment.              [update failed]    [23.1s]
  The following resource(s) failed to create: [DeleteMeRouteTable].                               
  - A Virtual Private Cloud to control networking of your AWS resources        [update complete]  [0.0s]
It seems like you have disabled automatic stack rollback for this deployment.
To debug, you can visit the AWS console to inspect the errors.
After fixing the deployment, you can:
1. Run `aws cloudformation rollback-stack --stack-name app-env --role-arn arn:aws:iam::123456789:role/app-env-CFNExecutionRole` to rollback the deployment.
2. Run `copilot env deploy` to make a new deployment.
✘ deploy environment env: stack app-env did not complete successfully and exited with status UPDATE_FAILED

➜ cl env deploy                                                                                                                                                                                              
✔ Proposing infrastructure changes for the app-env environment.
✘ deploy environment pdx: execute change set arn:aws:cloudformation:us-west-2:123456789:changeSet/copilot-id/id for stack app-env: ValidationError: This stack is currently in a non-terminal [UPDATE_FAILED] state. To update the stack from this state, please use the disable-rollback parameter with update-stack API. To rollback to the last known good state, use the rollback-stack API
        status code: 400, request id: id
```

iamhopaul123 (approved):

PR #4169, opened 2022-11-11
fix: access logs bucket policy applied after ALB update

dannyrandall:
Previously, turning on [access logs](https://aws.github.io/copilot-cli/docs/manifest/environment/#http-public-access-logs) for the ALB would result in the error like:
```
Access Denied for bucket: app-env-elbaccesslogsbucket-1234. Please check S3bucket permission (Service: AmazonElasticLoadBalancing; Status Code: 400; Error Code: InvalidConfigurationRequest; Request ID: id; Proxy: null)
```

This is because Cloudformation would update the ALB's settings _before_ creating the bucket policy that allows the ALB to send logs to the bucket.

Now, we add an explicit dependency to the ALB on the bucket policy if access logs is enabled so that the policy is created/applied before attempting to put objects in the bucket. 

---


efekarakus (approved):
Ty!

KollaAdithya (approved):

PR #4182, opened 2022-11-14
chore: set workspace clients for sub-commands only after `app init` in `init` 

Lou1415926:
This PR will be followed up with a refactoring to `init`. 


internal/pkg/term/selector/selector.go
```diff
@@ -196,6 +203,131 @@ type WorkspaceSelector struct {
...
+	if err != nil {
+		return "", fmt.Errorf("get custom Dockerfile path: %w", err)
+	}
+	return sel, nil
+}
+
+// ListDockerfiles returns the list of Dockerfiles within the current
+// working directory and a sub-directory level below. If an error occurs while
+// reading directories, or no Dockerfiles found returns the error.
+func (s *localFileSelector) listDockerfiles() ([]string, error) {
```
Lou1415926: This method is copied from https://github.com/aws/copilot-cli/blob/e7fadd11e2bea021a80c977fa210d3fe032b142a/internal/pkg/workspace/workspace.go#L608

I changed the received from `Workspace` to `localFileSelector` because listing/reading Dockerfile does not require the existence of a workspace.

internal/pkg/term/selector/selector_test.go
```diff
@@ -3127,3 +2990,214 @@ func TestTaskSelect_Task(t *testing.T) {
...
+			if tc.wantedErr != nil {
+				require.EqualError(t, err, tc.wantedErr.Error())
+			} else {
+				require.Equal(t, tc.wantedDockerfile, dockerfile)
+			}
+		})
+	}
+}
+
+func TestLocalFileSelector_listDockerfiles(t *testing.T) {
```
Lou1415926: Similarly, these test cases are (mostly) copied from https://github.com/aws/copilot-cli/blob/e7fadd11e2bea021a80c977fa210d3fe032b142a/internal/pkg/workspace/workspace_test.go#L1477

Except for `"exclude Dockerfiles in parent directories of the working dir"` which is a new test case.

internal/pkg/term/selector/selector.go
```diff
@@ -196,6 +203,131 @@ type WorkspaceSelector struct {
...
+
+// NewConfigurationSelector constructs a localFconfigurationSelectorileSelector.
+func NewConfigurationSelector(prompt prompter) *configurationSelector {
+	return &configurationSelector{
+		prompt: prompt,
+	}
+}
+
+// Schedule asks the user to select either a rate, preset cron, or custom cron.
+func (s *configurationSelector) Schedule(scheduleTypePrompt, scheduleTypeHelp string, scheduleValidator, rateValidator prompt.ValidatorFunc) (string, error) {
```
Lou1415926: This is just a change of receiver from `WorkspaceSelector` to `configurationSelector`.

internal/pkg/cli/svc_init.go
```diff
@@ -468,6 +472,9 @@ func (o *initSvcOpts) askImage() error {
 }
 
 func (o *initSvcOpts) shouldSkipAsking() (bool, error) {
```
efekarakus: should we rename this method to `doesManifestAlreadyExist`?

internal/pkg/term/selector/selector.go
efekarakus: What do you think of moving this comment to the `localFileSelector` struct? because it wasn't obvious to me that there was a difference between a workspace and local file selector but it makes sense to me after reading this.

internal/pkg/term/selector/selector.go
efekarakus: I'm not clear on why this is called a `configurationSelector` 🤔 why not a `scheduleSelector`?

internal/pkg/term/selector/selector.go
```diff
@@ -196,6 +203,131 @@ type WorkspaceSelector struct {
...
+	}
+	sort.Strings(dockerfiles)
+	return dockerfiles, nil
+}
+
+type configurationSelector struct {
+	prompt prompter
+}
+
+// NewConfigurationSelector constructs a localFconfigurationSelectorileSelector.
```
efekarakus: nit: there is a copy/pasta typo here
```suggestion
// NewConfigurationSelector constructs a configurationSelector.
```

internal/pkg/cli/svc_init.go
iamhopaul123: Maybe `manifestExists`? Having `does` in a function name seems a little odd to me 🤔

iamhopaul123 (commented):
LGTM overall 👍 the `selector.go` file seems to be huge right now. Given the fact that it is the only file in the package, should we consider moving ws package functions to a separate new file in `selector` package?

internal/pkg/term/selector/selector.go
Lou1415926: My thought is that this selector is actually used to ask any configurational questions that do not require any client to read from fs/ssm, etc. `Schedule` happens to be the only existing selector question that is like this.  I wanted to hint it in the name, so that in the future if we want to ask, for example, a `Rate` question, we don't create a `RateSelector` to be the receiver - the receiver could just be this struct. I am happy to change it to `scheduleSelector` though - we can rename it in the future if we load the struct with other questions.

internal/pkg/term/selector/selector.go
Lou1415926: Added a comment to the struct!
```
// localFileSelector selects from a local file system where a workspace does not necessarily exist.
type localFileSelector struct {
	prompt        prompter
	fs            *afero.Afero
	workingDirAbs string
}
```

internal/pkg/cli/svc_init.go
Lou1415926: Changed it to `manifestAlreadyExists` - a nice middle ground 🤣

internal/pkg/term/selector/selector.go
efekarakus: Gotcha that makes sense to me! I think my confusion was that we have a package called `config` which is about SSM 😅 so I wasn't sure if `configurationSelector` was meant to be about SSM or something else.

I'd prefer a `scheduleSelector` but I'll leave it up to you!

efekarakus (approved):

internal/pkg/term/selector/selector.go
Lou1415926: I renamed the selector so that it does not remind folks of SSM!

internal/pkg/cli/init.go
```diff
@@ -193,17 +184,36 @@ func newInitOpts(vars initVars) (*initOpts, error) {
...
+		ws, err := workspace.Use(fs)
+		if err != nil {
+			return err
+		}
+		sel := selector.NewLocalWorkloadSelector(prompt, configStore, ws)
+		initEnvCmd.manifestWriter = ws
+		deployEnvCmd.ws = ws
+		deploySvcCmd.ws = ws
+		deployJobCmd.ws = ws
+		deployJobCmd.sel = sel
```
dannyrandall, line 209: do we need to set `deploySvcCmd.sel` as well?

dannyrandall (approved):
looks good, just a question!

PR #4183, opened 2022-11-15
docs: add sc docs

iamhopaul123:

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -210,6 +210,7 @@ plugins:
             'docs/commands/pipeline-update.ja.md': 'docs/commands/pipeline-deploy.ja.md'
             'docs/developing/additional-aws-resources.en.md': 'docs/developing/addons/modeling.en.md'
             'docs/developing/additional-aws-resources.ja.md': 'docs/developing/addons/modeling.ja.md'
+            'docs/developing/service-discovery.en.md': 'docs/developing/internal-traffic.en.md'
     - i18n:
         default_language: en
         languages:
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -55,11 +55,11 @@ nav:
       - Custom Environment Resources: docs/developing/custom-environment-resources.en.md
       - Domain: docs/developing/domain.en.md
       - Internal Load Balancers: docs/developing/internal-albs.en.md
-      - Internal Traffic: docs/developing/internal-traffic.en.md
       - Manifest Environment Variables: docs/developing/manifest-env-var.en.md
       - Observability: docs/developing/observability.en.md
       - Publish/Subscribe: docs/developing/publish-subscribe.en.md
       - Secrets: docs/developing/secrets.en.md
+      - Service-to-Service Communication: docs/developing/svc-to-svc-communication.en.md
       - Sidecars: docs/developing/sidecars.en.md
       - Storage: docs/developing/storage.en.md
       - Task Definition Overrides: docs/developing/taskdef-overrides.en.md
@@ -210,7 +210,7 @@ plugins:
             'docs/commands/pipeline-update.ja.md': 'docs/commands/pipeline-deploy.ja.md'
             'docs/developing/additional-aws-resources.en.md': 'docs/developing/addons/modeling.en.md'
             'docs/developing/additional-aws-resources.ja.md': 'docs/developing/addons/modeling.ja.md'
-            'docs/developing/service-discovery.en.md': 'docs/developing/internal-traffic.en.md'
+            'docs/developing/service-discovery.en.md': 'docs/developing/svc-to-svc-communication.en.md'
     - i18n:
         default_language: en
         languages:
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -55,11 +55,11 @@ nav:
       - Custom Environment Resources: docs/developing/custom-environment-resources.en.md
       - Domain: docs/developing/domain.en.md
       - Internal Load Balancers: docs/developing/internal-albs.en.md
+      - Internal Traffic: docs/developing/internal-traffic.en.md
       - Manifest Environment Variables: docs/developing/manifest-env-var.en.md
       - Observability: docs/developing/observability.en.md
       - Publish/Subscribe: docs/developing/publish-subscribe.en.md
       - Secrets: docs/developing/secrets.en.md
-      - Service Discovery: docs/developing/service-discovery.en.md
       - Sidecars: docs/developing/sidecars.en.md
       - Storage: docs/developing/storage.en.md
       - Task Definition Overrides: docs/developing/taskdef-overrides.en.md
--- a/site/content/docs/developing/internal-traffic.en.md
+++ b/site/content/docs/developing/internal-traffic.en.md
@@ -1,4 +1,3 @@
-
 # Internal Traffic
 
 ## Service Discovery
@@ -47,16 +46,16 @@ When your environments are upgraded, Copilot will honor the service discovery na
 Similar to Service Discovery, ECS Service Connect provides a “one-click” experience to have your client services connect to their dependencies in a load-balanced and resilient fashion, simplifying the way of exposing those dependencies to their clients. With Service Connect, each service you create is given a private (in-VPC) alias by default, which is also configurable in your service manifests.
 
 ### How do I use Service Connect?
-Imagine we have the same app called `kudos` and two services: `api` and `front-end`. And both of them are deployed to the same environment. In order to use Service Connect, both services need to have
+Imagine we have the same app called `kudos` and two services: `api` and `front-end`, deployed in the same environment. In order to use Service Connect, both services' manifests need to have:
 
 ```yaml
 network:
-  connect: true # By default it is "false"
+  connect: true # Defaults to "false"
 ```
 
-in their manifests. And after deploying both services, they should be able to talk to each other using the default Service Connect endpoint, which is the same as its service name. For example, `front-end` service can simply call `http://api`.
+After deploying both services, they should be able to talk to each other using the default Service Connect endpoint, which is the same as its service name. For example, `front-end` service can simply call `http://api`.
 
-However, if you are already using Service Discovery right now and uncomfortable for make any code changes. You can configure the `connect.alias` field so that the Service Connect uses the same endpoint as Service Discovery. And if **both** client and caller have Service Connect enabled, they'll be talking in Service Connect instead of Service Discovery. For example, in the manifest of the `api` service we have
+However, if you are already using Service Discovery right now and want to avoid any code changes, you can configure [`network.connect.alias`](../manifest/lb-web-service.en.md#network-connect-alias) field so that the Service Connect uses the same alias as Service Discovery. And if **both** client and caller have Service Connect enabled, they'll connect via Service Connect instead of Service Discovery. For example, in the manifest of the `api` service we have
 
 ```yaml
 network:
--- a/site/content/docs/developing/internal-traffic.en.md
+++ b/site/content/docs/developing/internal-traffic.en.md
@@ -1,12 +1,15 @@
-# Service Discovery
+
+# Internal Traffic
+
+## Service Discovery
 
 Service Discovery is a way of letting services discover and connect with each other. Typically, services can only talk to each other if they expose a public endpoint - and even then, requests will have to go over the internet. With [ECS Service Discovery](https://docs.aws.amazon.com/whitepapers/latest/microservices-on-aws/service-discovery.html), each service you create is given a private address and DNS name - meaning each service can talk to another without ever leaving the local network (VPC) and without exposing a public endpoint.  
 
-## How do I use Service Discovery?
+### How do I use Service Discovery?
 
 Service Discovery is enabled for all services set up using the Copilot CLI. We'll show you how to use it by using an example. Imagine we have an app called `kudos` and two services: `api` and `front-end`.
 
-In this example we'll imagine our `front-end` service is deployed in the `test` environment, has a public endpoint and wants to call our `api` service using its service discovery endpoint. 
+In this example we'll imagine our `front-end` service is deployed in the `test` environment, has a public endpoint and wants to call our `api` service using its service discovery endpoint.
 
 ```go
 // Calling our api service from the front-end service using Service Discovery
@@ -32,10 +35,33 @@ endpoint := fmt.Sprintf("http://api.%s/some-request", os.Getenv("COPILOT_SERVICE
 
 `COPILOT_SERVICE_DISCOVERY_ENDPOINT` is a special environment variable that the Copilot CLI sets for you when it creates your service. It's of the format _{env name}.{app name}.local_ - so in this case in our _kudos_ app, when deployed in the _test_ environment, the request would be to `http://api.test.kudos.local/some-request`. Since our _api_ service is running on port 80, we're not specifying the port in the URL. However, if it was running on another port, say 8080, we'd need to include the port in the request, as well `http://api.test.kudos.local:8080/some-request`.
 
-When our front-end makes this request, the endpoint `api.test.kudos.local` resolves to a private IP address and is routed privately within your VPC. 
+When our front-end makes this request, the endpoint `api.test.kudos.local` resolves to a private IP address and is routed privately within your VPC.
 
-## Legacy Environments and Service Discovery
+### Legacy Environments and Service Discovery
 
 Prior to Copilot v1.9.0, the service discovery namespace used the format _{app name}.local_, without including the environment. This limitation made it impossible to deploy multiple environments in the same VPC. Any environments created with Copilot v1.9.0 and newer can share a VPC with any other environment.
 
-When your environments are upgraded, Copilot will honor the service discovery namespace that the environment was created with. That means that the endpoint your services are reachable at will not change. Any new environments created with Copilot v1.9.0 and above will use the _{env name}.{app name}.local_ format for service discovery, and can share VPCs with older environments. 
+When your environments are upgraded, Copilot will honor the service discovery namespace that the environment was created with. That means that the endpoint your services are reachable at will not change. Any new environments created with Copilot v1.9.0 and above will use the _{env name}.{app name}.local_ format for service discovery, and can share VPCs with older environments.
+
+## Service Connect
+Similar to Service Discovery, ECS Service Connect provides a “one-click” experience to have your client services connect to their dependencies in a load-balanced and resilient fashion, simplifying the way of exposing those dependencies to their clients. With Service Connect, each service you create is given a private (in-VPC) alias by default, which is also configurable in your service manifests.
+
+### How do I use Service Connect?
+Imagine we have the same app called `kudos` and two services: `api` and `front-end`. And both of them are deployed to the same environment. In order to use Service Connect, both services need to have
+
+```yaml
+network:
+  connect: true # By default it is "false"
+```
+
+in their manifests. And after deploying both services, they should be able to talk to each other using the default Service Connect endpoint, which is the same as its service name. For example, `front-end` service can simply call `http://api`.
+
+However, if you are already using Service Discovery right now and uncomfortable for make any code changes. You can configure the `connect.alias` field so that the Service Connect uses the same endpoint as Service Discovery. And if **both** client and caller have Service Connect enabled, they'll be talking in Service Connect instead of Service Discovery. For example, in the manifest of the `api` service we have
+
+```yaml
+network:
+  connect:
+    alias: ${COPILOT_SERVICE_NAME}.${COPILOT_ENVIRONMENT_NAME}.${COPILOT_APPLICATION_NAME}.local
+```
+
+and `front-end` can keep using the same endpoint to make API calls via Service Connect instead of Service Discovery.
--- a/site/content/docs/developing/svc-to-svc-communication.en.md
+++ b/site/content/docs/developing/svc-to-svc-communication.en.md
@@ -1,4 +1,38 @@
-# Internal Traffic
+# Service-to-Service Communication
+
+## Service Connect
+
+!!! info
+    This feature is supported since Copilot `v1.24`, which provides more integrated service-to-service experience compared to [Service Discovery](#service-discovery).
+
+ECS Service Connect enables your client services connect to their dependencies in a load-balanced and resilient fashion, simplifying the way of exposing those dependencies to their clients. With Service Connect in Copilot, each service you create is given a private alias by default: `http://<your service name>`.
+
+### How do I use Service Connect?
+Imagine we have an app called `kudos` and two services: `api` and `front-end`, deployed in the same environment. In order to use Service Connect, both services' manifests need to have:
+
+```yaml
+network:
+  connect: true # Defaults to "false"
+```
+
+After deploying both services, they should be able to talk to each other using the default Service Connect endpoint, which is the same as its service name. For example, `front-end` service can simply call `http://api`.
+
+```go
+// Calling the "api" service from the "front-end" service.
+resp, err := http.Get("http://api/")
+```
+
+### Upgrading from Service Discovery
+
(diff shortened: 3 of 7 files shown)


mkdocs.yml
```diff
@@ -55,11 +55,11 @@ nav:
       - Custom Environment Resources: docs/developing/custom-environment-resources.en.md
       - Domain: docs/developing/domain.en.md
       - Internal Load Balancers: docs/developing/internal-albs.en.md
+      - Internal Traffic: docs/developing/internal-traffic.en.md
```
dannyrandall: should probably add a redirect in `redirect_maps` (same file)

site/content/docs/developing/internal-traffic.en.md
```diff
@@ -1,12 +1,15 @@
-# Service Discovery
+
```
dannyrandall:
```suggestion
```

site/content/docs/developing/internal-traffic.en.md
```diff
@@ -32,10 +35,33 @@ endpoint := fmt.Sprintf("http://api.%s/some-request", os.Getenv("COPILOT_SERVICE
...
+When your environments are upgraded, Copilot will honor the service discovery namespace that the environment was created with. That means that the endpoint your services are reachable at will not change. Any new environments created with Copilot v1.9.0 and above will use the _{env name}.{app name}.local_ format for service discovery, and can share VPCs with older environments.
+
+## Service Connect
+Similar to Service Discovery, ECS Service Connect provides a “one-click” experience to have your client services connect to their dependencies in a load-balanced and resilient fashion, simplifying the way of exposing those dependencies to their clients. With Service Connect, each service you create is given a private (in-VPC) alias by default, which is also configurable in your service manifests.
+
+### How do I use Service Connect?
+Imagine we have the same app called `kudos` and two services: `api` and `front-end`. And both of them are deployed to the same environment. In order to use Service Connect, both services need to have
```
dannyrandall:
```suggestion
Imagine we have the same app called `kudos` and two services: `api` and `front-end`, deployed in the same environment. In order to use Service Connect, both services' manifests need to have:
```

site/content/docs/developing/internal-traffic.en.md
```diff
@@ -32,10 +35,33 @@ endpoint := fmt.Sprintf("http://api.%s/some-request", os.Getenv("COPILOT_SERVICE
...
+
+```yaml
+network:
+  connect: true # By default it is "false"
```
dannyrandall:
```suggestion
  connect: true # Defaults to "false"
```

site/content/docs/developing/internal-traffic.en.md
```diff
@@ -32,10 +35,33 @@ endpoint := fmt.Sprintf("http://api.%s/some-request", os.Getenv("COPILOT_SERVICE
...
+```
+
+in their manifests. And after deploying both services, they should be able to talk to each other using the default Service Connect endpoint, which is the same as its service name. For example, `front-end` service can simply call `http://api`.
```
dannyrandall:
```suggestion
After deploying both services, they should be able to talk to each other using the default Service Connect endpoint, which is the same as its service name. For example, `front-end` service can simply call `http://api`.
```

site/content/docs/developing/internal-traffic.en.md
```diff
@@ -32,10 +35,33 @@ endpoint := fmt.Sprintf("http://api.%s/some-request", os.Getenv("COPILOT_SERVICE
...
+
+However, if you are already using Service Discovery right now and uncomfortable for make any code changes. You can configure the `connect.alias` field so that the Service Connect uses the same endpoint as Service Discovery. And if **both** client and caller have Service Connect enabled, they'll be talking in Service Connect instead of Service Discovery. For example, in the manifest of the `api` service we have
```
dannyrandall:
```suggestion
However, if you are already using Service Discovery right now and want to avoid any code changes, you can configure [`network.connect.alias`](manifest doc link) so that Service Connect uses the same alias as Service Discovery. And if **both** client and caller have Service Connect enabled, they'll connect via Service Connect instead of Service Discovery. For example, in the manifest of the `api` service we have
```

site/content/docs/include/network.en.md
```diff
@@ -3,6 +3,14 @@
 <a id="network" href="#network" class="field">`network`</a> <span class="type">Map</span>      
 The `network` section contains parameters for connecting to AWS resources in a VPC.
 
+<span class="parent-field">network.</span><a id="network-connect" href="#network-connect" class="field">`connect`</a> <span class="type">Bool or Map</span>    
+Enable [Service Connect](../developing/internal-traffic.en.md#service-connect) for your service, which makes the traffic between services load balanced and more resilient. The default is `false`.
```
dannyrandall:
```suggestion
Enable [Service Connect](../developing/internal-traffic.en.md#service-connect) for your service, which makes the traffic between services load balanced and more resilient. Defaults to `false`.
```

also might be nice to have a link to ECS docs here (might not be possible currently)

site/content/docs/include/network.en.md
```diff
@@ -3,6 +3,14 @@
...
+
+When using it as a map, you can specify which alias to use for this service. Note that the alias must be unique within the environment.
```
dannyrandall, line 9:
```suggestion
When using it as a map, you can specify which alias to use for this service. Note that the alias must be unique within the environment and VPC.
```

site/content/docs/include/network.en.md
```diff
@@ -3,6 +3,14 @@
...
+
+<span class="parent-field">network.</span><a id="network-connect-alias" href="#network-connect-alias" class="field">`alias`</a> <span class="type">String</span>  
+A custom DNS name for your Service Connect.
```
dannyrandall:
```suggestion
A custom DNS name for this service exposed to Service Connect.
```

site/content/docs/include/network.en.md
iamhopaul123: yeah we can add later.

site/content/docs/include/network.en.md
iamhopaul123: strictly environment and has nothing to do with VPC because it is unique in a namespace.

mkdocs.yml
efekarakus: nit: I wonder if "Internal Traffic" is clear enough for folks? I think the "Service Discovery" name is the typical name for this concept: https://microservices.io/patterns/server-side-discovery.html alternatively, we can do:
```suggestion
      - Service-to-Service Communication: docs/developing/internal-traffic.en.md
```

site/content/docs/developing/internal-traffic.en.md
```diff
@@ -32,10 +34,33 @@ endpoint := fmt.Sprintf("http://api.%s/some-request", os.Getenv("COPILOT_SERVICE
...
 
-## Legacy Environments and Service Discovery
+### Legacy Environments and Service Discovery
 
 Prior to Copilot v1.9.0, the service discovery namespace used the format _{app name}.local_, without including the environment. This limitation made it impossible to deploy multiple environments in the same VPC. Any environments created with Copilot v1.9.0 and newer can share a VPC with any other environment.
 
-When your environments are upgraded, Copilot will honor the service discovery namespace that the environment was created with. That means that the endpoint your services are reachable at will not change. Any new environments created with Copilot v1.9.0 and above will use the _{env name}.{app name}.local_ format for service discovery, and can share VPCs with older environments. 
+When your environments are upgraded, Copilot will honor the service discovery namespace that the environment was created with. That means that the endpoint your services are reachable at will not change. Any new environments created with Copilot v1.9.0 and above will use the _{env name}.{app name}.local_ format for service discovery, and can share VPCs with older environments.
+
+## Service Connect
```
efekarakus: Can we move the Service Connect section to the top? I think that should be our recommended approach first.

site/content/docs/developing/internal-traffic.en.md
```diff
@@ -1,12 +1,14 @@
-# Service Discovery
+# Internal Traffic
+
+## Service Discovery
```
efekarakus: Can we add an [`info` admonition here](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#supported-types) recommending folks that they should prefer using service connect which was added with Copilot v1.24. Service connect provides a more resilient and easier to use communication alternative that doesn't require users to modify their code.

site/content/docs/developing/internal-traffic.en.md
efekarakus: I wonder if we can add a nice badge here saying "added in copilot1.24" like how Go does: https://pkg.go.dev/os#MkdirTemp

site/content/docs/developing/internal-traffic.en.md
```diff
@@ -32,10 +34,33 @@ endpoint := fmt.Sprintf("http://api.%s/some-request", os.Getenv("COPILOT_SERVICE
...
 
-When your environments are upgraded, Copilot will honor the service discovery namespace that the environment was created with. That means that the endpoint your services are reachable at will not change. Any new environments created with Copilot v1.9.0 and above will use the _{env name}.{app name}.local_ format for service discovery, and can share VPCs with older environments. 
+When your environments are upgraded, Copilot will honor the service discovery namespace that the environment was created with. That means that the endpoint your services are reachable at will not change. Any new environments created with Copilot v1.9.0 and above will use the _{env name}.{app name}.local_ format for service discovery, and can share VPCs with older environments.
+
+## Service Connect
+Similar to Service Discovery, ECS Service Connect provides a “one-click” experience to have your client services connect to their dependencies in a load-balanced and resilient fashion, simplifying the way of exposing those dependencies to their clients. With Service Connect, each service you create is given a private (in-VPC) alias by default, which is also configurable in your service manifests.
```
efekarakus:
```suggestion
ECS Service Connect enables your services to connect to their dependencies in a load-balanced and resilient fashion. With Service Connect, each service is given a private alias within the VPC accessible at `http://<your service name>` .
```

site/content/docs/developing/internal-traffic.en.md
```diff
@@ -32,10 +34,33 @@ endpoint := fmt.Sprintf("http://api.%s/some-request", os.Getenv("COPILOT_SERVICE
...
+
+### How do I use Service Connect?
+Imagine we have the same app called `kudos` and two services: `api` and `front-end`, deployed in the same environment. In order to use Service Connect, both services' manifests need to have:
+
+```yaml
+network:
+  connect: true # Defaults to "false"
+```
+
+After deploying both services, they should be able to talk to each other using the default Service Connect endpoint, which is the same as its service name. For example, `front-end` service can simply call `http://api`.
```
efekarakus:
```suggestion
After deploying both services, they should be able to talk to each other using the default Service Connect endpoint, which is the same as its service name. For example, `front-end` service can simply call `http://api`.

```go
// Calling the "api" service from the "front-end" service.
resp, err := http.Get("http://api/")
```
```

File: site/content/docs/developing/internal-traffic.en.md [hunk 14]
```diff
@@ -32,10 +34,33 @@ endpoint := fmt.Sprintf("http://api.%s/some-request", os.Getenv("COPILOT_SERVICE
...
+
+However, if you are already using Service Discovery right now and want to avoid any code changes, you can configure [`network.connect.alias`](../manifest/lb-web-service.en.md#network-connect-alias) field so that the Service Connect uses the same alias as Service Discovery. And if **both** client and caller have Service Connect enabled, they'll connect via Service Connect instead of Service Discovery. For example, in the manifest of the `api` service we have
```
efekarakus:
```suggestion

### Upgrading from Service Discovery
Prior to v1.24, Copilot enabled private service-to-service communication with [service discovery](#service-discovery). If you are already using Service Discover and want to avoid any code changes, you can configure [`network.connect.alias`](../manifest/lb-web-service.en.md#network-connect-alias) field so that the Service Connect uses the same alias as Service Discovery. And if **both** client and server have Service Connect enabled, they'll connect via Service Connect instead of Service Discovery. For example, in the manifest of the `api` service we have:
```

site/content/docs/developing/internal-traffic.en.md
```diff
@@ -32,10 +34,33 @@ endpoint := fmt.Sprintf("http://api.%s/some-request", os.Getenv("COPILOT_SERVICE
...
+
+```yaml
+network:
+  connect:
+    alias: ${COPILOT_SERVICE_NAME}.${COPILOT_ENVIRONMENT_NAME}.${COPILOT_APPLICATION_NAME}.local
+```
+
+and `front-end` can keep using the same endpoint to make API calls via Service Connect instead of Service Discovery.
```
efekarakus:
```suggestion
and `front-end` can keep using the same endpoint to make API calls via Service Connect instead of Service Discovery to leverage the benefits of load balancing and additional resiliency.
```

site/content/docs/include/network.en.md
```diff
@@ -3,6 +3,14 @@
 <a id="network" href="#network" class="field">`network`</a> <span class="type">Map</span>      
 The `network` section contains parameters for connecting to AWS resources in a VPC.
 
+<span class="parent-field">network.</span><a id="network-connect" href="#network-connect" class="field">`connect`</a> <span class="type">Bool or Map</span>    
+Enable [Service Connect](../developing/internal-traffic.en.md#service-connect) for your service, which makes the traffic between services load balanced and more resilient. Defaults to `false`.
```
efekarakus: Can we explain the default alias value when the value is set to `true`?

site/content/docs/include/network.en.md
```diff
@@ -3,6 +3,14 @@
...
+
+When using it as a map, you can specify which alias to use for this service. Note that the alias must be unique within the environment.
+
+<span class="parent-field">network.</span><a id="network-connect-alias" href="#network-connect-alias" class="field">`alias`</a> <span class="type">String</span>  
+A custom DNS name for this service exposed to Service Connect.
```
efekarakus: Can we show an example that folks would want to use here?

efekarakus:
Also thank you for the docs boss :)

site/content/docs/developing/internal-traffic.en.md
huanjani: and, iirc, Service Connect is free whereas Service Discovery isn't, right?

iamhopaul123: > Can we add an [info admonition here](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#supported-types) recommending folks that they should prefer using service connect which was added with Copilot v1.24.

But all the info's been mentioned in the doc 🤔 

> and, iirc, Service Connect is free whereas Service Discovery isn't, right?

I think both of them are "free", and both of them are subject to additional cost (sc could be even higher because of the additional cost caused by the envoy sidecar usage)

mkdocs.yml
huanjani: if we change it, we should also maybe change the file name?

mkdocs.yml
```diff
@@ -210,6 +210,7 @@ plugins:
             'docs/commands/pipeline-update.ja.md': 'docs/commands/pipeline-deploy.ja.md'
             'docs/developing/additional-aws-resources.en.md': 'docs/developing/addons/modeling.en.md'
             'docs/developing/additional-aws-resources.ja.md': 'docs/developing/addons/modeling.ja.md'
+            'docs/developing/service-discovery.en.md': 'docs/developing/internal-traffic.en.md'
```
huanjani: unrelated: these redirects aren't all working. maybe a problem we can solve separately.

site/content/docs/developing/internal-traffic.en.md
```diff
@@ -32,10 +34,33 @@ endpoint := fmt.Sprintf("http://api.%s/some-request", os.Getenv("COPILOT_SERVICE
...
 
-## Legacy Environments and Service Discovery
+### Legacy Environments and Service Discovery
 
 Prior to Copilot v1.9.0, the service discovery namespace used the format _{app name}.local_, without including the environment. This limitation made it impossible to deploy multiple environments in the same VPC. Any environments created with Copilot v1.9.0 and newer can share a VPC with any other environment.
 
-When your environments are upgraded, Copilot will honor the service discovery namespace that the environment was created with. That means that the endpoint your services are reachable at will not change. Any new environments created with Copilot v1.9.0 and above will use the _{env name}.{app name}.local_ format for service discovery, and can share VPCs with older environments. 
+When your environments are upgraded, Copilot will honor the service discovery namespace that the environment was created with. That means that the endpoint your services are reachable at will not change. Any new environments created with Copilot v1.9.0 and above will use the _{env name}.{app name}.local_ format for service discovery, and can share VPCs with older environments.
```
huanjani:
```suggestion
When your environments are upgraded, Copilot will honor the service discovery namespace that the environment was created with. That means that the endpoints for your services will not change. Any new environments created with Copilot v1.9.0 and above will use the _{env name}.{app name}.local_ format for service discovery, and can share VPCs with older environments.
```

site/content/docs/developing/internal-traffic.en.md
```diff
@@ -32,10 +34,33 @@ endpoint := fmt.Sprintf("http://api.%s/some-request", os.Getenv("COPILOT_SERVICE
...
+
+## Service Connect
+Similar to Service Discovery, ECS Service Connect provides a “one-click” experience to have your client services connect to their dependencies in a load-balanced and resilient fashion, simplifying the way of exposing those dependencies to their clients. With Service Connect, each service you create is given a private (in-VPC) alias by default, which is also configurable in your service manifests.
+
+### How do I use Service Connect?
+Imagine we have the same app called `kudos` and two services: `api` and `front-end`, deployed in the same environment. In order to use Service Connect, both services' manifests need to have:
```
huanjani:
```suggestion
Imagine we have the same `kudos` app with two services: `api` and `front-end`, deployed in the same environment. In order to use Service Connect, both services' manifests need to have:
```

huanjani (commented):
Looks good! Some small suggestions.

site/content/docs/include/network.en.md
iamhopaul123: We have the example in `../developing/internal-traffic.en.md#service-connect`. we probably don't need an example if we've already had a section explaining how it works.

site/content/docs/developing/svc-to-svc-communication.en.md
```diff
@@ -1,12 +1,48 @@
-# Service Discovery
+# Service-to-Service Communication
+
+## Service Connect
+
+!!! info
+    This feature is supported since Copilot `v1.24`, which provides more integrated service-to-service experience compared to [Service Discovery](#service-discovery).
+
+ECS Service Connect enables your client services connect to their dependencies in a load-balanced and resilient fashion, simplifying the way of exposing those dependencies to their clients. With Service Connect in Copilot, each service you create is given a private alias by default: `http://<your service name>`.
```
efekarakus: It would be cool if we can include a link to the ECS docs here, maybe we can ask them what the URL will be?
```suggestion
[ECS Service Connect](some-ecs-doc-url) enables your client services connect to their dependencies in a load-balanced and resilient fashion, simplifying the way of exposing those dependencies to their clients. With Service Connect in Copilot, each service you create is given a private alias by default: `http://<your service name>`.
```

efekarakus: There might be a word that's missing here 😅 
```suggestion
ECS Service Connect enables a client service to connect to its downstream services in a load-balanced and resilient fashion. Furthermore, it simplifies the way of exposing a service to its clients by specifying friendly aliases. With Service Connect in Copilot, each service you create is given the following private alias by default: `http://<your service name>`.
```

site/content/docs/developing/svc-to-svc-communication.en.md
```diff
@@ -1,12 +1,48 @@
...
+network:
+  connect: true # Defaults to "false"
+```
+
+After deploying both services, they should be able to talk to each other using the default Service Connect endpoint, which is the same as its service name. For example, `front-end` service can simply call `http://api`.
+
+```go
+// Calling the "api" service from the "front-end" service.
+resp, err := http.Get("http://api/")
+```
```
efekarakus, line 20: This reminds me (in a separate PR), I wonder if we should a recommended action after `copilot deploy` saying:
```
Your service is accessible within the VPC at: fmt.Sprintf(http://%s:%s/, name, port)
```

site/content/docs/developing/svc-to-svc-communication.en.md
```diff
@@ -1,12 +1,48 @@
...
+
+### Upgrading from Service Discovery
+
+Prior to v1.24, Copilot enabled private service-to-service communication with Service Discovery. If you are already using Service Discovery and want to avoid any code changes, you can configure [`network.connect.alias`](../manifest/lb-web-service.en.md#network-connect-alias) field so that the Service Connect uses the same alias as Service Discovery. And if **both** client and caller have Service Connect enabled, they'll connect via Service Connect instead of Service Discovery. For example, in the manifest of the `api` service we have
```
efekarakus: isn't client and caller the same thing?
```suggestion
Prior to v1.24, Copilot enabled private service-to-service communication with Service Discovery. If you are already using Service Discovery and want to avoid any code changes, you can configure [`network.connect.alias`](../manifest/lb-web-service.en.md#network-connect-alias) field so that the Service Connect uses the same alias as Service Discovery. And if **both**, the service and its client, have Service Connect enabled, they'll connect via Service Connect instead of Service Discovery. For example, in the manifest of the `api` service we have
```

site/content/docs/developing/svc-to-svc-communication.en.md
```diff
@@ -1,12 +1,48 @@
...
+
+```yaml
+network:
+  connect:
+    alias: ${COPILOT_SERVICE_NAME}.${COPILOT_ENVIRONMENT_NAME}.${COPILOT_APPLICATION_NAME}.local
+```
+
+and `front-end` can keep using the same endpoint to make API calls via Service Connect instead of Service Discovery to leverage the benefits of load balancing and additional resiliency.
```
efekarakus:
```suggestion
and `front-end` also has the same setting. Then, they can keep using the same endpoint to make API calls via Service Connect instead of Service Discovery to leverage the benefits of load balancing and additional resiliency.
```

efekarakus (commented):
Looks great! Just some tiny tweaks boss and then :shipit:

site/content/docs/developing/svc-to-svc-communication.en.md
```diff
@@ -1,12 +1,48 @@
-# Service Discovery
+# Service-to-Service Communication
+
+## Service Connect
+
+!!! info
+    This feature is supported since Copilot `v1.24`, which provides more integrated service-to-service experience compared to [Service Discovery](#service-discovery).
```
dannyrandall: what about https://github.com/aws/copilot-cli/blob/mainline/site/content/docs/developing/addons/package.en.md#L1 🥺

site/content/docs/developing/svc-to-svc-communication.en.md
```diff
@@ -1,12 +1,48 @@
...
+Imagine we have an app called `kudos` and two services: `api` and `front-end`, deployed in the same environment. In order to use Service Connect, both services' manifests need to have:
+
+```yaml
+network:
+  connect: true # Defaults to "false"
+```
+
+After deploying both services, they should be able to talk to each other using the default Service Connect endpoint, which is the same as its service name. For example, `front-end` service can simply call `http://api`.
+
+```go
```
dannyrandall, line 17: ❤️

dannyrandall (approved):
great docs!

site/content/docs/developing/svc-to-svc-communication.en.md
iamhopaul123: I think I added that like

```
Recommended follow-up action:
  - You can access your service at redis-server:6379 with service connect.
```

efekarakus (approved):
So stylish

PR #4185, opened 2022-11-15
fix: NLB use case for service connect

paragbhingre:
This PR adds Name parameter to the port mapping in the NLB use case which will be releasing as a part of Service Connect feature.

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/testdata/stacklocal/override-cf.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/stacklocal/override-cf.yml
@@ -40,7 +40,7 @@
               awslogs-stream-prefix: copilot
           PortMappings:
             - ContainerPort: !Ref ContainerPort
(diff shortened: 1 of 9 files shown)


iamhopaul123 (commented):
we need to fix the SC part to make sure we use the correct one (not "target" because it does not exist any more). But this 
makes me start to wonder what port should we use for it if both ALBtarget and NLBTarget exist?

I think basically we need to create two [SC service](https://github.com/aws/copilot-cli/blob/9fb61228da96befccf48c0dae4f526a1a5c0ca66/internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml#L47) in this scenario and one for ALBTarget and one for NLBTarget with correct alias port (used for SC endpoint like http://frontend:8080, which should be the same as the container port)

paragbhingre:
Deleting in favor of multiple ports. This will be taken care in multiple ports feature.

PR #4186, opened 2022-11-15
feat: add AutoScalingConfiguration to RDWS

### LINKED ISSUE #3200: App Runner : Please support AutoScalingConfiguration settings in manifest.yml
I want to use `min_size` of AutoScalingConfiguration setting.
I think 1 is default, but I want to increase it to 2. 
So, please support AutoScalingConfiguration settings in manifest.yml. 

refs:
- https://docs.aws.amazon.com/apprunner/latest/dg/manage-autoscaling.html
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/apprunner_auto_scaling_configuration_version
- https://docs.aws.amazon.com/apprunner/latest/api/API_AutoScalingConfiguration.html
- https://docs.aws.amazon.com/apprunner/latest/api/API_DescribeAutoScalingConfiguration.html

KollaAdithya:
This PR resolves #3200
With this PR we will have a new field `count` to the  Manifest of RequestDrivenService.
```yaml
count: 'high-availability/3'
```
In order to use this field customers  have to provide `AutoScalingConfigurationName/AutoScalingConfigurationRevision`  as input to`count`.

`count: AutoScalingConfigurationName/AutoScalingConfigurationRevision`

This field will help customers to reference existing autoscaling configurations for different services.


internal/pkg/template/templates/workloads/services/rd-web/cf.yml
```diff
@@ -150,6 +150,9 @@ Resources:
         ObservabilityEnabled: true
         ObservabilityConfigurationArn: !Sub 'arn:aws:apprunner:${AWS::Region}:${AWS::AccountId}:observabilityconfiguration/DefaultConfiguration/1/00000000000000000000000000000001'
       {{- end }}
+      {{- if .Count}}
```
paragbhingre, line 153: How about using isARN function to make sure customers are giving valid input?
You can refer this implementation in the [task cfn template](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/task/cf.yml#L180).

paragbhingre: Now that I think more, if we are accepting ARN input from the customer then we can also validate this before even coming till this far in the cfn template also. May be in the manifest validation.

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
```diff
@@ -69,7 +69,7 @@ Resources:
     Metadata:
       'aws:copilot:description': 'An App Runner service to run and manage your containers'
     Type: AWS::AppRunner::Service
-    Properties:
+    Properties:
```
paragbhingre:
```suggestion
    Properties:
```

internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-test.stack.yml
```diff
@@ -153,6 +153,7 @@ Resources:
         EgressConfiguration:
           EgressType: VPC
           VpcConnectorArn: !Ref VpcConnector
+      AutoScalingConfigurationArn: 'arn:aws:apprunner:us-east-1:123456789012:autoscalingconfiguration/high-availability/3'
```
efekarakus: Thank you for the PR Adi!

Do we need to accept an entire ARN? can we generate the ARN ourselves so that users don't have to write the account ID and region and instead they can pass just `highavailability/3`?

In cloudformation, we can do something like:
https://github.com/aws/copilot-cli/blob/5781c535bc78e5e3a544b855d930256562e52b63/internal/pkg/deploy/cloudformation/stack/testdata/pipeline/cc_template.yaml#L97

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
KollaAdithya: With @efekarakus suggestion input is not going to be complete ARN. Still do we need to validate the ARN ?

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
```diff
@@ -150,6 +150,9 @@ Resources:
         ObservabilityEnabled: true
         ObservabilityConfigurationArn: !Sub 'arn:aws:apprunner:${AWS::Region}:${AWS::AccountId}:observabilityconfiguration/DefaultConfiguration/1/00000000000000000000000000000001'
       {{- end }}
+      {{- if .Count}}   
+      AutoScalingConfigurationArn: !Sub 'arn:aws:apprunner:${AWS::Region}:${AWS::AccountId}:autoscalingconfiguration/{{.Count}}'
```
iamhopaul123:
```suggestion
      AutoScalingConfigurationArn: !Sub 'arn:${AWS::Partition}:apprunner:${AWS::Region}:${AWS::AccountId}:autoscalingconfiguration/{{.Count}}'
```

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
paragbhingre: Nope, with the new implementation it is not required

efekarakus (approved):
🚀

paragbhingre (approved):
:shipit:

miketheman:
Excellent to see this get merged! Is there a separate process to add the `count` syntax [to the docs](https://github.com/aws/copilot-cli/blob/6a50bcb45f4aafb723df31b9cf6f56954174dd24/site/content/docs/manifest/rd-web-service.en.md)?

PR #4187, opened 2022-11-15
chore: set tasks fewer than `spot_from` to fargate cp when `min` and `spot_from` equal

huanjani:
Addresses: https://gitter.im/aws/copilot-cli?at=637244e418f21c023bb057cc

Previously, the logic to scale into Fargate Spot only applied if the `spot_from` value was greater than the `min` value. So in edge case of users having equal values greater than 1 for those fields, as in the case linked to above, all of the tasks would have Fargate Spot as their capacity provider, rather than the `spot_from` value determining the threshold for Fargate vs Fargate Spot. This kind of config is somewhat atypical, as the `min` value isn't guaranteed if that number task has Fargate Spot as a capacity provider, but if users do set up their scaling this way, at least we are honoring the `spot_from` value.

With this change:
`min: 3` and `spot_from: 3` --> 2 Fargate, 1 Spot
`min: 2` and `spot_from: 2` --> 1 Fargate, 1 Spot
`min: 1` and `spot_from: 1` --> 0 Fargate, 1 Spot
`min: 0` and `spot_from: 0` --> 0 Fargate, 1 Spot

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -205,11 +205,14 @@ func convertCapacityProviders(a manifest.AdvancedCount) []*template.CapacityProv
 	// Scaling with spot
 	spotFrom := aws.IntValue(rc.SpotFrom)
 	min := aws.IntValue(rc.Min)
-	// If spotFrom value is greater than or equal to the autoscaling min, 
+	// If spotFrom value is greater than or equal to the autoscaling min,
 	// then the base value on the Fargate capacity provider must be set
 	// to one less than spotFrom
 	if spotFrom >= min {
 		base := spotFrom - 1
+		if base < 0 {
+			base = 0
+		}
 		fgCapacity := &template.CapacityProviderStrategy{
 			Base:             aws.Int(base),
 			Weight:           aws.Int(0),
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -206,7 +206,7 @@ func convertCapacityProviders(a manifest.AdvancedCount) []*template.CapacityProv
 	spotFrom := aws.IntValue(rc.SpotFrom)
 	min := aws.IntValue(rc.Min)
 	// If spotFrom value is greater than or equal to the autoscaling min, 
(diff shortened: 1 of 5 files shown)


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -198,17 +198,17 @@ func convertCapacityProviders(a manifest.AdvancedCount) []*template.CapacityProv
...
 	spotFrom := aws.IntValue(rc.SpotFrom)
 	min := aws.IntValue(rc.Min)
-	// If spotFrom value is not equal to the autoscaling min, then
-	// the base value on the Fargate Capacity provider must be set
+	// If spotFrom value is greater than or equal to the autoscaling min, 
+	// then the base value on the Fargate capacity provider must be set
 	// to one less than spotFrom
-	if spotFrom > min {
+	if spotFrom >= min {
 		base := spotFrom - 1
```
efekarakus, line 212: huh so you can you can have `spotFrom` and `min` be 0? 
and it doesn't error if the base ends up being -1?

According to the commit message, we get:
>min: 0 and spot_from: 0 --> 0 Fargate, 1 Spot 

I think this is pretty strange, I would have expected 0 Fargate and 0 Spot here.

edit: I take this back 1 fargate spot makes sense if your cpu_percentage is like 70% because when you have 0 tasks the avg cpu is at 0%, so ECS has to scale up resulting in 1 spot task.

efekarakus: Side-note: should we make sure that `min`, `max` and `spot_from` is >= 0 here:
https://github.com/aws/copilot-cli/blob/5781c535bc78e5e3a544b855d930256562e52b63/internal/pkg/manifest/validate.go#L1077

efekarakus (commented):
Awesome, ty Janice!

iamhopaul123 (approved):

internal/pkg/deploy/cloudformation/stack/transformers.go
huanjani: Screen Shot 2022-11-16 at 3 33 02 PM

efekarakus: this has got to be a bug lol 😂 like what could -1 base mean conceptually

efekarakus (approved):

PR #4189, opened 2022-11-16
chore: generic union transformer and detect zero using reflection

dannyrandall:
This PR contains three changes:
1. A generic `Union` transformer - Union types no longer have to be added individually to the transformers list, as this transformer will handle all cases.
2. When unmarshalling, `Union` now uses `reflect.IsZero` to determine if a value hasn't been set. This means types used in a Union will need to be pointers if the zero value of a type means its been set (e.g., `key: false` will be considered unset unless the Union type is `*bool`)
3. ~Enables strict unmarshaling of types embedded in `Union`. This solves an issue where invalid yaml for a struct results in a zero value, requiring most struct types to have their own `IsZero` function to handle this case. For example, take this struct and unmarshal the below yaml into `Union[*string, complex]`:~


internal/pkg/manifest/transform.go
```diff
@@ -291,29 +292,35 @@ func (t subnetListOrArgsTransformer) Transformer(typ reflect.Type) func(dst, src
...
 
-func (t unionTransformer[Basic, Advanced]) Transformer(typ reflect.Type) func(dst, src reflect.Value) error {
-	if typ != reflect.TypeOf(Union[Basic, Advanced]{}) {
+// Transformer returns custom merge logic for union types.
+func (t unionTransformer) Transformer(typ reflect.Type) func(dst, src reflect.Value) error {
+	// :sweat_smile: https://github.com/golang/go/issues/54393
+	// reflect currently doesn't have support for getting type parameters
+	// or checking if a type is a non-specific instantiation of a generic type
+	// (i.e., no way to tell if the type Union[string, bool] is a Union)
+	isUnion := strings.HasPrefix(typ.String(), "manifest.Union[")
```
iamhopaul123: 🥲

internal/pkg/manifest/union.go
```diff
@@ -62,48 +66,64 @@ func (t Union[_, _]) IsAdvanced() bool {
...
-		return false
+// isZero returns true if:
+//   - v is a yaml.Zeroer and IsZero().
+//   - v is not a yaml.Zeroer and determined to be zero via reflection.
+func isZero(v any) bool {
+	if z, ok := v.(yaml.IsZeroer); ok {
+		return z.IsZero()
 	}
-	return z.IsZero()
+	return reflect.ValueOf(v).IsZero()
```
iamhopaul123, line 115: Could this panic 🤔

internal/pkg/manifest/union.go
```diff
@@ -62,48 +66,64 @@ func (t Union[_, _]) IsAdvanced() bool {
 }
 
 // UnmarshalYAML decodes value into either type Basic or Advanced, and stores that value
-// in t. value is first decoded into type Basic, and t will hold type Basic if:
-//   - There was no error decoding value into type Basic
-//   - Basic.IsZero() returns false OR Basic does not support IsZero()
+// in t. value is first decoded into type Basic, and t will hold type Basic if
```
iamhopaul123:
```suggestion
// in t. Value is first decoded into type Basic, and t will hold type Basic if
```

internal/pkg/manifest/union.go
```diff
@@ -62,48 +66,64 @@ func (t Union[_, _]) IsAdvanced() bool {
...
+		return err
+	}
 
-	sErr := value.Decode(&simple)
-	if sErr == nil && !isZeroerAndZero(simple) {
-		t.isBasic, t.Basic = true, simple
+	var basic Basic
+	dec := yaml.NewDecoder(bytes.NewReader(b))
+	dec.KnownFields(true)
+	bErr := dec.Decode(&basic)
```
iamhopaul123: Why do we do this 🤔 What will happen if we don't use `KnownFields`?

internal/pkg/manifest/transform.go
efekarakus: My slight concern with this is that doing a refactor and renaming the type could result in this breaking.

What do you think of the following? this way the programmer will be forced to refactor this declaration:
```go
var dummy Union[any,any]
typeName, _, _ := strings.Cut(reflect.TypeOf(dummy).String(), "[")
isUnion := strings.HasPrefix(typ.String(), typeName)
```

internal/pkg/manifest/transform.go
```diff
@@ -291,29 +292,35 @@ func (t subnetListOrArgsTransformer) Transformer(typ reflect.Type) func(dst, src
...
-		if srcStruct.IsBasic() {
-			var zero Advanced
-			dstStruct.isAdvanced, dstStruct.Advanced = false, zero
-			dstStruct.isBasic = true
-		} else if srcStruct.IsAdvanced() {
-			var zero Basic
-			dstStruct.isBasic, dstStruct.Basic = false, zero
-			dstStruct.isAdvanced = true
+		isBasic := src.MethodByName("IsBasic").Call(nil)[0].Bool()
+		isAdvanced := src.MethodByName("IsAdvanced").Call(nil)[0].Bool()
```
efekarakus, line 320: seems totally safe 😱 
I wonder if it it's worthwhile catching the whole block in a defer+recover?
Something like:
1. Recover
2. Return `nil` if we deem it's better to skip or return an error saying the field "could not be overriden please don't override it?"

I'm leaning more towards the error and suggest them not to override route?

internal/pkg/manifest/union.go
```diff
@@ -62,48 +66,64 @@ func (t Union[_, _]) IsAdvanced() bool {
...
 // decoding, t will not hold any value.
 func (t *Union[Basic, Advanced]) UnmarshalYAML(value *yaml.Node) error {
-	// reset struct
-	var simple Basic
-	var advanced Advanced
-	t.isBasic, t.Basic = false, simple
-	t.isAdvanced, t.Advanced = false, advanced
+	// convert value to []byte so we can use KnownFields()
+	// see https://github.com/go-yaml/yaml/issues/460
+	// side effect: TypeError line numbers are reset.
```
efekarakus: I don't know what this means :P can you elaborate in the comments?

internal/pkg/manifest/union.go
```diff
@@ -62,48 +66,64 @@ func (t Union[_, _]) IsAdvanced() bool {
...
 	}
 
-	aErr := value.Decode(&advanced)
-	if aErr == nil && !isZeroerAndZero(advanced) {
-		t.isAdvanced, t.Advanced = true, advanced
+	var advanced Advanced
+	dec = yaml.NewDecoder(bytes.NewReader(b))
+	dec.KnownFields(true)
+	aErr := dec.Decode(&advanced)
+	if aErr == nil && !isZero(advanced) {
```
efekarakus: I wonder if this works?
```suggestion
	if dec.Decode(&advanced) == nil && !isZero(advanced) {
```

efekarakus (commented):
Thanks Danny! Let me know what you think of the suggestions!

internal/pkg/manifest/union.go
dannyrandall: ideally, it's something you set at the top level and the setting propagates down to custom `UnmarshalYAML()` functions. However, due to https://github.com/go-yaml/yaml/issues/460, the `value` passed in to `UnmarshalYAML()` doesn't retain the setting.

If we don't set `KnownFields()`, and we were unmarshalling
```yaml
key:
  invalid_key: asdf
```
into `Union[string, aStruct]`, then the value of Union would be aStruct, even though it's zero.

---

writing this, i realized that this statement:
> then the value of Union would be aStruct, even though it's zero.

is no longer true, because we are now using `reflect.IsZero()` when checking if a value is zero, which would be true if the struct is full of zero value fields. i guess using `KnownFields()` is just a semantic change in that case - erroring if an invalid field is set.

internal/pkg/manifest/union.go
Lou1415926: Do you know if a pointer type always implements `yaml.IsZerorer`? Are we falling through `reflect` so that we don't have to implement `IsZero` for every single custom `struct`? If so, I wonder if it actually makes sense to require the developers to implement `IsZero` themselves?

dannyrandall: it is possible if you called `isZero(nil)` - but since it's a private function and only used here i think it's probably fine - happy to add some extra logic if you think we should though.

dannyrandall: > Do you know if a pointer type always implements yaml.IsZerorer

No, only if there is a function like `(c *Config) IsZero() bool`

> Are we falling through reflect so that we don't have to implement IsZero for every single custom struct

Yes, but also for comparison to builtin types. i.e., you can use `Union[string, []string]` without having to create a custom `string` type that has `IsZero()`

internal/pkg/manifest/union.go
dannyrandall: the error values are used later in the function!

internal/pkg/manifest/union.go
Lou1415926: Sounds good!

I was trying to  think of a scenario where a `struct`'s zero value determined by `reflect` isn't really the zero value that we agree with. If this is potentially possible, then it might make sense to figure out a way to force developers to implement `IsZero` instead of automatically falling through to `reflect`. But I don't think I have thought of any, so we  should be good here.

efekarakus (dismissed):

internal/pkg/manifest/union.go
```diff
@@ -62,48 +65,52 @@ func (t Union[_, _]) IsAdvanced() bool {
...
+		aErr = fmt.Errorf("is zero")
+	}
+	if bErr == nil {
+		bErr = fmt.Errorf("is zero")
 	}
-	return aErr
+
+	// multiline error because yaml.TypeError (which this likely is)
+	// is already a multiline error
+	return fmt.Errorf("%T: %s\n%T: %s", t.Basic, bErr, t.Advanced, aErr)
```
efekarakus: I'm trying to think what are the scenarios that can happen here. Do you think this structure would be more helpful for customers?

```go
switch {
case aErr == nil && bErr == nil: // The basic form and advanced form were empty.
  return errors.New("ambiguous value: neither the simple or advanced form for the field was set")
case aErr != nil:
   return fmt.Errorf("unmarshal to %T advanced form: %w", t.Advanced, aErr)
default:
  return fmt.Errorf("unmarshal to %T basic form: %w", t.Basic, bErr)
}
```

efekarakus (commented):
looks good to me! Thanks for reverting the changes, I think this looks great.

I just have one suggestion, let me know your thoughts!

internal/pkg/manifest/union.go
dannyrandall: i like the proposed error first case (both values are zero), but the others i'm not quite as sold on 🤔. Say the union type is `Union[*string, aStruct]`. Given the yaml:
```
key:
  keyNotInaStruct: value
```
then `bErr == "unable to unmarshal !!map into string"` and `aErr == nil`. the proposed error would state `"unmarshal to string: <error>"`, which is true, but might be confusing along with the context that they were likely _trying_ to use the advanced form. i figured printing both the error for the string unmarshal _and_ the reasoning that the advanced type is zero might be more helpful - though it might be nice to include the "basic" and "advanced" type distinction. what do you think?

efekarakus: oh I see what you mean 👍 that makes sense to me!!
Perhaps with just some slight tweak words but your reasoning sounds good:
```go
return fmt.Errorf("unmarshal to %T basic form: %v\nunmarshal to %T advanced form: %v", t.Basic, bErr, t.Advanced, aErr)
```

iamhopaul123 (approved):

efekarakus (approved):

Issue #4170, opened 2022-11-11
Proposal: Support multiple ports


paragbhingre:
Proposal for #1783 , #3594,  #4445 
**Problem Statement**

Today in Copilot, a container can expose only one port. Below is an example manifest. Requests on “/” to the ALB  will be routed to the frontend main container’s port `8080`.
```yaml
name: 'frontend'
type: 'Load Balanced Web Service'
 
image:
  build: './frontend/Dockerfile'
  port: 8080
  
http:
  path: '/'
```

Copilot is adding support for multiple ports. This feature will enable exposing multiple ports from the containers, which can be opened to the application load balancer or the network load balancer. 

**Proposal**

To expose multiple ports, we will configure additional listener rules through a new field called `additional_rules` . 

For example, say we want to expand the above manifest such that it opens up port `8081` on the main service container, and `8082` on the sidecar container, in addition to the existing port `8080`.

```yaml
# Example 1
name: 'frontend'
type: 'Load Balanced Web Service'
 
image:
  build: './frontend/Dockerfile'
  port: 8080
  
http:
  path: '/'
  additional_rules: # The new field "additional_rules",
    - target_port: 8081
      path: '/customerdb'
    - target_port: 8082
      target_container: sidecarA
      path: '/admin'
```

With this manifest, requests to “/” will still be routed to the main container’s port 8080. Requests to "/customerdb" will be route to the main container’s `8081`, and "/admin" to “sidecarA”‘s port `8082`.

It is also possible to configure the container port that handles the requests to “/”:

```yaml
# Example 2
name: 'frontend'
type: 'Load Balanced Web Service'
 
image:
  build: './frontend/Dockerfile'
  port: 8080 # Some port that's exposed internally and may not handle public `http` traffic
  
http:
  path: '/'
  target_port: 8083  # Defaulted to `image.port` (8080) in "Example 1".
  additional_rules:
    - target_port: 8081
      path: '/customerdb'
    - target_port: 8082
      target_container: sidecarA 
      path: '/admin'
```

With this manifest, requests to “/” will now be routed to the main container’s port `8083`, instead of `8080`.

This feature will also allow us to expose multiple ports of the container through NLB ports. For example, say we want to expose `8080` and `8081` on the main service container and `8085` on the sidecar container.

```yaml
name: 'frontend'
type: 'Load Balanced Web Service'

image:
  build: Dockerfile
 
http: false
nlb:
  additional_listeners: 
    - port: 8081/tcp
[...]

PR #4192, opened 2022-11-17
chore: add target_port field in http and refactor portmappings and Service Connect

paragbhingre:
This PR is a part of 1st milestone of multiple ports i.e. converting Cloudformation such as way that it accepts list instead of just one object/parameter for resources like PortMappings, LoadBalancers, TargetGroup etc.

In this PR specifically 
1. I have addressed converting PortMappings into a list of objects, 
2. And also introduced a new field inside `http` i.e. `target_port` using which we can start exposing 2 ports internally. We can expose 1 port from `image.port` and another port from `http.target_port`. This will unblock us for one of the scenarios where we want to expose 2 ports internally, which we can still make happen through taskdef override, but now we have `http.target_port` for the rescue.
3. Service connect changes to incorporate multiple ports. 

Related #4170 


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -67,7 +67,8 @@ var (
 )
 
 // convertSidecar converts the manifest sidecar configuration into a format parsable by the templates pkg.
-func convertSidecar(s map[string]*manifest.SidecarConfig) ([]*template.SidecarOpts, error) {
+func convertSidecar(s map[string]*manifest.SidecarConfig, lbmft *manifest.LoadBalancedWebService, bsmft *manifest.BackendService, sjmft *manifest.ScheduledJob, wsmft *manifest.WorkerService) ([]*template.SidecarOpts, error) {
```
iamhopaul123: Instead of having a giant function to handle all type of manifest. Can we have one for each manifest type and within each function we can extract common logic to avoid duplication? Plus, I feel like, for example lbws, we just need `LoadBalancedWebServiceConfig` as the only input.

paragbhingre: I get your point, but the trouble here is we are getting `manifest.SidecarConfig` as an input to this function and we need parent struct of it which is `LoadBalancedWebServiceConfig` which you mentioned in the later part. So even if we create one for each manifest type then also inside sidecar we need to find out which manifest type this sidecar belongs to? For that we need manifest type as an input to the `convertSidecar` function. Does that make sense? Or you are suggesting something different than what I have understood?

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
```diff
@@ -44,11 +46,18 @@
 {{- end}}
 {{- end}} {{/* end if eq .WorkloadType "Load Balanced Web Service"*/}}
 {{- if eq .WorkloadType "Backend Service"}}
-{{- if eq .HTTPTargetContainer.Name .WorkloadName}} 
+{{- if eq .HTTPTargetContainer.Name .WorkloadName}}
```
Lou1415926: This is probably unrelated to your change: this seems to say that we expose the port via `PortMappings` for a backend service only if the target container is the main container. I wonder why is that - can't backend service expose a port from a sidecar 🤔 ?

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
```diff
@@ -44,11 +46,18 @@
...
 {{- if and .ServiceConnect .SCFeatureFlag}} # remove when release
-  PortMappings: !If [ExposePort, [{ContainerPort: !Ref ContainerPort, Name: target}], !Ref "AWS::NoValue"]
+  PortMappings:
+    - !If
+      - ExposePort
```
Lou1415926: Do we still need this condition? Is `{{- if .PortMappings}}` equivalent?

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
```diff
@@ -28,11 +28,13 @@
 {{- end}}
 {{- if eq .WorkloadType "Load Balanced Web Service"}}
   PortMappings:
-    - ContainerPort: !Ref ContainerPort
-    {{- if and .ServiceConnect .SCFeatureFlag}} # remove when release
+    {{- range $portMapping := .PortMappings}}
+    - ContainerPort: {{$portMapping.ContainerPort}}
+    {{- if and $.ServiceConnect $.SCFeatureFlag}} # remove when release
     {{/* Multiple exposed port is not supported yet. Therefore, if the target container is the sidecar, the target port must be the one and only one port that the container exposes. */}}
```
Lou1415926: Is this comment still useful?

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -260,6 +261,22 @@ func (s *LoadBalancedWebService) Template() (string, error) {
 	return string(overriddenTpl), nil
 }
 
+func (s *LoadBalancedWebService) getPrimaryContainerPortMappings() []*template.PortMapping {
```
Lou1415926: Is it possible that we let manifest do all of this:
```
func (s *LoadBalancedWebServiceManifest) portMappings() []template.PortMapping {
  // Read `image.port`
  // Read `http.target_port`
  // Read `sidecars.port`
  // Produce a list of port mappings where the port number is unique.
}
```

Lou1415926 (commented):
Is my understanding correct that after this PR:
1. We still have only one target container
2. We still have only one target port (but the number is configurable now - it used to be `image.port` or the target sidecar's port)
3. We can expose multiple ports from multiple containers

paragbhingre:
> my understanding correct that after this PR:
> 
> 1. We still have only one target container - ✔️ 
> 2. We still have only one target port (but the number is configurable now - it used to be `image.port` or the target sidecar's port) - ✔️ 
> 3. We can expose multiple ports from multiple containers - With this PR customer will be able to expose one additional port of the target container from `target_port` field. To expose more that **two** ports they will still have to wait for `additional_rule` field to come out in the upcoming PRs

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
paragbhingre: Great question: It looks like the code is saying that we expose the port via PortMappings for a backend service only if the target container is the main container. But in fact, the previous code was saying that if the target container is main container then add new field `Name` in the `portmapping` because SC will use that field for the communication purpose otherwise just use portmapping without `Name` field in it. So I just kept that condition as is for now because with SC feature release this  `{{- if eq .HTTPTargetContainer.Name .WorkloadName}}` condition will go away and we will be able to set portmappings irrespective of anything. Does that make sense? I know it is little confusing but happy to explain offline as well.

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
paragbhingre: Yeah we can remove this comment, I will do it.

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
paragbhingre: Hmm, that sounds interesting idea to me 😍 
I can check if we can do that.

internal/pkg/deploy/cloudformation/stack/transformers.go
iamhopaul123: I think Im pretty sure the signature can be much simpler than this. It can be either

```golang
func convertLBWSSidecar(config *manifest.LoadBalancedWebServiceConfig) ([]*template.SidecarOpts, error)
```

or a generic one

```golang
func convertSidecars(config map[string]*manifest.SidecarConfig, opts ...additionalSidecarOpts) ([]*template.SidecarOpts, error) {
  ......
  for _, opt := range opts {
     opt(sidecars)
  }
  return sidecars, nil
}

type additionalSidecarOpts func(opts []*template.SidecarOpts)

func withAdditionalLBPortMapping(routeingRules) additionalSidecarOpts

// and `withAdditionalSCPortMapping` stuff like that
```
and in additionalSidecarOpts we define the logic how we can

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -208,6 +209,23 @@ func (s *BackendService) Template() (string, error) {
 	return string(overriddenTpl), nil
 }
 
+func (s *BackendService) getPrimaryContainerPortMappings() []*template.PortMapping {
```
efekarakus: nit:
```suggestion
func (s *BackendService) primaryContainerPortMappings() []*template.PortMapping {
```

efekarakus: I wonder if most of our "convert*" functions should just be very dumb type conversions from a `manifest` type to a `template` type 💭 
Like we should move most of our smart logic to the `manifest` type itself whereas the `stack` package is just responsible to convert `manifest` types to `template` types.

Here is what I'm thinking a bit more concretely:
```go
package manifest

type ExposedPort struct {
   ContainerName string  // The name of the container that exposes this port.
   Port          int     // The port number.
   Protocol      string  // Either "tcp" or "udp", empty means the default value that the underlying service provides.
}

func (cfg *BackendServiceConfig) ExposedPorts() []ExposedPort

----

package stack

// convertPrimaryExposedPorts filters the exposed ports to only the main container ports and translates to the CFN template port mapping type.
func convertPrimaryExposedPorts(ports []manifest.ExposedPorts) []template.PortMapping
```

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -208,6 +209,23 @@ func (s *BackendService) Template() (string, error) {
...
+	containerPort := aws.String(s.containerPort())
+	var portMappings []*template.PortMapping
+	if aws.StringValue(containerPort) != NoExposedContainerPort {
+		portMappings = append(portMappings, ConvertPortMapping(containerPort, aws.String("tcp"), s.name))
+	}
+	if s.isTargetPortDifferentThanPrimaryContainerPort(containerPort) {
```
efekarakus: hopefully this method will get removed and most of the remaining logic will be simpler after the change proposed above 👍

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -219,6 +237,11 @@ func (s *BackendService) httpLoadBalancerTarget() (targetContainer *string, targ
 		targetPort = s.manifest.Sidecars[aws.StringValue(targetContainer)].Port
 	}
 
+	// Route load balancer traffic to the target_port if mentioned.
+	if s.manifest.RoutingRule.TargetPort != nil {
+		targetPort = s.manifest.RoutingRule.TargetPort
+	}
```
efekarakus: Do we not need to figure out the `targetContainer` if a user specified a `targetPort`? 

My understanding is that we still need to provide the appropriate ContainerName in the `loadBalancers` setting of the service so we would need a way of looking the container name of the port.

If I have the following manifest for example:
```
image:
  build: './Dockerfile'
  port: 8080 
  
http:
  path: '/'
  target_port: 80

sidecars:
  nginx:
    port: 80
```
In this situation, I would expect the ContainerName to be set to nginx rather than the main container because my `target_port` matches the exposed port in the manifest.

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -227,7 +250,7 @@ func (s *BackendService) containerPort() string {
 	if s.manifest.BackendServiceConfig.ImageConfig.Port != nil {
 		port = strconv.FormatUint(uint64(aws.Uint16Value(s.manifest.BackendServiceConfig.ImageConfig.Port)), 10)
 	}
-
+	// TODO: @pbhingre add another else condition to check additional_rules != nil when implemented in the upcoming PRs for multiple ports*/
```
efekarakus: nit: can we remove this TODO, since you'll be touching this file anyway

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
efekarakus: Yay!  Wanxian's feedback looks coincidentally like my comment above.

My slight tweak is that I don't think the `manifest` pkg should depend on `template` types.
The `template` types like `template.PortMapping` in my mind are tightly coupled with CFN/AWS API.
Whereas the `manifest` methods should return I think its own types that can then be converted to other data types, like CFN or Terraform or ...

internal/pkg/deploy/cloudformation/stack/transformers.go
efekarakus: If my intuition is correct, I think this function also get simplified with the refactor suggested by Wanxian and I above. The logic will be:

```go
var out []*template.PortMapping
for _, name := range sidecarNames {
   for _, port := range exposedPorts {
       if port.ContainerName != name {
           continue
       }
       out = append(out, convertExposedPort(port))
   }
}
```

So there won't be a need to distinguish between types because each manifest type will implement the `ExposedPorts` method.

internal/pkg/template/template_functions.go
```diff
@@ -140,6 +140,12 @@ func generateMountPointJSON(mountPoints []*MountPoint) string {
 
 }
 
+// PortMappingName returns name for the port mapping.
+func PortMappingName(containerName string, containerPort string) string {
+	fmt.Println("I was here ", containerName+containerPort)
```
efekarakus:
```suggestion
```

internal/pkg/template/template_functions.go
```diff
@@ -140,6 +140,12 @@ func generateMountPointJSON(mountPoints []*MountPoint) string {
...
+	return containerName + containerPort
```
efekarakus: Is it okay according to the ECS API to have digits in the name of a port?

paragbhingre: yes it is allowed - The characters can include lowercase letters, numbers, underscores (_), and hyphens (-). A hyphen can't be the first character.

internal/pkg/deploy/cloudformation/stack/backend_svc.go
paragbhingre: Yes you are correct. I am making that change so that target container is set to the right container name. 
Also just wants to make sure my understanding aligns with yours. There could be following scenarios --
1. if `target_container` is set then target container's port is target port. 
2. if `target_port` is set then we will check if `target_container` is set? if yes, then we know target container and target port. If `target_container` is not set then we will check if sidecar container port is same as that of `target_port` and if found then set that sidecar container name as target container name. Otherwise we will set main container name as target container name.

internal/pkg/deploy/cloudformation/stack/backend_svc.go
paragbhingre: I have made changes as per your suggestion and moved all the smart logic inside manifest.

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: I have made changes as per the discussion here. Do let me know if my changes infers your expectations.

internal/pkg/deploy/cloudformation/stack/backend_svc.go
efekarakus: Your summary sounds good to me :shipit:

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -271,6 +277,29 @@ func (s *LoadBalancedWebService) Template() (string, error) {
...
+	for _, portConfig := range exposedPorts {
+		if portConfig.ContainerName == s.name {
+			primaryPorts = append(primaryPorts, portConfig)
+		} else {
+			sidecarPorts = append(sidecarPorts, portConfig)
+		}
+	}
+
+	return ConvertPortMapping(primaryPorts), ConvertPortMapping(sidecarPorts), nil
+}
```
iamhopaul123: Can we make it a separate function to reduce duplication?

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -212,6 +239,18 @@ func (s *BackendService) httpLoadBalancerTarget() (targetContainer *string, targ
...
+	if s.manifest.RoutingRule.TargetPort != nil {
+		targetPort = aws.String(strconv.Itoa(aws.IntValue(s.manifest.RoutingRule.TargetPort)))
+		exposedPorts, err := s.manifest.ExposedPorts()
+		if err == nil {
+			for _, portConfig := range exposedPorts {
+				if portConfig.Port == aws.IntValue(s.manifest.RoutingRule.TargetPort) {
+					targetContainer = aws.String(portConfig.ContainerName)
+				}
+			}
+		}
```
iamhopaul123: I feel like if they only specify target port we should just assume target container is the main container 🤔 And we need manifest validation making sure they don't specify the same port.

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -201,6 +207,27 @@ func (s *BackendService) Template() (string, error) {
 	return string(overriddenTpl), nil
 }
 
+func (s *BackendService) containerPortMappings() ([]*template.PortMapping, []*template.PortMapping, error) {
```
iamhopaul123: how about making the output to be a map? so the primary container port mapping is `output[workloadName]`. And also we can make `convertSidecar` much more efficient.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -81,6 +82,12 @@ func convertSidecar(s map[string]*manifest.SidecarConfig) ([]*template.SidecarOp
 
 	var sidecars []*template.SidecarOpts
 	for _, name := range keys {
+		var sidecarPortMappings []*template.PortMapping
+		for _, portConfig := range portMappings {
```
iamhopaul123: Maybe think twice when it is O(n^2) unless we can't mitigate this.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -112,11 +119,27 @@ func convertSidecar(s map[string]*manifest.SidecarConfig) ([]*template.SidecarOp
...
 			HealthCheck:  convertContainerHealthCheck(config.HealthCheck),
 			Command:      command,
+			PortMappings: sidecarPortMappings,
 		})
 	}
 	return sidecars, nil
 }
 
+// ConvertPortMapping return the port mapping object based on the given input.
+func ConvertPortMapping(exposedPorts []manifest.ExposedPort) []*template.PortMapping {
```
iamhopaul123: Why do we make this a public func? Also this signature looks odd. I would expect

```suggestion
func ConvertPortMapping(exposedPort manifest.ExposedPort) *template.PortMapping {
```

otherwise most likely we are gonna loop twice.

iamhopaul123 (commented):
Can we split this PR into smaller ones? Like one for manifest and one for stack changes, one for CFN template. Just to be safe 🥺

internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
```diff
@@ -45,17 +45,21 @@
   Essential: {{$sidecar.Essential}}
 {{- end}}
 {{include "image-overrides" . | indent 2}}  
-{{- if $sidecar.Port}}
+{{- if and $sidecar.PortMappings (gt (len $sidecar.PortMappings) 0)}}
```
Lou1415926: Is this equivalent to `{{- if $sidecar.PortMappings }}` ?

internal/pkg/template/workload.go
```diff
@@ -387,6 +388,13 @@ type NetworkLoadBalancer struct {
 	MainContainerPort string
 }
 
+// PortMapping holds configuration that exposes container ports internally.
+type PortMapping struct {
+	Protocol      *string
+	ContainerPort *string
+	Name          string
```
Lou1415926, line 395: Am I understanding correctly that `Name` holds the name of the container? If so, can we rename the field to `ContainerName`? I thought at first this was the port mapping's name and got confused when reading the template.

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
```diff
@@ -28,10 +28,11 @@
 {{- end}}
 {{- if eq .WorkloadType "Load Balanced Web Service"}}
   PortMappings:
-    - ContainerPort: !Ref ContainerPort
-    {{/* Multiple exposed port is not supported yet. Therefore, if the target container is the sidecar, the target port must be the one and only one port that the container exposes. */}}
-    {{- if eq .HTTPTargetContainer.Name .WorkloadName}}
-      Name: target
+    {{- range $portMapping := .PortMappings}}
+    - ContainerPort: {{$portMapping.ContainerPort}}
+    {{- if eq $.HTTPTargetContainer.Name $.WorkloadName}}
```
Lou1415926, line 33: Just making sure: is my understanding correct that we only assign names to ports if the current container is the target container?

internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
```diff
@@ -45,17 +45,21 @@
...
   PortMappings:
-    - ContainerPort: {{$sidecar.Port}}
+    {{range $portMapping := $sidecar.PortMappings}}
+    {{- if eq $portMapping.Name $sidecar.Name}}
```
Lou1415926: qq: is this if statement redundant?

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -212,6 +239,18 @@ func (s *BackendService) httpLoadBalancerTarget() (targetContainer *string, targ
 		targetPort = s.manifest.Sidecars[aws.StringValue(targetContainer)].Port
 	}
 
+	// Route load balancer traffic to the target_port if mentioned.
+	if s.manifest.RoutingRule.TargetPort != nil {
+		targetPort = aws.String(strconv.Itoa(aws.IntValue(s.manifest.RoutingRule.TargetPort)))
+		exposedPorts, err := s.manifest.ExposedPorts()
+		if err == nil {
```
Lou1415926: Is it safe to swallow the error here?

Or it is, we could just do `exposedPorts, _ := s.manifest.ExposedPorts()`

internal/pkg/deploy/cloudformation/stack/backend_svc.go
paragbhingre: [Here](https://github.com/aws/copilot-cli/pull/4192#discussion_r1042810508) is some explanation on what are the assumptions we want consider when only `target_port` is mentioned. Do let me know if that solves your concern.

Lou1415926: What would happen if a user has an incorrectly configured manifest?
```yaml
http:
  path: '/'
  target_port: 80
  target_container: another_sidecar

sidecars:
  nginx:
    port: 80
  another_sidecar:
    port: 8080
```



@paragbhingre --> I think in this case target_container would be another_sidecar and `target_port` would be 80 which is the wrong configuration but then we need to take care of this during validation stage. Even in NLB this configuration can be set and application would not work even though service will be up and running.

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
paragbhingre: Sorry but I did not understand what duplication you are referring to in here? Could you please elaborate?

iamhopaul123: everything after we get all the exposed ports from the manifest https://github.com/aws/copilot-cli/pull/4192/files#diff-7e36dd6cfea6499215e00fda7bd632bec4a52ffec8428ebe43981553b39c0cd2R287-R300 because they'll not be relavant to the manifest type anymore

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: I think I can use map for portMappings and get rid of this looping. I will do that.

internal/pkg/manifest/backend_svc.go
```diff
@@ -99,6 +101,61 @@ func (s *BackendService) requiredEnvironmentFeatures() []string {
...
+		if cfg.RoutingRule.TargetContainer != nil {
+			exposedPorts[targetPort] = ExposedPort{
+				Port:          targetPort,
+				Protocol:      "tcp",
+				ContainerName: aws.StringValue(cfg.RoutingRule.TargetContainer),
+			}
+
+		} else {
+			// if target_container is nil then by default set container name as main container name.
+			exposedPorts[targetPort] = ExposedPort{
```
Lou1415926: It's quite tricky that this block is required to be executed before L146 in order for us to compute the correct set of exposed ports!

We probably wants to either add a comment or write the logic in an order so that it's clear that "we only assume the target port is exposed on the main container if the target container is explicitly main, or the target port isn't exposed in any sidecar"

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -282,6 +311,18 @@ func (s *LoadBalancedWebService) httpLoadBalancerTarget() (targetContainer *stri
 		targetPort = s.manifest.Sidecars[aws.StringValue(targetContainer)].Port
 	}
 
+	// Route load balancer traffic to the target_port if mentioned.
+	if s.manifest.RoutingRule.TargetPort != nil {
```
efekarakus, line 291: similar to Penghao's comment above, can we think of ways of reducing this duplication? for example we can have a function like:
```go
// findContainerNameGivenPort searches through exposed ports in the manifest to find the container name that matches the given port.
// If no port is found, then an error is returned. 
func findContainerNameGivenPort(port string, exposedPorts []manifest.ExposedPort) (string, error)
```

This function might be also nice to capture our default: if there is no match, the main workload container is assumed.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -67,7 +67,8 @@ var (
 )
 
 // convertSidecar converts the manifest sidecar configuration into a format parsable by the templates pkg.
-func convertSidecar(s map[string]*manifest.SidecarConfig) ([]*template.SidecarOpts, error) {
+func convertSidecar(s map[string]*manifest.SidecarConfig, portMappings []*template.PortMapping) ([]*template.SidecarOpts, error) {
```
efekarakus: I know this isn't related to your change, but can we rename the `map` variable 😅 it doesn't tell us much. Perhaps just:
```suggestion
func convertSidecar(sidecars map[string]*manifest.SidecarConfig, portMappings []*template.PortMapping) ([]*template.SidecarOpts, error) {
```

internal/pkg/deploy/cloudformation/stack/worker_svc.go
```diff
@@ -180,3 +185,15 @@ func (s *WorkerService) Parameters() ([]*cloudformation.Parameter, error) {
 func (s *WorkerService) SerializedParameters() (string, error) {
 	return serializeTemplateConfig(s.wkld.parser, s)
 }
+
+func (s *WorkerService) containerPortMappings() ([]*template.PortMapping, error) {
```
efekarakus: similar to previous duplication comments, can we think of ways such that workerService and scheduledJob uses the same function?

internal/pkg/manifest/backend_svc.go
```diff
@@ -99,6 +101,61 @@ func (s *BackendService) requiredEnvironmentFeatures() []string {
 	return features
 }
 
+// ExposedPorts returns all the ports that are container ports available to receive traffic.
+func (cfg *BackendService) ExposedPorts() ([]ExposedPort, error) {
+	exposedPorts := make(map[int]ExposedPort)
```
efekarakus: I don't quite follow why we need the map, can't we just use the slice?

I see that we have validation that ensures that no two ports can be declared twice in the manifest, due to that can this method just assume that ports are unique?

internal/pkg/manifest/job.go
```diff
@@ -192,3 +194,23 @@ func newDefaultScheduledJob() *ScheduledJob {
 		Environments: map[string]*ScheduledJobConfig{},
 	}
 }
+
+// ExposedPorts returns all the ports that are container ports available to receive traffic.
+func (cfg *ScheduledJob) ExposedPorts() ([]ExposedPort, error) {
```
efekarakus: Huh can jobs expose any ports? do we allow that today I wouldn't have expected that

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -70,6 +72,63 @@ type LoadBalancedWebServiceProps struct {
 	Platform    PlatformArgsOrString // Optional platform configuration.
 }
 
+// ExposedPorts returns all the ports that are container ports available to receive traffic.
+func (cfg *LoadBalancedWebService) ExposedPorts() ([]ExposedPort, error) {
```
efekarakus: Is this identical to backend service, can we get rid of this duplication?
For example, a function that takes as input the image, http, and sidecar configurations.

internal/pkg/manifest/validate.go
```diff
@@ -1707,6 +1729,29 @@ func validateContainerDeps(opts validateDependenciesOpts) error {
...
+	for _, sidecar := range opts.sidecarConfig {
+		if sidecar.Port != nil {
+			port, err := strconv.Atoi(aws.StringValue(sidecar.Port))
+			if err != nil {
+				return err
+			}
+			if _, ok := exposedPorts[uint16(port)]; !ok {
+				exposedPorts[uint16(port)] = true
+			} else {
+				return fmt.Errorf("multiple containers are exposing the same port")
```
efekarakus: is there any chance we can tell them which two container is conflicting?

internal/pkg/manifest/worker_svc.go
```diff
@@ -42,6 +44,26 @@ func (s *WorkerService) subnets() *SubnetListOrArgs {
 	return &s.Network.VPC.Placement.Subnets
 }
 
+// ExposedPorts returns all the ports that are container ports available to receive traffic.
+func (cfg *WorkerService) ExposedPorts() ([]ExposedPort, error) {
```
efekarakus: Similar to jobs, I think worker services shouldn't allow ports to be exposed 💭

internal/pkg/template/workload.go
paragbhingre: you are right, actually I am contemplating between changing the name of the parameter to ContiainerName or just keep the value of the Name to be partmapping name i.e., containerName + port. But I think doing later one is correct and depicts the right thing. So I will change the calue of the Name to  containerName + port while converting `ExposedPorts` from manifest to `PortMappings` in template.

internal/pkg/template/templates/workloads/partials/cf/workload-container.yml
paragbhingre: Great question Wan: for now until we don't support multiple ports this is true because HTTPTargetContainer is the only container who is exposing port that too only one port. But when multiple ports come into the picture then we will remove this condition as HTTPTargetContainer is not the only container who is exposing the ports. does that make sense?

internal/pkg/deploy/cloudformation/stack/backend_svc.go
paragbhingre: I think it is safe to sqallow error because the `ExposedPorts()` is throwing the same error that [validExposedPort](https://github.com/aws/copilot-cli/pull/4192/files#diff-5fc13795122fe030799a09ef956a8b66cec3d40bf1a81266323edbf3626b5685R1732) is throwing. So I think if error has to come then it will be surfaced during validation phase itself.

internal/pkg/manifest/backend_svc.go
paragbhingre: Umm, yeah it is kinda tricky to understand what are the use cases being tackled here with the logic. I think changing the order of reading ports from different places will mess up the use cases. So I will just put better comments to make reader understand why is the code written that way.

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: it should be list because we are looping over it to convert all the exposed ports to port mappings. I will make the function private.

internal/pkg/manifest/backend_svc.go
paragbhingre: So the reason why I took this as map because we are writing some logic here that determines the correct set of exposed ports for example, if `target_port` is given and `target_container` is nil then we try our best to figure out the `target_container` from the `target_port`. In that case, in the for loop of sidecars  if `target_port` is same as that of the sidecar port then we set the exposedPorts[target_port] to sidecar container which was earlier set as `exposedPorts[target_port] = primary container`. That is how we are making a use of the map here. Else we will have to write extra logic to avoid duplication of the exposedPorts entry. Does that make sense?

internal/pkg/manifest/job.go
paragbhingre: I was also surprised to see that Jobs and Worker service has sidecars with the `port` field on it. I was not sure how customers must be writing their sidecars in terms of jobs and worker services hence I decide keep its logic consistent with LBWS and Backend Service.

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
paragbhingre: we already set main workload container as default while creating the exposedPorts. So this method will only will have logic to get the container name based on the target port.

internal/pkg/deploy/cloudformation/stack/scheduled_job.go
```diff
@@ -136,7 +136,11 @@ func (j *ScheduledJob) Template() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	sidecars, err := convertSidecar(j.manifest.Sidecars)
+	portMappings, err := containerPortMappings(j.name, nil, nil, nil, j.manifest.Sidecars)
```
efekarakus, line 139: If we catch ourselves writing functions/methods where we have to pass `nil` values or booleans, typically we should go into suspicion-mode 🤨 and think if there is an alternative way of writing it: https://martinfowler.com/bliki/FlagArgument.html

For example here this might highlight that there is a need for more functions:
```go
func portMappingsWithLoadBalancer(name, port, targetPort, targetContainer, sidecars)
func portMappings(name, port, sidecars)
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -112,11 +117,55 @@ func convertSidecar(s map[string]*manifest.SidecarConfig) ([]*template.SidecarOp
...
 }
 
+// findContainerNameGivenPort searches through exposed ports in the manifest to find the container name that matches the given port.
+func findContainerNameGivenPort(port int, exposedPorts []manifest.ExposedPort) *string {
+	for _, portConfig := range exposedPorts {
+		if portConfig.Port == port {
+			return aws.String(portConfig.ContainerName)
+		}
+	}
+	return nil
```
efekarakus, line 133: Why are we returning a `*string` instead of a `string`? 
Should the default be the name container if the port field is not present in the manifest? If that's already the case then I think `string` captures better the idea that the returned value is not optional.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -112,11 +117,55 @@ func convertSidecar(s map[string]*manifest.SidecarConfig) ([]*template.SidecarOp
...
+	for _, exposedPort := range exposedPorts {
+		port := aws.String(strconv.Itoa(exposedPort.Port))
+		if config, ok := portMapping[exposedPort.ContainerName]; ok {
+			portMapping[exposedPort.ContainerName] = append(portMapping[exposedPort.ContainerName], &template.PortMapping{
+				ContainerPort: port,
+				Protocol:      aws.String(exposedPort.Protocol),
+				Name:          exposedPort.ContainerName + aws.StringValue(port),
+			})
+		} else {
+			portMapping[exposedPort.ContainerName] = append(config, &template.PortMapping{
```
efekarakus, line 159: wut, it doesn't make sense to use `config` in the `else` block no? can't we eliminate the if-else statements:
When we see duplication like this, let's try to find ways to see if it can be eliminated 💭 
In this scenario, we can see both the `if` and `else` statements have the same declaration, that means we can move this declaration outside of it:
```go
for _, in := range exposedPorts {
   port := strconv.Itoa(in.Port)
   out := &template.PortMapping{
      ContainerPort: aws.String(port),
      Protocol: aws.String(in.Protocol),
      Name: in.ContainerName + port,
   }
   portMapping[in.ContainerName] = append(portMapping[in.ContainerName], out) // should work even when nil
}
```

efekarakus: As a side-note, we should 🤨 when we see an `else` statement. Else-statements are great if we need the performance benefit but typically we can write the code like this:
```go
value := commonCase()
if condition {
   value = specialCase()
}
// process value...
```
Instead of:
```
if condition {
  value := specialCase()
  // process value...
} else {
  value := commonCase()
  // process value...
}
```

internal/pkg/manifest/svc.go
```diff
@@ -55,6 +55,64 @@ func (r *Range) Parse() (min int, max int, err error) {
 	return aws.IntValue(r.RangeConfig.Min), aws.IntValue(r.RangeConfig.Max), nil
 }
 
+// ExposedPorts returns all the ports that are container ports available to receive traffic.
+func ExposedPorts(workloadName string, port *uint16, targetPort *int, targetContainer *string, sidecars map[string]*SidecarConfig) ([]ExposedPort, error) {
```
efekarakus, line 59: Long parameter lists is also a code smell. I'll quote from the [Refactoring book from Martin Fowler](https://www.amazon.com/Refactoring-1st-first-Text-Only/dp/B004PQQRK2):

> Long parameter lists are hard to understand, because they become inconsistent and difficult to use, and because you are forever changing them as you need more data.

The solutions to this are:
> Use _Replace Parameter with Method_ when you can get the data in one parameter by making a request of an object you already know about. This object might be a field or it might be another parameter.
> Use _Preserve Whole Object_ to take a bunch of data glened from an object and replace it with the object itself.
> If you have several data items with no logical object, use _Introduce Parameter Object_.

In our situation, your original submission was the way to go with `Replace Parameter with Method`:

```go
func (lbws *LoadBalancedWebService) ExposedPorts() ([]ExposedPort, error)
func (worker *WorkerService) ExposedPorts() ([]ExposedPort, error)
```

You can remove duplication by then having each sub-type implement its own `exportPorts` method:
```go
func (lbws *LoadBalancedWebService) ExposedPorts() ([]ExposedPort, error) {
  var exposedPorts []ExposedPort
  exposedPorts = append(exposedPorts, lbws.ImageConfig.exposedPorts()...)
  exposedPorts = append(exposedPorts, lbws.RoutingRule.exposedPorts()...)
  exposedPorts = append(exposedPorts, lbws.Sidecars.exposedPorts()...)
  return exposedPorts
}


func (ws *WorkerService) ExposedPorts() ([]ExposedPort, error) {
  var exposedPorts []ExposedPort
  exposedPorts = append(exposedPorts, lbws.ImageConfig.exposedPorts()...)
  exposedPorts = append(exposedPorts, lbws.Sidecars.exposedPorts()...)
  return exposedPorts
}
```

efekarakus (commented):
This PR is really big 🥺 It results in the reviewer to either:
1. Leave only high level feedback
2. Or if I want to leave detailed feedback, it takes too long to review.

Can we please close the PR and send a separate PR per package?
Here is a rough outline:
1. Refactor: `template` pkg and cfn go templates under `templates/` 
2. Refactor: `manifest` pkg
3. Refactor: `cloudformation/stack` transformers
2. Refactor: ` deploy` if any changes are needed

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: I will keep this thing in mind.

PR #4193, opened 2022-11-17
chore: reference values from a different stack for env vars from service manifest

Lou1415926:
This PR adds `from_environment` under `variables` field.

Should probably get merged only after #4189 so that `Union` type can be used in this PR.

Support #4220 


internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
```diff
@@ -52,6 +52,8 @@ environments:
     command: crazy diamond
     variables:
       LOG_LEVEL: debug        # Log level for the "test" environment.
+      DB_NAME:
```
Lou1415926, line 57: <strike>This change and `svc-manifest`'s `nginx` sidecar manifest change are not reflected in   `rdws-test.stack.yml` or `svc-prod.stack.yml` respectively. They are kept out by the feature flag. This is to show that they are indeed kept out.</strike>

LBWS has the feature flag, so LBWS integ test on `prod` env should have it enable on main container, `nginx` sidecar and logging container.

Job on the other hand does not have the feautre flag enabled. Hence even though `from_cfn` appears in its manifest, it does not render in the CFN template.

internal/pkg/template/templates/workloads/partials/cf/envvars-container.yml
```diff
@@ -1,6 +1,16 @@
 {{if .Variables}}{{range $name, $value := .Variables}}
+{{if $.EnvAddonsFeatureFlag}}
 - Name: {{$name}}
-  Value: {{$value | printf "%q"}}{{end}}{{end}}
+  Value: {{if $value.RequiresImport}}
+    Fn::ImportValue: {{$value.Value | printf "%q" -}}
```
iamhopaul123: can we use quote?

```suggestion
    Fn::ImportValue: {{quote $value.Value -}}
```

internal/pkg/template/templates/workloads/partials/cf/envvars-container.yml
```diff
@@ -1,6 +1,16 @@
 {{if .Variables}}{{range $name, $value := .Variables}}
+{{if $.EnvAddonsFeatureFlag}}
 - Name: {{$name}}
-  Value: {{$value | printf "%q"}}{{end}}{{end}}
+  Value: {{if $value.RequiresImport}}
```
iamhopaul123: maybe we can make it to be a partial file for easier maintenance 🤔

iamhopaul123 (commented):
LGTM overall only two nits.

internal/pkg/deploy/cloudformation/stack/lb_web_service_integration_test.go
```diff
@@ -114,6 +114,7 @@ func TestLoadBalancedWebService_TemplateInteg(t *testing.T) {
 			},
 		})
 		serializer.SCFeatureFlag = true
+		serializer.EnvAddonsFeatureFlag = true
```
efekarakus, line 116: Is the advantage of putting the feature flag at the `stack` lvl the fact that we can have our integ tests done ahead of time and just remove these lines once the feature is ready?

internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-prod.stack.yml
```diff
@@ -87,6 +87,9 @@ Resources: # If a bucket URL is specified, that means the template exists.
               Value: '{"givesdogs":"arn:aws:sns:us-west-2:123456789123:my-app-prod-fe-givesdogs","mytopic.fifo":"arn:aws:sns:us-west-2:123456789123:my-app-prod-fe-mytopic.fifo"}'
             - Name: COPILOT_LB_DNS
               Value: !GetAtt EnvControllerAction.PublicLoadBalancerDNSName
+            - Name: DB_NAME
+              Value:
+                Fn::ImportValue: "MyDB"
```
efekarakus, line 92: oh interesting 🤔 Did you decide to not automatically include the `!Sub "${App}-${Env}-` prefix for import value?

The reason why I'm mentioning is this is that to me the current implemenation looks more like:
```yaml
variables:
  DB_NAME:
    from_cfn_output: MyDB
```
There is nothing necessarily typing it back to the "environment" concept.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -997,6 +997,21 @@ func convertHTTPVersion(protocolVersion *string) *string {
 	return &pv
 }
 
+func convertEnvVars(variables map[string]manifest.Variable) map[string]template.Variable {
+	if len(variables) == 0 {
+		return nil
+	}
+	m := make(map[string]template.Variable)
```
```suggestion
	m := make(map[string]template.Variable, len(variables))
```

internal/pkg/manifest/applyenv_test.go
```diff
@@ -790,3 +790,175 @@ func TestApplyEnv_MapToPStruct(t *testing.T) {
 		})
 	}
 }
+
+func TestApplyEnv_MapToStruct(t *testing.T) {
```
efekarakus, line 794: nit: maybe just
```suggestion
func TestApplyEnv_EnvVars(t *testing.T) {
```

internal/pkg/manifest/svc.go
```diff
@@ -416,3 +416,35 @@ func ParsePortMapping(s *string) (port *string, protocol *string, err error) {
 		return nil, nil, fmt.Errorf("cannot parse port mapping from %s", *s)
 	}
 }
+
+type fromEnvironment struct {
+	Name *string `yaml:"from_environment"`
```
efekarakus: I always forget, do we opt-in to `*string` over `string` due to mergo?

internal/pkg/manifest/validate.go
```diff
@@ -1599,6 +1604,10 @@ func (r OverrideRule) validate() error {
 	return nil
 }
 
+func (v Variable) validate() error {
+	return nil
```
efekarakus, line 1631: do we need to check if `from_environment`'s length is > 0?

internal/pkg/manifest/validate.go
```diff
@@ -819,6 +819,11 @@ func (t TaskConfig) validate() error {
 	if err = t.Storage.validate(); err != nil {
 		return fmt.Errorf(`validate "storage": %w`, err)
 	}
+	for n, v := range t.Variables {
+		if err := v.validate(); err != nil {
+			return fmt.Errorf(`validate %q "variables"  : %w`, n, err)
```
```suggestion
			return fmt.Errorf(`validate %q "variables": %w`, n, err)
```

internal/pkg/template/workload.go
```diff
@@ -205,7 +206,7 @@ type LogConfigOpts struct {
 	EnableMetadata *string
 	SecretOptions  map[string]Secret
 	ConfigFile     *string
-	Variables      map[string]string
+	Variables      map[string]Variable
```
efekarakus: Was there a reason why you preferred using a `Variable` interface instead of making the `envVar` type public and getting rid of the interfaces?

internal/pkg/deploy/cloudformation/stack/lb_web_service_integration_test.go
Lou1415926: yeah. an alternative is putting on `variables` level, but in order to have the integ test done for the feature, we'd need to pass this feature flag across multiple packages - which is error prone when the time comes to remove them.

internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-prod.stack.yml
Lou1415926: Good question! Discussed offline and this new design here #4220 is the result. This PR will be updated accordingly.

internal/pkg/manifest/svc.go
Lou1415926: yesssss

internal/pkg/template/workload.go
Lou1415926: This is just so that the template YAML files would have a sort-of-stable interface to interact with. We can change the underlying implementation of the `envVar`'s method but the interface does not change. The benefit is marginal right now, but could be significant in the future if/when we let users bring their own go template. Plus this is largely inspired by https://github.com/aws/copilot-cli/blob/2d3a3f8036941d4f366fc0888b03c7f8eb573f3b/internal/pkg/template/workload.go#L255

internal/pkg/manifest/applyenv_test.go
Lou1415926: The intention of the naming is that if this unit test pass, then we should be confident to say that other `mapToStruct` would be correctly overridden as well.  `EnvVars` is sort of like a representative 😂 here... This way we can test that `applyEnv` is applying necessary `mergo` transformers correctly, without having verbose tests for every single possible fields.

internal/pkg/deploy/cloudformation/stack/lb_web_service_integration_test.go
Lou1415926: I've changed the feautre flag to be container-level!

internal/pkg/template/templates/workloads/partials/cf/envvars-container.yml
Lou1415926: done!

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
```diff
@@ -101,8 +101,17 @@ Resources:
               {{- end }}
               {{- if .Variables}}
               {{- range $name, $value := .Variables}}
+              {{if $.EnvAddonsFeatureFlag}}
```
iamhopaul123: Can we not apply the same vairables.yml to rdws?

internal/pkg/deploy/cloudformation/stack/lb_web_service_integration_test.go
efekarakus: Interesting 🤔 do you think there is a benefit for container level? I feel like the previous setup was simpler and equivalent.
Sidecars should still be able to reference `EnvAddonsFeatureFlag` and decide whether a field should be supported or not

internal/pkg/manifest/svc.go
```diff
@@ -410,3 +410,35 @@ func ParsePortMapping(s *string) (port *string, protocol *string, err error) {
 		return nil, nil, fmt.Errorf("cannot parse port mapping from %s", *s)
 	}
 }
+
+type fromCFN struct {
+	Name *string `yaml:"from_cfn"`
+}
+
+func (e *fromCFN) isEmpty() bool {
```
efekarakus: nit: 😂 why `e` for the receiver?
```suggestion
func (cfg *fromCFN) isEmpty() bool {
```

internal/pkg/template/workload.go
```diff
@@ -252,6 +255,48 @@ func (h HTTPHealthCheckOpts) IsHTTPS() bool {
...
+}
+
+type envVar struct {
+	value      string
+	importName string
+}
+
+// RequiresImport returns true if the environment variable is imported from a stack.
+func (v envVar) RequiresImport() bool {
+	return v.importName != ""
```
efekarakus: nit: I wonder if instead of putting two mutually exclusive fields in a single struct we can split the struct:
```go
type plainEnvVar string
func (var plainEnvVar) RequiresImport() bool { return false }
func (var plainEnvVar) Value() string { return var }

type importedEnvVar string
func (var importedEnvVar) RequiresImport() bool { return true }
func (var importedEnvVar) Value() string { return var } 
```

efekarakus (commented):
Looks great! just some tiny feedback, let me know what you think

internal/pkg/deploy/cloudformation/stack/lb_web_service_integration_test.go
Lou1415926: > Sidecars should still be able to reference EnvAddonsFeatureFlag and decide whether a field should be supported or not

This is not quite the case unless we change how we organize the `partials`s, or modify `include` template function such that it is able to handle multiple parameters. I am not sure if the benefit overcomes the amount of application code change that need to be made.

For example, when the feature flag is service-level, the sidecar section won't be able to read the feature flag data because what's passed into the `envvars-container`  partial template is the sidecar container's data, not the service's.

efekarakus (approved):

iamhopaul123 (approved):

PR #4201, opened 2022-11-21
chore: sc should work with no exposed port

iamhopaul123:

### DIFF
--- a/internal/pkg/manifest/validate.go
+++ b/internal/pkg/manifest/validate.go
@@ -204,9 +204,9 @@ func (b BackendService) validate() error {
 	}); err != nil {
 		return fmt.Errorf("validate HTTP load balancer target: %w", err)
 	}
-	if b.Network.Connect.Enabled() {
+	if b.Network.Connect.Alias != nil {
 		if b.RoutingRule.GetTargetContainer() == nil && b.ImageConfig.Port == nil {
-			return fmt.Errorf(`cannot enable "network.connect" when no port exposed`)
+			return fmt.Errorf(`cannot set "network.connect.alias" when no port exposed`)
 		}
 	}
 	if err = validateContainerDeps(validateDependenciesOpts{
@@ -1083,8 +1083,8 @@ func (r RangeConfig) validate() error {
 	min, max, spotFrom := aws.IntValue(r.Min), aws.IntValue(r.Max), aws.IntValue(r.SpotFrom)
 	if min < 0 || max < 0 || spotFrom < 0 {
 		return &errRangeValueLessThanZero{
-			min: min,
-			max: max,
+			min:      min,
+			max:      max,
 			spotFrom: spotFrom,
 		}
 	}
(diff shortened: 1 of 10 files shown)


internal/pkg/manifest/validate.go
```diff
@@ -204,9 +204,9 @@ func (b BackendService) validate() error {
 	}); err != nil {
 		return fmt.Errorf("validate HTTP load balancer target: %w", err)
 	}
-	if b.Network.Connect.Enabled() {
+	if b.Network.Connect.Alias != nil {
 		if b.RoutingRule.GetTargetContainer() == nil && b.ImageConfig.Port == nil {
-			return fmt.Errorf(`cannot enable "network.connect" when no port exposed`)
+			return fmt.Errorf(`cannot set "network.connect.alias" when no port exposed`)
```
```suggestion
			return fmt.Errorf(`cannot set "network.connect.alias" when no ports are exposed`)
```

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
```diff
@@ -43,6 +43,7 @@ ServiceConnectConfiguration:
       awslogs-region: !Ref AWS::Region
       awslogs-group: !Ref LogGroup
       awslogs-stream-prefix: copilot
+  {{- if and (ne .HTTPTargetContainer.Port "") (ne .HTTPTargetContainer.Port "-1")}}
```
efekarakus: Should this be a method on the `HTTPTargetContainer` type?
```go
// Exposed returns true if the target container has an accessible port to receive traffic.
func (tg HTTPTargetContainer) Exposed() bool
```

efekarakus (approved):
Looks great, thank you boss! 
Just two minor changes. Don't block on me, so :shipit: and remove the do-not-merge label afterwrads

internal/pkg/manifest/validate.go
```diff
@@ -204,9 +204,9 @@ func (b BackendService) validate() error {
 	}); err != nil {
 		return fmt.Errorf("validate HTTP load balancer target: %w", err)
 	}
-	if b.Network.Connect.Enabled() {
+	if b.Network.Connect.Alias != nil {
 		if b.RoutingRule.GetTargetContainer() == nil && b.ImageConfig.Port == nil {
-			return fmt.Errorf(`cannot enable "network.connect" when no port exposed`)
+			return fmt.Errorf(`cannot set "network.connect.alias" when no ports are exposed`)
```
dannyrandall, line 209: shouldn't this error also happen on `network.connect = true`? or no?

dannyrandall (commented):
looks good, but just a question!

internal/pkg/manifest/validate.go
iamhopaul123: nope only for `alias`. because they can still turn SC with no exposed endpoint so that it is a client-only service.

dannyrandall (approved):

PR #4202, opened 2022-11-22
chore(workspace): fix `Use` to return an error if the ws has no summary 

efekarakus:
Currently, running `copilot app init` in an empty workspace results in the creation of a .workspace file with an empty "" app name. This is because `Use` did not error on empty workspace and incorrectly assumed that an empty `copilot/` directory was a valid workspace which resulted in the app name to be empty.
Instead, we want to prompt for an app name when the workspace is empty.

### DIFF
--- a/internal/pkg/workspace/errors.go
+++ b/internal/pkg/workspace/errors.go
@@ -44,11 +44,19 @@ func (e *ErrWorkspaceNotFound) Error() string {
 		e.CurrentDirectory)
 }
 
+func (_ *ErrWorkspaceNotFound) empty() bool {
+	return true
+}
+
 // ErrNoAssociatedApplication means we couldn't locate a workspace summary file.
 type ErrNoAssociatedApplication struct{}
 
 func (e *ErrNoAssociatedApplication) Error() string {
 	return "couldn't find an application associated with this workspace"
+}
+
+func (_ *ErrNoAssociatedApplication) empty() bool {
+	return true
 }
 
 // errHasExistingApplication means we tried to create a workspace that belongs to another application.
@@ -64,4 +72,13 @@ func (e *errHasExistingApplication) Error() string {
 		relPath = e.summaryPath
 	}
 	return fmt.Sprintf("workspace is already registered with application %s under %s", e.existingAppName, relPath)
+}
+
+// IsEmptyErr returns true if the error is related to an empty workspace.
+func IsEmptyErr(err error) bool {
+	type empty interface {
+		empty() bool
+	}
+	var emptyWs empty
+	return errors.As(err, &emptyWs)
 }
--- a/internal/pkg/workspace/errors.go
+++ b/internal/pkg/workspace/errors.go
@@ -7,6 +7,8 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
+
+	"github.com/aws/copilot-cli/internal/pkg/term/color"
 )
 
 // ErrNoPipelineInWorkspace means there was no pipeline manifest in the workspace dir.
@@ -44,9 +46,13 @@ func (e *ErrWorkspaceNotFound) Error() string {
 		e.CurrentDirectory)
 }
 
-func (_ *ErrWorkspaceNotFound) empty() bool {
-	return true
+// RecommendActions suggests steps clients can take to mitigate the copilot/ directory not found error.
+func (_ *ErrWorkspaceNotFound) RecommendActions() string {
+	return fmt.Sprintf("Run %s to create an application.", color.HighlightCode("copilot app init"))
 }
+
+// empty denotes that this error represents an empty workspace.
+func (_ *ErrWorkspaceNotFound) empty() {}
 
 // ErrNoAssociatedApplication means we couldn't locate a workspace summary file.
 type ErrNoAssociatedApplication struct{}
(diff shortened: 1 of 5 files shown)


internal/pkg/workspace/errors.go
```diff
@@ -65,3 +73,12 @@ func (e *errHasExistingApplication) Error() string {
 	}
 	return fmt.Sprintf("workspace is already registered with application %s under %s", e.existingAppName, relPath)
 }
+
+// IsEmptyErr returns true if the error is related to an empty workspace.
+func IsEmptyErr(err error) bool {
+	type empty interface {
+		empty() bool
+	}
```
dannyrandall: this is a really cool idea! 😍

dannyrandall: the only thing i wonder - it's kinda like "is empty if the function exists", so maybe it could just be `empty()` (no bool)?

also, prob don't need to define a type for `empty`, just:
```
var emptyWs interface {
  empty()
}
```

internal/pkg/workspace/workspace.go
```diff
@@ -147,19 +158,23 @@ func Create(appName string, fs afero.Fs) (*Workspace, error) {
 
 // Summary returns a summary of the workspace. The method assumes that the workspace exists and the path is known.
 func (ws *Workspace) Summary() (*Summary, error) {
-	summaryPath := filepath.Join(ws.copilotDirAbs, SummaryFileName) // Assume `copilotDirAbs` is always present.
-	summaryFileExists, _ := ws.fs.Exists(summaryPath)               // If an err occurs, return no applications.
-	if summaryFileExists {
-		value, err := ws.fs.ReadFile(summaryPath)
+	ws.summarizeOnce.Do(func() {
```
dannyrandall, line 158: it's so beautiful

internal/pkg/workspace/workspace.go
```diff
@@ -60,16 +62,20 @@ const (
...
 // Workspace typically represents a Git repository where the user has its infrastructure-as-code files as well as source files.
 type Workspace struct {
 	workingDirAbs string
 	copilotDirAbs string
-	fs            *afero.Afero
-	logger        func(format string, args ...interface{})
+
+	summary       *Summary
+	summaryErr    error
+	summarizeOnce sync.Once
```
dannyrandall, line 72: maybe a comment that these should only be accessed via `Summary()`, not directly

internal/pkg/workspace/workspace.go
```diff
@@ -60,16 +62,20 @@ const (
...
 
 // Workspace typically represents a Git repository where the user has its infrastructure-as-code files as well as source files.
 type Workspace struct {
 	workingDirAbs string
 	copilotDirAbs string
-	fs            *afero.Afero
-	logger        func(format string, args ...interface{})
+
+	summary       *Summary
+	summaryErr    error
```
Lou1415926, line 71: I'm probably missing some thing important - why do we want to store `summaryErr` in `workspace` ?

Lou1415926 (commented):
Thanks for the fix! LGTM just a small question.

iamhopaul123 (approved):
💯

internal/pkg/workspace/errors.go
efekarakus: Done!

internal/pkg/workspace/workspace.go
efekarakus: Done!

internal/pkg/workspace/workspace.go
efekarakus: We're now using [`sync.Once`](https://pkg.go.dev/sync#Once) in the [Summary method](https://github.com/aws/copilot-cli/pull/4202/files/c47847b0d5e92c2b67851f261c478462bb28aa2a#diff-8646fd3cdd4c58b0a1c3c3c09348d53c9fe5d6231b922f87f91c02f90b57f8d2R161) to read the `.workspace` file only once.
So we need to store its results in the struct: the successful parsed file or any error that occurred while reading the file.
This way, when `Summary` is invoked again then we can return the [saved values](https://github.com/aws/copilot-cli/pull/4202/files/c47847b0d5e92c2b67851f261c478462bb28aa2a#diff-8646fd3cdd4c58b0a1c3c3c09348d53c9fe5d6231b922f87f91c02f90b57f8d2R177).

If we don't save the error, say we have:
```go
var summaryErr error
summarizeOnce.Do(func() {
    // existing logic...
})
return ws.summary, summaryErr
```
Then the first time a caller invokes `Summary` we will return an error appropriately, but the second time the error will be `nil` which is not what we want.


Does this help?

internal/pkg/workspace/errors_test.go
```diff
@@ -0,0 +1,41 @@
...
+		"should return true when ErrWorkspaceNotFound": {
+			err:    &ErrWorkspaceNotFound{},
+			wanted: true,
+		},
+		"should return true when ErrNoAssociatedApplication": {
+			err:    &ErrNoAssociatedApplication{},
+			wanted: true,
+		},
+		"should return true when an empty workspace error is wrapped": {
+			err:    fmt.Errorf("burrito: %w", &ErrNoAssociatedApplication{}),
```
dannyrandall, line 28: :joy:

dannyrandall (approved):

PR #4203, opened 2022-11-22
chore: remove unnecessary health check `IsZero()` function

dannyrandall:
Due to changes in #4189, a `*string` is now required to determine if `healthcheck` has been configured using the basic configuration.

### DIFF
--- a/internal/pkg/manifest/svc.go
+++ b/internal/pkg/manifest/svc.go
@@ -368,13 +368,13 @@ type HTTPHealthCheckArgs struct {
 // HealthCheckArgsOrString is a custom type which supports unmarshaling yaml which
 // can either be of type string or type HealthCheckArgs.
 type HealthCheckArgsOrString struct {
-	Union[*string, HTTPHealthCheckArgs]
+	Union[string, HTTPHealthCheckArgs]
 }
 
 // Path returns the default health check path if provided otherwise, returns the path from the advanced configuration.
 func (hc *HealthCheckArgsOrString) Path() *string {
 	if hc.IsBasic() {
-		return hc.Basic
+		return aws.String(hc.Basic)
 	}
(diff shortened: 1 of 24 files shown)


internal/pkg/manifest/svc.go
```diff
@@ -365,22 +365,16 @@ type HTTPHealthCheckArgs struct {
 	GracePeriod        *time.Duration `yaml:"grace_period"`
 }
 
-// IsZero returns true if all of h's fields are unset. Implements yaml.IsZeroer.
-func (h *HTTPHealthCheckArgs) IsZero() bool {
```
efekarakus, line 369: aww yea

efekarakus (commented):
The code looks good but I don't quite follow why this change is necessary 🤔 

1. If I follow correctly, **now**:
```yaml
http:
   path: '/'
   healthcheck: ''
```
will result in healthcheck having the path "", whereas if we kept it as `string` then it results in an error because healthcheck is zero and advanced is an error. Is that correct?

2. How can we add validation on the basic `*string` type? so that we don't allow empty value. Would that require a custom type like `type nonEmptyStringPointer *string`

dannyrandall:
Mmm that's a fair point - if we don't want to allow the empty string anyways, then we can leave it as `string`. However, in previous versions of Copilot we do allow `healthcheck: ""`, so maybe we should leave it as allowed for backwards compatibility?

dannyrandall:
Actually I just tested it - although we did allow it, the CFN fails :) So I guess #4189 was a bug fix!🤓 I'll change it back to `string` here.

huanjani (approved):
This felt like a sheep in wolf's clothing or something. 😆

efekarakus (approved):
💥

PR #4213, opened 2022-11-23
feat: set placement field to private in both Service and Job manifest

KollaAdithya:
1. Currently, when the environment has only `private subnets` and the customer wants to deploy a service in that environment. By default, the `network.vpc.placement` is set to public. Customers are experiencing failed deployments because the placement is set to public by default.

2. To fix this issue, this PR includes implementation of setting `placement: private` in the `environments` section of Service manifest and Job manifest by checking all the environments in the application to see whether they have only private subnets. In this way, we will override the placement to private in the `environments` section.

3. For example, let's say we have two environments test, prod that are deployed, where the test environment has only private subnets. When you run `copilot svc init` for any Service available in Copilot or `copilot job init` for a Scheduled Job, the environments section in Service or Job manifest contains:

```yaml
environments:
  test:
    network:
      vpc:
        placement: 'private'
```
     


internal/pkg/cli/job_init.go
```diff
@@ -79,6 +80,8 @@ type initJobOpts struct {
 
 	// Init a Dockerfile parser using fs and input path
 	initParser func(string) dockerfileParser
+	// Init a new EnvDescriber using environment name.
+	envDescriber func(string) (envDescriber, error)
```
huanjani: Ah, I like this much better than where you had it before!

internal/pkg/cli/job_init.go
```diff
@@ -211,6 +225,35 @@ func (o *initJobOpts) Ask() error {
 	return nil
 }
 
+// isSubnetsOnlyPrivate returns the list of environments names deployed that contains only private subnets.
+func (o *initJobOpts) isSubnetsOnlyPrivate() ([]string, error) {
```
huanjani: nit: what do you think of `hasOnlyPrivateSubnets()` as a method name instead?

internal/pkg/template/templates/workloads/jobs/scheduled-job/manifest.yml
```diff
@@ -46,7 +46,21 @@ platform: {{.Platform.PlatformString}}   # See https://aws.github.io/copilot-cli
...
-#    cpu: 2048               # Larger CPU value for prod environment 
+#    cpu: 2048               # Larger CPU value for prod environment.
+{{- else}}
+
+# You can override any of the values defined above by environment.
+environments: {{ range $key, $value := .Environments}} 
+  {{$key}}:
+    network:
+      vpc:
+        placement: '{{$value.Network.VPC.Placement.PlacementString}}' # Placement is set to private for environment "{{$key}}" containing only private subnets.
```
huanjani: nit (here and below):
```suggestion
        placement: '{{$value.Network.VPC.Placement.PlacementString}}' # Placement is set to 'private' for environment "{{$key}}", which contains only private subnets.
```

internal/pkg/template/templates/workloads/jobs/scheduled-job/manifest.yml
```diff
@@ -46,7 +46,21 @@ platform: {{.Platform.PlatformString}}   # See https://aws.github.io/copilot-cli
 #secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store.
 #  GITHUB_TOKEN: GITHUB_TOKEN  # The key is the name of the environment variable, the value is the name of the SSM parameter.
 
+{{- if not .Environments}}
```
huanjani, line 49: I wonder if we should make this more specific, in case we add other config to `.Environments`? Or we could change it later.

huanjani (commented):
I've just done a first read-through. Great tests! Thanks for remembering to do the integ tests, too!

internal/pkg/template/templates/workloads/jobs/scheduled-job/manifest.yml
KollaAdithya: I will  change this one to be more specific.

KollaAdithya: _Correct if am wrong:_

Thinking about it more, `{{- if not .Environments}}` makes sense because we have `Environments : map[string]*LoadBalancedWebServiceConfig`. So, whenever we have `Environments`  empty, we generate the default manifest file, the same as before, which supports backward compatibility.

If we have to add any other config just like `Placement` in future, we can add those configs in the `else` case when the Environments is not empty.

internal/pkg/cli/job_init.go
```diff
@@ -121,6 +124,17 @@ func newInitJobOpts(vars initJobVars) (*initJobOpts, error) {
 		initParser: func(path string) dockerfileParser {
 			return dockerfile.New(fs, path)
 		},
+		envDescriber: func(envName string) (envDescriber, error) {
+			envDescriber, err := describe.NewEnvDescriber(describe.NewEnvDescriberConfig{
```
Lou1415926: Is it possible that `vars.appName` is `""` now, until `Validate()` where we assign `o.appName` with `o.wsAppName`? 

If a value is only determined after `Validate/Ask` (like `envName` right now), we should probably pass them as parameters:
```
envDescriber: func(appName, envName string) (envDescriber, error) {
```

lmk if this makes sense!

internal/pkg/cli/job_init.go
```diff
@@ -121,6 +124,17 @@ func newInitJobOpts(vars initJobVars) (*initJobOpts, error) {
...
+				App:         vars.appName,
+				Env:         envName,
+				ConfigStore: store,
+			})
+			if err != nil {
+				return nil, err
```
Lou1415926: We probably can skip wrapping an error if the `err` message is already clear enough. Is this the case?

internal/pkg/cli/job_init.go
```diff
@@ -214,6 +228,35 @@ func (o *initJobOpts) Ask() error {
 	return nil
 }
 
+// hasOnlyPrivateSubnets returns the list of environments names deployed that contains only private subnets.
+func (o *initJobOpts) hasOnlyPrivateSubnets() ([]string, error) {
+	envs, err := o.store.ListEnvironments(o.appName)
+	if err != nil {
+		return nil, err
```
Lou1415926: We should probably wrap the error here!

internal/pkg/cli/job_init.go
```diff
@@ -214,6 +228,35 @@ func (o *initJobOpts) Ask() error {
 	return nil
 }
 
+// hasOnlyPrivateSubnets returns the list of environments names deployed that contains only private subnets.
+func (o *initJobOpts) hasOnlyPrivateSubnets() ([]string, error) {
```
Lou1415926: The name `hasOnlyPrivateSubnets` sounds like the method should return a `bool`. What do you think of something like `envsWithPrivateSubnetsOnly` or `privateSubnetsOnlyEnvs`.

internal/pkg/cli/job_init.go
```diff
@@ -80,6 +81,8 @@ type initJobOpts struct {
 
 	// Init a Dockerfile parser using fs and input path
 	initParser func(string) dockerfileParser
+	// Init a new EnvDescriber using environment name.
+	envDescriber func(string) (envDescriber, error)
```
iamhopaul123:
```suggestion
	initEnvDescriber func(string) (envDescriber, error)
```

internal/pkg/cli/job_init.go
```diff
@@ -214,6 +228,35 @@ func (o *initJobOpts) Ask() error {
...
 }
 
+// hasOnlyPrivateSubnets returns the list of environments names deployed that contains only private subnets.
+func (o *initJobOpts) hasOnlyPrivateSubnets() ([]string, error) {
+	envs, err := o.store.ListEnvironments(o.appName)
+	if err != nil {
+		return nil, err
+	}
+	var privateOnlyEnvs []string
+	for i := range envs {
```
Lou1415926: Generally, I think it's safer to range via ⬇️   because `env` would be a "copy" of `envs[i]`. 
```go
for _, env := range envs{
   envDescriber, err := o.envDescriber(env.Name)
}
```

This prevents you from unexpectedly changing the value of `envs[i]`. For example:
```go
printEnvName(envs) // print "test", "prod"

for _, env := range envs{
   env.Name = "prod"
}

printEnvName(envs) // still print "test", "prod"


for i := range envs{
   envs[i].Name = "prod"
}

printEnvName(envs) // print "prod", "prod"

```

internal/pkg/cli/job_init.go
```diff
@@ -214,6 +228,35 @@ func (o *initJobOpts) Ask() error {
...
+		envDescriber, err := o.envDescriber(envs[i].Name)
+		if err != nil {
+			return nil, fmt.Errorf("initiate env describer: %w", err)
+		}
```
Lou1415926: I think generally, we wrap an error that is returned from an exported method in another package, and we do not wrap an error otherwise. If we want to wrap this error anyway, it's probably better to wrap it inside of `o.envDescriber()` instead of here, because this way we'd be able to adhere to the error wrapping agreement.

internal/pkg/cli/job_init.go
```diff
@@ -214,6 +228,35 @@ func (o *initJobOpts) Ask() error {
...
+		envConfig, err := manifest.UnmarshalEnvironment(mft)
+		if err != nil {
+			return nil, fmt.Errorf("unmarshal the manifest used to deploy environment %s: %w", envs[i].Name, err)
+		}
+		subnets := envConfig.Network.VPC.Subnets
+
+		if len(subnets.Public) == 0 && len(subnets.Private) != 0 {
+			privateOnlyEnvs = append(privateOnlyEnvs, envs[i].Name)
+		}
```
Lou1415926: At some point we will need to refactor the code so that the manifest structure isn't exposed everywhere 😞

internal/pkg/cli/job_init.go
```diff
@@ -235,6 +278,11 @@ func (o *initJobOpts) Execute() error {
 			o.platform = &platform
 		}
 	}
+	// Environments that are deployed have​ only private subnets.
```
Lou1415926: We *should* be able to get rid of this comment if we give `hasOnlyPrivateSubnets` a more descriptive name!

internal/pkg/cli/job_init_test.go
```diff
@@ -550,12 +550,25 @@ type: Scheduled Job`), nil)
 }
 
 func TestJobInitOpts_Execute(t *testing.T) {
+	mockmanifest := []byte(`name: test
```
Lou1415926:
```suggestion
	mockEnvironmentManifest := []byte(`name: test
```

internal/pkg/cli/job_init_test.go
```diff
@@ -701,6 +741,17 @@ func TestJobInitOpts_Execute(t *testing.T) {
 					},
 				}).Return("manifest/path", nil)
 			},
+			mockStore: func(m *mocks.Mockstore) {
+				m.EXPECT().ListEnvironments("sample").Return([]*config.Environment{
```
Lou1415926: I think to simplify these mocking in unit tests, you can return
```go
m.EXPECT().ListEnvironments("sample").Return(nil, nil)
```
It's entirely possible that when people run `copilot svc init` there are no existing environment yet.

internal/pkg/cli/job_init.go
iamhopaul123: It seems like this method is very similar to the one in `svc_init`, can we try to reduce the duplication?

internal/pkg/cli/job_init.go
```diff
@@ -214,6 +228,35 @@ func (o *initJobOpts) Ask() error {
...
+func (o *initJobOpts) hasOnlyPrivateSubnets() ([]string, error) {
+	envs, err := o.store.ListEnvironments(o.appName)
+	if err != nil {
+		return nil, err
+	}
+	var privateOnlyEnvs []string
+	for i := range envs {
+		envDescriber, err := o.envDescriber(envs[i].Name)
+		if err != nil {
+			return nil, fmt.Errorf("initiate env describer: %w", err)
```
iamhopaul123: no need to wrap this error from private function.

internal/pkg/manifest/job.go
```diff
@@ -67,11 +67,12 @@ type JobFailureHandlerConfig struct {
...
-	Timeout     string
-	HealthCheck ContainerHealthCheck // Optional healthcheck configuration.
-	Platform    PlatformArgsOrString // Optional platform configuration.
-	Retries     int
+	Schedule                string
+	Timeout                 string
+	HealthCheck             ContainerHealthCheck // Optional healthcheck configuration.
+	Platform                PlatformArgsOrString // Optional platform configuration.
+	Retries                 int
+	PrivateOnlyEnvironments []string
```
Lou1415926: If this applies to all workloads, then it probably makes sense to be a field of `WorkloadProps`

internal/pkg/template/templates/workloads/jobs/scheduled-job/manifest.yml
```diff
@@ -46,7 +46,23 @@ platform: {{.Platform.PlatformString}}   # See https://aws.github.io/copilot-cli
...
+#    cpu: 2048               # Larger CPU value for prod environment.
+{{- else}}
+
+# You can override any of the values defined above by environment.
+environments: {{ range $key, $value := .Environments}} 
+  {{$key}}:
+{{- if $value.Network.VPC.Placement.PlacementString}}
+    network:
+      vpc:
+        placement: '{{$value.Network.VPC.Placement.PlacementString}}' # Placement is set to 'private' for environment "{{$key}}", which contains only private subnets.
```
Lou1415926: Adding the comment is nice!! We could probably reword it to something like "# The service will be placed in private subnets...."

internal/pkg/initialize/workload.go
```diff
@@ -60,14 +60,15 @@ type Prog interface {
...
-	Queue          manifest.SQSQueue
+	App                     string
+	Type                    string
+	Name                    string
+	DockerfilePath          string
+	Image                   string
+	Platform                manifest.PlatformArgsOrString
+	Topics                  []manifest.TopicSubscription
+	Queue                   manifest.SQSQueue
+	PrivateOnlyEnvironments []string
```
iamhopaul123, line 71: Maybe in the future if we are doing more "magic" setting to the manifest (per environment), we could further extend this field to something like

```golang
Environments map[string]EnvOverrideSetting
```

and then unwrap it in the manifest package. But I think right now it is good enough.

internal/pkg/template/templates/workloads/services/rd-web/manifest.yml
```diff
@@ -66,8 +66,25 @@ network:
...
+{{- if not .Environments}}
+
 # You can override any of the values defined above by environment.
 # environments:
 #   test:
 #     variables:
 #       LOG_LEVEL: debug        # Log level for the "test" environment.
+{{- else}}
+
+# You can override any of the values defined above by environment.
```
Lou1415926: This is probably not needed for RDWS because https://aws.github.io/copilot-cli/docs/manifest/rd-web-service/#network-vpc-placement

Lou1415926 (commented):
Nice work - Thank you Adi!

internal/pkg/manifest/backend_svc.go
```diff
@@ -62,6 +63,17 @@ func NewBackendService(props BackendServiceProps) *BackendService {
 		svc.BackendServiceConfig.TaskConfig.Memory = aws.Int(MinWindowsTaskMemory)
 	}
 	svc.parser = template.New()
+	for _, envName := range props.PrivateOnlyEnvironments {
```
iamhopaul123, line 65: <strike>Can we make it a function so that we don't repeat ourselves? or like i mentioned earlier we can make `PrivateOnlyEnvironments` a field not dedicated to private subnets setting only and then we can have a function to apply the env override setting to the manifest struct.</strike>

It's ok I think it'll make things more complicated.

iamhopaul123 (approved):
LGTM great work!

internal/pkg/cli/svc_init.go
```diff
@@ -327,6 +341,11 @@ func (o *initSvcOpts) Execute() error {
 			o.platform = &platform
 		}
 	}
+	// Environments that are deployed have​ only private subnets.
```
huanjani: nit:
```suggestion
	// Environments that are deployed and have​ only private subnets.
```

huanjani (approved):
Added the DNM label for that one tiny nit. Feel free to remove it when you're ready!

PR #4214, opened 2022-11-23
chore: appinit sub command in init

KollaAdithya:
Currently, when you run `copilot init` it errors out because iam and iamRoleManager are not initialized.

### DIFF
--- a/internal/pkg/cli/init.go
+++ b/internal/pkg/cli/init.go
@@ -110,7 +110,7 @@ func newInitOpts(vars initVars) (*initOpts, error) {
 	spin := termprogress.NewSpinner(log.DiagnosticWriter)
 	id := identity.New(defaultSess)
 	deployer := cloudformation.New(defaultSess, cloudformation.WithProgressTracker(os.Stderr))
-	iam := iam.New(defaultSess)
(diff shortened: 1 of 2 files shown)


internal/pkg/cli/init.go
```diff
@@ -109,6 +110,7 @@ func newInitOpts(vars initVars) (*initOpts, error) {
 	spin := termprogress.NewSpinner(log.DiagnosticWriter)
 	id := identity.New(defaultSess)
 	deployer := cloudformation.New(defaultSess, cloudformation.WithProgressTracker(os.Stderr))
+	iam := iam.New(defaultSess)
```
Lou1415926: Just noticed the block from L114-116 😂 Probably an orphan as a result of moving codes around. Can we remove that block plzz?

Lou1415926 (approved):
LGTM just one minor request!

internal/pkg/cli/init.go
iamhopaul123: Can we avoid using the package name as the var name? for both init and app init 🙏

iamhopaul123 (approved):

PR #4217, opened 2022-11-23
chore: add sc alias validation for worker svc

iamhopaul123:

### DIFF
--- a/internal/pkg/manifest/validate.go
+++ b/internal/pkg/manifest/validate.go
@@ -204,11 +204,6 @@ func (b BackendService) validate() error {
 	}); err != nil {
 		return fmt.Errorf("validate HTTP load balancer target: %w", err)
 	}
-	if b.Network.Connect.Alias != nil {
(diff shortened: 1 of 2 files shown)


huanjani (approved):

internal/pkg/manifest/validate.go
```diff
@@ -252,6 +247,11 @@ func (b BackendServiceConfig) validate() error {
 	if err = b.Network.validate(); err != nil {
 		return fmt.Errorf(`validate "network": %w`, err)
 	}
+	if b.Network.Connect.Alias != nil {
+		if b.RoutingRule.GetTargetContainer() == nil && b.ImageConfig.Port == nil {
```
efekarakus, line 251: This might be of interest to @paragbhingre.
If we do the changes here: https://github.com/aws/copilot-cli/pull/4192#discussion_r1026933083

Then we can replace this section with `if len(manifest.ExposedPorts()) > 0` and that would greatly simplify things I think

efekarakus (approved):

Issue #4219, opened 2022-11-23
Support Environment Addons


Lou1415926:
### Problem Statement
Today, Copilot supports workload-level addons. With this feature, user can add to a service/job [additional AWS resources](https://aws.github.io/copilot-cli/docs/developing/addons/modeling/) that are not integrated to a service/job manifest by default. 

This works nice for resources that are meant to be tied to a single service.  However, there are resources such as AppMesh or Load Balancers that are designed to be shared by multiple consumers; and resources whose lifecycle should not be tied to that of a single service, but rather an environment. 

To address this, we propose the environment addons feature - a user experience for associating additional AWS resources with environments and hence across workloads.

### Proposal

The proposal can be broken down into three parts: 
1. Deploying environment addons 
2. Using values from environment addons in a workload-level addon 
3. Using values from environment addons in a service/job.

#### Deploying Environment Addons

The experience to deploy environment addons will be mirroring that of the existing workload-level addons:

1. Users will write addon templates as a CloudFormation template in YAML format.
2. Users can [customize the `Parameters` section](https://aws.github.io/copilot-cli/docs/developing/addons/modeling/#customizing-the-parameters-section).
5. Users will then store the addon templates in `copilot/environments/addon/myaddon.yml` folder.
6. User will run `copilot env deploy` to deploy an environment along with the addons.
7. `copilot env package --upload-assets` will upload the addon templates. User will be able to deploy environment addons in their pipelines.

##### File Structure

To elaborate on 2. here is an example. Suppose we have a "test" environment, a "dev" environment and a "web" service. Now we want to create an `AppMesh` addon and a few `WAF` rules on environment level:
```shell
copilot/
├── environments/ 
│   ├── addons/                # <- You can define multiple addons. 
│   │     ├── appmesh.yml      # <- The `AppMesh` addon.
│   │     └── waf.yml          # <- The WAF rules.  
│   ├── test/ 
│   │     └─── manifest.yml 
│   └── dev/ 
│         └── manifest.yml
└── web
    ├── addons/
    │     └── appmesh_node.yml
    └─── manifest.yml 
  ```

All environments (in this example, both "test" and "dev") will share the same addon templates. Just like today’s workload-level addons, any environment-specific configuration should be specified inside of the addon templates via the `Conditions` and `Mappings` sections. This follows [CFN’s best practice of reusing templates](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/best-practices.html#reuse):
```yaml
[...]

PR #4221, opened 2022-11-23
chore: parse and package environment addons

Lou1415926:
This PR enable parsing of environment addons, as long as customized parameters and packaging.

Related: #4219 


internal/pkg/addon/addons.go
```diff
@@ -33,49 +39,115 @@ var (
...
 	ListFiles(dirPath string) ([]string, error)
 	ReadFile(fPath string) ([]byte, error)
 }
 
-// Stack represents a CloudFormation stack.
-type Stack struct {
-	template     *cfnTemplate
-	parameters   yaml.Node
+// WorkloadAddonStack represents a CloudFormation stack for workload addons.
+type WorkloadAddonStack struct {
```
efekarakus: nit: What do you think of dropping "Addon" from the name so that it reads `addon.WorkloadStack` and `addon.EnvironmentStack`

internal/pkg/addon/addons.go
```diff
@@ -33,49 +39,115 @@ var (
...
+	stack
 	workloadName string
 }
 
+type EnvironmentAddonStack struct {
```
efekarakus: doc comment 🥺

internal/pkg/addon/addons.go
```diff
@@ -33,49 +39,115 @@ var (
...
+	filePath           func(fName string) string
+	validateParameters func(params yaml.Node, paramFilePath string) error
+}
+
 // Parse parses the 'addon/' directory for the given workload
 // and returns a Stack created by merging the CloudFormation templates
 // files found there. If no addons are found, Parse returns a nil
 // Stack and ErrAddonsNotFound.
-func Parse(workloadName string, ws workspaceReader) (*Stack, error) {
+func Parse(workloadName string, ws workspaceReader) (*WorkloadAddonStack, error) {
```
efekarakus: nit: should this be renamed to `ParseWorkload` or `ParseFromWorkload`

internal/pkg/addon/addons.go
```diff
@@ -33,49 +39,115 @@ var (
...
+		},
 		workloadName: workloadName,
 	}, nil
 }
 
+// ParseEnvAddon parses the 'addon/' directory for environments
+// and returns a Stack created by merging the CloudFormation templates
+// files found there. If no addons are found, Parse returns a nil
+// Stack and ErrAddonsNotFound.
+func ParseEnvAddon(ws workspaceReader) (*EnvironmentAddonStack, error) {
```
efekarakus: nit: can we rename to `ParseFromEnv` so that it reads `envAddon := addon.ParseFromEnv(ws)`

internal/pkg/addon/addons.go
```diff
@@ -105,30 +177,28 @@ func (s *Stack) encode(v any) (string, error) {
 	return str.String(), nil
 }
 
-// parseTemplate merges CloudFormation templates under the "addons/" directory of a workload
-// into a single CloudFormation template and returns it.
+// parseWorkloadTemplate merges CloudFormation templates under the "addons/" directory  into a single CloudFormation
```
efekarakus:
```suggestion
// parseTemplate merges CloudFormation templates under the "addons/" directory  into a single CloudFormation 
```

internal/pkg/addon/addons.go
```diff
@@ -137,55 +207,55 @@ func parseTemplate(fNames []string, workloadName string, ws workspaceReader) (*c
...
-		Name yaml.Node `yaml:"Name"`
-	}{}
+func validateReservedParameters(params yaml.Node, reservedKeys []string) error {
+	content := make(map[string]yaml.Node, len(reservedKeys))
+	for _, key := range reservedKeys {
+		content[key] = yaml.Node{}
+	}
 	if err := params.Decode(&content); err != nil {
-		return fmt.Errorf("decode content of parameters file %s under %s addons/", fname, workloadName)
+		return fmt.Errorf("decode content of the parameters file: %w", err)
```
efekarakus: is it really a file or just the `Parameters` section in the template?

internal/pkg/addon/addons.go
```diff
@@ -137,55 +207,55 @@ func parseTemplate(fNames []string, workloadName string, ws workspaceReader) (*c
...
 
-func validateReservedParameters(params yaml.Node, fname, workloadName string) error {
-	content := struct {
-		App  yaml.Node `yaml:"App"`
-		Env  yaml.Node `yaml:"Env"`
-		Name yaml.Node `yaml:"Name"`
-	}{}
+func validateReservedParameters(params yaml.Node, reservedKeys []string) error {
+	content := make(map[string]yaml.Node, len(reservedKeys))
+	for _, key := range reservedKeys {
```
efekarakus: so the `yaml` struct tags aren't needed if the field names match exactly?

internal/pkg/template/artifactpath/artifactpath.go
```diff
@@ -38,6 +39,12 @@ func AddonAsset(workloadName, hash string) string {
 	return path.Join(s3ArtifactDirName, s3ArtifactAddonsDirName, workloadName, s3ArtifactAddonAssetDirName, hash)
 }
 
+// EnvironmentAddonAsset returns the path to store an addon asset file for an environment addon.
+// Example: manual/addons/environments/assets/668e2b73ac.
```
efekarakus, line 43: I don't know why we put `assets` in there 😂 for workloads - seems redundant 🤔

efekarakus (commented):
Looks great! just a bunch of tiny renames.

Do not block on me for approval, I won't be online properly till Friday

internal/pkg/addon/addons.go
```diff
@@ -33,49 +39,115 @@ var (
...
 // and returns a Stack created by merging the CloudFormation templates
 // files found there. If no addons are found, Parse returns a nil
 // Stack and ErrAddonsNotFound.
-func Parse(workloadName string, ws workspaceReader) (*Stack, error) {
+func Parse(workloadName string, ws workspaceReader) (*WorkloadAddonStack, error) {
 	fNames, err := ws.ListFiles(ws.WorkloadAddonsPath(workloadName))
 	if err != nil {
-		return nil, &ErrAddonsNotFound{
-			WlName:    workloadName,
+		return nil, fmt.Errorf("list addons for workload %s: %w", workloadName, &ErrAddonsNotFound{
```
dannyrandall: nit: `%q`

internal/pkg/addon/addons.go
```diff
@@ -33,49 +39,115 @@ var (
...
 			ParentErr: err,
-		}
+		})
 	}
 
-	template, err := parseTemplate(fNames, workloadName, ws)
+	parser := parser{
```
dannyrandall, line 76: i wonder if parser should just have a `stack()` function that returns a `stack` 🤔 and also takes a `ListFiles` function? then these two functions (`addon.ForWorkload()` and `addon.ForEnvironment()`) would look like:
```
parser := parser{
  // omitted
  listFiles: ws.ListFiles,
}

stack, err := parser.stack()
return &Workload/EnvStack{
  stack: stack,
}
```

`listFiles` may or may not make sense to include, just due to the slightly different errors, but i think the `stack()` func would be nice

internal/pkg/addon/addons.go
```diff
@@ -33,49 +39,115 @@ var (
...
+	if err != nil {
+		return nil, fmt.Errorf("list addons for environments: %w", &ErrAddonsNotFound{
+			ParentErr: err,
+		})
+	}
+	parser := parser{
+		ws: ws,
+		filePath: func(fName string) string {
+			return ws.EnvAddonFilePath(fName)
+		},
```
dannyrandall: nit: would this work?
```suggestion
		filePath: ws.EnvAddonFilePath,
```

internal/pkg/addon/addons.go
```diff
@@ -105,30 +177,28 @@ func (s *Stack) encode(v any) (string, error) {
...
 
 	mergedTemplate := newCFNTemplate("merged")
 	for _, fname := range templateFiles {
-		path := ws.WorkloadAddonFilePath(workloadName, fname)
-		out, err := ws.ReadFile(path)
+		path := p.filePath(fname)
+		out, err := p.ws.ReadFile(path)
 		if err != nil {
-			return nil, fmt.Errorf("read addon %s under %s: %w", fname, workloadName, err)
+			return nil, fmt.Errorf("read addon %s under path %s: %w", fname, path, err)
```
dannyrandall: whats the difference between `fname` and `path`? is `fname` just the name of the file without the leading path?

internal/pkg/addon/addons.go
```diff
@@ -228,3 +298,15 @@ func contains(arr []string, el string) bool {
 	}
 	return false
 }
+
+func quoteSlice(elems []string) []string {
+	var quotedElems []string
+	if len(elems) == 0 {
+		return quotedElems
+	}
+	quotedElems = make([]string, len(elems))
```
dannyrandall: nit
```suggestion
	if len(elems) == 0 {
		return nil
	}
	quotedElems := make([]string, len(elems))
```

internal/pkg/addon/package.go
```diff
@@ -197,14 +197,28 @@ type PackageConfig struct {
 	WorkspacePath string
 	FS            afero.Fs
 
-	workloadName string
+	toPath func(hash string) string
```
dannyrandall: nit: `s3Path()`?

internal/pkg/addon/addons.go
Lou1415926: This is decoding the `Parameters` section of the `xx.params.yaml` file (not the template), and `Parameters` is pretty much the only section that's in `xx.params.yaml` file I think. I have updated the error message (will push later) - see if you like it better!

internal/pkg/addon/addons.go
Lou1415926: Not if it's decoding into a `map`. I think it's similar to how arbitrary keys under `variables` in manifest can be recognized as the key of a map. In fact, it might be clear how it works if I've written the code this way:
```go
        content := make(map[string]yaml.Node, len(reservedKeys))
	// for _, key := range reservedKeys {
	// 	content[key] = yaml.Node{}
	// }  Remove this for loop - it is not necessary.
	if err := params.Decode(&content); err != nil {
		return fmt.Errorf("decode \"Parameters\" section of the parameters file: %w", err)
	}
 	for _, key := range reservedKeys {
		if _, ok := content[key]; ok { // Instead of checking `field.IsZero`, we simply check if the key exists in the map.
                      	return fmt.Errorf("reserved parameters %s cannot be declared", english.WordSeries(quoteSlice(reservedKeys), "and"))
                 }
	}
```

internal/pkg/template/artifactpath/artifactpath.go
Lou1415926: I think it's storing the packaged content inside of an addon, not the addon itself. Hence "assets" to addons.

internal/pkg/addon/addons.go
Lou1415926: right - `fname` is agnostic of its parent (which workload it belongs to, whether `addons` is the name of an addon folder).

internal/pkg/addon/addons.go
efekarakus: 🤩 yea this way is a lot clearer! 🚀

internal/pkg/addon/addons.go
Lou1415926: updated!

internal/pkg/addon/addons.go
```diff
@@ -137,55 +200,50 @@ func parseTemplate(fNames []string, workloadName string, ws workspaceReader) (*c
 	return mergedTemplate, nil
 }
 
-// parseParameters returns the content of user-defined additional CloudFormation Parameters
+// parseWorkloadParameters returns the content of user-defined additional CloudFormation Parameters
```
efekarakus:
```suggestion
// parseParameters returns the content of user-defined additional CloudFormation Parameters
```

efekarakus (approved):

internal/pkg/addon/package.go
```diff
@@ -197,14 +197,28 @@ type PackageConfig struct {
...
 }
 
 // Package finds references to local files in Stack's template, uploads
 // the files to S3, and replaces the file path with the S3 location.
-func (s *Stack) Package(cfg PackageConfig) error {
-	cfg.workloadName = s.workloadName
+func (s *EnvironmentStack) Package(cfg PackageConfig) error {
+	cfg.s3Path = func(hash string) string {
+		return artifactpath.EnvironmentAddonAsset(hash)
+	}
```
dannyrandall:
```suggestion
	cfg.s3Path = artifactpath.EnvironmentAddonAsset
```

dannyrandall (approved):

PR #2537, opened 2021-06-28
fix(stream): wrap close done ch with sync.Once

efekarakus:
Today, `stream.Stream` is structured as:
```go
case <-streamer.Done():
  // call Close()
case <-time.After(fetchDelay):
  // call Fetch()
```
The problem is that it's possible that the `Done()` event is triggered at
the same time as `time.After(fetchDelay)` and in that scenario the
behavior is undeterministic.
It's possible that we call `Fetch` a second time. For the ECS deployment
streamer, this means that we would previous close the `s.done` channel
twice. Wrapping the call with a `sync.Once` removes this concern.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4231, opened 2022-11-29
fix(stream): should not try to fetch from a closed streamer

iamhopaul123:
Currently when using progress tracker (especially for tracking StackSet events), it is possible that the program panics because we try to close a closed channel.

Behind the scene, it is because we use
```golang
	waitMs :=
		rand(
			min(
				streamerMaxFetchIntervalDurationMs,
				streamerFetchIntervalDurationMs*(1<<retries),
			)
		)
```
as `Fetch` interval, which is possible to set the wait time to be very close to 0s if `retries` is 0. And then we do the last Fetch (when the stack/stackset/ecs deployment is done), we always set `retries` to 0.

In the caller function `Stream`, we have a select logic as below

```golang
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-streamer.Done():
			// No more events to Fetch, flush and exit successfully.
			streamer.Notify()
			return nil
		case <-time.After(fetchDelay):
			next, err = streamer.Fetch()
			if err != nil {
				return err
			}
			streamer.Notify()
		}
```
When `fetchDelay` is 0, it could happen that both `streamer.Done()` and `time.After(fetchDelay)` are `true`. And in this scenario, golang will [select one of them randomly](https://go.dev/tour/concurrency/5) (doom when the one with `time.After(fetchDelay)` condition is select).

This fix will set a `streamerMinFetchIntervalDurationMs` to make sure we don't fetch immediately after the previous fetch. It could also help reduce the throttling error we might receive without hurting the current performance.


internal/pkg/stream/stream.go
```diff
@@ -64,18 +65,19 @@ func Stream(ctx context.Context, streamer Streamer) error {
...
-	)
+	// 	)-streamerMinFetchIntervalDurationMs
+	// ) + streamerMinFetchIntervalDurationMs
+	waitMs :=
+		rand(
+			min(
+				streamerMaxFetchIntervalDurationMs,
+				streamerFetchIntervalDurationMs*(1<<retries),
+			)-streamerMinFetchIntervalDurationMs,
+		) + streamerMinFetchIntervalDurationMs
```
dannyrandall, line 71: while i think, in practice, this will always work - what we are doing here is effectively hoping that we'll give the scheduler enough time to give the `select` block a chance to run again and select the `Done()` case before the time returned has passed, which isn't necessarily guaranteed.

i wonder if it makes more sense to handle the "closed" case from within the `Fetch()` functions. for example, we could change the logic in `Fetch()` to something like:
```
Fetch() {
  if we closed the chan {
    return time.Time{}
  }

  return nextFetchDate()
}
```

And then in the `select` block, we could change the fetch delay case to:
```
case <-time.After(fetchDelay):
	next, err = streamer.Fetch()
	if err != nil {
		return err
	}
	streamer.Notify()
    if next.IsZero() {
		return nil
    }
```
Something like that? We could also return a "Stop fetching" error or something like that, or a third return value. Something other than just a time that should be long enough.

dannyrandall:
Great job finding the root cause of this!🥳

internal/pkg/stream/stream.go
iamhopaul123: That sounds good to me. I'll still leave the min fetch time so that we don't fetch too often, because right now it is almost guaranteed that we'll receive the throttling error.

internal/pkg/stream/cfnstack.go
```diff
@@ -130,9 +128,9 @@ func (s *StackStreamer) Fetch() (next time.Time, err error) {
 			// Check for throttles and wait to try again using the StackStreamer's interval.
 			if request.IsErrorThrottle(err) {
 				s.retries += 1
-				return nextFetchDate(s.clock, s.rand, s.retries), nil
+				return nextFetchDate(s.clock, s.rand, s.retries), done, nil
```
dannyrandall: nit: maybe just explicitly return false, just helps me know that `done` hasn't been modified somewhere up above and you're returning its value (same L133)

internal/pkg/stream/ecs.go
```diff
@@ -143,13 +140,7 @@ func (s *ECSDeploymentStreamer) Fetch() (next time.Time, err error) {
...
 		deployments = append(deployments, rollingDeploy)
 		if isDeploymentDone(rollingDeploy, s.deploymentCreationTime) {
-			// The deployment is done, notify that there is no need for another Fetch call beyond this point.
-			// In stream.Stream, it's possible that both the <-Done() event is available as well as another Fetch()
-			// call. In order to guarantee that we don't try to close the same stream multiple times, we wrap it with a
-			// sync.Once.
-			s.once.Do(func() {
-				close(s.done)
-			})
+			done = true
```
dannyrandall, line 143:
```suggestion
			done = isDeploymentDone(rollingDeploy, s.deploymentCreationTime)
```

internal/pkg/stream/stream.go
```diff
@@ -38,6 +36,7 @@ func Stream(ctx context.Context, streamer Streamer) error {
 	defer streamer.Close()
 
 	var next time.Time
+	var done bool
 	var err error
```
dannyrandall: nit: remove these declarations and move the
```
if now := time.Now(); next.After(now) {
	fetchDelay = next.Sub(now)
}
```
into the `time.After` case

dannyrandall: oh also nit: i think that block can just be `fetchDelay = time.Until(next)`

internal/pkg/stream/stream.go
```diff
@@ -48,34 +47,34 @@ func Stream(ctx context.Context, streamer Streamer) error {
...
-		),
-	)
+	// 	)-streamerMinFetchIntervalDurationMs
+	// ) + streamerMinFetchIntervalDurationMs
+	waitMs :=
+		rand(
+			min(
+				streamerMaxFetchIntervalDurationMs,
+				streamerFetchIntervalDurationMs*(1<<retries),
+			)-streamerMinFetchIntervalDurationMs,
```
dannyrandall, line 70: if min generated a number less than `streamerMinFetchIntervalDurationMs`, `rand` would get passed a negative number and panic - unless i'm reading that wrong?

iamhopaul123: min can't generate a number less than `streamerMinFetchIntervalDurationMs`. so it will be

`[0, min(streamerMaxFetchIntervalDurationMs, streamerFetchIntervalDurationMs*(1<<retries))-streamerMinFetchIntervalDurationMs)` for the `rand` and then
`[streamerMinFetchIntervalDurationMs, min(streamerMaxFetchIntervalDurationMs, streamerFetchIntervalDurationMs*(1<<retries)))` after the addition.

dannyrandall: ahhh yeah i gotcha, i was thinking `rand` logic on the inside instead of min.👍

dannyrandall (approved):
lgtm! thanks for fixing this bug!

efekarakus:
huh I thought this issue was fixed with: https://github.com/aws/copilot-cli/pull/2537 but I think it's that I forgot to wrap the close call with `sync.Once` for the stackset streamer.

Could the same solution be applied here instead of changing the API?

iamhopaul123:
> Could the same solution be applied here instead of changing the API?

thought about that but it doesn't sound to be the right approach to me because when the stack is done we shouldn't fetch any more info again. `once` just makes sure we don't close the same channel again. Besides for the same streamer `Fetch` is always called sequentially and we probably don't need to use `done` as a channel. This PR should simplify the logic and make it more easy to maintain too.

internal/pkg/stream/cfnstack.go
```diff
@@ -113,8 +111,8 @@ func (s *StackStreamer) Subscribe() <-chan StackEvent {
 
 // Fetch retrieves and stores any new CloudFormation stack events since the ChangeSetCreationTime in chronological order.
 // If an error occurs from describe stack events, returns a wrapped error.
-// Otherwise, returns the time the next Fetch should be attempted.
-func (s *StackStreamer) Fetch() (next time.Time, err error) {
+// Otherwise, returns the time the next Fetch should be attempted, and if the stack is done.
```
efekarakus:
```suggestion
// Otherwise, returns the time the next Fetch should be attempted and whether or not there are more events to fetch.
```

internal/pkg/stream/cfnstackset.go
```diff
@@ -118,24 +114,22 @@ func (s *StackSetStreamer) Subscribe() <-chan StackSetOpEvent {
 }
 
 // Fetch retrieves the latest stack set operation and buffers it.
```
efekarakus: Can we update the comment in the other `Fetch` methods too to explain the purpose of the boolean return variable 🥺

internal/pkg/stream/stream.go
```diff
@@ -13,69 +13,61 @@ import (
...
```
efekarakus, line 70: I am having difficulty parsing this 😅 even after reading the comments above from Danny.
After plugging in the numbers it makes sense! I think this might be equivalent too:
```
jitter := rand(
	min(
		streamerMaxFetchIntervalDurationMs - streamerMinFetchIntervalDurationMs,
		streamerFetchIntervalDurationMs*(1<<retries),
	)
)
waitMs := streamerMinFetchIntervalDurationMs + jitter
```

efekarakus (approved):
Looks good boss! ty for fixing this.

Just a doc comment suggestion and a readability one

internal/pkg/stream/stream.go
iamhopaul123: hmm it's not equivalent when `retries` is a small number like 0. In this case the current implementation has 
```
[streamerMinFetchIntervalDurationMs, streamerFetchIntervalDurationMs*(1<<retries))
```
where as yours is
```
[streamerMinFetchIntervalDurationMs, streamerFetchIntervalDurationMs*(1<<retries)+streamerMinFetchIntervalDurationMs)
```

I'll add a URL with more info about this https://www.educative.io/answers/how-to-generate-random-numbers-in-a-given-range-in-go

PR #4234, opened 2022-11-30
test(e2e): apprunner private service test

dannyrandall:
Also removes db addon test from apprunner suite to clean up deps and keep running time reasonable.


iamhopaul123 (commented):
I like how it helps to reduce the running time for `apprunner` suite which takes most time right now, but for those removed blocks should we add them to the other suite or they are already there in some other suites?

dannyrandall:
we test an RDS and S3 addon in the `addons` e2e test!

iamhopaul123 (approved):

e2e/apprunner/front-end/main.go
```diff
@@ -5,129 +5,68 @@ package main
...
+	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
-		log.Printf("🚨 sql open: err=%s\n", err)
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+		handleErr("do request: %s", err)
 		return
 	}
-	defer db.Close()
+	defer resp.Body.Close()
+	log.Printf("Did request")
```
Lou1415926: nit (and a few other places where `f` isn't needed):
```suggestion
	log.Print("Did request")
```

Lou1415926 (approved):
Great test!

e2e/apprunner/apprunner_test.go
```diff
@@ -86,39 +86,6 @@ var _ = Describe("App Runner", func() {
...
-			deployErr        error
-		)
-
-		BeforeAll(func() {
-			_, s3StorageInitErr = cli.StorageInit(&client.StorageInitRequest{
-				StorageName:  "s3storage",
-				StorageType:  "S3",
-				WorkloadName: feSvcName,
-			})
-			_, auroraInitErr = cli.StorageInit(&client.StorageInitRequest{
```
Lou1415926, line 102: The `addon` test suite only test LBWS. I think s3 addon is probably fine, but there are something different with RDWS when it comes to RDS (iirc you need to have `network.placement: 'private'` to be able to send egress from RDWS to the DB). I think this makes RDWS+RDS special - which was why we added RDS to `apprunner` test suite in the first place. Probably also why we kept `copilot/front-end/manifest.yml` as we needed to keep that `network.placement: 'private'` bit.

I think if we do not want to keep it here, we can move the RDS+RDWS test to `addon` package (in a separate PR). Either works for me!

dannyrandall: yeah - i figured that we were effectively testing the same thing (RDWS->VPC) because that's required for a RDWS service to reach a _private_ RDWS service. Traffic flows like: `front end -> VPC Connector -> VPC -> App Runner VPC Endpoint -> back-end`. So at least from a reachability standpoint, this change tests the same thing (it does stop testing RDWS+addons, but I think that's an ok tradeoff?)

PR #4237, opened 2022-11-30
ci: include local-integ in mainline commit coverage

dannyrandall:


.github/workflows/commit.yml
```diff
@@ -7,6 +7,7 @@ on:
 
 jobs:
   test: # same as ci/test
+    if: github.repository == "aws/copilot-cli" # don't run on forks
```
efekarakus: Why are we making this change?

dannyrandall: i noticed it's running on all forks and i'm not sure there's a reason for them to all have their own coverage reports? i guess it doesn't really matter if they do 🤔

efekarakus (approved):

Lou1415926 (approved):

PR #4240, opened 2022-12-01
ci: compute size diff between PR and last release

Lou1415926:
instead of mainline vs. last release


dannyrandall:
This workflow is making me wonder if we can use a different event: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ 🤔

efekarakus (approved):

huanjani (approved):

Lou1415926:
I updated this PR so that we now have two workflows:
1. `Compute bin sizes` is triggered by `pull_request` event, which is a read-only trigger. This workflow computes the sizes of old/new binaries, and upload the results as artifacts.
2. `Comment on PR` is triggered by the completion of `Compute bin sizes`. It reads the results that were uploaded as artifacts, and comment on the PR.

I already have two approvals for the previous version, but this change is too large to assume that the approvals are still in effect 👀  Can I please have two more approvals before I can remove the label?

.github/workflows/ci_size_computer.yml
```diff
@@ -0,0 +1,97 @@
...
+          echo "MAC_ARM_KIB=$(du -k ./bin/local/copilot-darwin-arm64 | awk '{ print $1}')" >> $GITHUB_OUTPUT
+          echo "LINUX_AMD_KIB=$(du -k ./bin/local/copilot-linux-amd64 | awk '{ print $1}')" >> $GITHUB_OUTPUT
+          echo "LINUX_ARM_KIB=$(du -k ./bin/local/copilot-linux-arm64 | awk '{ print $1}')" >> $GITHUB_OUTPUT
+          echo "WINDOWS_AMD_KIB=$(du -k ./bin/local/copilot.exe | awk '{ print $1}')" >> $GITHUB_OUTPUT
+
+      - name: Save sizes and PR number
+        run: |
+          mkdir -p artifacts/old artifacts/new
+          sizes_json="{
+            \"macOS\": {
```
efekarakus, line 64: lol, this must have been fun to write

efekarakus (approved):
Really cool! Thanks Wanxian!

.github/workflows/ci_size_computer.yml
```diff
@@ -0,0 +1,97 @@
...
+          make release
+          echo "MAC_AMD_KIB=$(du -k ./bin/local/copilot-darwin-amd64 | awk '{ print $1}')" >> $GITHUB_OUTPUT
+          echo "MAC_ARM_KIB=$(du -k ./bin/local/copilot-darwin-arm64 | awk '{ print $1}')" >> $GITHUB_OUTPUT
+          echo "LINUX_AMD_KIB=$(du -k ./bin/local/copilot-linux-amd64 | awk '{ print $1}')" >> $GITHUB_OUTPUT
+          echo "LINUX_ARM_KIB=$(du -k ./bin/local/copilot-linux-arm64 | awk '{ print $1}')" >> $GITHUB_OUTPUT
+          echo "WINDOWS_AMD_KIB=$(du -k ./bin/local/copilot.exe | awk '{ print $1}')" >> $GITHUB_OUTPUT
+
+      - name: Save sizes and PR number
+        run: |
+          mkdir -p artifacts/old artifacts/new
```
dannyrandall: why do we need this line, and not just `mkdir artifacts`?

dannyrandall (approved):
just a question, otherwise looks great!🚀

.github/workflows/ci_size_computer.yml
Lou1415926: good catch! I changed the implementation and forgot to remove this line

PR #4250, opened 2022-12-05
fix: Better error message for alias validation against imported certificate

KollaAdithya:
Currently when customer import a certificate and specify alias in their service manifest, Copilot validates the aliases against that imported certificate. For example, if the certificate allows `images.example.com`, then an alias of `test.images.example.com` is not valid.

Copilot error out saying something like `"test.images.example.com" is not a valid domain against <certificate ARN>`.  But we don’t really say what’s in the certificate.  

With this PR we log out the  details  of the imported `certificate ARN` and all the allowed domains in the Certificate.


internal/pkg/aws/acm/acm.go
```diff
@@ -66,23 +71,31 @@ func (a *ACM) ValidateCertAliases(aliases []string, certs []string) error {
...
+func (a *ACM) findValidAliasesAndAllowedDomainsAgainstCert(ctx context.Context, aliases []string, cert string) ([]string, map[string][]string, error) {
 	resp, err := a.client.DescribeCertificateWithContext(ctx, &acm.DescribeCertificateInput{
 		CertificateArn: aws.String(cert),
 	})
 	if err != nil {
-		return nil, fmt.Errorf("describe certificate %s: %w", cert, err)
+		return nil, nil, fmt.Errorf("describe certificate %s: %w", cert, err)
 	}
 	domainSet := make(map[string]bool)
+	domainsperCert := make(map[string][]string)
```
Lou1415926: I might be missing something - it seems like the only key would be `cert`. If this is indeed the case, why do we need a `map`?

internal/pkg/aws/acm/acm.go
```diff
@@ -66,23 +71,31 @@ func (a *ACM) ValidateCertAliases(aliases []string, certs []string) error {
 	}
 	for _, alias := range aliases {
 		if !validAliases[alias] {
+			for certARN, domains := range AllowedDomainsCert {
+				log.Infoln(fmt.Sprintf("Your imported Certificate '%s' protects the following domains:", certARN))
```
Lou1415926: I think we should avoid `log`ging in arbitrary packages. What do you think of making the returned error a custom type and putting these in its `RecommendedActions`?

internal/pkg/aws/acm/acm.go
```diff
@@ -42,14 +43,15 @@ func New(s *session.Session) *ACM {
 // ValidateCertAliases validates if aliases are all valid against the provided ACM certificates.
 func (a *ACM) ValidateCertAliases(aliases []string, certs []string) error {
 	validAliases := make(map[string]bool)
+	AllowedDomainsCert := make(map[string][]string)
```
Lou1415926: perhaps rename `AllowedDomainsCert` to something like `sansOfCert`
so that when it reads:
```go
sans := sansOfCert[cert]
```

internal/pkg/aws/acm/acm.go
```diff
@@ -42,14 +43,15 @@ func New(s *session.Session) *ACM {
...
 	ctx, cancelWait := context.WithTimeout(context.Background(), waitForFindValidAliasesTimeout)
 	defer cancelWait()
 	g, ctx := errgroup.WithContext(ctx)
 	var mux sync.Mutex
 	for i := range certs {
 		cert := certs[i]
 		g.Go(func() error {
-			validCertAliases, err := a.findValidAliasesAgainstCert(ctx, aliases, cert)
+			validCertAliases, domainsperCert, err := a.findValidAliasesAndAllowedDomainsAgainstCert(ctx, aliases, cert)
```
Lou1415926:
```suggestion
			validCertAliases, sans, err := a.findValidAliasesAndAllowedDomainsAgainstCert(ctx, aliases, cert)
```
The `perCert` part is probably not needed if all the SANS that get returned are for the same `cert`.

Lou1415926 (commented):
Thanks Adi!

internal/pkg/aws/acm/acm.go
```diff
@@ -66,23 +71,31 @@ func (a *ACM) ValidateCertAliases(aliases []string, certs []string) error {
...
 	resp, err := a.client.DescribeCertificateWithContext(ctx, &acm.DescribeCertificateInput{
 		CertificateArn: aws.String(cert),
 	})
 	if err != nil {
-		return nil, fmt.Errorf("describe certificate %s: %w", cert, err)
+		return nil, nil, fmt.Errorf("describe certificate %s: %w", cert, err)
 	}
 	domainSet := make(map[string]bool)
+	domainsperCert := make(map[string][]string)
 	domainSet[aws.StringValue(resp.Certificate.DomainName)] = true
```
paragbhingre: Are we missing this one entry inside `domainsperCert` ?

paragbhingre (commented):
This is so cool and helpful 🤩

internal/pkg/aws/acm/acm.go
KollaAdithya: Just changed the map to [ ]string

internal/pkg/aws/acm/acm.go
```diff
@@ -66,23 +68,28 @@ func (a *ACM) ValidateCertAliases(aliases []string, certs []string) error {
...
+func (a *ACM) findValidAliasesAndAllowedDomainsAgainstCert(ctx context.Context, aliases []string, cert string) ([]string, []string, error) {
 	resp, err := a.client.DescribeCertificateWithContext(ctx, &acm.DescribeCertificateInput{
 		CertificateArn: aws.String(cert),
 	})
 	if err != nil {
-		return nil, fmt.Errorf("describe certificate %s: %w", cert, err)
+		return nil, nil, fmt.Errorf("describe certificate %s: %w", cert, err)
 	}
 	domainSet := make(map[string]bool)
+	var domainsperCert []string
```
huanjani: nit for readability:
```suggestion
	var domainsPerCert []string
```

internal/pkg/aws/acm/acm.go
```diff
@@ -92,5 +99,23 @@ func (a *ACM) findValidAliasesAgainstCert(ctx context.Context, aliases []string,
...
+}
+
+func (e *errInValidAliasAgainstCert) Error() string {
+	return fmt.Sprintf("%s is not a valid domain against %s", e.alias, strings.Join(e.certs, ","))
+}
+
+func (e *errInValidAliasAgainstCert) RecommendActions() string {
+	var logMsg string
+	for cert, sans := range e.sansOfCert {
+		logMsg += fmt.Sprintf("Your imported Certificate '%s' protects the following domains:\n%s\n", cert, strings.Join(sans, "\n"))
```
huanjani: tiny nit:
```suggestion
		logMsg += fmt.Sprintf("Your imported certificate '%s' protects the following domains:\n%s\n", cert, strings.Join(sans, "\n"))
```

internal/pkg/aws/acm/acm.go
```diff
@@ -66,23 +68,28 @@ func (a *ACM) ValidateCertAliases(aliases []string, certs []string) error {
...
+				certs:      certs,
+				alias:      alias,
+				sansOfCert: sansOfCert}
 		}
 	}
 	return nil
 }
 
-func (a *ACM) findValidAliasesAgainstCert(ctx context.Context, aliases []string, cert string) ([]string, error) {
+func (a *ACM) findValidAliasesAndAllowedDomainsAgainstCert(ctx context.Context, aliases []string, cert string) ([]string, []string, error) {
```
Lou1415926: Let's try to keep each method single purposed! What do you think of the following proposal:
```go
func (a *ACM) validDomainsOfCert(ctx context.Context, cert string) ([]string, err) {
   // Finds the domain and the SANs of the cert
}

func filterValidAliases(domains[]string, aliases[]string) )[]string {
  // Filter `aliases` based on `domains`
}

   // ...
   for i := range certs {
		cert := certs[i]
		g.Go(func() error {
			domains, err := a.validDomainsOfCert(ctx, cert)
			if err != nil {
				return err
			}
			validCertAliases := filterValidAliases()

                         mux.Lock()
			defer mux.Unlock()
			
			domainsOfCert[cert] = domains
			for _, alias := range validCertAliases {
				validAliases[alias] = true
			}
			return nil
                         
                         
	}
   // ...
}

```

KollaAdithya: Done

internal/pkg/aws/acm/acm.go
```diff
@@ -66,23 +69,35 @@ func (a *ACM) ValidateCertAliases(aliases []string, certs []string) error {
...
 	if err != nil {
 		return nil, fmt.Errorf("describe certificate %s: %w", cert, err)
 	}
-	domainSet := make(map[string]bool)
-	domainSet[aws.StringValue(resp.Certificate.DomainName)] = true
+	var domainsOfCert []string
+	domainsOfCert = append(domainsOfCert, aws.StringValue(resp.Certificate.DomainName))
 	for _, san := range resp.Certificate.SubjectAlternativeNames {
-		domainSet[aws.StringValue(san)] = true
+		domainsOfCert = append(domainsOfCert, aws.StringValue(san))
```
Lou1415926: We can consider using [`aws.StringValueSlice`](https://docs.aws.amazon.com/sdk-for-go/api/aws/#StringValueSlice)

internal/pkg/aws/acm/acm.go
```diff
@@ -66,23 +69,35 @@ func (a *ACM) ValidateCertAliases(aliases []string, certs []string) error {
...
+	}
+	return domainsOfCert, err
+}
+
+func filterValidAliases(domains []string, aliases []string) []string {
+	domainSet := make(map[string]bool)
```
Lou1415926: Since we know the length in advance, there is a chance for a bit of optimization:
```suggestion
	domainSet := make(map[string]bool, len(domains))
```
When we allocate the map with a initial size, we save the computation cost of resizing the map for the first `len(domains)` items.

internal/pkg/aws/acm/acm.go
```diff
@@ -92,5 +107,23 @@ func (a *ACM) findValidAliasesAgainstCert(ctx context.Context, aliases []string,
...
+
+func (e *errInValidAliasAgainstCert) Error() string {
+	return fmt.Sprintf("%s is not a valid domain against %s", e.alias, strings.Join(e.certs, ","))
+}
+
+func (e *errInValidAliasAgainstCert) RecommendActions() string {
+	var logMsg string
+	for cert, sans := range e.domainsOfCert {
+		logMsg += fmt.Sprintf("Your imported certificate '%s' protects the following domains:\n%s\n", cert, strings.Join(sans, "\n"))
+	}
```
Lou1415926: nit: When there are multiple certs, this reads:
```
Your imported cert 'arn:aws:acm:us-west-2:111111111111:certificate/5033ba2b-09b2-4bc0-8dbb-51adf3e12345' protects the following domains:
a.com, b.com
Your imported cert 'arn:aws:acm:us-west-2:111111111111:certificate/5033ba2b-09b2-4bc0-8dbb-51adf3e54321' protects the following domains:
c.com, d.com
```
The repetition *might* sound a little robotic?
What do you think of something like:
```
Please use aliases that are protected by the certificates your imported:
'5033ba2b-09b2-4bc0-8dbb-51adf3e12345': a.com, b.com
'5033ba2b-09b2-4bc0-8dbb-51adf3e54321': c.com, d.com
```

We can take advantage of [`english.Plural`](https://pkg.go.dev/glitch.sh/go/humanize/english#Plural) to use "certificate" (singular) or "certificates" (plural) depending on the number of certs imported.
https://pkg.go.dev/glitch.sh/go/humanize/english#WordSeries

Lou1415926 (approved):

internal/pkg/aws/acm/acm.go
```diff
@@ -92,5 +106,24 @@ func (a *ACM) findValidAliasesAgainstCert(ctx context.Context, aliases []string,
...
+
+func (e *errInValidAliasAgainstCert) Error() string {
+	return fmt.Sprintf("%s is not a valid domain against %s", e.alias, strings.Join(e.certs, ","))
+}
+
+func (e *errInValidAliasAgainstCert) RecommendActions() string {
+	var logMsg string
+	logMsg = fmt.Sprintf("Please use aliases that are protected by %s your imported:\n", english.Plural(len(e.certs), "certificate", ""))
+	for cert, sans := range e.domainsOfCert {
+		logMsg += fmt.Sprintf("'%s': %s\n", cert, english.WordSeries(sans[1:], ","))
```
Lou1415926:
```suggestion
		logMsg += fmt.Sprintf("%q: %s\n", cert, english.WordSeries(sans[1:], ","))
```

Lou1415926: Why `sans[1:]` instead of just `sans`?

Lou1415926 (approved):

internal/pkg/aws/acm/acm.go
KollaAdithya: `aws acm describe-certificate --certificate-arn` fetches the output something like below where Domain Name is repetitive in SubjectAlternativeNames. so that is why I used `sans[1:]`.

        "CertificateArn": "arn:aws:acm:us-west-2:535307839156:certificate/c00251c7-5642-4a1e-8c19-7cf0daf92c2e",
        "DomainName": "gingo.adi.peanutsninja.com",
        "SubjectAlternativeNames": [
            "gingo.adi.peanutsninja.com",
            "gingo.bong.adi.peanutsninja.com"
        ],

Lou1415926: I see. From the [doc](https://docs.aws.amazon.com/acm/latest/APIReference/API_CertificateDetail.html#ACM-Type-CertificateDetail-SubjectAlternativeNames): 
> The subject alternative names include the canonical domain name (CN) of the certificate and additional domain names that can be used to connect to the website.

internal/pkg/aws/acm/acm.go
```diff
@@ -66,23 +70,33 @@ func (a *ACM) ValidateCertAliases(aliases []string, certs []string) error {
...
 	})
 	if err != nil {
 		return nil, fmt.Errorf("describe certificate %s: %w", cert, err)
 	}
-	domainSet := make(map[string]bool)
-	domainSet[aws.StringValue(resp.Certificate.DomainName)] = true
-	for _, san := range resp.Certificate.SubjectAlternativeNames {
-		domainSet[aws.StringValue(san)] = true
+	var domainsOfCert []*string
+	domainsOfCert = append(domainsOfCert, resp.Certificate.DomainName)
```
Lou1415926: Do you think we need this line then? Based on the discussion here https://github.com/aws/copilot-cli/pull/4250#discussion_r1053572593

KollaAdithya: removed that line as appending of canonical Domain name is not needed

internal/pkg/aws/acm/acm.go
```diff
@@ -66,23 +70,32 @@ func (a *ACM) ValidateCertAliases(aliases []string, certs []string) error {
...
 	})
 	if err != nil {
 		return nil, fmt.Errorf("describe certificate %s: %w", cert, err)
 	}
-	domainSet := make(map[string]bool)
-	domainSet[aws.StringValue(resp.Certificate.DomainName)] = true
-	for _, san := range resp.Certificate.SubjectAlternativeNames {
-		domainSet[aws.StringValue(san)] = true
+	var domainsOfCert []*string
+	domainsOfCert = append(domainsOfCert, resp.Certificate.SubjectAlternativeNames...)
```
paragbhingre, line 91: Are we sure that we are also including the `DomainName` as well that we are getting in the resp? Because when I looked at the tests I got little confused that we have removed `DomainName` entry and added it's value inside `SubjectAlternativeNames`. It should be good as long as `SubjectAlternativeNames` has DomainName already included in it. Else we will lose one entry every time.

KollaAdithya: From this [doc](https://docs.aws.amazon.com/acm/latest/APIReference/API_CertificateDetail.html#ACM-Type-CertificateDetail-SubjectAlternativeNames) it is mentioned that:
The subject alternative names include the canonical domain name (CN) of the certificate and additional domain names that can be used to connect to the website. 
So the `DomainName` is already included in the `SubjectAlternativeNames`. That is the reason why I removed the `DomainName` and  added to the `SubjectAlternativeNames` in the test cases.

paragbhingre (approved):
:shipit:

PR #4255, opened 2022-12-06
chore(cli): better error msg if workload deploy/package to an undeployed env

iamhopaul123:


dannyrandall (approved):
awesome!!

internal/pkg/cli/svc_deploy.go
```diff
@@ -395,6 +396,13 @@ func workloadManifest(in *workloadManifestInput) (manifest.DynamicWorkload, erro
 }
 
 func validateWorkloadManifestCompatibilityWithEnv(ws wsEnvironmentsLister, env versionCompatibilityChecker, mft manifest.DynamicWorkload, envName string) error {
+	currVersion, err := env.Version()
+	if err != nil {
+		return fmt.Errorf("get environment %q version: %w", envName, err)
+	}
```
Lou1415926, line 402: Do we have any super old environment template that doesn't have version?

Lou1415926 (approved):

internal/pkg/cli/svc_deploy.go
iamhopaul123: Good call. I think it is ok because we'll return `deploy.LegacyEnvTemplateVersion` which is `v0.0.0`

Issue #4208, opened 2022-11-22
Extending Copilot with the CDK


efekarakus:
Hi folks! This issue is a design proposal for overriding Copilot abstracted resources using the [AWS Cloud Development Kit (CDK)](https://docs.aws.amazon.com/cdk/v2/guide/home.html). The goal is to provide a "break the glass" mechanism to access and configure functionality that is not surfaced by [Copilot manifests](https://aws.github.io/copilot-cli/docs/manifest/overview/) by leveraging the expressive power of a programming language.

### Problem statement
Clients can describe their application configuration with declarative [`manifest.yml` files](https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/). Manifests enable users to define their application in terms of [“architecture as code”](https://aws.amazon.com/blogs/containers/containers-and-infrastructure-as-code-like-peanut-butter-and-jelly/) that gets transformed into AWS CloudFormation (CFN) templates by Copilot. Today, customers have the ability to define additional AWS resources with [“addons”](https://aws.github.io/copilot-cli/docs/developing/addons/modeling/) CFN templates. However, the resources encapsulated by manifests remain unmodifiable by users. We're looking into providing extension solutions for customers to self-unblock themselves and access the internal layers of Copilot.

This document proposes a user experience for extending manifests with the CDK.
### Proposal
1. **Copilot will introduce a new command `copilot [noun] override`** to create a CDK application that overrides Copilot resources, for example:
   ```console
   $ copilot svc override
   > Which service's resources would you like to override?
   * frontend
   * api

   > Which infrastructure as code tool would you like to use to override "frontend"?
   * AWS Cloud Development Kit (CDK)
   * CloudFormation YAML patches

   > Which programming language would you like to use with the CDK?
   * Typescript(ts)

   > Which resources in "frontend" would you like to to override?
   [ ] DiscoveryService          (AWS::ServiceDiscovery::Service)
   [x] Service                   (AWS::ECS::Service)
   [x] PublicNetworkLoadBalancer  (AWS::ElasticLoadBalancingV2::LoadBalancer)
   [ ] TaskDefinition            (AWS::ECS::TaskDefinition)
   ✅ Created a new CDK app under `copilot/frontend/overrides/` to override the resources:
   .
   ├── .gitignore
   ├── .build/
   ├── bin/
   │   └── override.ts
   ├── cdk.json
   ├── package.json
   ├── stack.ts
   ├── tsconfig.json
   └── README.md

[...]

PR #4256, opened 2022-12-08
chore(override): add skeleton package for CDK overrides

efekarakus:
Implements `Install` and `Override` for extending Copilot with the CDK.


internal/pkg/override/cdk.go
```diff
@@ -0,0 +1,85 @@
...
+type CDK struct {
+	rootAbsPath string // Absolute path to the overrides/ directory.
+
+	out  io.Writer  // Writer for any os/exec calls.
+	fs   afero.Fs   // OS file system.
+	exec Executable // For testing os/exec calls.
+}
+
+// NewCDK instantiates a new CDK Overrider with the file system pointing ot the overrides/ dir.
+func NewCDK(root string, stdout io.Writer, fs afero.Fs, exec Executable) *CDK {
```
efekarakus, line 33: I will get rid of this constructor in an upcoming PR same with the `Executable` interface, I just created them for the sake of writing unit tests without depending on private struct fields

internal/pkg/override/cdk.go
```diff
@@ -0,0 +1,85 @@
...
+	if err := cdk.fs.MkdirAll(buildPath, fs.ModeDir); err != nil {
+		return nil, fmt.Errorf("create %s directory to store the CloudFormation template body: %w", buildPath, err)
+	}
+	inputPath := filepath.Join(buildPath, "in.yml")
+	if err := afero.WriteFile(cdk.fs, inputPath, body, 0644); err != nil {
+		return nil, fmt.Errorf("write CloudFormation template body content at %s: %w", inputPath, err)
+	}
+
+	// We assume that a node_modules/ dir is present with the CDK downloaded after running "npm install".
+	// This way clients don't need to install the CDK toolkit separately.
```
paragbhingre, line 81: ❤️

internal/pkg/override/cdk.go
```diff
@@ -0,0 +1,85 @@
...
+		rootAbsPath: root,
+		out:         stdout,
+		fs:          fs,
+		exec:        exec,
+	}
+}
+
+// Install runs "npm install" insides the overrides/ directory to pull the latest dependencies.
+func (cdk *CDK) Install() error {
+	if _, err := cdk.exec.LookPath("npm"); err != nil {
```
paragbhingre, line 57: Does this look for "npm" inside PATH env variable?

internal/pkg/override/cdk.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, fmt.Errorf("create %s directory to store the CloudFormation template body: %w", buildPath, err)
+	}
+	inputPath := filepath.Join(buildPath, "in.yml")
+	if err := afero.WriteFile(cdk.fs, inputPath, body, 0644); err != nil {
+		return nil, fmt.Errorf("write CloudFormation template body content at %s: %w", inputPath, err)
+	}
+
+	// We assume that a node_modules/ dir is present with the CDK downloaded after running "npm install".
+	// This way clients don't need to install the CDK toolkit separately.
+	cmd := cdk.exec.Command(filepath.Join("node_modules", "aws-cdk", "bin", "cdk"), "synth", "--no-version-reporting")
```
paragbhingre, line 82: override.ts file creation is not a part of this PR right? CDK synth will be applied on the override.ts file which will be created using cloudformation_include module right?

internal/pkg/override/cdk.go
efekarakus: yup, the implementation will be https://pkg.go.dev/os/exec#LookPath, we don't call the function directly to just facilitate testing

internal/pkg/override/cdk.go
efekarakus: Right, I'm planning to send a separate PR for each of the following sections:
1.  Add to `templates/` the scaffold code for the CDK: `override.ts` and the other files.
2. Add a `ScaffoldWithCDK` func in the `override` pkg
3. Update `svc_package.go` and `svc_deploy.go` to use `override`
4. Create `svc_override.go`, `job_override.go` and so on to use `override.ScaffoldWithCDK`.

internal/pkg/override/cdk.go
```diff
@@ -0,0 +1,85 @@
...
+// Override should be invoked after Install.
+func (cdk *CDK) Override(body []byte) ([]byte, error) {
+	buildPath := filepath.Join(cdk.rootAbsPath, ".build")
+	if err := cdk.fs.MkdirAll(buildPath, fs.ModeDir); err != nil {
+		return nil, fmt.Errorf("create %s directory to store the CloudFormation template body: %w", buildPath, err)
+	}
+	inputPath := filepath.Join(buildPath, "in.yml")
+	if err := afero.WriteFile(cdk.fs, inputPath, body, 0644); err != nil {
+		return nil, fmt.Errorf("write CloudFormation template body content at %s: %w", inputPath, err)
+	}
```
dannyrandall, line 78: i wonder if we should delete this file/directory after running the command 🤔 just because it's probably _not_ covered by anyone's gitignore normally...

internal/pkg/override/override.go
```diff
@@ -0,0 +1,37 @@
...
+
+// An Overrider is the interface to apply transformations to a CloudFormation template body.
+type Overrider interface {
+	Override(body []byte) (transformed []byte, err error)
+}
+
+// Bytes applies an Overrider to a given byte slice and returns the transformed byte slice.
+// If the Overrider has to install any dependencies, then Install is invoked before Override.
+// If the Overrider has to clean up data from the transformed template, then CleanUp is invoked before exiting.
+func Bytes(body []byte, ovrdr Overrider) ([]byte, error) {
```
dannyrandall: i sorta feel like this might be a premature abstraction - CDK is the only known `Overrider` impl right now that will use `Install()`/`CleanUp()`, I wonder if it should just live in the `Override()` function of `CDK{}`? Then if it's a common pattern we can extract then? 

if you think YAML overlays will use these, then i think this makes sense.

dannyrandall (commented):
looks great overall! just a couple thoughts

internal/pkg/override/override_test.go
```diff
@@ -0,0 +1,79 @@
...
+		ovrdr := &mockOverriderWithDeps{
+			mockOverrider: &mockOverrider{},
+			cleanUp: func(_ []byte) ([]byte, error) {
+				return nil, errors.New("unmarshal yaml")
+			},
+		}
+		_, err := Bytes(nil, ovrdr)
+		require.EqualError(t, err, "clean up overriden document: unmarshal yaml")
+	})
+	t.Run("should install, override, and then return cleaned up template", func(t *testing.T) {
```
Lou1415926: Do we also want to check the mock counter?

internal/pkg/override/override_test.go
```diff
@@ -0,0 +1,79 @@
...
+		ovrdr := &mockOverriderWithDeps{
+			mockOverrider: &mockOverrider{out: []byte("hello")},
+			cleanUp: func(in []byte) ([]byte, error) {
```
Lou1415926: Is the intention here to test whether `CleanUp` is called?

internal/pkg/override/cdk.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// Install runs "npm install" insides the overrides/ directory to pull the latest dependencies.
+func (cdk *CDK) Install() error {
+	if _, err := cdk.exec.LookPath("npm"); err != nil {
+		return &errNPMUnavailable{parent: err}
+	}
+
+	cmd := cdk.exec.Command("npm", "install")
```
Lou1415926, line 61: Is `package.json` a part of the scaffold code that gets generated along with `override.ts`?

Lou1415926 (commented):
LGTM! just a few questsion!

internal/pkg/override/errors.go
```diff
@@ -0,0 +1,19 @@
...
+package override
+
+import "fmt"
+
+type errNPMUnavailable struct {
+	parent error
+}
+
+func (err *errNPMUnavailable) Error() string {
+	return fmt.Sprintf(`"npm" is required to override with the Cloud Development Kit: %v`, err.parent)
```
iamhopaul123: Maybe make this error msg more explicit. Like starting with
`"npm" is not available ...` or `cannot find "npm"...`

internal/pkg/override/cdk.go
```diff
@@ -0,0 +1,85 @@
...
+	cmd.Stdout = io.MultiWriter(buf, cdk.out)
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf(`run %q: %w`, cmd.String(), err)
+	}
+	return buf.Bytes(), nil
+}
+
+// CleanUp removes any CDK metadata resource from the CloudFormation template body.
+func (cdk *CDK) CleanUp(body []byte) ([]byte, error) {
+	return nil, errors.New("implement me")
```
iamhopaul123: 👍

internal/pkg/override/cdk.go
efekarakus: We will be writing a .gitignore file for the users when scaffolding the `overrides/` directory with `copilot [noun] override` (see [here](https://github.com/aws/copilot-cli/issues/4208) for all the files that will be generated).

Does that mitigate your concern?

internal/pkg/override/override.go
efekarakus: Done!

internal/pkg/override/override_test.go
efekarakus: I followed Danny's feedback and completely removed the `override.Bytes` function 😂 now all the logic exists inside the `Override` method part of `*CDK`. 

Let me know what you think

internal/pkg/override/cdk.go
efekarakus: Yup, see [here](https://github.com/aws/copilot-cli/issues/4208) for a list of files that will be generated part of scaffold.

internal/pkg/override/errors.go
efekarakus: Done, prefixed with "npm cannot be found:"

internal/pkg/override/cdk.go
dannyrandall: oh sweet! yeah, works for me 😊

internal/pkg/override/cdk.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	return cdk.cleanUp(out)
+}
+
+func (cdk *CDK) install() error {
+	if _, err := cdk.exec.LookPath("npm"); err != nil {
+		return &errNPMUnavailable{parent: err}
+	}
+
+	cmd := cdk.exec.Command("npm", "install")
```
dannyrandall, line 61: nit
```suggestion
	cmd := cdk.exec.Command("npm", "install")
	if cmd.Err != nil {
	  return &errNPMUnavailable{parent: err}
	}
```

dannyrandall (approved):
🚀

Lou1415926 (approved):

PR #4258, opened 2022-12-08
feat: deploy environment addons with customized parameters

Lou1415926:
Support #4219.


internal/pkg/addon/addons.go
```diff
@@ -21,7 +21,7 @@ const (
 
 var (
 	wkldAddonsParameterReservedKeys = []string{"App", "Env", "Name"}
-	envAddonsParameterReservedKeys  = []string{"App", "Name"}
+	envAddonsParameterReservedKeys  = []string{"App", "Env"}
```
efekarakus, line 24: Interesting, why did you prefer this change?

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -646,6 +646,20 @@ Resources:
                   - logs:DescribeLogStreams
                 Resource: "*"
 {{- end}}
+  {{- if .Addons}}
+  AddonsStack:
+    Metadata:
+      'aws:copilot:description': 'An Addons CloudFormation Stack for your additional AWS resources'
```
huanjani, line 652: nit:
```suggestion
      'aws:copilot:description': 'An addons CloudFormation stack for your additional AWS resources'
```

huanjani (approved):
LGTM!

internal/pkg/cli/deploy/env.go
```diff
@@ -118,15 +132,38 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
 	if err != nil {
 		return nil, fmt.Errorf("initialize env describer: %w", err)
 	}
+	ws, err := workspace.Use(afero.NewOsFs())
```
efekarakus, line 135: nit: should we use pass down the dependency to the constructor with the `NewEnvDeployerInput` since this is most likely already initialized in the `cli` pkg?

internal/pkg/cli/deploy/env.go
```diff
@@ -118,15 +132,38 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+	if err != nil {
+		return nil, err
+	}
 	cfnClient := deploycfn.New(envManagerSession, deploycfn.WithProgressTracker(os.Stderr))
-	return &envDeployer{
+
+	var addons addons
+	if in.EnvAddonsFeatureFlag {
+		addonsStack, err := addon.ParseFromEnv(ws)
```
efekarakus, line 143: This seems OK just a bit unusual to have this done within the constructor 💭 but I think it makes sense given that `UploadArtifacts` and `AddonsTemplate` both require it.

Perhaps an alternative way of writing would be:
```go
type envDeployer struct {
  // existing fields...
  
  parseAddonsOnce sync.Once
  addonsStack  stackBuilder
  addonsErr  error
}


func (d *envDeployer) parseAddons() (stackBuilder, error) {
  d.parseAddonsOnce.Do(func() {
    // parse and assign to fields.
  })
  return d.addonsStack, d.addonsErr
}
```
The main difference is that the parsing-cost is only incurred if parsing addons is necessary. However, if you deem all methods need the addon template then keeping it in the constructor like you did is better.

internal/pkg/cli/deploy/env.go
```diff
@@ -142,14 +179,213 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+	CustomResourcesURLs map[string]string
+	Manifest            *manifest.Environment
+	ForceNewUpdate      bool
+	RawManifest         []byte
+	PermissionsBoundary string
+	DisableRollback     bool
+}
+
+// GenerateCloudFormationTemplate returns the environment stack's template and parameter configuration.
+func (d *envDeployer) GenerateCloudFormationTemplate(in *DeployEnvironmentInput) (*GenerateCloudFormationTemplateOutput, error) {
```
efekarakus, line 241: Is there any difference here? (just asking to make sure I'm not overlooking something)

internal/pkg/cli/deploy/env.go
```diff
@@ -142,14 +179,213 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+		return nil, fmt.Errorf("generate stack template parameters: %w", err)
+	}
+	return &GenerateCloudFormationTemplateOutput{
+		Template:   tpl,
+		Parameters: params,
+	}, nil
+}
+
+// DeployEnvironment deploys an environment using CloudFormation.
+func (d *envDeployer) DeployEnvironment(in *DeployEnvironmentInput) error {
```
efekarakus, line 270: Is there any difference here? (just asking to make sure I'm not overlooking something)

internal/pkg/deploy/env.go
```diff
@@ -18,6 +18,16 @@ const (
 	EnvTemplateVersionBootstrap = "bootstrap"
 )
 
+type stackConfigurer interface {
+	Parameters() (string, error)
```
efekarakus, line 22: Why is `Parameters` a single `string`? I probably wrote somethign like for the previous addons 😅 feels strange maybe there was a good reason

internal/pkg/deploy/env.go
```diff
@@ -18,6 +18,16 @@ const (
...
+}
+
+// Addons holds information about the addons.
+type Addons struct {
+	URL    string
+	Config stackConfigurer
```
efekarakus, line 28: Do we want the interface to be private?

internal/pkg/deploy/env.go
Lou1415926: Perhaps we could rename the interface:
```go
type nestedStackConfigurer interface { // Rename the interface because `CustomParameters` is specific to nested stacks.
  Parameters() (string, error)  // Rename the method.
}
```


Basically, this `Parameters()` method is slightly different than the `Parameters()` of other "stack configurer"s - for example `EnvStackConfig` - that returns a list of `cloudformation.Parameter`s. Those `Parameters()` gives values to the "Parameters" section of a stack. They are later fed into an API call. 

This `Parameters()` method is meant to pass values from the parent stack to the addons stack's parameters, and is later rendered into the template.
```yaml
{{- if .Addons}}
  AddonsStack:
    Metadata:
      'aws:copilot:description': 'An addons CloudFormation stack for your additional AWS resources'
    Type: AWS::CloudFormation::Stack
    Properties:
      Parameters:
        App: !Ref AppName
        Env: !Ref EnvironmentName
        {{- if .Addons.ExtraParams }} # <- This is from `Parameters() (string, error)`
        {{ .Addons.ExtraParams }}
        {{- end}}
      TemplateURL: {{.Addons.URL}}
{{- end }}
```

This is due to the different ways of how a stack is configured (via API) vs. how a nested stack is configured (from parent).

internal/pkg/addon/addons.go
Lou1415926: Given that people have been writing workload addons, in which they have been referring to env via `${Env}` in their workload addons, I was assuming they'd more naturally use `${Env}`  instead of `${Name}` to refer to the name of an env in the env addons as well.

The example that I was mentally running through was when users want to export some `Outputs`, they'd probably write:
```yaml
Outputs:
  DBSecurityGroup:
     Value: !DBSecurityGroup
     Name: !Ref ${App}-${Env}-SecurityGroup # vs  Name: !Ref ${App}-${Name}-SecurityGroup
```

The `${Env}` version is clear in that the output is namespaced by env name.

internal/pkg/template/templates/environment/cf.yml
Lou1415926: I addressed it in #4264!

internal/pkg/cli/deploy/env.go
Lou1415926: This will be addressed in a PR that follows #4264 !

internal/pkg/cli/deploy/env.go
Lou1415926: sgtm - will address in a PR that follows https://github.com/aws/copilot-cli/pull/4264 !

internal/pkg/cli/deploy/env.go
Lou1415926: No difference! I also added a comment in https://github.com/aws/copilot-cli/pull/4264#discussion_r1046571293 so that it's clear for reviewers of the new PR.

Lou1415926:
Closing in favor of smaller PR that touches less packages #4264 !

PR #4263, opened 2022-12-13
chore(override): implement `Lookup` to check if the path is CDK or YAML

efekarakus:


internal/pkg/override/override.go
```diff
@@ -0,0 +1,77 @@
...
+// If path does not exist or is an empty directory, then return an ErrNotExist.
+// If path is a YAML patch document, then IsYAMLPatch evaluates to true.
+// If path is a directory that contains a cdk.json file, then IsCDK evaluates to true.
+func Lookup(path string, fs afero.Fs) (Info, error) {
+	stat, err := fs.Stat(path)
+	if err != nil {
+		return unknownOverrider, &ErrNotExist{parent: err}
+	}
+
+	if stat.IsDir() {
```
paragbhingre, line 46: Correct me if I am wrong. Isn't yaml patch is also going to live under `overrides` directory? Just curious to know the reason behind putting `lookupCDK` inside `IsDir` if condition.

internal/pkg/override/errors.go
```diff
@@ -17,3 +19,12 @@ func (err *errNPMUnavailable) Error() string {
 func (err *errNPMUnavailable) RecommendActions() string {
 	return fmt.Sprintf(`Please follow instructions at: %q to install "npm"`, "https://docs.npmjs.com/downloading-and-installing-node-js-and-npm")
 }
+
+// ErrNotExist occurs when an overrides/ directory is either empty or does not exist.
+type ErrNotExist struct {
```
paragbhingre, line 24: How about renaming this as `ErrDirNotExist` for clearer understanding of what does not exist.

internal/pkg/override/cdk.go
```diff
@@ -29,8 +29,8 @@ type CDK struct {
 	exec Executable // For testing os/exec calls.
 }
 
-// NewCDK instantiates a new CDK Overrider with the file system pointing ot the overrides/ dir.
-func NewCDK(root string, stdout io.Writer, fs afero.Fs, exec Executable) *CDK {
```
iamhopaul123, line 33: I feel like `NewCDK` seems to be more consistent with our naming convention 🤔

internal/pkg/override/override.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/spf13/afero"
+)
+
+// Info holds metadata about an overrider.
+type Info int
+
+const (
+	unknownOverrider Info = iota
+	cdkOverrider
+	yamlPatchOverrider
```
iamhopaul123, line 23: Why do we use patch instead of overlay?

iamhopaul123 (approved):
LGTM overall!

internal/pkg/override/override.go
```diff
@@ -0,0 +1,77 @@
...
+func Lookup(path string, fs afero.Fs) (Info, error) {
+	stat, err := fs.Stat(path)
+	if err != nil {
+		return unknownOverrider, &ErrNotExist{parent: err}
+	}
+
+	if stat.IsDir() {
+		return lookupCDK(path, fs)
+	}
+	return lookupYAMLPatch(path, fs)
```
Lou1415926, line 49: Should we say that it's a YAML patch only if `path` points to a file & it ends with `.yml/.yaml`? Or is this type of filename parsing done in upstream?

internal/pkg/override/cdk.go
efekarakus: My reasoning is that I got inspired by the context package: https://pkg.go.dev/context (e.g. `context.WithTimeout`).
Reading `cdk := override.WithCDK("path/to/overrides", buf, fs, os.Exec)` seems easier to understand than `cdk := override.NewCDK(...)`.

I'll keep it as is unless you feel strongly against it.

internal/pkg/override/errors.go
efekarakus: I've updated the comment to make this a bit clearer. It's not necessarily a directory 😅 It's possible that path is just `path/to/cfn.patch.yml` (I took the same err name as [`fs.ErrNotExist`](https://pkg.go.dev/io/fs#pkg-variables)) so I kept this as is.

internal/pkg/override/override.go
efekarakus: They're the same thing, the [RFC](https://www.rfc-editor.org/rfc/rfc6902) is called "JavaScript Object Notation (JSON) Patch" which is why I'm calling it "YAML Patch"

internal/pkg/override/override.go
efekarakus: When scaffolding we will create the `cfn.patch.yml` under the `overrides/` dir that's correct but we want users to be able to customize the location of this file and just move it wherever makes sense for them (e.g. they might want all manifests to use the same yaml patch file).

This is why the input to `Lookup` is any path, we don't make any assumtion about an `overrides` dir at all

internal/pkg/override/override.go
efekarakus: Updated!

Lou1415926 (approved):
:shipit:

PR #4264, opened 2022-12-13
chore: conditionally render environment addons as a nested stack

Lou1415926:
This PR is strictly a part of the original #4258. 

Support #4219.


internal/pkg/cli/deploy/env.go
```diff
@@ -150,108 +150,6 @@ func (d *envDeployer) Validate(mft *manifest.Environment) error {
 	return d.validateCDN(mft)
 }
 
-func (d *envDeployer) validateCDN(mft *manifest.Environment) error {
```
Lou1415926, line 153: There isn't any code change in `cli/deploy/env.go`. These are just moving codes around so that private methods are moved to the end of the file.

iamhopaul123: I personally prefer to put all the methods that belong to one struct together and then sort them by public -> private, because they form a more consistent story (and consistent with most files in our codebase).

internal/pkg/addon/addons.go
```diff
@@ -21,7 +21,7 @@ const (
 
 var (
 	wkldAddonsParameterReservedKeys = []string{"App", "Env", "Name"}
-	envAddonsParameterReservedKeys  = []string{"App", "Name"}
+	envAddonsParameterReservedKeys  = []string{"App", "Env"}
```
iamhopaul123, line 24: Why do we make this change?

internal/pkg/cli/env_package.go
```diff
@@ -65,7 +65,7 @@ type packageEnvOpts struct {
 	paramsWriter io.WriteCloser
 
 	newInterpolator func(appName, envName string) interpolator
-	newEnvDeployer  func() (envPackager, error)
```
iamhopaul123, line 68: 🤦

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -646,6 +646,21 @@ Resources:
                   - logs:DescribeLogStreams
                 Resource: "*"
 {{- end}}
+{{- if .Addons}}
+  AddonsStack:
+    Metadata:
+      'aws:copilot:description': 'An addons CloudFormation stack for your additional AWS resources'
```
iamhopaul123: Maybe
```suggestion
      'aws:copilot:description': 'A CloudFormation nested stack stack for your addons resources'
```
?

internal/pkg/template/env.go
```diff
@@ -94,6 +94,12 @@ var (
 	}
 )
 
+// Addons holds data about an aggregated addons stack.
+type Addons struct {
```
iamhopaul123, line 98: Does this only apply to env addons or any addons?

internal/pkg/addon/addons.go
```diff
@@ -36,7 +36,7 @@ var (
 	}()
 )
 
-type workspaceReader interface {
+type workspaceAddonsReader interface {
```
efekarakus: nit: should the interface be public since it's used by exported functions?

efekarakus (approved):
Looks good to me! I have the same question about the `Addons` type in `template` as @iamhopaul123 but don't block on me

internal/pkg/addon/addons.go
Lou1415926: I replied to Efe here: https://github.com/aws/copilot-cli/pull/4258#discussion_r1046557439

internal/pkg/template/env.go
Lou1415926: This is generic enough for any addons. For now it's only used for env addons.

internal/pkg/cli/deploy/env.go
Lou1415926: All of the methods' receivers are the same in this file though...so my change here doesn't contradict with your prefernce 🤔

iamhopaul123: oh gotcha sorry i misread

internal/pkg/addon/addons.go
Lou1415926: good catch - addressed!

internal/pkg/template/templates/environment/cf.yml
Lou1415926: I updated it to `'A CloudFormation nested stack for your additional AWS resources'`!

iamhopaul123 (approved):

PR #4267, opened 2022-12-13
chore: add CFN fields for deployment alarms

huanjani:
These changes allow lists of alarm names as string values in the `deployment.rollback_alarms` manifest field, and transform them into the corresponding CFN fields. While map values are currently accepted, those are currently a no-op, but will soon be used for basic alarm creation.


internal/pkg/deploy/cloudformation/stack/transformers_test.go
```diff
@@ -2053,6 +2053,61 @@ func Test_convertSubscribe(t *testing.T) {
...
+			in: manifest.DeploymentConfiguration{},
+			out: template.DeploymentConfigurationOpts{
+				MinHealthyPercent: minHealthyPercentDefault,
+				MaxPercent:        maxPercentDefault,
+			},
+		},
+		"if alarm names entered, populate directly": {
+			in: manifest.DeploymentConfiguration{
+				Alarms: manifest.RollbackAlarmArgsOrNames{
+					AlarmNames: []*string{
```
paragbhingre, line 2089: What happens when AlarmNames and deployment strategy both are specified?

internal/pkg/manifest/workload_ecs.go
```diff
@@ -85,13 +85,57 @@ type ImageWithPortAndHealthcheck struct {
...
+// It supports unmarshalling yaml which can either be of type AlarmArgs or a list of strings.
+type RollbackAlarmArgsOrNames struct {
+	AlarmNames []*string
+	AlarmArgs
+}
+
+// UnmarshalYAML overrides the default YAML unmarshaling logic for the RollbackAlarmArgsOrNames
+// struct, allowing it to perform more complex unmarshaling behavior.
+// This method implements the yaml.Unmarshaler (v3) interface.
+func (r *RollbackAlarmArgsOrNames) UnmarshalYAML(value *yaml.Node) error {
```
paragbhingre: Can we try using Danny's Union implementation here? That will help get rid of this method.

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
```diff
@@ -16,6 +16,12 @@ DeploymentConfiguration:
     Rollback: true
   MinimumHealthyPercent: {{ .DeploymentConfiguration.MinHealthyPercent }}
   MaximumPercent: {{ .DeploymentConfiguration.MaxPercent }}
+  {{- if .DeploymentConfiguration.RollbackAlarms }}
+  Alarms:
+    AlarmNames: {{ .DeploymentConfiguration.RollbackAlarms }}
```
Lou1415926, line 21: Just double checking: does  `.RollbackAlarms` - which is `[]*string` instead of `[]string` -  render properly here?

internal/pkg/manifest/validate.go
```diff
@@ -115,6 +115,24 @@ func (d DeploymentConfiguration) validate() error {
 	return fmt.Errorf("invalid rolling deployment strategy %s, must be one of %s",
 		aws.StringValue(d.Rolling),
 		english.WordSeries(ecsRollingUpdateStrategies, "or"))
+	if err := d.Alarms.validate(); err != nil {
+		return fmt.Errorf("validate rollback alarms: %w", err)
```
```suggestion
		return fmt.Errorf(`validate "rollback_alarms": %w`, err)
```

internal/pkg/manifest/validate.go
```diff
@@ -115,6 +115,24 @@ func (d DeploymentConfiguration) validate() error {
...
+	}
+	return nil
+}
+
+func (r RollbackAlarmArgsOrNames) validate() error {
+	if r.AlarmArgs.isEmpty() {
+		return nil
+	}
```
Lou1415926: nit: I feel like this block could be written in L134 in `func (a AlarmArgs) validate() errror`  - my reasoning is that whether being empty is valid or not sound like a judgement that `AlarmArgs` should make.

internal/pkg/manifest/workload_ecs.go
```diff
@@ -85,13 +85,57 @@ type ImageWithPortAndHealthcheck struct {
...
+// AlarmArgs represents specs of CloudWatch alarms for deployment rollbacks.
+type AlarmArgs struct {
+	CPUUtilization    *int `yaml:"cpu_utilization"`
+	MemoryUtilization *int `yaml:"memory_utilization"`
+}
+
+// RollbackAlarmArgsOrNames represents alarms for rolling back deployments.
+// It supports unmarshalling yaml which can either be of type AlarmArgs or a list of strings.
+type RollbackAlarmArgsOrNames struct {
+	AlarmNames []*string
```
Lou1415926: Why is this a list of pointer to string? Is this because of `mergo`?

internal/pkg/manifest/workload_ecs.go
```diff
@@ -85,13 +85,57 @@ type ImageWithPortAndHealthcheck struct {
...
 // DeploymentConfiguration represents the deployment strategies for a service.
 type DeploymentConfiguration struct {
-	Rolling *string `yaml:"rolling"`
+	Rolling *string                  `yaml:"rolling"`
+	Alarms  RollbackAlarmArgsOrNames `yaml:"rollback_alarms"`
 }
 
 func (d *DeploymentConfiguration) isEmpty() bool {
-	return d == nil || d.Rolling == nil
+	return d.Rolling == nil && d.Alarms.IsEmpty()
```
Lou1415926: Can we make sure that whoever calls `d.isEmpty()` is not counting on `isEmpty()` to determine whether `d==nil`? That is, I'm worried that we have codes like this:
```go
// d is a (*DeploymentConfiguration)(nil)
if !d.isEmpty() {}
```

This code wouldn't panic before - `d.isEmpty()` would have evaluated `d==nil` and return `true`. Now it would panic, because `d.isEmpty` will try to evaluate `d.Rolling` which is fetching data from a `nil` pointer.

internal/pkg/manifest/workload_ecs.go
huanjani: done

internal/pkg/manifest/workload_ecs.go
huanjani: changed back 👍🏼

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
huanjani: it's now `[]*string` in `manifest` but `[]string` in template

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -812,6 +812,16 @@ func convertDeploymentConfig(deploymentConfig manifest.DeploymentConfiguration)
 		deployConfigs.MinHealthyPercent = minHealthyPercentDefault
 		deployConfigs.MaxPercent = maxPercentDefault
 	}
+	if !deploymentConfig.Alarms.IsZero() {
```
dannyrandall: i think you could probably remove this condition and just use `IsBasic()` and `IsAdvanced()`

internal/pkg/manifest/workload_ecs.go
```diff
@@ -85,13 +85,26 @@ type ImageWithPortAndHealthcheck struct {
...
+// It supports unmarshalling yaml which can either be of type AlarmArgs or a list of strings.
+type RollbackAlarmArgsOrNames struct {
+	Union[[]*string, AlarmArgs]
+}
+
 // DeploymentConfiguration represents the deployment strategies for a service.
 type DeploymentConfiguration struct {
-	Rolling *string `yaml:"rolling"`
+	Rolling *string                  `yaml:"rolling"`
+	Alarms  RollbackAlarmArgsOrNames `yaml:"rollback_alarms"`
```
dannyrandall: based on the usage - i don't think you need a separate struct here!
```suggestion
	Alarms  Union[[]*string, AlarmArgs] `yaml:"rollback_alarms"`
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -812,6 +812,16 @@ func convertDeploymentConfig(deploymentConfig manifest.DeploymentConfiguration)
...
+		if deploymentConfig.Alarms.IsBasic() {
+			alarmStrings := make([]string, len(deploymentConfig.Alarms.Basic))
+			for i, alarm := range deploymentConfig.Alarms.Basic {
+				alarmStrings[i] = aws.StringValue(alarm)
+			}
```
dannyrandall: :) https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.44.161/aws#StringValueSlice

internal/pkg/manifest/workload_ecs.go
dannyrandall: also - i think the first type can just be `[]string`, not `[]*string`

internal/pkg/template/workload.go
```diff
@@ -457,7 +457,8 @@ type DeploymentConfigurationOpts struct {
 	// The lower limit on the number of tasks that should be running during a service deployment or when a container instance is draining.
 	MinHealthyPercent int
 	// The upper limit on the number of tasks that should be running during a service deployment or when a container instance is draining.
-	MaxPercent int
+	MaxPercent     int
```
dannyrandall, line 460: i assume you'll use this in a future PR?

internal/pkg/manifest/validate.go
```diff
@@ -107,14 +107,38 @@ func (d DeploymentConfiguration) validate() error {
...
+			}
+		}
+		return fmt.Errorf("invalid rolling deployment strategy %q, must be one of %s",
+			aws.StringValue(d.Rolling),
+			english.WordSeries(ecsRollingUpdateStrategies, "or"))
+	}
+	return nil
+}
+
+func (r RollbackAlarmArgsOrNames) validate() error {
```
dannyrandall: if we drop the `RollbackAlarmArgsOrNames` struct, we can remove this function and just leave the `AlarmArgs` validation

internal/pkg/deploy/cloudformation/stack/transformers_test.go
```diff
@@ -2053,6 +2053,55 @@ func Test_convertSubscribe(t *testing.T) {
 	}
 }
 
+func Test_convertDeploymentConfig(t *testing.T) {
```
Lou1415926, line 2056: Thanks for adding these tests!!!

internal/pkg/manifest/validate.go
```diff
@@ -107,14 +107,25 @@ func (d DeploymentConfiguration) validate() error {
 	if d.isEmpty() {
 		return nil
 	}
-	for _, validStrategy := range ecsRollingUpdateStrategies {
-		if strings.EqualFold(aws.StringValue(d.Rolling), validStrategy) {
-			return nil
+	if err := d.RollbackAlarms.validate(); err != nil {
+		return fmt.Errorf(`validate "rollback alarms": %w`, err)
```
```suggestion
		return fmt.Errorf(`validate "rollback_alarms": %w`, err)
```

Lou1415926 (approved):
:shipit:

paragbhingre (approved):
Looks good to me 🤩

PR #4268, opened 2022-12-13
refactor: manifest pkg to support multiple ports

paragbhingre:
This PR refactors the manifest package in order to create a list of exposed ports. Though we are still accepting a single port from the customer. Also introducing new field `target_port` under `http` that can open another port internally from either primary container or sidecar container. In this PR because there are no CFN changes so nothing really reflects in the CFN. 

This PR is a part of #4192 


internal/pkg/manifest/job.go
```diff
@@ -192,3 +193,23 @@ func newDefaultScheduledJob() *ScheduledJob {
 		Environments: map[string]*ScheduledJobConfig{},
 	}
 }
+
+// ExposedPorts returns all the ports that are sidecar container ports available to receive traffic.
+func (j *ScheduledJob) ExposedPorts() []ExposedPort {
+
```
dannyrandall: super nit but these spaces 😭 they just don't match the rest of the codebase
```suggestion
```

internal/pkg/manifest/backend_svc.go
```diff
@@ -188,3 +189,26 @@ func newDefaultBackendService() *BackendService {
...
+
+// ExposedPorts returns all the ports that are container ports available to receive traffic.
+func (b *BackendService) ExposedPorts() []ExposedPort {
+
+	var exposedPortList []ExposedPort
+	exposedPorts := make(map[int]ExposedPort)
+
+	workloadName := aws.StringValue(b.Name)
+	b.ImageConfig.exposedPorts(exposedPorts, workloadName)
+	b.RoutingRule.exposedPorts(exposedPorts, workloadName)
```
dannyrandall: maybe these are from a different PR, but how are we handling conflicts? shouldn't that be an error when parsing the manifest? 

if we already know the port numbers are unique (i.e., errors occurred before this is called), then i'd prefer to make this return a slice instead of a map so we can just have a single `exposedPorts` slice instead of a map+slice

internal/pkg/manifest/workload_ecs.go
```diff
@@ -94,6 +94,13 @@ func (d *DeploymentConfiguration) isEmpty() bool {
 	return d == nil || d.Rolling == nil
 }
 
+// ExposedPort will hold the port mapping configuration.
+type ExposedPort struct {
+	ContainerName string // The name of the container that exposes this port.
+	Port          int    // The port number.
```
dannyrandall: maybe a `uint16`?

internal/pkg/manifest/validate.go
```diff
@@ -1721,6 +1738,47 @@ func validateDepsForEssentialContainers(deps map[string]containerDependency) err
...
+func validateExposedPorts(opts validateExposedPortsOpts) error {
+	exposedPorts := make(map[uint16]string)
+
+	if opts.mainContainerPort != nil {
+		exposedPorts[aws.Uint16Value(opts.mainContainerPort)] = opts.mainContainerName
+	}
+
+	for name, sidecar := range opts.sidecarConfig {
+		if sidecar.Port != nil {
+			port, err := strconv.Atoi(aws.StringValue(sidecar.Port))
```
dannyrandall: does `Atoi` work if the port is `8080/tcp`?

internal/pkg/manifest/validate.go
```diff
@@ -1721,6 +1738,47 @@ func validateDepsForEssentialContainers(deps map[string]containerDependency) err
...
 
+func validateExposedPorts(opts validateExposedPortsOpts) error {
+	exposedPorts := make(map[uint16]string)
+
+	if opts.mainContainerPort != nil {
+		exposedPorts[aws.Uint16Value(opts.mainContainerPort)] = opts.mainContainerName
+	}
+
+	for name, sidecar := range opts.sidecarConfig {
+		if sidecar.Port != nil {
```
dannyrandall: instead of nesting all of this logic, it'd be nice to just `continue` in the opposite case:
```suggestion
		if sidecar.Port == nil {
		  continue
		}
```

internal/pkg/manifest/validate.go
```diff
@@ -1721,6 +1738,47 @@ func validateDepsForEssentialContainers(deps map[string]containerDependency) err
...
+			port, err := strconv.Atoi(aws.StringValue(sidecar.Port))
+			if err != nil {
+				return err
+			}
+			if _, ok := exposedPorts[uint16(port)]; !ok {
+				exposedPorts[uint16(port)] = name
+			} else {
```
dannyrandall: same nesting idea here, this could be:
```suggestion
			if _, ok := exposedPorts[uint16(port)]; ok {
				return err
			}
			exposedPorts[uint16(port)] = name
```

dannyrandall (approved):
couple of thoughts from the first pass!

internal/pkg/manifest/backend_svc.go
dannyrandall: looks like validate will have already run (i'm assuming) - so maybe we can drop the map and just create a single slice?

paragbhingre: Yeah validation will take care of the duplicates, so probably I can remove map from here and just use slice.

internal/pkg/manifest/validate.go
paragbhingre: Hmm good catch. I will parse it before directly using it.

internal/pkg/manifest/svc.go
```diff
@@ -442,3 +442,62 @@ func (s *stringOrFromCFN) UnmarshalYAML(value *yaml.Node) error {
 	}
 	return nil
 }
+
+func (cfg ImageWithPortAndHealthcheck) exposedPorts(workloadName string) (exposedPort []ExposedPort) {
```
efekarakus: nit: (same for the other methods)
```suggestion
func (cfg ImageWithPortAndHealthcheck) exposedPorts(workloadName string) (exposedPorts []ExposedPort) {
```
or
```suggestion
func (cfg ImageWithPortAndHealthcheck) exposedPorts(workloadName string) (ports []ExposedPort) {
```

internal/pkg/manifest/svc.go
```diff
@@ -442,3 +442,62 @@ func (s *stringOrFromCFN) UnmarshalYAML(value *yaml.Node) error {
...
+			Protocol:      "tcp",
+			ContainerName: containerName,
+		})
+	}
+	return exposedPort
+}
+
+func (sidecar SidecarConfig) exposedPorts(sidecarName string) (exposedPort []ExposedPort) {
+	if sidecar.Port != nil {
+		sidecarPort, protocol, _ := ParsePortMapping(sidecar.Port)
```
efekarakus: is it not possible that `sidecarPort` is `nil` on error, should we skip and return nil if it's malformed?

internal/pkg/manifest/svc.go
```diff
@@ -442,3 +442,62 @@ func (s *stringOrFromCFN) UnmarshalYAML(value *yaml.Node) error {
...
+		if protocol == nil {
+			protocol = aws.String("tcp")
+		}
+		port, _ := strconv.Atoi(aws.StringValue(sidecarPort))
```
efekarakus: same here

internal/pkg/manifest/validate.go
dannyrandall: it may work fine, i just wasn't sure!

internal/pkg/manifest/validate.go
```diff
@@ -1721,6 +1738,52 @@ func validateDepsForEssentialContainers(deps map[string]containerDependency) err
 	return nil
 }
 
+func validateExposedPorts(opts validateExposedPortsOpts) error {
```
efekarakus, line 1768: can we potentially simplify by leveraging the new `ExposedPort` method on all types?

```go
if err := validateExposedPorts(mft.ExposedPorts(), &mft.RoutingRule); err != nil  { return err }

func validateExposedPorts(ports []ExposedPort, http *RoutingRuleConfiguration) error {
  // (1) check duplicate in ports
  
  // (2) check if http.targetPort and http.targetContainer are compatible
}
```

internal/pkg/manifest/validate.go
dannyrandall: ok jk i just tested and it doesn't work :)

efekarakus (commented):
Looks great! Thank you so much for splitting up the PR. Just tiny suggestions and then :shipit:

internal/pkg/manifest/errors.go
```diff
@@ -189,3 +189,13 @@ func quoteStringSlice(in []string) []string {
...
 }
+
+type errContainersExposingSamePort struct {
+	firstContainer  string
+	secondContainer string
+	port            string
+}
+
+func (e *errContainersExposingSamePort) Error() string {
+	return fmt.Sprintf(`container %s and container %s are exposing the same port %s`, e.firstContainer, e.secondContainer, e.port)
```
dannyrandall:
```suggestion
	return fmt.Sprintf(`containers %q and %q are exposing the same port %d`, e.firstContainer, e.secondContainer, e.port)
```

internal/pkg/manifest/validate.go
```diff
@@ -1721,6 +1738,52 @@ func validateDepsForEssentialContainers(deps map[string]containerDependency) err
...
+	for name, sidecar := range opts.sidecarConfig {
+		if sidecar.Port == nil {
+			continue
+		}
+		sidecarPort, _, err := ParsePortMapping(sidecar.Port)
+		if err != nil {
+			return err
+		}
+
+		port, err := strconv.Atoi(aws.StringValue(sidecarPort))
```
```suggestion
		port, err := strconv.ParseUint(aws.StringValue(sidecarPort), 10, 16)
```

internal/pkg/manifest/validate.go
```diff
@@ -1721,6 +1738,52 @@ func validateDepsForEssentialContainers(deps map[string]containerDependency) err
...
+
+	if opts.mainContainerPort != nil {
+		exposedPorts[aws.Uint16Value(opts.mainContainerPort)] = opts.mainContainerName
+	}
+
+	for name, sidecar := range opts.sidecarConfig {
+		if sidecar.Port == nil {
+			continue
+		}
+		sidecarPort, _, err := ParsePortMapping(sidecar.Port)
```
dannyrandall: do we need to validate the protocol here?

internal/pkg/manifest/svc.go
paragbhingre: So here and below [here](https://github.com/aws/copilot-cli/pull/4268#discussion_r1048995639) this same implementation is validated inside validation and returns error if port is malformed. So we do not need to worry here as validation has already taken care of those scenarios. Here is [validation code](https://github.com/aws/copilot-cli/pull/4268/files#diff-5fc13795122fe030799a09ef956a8b66cec3d40bf1a81266323edbf3626b5685R1752-R1757) for your reference.

internal/pkg/manifest/validate.go
```diff
@@ -1721,6 +1738,52 @@ func validateDepsForEssentialContainers(deps map[string]containerDependency) err
...
+	if opts.targetPort != nil && opts.targetContainer != nil {
+		if exposedPorts[aws.Uint16Value(opts.targetPort)] != aws.StringValue(opts.targetContainer) {
+			return &errContainersExposingSamePort{
+				firstContainer:  aws.StringValue(opts.targetContainer),
+				secondContainer: exposedPorts[aws.Uint16Value(opts.targetPort)],
+				port:            strconv.Itoa(int(aws.Uint16Value(opts.targetPort))),
+			}
+		}
+	}
+
```
dannyrandall, line 1855: do we need to handle these cases? or maybe they're handled somewhere else?
1. `target_port` not in `exposedPorts`
2. `target_container` not exposing any ports (or not existing?)

internal/pkg/manifest/validate.go
paragbhingre: I think the current implementation is easier for us validate, because we are validating against the manifest inputs. Once we call `ExposePorts()` then there is only `ExposedPort` objects and to validate those objects for duplicates and `http.targetPort` and `http.targetContainer` are compatible use case will be same amount of code and effort. Instead current validation keeps it easy. We just need to call  `validateExposedPorts` with right inputs from all the types of services. Let me know what you think.

internal/pkg/manifest/validate.go
paragbhingre: Yes we are taking care of invalid target_port and target_container configuration in here.

internal/pkg/manifest/svc.go
```diff
@@ -442,3 +442,62 @@ func (s *stringOrFromCFN) UnmarshalYAML(value *yaml.Node) error {
...
+	// Read `image.port`
+	if cfg.Port != nil {
+		exposedPorts = append(exposedPorts, ExposedPort{
+			Port:          aws.Uint16Value(cfg.Port),
+			Protocol:      "tcp",
+			ContainerName: workloadName,
+		})
+	}
+	return exposedPorts
+}
```
dannyrandall: https://google.github.io/styleguide/go/decisions#named-result-parameters

> Don’t use named result parameters when the names produce [unnecessary repetition](https://google.github.io/styleguide/go/decisions#repetitive-with-type).

(also, we're not really saving any amount of code with it)
```suggestion
func (cfg ImageWithPortAndHealthcheck) exposedPorts(workloadName string) () {
	// Read `image.port`
	if cfg.Port == nil {
	    return nil
	}
	return []ExposedPort{
	    {
		    Port:          aws.Uint16Value(cfg.Port),
			Protocol:      "tcp",
			ContainerName: workloadName,
	    }
	}
}
```

dannyrandall (approved):
lgtm! just one small thing

internal/pkg/manifest/svc.go
efekarakus: Gotcha!

I'm just always paranoid that something will go wrong if the validation step gets skipped :mildpanic-intensifies:
(this has nothing to do with your PR): We've been creating temporal coupling between these methods where it's assume that `Validate` is invoked before anything.

Perhaps, this is fine - would you mind adding a comment to the `ExposedPorts` methods saying that `Validate` should be invoked before calling this method?

efekarakus (approved):
👍 after addressing Danny's feedback and I just have a small comment update request 🥺

paragbhingre:
> something will go wrong if the validation step gets skipped :mildpanic-intensifies:

@efekarakus  I think I agree with this and it also intensified my mildpanic 😁. It is better to have error returned. I have made the required changes.

internal/pkg/manifest/svc.go
```diff
@@ -442,3 +443,90 @@ func (s *stringOrFromCFN) UnmarshalYAML(value *yaml.Node) error {
 	}
 	return nil
 }
+
+func (cfg ImageWithPortAndHealthcheck) exposedPorts(workloadName string) []ExposedPort {
+	// Read `image.port`
```
efekarakus: nit: can we remove these `// Read` comments in this file

internal/pkg/manifest/svc.go
```diff
@@ -442,3 +443,90 @@ func (s *stringOrFromCFN) UnmarshalYAML(value *yaml.Node) error {
...
+	return []ExposedPort{
+		{
+			Port:          aws.Uint16Value(cfg.Port),
+			Protocol:      "tcp",
+			ContainerName: workloadName,
+		},
+	}
+}
+
+func (rr RoutingRuleConfiguration) exposedPorts(exposedPorts []ExposedPort, workloadName string) []ExposedPort {
```
efekarakus, line 476: Since this is really the only one that's particularly tricky, I think it's worthwhile adding a comment explaining why this method takes an extra parameter.
```suggestion
// exportPorts returns any new ports that should be exposed given the load balancer configuration that's not part of the existing containerPorts.
func (rr RoutingRuleConfiguration) exposedPorts(containerPorts []ExposedPort, defaultContainerName string) []ExposedPort {
```

internal/pkg/manifest/validate.go
```diff
@@ -1721,6 +1754,79 @@ func validateDepsForEssentialContainers(deps map[string]containerDependency) err
...
+	if opts.mainContainerPort != nil {
+		exposedPorts[aws.Uint16Value(opts.mainContainerPort)] = opts.mainContainerName
+	}
+
+	for name, sidecar := range opts.sidecarConfig {
+		if sidecar.Port == nil {
+			continue
+		}
+		sidecarPort, protocol, err := ParsePortMapping(sidecar.Port)
```
efekarakus, line 1792: huh do we even accept `port/protocol` for sidecars, I thought it was strictly only for `nlb.port`?

It's odd to check for `nlbValidProtocols` for all sidecars because they might not be an nlb in the manifest at all 🤔

internal/pkg/manifest/validate.go
```diff
@@ -1721,6 +1754,79 @@ func validateDepsForEssentialContainers(deps map[string]containerDependency) err
 	return nil
 }
 
+func validateExposedPorts(opts validateExposedPortsOpts) error {
```
efekarakus, line 1768: What do you think of a structure like this:
```
func validateExposedPorts(opts validateExportOptsOpts) error {
   containerNameFor := make(map[uint16]string)
   if err := validateMainContainerPort(containerNameFor, opts); err != nil { return err }
   if err := validateSidecarContainerPorts(containerNameFor, opts); err != nil { return err }
   if err := validateALBPorts(containerNameFor, opts); err != nil { return err }
   return nil
}
```

internal/pkg/manifest/validate.go
```diff
@@ -1721,6 +1754,79 @@ func validateDepsForEssentialContainers(deps map[string]containerDependency) err
...
+						port:            aws.Uint16Value(alb.TargetPort),
+					}
+				}
+			}
+		} else if alb.TargetContainer != nil {
+			containerName = aws.StringValue(alb.TargetContainer)
+		}
+		exposedPorts[aws.Uint16Value(alb.TargetPort)] = containerName
+	}
+	return nil
```
efekarakus, line 1856: how about NLB validation?

paragbhingre: I am cutting a new PR for adding NLB validation and adding NLB exposed ports to the []ExposedPort slice. I had ask the same question earlier if we should add NLB config to the same PR or as a separate PR.

internal/pkg/manifest/validate.go
paragbhingre: I think we will keep the implementation as is for now as we might support `udp` and `tcp` protocols for service connect in the future. Do you want to make separate valid protocols list for sidecars? I thought of reusing the existing ones. But happy to make separate list as well.

efekarakus: maybe while working on the NLB validation, we can rename it to `validPortProtocols` or something

internal/pkg/manifest/validate.go
efekarakus: 🤦 sorry my bad, I forgot we discussed that earlier, sounds good!

internal/pkg/manifest/validate.go
```diff
@@ -1755,12 +1755,24 @@ func validateDepsForEssentialContainers(deps map[string]containerDependency) err
 }
 
 func validateExposedPorts(opts validateExposedPortsOpts) error {
-	exposedPorts := make(map[uint16]string)
+	containerNameFor := make(map[uint16]string)
+	validateMainContainerPort(containerNameFor, opts)
+	if err := validateSidecarContainerPorts(containerNameFor, opts); err != nil {
+		return err
+	}
+	if err := validateALBPorts(containerNameFor, opts); err != nil {
```
efekarakus: nit: sorry about the rename, perhaps this is more accurate:
```
	populateMainContainerPort(containerNameFor, opts)
	if err := populateSidecarContainerPortsAndValidate(containerNameFor, opts); err != nil {
		return err
	}
	if err := populateALBPortsAndValidate(containerNameFor, opts); err != nil {
```

internal/pkg/manifest/validate.go
```diff
@@ -1755,12 +1755,24 @@ func validateDepsForEssentialContainers(deps map[string]containerDependency) err
...
+		return err
+	}
+	return nil
+}
 
+func validateMainContainerPort(containerNameFor map[uint16]string, opts validateExposedPortsOpts) {
 	if opts.mainContainerPort != nil {
```
efekarakus: can we flip the if statement to create guard-statements

```go
if opts.mainContainerPort == nil { return }
```

internal/pkg/manifest/validate.go
```diff
@@ -1787,16 +1799,19 @@ func validateExposedPorts(opts validateExposedPortsOpts) error {
...
 				port:            uint16(port),
 			}
 		}
-		exposedPorts[uint16(port)] = name
+		containerNameFor[uint16(port)] = name
 	}
+	return nil
+}
 
+func validateALBPorts(containerNameFor map[uint16]string, opts validateExposedPortsOpts) error {
```
efekarakus: same here can we flip the if-statement to reduce nesting:
```
if opts.alb == nil { return nil }
```

efekarakus (approved):
Looks good! just two tiny refactors

PR #4265, opened 2022-12-13
refactor: reduce `stack` package's dependency on `deploy` package

Lou1415926:
The refactorsin this PR are:
1. Changing the signature of `CreateAndRenderEnvironment`
2. Move `CreateEnvironmentInput` from `deploy` package to `stack` package
3. Rename `CreateEnvironmentInput` to `EnvConfig`

The original struct `CreateEnvironmentInput` is only used to construct `stack.Env`. Therefore, it shouldn't live in `deploy` package, and should instead be considered an input struct to `stack.Env`'s constructor.

This is a peripheral effort to support #4219.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the Apache 2.0 License.

PR #4271, opened 2022-12-14
chore(stack): pass addons data to template

Lou1415926:
Part of the support for #4219.

Also includes a fix for an integration test failure caused by #4265.

### DIFF
--- a/internal/pkg/deploy/cloudformation/cloudformation_integration_test.go
+++ b/internal/pkg/deploy/cloudformation/cloudformation_integration_test.go
@@ -433,9 +433,9 @@ func Test_Environment_Deployment_Integration(t *testing.T) {
 	}()
 
 	t.Run("Deploys bootstrap resources for the environment to CloudFormation", func(t *testing.T) {
-		bucketARN := "arn:aws:kms:us-west-2:111122223333:key/1234abcd-12ab-34cd-56ef-1234567890ab"
-		environmentToDeploy.ArtifactBucketKeyARN = bucketARN
-		environmentToDeploy.ArtifactBucketARN = fmt.Sprintf("arn:aws:s3:::%s", bucketName)
+		bucketARN := fmt.Sprintf("arn:aws:s3:::%s", bucketName)
+		environmentToDeploy.ArtifactBucketKeyARN = "arn:aws:kms:us-west-2:111122223333:key/1234abcd-12ab-34cd-56ef-1234567890ab"
+		environmentToDeploy.ArtifactBucketARN = bucketARN
 
 		// Deploy the environment and wait for it to be complete
 		require.NoError(t, deployer.CreateAndRenderEnvironment(stack.NewBootstrapEnvStackConfig(&environmentToDeploy), bucketARN))
--- a/internal/pkg/deploy/cloudformation/stack/env.go
+++ b/internal/pkg/deploy/cloudformation/stack/env.go
@@ -74,6 +74,7 @@ type EnvConfig struct {
 	PermissionsBoundary  string                // Optional. An IAM Managed Policy name used as permissions boundary for IAM roles.
 
 	// Runtime configurations.
+	Addons              *NestedStackConfigurer
 	CustomResourcesURLs map[string]string //  Mapping of Custom Resource Function Name to the S3 URL where the function zip file is stored.
 
 	// User inputs.
--- a/internal/pkg/deploy/cloudformation/stack/env.go
+++ b/internal/pkg/deploy/cloudformation/stack/env.go
@@ -133,12 +133,6 @@ func (e *Env) Template() (string, error) {
 		}
 		forceUpdateID = id.String()
 	}
-
-	publicHTTPConfig, err := e.publicHTTPConfig()
-	if err != nil {
-		return "", err
-	}
-
 	vpcConfig, err := e.vpcConfig()
 	if err != nil {
 		return "", err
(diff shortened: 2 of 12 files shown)


internal/pkg/deploy/cloudformation/stack/workload.go
```diff
@@ -101,7 +101,8 @@ func (i ECRImage) GetLocation() string {
 	return fmt.Sprintf("%s:%s", i.RepoURL, "latest")
 }
 
-type addons interface {
+// NestedStackConfigurer configures a nested stack that deploys addons.
+type NestedStackConfigurer interface {
```
Lou1415926, line 105: The reason why I exported this interface is because it is a part of the input for the constructors of `Env`, `LBWS`, `RDWS` etc.

internal/pkg/deploy/cloudformation/cloudformation_integration_test.go
```diff
@@ -433,9 +433,9 @@ func Test_Environment_Deployment_Integration(t *testing.T) {
 	}()
 
 	t.Run("Deploys bootstrap resources for the environment to CloudFormation", func(t *testing.T) {
-		bucketARN := "arn:aws:kms:us-west-2:111122223333:key/1234abcd-12ab-34cd-56ef-1234567890ab"
```
efekarakus, line 436: 🤣

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -74,6 +79,7 @@ type EnvConfig struct {
 	PermissionsBoundary  string                // Optional. An IAM Managed Policy name used as permissions boundary for IAM roles.
 
 	// Runtime configurations.
+	Addons              *addons
 	CustomResourcesURLs map[string]string //  Mapping of Custom Resource Function Name to the S3 URL where the function zip file is stored.
```
efekarakus, line 84: Since `EnvConfig` is the input to the constructor should the `addons` type be public?

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -61,6 +61,11 @@ type envReadParser interface {
 	ParseEnvBootstrap(data *template.EnvOpts, options ...template.ParseOption) (*template.Content, error)
 }
 
+type addons struct {
+	URL    string
+	Config NestedStackConfigurer
```
efekarakus: 1. If clients will populate these fields can we add doc comments 🥺 
2. What do you think of these alternative names?
```suggestion
	S3ObjectURL    string
	Stack NestedStackConfigurer
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -368,20 +368,20 @@ type networkLoadBalancerConfig struct {
 	appDNSName           *string
 }
 
-func convertELBAccessLogsConfig(mft *manifest.Environment) (*template.ELBAccessLogs, error) {
+func convertELBAccessLogsConfig(mft *manifest.Environment) *template.ELBAccessLogs {
 	elbAccessLogsArgs, isELBAccessLogsSet := mft.ELBAccessLogs()
 	if !isELBAccessLogsSet {
-		return nil, nil
```
efekarakus, line 374: :cool-crying:

internal/pkg/deploy/cloudformation/stack/workload.go
efekarakus: nit: What do you think of renaming the the interface to
```suggestion
type StringSerializer interface {
```
Since it seem to be more flexible than just for nested stacks

Lou1415926: I think the intended usage of `Parameters() (string, error)` is pretty specific to nested stacks 🤔  Only nested stacks would need to render for example, the below snippet
```yaml
Extra1: !Ref Somthing
Extra2: !Ref SomethingElse
```
goes [here](https://github.com/aws/copilot-cli/blob/534e4296060a40dd746420154fc891062e6634fa/internal/pkg/template/templates/environment/cf.yml#L659).

For a regular stack, the values of parameters are passed in API calls, not in the plain string template.

---

Instead of `Parameters() (string, error)`, maybe a better name for the function would be `ExtraParameters() (string, error)` - because it returns only the extra parameters:
```yaml
Extra1: !Ref Somthing
Extra2: !Ref SomethingElse
```
Whereas, a function named `Parameters() (string, error)` (or better, [`SerializedParameters() (string, error)`](https://github.com/aws/copilot-cli/blob/821c7086ff66c7c1760c0f5027b4a1a6fe050746/internal/pkg/deploy/cloudformation/stack/worker_svc.go#L180)) should return both the extra parameters and the Copilot-required ones:
```yaml
{
  "Parameters": {
    "App": "!Ref AppName",
    "Env": "!Ref Name",
    "Extra1": "!Ref Something",
    "Extra2": "!Ref SomethingElse",
  },
}
```

The name `StringSerializer` sounds like it should have `Template()(string, error)` and `SerializedParameters() (string, error)`; an interface that contains `Template()(string, error)` and `ExtraParameters() (string, error)` should probably be named specifically after "nested stacks"

iamhopaul123 (approved):
:shipit:

internal/pkg/deploy/cloudformation/stack/workload.go
efekarakus: Gotcha yea, then `ExtraParameters` would have been perfect 😭 If it's not too much work afterwards we can refactor it to that

efekarakus (approved):

PR #4277, opened 2022-12-15
chore(manifest): add cdn static support for existing S3

iamhopaul123:

### DIFF
--- a/internal/pkg/manifest/env.go
+++ b/internal/pkg/manifest/env.go
@@ -216,7 +216,7 @@ type EnvironmentCDNConfig struct {
 type AdvancedCDNConfig struct {
 	Certificate  *string         `yaml:"certificate,omitempty"`
 	TerminateTLS *bool           `yaml:"terminate_tls,omitempty"`
-	Static       CDNStaticConfig `yaml:"static,omitempty"`
+	Static       CDNStaticConfig `yaml:"static_assets,omitempty"`
 }
 
 // IsEmpty returns whether environmentCDNConfig is empty.
--- a/internal/pkg/manifest/env.go
+++ b/internal/pkg/manifest/env.go
@@ -214,8 +214,9 @@ type EnvironmentCDNConfig struct {
 
 // AdvancedCDNConfig represents an advanced configuration for a Content Delivery Network.
 type AdvancedCDNConfig struct {
-	Certificate  *string `yaml:"certificate,omitempty"`
-	TerminateTLS *bool   `yaml:"terminate_tls,omitempty"`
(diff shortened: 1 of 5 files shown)


internal/pkg/manifest/env.go
```diff
@@ -270,7 +271,19 @@ func (cfg *EnvironmentCDNConfig) UnmarshalYAML(value *yaml.Node) error {
 	return nil
 }
 
-// IsEmpty returns true if vpc is not configured.
+// CDNStaticConfig represents the static config for CDN.
+type CDNStaticConfig struct {
+	Location string  `yaml:"location,omitempty"`
```
Lou1415926, line 276: Does this work nicely with environment overrides `mergo`?

Lou1415926 (commented):
LGTM just one questeion!

internal/pkg/manifest/env.go
iamhopaul123: This is an env manifest field. How do we use env overrides 🤔

Lou1415926: :homer-disappearing

Lou1415926 (approved):
:shipit:

internal/pkg/manifest/env.go
```diff
@@ -214,8 +214,9 @@ type EnvironmentCDNConfig struct {
 
 // AdvancedCDNConfig represents an advanced configuration for a Content Delivery Network.
 type AdvancedCDNConfig struct {
-	Certificate  *string `yaml:"certificate,omitempty"`
-	TerminateTLS *bool   `yaml:"terminate_tls,omitempty"`
+	Certificate  *string         `yaml:"certificate,omitempty"`
+	TerminateTLS *bool           `yaml:"terminate_tls,omitempty"`
+	Static       CDNStaticConfig `yaml:"static,omitempty"`
```
efekarakus: What do you think of the following for the key name?
```suggestion
	Static       CDNStaticConfig `yaml:"static_website,omitempty"`
```
It's slightly more verbose but I think more explicit 🤔

efekarakus (approved):
Looks great, approved with do-not-merge to hear your thoughts on the field name

internal/pkg/manifest/env.go
iamhopaul123: It's actually where their static assets go. I wonder if `static_website` can correctly convey the meaning 🤔

efekarakus: That totally makes sense! How about `static_content` or `static_assets` then?

PR #4278, opened 2022-12-15
fix: change invalid property SSLPolicy to SslPolicy in CloudFormation template

### LINKED ISSUE #4269: Specifying http.public.ssl_policy in the environment manifest creates CloudFormation with incorrect key of SSLPolicy.
Specifying the AWS SSL Security policy in the copilot environment manifest will  generate  CloudFormation in which the AWS::ElasticLoadBalancingV2::Listener resource  has a property key of "SSLPolicy" which will generate the following validation failure:

Resource template validation failed for resource HTTPSListener as the template has invalid properties. Please refer to the resource documentation to fix the template. Properties validation failed for resource HTTPSListener with message: #: extraneous key [SSLPolicy] is not permitted

The correct key value  should be "SslPolicy".

The CloudFormation does update the resource as expected, but will  detect drift in the AWS::ElasticLoadBalancingV2::Listener  due to the  validation issue.

KollaAdithya:
This PR resolves #4269


iamhopaul123 (approved):

efekarakus (approved):
ty!!

PR #1108, opened 2020-07-10
fix(storage): add bucket policy to force SSL connections

bvtujo:
This PR adds a bucket policy to the CF generated by `storage init`. It now requires SSL connections in order to perform any action on a bucket. 

Related #769 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4279, opened 2022-12-15
chore(templates): create CFN template for environment-level s3 storage option

Lou1415926:
With support of #4219, it is possible now to create environment-level storage options. This is the s3 storage option.


Lou1415926:
View https://github.com/aws/copilot-cli/pull/4279/commits/c921dfb9ad9b86ddb2f00a5fa709b9ba73cfc161 to see the difference between templates of the environment-level version and the workload-level version.

internal/pkg/template/templates/addons/s3/env/cf.yml
```diff
@@ -0,0 +1,54 @@
+Parameters:
+  App:
+    Type: String
+    Description: Your application's name.
+  Env:
+    Type: String
+    Description: The name of the environment being deployed.
```
iamhopaul123, line 7: Maybe just
```suggestion
    Description: Your environment's name.
```
🤔

internal/pkg/template/templates/addons/s3/env/cf.yml
```diff
@@ -0,0 +1,54 @@
...
+
+Resources:
+  {{logicalIDSafe .Name}}Bucket:
+    Metadata:
+      'aws:copilot:description': 'An Amazon S3 bucket to store and retrieve objects for {{.Name}}'
```
iamhopaul123:
```suggestion
      'aws:copilot:description': 'An Amazon S3 bucket to store and retrieve objects'
```
It doesn't feel `for {{.Name}}` needs to be there.

internal/pkg/template/templates/addons/s3/env/cf.yml
```diff
@@ -0,0 +1,54 @@
...
+            SSEAlgorithm: AES256
+      PublicAccessBlockConfiguration:
+        BlockPublicAcls: true
+        BlockPublicPolicy: true
+
+  {{logicalIDSafe .Name}}BucketPolicy:
+    Metadata:
+      'aws:copilot:description': 'A bucket policy to deny unencrypted access to the bucket and its contents'
+    Type: AWS::S3::BucketPolicy
+    DeletionPolicy: Retain
```
iamhopaul123, line 31: why do we retain the bucket policy?

Lou1415926: good question!!! It was initially added by [this commit](https://github.com/aws/copilot-cli/pull/1108/files#diff-f7921c17445c4b5d5487a2de5974851ba2a710e56f491f66d030c01ffc86877fR24) for workload addons.

My guess is that since this bucket policy has a dependency on the bucket, it will get deleted before the bucket. This creates a window during which the bucket exists while the SSL connection is not enforced (because the policy has been deleted). To prevent this from happening, we add the retain policy.

I can double check with @bvtujo ! But this is my guess!

internal/pkg/template/templates/addons/s3/env/cf.yml
Lou1415926: I am debating internally, probably to an unnecessary extent 😅  I wanted to frame this description so that it implies that people can write, for example, `Mappings` depending on the value of `Env`. "Your environment's name" makes it sound like the env addon template is going to be deployed with one specific env; while in reality, it will get deployed with any envs... I might be overthinking this way too much. But do you think this concern make sense?

internal/pkg/template/templates/addons/s3/env/cf.yml
Lou1415926: I wanted to keep `{{.Name}}` so that when people look at the progress tracker they see the name that they put in for the storage! Also, if people create multiple s3 buckets as addons, it'd be helpful for them to differentiate. I rephrased the description a little bit with `{{.Name}}` kept!

internal/pkg/template/templates/addons/s3/env/cf.yml
iamhopaul123: > "Your environment's name" makes it sound like the env addon template is going to be deployed with one specific env; while in reality, it will get deployed with any envs

That sounds like too much thinking...Workload addons is the same but we won't think it is for a specific workload (some wkld addons can be agnostic to wkld itself).

internal/pkg/template/templates/addons/s3/env/cf.yml
iamhopaul123: Let's say the bucket name is "myBucket". `'An Amazon S3 bucket to store and retrieve objects for myBucket'` doesn't make sense to me since "myBucket" is not technically a pattern name but a bucket name. When doing `storage init` we prompt them to input the name we ask for the bucket name as well.

I feel like this sounds more like a general question about showing the logical resource name in progress tracker instead of putting the name into the description. Or maybe rephrase it to `"An Amazon S3 bucket named {{.Name}} to store and retrieve objects"`? TBH "to store and retrieve objects" sounds redundant too...

internal/pkg/template/templates/addons/s3/env/cf.yml
Lou1415926: In the workload version, the description to the `Name` parameter is actually `Description: The name of the service, job, or workflow being deployed.` instead of "Your workload's name". Although I think it makes sense in this case to describe it as "Your workload's name" because a workload addon is indeed associated with a specific workload in most cases. But for environment addons, it is different. 

Anyway, I can change it to "Your environment's name".

internal/pkg/template/templates/addons/s3/env/cf.yml
iamhopaul123: Gotcha. It's ok they can update the template themselves and we just promote the best practice.

internal/pkg/template/templates/addons/s3/env/cf.yml
iamhopaul123: nvm let's keep it as is for now. I don't think it is a big one.

Lou1415926: I will keep it as is then

internal/pkg/template/templates/addons/s3/env/cf.yml
huanjani: I like it as is. 🤷🏼‍♀️

internal/pkg/template/templates/addons/s3/env/cf.yml
```diff
@@ -9,7 +9,7 @@ Parameters:
 Resources:
   {{logicalIDSafe .Name}}Bucket:
     Metadata:
-      'aws:copilot:description': 'An Amazon S3 bucket to store and retrieve objects for {{.Name}}'
+      'aws:copilot:description': 'An Amazon S3 bucket as your {{.Name}} storage to store and retrieve objects'
```
huanjani: I'm curious about this change. I thought it made more sense before the rephrasing.

huanjani: Oh, just saw the background. Maybe: 
```suggestion
      'aws:copilot:description': 'An Amazon S3 bucket, {{.Name}}, for storing and retrieving objects'
```

internal/pkg/template/templates/addons/s3/env/cf.yml
Lou1415926: I will keep it as is!

internal/pkg/template/templates/addons/s3/env/cf.yml
```diff
@@ -0,0 +1,53 @@
...
+  Env:
+    Type: String
+    Description: The name of the environment being deployed.
+
+Resources:
+  {{logicalIDSafe .Name}}Bucket:
+    Metadata:
+      'aws:copilot:description': 'An Amazon S3 bucket, {{.Name}}, for storing and retrieving objects'
+    Type: AWS::S3::Bucket
+    Properties:
```
efekarakus, line 14: Can we add this to our S3 bucket templates too 🥺 
https://github.com/aws/copilot-cli/blob/8d4f503dfb14a7e5985831aa191fc93175fbe6b7/internal/pkg/template/templates/app/cf.yml#L92-L94

It's okay if the service addon one is done separately

internal/pkg/template/templates/addons/s3/env/cf.yml
efekarakus: Wouldn't `DependsOn` instead handle this problem rather than `DeletionPolicy: Retain`? 🥺

efekarakus (commented):
LGTM! Just two 🥺 requests

internal/pkg/template/templates/addons/s3/env/cf.yml
Lou1415926: Do you mean like this?
```yaml
{{logicalIDSafe .Name}}Bucket:
    Metadata:
      'aws:copilot:description': 'An Amazon S3 bucket, {{.Name}}, for storing and retrieving objects'
    Type: AWS::S3::Bucket
    DependsOn: {{logicalIDSafe .Name}}BucketPolicy
```

That would introduce circular dependency! BucketPolicy has `!Sub ${ {{logicalIDSafe .Name}}Bucket.Arn}/*`

internal/pkg/template/templates/addons/s3/env/cf.yml
Lou1415926: Added this to both env addon and workload addon!

efekarakus (approved):

internal/pkg/template/templates/addons/s3/env/cf.yml
efekarakus: nono i meant the other way around:
```
 {{logicalIDSafe .Name}}BucketPolicy:
    Metadata:
      'aws:copilot:description': 'A bucket policy to deny unencrypted access to the bucket and its contents'
    Type: AWS::S3::BucketPolicy
    DependsOn: {{logicalIDSafe .Name}}Bucket
```

Wouldn't that solve the concern in the above comment?

(edit: I'm dumb - this won't work the policy will get deleted first 🤦 )

iamhopaul123 (approved):

PR #4280, opened 2022-12-16
chore(cli): deploy environment addons

Lou1415926:
In support of #4219 


internal/pkg/cli/deploy/env.go
```diff
@@ -142,24 +165,78 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+func (d *envDeployer) AddonsTemplate() (string, error) {
+	addons, err := d.parseAddons()
+	if err != nil {
+		var notFoundErr *addon.ErrAddonsNotFound
+		if !errors.As(err, &notFoundErr) {
+			return "", fmt.Errorf("parse environment addons: %w", err)
+		}
+		return "", nil
+	}
+	return addons.Template()
```
iamhopaul123: Do we need to wrap the error here?

iamhopaul123 (approved):
Awesome! LGTM overall

internal/pkg/cli/deploy/env.go
```diff
@@ -142,24 +165,82 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
-	}, nil
+
+		ws: in.Workspace,
+	}
+	deployer.parseAddons = func() (stackBuilder, error) {
+		if !in.EnvAddonsFeatureFlag {
+			return nil, &addon.ErrAddonsNotFound{}
+		}
+		deployer.parseAddonsOnce.Do(func() {
+			deployer.addons.stack, deployer.addons.err = addon.ParseFromEnv(deployer.ws)
```
efekarakus, line 176: do we need to wrap the error message or is it clear enough?

internal/pkg/cli/deploy/env.go
```diff
@@ -142,24 +165,82 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
...
+	}
+	return deployer, nil
 }
 
 // Validate returns an error if the environment manifest is incompatible with services and application configurations.
 func (d *envDeployer) Validate(mft *manifest.Environment) error {
 	return d.validateCDN(mft)
 }
 
+// UploadEnvArtifactsOutput is the output of UploadArtifacts.
```
efekarakus: nit: 😂 maybe to be a bit less redundant:
```suggestion
// UploadEnvArtifactsOutput holds URLs of artifacts pushed to S3 buckets, ECR repositories...
```

internal/pkg/cli/env_deploy.go
```diff
@@ -138,13 +139,14 @@ func (o *deployEnvOpts) Execute() error {
 	if err := deployer.Validate(mft); err != nil {
 		return err
 	}
-	urls, err := deployer.UploadArtifacts()
+	uploadArtifactsOut, err := deployer.UploadArtifacts()
```
efekarakus: nit:
```suggestion
	artifacts, err := deployer.UploadArtifacts()
```

internal/pkg/cli/env_package.go
```diff
@@ -152,23 +156,25 @@ func (o *packageEnvOpts) Execute() error {
...
 	}
-	if err := deployer.Validate(mft); err != nil {
+	if err := packager.Validate(mft); err != nil {
 		return err
 	}
-	var urls map[string]string
+	var uploadArtifactsOut deploy.UploadEnvArtifactsOutput
 	if o.uploadAssets {
-		urls, err = deployer.UploadArtifacts()
+		out, err := packager.UploadArtifacts()
```
efekarakus, line 168: Let me know if this train of thought makes sense to you too:
1. Since all the hashes and the URL of these artifacts are deterministic, I wonder if this signature makes more sense:
   ```
   err := packager.UploadArtifacts()
   ```
2. Instead `GenerateCloudFormationTemplate` holds the responsibility of `parseAddons` and calling `artifactpath.EnvironmentAddons` at all times so that the template output is always the same regardless of using the `--upload-assets` flag

efekarakus: It's okay if this is not done part of this PR as it's larger change, let me know what you think!

The extensions project would need to make the change above anyway, so I feel like it'd be nice to introduce it

efekarakus (approved):
Looks good!

internal/pkg/cli/env_package.go
Lou1415926: SGTM! I should do it in a separate PR so that it's clearer to review!

Issue #4270, opened 2022-12-14
copilot app init --domain followed by copilot init fails due to HTTPSCert


lexicalunit:
I ran the following commands:

```shell
$ copilot app init --domain aws.spellbot.io
$ copilot init

Use existing application: No
Workload type: Backend Service
Service name: spellbot
Dockerfile: ./Dockerfile.bot
Deploy: Yes
```

And encountered the following errors. Everything else seemed to work except HTTPSCert.

```
The following resource(s) failed to create: [HTTPSCert]

- A Route 53 Hosted Zone for the environment's subdomain:

    The specified hosted zone contains non-required resource record sets
    and so cannot be deleted. (Service: Route53, Status Code: 400, Request
    ID: a28557c2-1a40-458a-a49e-27b84bb702c6, Extended Request ID: null)
    The specified hosted zone contains non-required resource record sets
    and so cannot be deleted. (Service: Route53, Status Code: 400, Request
    ID: 586c74d4-e434-42ce-9a67-e13c24fd4a1a, Extended Request ID: null)

- Request and validate an ACM certificate for your domain:

    Received response status [FAILED] from custom resource. Message return
    ed: Resource is not in the state certificateValidated (Log: /aws/lambd
    a/spellbot-test-CertificateValidationFunction-RYnN39hsF5fa/2022/12/14/
    [$LATEST]1d2aba50501a4172a4c4c1f0518ca852) (RequestId: f03478d7-1e1b-4
    164-a6ab-4656db5ab632)
    Received response status [FAILED] from custom resource. Message return
    ed: Cannot read properties of undefined (reading 'Name') (Log: /aws/la
    mbda/spellbot-test-CertificateValidationFunction-RYnN39hsF5fa/2022/12/
    14/[$LATEST]1d2aba50501a4172a4c4c1f0518ca852) (RequestId: 4f475642-c54
    4-4b6f-b466-98691b1a0fc0)
```

After hanging for over 20 minutes the process started emitting a line like the following 10 times a second until I killed it.
```log
- Creating the infrastructure for the spellbot-test environment.                     [update rollback complete cleanup in progress]  [1159.9s]
```

Note that I have 3 hosted zones, and one of them is my `aws.spellbot.io` HZ that I created manually before running the `copilot` commands. I then set up NS records for the Name servers created by that HZ in my DNS management tool of my personal website.

Should I have not done that? Is it confused because I manually set up the HZ myself?

lexicalunit:
[...]

PR #4281, opened 2022-12-16
chore: lookup NS records of domain to validate route53 as DNS service

### LINKED ISSUE #4172: Trouble with domain added with manual name servers
Hey Copilot team,

I am trying to create an app with the --domain flag. This is a domain that was not registered through Route 53. I followed the guides to add the correct name servers to a hosted zone in Route53. However, I still get the below error message:

```
Note: The account does not seem to own the domain that you entered.
Please make sure that MY_DOMAIN_NAME is registered with Route53 in your account, or that your hosted zone has the appropriate NS records.
To transfer domain registration in Route53, see:
https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/domain-transfer-to-route-53.html
To update the NS records in your hosted zone, see:
https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/SOA-NSrecords.html#NSrecords
```

I tried the links in the error message, but they are the same steps I followed to create the hosted zone in the first place. Any help to point me in the right direction would be much appreciated.

Thanks!

efekarakus:
Previously, we would log a warning when a user bought a domain outside of Route53 but correctly migrated to using Route53 as the DNS service. This is totally fine and we shouldn't scare users from using Copilot that way :) 

We now instead lookup NS records and ensure they match the hosted zone in the account. All of this functionality is optional and won't stop the CLI.

Resolves #4172, related #4270


internal/pkg/aws/route53/errors.go
```diff
@@ -5,22 +5,47 @@ package route53
...
+	r53Records   []string
+	dnsRecords   []string
+}
+
+func (err *ErrUnmatchedNSRecords) Error() string {
+	return fmt.Sprintf("name server records for %q do not match records for hosted zone %q", err.domainName, err.hostedZoneID)
+}
+
+// RecommendActions implements the main.actionRecommender interface.
+func (err *ErrUnmatchedNSRecords) RecommendActions() string {
```
huanjani, line 44: really nice messaging! ❤️

internal/pkg/aws/route53/errors.go
```diff
@@ -5,22 +5,47 @@ package route53
...
+	return fmt.Sprintf(`Domain name %q has the following name server records: %s
+Whereas the hosted zone ID %q for the domain has: %s
+Copilot will proceed but to use Route 53 as the DNS service
```
huanjani: nit:
```suggestion
Copilot will proceed, but to use Route 53 as the DNS service, 
```

internal/pkg/aws/route53/route53.go
```diff
@@ -20,47 +23,130 @@ const (
...
+	if err != nil {
+		return err
+	}
+
+	actual, err := r53.lookupNSRecords(domainName)
+	if err != nil {
+		return err
+	}
+
+	if !eqStringSlices(wanted, actual) {
```
Lou1415926: Can `wanted` be a subset of `actual` instead of a equal set?

internal/pkg/aws/route53/route53.go
```diff
@@ -20,47 +23,130 @@ const (
...
+type nameserverResolver interface {
+	LookupNS(ctx context.Context, name string) ([]*net.NS, error)
 }
 
 // Route53 wraps an Route53 client.
 type Route53 struct {
 	client api
+	dns    nameserverResolver
+
+	hostedZoneIDFor map[string]string
```
huanjani, line 39: This is an interesting naming convention. What situations does `...For` apply to?

Lou1415926 (commented):
LGTM just one question!

internal/pkg/aws/route53/route53.go
```diff
@@ -79,3 +165,28 @@ func matchesDomain(domain string) filterZoneFunc {
...
 }
+
+func cleanNSRecord(record string) string {
+	if !strings.HasSuffix(record, ".") {
+		return record
+	}
+	return record[:len(record)-1]
+}
+
+func eqStringSlices(first, second []string) bool {
```
huanjani: This function seems useful beyond this application-- I wonder if we should move it to somewhere more general?

internal/pkg/aws/route53/route53.go
efekarakus: Huh great question that I haven't thought of 👍 

My understanding is that they shouldn't do this: https://serverfault.com/a/796311 if they want `example.com` to be managed by R53 they should replace all of its NS records. I think you're right that this is OK if they write a subset (with a minimum of 2 records) as long as the hosts contain the same info.

On the other hand, if the hosts contain different information (this person says [it'd be terrible if the content](https://qr.ae/prNPbH) is different) then I don't get the behavior 👻 .

efekarakus: OK I updated so that actual must be a strict subset of wanted (with no other records besides the ones in actual)

internal/pkg/aws/route53/route53.go
efekarakus: I'm struggling to find a link to the original writeup 😭 but this is a naming convention I stole from Martin Fowler and Kent Beck I think.

They use `xxxFor[name]` or `yyyOn[date]` style naming for variables that represent `maps`

internal/pkg/aws/route53/route53.go
efekarakus: Maybe someday these generic functions will get promoted to the stdlib: https://github.com/golang/exp/blob/master/slices/slices.go

I think today it's better to just do a little duplication for these basic functions rather than having a "utils" style package for slices/sets

internal/pkg/aws/route53/errors.go
```diff
@@ -5,22 +5,47 @@ package route53
...
+}
+
+// RecommendActions implements the main.actionRecommender interface.
+func (err *ErrUnmatchedNSRecords) RecommendActions() string {
+	return fmt.Sprintf(`Domain name %q has the following name server records: %s
+Whereas the hosted zone ID %q for the domain has: %s
+Copilot will proceed, but to use Route 53 as the DNS service, 
+please ensure the name server records are mapped correctly:
+- https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/migrate-dns-domain-in-use.html#migrate-dns-change-name-servers-with-provider`,
+		err.domainName, strings.Join(err.dnsRecords, ", "),
```
Lou1415926: what about `english.WordSeries` 🪄

internal/pkg/aws/route53/route53.go
```diff
@@ -79,3 +165,28 @@ func matchesDomain(domain string) filterZoneFunc {
...
+	}
+	return record[:len(record)-1]
+}
+
+func isStrictSubset(subset, superset []string) bool {
+	if len(subset) > len(superset) {
+		return false
+	}
+
+	exists := make(map[string]struct{}, len(superset))
```
Lou1415926: nit: what do you think of naming it to something like this:
```suggestion
	isInSuperset := make(map[string]struct{}, len(superset))
```

internal/pkg/aws/route53/route53.go
Lou1415926: Makes sense to me! We want the `actual` to contain only NS records from `wanted`, so that all the NS points to the R53 servers.

Lou1415926 (approved):
🚢 :shipit:

internal/pkg/aws/route53/errors.go
efekarakus: Done!

internal/pkg/aws/route53/route53.go
efekarakus: Renamed to `isMember`

internal/pkg/aws/route53/route53.go
```diff
@@ -173,18 +173,18 @@ func cleanNSRecord(record string) string {
 	return record[:len(record)-1]
 }
 
-func eqStringSlices(first, second []string) bool {
-	if len(first) != len(second) {
+func isStrictSubset(subset, superset []string) bool {
```
huanjani, line 176: ❤️

huanjani (approved):
🚢

PR #4282, opened 2022-12-16
chore: cli stack

dannyrandall:
This (incomplete) PR refactors the `cli` and `cli/deploy` packages to make the `addon` and `main` stacks visible to the `cli` package. The idea is to combine similar functionality between `package` and `deploy` commands to follow the same pattern of generate stack -> do something with it (deploy or print). This makes it so that we can add steps _after_ generating the stack but _before_ doing the command task, like printing a diff or overriding a template.

### DIFF
--- a/internal/pkg/addon/addons.go
+++ b/internal/pkg/addon/addons.go
@@ -51,6 +51,10 @@ type WorkloadStack struct {
 	workloadName string
 }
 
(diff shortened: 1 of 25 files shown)


efekarakus:
I'll close the PR and steal commits individually 😈 as needed for packaging templates. Thanks Danny!

PR #4289, opened 2022-12-19
chore(templates): create CFN template for environment-level ddb storage option

Lou1415926:
With support of https://github.com/aws/copilot-cli/issues/4219, it is possible now to create environment-level storage options. This is the ddb storage option.


Lou1415926:
View https://github.com/aws/copilot-cli/pull/4289/commits/094c81d804eaf26c9faaf06863685f850176d151 to see the difference between templates of the environment-level version and the workload-level version.

internal/pkg/template/templates/addons/ddb/env/cf.yml
```diff
@@ -0,0 +1,45 @@
...
+              KeyType: HASH
+            - AttributeName: {{.SortKey}}
+              KeyType: RANGE
+          Projection:
+            ProjectionType: ALL{{end}}{{end}}
+
+Outputs:
+  {{envVarName .Name}}:
+    Description: "The name of this DynamoDB."
+    Value: !Ref {{logicalIDSafe .Name}}
```
iamhopaul123: Noticed this is an existing one for wkld addons but I am not sure why we want to output this...how do we expect users to use it 🤔

Lou1415926: In their application code, people need the name of the table in order to connect. For example, with [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Table), you may write
```python
user_table = dynamo_client.Table(os.environ.get('USER_TABLE_NAME'))
```

With env-level ddb, you'd 
1. Export the ddb table name
2. Write in service manifest:
```yaml
variables:
   USER_TABLE_NAME:
      from_cfn: UserDDBTableName
```

Maybe I can add the commented `Export` block for the DDB name output as well.

iamhopaul123: Maybe we should export `TableName: !Sub ${App}-${Env}-{{.Name}}` which is the real table name? Or export "AWS::DynamoDB::Table" which is https://github.com/aws/copilot-cli/pull/4289/files#diff-f3cbab28817cc024d734b1c7fddb2144cf94e7c5af19eb69549833492e309d64R39-R44

Lou1415926: I'm not sure I follow! 

The current implementation uses `!Ref LogicalID` which returns the real table name: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dynamodb-table.html#aws-resource-dynamodb-table-return-values. Can you clarify the difference between this and your suggestion?

internal/pkg/template/templates/addons/ddb/env/cf.yml
```diff
@@ -0,0 +1,48 @@
...
+#    # Uncomment the "Export" block if you need to inject the table name as an environment variable in a workload. 
+#    Export: 
+#      Name: !Sub ${Env}-{{logicalIDSafe .Name}}TableName
+  {{logicalIDSafe .Name}}DynamoDBTableARN:
+    Description: "The ARN of the {{.Name}} DynamoDB table."
+    Value: !GetAtt {{logicalIDSafe .Name}}.Arn
+#    # Uncomment the "Export" block if you need to reference the DynamoDB table ARN in a workload addon.
+#    Export: 
+#      Name: !Sub ${Env}-{{logicalIDSafe .Name}}ARN
```
efekarakus, line 47: nit: I wonder if for env addons we should just leave this uncommented

efekarakus (approved):

internal/pkg/template/templates/addons/ddb/env/cf.yml
Lou1415926: Good point! 

There are two benefits that I had in mind for having this commented.

The first is to familiarize users with the "export + `from_cfn`" procedure. Instead of having the template connectable to workloads out-of-the-box, this provides a hands-on guidance. 

The second benefit is that, this (hopefully) incentivizes users to look at what an `Export` is, and signals that `Export` is only needed if the value will be used in a workload. Since this also serves as an example of how users could write an environment addon, I'm worried that users would just `Export` every single `Output` in any env addons that they write. 

The benefit of leaving this uncommented is obvious - it saves users some extra steps 👍🏼  I think maybe we can take a middle ground here:
```yaml
  {{envVarName .Name}}:
    Description: "The name of this DynamoDB table."
    Value: !Ref {{logicalIDSafe .Name}}
    # NOTE: 
    # You can inject the table name as an environment variable into a workload:
    #  ```
    # variables:
    #    {{logicalIDSafe .Name}}_TABLE_NAME: 
    #        from_cfn: ${COPILOT_ENVIRONMENT_NAME}-{{logicalIDSafe .Name}}TableName
    # ```
    # Remove the "Export" if none of your workload needs this as an environment variable.
    Export:    
      Name: !Sub ${Env}-{{logicalIDSafe .Name}}TableName
  {{logicalIDSafe .Name}}DynamoDBTableARN:
    Description: "The ARN of the {{.Name}} DynamoDB table."
    Value: !GetAtt {{logicalIDSafe .Name}}.Arn
    # NOTE: ...similar but with an example to workload addon
    Export: 
      Name: !Sub ${Env}-{{logicalIDSafe .Name}}ARN
```

internal/pkg/template/templates/addons/ddb/env/cf.yml
iamhopaul123: :homer-disappearing:

iamhopaul123 (approved):

PR #4294, opened 2022-12-20
chore: add alarm creation for ABR

huanjani:
Related: #4267.

This PR adds the `advanced` format of `rollback_alarms`, `AlarmArgs`.
With these changes, users may create alarms for `cpu_utilization` and/or `memory_utilization` metrics and the deployment config will refer to them by name.

Please pay special attention to/give feedback about the alarm defaults related to how often and for how long the alarm should be evaluated. 

The Deployments team has a built-in "bake time" for polling the alarm after successful deployment that is based on the evaluation time of the alarm(s). (With multiple different times, the longest is used.)

Screen Shot 2022-12-20 at 3 52 56 PM

Next up: adding an alarm option for worker services.


internal/pkg/deploy/cloudformation/stack/transformers_test.go
```diff
@@ -2091,7 +2091,22 @@ func Test_convertDeploymentConfig(t *testing.T) {
 			out: template.DeploymentConfigurationOpts{
 				MinHealthyPercent: minHealthyPercentDefault,
 				MaxPercent:        maxPercentDefault,
-				RollbackAlarms:    []string{"alarmName1", "alarmName2"},
+				RollbackAlarms:    "[alarmName1, alarmName2]",
```
paragbhingre: Why did we change `RollbackAlarms` to a string from the slice?

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
```diff
@@ -21,6 +21,17 @@ DeploymentConfiguration:
     AlarmNames: {{ .DeploymentConfiguration.RollbackAlarms }}
     Enable: true
     Rollback: true
+  {{- else if or .DeploymentConfiguration.CPUUtilization .DeploymentConfiguration.MemoryUtilization }}
+  Alarms:
+    AlarmNames:
```
paragbhingre, line 24: I can see only `AlarmNames` is changing in both the scenarios i.e. basic and advanced. So can we club them here to reduce the code duplicity?

```
  {{- if or .DeploymentConfiguration.RollbackAlarms .DeploymentConfiguration.CPUUtilization .DeploymentConfiguration.MemoryUtilization }}
  Alarms:
    AlarmNames: {{- if .DeploymentConfiguration.RollbackAlarms }}{{ .DeploymentConfiguration.RollbackAlarms }}
    {{ - else if .DeploymentConfiguration.CPUUtilization }}
      - !Sub ${AWS::StackName}-CPUAlarm
    {{- else }}
      - !Sub ${AWS::StackName}-MemAlarm
    {{- end }}
    Enable: true
    Rollback: true
  {{- end }}
```

internal/pkg/template/templates/workloads/partials/cf/rollback-alarms.yml
```diff
@@ -0,0 +1,45 @@
+{{- if .DeploymentConfiguration.CPUUtilization}}
+CPURollbackAlarm:
+  Metadata:
+    'aws:copilot:description': "A CloudWatch alarm associated with CPU utilization for deployment rollbacks"
+  Type: AWS::CloudWatch::Alarm
+  DependsOn: Service
+  Properties:
+    AlarmDescription: "Roll back ECS service if CPU utilization is greater or equal to {{.DeploymentConfiguration.CPUUtilization}}% twice in 90 seconds."
```
paragbhingre: Is 90 secs default value given by the deployments team?

paragbhingre (commented):
Just a question and a suggestion.

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
paragbhingre: We can also write a `DeploymentRollbackAlarmsEnabled` function for the `{{- if or .DeploymentConfiguration.RollbackAlarms .DeploymentConfiguration.CPUUtilization .DeploymentConfiguration.MemoryUtilization }}`

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
```diff
@@ -21,6 +21,17 @@ DeploymentConfiguration:
...
+      {{- if .DeploymentConfiguration.CPUUtilization }}
+      - !Sub ${AWS::StackName}-CPUAlarm
```
Lou1415926: Will CFN support something like `!GetAtt CPURollbackAlarm.Name`? With `!Sub ${AWS::StackName}-CPUAlarm`, we will have to keep in mind that this needs to match the actual alarm name written in `rollback-alarms` - which is probably not ideal in the long run.

internal/pkg/template/templates/workloads/partials/cf/rollback-alarms.yml
```diff
@@ -0,0 +1,45 @@
+{{- if .DeploymentConfiguration.CPUUtilization}}
+CPURollbackAlarm:
+  Metadata:
+    'aws:copilot:description': "A CloudWatch alarm associated with CPU utilization for deployment rollbacks"
+  Type: AWS::CloudWatch::Alarm
+  DependsOn: Service
```
Lou1415926: Is this because of `!Ref Service`?

iirc with the `!Ref Service`, CFN should be able to work out the dependency graph already - i.e. this alarm will be created *after* service. So we probably don't need `DependsOn: Service` in this case!

Lou1415926: However, this makes me wonder - if the `Service` creation/update is already completed, does that mean the deployment already stabilized before the  creation of the alarm? Should the alarm have been created.updated before the `Service` instead?

internal/pkg/deploy/cloudformation/stack/transformers_test.go
huanjani: Good question-- for ease of use in the CFN template. `FmtSliceFunc` adds the brackets and commas.

internal/pkg/template/templates/workloads/partials/cf/rollback-alarms.yml
huanjani: It's a very funny thing-- the service has to exist before the alarm can be created; but unless we want users to have to deploy once, then add the alarm name, then deploy again, we need to make sure the service is created before the alarm, and assign the Copilot-created name in the `deployments` field. 

The alarm names can exist/ABR can be enabled during service creation, but ABR is only applied during `UpdateService` because you can only roll back to a previous version, so ABR doesn't apply to the first deployment.

EDS (ECS Deployments Service) doesn't validate that the given alarm exists in an account (though I'm not sure how the bake time is calculated in this case), it just tries to poll for a non-existent alarm. So in theory the service can exist before the alarm, but the alarm still can't exist before the service, hence the `DependsOn`. I plan to add a validation of any passed-in `AlarmNames` in the interest of failing faster.

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
huanjani: It appears the only available attribute for that resource is `ARN`, unfortunately.

internal/pkg/template/templates/workloads/partials/cf/rollback-alarms.yml
Lou1415926: Interesting. Just validating my understanding. If user changes `cpu_utilization` as well as some application code, then the `CPUUtilizationAlarm` will actually get updated *after* the service, meaning that the alarm will not be used in this deployment, not until the next deployment. Is that correct?

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
Lou1415926: Yeah also it'd introduce a circular dependency between the `CPUUtilizationAlarm` and `Service` : (

internal/pkg/deploy/cloudformation/stack/transformers_test.go
huanjani: I could also switch it back and use `range` in the CFN template.

Lou1415926: We can probably keep it as `[]string` and call `fmtSliceFunc .Alarms` from within the CFN template like this https://github.com/aws/copilot-cli/blob/3f458da01ce1cb412530c84508af9b4eda1e1c2a/internal/pkg/template/templates/workloads/partials/cf/https-listener.yml#L109

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -813,9 +813,12 @@ func convertDeploymentConfig(deploymentConfig manifest.DeploymentConfiguration)
 		deployConfigs.MaxPercent = maxPercentDefault
 	}
 	if deploymentConfig.RollbackAlarms.IsBasic() {
-		deployConfigs.RollbackAlarms = deploymentConfig.RollbackAlarms.Basic
+		deployConfigs.RollbackAlarms = template.FmtSliceFunc(deploymentConfig.RollbackAlarms.Basic)
```
efekarakus: nit: what do you think of renaming the variables like 👇 since the manifest and template variables are so similar to each other
```go
func convertDeploymentConfig(in manifest.DeploymentConfiguration) template.DeploymentConfigurationOpts {
   var out template.DeploymentConfigurationOpts
   // same stuff..
}
```

internal/pkg/manifest/validate.go
```diff
@@ -124,7 +124,6 @@ func (d DeploymentConfiguration) validate() error {
 }
 
 func (a AlarmArgs) validate() error {
-	// TODO(jwh)
```
efekarakus, line 135: 👏 👏 👏

internal/pkg/template/workload.go
```diff
@@ -452,13 +453,15 @@ type ObservabilityOpts struct {
...
 type DeploymentConfigurationOpts struct {
 	// The lower limit on the number of tasks that should be running during a service deployment or when a container instance is draining.
 	MinHealthyPercent int
 	// The upper limit on the number of tasks that should be running during a service deployment or when a container instance is draining.
-	MaxPercent     int
-	RollbackAlarms []string
+	MaxPercent        int
+	RollbackAlarms    string
+	CPUUtilization    int
+	MemoryUtilization int
```
efekarakus: What do you think of moving these three fields into its own data type?
```go
type DeploymentConfig struct {
   MinHealthyPercent   int
   MaxHealthyPercent  int
   
   Rollback   RollingUpdateRollbackConfig
}

type RollingUpdateRollackConfig struct {
   AlarmARNs []string
   
   // Custom alarms to create.
   AvgCPUUtilization int
   AvgMemUtilization int
}

// HasCustomAlarms returns true if the client has configured to use Copilot generated CW alarms.
func (cfg RollingUpdateRollackConfig) HasCustomAlarms() bool 
```

internal/pkg/template/templates/workloads/partials/cf/rollback-alarms.yml
```diff
@@ -0,0 +1,45 @@
...
+  Properties:
+    AlarmDescription: "Roll back ECS service if CPU utilization is greater or equal to {{.DeploymentConfiguration.CPUUtilization}}% twice in 3 minutes."
+    AlarmName: !Sub ${AWS::StackName}-CPUAlarm
```
efekarakus: I just want to be as specific as possible since the customer can add its own custom alarms and we don't want to clash with names: (same for the others)
```suggestion
    AlarmName: !Sub ${AWS::StackName}-CopilotRollbackCPUAlarm
```

efekarakus: Also I wonder if we should do some truncating here given that alarms have a limit of 255 chars 🤔 

Something like this (I'm making this up, feel free to do something else):
```suggestion
    AlarmName: {{.DeploymentConfiguration.AlarmName .App .Env .Name "CopilotRollbackCPUAlarm"}}
```

```go
func (cfg DeploymentConfig) AlarmName(app, env, name, suffix string) string {
   maxChunk := (255 - len(suffix) - 3 /* number of dashes */) / 3
   return fmt.Sprintf("%s-%s-%s-%s", app[:maxChunk], env[:maxChunk], name[:maxChunk], suffix)
}
```

internal/pkg/template/templates/workloads/partials/cf/rollback-alarms.yml
huanjani: I tested it and it was used in the current deployment. There's enough time (the deployment becomes long with ABR enabled) for the alarm to get created/updated.
"The bake time is a period of time after the new service version has reached steady state, during which Amazon ECS continues to monitor the alarm associated with the deployment." from https://aws.amazon.com/about-aws/whats-new/2022/12/amazon-ecs-cloudwatch-alarms-safety-deployments/ and for more, https://aws.amazon.com/blogs/containers/automate-rollbacks-for-amazon-ecs-rolling-deployments-with-cloudwatch-alarms/

What isn't great is that not a lot of info is surfaced by the progress tracker so it just seems to be taking forever/stalling and when a rollback happens, it's not obvious. I'll address that in a coming PR.

efekarakus (commented):
Looks great!

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
huanjani: Thanks! I did a version of this that isn't XOR and while has an extra line for the sake of readability.

internal/pkg/template/templates/workloads/partials/cf/rollback-alarms.yml
huanjani: Yeah, good call!
Is it too late to do something similar [here](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/environment/partials/ar-vpc-connector.yml#L7) and [here](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/environment/partials/ar-vpc-connector.yml#L11) so it's not the end that gets lopped off?

efekarakus: If the updates don't require replacement it'd be great!

internal/pkg/template/templates/workloads/partials/cf/rollback-alarms.yml
huanjani: No, the deployments team has the bake time adjusting according to alarms. We need to make the evaluation time long enough to get proper data, but not so long that deployments always take forever. It's hard to know how to hit the balance. ECS emits data to CloudWatch each minute, so I changed this from 90 seconds.

internal/pkg/template/templates/workloads/partials/cf/rollback-alarms.yml
huanjani: Actually, upon further testing, you're right, @Lou1415926. I watched the multiple consoles and the alarm values didn't update until afterward.

huanjani: This is the current state for rollbacks:
ECS Console:
Screen Shot 2022-12-23 at 9 52 42 AM

Progress tracker:
Screen Shot 2022-12-23 at 9 53 19 AM

efekarakus: 😭

Lou1415926: What about without `--force` flag? iirc, with `--force` flag, Copilot just waits for `UpdateService` to return, whereas without `--force` flag Copilot keeps reading information about the service and sort of "judge" whether the deployment is finished from the service information.

huanjani: without the `DependsOn:`, got into a `UPDATE_ROLLBACK_FAILED` state

huanjani: The larger problem I'm having right now. Circuit Breaker and ABR aren't playing nicely together even though [they should](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/deployment-type-ecs.html#deployment-failure-detection). 😭 
Screen Shot 2022-12-23 at 10 07 10 AM

huanjani: Update:
1. I'm removing the `DependsOn` because alarms aren't polled when a service is created, but this way they can be polled (if updated) in subsequent deployments
2. EDS is working on changing the Circuit Breaker error message that was being returned even if CB was disabled. 
3. @Lou1415926 was right-- the difference in behavior with and without `--force` is the main issue right now. Without `--force`, CFN gets in an `UPDATE_ROLLBACK_FAILED` every time an alarm is triggered.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -803,20 +803,23 @@ func convertEntryPoint(entrypoint manifest.EntryPointOverride) ([]string, error)
...
+	if in.RollbackAlarms.IsBasic() {
+		out.Rollback.AlarmNames = in.RollbackAlarms.Basic
 	}
-	// TODO (jwh): if AlarmArgs, create alarms and pass their Copilot-created names in CFN template.
-	return deployConfigs
+	if in.RollbackAlarms.IsAdvanced() {
+		out.Rollback.CPUUtilization = in.RollbackAlarms.Advanced.CPUUtilization
+		out.Rollback.MemoryUtilization = in.RollbackAlarms.Advanced.MemoryUtilization
+	}
+	return out
```
efekarakus: nit: a slightly different way of writing which sets all the defaults first and then configures only the unusual cases:
```go
out := template.DeploymentConfigurationOpts{
	MinHealthyPercent: minHealthyPercentDefault,	
	MaxPercent: maxPercentDefault,
	Rollback: RollingUpdateRollbackConfig{
		AlarmNames: in.RollbackAlarms.Basic,
		CPUUtilization: in.RollbackAlarms.Advanced.CPUUtilization,
		MemoryUtilization: in.RollbackAlarms.Advanced.MemoryUtilization,
	},
}

if strings.EqualFold(aws.StringValue(in.Rolling), manifest.ECSRecreateRollingUpdateStrategy) {
	out.MinHealthyPercent = minHealthyPercentRecreate
	out.MaxPercent = maxPercentRecreate
}
return out
```

internal/pkg/manifest/workload_ecs.go
```diff
@@ -87,14 +87,14 @@ type ImageWithPortAndHealthcheck struct {
 
 // AlarmArgs represents specs of CloudWatch alarms for deployment rollbacks.
 type AlarmArgs struct {
-	CPUUtilization    *int `yaml:"cpu_utilization"`
-	MemoryUtilization *int `yaml:"memory_utilization"`
+	CPUUtilization    *float64 `yaml:"cpu_utilization"`
```
efekarakus, line 90: Aww yeah thanks for catching this!

internal/pkg/manifest/workload_ecs.go
```diff
@@ -87,14 +87,14 @@ type ImageWithPortAndHealthcheck struct {
...
+	MemoryUtilization *float64 `yaml:"memory_utilization"`
 }
 
 // DeploymentConfiguration represents the deployment strategies for a service.
 type DeploymentConfiguration struct {
 	Rolling        *string                    `yaml:"rolling"`
-	RollbackAlarms Union[[]string, AlarmArgs] `yaml:"rollback_alarms"`
+	RollbackAlarms Union[[]string, AlarmArgs] // `yaml:"rollback_alarms"`
```
efekarakus: Can you drop a comment here explaining why we are not enabling this feature.

internal/pkg/template/templates/workloads/partials/cf/rollback-alarms.yml
```diff
@@ -0,0 +1,51 @@
...
+    AlarmDescription: "Roll back ECS service if CPU utilization is greater than or equal to {{.DeploymentConfiguration.Rollback.CPUUtilization}}% twice in 2 minutes."
+    AlarmName: {{.DeploymentConfiguration.Rollback.TruncateAlarmName .AppName .EnvName .WorkloadName "CopilotRollbackCPUAlarm"}}
+    Namespace: 'AWS/ECS'
+    Dimensions:
+      - Name: ClusterName
+        Value:
+          Fn::ImportValue:
+            !Sub '${AppName}-${EnvName}-ClusterId'
+      - Name: ServiceName
+        Value: !Select [2, !Split ["/", !Ref Service]]
```
efekarakus, line 16: Just to make sure the CFN stack did deploy right with `rollback_alarms` for a brand new LBWS?
I wanted to make sure that CFN created the alarms _after_ the service creation so the `DependsOn` is not needed

internal/pkg/template/templates/workloads/partials/cf/rollback-alarms.yml
```diff
@@ -0,0 +1,51 @@
...
+    MetricName: 'CPUUtilization'
+    ComparisonOperator: 'GreaterThanOrEqualToThreshold'
+    DatapointsToAlarm: 2
+    EvaluationPeriods: 2
+    Period: 60
```
efekarakus, line 21: I'm slightly worried that 2 mins as a default setting isn't enough to judge if a deployment was successful. Would you mind reaching out to @nathanpeck to see what defaults he would like to see in a tool like Copilot and CDK extensions?

(don't block on this feedback) you can apply it retroactively in a separate PR.

My preference for a default would be:
```
    DatapointsToAlarm: 8
    EvaluationPeriods: 10
    Period: 60
```

internal/pkg/template/workload_test.go
```diff
@@ -426,3 +428,35 @@ func TestEnvControllerParameters(t *testing.T) {
...
+			inSvc: "shortSvcName",
+			inAlarmType: "CopilotRollbackMemAlarm",
+			expected: "shortAppName-shortEnvName-shortSvcName-CopilotRollbackMemAlarm",
+		},
+		"with need to truncate at 76 chars per element": {
+			inApp: "12345678911234567892123456789312345678941234567895123456789612345678971234567898",
+			inEnv: "12345678911234567892123456789312345678941234567895123456789612345678971234567898",
+			inSvc: "12345678911234567892123456789312345678941234567895123456789612345678971234567898",
+			inAlarmType: "CopilotRollbackCPUAlarm",
+			expected: "1234567891123456789212345678931234567894123456789512345678961234567897123456-1234567891123456789212345678931234567894123456789512345678961234567897123456-1234567891123456789212345678931234567894123456789512345678961234567897123456-CopilotRollbackCPUAlarm",
```
efekarakus, line 453: 🤣

efekarakus (approved):
Looks great! just a small feedback, feel free to remove the label after addressing

Lou1415926 (approved):
LGTM! I have nothing to add in addition to Efe's comments :shipit:

internal/pkg/template/templates/workloads/partials/cf/rollback-alarms.yml
huanjani: [pending]

internal/pkg/template/templates/workloads/partials/cf/rollback-alarms.yml
huanjani: Yup yup!

PR #4295, opened 2022-12-21
chore(templates): create CFN template for environment-level aurora serverless v2 storage option

Lou1415926:
With support of https://github.com/aws/copilot-cli/issues/4219, it is possible now to create environment-level storage options. This is the aurora serverless v2 storage option.


internal/pkg/template/templates/addons/aurora/env/serverlessv2.yml
```diff
@@ -0,0 +1,178 @@
...
+  {{logicalIDSafe .ClusterName}}WorkloadSecurityGroup:
+    Metadata:
+      'aws:copilot:description': 'A security group for one or more workloads to access the Aurora Serverless v2 cluster {{logicalIDSafe .ClusterName}}'
+    Type: 'AWS::EC2::SecurityGroup'
+    Properties:
+      GroupDescription: !Sub 'The Security Group to access Aurora Serverless v2 cluster {{logicalIDSafe .ClusterName}}.'
+      VpcId: !Ref VPCID
+      Tags:
+        - Key: Name
+          Value: !Sub 'copilot-${App}-${Env}-{{ .ClusterName }}AuroraCluster'
```
iamhopaul123: Why do we use `{{ .ClusterName }}` instead of `{{ .Name }}` to be consistent with the wkld addons?

internal/pkg/template/templates/addons/aurora/env/serverlessv2.yml
```diff
@@ -0,0 +1,178 @@
...
+  {{logicalIDSafe .ClusterName}}Secret: # injected as {{envVarSecret .ClusterName | toSnakeCase}} environment variable by Copilot.
+    Description: "The JSON secret that holds the database username and password. Fields are 'host', 'port', 'dbname', 'username', 'password', 'dbClusterIdentifier' and 'engine'"
+    Value: !Ref {{logicalIDSafe .ClusterName}}AuroraSecret
+    Export:
+      Name: !Sub ${Env}-{{logicalIDSafe .ClusterName}}AuroraSecret
+  {{logicalIDSafe .ClusterName}}SecurityGroup:
+    Description: "The security group to attach to the workload."
+    Value: !Ref {{logicalIDSafe .ClusterName}}WorkloadSecurityGroup  
+    Export:
+      Name: !Sub ${Env}-{{logicalIDSafe .ClusterName}}SecurityGroup
```
iamhopaul123: Do we want to namespace the stack output such as `copilot-${App}-${Env}-{{logicalIDSafe .ClusterName}}SecurityGroup`

iamhopaul123 (commented):
LGTM overall with a general question: do we want to keep the env addons and wkld addons consistent and add the tag and export the outputs as well for the wkld addons?

internal/pkg/template/templates/addons/aurora/env/serverlessv2.yml
Lou1415926: What do you think of `${App}-${Env}-{{logicalIDSafe .ClusterName}}SecurityGroup` instead so that people can write in manifest:
```
security_groups: 
  - from_cfn: ${COPILOT-APPLICATION-NAME}...
```
instaed of 
```
security_groups: 
  - from_cfn: copilot-${COPILOT-APPLICATION-NAME}...
```

internal/pkg/template/templates/addons/aurora/env/serverlessv2.yml
Lou1415926: The field name is [`ClusterName`](https://github.com/aws/copilot-cli/blob/1e43932a3c6f2ff870bb25d743d53edee42a8756/internal/pkg/addon/storage.go#L163) instead of `Name`. We probably could rename the field to `Name` - but whether we name the field `ClusterName` or `Name` makes no difference to users. 

I *think* (but can't recall exactly) the intention behind `ClusterName` is to be clear (to us devs) that this would be the *exact* name of the cluster - the actual name of the cluster will be `ClusterName`, instead of `ClusterName-RANDOMSTRING`.

Lou1415926:
> do we want to keep the env addons and wkld addons consistent and add the tag 

Do you mean the tags [here](https://github.com/aws/copilot-cli/blob/4c58b22f8067f7bce1be65f5daf5cf957eb79a76/internal/pkg/template/templates/addons/aurora/env/serverlessv2.yml#L52) and [here](https://github.com/aws/copilot-cli/blob/4c58b22f8067f7bce1be65f5daf5cf957eb79a76/internal/pkg/template/templates/addons/aurora/env/serverlessv2.yml#L63)?

> and export the outputs as well for the wkld addons?

I don't think we want to export the outputs for wkld addons since there is no need. We only export the outputs here for a better integration with svc manifest.

iamhopaul123:
> > do we want to keep the env addons and wkld addons consistent and add the tag
> 
> Do you mean the tags [here](https://github.com/aws/copilot-cli/blob/4c58b22f8067f7bce1be65f5daf5cf957eb79a76/internal/pkg/template/templates/addons/aurora/env/serverlessv2.yml#L52) and [here](https://github.com/aws/copilot-cli/blob/4c58b22f8067f7bce1be65f5daf5cf957eb79a76/internal/pkg/template/templates/addons/aurora/env/serverlessv2.yml#L63)?
> 
> > and export the outputs as well for the wkld addons?
> 
> I don't think we want to export the outputs for wkld addons since there is no need. We only export the outputs here for a better integration with svc manifest.

Gotcha. How about the tags?

internal/pkg/template/templates/addons/aurora/env/serverlessv2.yml
iamhopaul123: I was trying to say `${Name}` but realized it is the wkld name. It reminds me tho since it is an addons that belongs to an env, maybe tagging it with  `copilot-${App}-${Env}-AuroraCluster` should be good enough?

Lou1415926:
> Gotcha. How about the tags?

If you were asking about the tags on `{{logicalIDSafe .ClusterName}}WorkloadSecurityGroup` and `{{logicalIDSafe .ClusterName}}DBClusterSecurityGroup` - 

We already have the tags in the workload addons for the `{{logicalIDSafe .ClusterName}}WorkloadSecurityGroup` counterpart:  https://github.com/aws/copilot-cli/blob/1e43932a3c6f2ff870bb25d743d53edee42a8756/internal/pkg/template/templates/addons/aurora/serverlessv2.yml#L44

We don't, however, have the tags for https://github.com/aws/copilot-cli/blob/1e43932a3c6f2ff870bb25d743d53edee42a8756/internal/pkg/template/templates/addons/aurora/serverlessv2.yml#L47 


and yeah I think we can add it!

iamhopaul123 (approved):

huanjani (approved):
LGTM! Great balance of defaults/guidance for configuration.

PR #4297, opened 2022-12-21
refactor: manifest pkg to add nlb config for multiple ports

paragbhingre:
This PR refactors the manifest package in order to add NLB exposed ports of the containers to the exposedPorts list that we created in the PR #4268 


internal/pkg/manifest/lb_web_svc.go
```diff
@@ -254,6 +254,10 @@ func (lbws *LoadBalancedWebService) ExposedPorts() ([]ExposedPort, error) {
 		exposedPorts = append(exposedPorts, out...)
 	}
 	exposedPorts = append(exposedPorts, lbws.RoutingRule.exposedPorts(exposedPorts, workloadName)...)
-
+	out, err := lbws.NLBConfig.exposedPorts(exposedPorts, workloadName)
```
iamhopaul123, line 257: Do we need this for the other svc types like backend svc?

internal/pkg/manifest/svc.go
```diff
@@ -495,6 +495,46 @@ func (rr RoutingRuleConfiguration) exposedPorts(exposedPorts []ExposedPort, work
...
+		return nil, err
+	}
+	if protocol == nil {
+		protocol = aws.String("tcp")
+	}
+
+	port, err := strconv.ParseUint(aws.StringValue(nlbPort), 10, 16)
+	if err != nil {
+		return nil, err
+	}
```
iamhopaul123, line 519: I feel like we don't need this whole thing if `cfg.TargetPort != nil`

internal/pkg/manifest/svc.go
```diff
@@ -495,6 +495,46 @@ func (rr RoutingRuleConfiguration) exposedPorts(exposedPorts []ExposedPort, work
...
+		containerPort = uint16(aws.IntValue(cfg.TargetPort))
+	}
+	for _, exposedPort := range exposedPorts {
+		if containerPort == exposedPort.Port && cfg.TargetContainer == nil {
+			return nil, nil
+		}
+	}
+	return []ExposedPort{
+		{
+			Port:          containerPort,
```
iamhopaul123: qq: why do we keep `Port` to be `uint16` instead of `string`?

internal/pkg/manifest/svc.go
```diff
@@ -495,6 +495,46 @@ func (rr RoutingRuleConfiguration) exposedPorts(exposedPorts []ExposedPort, work
...
+	port, err := strconv.ParseUint(aws.StringValue(nlbPort), 10, 16)
+	if err != nil {
+		return nil, err
+	}
+	containerPort := uint16(port)
+	if cfg.TargetPort != nil {
+		containerPort = uint16(aws.IntValue(cfg.TargetPort))
+	}
+	for _, exposedPort := range exposedPorts {
+		if containerPort == exposedPort.Port && cfg.TargetContainer == nil {
```
iamhopaul123: what if they set `TargetContainer` to be the main container? This condition would not be satisfied and we'll end up exposing a duplicated port for the main container 🤔

internal/pkg/manifest/validate.go
```diff
@@ -1856,6 +1860,62 @@ func populateALBPortsAndValidate(containerNameFor map[uint16]string, opts valida
...
+		protocolVal := aws.StringValue(protocol)
+		var isValidProtocol bool
+		for _, valid := range nlbValidProtocols {
+			if strings.EqualFold(protocolVal, valid) {
+				isValidProtocol = true
+				break
+			}
+		}
+		if !isValidProtocol {
+			return fmt.Errorf(`invalid protocol %s; valid protocols include %s`, protocolVal, english.WordSeries(nlbValidProtocols, "and"))
```
iamhopaul123: Isn't NLB port already validated [here](https://github.com/aws/copilot-cli/blob/856e9dd0cdb46077e3db77e4745d0e60c106dd8a/internal/pkg/manifest/validate.go#L806)?

internal/pkg/manifest/validate.go
```diff
@@ -1856,6 +1860,62 @@ func populateALBPortsAndValidate(containerNameFor map[uint16]string, opts valida
...
+	containerPort := uint16(port)
+	if nlb.TargetPort != nil {
+		containerPort = uint16(aws.IntValue(nlb.TargetPort))
+	}
+
+	containerName := opts.mainContainerName
+	existingContainerName := containerNameFor[containerPort]
+
+	if existingContainerName != "" {
+		containerName = existingContainerName
```
iamhopaul123: can we remove this line?

internal/pkg/manifest/validate.go
```diff
@@ -1856,6 +1860,62 @@ func populateALBPortsAndValidate(containerNameFor map[uint16]string, opts valida
...
+		if nlb.TargetContainer != nil {
```
iamhopaul123, line 1902: I don't think this condition is needed here.

internal/pkg/manifest/lb_web_svc.go
paragbhingre: We support NLB only for LBWS at the moment.

internal/pkg/manifest/svc.go
paragbhingre: Do you think following is a good idea?

```
if cfg.TargetPort == nil {
  nlbPort, protocol, err := ParsePortMapping(cfg.Port)
	  if err != nil {
		  return nil, err
	  }
	  if protocol == nil {
		  protocol = aws.String("tcp")
	  }
  
	  port, err := strconv.ParseUint(aws.StringValue(nlbPort), 10, 16)
	  if err != nil {
		  return nil, err
	  }
} else {
  containerPort = uint16(aws.IntValue(cfg.TargetPort))
}
```

internal/pkg/manifest/svc.go
paragbhingre: There is no solid reason behind keeping it as `uint16` but we accept primary container's port as `uint16` so I thought I should keep it consistent and use as much as `uint16` everywhere so that there won't frequent conversion issue going forward from `uint16` to string and vice a versa.

internal/pkg/manifest/svc.go
paragbhingre: I think removing `cfg.TargetContainer == nil` resolves all the problems. I thought I committed the change of removing this `cfg.TargetContainer == nil` condition before going on a vacation. I will commit it asap.

internal/pkg/manifest/validate.go
paragbhingre: good catch, I will remove this part as it is not required.

internal/pkg/manifest/validate.go
paragbhingre: May I ask why do we want to remove this line?

internal/pkg/manifest/svc.go
```diff
@@ -495,6 +495,46 @@ func (rr RoutingRuleConfiguration) exposedPorts(exposedPorts []ExposedPort, work
 	}
 }
 
+// exportPorts returns any new ports that should be exposed given the network load balancer
+// configuration that's not part of the existing containerPorts.
+func (cfg NetworkLoadBalancerConfiguration) exposedPorts(exposedPorts []ExposedPort, workloadName string) ([]ExposedPort, error) {
```
Lou1415926, line 504: 💭  I wonder if this ⬇️ can simplify the logic a little... It looks a bit tricky that we are appending and passing `exposedPorts` as a parameter around.
```go
// exportPorts returns any new ports that should be exposed given the application load balancer configuration.
func (rr ALBConfig) exposedPorts() ([]ExposedPort, error)

// exportPorts returns any new ports that should be exposed given the network load balancer configuration.
func (cfg NLBConfig) exposedPorts() ([]ExposedPort, error)

// exportPorts returns any new ports that should be exposed given the image configuration.
func (cfg ImageWithPortAndHealthcheck) exposedPorts(workloadName string) []ExposedPort

// exportPorts returns any new ports that should be exposed given the image configuration.
func (cfg ImageWithHealthcheckAndOptionalPort) exposedPorts(workloadName string) []ExposedPort

func (b *BackendService) ExposedPorts() ([]ExposedPort, error) {
	var exposedPorts []ExposedPort
	workloadName := aws.StringValue(b.Name)
	exposedPorts = append(exposedPorts, b.ImageConfig.exposedPorts(workloadName)...)
	for name, sidecar := range b.Sidecars {
		out, err := sidecar.exposedPorts(name)
		if err != nil {
			return nil, err
		}
		exposedPorts = append(exposedPorts, out...)
	}
	exposedPorts = append(exposedPorts, b.RoutingRule.exposedPorts(workloadName)...)
	return sortExposedPorts(exposedPorts), nil
}

func sortExposedPorts(ports []ExposedPorts) []ExposedPort {
   // Remove repeatedly exposed ports.
   // Sort
}
```

internal/pkg/manifest/validate.go
iamhopaul123: because `existingContainerName` would be a better name and easier to understand than just `containerName`

internal/pkg/manifest/validate.go
```diff
@@ -1774,6 +1775,9 @@ func validateExposedPorts(opts validateExposedPortsOpts) error {
 	if err := populateALBPortsAndValidate(containerNameFor, opts); err != nil {
 		return err
 	}
+	if err := populateNLBPortsAndValidate(containerNameFor, opts); err != nil {
```
Lou1415926, line 1806: quick question: what is the responsibility of `validateExposedPorts`?
1. validate that no port is exposed by two containers
2. validate that the protocol is valid
Is this the complete list of responsibilities?

internal/pkg/manifest/svc.go
iamhopaul123: hmm it's ok maybe stay as is

internal/pkg/manifest/svc.go
paragbhingre: I had the exact same logic written before but the only problem with this logic is sortExposedPorts doesn't know which duplicate port to remove out of two ports. 

Let me take an example manifest
```
name: lbservice

image:
  port: 80

http:
  path: '/'
  target_port: 81
  
sidecars:
  nginx:
    image:
      port: 81
```

In the following manifest we should expose following ports 
lbservice - 80
nginx - 81

But with the logic mentioned above we will get following exposed ports 
lbservice - 80
lbservice - 81
nginx - 81

And now we do not know which duplicate port to be removed from the list? Does that make sense?

internal/pkg/manifest/validate.go
paragbhingre: Yes making sure that we error out when customer defines duplicate ports. 
Validating protocol is extra thing that I was doing here which is already being done in the validate NLB. I will remove that code in the next commit. 

Is there anything else you would want me to take care as a part of `validateExposedPorts`?

Lou1415926: no no I think it makes sense that "1. validate duplicate ports" is the sole purpose of `validateExposedPorts`. In fact, in addition to NLB, what do you think of moving the protocol check in `validateSidecarExposedPorts` to sidecar level as well? If I understand the design correctly,  you don't need any service-level information in order to know a protocol is invalid. 

If you do decide to do this, don't worry about adding it to this PR! We can do it in a follow-up PR.

internal/pkg/manifest/validate.go
```diff
@@ -1856,6 +1860,62 @@ func populateALBPortsAndValidate(containerNameFor map[uint16]string, opts valida
...
+	}
+	containerPort := uint16(port)
+	if nlb.TargetPort != nil {
+		containerPort = uint16(aws.IntValue(nlb.TargetPort))
+	}
+
+	containerName := opts.mainContainerName
+	existingContainerName := containerNameFor[containerPort]
+
+	if existingContainerName != "" {
```
Lou1415926: Is the following snippet equivalent?
```go
	if exposed, ok := containerNameFor[containerPort]; ok {
		if nlb.TargetContainer != nil && aws.StringValue(nlb.TargetContainer) != exposed {
			return &errContainersExposingSamePort{
				firstContainer:  aws.StringValue(nlb.TargetContainer),
				secondContainer: containerName,
				port:            containerPort,
			} 
		}
		return nil
	}
	
        // The port isn't exposed. Add it to the map of exposed ports. 
	targetContainer := opts.mainContainerName
	if nlb.TargetContainer != nil {
		targetContainer = aws.StringValue(nlb.TargetContainer)
	}
	containerNameFor[containerPort] = targetContainer
```

paragbhingre: yes looks good to me. Just made little modification in it. Changed `containerName` to `targetContainerName` for better understanding. And `secondContainer: containerName` changed to `secondContainer: exposed`

internal/pkg/manifest/validate.go
paragbhingre: Done

iamhopaul123 (approved):

internal/pkg/manifest/svc.go
```diff
@@ -495,6 +495,46 @@ func (rr RoutingRuleConfiguration) exposedPorts(exposedPorts []ExposedPort, work
...
+	if cfg.IsEmpty() {
+		return nil, nil
+	}
+	containerName := workloadName
```
Lou1415926: I have a few question/nits:
1. Would `targetContainer` be an accurate name for `containerName`? Would `targetPort` be an accurate name for `port`? If so, can we rename the variables to `targetContainer` and `targetPort`?
2. If `containerName` is only used to add the port to `exposedPort` then can we parse its value toward the end, just before `return []ExposedPort{...}`? My reasoning is that this increases code readability. When I was reading the snippet, I kept wondering "what is `containerName` for" and didn't find out why until I reached the end 😂 

For example,
```go
if cfg.IsEmpty() {
  return nil, nil
}

// parse target port
// check if port is already exposed
// port isn't exposed yet, then...
targetContainer := workloadName
if cfg.TargetContaner != nil {
  targetContainer = aws.StringValue(cfg.TargetContainer)
}
return []ExposedPort{
  {
      Port: targetPort,
      Protocol:      aws.StringValue(protocol),
      ContainerName: targetContainer,
   }
}
```

internal/pkg/manifest/validate_test.go
```diff
@@ -3548,6 +3548,76 @@ func TestValidateExposedPorts(t *testing.T) {
...
+				},
+				alb: &RoutingRuleConfiguration{
+					TargetPort: aws.Uint16(5001),
+				},
+				nlb: &NetworkLoadBalancerConfiguration{
+					Port:       aws.String("5001/tcp"),
+					TargetPort: aws.Int(5001),
+				},
+			},
+			wanted: nil,
```
Lou1415926: nit:
```suggestion
```

internal/pkg/manifest/validate_test.go
```diff
@@ -3548,6 +3548,76 @@ func TestValidateExposedPorts(t *testing.T) {
...
+					TargetPort:      aws.Uint16(5001),
+					TargetContainer: aws.String("foo"),
+				},
+				nlb: &NetworkLoadBalancerConfiguration{
+					Port:            aws.String("5001/tcp"),
+					TargetPort:      aws.Int(5001),
+					TargetContainer: aws.String("foo"),
+				},
+			},
+			wanted: nil,
```
Lou1415926: nit:
```suggestion
```

internal/pkg/manifest/validate_test.go
```diff
@@ -3548,6 +3548,76 @@ func TestValidateExposedPorts(t *testing.T) {
...
+		},
+		"should return an error if alb and nlb target_port trying to expose same container port of different containers": {
```
Lou1415926, line 3595: nice test!

internal/pkg/manifest/lb_web_svc_test.go
```diff
@@ -1935,6 +1935,241 @@ func TestLoadBalancedWebService_ExposedPorts(t *testing.T) {
 				},
 			},
 		},
+		"expose new primary container port only once through ALB and NLB config": {
```
Lou1415926: nit:
```suggestion
		"ALB and NLB exposes the same additional port on the main container": {
```

internal/pkg/manifest/lb_web_svc_test.go
```diff
@@ -1935,6 +1935,241 @@ func TestLoadBalancedWebService_ExposedPorts(t *testing.T) {
...
+					Protocol:      "tcp",
+				},
+				{
+					Port:          2000,
+					ContainerName: "xray",
+					Protocol:      "tcp",
+				},
+			},
+		},
+		"expose two new primary container port through ALB and NLB config": {
```
Lou1415926: nit:
```suggestion
		"ALB and NLB exposes two different ports on the main container": {
```

internal/pkg/manifest/lb_web_svc_test.go
```diff
@@ -1935,6 +1935,241 @@ func TestLoadBalancedWebService_ExposedPorts(t *testing.T) {
...
+					Name: aws.String("frontend"),
+				},
+				LoadBalancedWebServiceConfig: LoadBalancedWebServiceConfig{
+					ImageConfig: ImageWithPortAndHealthcheck{
+						ImageWithPort: ImageWithPort{
+							Port: aws.Uint16(8080),
+						},
+					},
+					NLBConfig: NetworkLoadBalancerConfiguration{
+						Port:            aws.String("8082"),
```
Lou1415926: Let's also test `TargetPort` for `NLBConfig` in one of these tests:
```suggestion
						Port:            aws.String("8082/tcp"),
						TargetPort: aws.String("8083")
```

internal/pkg/manifest/lb_web_svc_test.go
Lou1415926: Also, nice tests!!

Lou1415926 (approved):
LGTM! I just have a few nits. Feel free to remove the label when you are ready.

internal/pkg/manifest/svc.go
paragbhingre: Changing `containerName` to `targetContainer` and `containerPort` to `targetPort`.

PR #4301, opened 2022-12-22
chore(template): add WalkOverridesCDKDir method

efekarakus:
Related #4208 

### DIFF
--- a/internal/pkg/template/overrides.go
+++ b/internal/pkg/template/overrides.go
@@ -16,7 +16,7 @@ const (
 )
 
 var (
-	shortenServiceName = map[string]string{
+	cdkAliasForService = map[string]string{
 		"ApiGatewayV2":           "apigwv2",
 		"AppRunner":              "ar",
 		"AutoScalingPlans":       "asgplans",
@@ -72,7 +72,7 @@ func (t CFNType) ImportShortRename() string {
 	parts := strings.Split(string(t), "::")
 	name := parts[1]
 
-	if rename, ok := shortenServiceName[name]; ok {
+	if rename, ok := cdkAliasForService[name]; ok {
 		return rename
 	}
 	return strings.ToLower(name)
@@ -98,15 +98,15 @@ func (t *Template) WalkOverridesCDKDir(resources []CFNResource, fn WalkDirFunc) 
 	}, fn, WithFuncs(
 		map[string]interface{}{
 			// transform all the initial capital letters into lower letters.
-			"camelCase": func(pascal string) string {
-				firstSmall := len(pascal)
-				for i, r := range pascal {
+			"lowerInitialLetters": func(serviceName string) string {
+				firstSmall := len(serviceName)
+				for i, r := range serviceName {
 					if unicode.IsLower(r) {
 						firstSmall = i
 						break
 					}
 				}
-				return strings.ToLower(pascal[:firstSmall]) + pascal[firstSmall:]
+				return strings.ToLower(serviceName[:firstSmall]) + serviceName[firstSmall:]
 			},
 		},
 	))
--- a/internal/pkg/template/overrides.go
+++ b/internal/pkg/template/overrides.go
@@ -15,6 +15,28 @@ const (
 	cdkTemplatesPath        = "overrides/cdk"
 )
 
+var (
+	shortenServiceName = map[string]string{
+		"ApiGatewayV2":           "apigwv2",
+		"AppRunner":              "ar",
+		"AutoScalingPlans":       "asgplans",
+		"ApplicationAutoScaling": "appasg",
+		"AutoScaling":            "asg",
+		"CertificateManager":     "acm",
+		"CloudFormation":         "cfn",
+		"CloudFront":             "cf",
+		"ServiceDiscovery":       "sd",
+		"CloudWatch":             "cw",
+		"CodeBuild":              "cb",
+		"CodePipeline":           "cp",
+		"DynamoDB":               "ddb",
+		"ElasticLoadBalancingV2": "elbv2",
+		"OpenSearchService":      "oss",
+		"Route53":                "r53",
+		"StepFunctions":          "sfn",
+	}
+)
+
 // CFNResource represents a resource rendered in a CloudFormation template.
 type CFNResource struct {
 	Type      CFNType
@@ -23,8 +45,8 @@ type CFNResource struct {
 
 type cfnResources []CFNResource
 
-// Types returns the list of unique CFN types.
-func (rs cfnResources) Types() []CFNType {
+// UniqueTypes returns the list of unique CFN types.
+func (rs cfnResources) UniqueTypes() []CFNType {
 	var uniqueTypes []CFNType
 	seen := make(map[CFNType]struct{})
 	for _, r := range rs {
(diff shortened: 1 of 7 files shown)


internal/pkg/template/overrides.go
```diff
@@ -0,0 +1,100 @@
...
+// CFNResource represents a resource rendered in a CloudFormation template.
+type CFNResource struct {
+	Type      CFNType
+	LogicalID string
+}
+
+type cfnResources []CFNResource
+
+// Types returns the list of unique CFN types.
+func (rs cfnResources) Types() []CFNType {
```
iamhopaul123: Maybe `UniqueTypes()`?

internal/pkg/template/overrides.go
```diff
@@ -0,0 +1,100 @@
...
+	return uniqueTypes
+}
+
+// CFNType is a CloudFormation resource type such as "AWS::ECS::Service".
+type CFNType string
+
+// ImportName returns the name of the CDK package for the given CloudFormation type.
+func (t CFNType) ImportName() string {
+	parts := strings.Split(strings.ToLower(string(t)), "::")
+	return fmt.Sprintf("aws_%s", parts[1])
```
iamhopaul123, line 67: Could it causes panic?

internal/pkg/template/overrides.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// ImportShortRename returns a human-friendly shortened rename of the CDK package for the given CloudFormation type.
+func (t CFNType) ImportShortRename() string {
+	parts := strings.Split(string(t), "::")
+
+	// If the service name is camel case values, like ElasticLoadBalancingV2, then only grab the upper case
+	// letters and numbers to generate "elbv2"
```
iamhopaul123: Could it cause conflits potentially? Also I feel like we don't need this method because CFN generally does a pretty good job trying to make the service name short for their resources. I wonder if ELBV2 would be the only one, and having this function might cause unexpected error in the future.

internal/pkg/template/overrides.go
```diff
@@ -0,0 +1,100 @@
...
+		if _, ok := seen[r.Type]; !ok {
+			uniqueTypes = append(uniqueTypes, r.Type)
+		}
+		seen[r.Type] = struct{}{}
+	}
+	return uniqueTypes
+}
+
+// CFNType is a CloudFormation resource type such as "AWS::ECS::Service".
+type CFNType string
```
iamhopaul123, line 62: I feel like we need a mechanism to make sure the `string` _is_ a `CFNType`. Maybe regex validation etc, so that we won't cause panic when calling `ImportName()` etc.

internal/pkg/template/overrides.go
```diff
@@ -0,0 +1,100 @@
...
+		Resources         cfnResources
+	}
+	return t.walkDir(cdkTemplatesPath, cdkTemplatesPath, metadata{
+		Version:           cdkVersion,
+		ConstructsVersion: cdkConstructsMinVersion,
+		Resources:         resources,
+	}, fn, WithFuncs(
+		map[string]interface{}{
+			// transform all the initial capital letters into lower letters.
+			"camelCase": func(pascal string) string {
```
iamhopaul123: Does it work based on an assumption that `pascal` has been processed by `ImportShortRename`?

internal/pkg/template/overrides.go
efekarakus: Done!

internal/pkg/template/overrides.go
efekarakus: I think we are good boss and there is no need for validation here.
The cloudformation types won't be passed by users, they are retrieved from our CloudFormation templates.

And they are guaranteed to be of this format:
> Resource type identifiers **always** take the following form:
> ```
> service-provider::service-name::data-type-name
> ```
> https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html

internal/pkg/template/overrides.go
efekarakus: Nope, see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html

internal/pkg/template/overrides.go
efekarakus: > because CFN generally does a pretty good job trying to make the service name short for their resources

😅 lol it's okay if we disagree here see the unit tests for sample resources that we use that wouldn't be nice without a short form (AutoScaling, ServiceDiscovery, ElasticLoadBalancingV2, CloudFormation, CloudFront, CertificateManager). 
But I totally agree with your concern about this being potentially too much magic leading to a conflict!

I changed the implementation to be simple by looking up a short form from a `map`. Let me know what you think!

internal/pkg/template/overrides.go
efekarakus: nono this is just making the assumption that the first letters of the service name is capital really: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html

like `SSM` or `ApiGatewayV2`, it transform all the initial uppercase letters to small

internal/pkg/template/overrides.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+const (
+	cdkVersion              = "2.56.0"
+	cdkConstructsMinVersion = "10.0.0"
+	cdkTemplatesPath        = "overrides/cdk"
+)
+
+var (
+	shortenServiceName = map[string]string{
```
Lou1415926: tiny nit: do boss know if there is any convention around starting a var name with a verb? What do you think of `shortenedServiceName` or `shortServiceName`?

Besides, these short names seem to be cdk-specific, do you think we want to add the `CDK` keyword in the var name?

internal/pkg/template/overrides.go
```diff
@@ -0,0 +1,113 @@
...
+	name := parts[1]
+
+	if rename, ok := shortenServiceName[name]; ok {
+		return rename
+	}
+	return strings.ToLower(name)
+}
+
+// L1ConstructName returns the name of the L1 construct representing the CloudFormation type.
+func (t CFNType) L1ConstructName() string {
```
Lou1415926, line 82: These seem pretty CDK-specific. Do we plan to rename `overrides.go` to `overrides/cdk.go` in the future when/if we support other providers?

internal/pkg/template/overrides.go
Lou1415926: > just making the assumption that the first letters of the service name is capital

Is it possible that there is a service name that capitalize the first few letters - not just the first - e.g. `"ECSServiceConfig"`? `camelCase` will transform this into `ecsserviceConfig`

iamhopaul123: It's fine it's gonna be `ecsServiceConfig`

iamhopaul123 (approved):

internal/pkg/template/overrides.go
efekarakus: renamed to `cdkAliasForService`!

internal/pkg/template/overrides.go
efekarakus: This is the `template` package 😄
My plan was to shove both CDK and YAMLPatch types into this file, does that make sense to you or would you prefer a `template/cdk_overrides.go` file instead?
I'm fine either way.

internal/pkg/template/overrides.go
efekarakus: > Is it possible that there is a service name that capitalize the first few letters - not just the first - e.g. "ECSServiceConfig"? camelCase will transform this into ecsserviceConfig

Looks like we do have some resources where that's possible, the first one that I saw is this: 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/AWS_EMRContainers.html

I am not super worried though! This fn is just a nice to have and it's used to set a variable name, as long as the output roughly makes sense i think it's fine:
https://github.com/aws/copilot-cli/pull/4301/files/c641f377464c98b495e53df5f6814a0dca76fa41#diff-e1d42ac36c232f8ba809aedec6dfb05e794faec45561621cbac1a622ec155c96R23

I renamed the function to be more explicit:
```go
"lowerInitialLetters": func(serviceName string) string 
```

internal/pkg/template/overrides.go
Lou1415926: `template/overrides.go` is fine with me! If it gets too bulky in the future it's easy to restructure anyway!

Lou1415926 (approved):

PR #4303, opened 2022-12-23
chore: fix parsing of environment addons

Lou1415926:
Patching #4280 


iamhopaul123 (approved):

internal/pkg/cli/deploy/env.go
```diff
@@ -370,16 +370,16 @@ func (d *envDeployer) buildStackInput(in *DeployEnvironmentInput) (*cfnstack.Env
 
 	var addons *cfnstack.Addons
```
efekarakus: tiny nit: can we move this variable declaration to after l.376, once the parsing is known to be either successfull or a `NotFoundError`

efekarakus (approved):

Issue #4220, opened 2022-11-23
Import Values From Other CFN Stacks in Workload Manifests


Lou1415926:
### Proposal

Today, user can use [`from_tags`](https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#network-vpc-placement-subnets-from-tags) (https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#network-vpc-placement-subnets-from-tags)  in a workload manifest under `network.vpc.placement.subnets` to specify external subnets for a service/job based on AWS tags. This proposal introduces `from_cfn`, an additional way to connect resources from other CFN stack outputs into a service/job.

We expect the `from_cfn` feature to be useful if you have existing AWS CloudFormation stacks that you’d like to connect with your Copilot workloads or if you’re using [environment addons #4219](https://github.com/aws/copilot-cli/issues/4219).

### Experience

To reference a value from another CFN stack, users should first export the output value from the source stack:
```yaml
# In a different stack, e.g. named "appmesh-stack".
Outputs:
  UserDBAccessSecurityGroup:
     Value: !GetAtt UserDBAccessSecurityGroup.Id
     Export: 
        Name: MyUserDBAccessSecurityGroup # <- Unique export name within the region.
```
Now, suppose I want to associate this security group with my `web` service, I will use `from_cfn`:
```yaml
Name: web
Type: Load Balanced Web Service

network:
  vpc:
    security_groups:
       - sg-1234
       - from_cfn: MyUserDBAccessSecurityGroup # <- Import the value from the other stack.
```
This will render `!ImportValue MyUserDBAccessSecurityGroup` into `web`’s service stack. Seeing `!ImportValue`, CloudFormation will try to fetch the value and use it as a security group ID for the web service. 

For now, `from_cfn` will be added to the following manifest fields only:

1. `network.vpc.security_groups`
2. `variables`
3. `secrets`

### Use `from_cfn` to reference a value from environment addons

This feature can work in conjunction with [environment addons #4219](https://github.com/aws/copilot-cli/issues/4219) to provide a full story of creating additional environment-level resources. After user exports a value from an environment, they can use that export name in a service manifest via `from_cfn`.

Following the example started in the [environment addons proposal #4219](https://github.com/aws/copilot-cli/issues/4219) - say we now want to inject `MyAppMeshName` into my `web` service as an environment variable, we can write:
```yaml
Name: web
Type: Load Balanced Web Service

variables:
[...]

PR #4314, opened 2023-01-03
chore: implement `from_cfn` for security groups to import values from other CFN stacks

KollaAdithya:
This PR implements #4220

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/testdata/workloads/job-manifest.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/workloads/job-manifest.yml
@@ -58,6 +58,14 @@ sidecars:
     labels:
       com.amazonaws.ecs.copilot.sidecars.nginx.description: tricky
 
+network:
+  vpc:
+    security_groups:
+      - sg-0c10c4fe23f5e5361
+      - sg-09295097b2a41b59d
+      - from_cfn: MyUserDBAccessSecurityGroup1
+      - from_cfn: MyUserDBAccessSecurityGroup2
+
 environments:
   test:
     image:
--- a/internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-manifest.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-manifest.yml
@@ -40,13 +40,6 @@ publish:
     - name: givesdogs
     - name: mytopic
       fifo: true
-network:
-  vpc:
-    security_groups:
-      - sg-0c10c4fe23f5e5361
-      - sg-09295097b2a41b59d
-      - from_cfn: MyUserDBAccessSecurityGroup1
-      - from_cfn: MyUserDBAccessSecurityGroup2
 # Optional fields for more advanced use-cases.
 #
 variables:                    # Pass environment variables as key value pairs.
@@ -56,6 +49,14 @@ variables:                    # Pass environment variables as key value pairs.
 #  GITHUB_TOKEN: GITHUB_TOKEN  # The key is the name of the environment variable, the value is the name of the SSM parameter.
 # You can override any of the values defined above by environment.
 environments:
+  test:
+    network:
+      vpc:
+        security_groups:
+          - sg-0c10c4fe23f5e5361
+          - sg-09295097b2a41b59d
+          - from_cfn: MyUserDBAccessSecurityGroup1
+          - from_cfn: MyUserDBAccessSecurityGroup2
   staging:
     image:
       location: 123456789000.dkr.ecr.us-east-1.amazonaws.com/vault/e2e:${TAG}
--- a/internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-manifest.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-manifest.yml
@@ -40,6 +40,13 @@ publish:
     - name: givesdogs
     - name: mytopic
       fifo: true
+network:
+  vpc:
+    security_groups:
+      - sg-0c10c4fe23f5e5361
+      - sg-09295097b2a41b59d
+      - from_cfn: MyUserDBAccessSecurityGroup1
+      - from_cfn: MyUserDBAccessSecurityGroup2
 # Optional fields for more advanced use-cases.
 #
 variables:                    # Pass environment variables as key value pairs.
--- a/internal/pkg/deploy/cloudformation/stack/testdata/workloads/worker-manifest.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/workloads/worker-manifest.yml
@@ -72,6 +72,12 @@ subscribe:
       queue:
         fifo: false
 
+network:
+  vpc:
+    security_groups:
+      - sg-0c10c4fe23f5e5361
+      - from_cfn: MyUserDBAccessSecurityGroup1
+
 # Optional fields for more advanced use-cases.
 #
 #variables:                    # Pass environment variables as key value pairs.
--- a/internal/pkg/deploy/cloudformation/stack/testdata/workloads/worker-manifest.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/workloads/worker-manifest.yml
@@ -72,12 +72,6 @@ subscribe:
       queue:
         fifo: false
 
-network:
-  vpc:
-    security_groups:
-      - sg-0c10c4fe23f5e5361
-      - from_cfn: MyUserDBAccessSecurityGroup1
-
 # Optional fields for more advanced use-cases.
 #
 #variables:                    # Pass environment variables as key value pairs.
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -750,16 +750,16 @@ func convertNetworkConfig(network manifest.NetworkConfig) template.NetworkOpts {
 		AssignPublicIP: template.EnablePublicIP,
 		SubnetsType:    template.PublicSubnetsPlacement,
 	}
-	mftSecurityGroups := network.VPC.SecurityGroups.GetIDs()
-	var tplSecurityGroups []template.SecurityGroup
-	for _, sg := range mftSecurityGroups {
+	inSGs := network.VPC.SecurityGroups.GetIDs()
+	outSGs := make([]template.SecurityGroup, len(inSGs))
+	for i, sg := range inSGs {
 		if sg.Plain != nil {
-			tplSecurityGroups = append(tplSecurityGroups, template.PlainSecurityGroup(aws.StringValue(sg.Plain)))
+			outSGs[i] = template.PlainSecurityGroup(aws.StringValue(sg.Plain))
 		} else {
-			tplSecurityGroups = append(tplSecurityGroups, template.ImportedSecurityGroup(aws.StringValue(sg.FromCFN.Name)))
+			outSGs[i] = template.ImportedSecurityGroup(aws.StringValue(sg.FromCFN.Name))
 		}
 	}
-	opts.SecurityGroups = tplSecurityGroups
+	opts.SecurityGroups = outSGs
 	opts.DenyDefaultSecurityGroup = network.VPC.SecurityGroups.IsDefaultSecurityGroupDenied()
 
 	placement := network.VPC.Placement
--- a/internal/pkg/deploy/cloudformation/stack/transformers.go
+++ b/internal/pkg/deploy/cloudformation/stack/transformers.go
@@ -812,20 +812,22 @@ func convertEntryPoint(entrypoint manifest.EntryPointOverride) ([]string, error)
 	return out, nil
 }
 
-func convertDeploymentConfig(deploymentConfig manifest.DeploymentConfiguration) template.DeploymentConfigurationOpts {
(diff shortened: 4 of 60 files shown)


internal/pkg/deploy/cloudformation/stack/worker_svc_test.go
```diff
@@ -165,7 +165,6 @@ Outputs:
 						Network: template.NetworkOpts{
 							AssignPublicIP: template.DisablePublicIP,
 							SubnetsType:    template.PrivateSubnetsPlacement,
-							SecurityGroups: []string{"sg-1234"},
```
paragbhingre, line 168: Why are we removing this from all the tests? can't we add `[]stringOrFromCFN{}`?

internal/pkg/template/workload.go
```diff
@@ -540,8 +540,9 @@ type DeadLetterQueue struct {
 
 // NetworkOpts holds AWS networking configuration for the workloads.
 type NetworkOpts struct {
-	SecurityGroups []string
-	AssignPublicIP string
+	PlainSecurityGroups    []string
+	ImportedSecurityGroups []string
```
Lou1415926: what do you think of doing something similar with `Variable`? https://github.com/aws/copilot-cli/blob/e5c44581e69659a874f5d7ebadd7119dc551309e/internal/pkg/template/workload.go#L263

For example,

```go
type NetworkOpts struct {
  SecurityGroups []SecurityGroup
}

type SecurityGroup interface {
  importable
  Value()
}

type PlainSecurityGroup string

func (sg PlainSecurityGroup) RequiresImport() bool {
  return false
}

type ImportedSecurityGroup string

func (sg ImportedSecurityGroup) RequiresImport() bool {
  return true
}
```

internal/pkg/deploy/cloudformation/stack/worker_svc_test.go
KollaAdithya: I think stringOrFromCFN{} can not be exported from manifest package. So that is the reason for removing this test case.

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
```diff
@@ -83,7 +83,10 @@ NetworkConfiguration:
       - Fn::ImportValue: !Sub '${AppName}-${EnvName}-EnvironmentSecurityGroup'
       {{- end}}
       {{- range $sg := .Network.SecurityGroups}}
-      - {{$sg}}
+      {{- if not $sg.RequiresImport}}
+      - {{$sg.Value}}
+      {{- else}}
+      - Fn::ImportValue: {{$sg.Value}} {{- end}}
```
Lou1415926, line 99: yay this looks nice!

internal/pkg/template/workload.go
```diff
@@ -540,14 +540,54 @@ type DeadLetterQueue struct {
...
-	SecurityGroups []string
+	SecurityGroups []SecurityGroup
 	AssignPublicIP string
 	// SubnetsType and SubnetIDs are mutually exclusive. They won't be set together.
 	SubnetsType              string
 	SubnetIDs                []string
 	DenyDefaultSecurityGroup bool
 }
 
+// SecurityGroup represents the value of an additional security IDs associated with your tasks.
```
```suggestion
// SecurityGroup represents the ID of an additional security group associated with the tasks.
```

internal/pkg/template/template_integration_test.go
```diff
@@ -419,10 +419,13 @@ func TestTemplate_ParseNetwork(t *testing.T) {
...
 				AssignPublicIP: "DISABLED",
 				SubnetsType:    "PrivateSubnets",
-				SecurityGroups: []string{
+				PlainSecurityGroups: []string{
 					"sg-1bcf1d5b",
 					"sg-asdasdas",
 				},
+				ImportedSecurityGroups:[]string{
+					"mydb-sg001"
+				}
```
Lou1415926: Now that we've changed the implementations, I suppose these need to be changed as well 👀 ?

internal/pkg/deploy/cloudformation/stack/testdata/workloads/worker-manifest.yml
```diff
@@ -72,6 +72,12 @@ subscribe:
       queue:
         fifo: false
 
+network:
+  vpc:
+    security_groups:
+      - sg-0c10c4fe23f5e5361
+      - from_cfn: MyUserDBAccessSecurityGroup1
```
Lou1415926: nice!

internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-manifest.yml
```diff
@@ -40,6 +40,13 @@ publish:
     - name: givesdogs
     - name: mytopic
       fifo: true
+network:
```
Lou1415926: we can add this chunk to 
```yaml
environments:
  test:
     network:...
```

The idea is that it's sufficient to test this functionality in just one environment. 

Also, we probably only need one test case in one of `worker`, `svc`, not both - since they are both under the hood ECS services.

KollaAdithya: Just removed the test cases for `worker and backend` service. Now we have only one test case for `svc` in just `test` environment to test this functionality.

internal/pkg/template/template_integration_test.go
KollaAdithya: updated the integration test as per the new implementation.

internal/pkg/manifest/validate.go
```diff
@@ -1651,13 +1651,23 @@ func (r OverrideRule) validate() error {
...
 func (v Variable) validate() error {
 	if err := v.FromCFN.validate(); err != nil {
 		return fmt.Errorf(`validate "from_cfn": %w`, err)
 	}
 	return nil
 }
 
+// validate returns nil if stringorFromCFN is configured correctly.
+func (s stringOrFromCFN) validate() error {
+	if s.FromCFN.isEmpty() && s.Plain == nil {
```
Lou1415926: I see this line is saying "if stringOrFromCFN is empty, return nil" - this makes sense to me! However, can we instead implement
```go
func (s stringOrFromCFN) isEmpty() bool {
  return s.FromCFN.isEmpty() && s.Plain == nil
}
```
and call it here:
```go
func (s stringOrFromCFN) validate() error {
  if s.isEmpty() {
     return nil  
  }
  ...
}
```
This way, it's clear what the logic is. In addition, `s.isEmpty()` can potentially be reused somewhere else - *if* in the future we have more fields in `stringOrFromCFN`, we'd only need too update `isEmpty`, instead of every `if s.FromCFN.isEmpty() && s.Plain == nil`

internal/pkg/manifest/workload_test.go
```diff
@@ -998,16 +1056,30 @@ network:
...
-						IDs:            []string{"sg-1234", "sg-4567"},
+						IDs: []stringOrFromCFN{
+							{
+								Plain: aws.String("sg-1234"),
+							},
+							{
+								Plain: aws.String("sg-4567"),
+							},
+							{
+								Plain: nil,
```
Lou1415926: nit:
```suggestion
```

internal/pkg/manifest/workload_test.go
```diff
@@ -1047,8 +1134,15 @@ network:
...
-							DenyDefault:    nil,
+							SecurityGroups: []stringOrFromCFN{
+								{
+									Plain: aws.String("sg-1234"),
+								},
+								{
+									Plain: aws.String("sg-4567"),
+								},
+							},
+							DenyDefault: nil,
```
Lou1415926: nit:
```suggestion
```

Lou1415926 (approved):

internal/pkg/template/workload.go
```diff
@@ -540,14 +540,54 @@ type DeadLetterQueue struct {
...
 }
 
+// SecurityGroup represents the ID of an additional security group associated with the tasks.
+type SecurityGroup interface {
+	importable
+	Value() string
+}
+
+// PlainSecurityGroup returns a SecurityGroup that is a plain string value.
+func PlainSecurityGroup(value string) Variable {
```
iamhopaul123: `plainSecurityGroup` implementing `Variable` interface which "represents the value of an environment variable" looks weird to me...

iamhopaul123: I feel like maybe `Variable` is not a generic enough name for what methods (`RequiresImport()`, `Value()`) the interface defines. It sounds more like a concrete struct name.

KollaAdithya: Good catch! it is a mistake from my end the function is supposed to be
`func PlainSecurityGroup(value string) SecurityGroup` which is implementing `SecurityGroup` interface.

SecurityGroup interface looks something like this
```
type SecurityGroup interface {
	importable
	Value() string
}
```
 I will make `plainSecurityGroup` and `importedSecurityGroup` public and remove these setter functions
`1. func PlainSecurityGroup(value string) SecurityGroup` 
`2. func ImportedSecurityGroup(name string) SecurityGroup`

iamhopaul123: Hmm it doesn't sound necessary tho because both `SecurityGroup` and `Variable` interface will be used only once (aka not sharable and generic enough to justify their existence). I would suggest we **either**

1. [Recommended] Rename `Variable` interface to a more generic name (e.g., `Importable` or a better name) reflecting what methods it contains
2. Remove interfaces and use structs only

Lou1415926: Perhaps we can do:
```go
type importableValue interface {
  importable
  Value() string
}

type Variable importableValue

type SecurityGroup importableValue
```

KollaAdithya: Done.

internal/pkg/deploy/cloudformation/stack/testdata/workloads/job-manifest.yml
```diff
@@ -58,6 +58,14 @@ sidecars:
     labels:
       com.amazonaws.ecs.copilot.sidecars.nginx.description: tricky
 
+network:
+  vpc:
+    security_groups:
+      - sg-0c10c4fe23f5e5361
+      - sg-09295097b2a41b59d
+      - from_cfn: MyUserDBAccessSecurityGroup1
+      - from_cfn: MyUserDBAccessSecurityGroup2
```
efekarakus, line 67: so kewl 🤩

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -750,7 +750,16 @@ func convertNetworkConfig(network manifest.NetworkConfig) template.NetworkOpts {
 		AssignPublicIP: template.EnablePublicIP,
 		SubnetsType:    template.PublicSubnetsPlacement,
 	}
-	opts.SecurityGroups = network.VPC.SecurityGroups.GetIDs()
+	mftSecurityGroups := network.VPC.SecurityGroups.GetIDs()
+	var tplSecurityGroups []template.SecurityGroup
+	for _, sg := range mftSecurityGroups {
```
efekarakus: nit: these are just recommendations, not blockers!

1. For functions that are of the form `transformXToY` I like naming the variables `in` and `out`, I feel like it makes it pretty clear which one is the transformed version.
2. If the size of the slice is known ahead of time, consider using `make([]T, len(in))` for performance reasons.

```suggestion
	inSGs := network.VPC.SecurityGroups.GetIDs()
	outSGs :=  make([]template.SecurityGroup, len(inSGs))
	for i, sg := range inSGs {
		outSGs[i] = //...
```

internal/pkg/manifest/workload.go
```diff
@@ -680,7 +680,7 @@ func (s *SecurityGroupsIDsOrConfig) UnmarshalYAML(value *yaml.Node) error {
 
 // GetIDs returns security groups from SecurityGroupsIDsOrConfig that are attached to task.
 // nil is returned if no security groups are specified.
-func (s *SecurityGroupsIDsOrConfig) GetIDs() []string {
+func (s *SecurityGroupsIDsOrConfig) GetIDs() []stringOrFromCFN {
```
efekarakus, line 683: nit: it's a bit unusual to return a private type from a public method 😅 but I think it's fine since we don't expect `manifest` clients to initialize this type.

internal/pkg/template/template_integration_test.go
```diff
@@ -435,15 +436,16 @@ func TestTemplate_ParseNetwork(t *testing.T) {
...
 		"should render AWS VPC configuration without default environment security group": {
 			input: template.NetworkOpts{
 				AssignPublicIP: "DISABLED",
 				SubnetsType:    "PrivateSubnets",
-				SecurityGroups: []string{
-					"sg-1bcf1d5b",
-					"sg-asdasdas",
+				SecurityGroups:[]template.SecurityGroup{
+					template.plainSecurityGroup("sg-1bcf1d5b"),
+					template.plainSecurityGroup("sg-asdasdas"),
```
efekarakus: wat 🤯 am I dreaming but how does this work, isn't this type private?? I would have expected
```suggestion
					template.PlainSecurityGroup("sg-1bcf1d5b"),
					template.PlainSecurityGroup("sg-asdasdas"),
```

efekarakus (commented):
Looks great!

iamhopaul123 (approved):

efekarakus (approved):

PR #4315, opened 2023-01-03
chore: implement `from_cfn` for secrets to import value from other CFN stacks 

KollaAdithya:
Related #4220 


internal/pkg/template/workload.go
```diff
@@ -302,54 +302,119 @@ func (v importedEnvVar) Value() string {
 // A Secret represents an SSM or SecretsManager secret that can be rendered in CloudFormation.
 type Secret interface {
 	RequiresSub() bool
+	RequiresImport() bool
```
Lou1415926: nit: maybe we can just embed the `importable` interface here
```suggestion
	importable
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -1039,13 +1039,24 @@ func convertSecrets(secrets map[string]manifest.Secret) map[string]template.Secr
 	if len(secrets) == 0 {
 		return nil
 	}
-	m := make(map[string]template.Secret)
+	m := make(map[string]template.Secret, len(secrets))
```
Lou1415926, line 1054: nice!

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -1039,13 +1039,24 @@ func convertSecrets(secrets map[string]manifest.Secret) map[string]template.Secr
...
 	for name, mftSecret := range secrets {
-		var tplSecret template.Secret = template.SecretFromSSMOrARN(mftSecret.Value())
+		var tplSecret template.Secret
 		if mftSecret.IsSecretsManagerName() {
-			tplSecret = template.SecretFromSecretsManager(mftSecret.Value())
+			if mftSecret.SecretsManagerRequiresImport() {
```
Lou1415926: It seems like a weird case for me when both `mftSecret.IsSecretManagerName()` and `mftSecret.SecretsManagerRequiresImport()` return `true`. What does the manifest look like? Was this what you had in mind ⬇️ ?
```yaml
secrets:
  MYSECRET:
    secretsmanager:
      from_cfn: MYSTACK-secretsname
```

KollaAdithya: That is the exactly the case that I taught where the secret is from `SecretsManager` and also it requires an import from other CloudFormation Stack.

Lou1415926: This is not a super useful case I think! Here is my reasonings. If user create a secret in a stack (which could either be an env addon, or just an arbitrary stack that they manage on their own),they most likely export the secret like this:
```yaml
Outputs:
  MySecret:
     Value: !Ref MySecret
     Export:
       Name: !Sub ${AWS::StackName}-MySecret
```

The [`!Ref`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-secretsmanager-secret.html#aws-resource-secretsmanager-secret-return-values) is going to have the ARN of the secret, not the name.  From the same [doc](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-secretsmanager-secret.html#aws-resource-secretsmanager-secret-return-values), we know that for secrets, you can only use `!Ref` which returns ARN, i.e. you can't `!GetAtt MySecret.Name`. 

Therefore, chances are, people don't export a secret's name, but a secret ARN. 

On the other hand, `secretsmanager` accepts the **name** of a secret, not the  **arn**. Therefore, `secretsmanager.from_cfn` isn't super useful.

I'd say we only need to take care of the following cases:
```yaml
secrets:
  MYSECRETIMPORTED:
    from_cfn: MYSTACK-secretsExportName
  MYSECRETPLAIN: 'arn'
  MYSECRETSMANAGER:
    secretsmanager: 'secretname'
  
```

KollaAdithya: As discussed removed the implementation of `from_cfn` for `secretsmanger` that takes the name of a secret.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -1039,13 +1039,19 @@ func convertSecrets(secrets map[string]manifest.Secret) map[string]template.Secr
...
+		var tplSecret template.Secret
 		if mftSecret.IsSecretsManagerName() {
 			tplSecret = template.SecretFromSecretsManager(mftSecret.Value())
+			m[name] = tplSecret
+		} else if mftSecret.SSMRequiresImport() {
+			tplSecret = template.ImportedSecretFromSSMOrARN(mftSecret.Value())
+			m[name] = tplSecret
+		} else {
+			tplSecret = template.PlainSecretFromSSMOrARN(mftSecret.Value())
+			m[name] = tplSecret
```
```suggestion
		var tplSecret template.Secret
		if mftSecret.IsSecretsManagerName() {
			tplSecret = template.SecretFromSecretsManager(mftSecret.Value())
		} else if mftSecret.SSMRequiresImport() {
			tplSecret = template.ImportedSecretFromSSMOrARN(mftSecret.Value())
		} else {
			tplSecret = template.PlainSecretFromSSMOrARN(mftSecret.Value())
		}
		m[name] = tplSecret
```

internal/pkg/template/templates/workloads/partials/cf/secrets.yml
```diff
@@ -1,8 +1,14 @@
...
 {{- range $name, $secret := .Secrets}}
+{{if not $secret.RequiresImport}}
 - Name: {{$name}}
-  ValueFrom: {{if not $secret.RequiresSub }}{{$secret.ValueFrom}}{{- else}} !Sub 'arn:${AWS::Partition}:{{$secret.Service}}:${AWS::Region}:${AWS::AccountId}:{{$secret.ValueFrom}}' {{- end }}
+  ValueFrom: {{if not $secret.RequiresSub }} {{$secret.ValueFrom}} {{else}} !Sub 'arn:${AWS::Partition}:{{$secret.Service}}:${AWS::Region}:${AWS::AccountId}:{{$secret.ValueFrom}}' {{end}}
+{{- else}}
+- Name: {{$name}}
+  ValueFrom:
+    Fn::ImportValue: {{ quote $secret.ValueFrom -}}
+{{- end}}
```
Lou1415926: nit: can we switch the order and remove the `not` in `{{if not $secret.RequiresImport}}`? After that, maybe we can try moving `{{if $secret.RequiresImport}}` to after `Name: {{$name}}` so that `Name: {{$name}}` doesn't appear twice.

internal/pkg/template/workload.go
```diff
@@ -302,27 +302,60 @@ func (v importedEnvVar) Value() string {
...
+	return false
+}
+
+// ValueFrom returns the plain string value of the secret.
+func (s plainSSMOrSecretARN) ValueFrom() string {
+	return s.value
+}
+
+// PlainSecretFromSSMOrARN returns a Secret that refers to an SSM parameter or a secret ARN.
+func PlainSecretFromSSMOrARN(value string) plainSSMOrSecretARN {
```
```suggestion
func SecretFromPlainSSMOrARN(value string) plainSSMOrSecretARN {
```

internal/pkg/template/workload.go
```diff
@@ -302,27 +302,60 @@ func (v importedEnvVar) Value() string {
...
+// ValueFrom returns the name of the import value of the Secret.
+func (s importedSSMorSecretARN) ValueFrom() string {
 	return s.value
 }
 
-// SecretFromSSMOrARN returns a Secret that refers to an SSM parameter or a secret ARN.
-func SecretFromSSMOrARN(value string) ssmOrSecretARN {
-	return ssmOrSecretARN{
+// ImportedSecretFromSSMOrARN returns a Secret that refers to imported name of SSM parameter or a secret ARN.
+func ImportedSecretFromSSMOrARN(value string) importedSSMorSecretARN {
```
Lou1415926:
```suggestion
func SecretFromImportedSSMOrARN(value string) importedSSMorSecretARN {
```

internal/pkg/template/workload_test.go
```diff
@@ -265,18 +265,38 @@ func TestHTTPTargetContainer_IsHTTPS(t *testing.T) {
...
-func TestSsmOrSecretARN_RequiresSub(t *testing.T) {
-	require.False(t, ssmOrSecretARN{}.RequiresSub(), "SSM Parameter Store or secret ARNs do not require !Sub")
+func TestPlainSSMOrSecretARN_RequiresSub(t *testing.T) {
+	require.False(t, plainSSMOrSecretARN{}.RequiresSub(), "SSM Parameter Store or secret ARNs do not require !Sub")
 }
 
-func TestSsmOrSecretARN_ValueFrom(t *testing.T) {
-	require.Equal(t, "/github/token", SecretFromSSMOrARN("/github/token").ValueFrom())
+func TestPlainSSMOrSecretARN_RequiresImport(t *testing.T) {
+	require.False(t, plainSSMOrSecretARN{}.RequiresImport(), "secret name that refers to a value imported from CloudFormation Stack")
```
Lou1415926: maybe we need a better description of the test case:
```suggestion
	require.False(t, plainSSMOrSecretARN{}.RequiresImport(), "plain SSM Parameter Store or secret ARNs do not require !ImportValue")
```
and similar for the other test cases.

internal/pkg/manifest/workload_ecs_test.go
```diff
@@ -218,13 +218,28 @@ func TestSecret_UnmarshalYAML(t *testing.T) {
...
 		},
-		"should be able to unmarshal an SSM parameter name": {
-			in:     "/github/token",
-			wanted: Secret{from: aws.String("/github/token")},
+		"should be able to unmarshal an plain SSM parameter name": {
+			in: "/github/token",
+			wanted: Secret{from: stringOrFromCFN{
+				Plain: aws.String("/github/token"),
+			},
+			},
```
Lou1415926: Can we fix the `{` and `}` indentations?

```suggestion
			wanted: Secret{
			    from: stringOrFromCFN{
				Plain: aws.String("/github/token"),
			     },
			},
```

internal/pkg/manifest/workload_ecs_test.go
```diff
@@ -218,13 +218,28 @@ func TestSecret_UnmarshalYAML(t *testing.T) {
...
-		"should be able to unmarshal a SecretsManager ARN": {
-			in:     "arn:aws:secretsmanager:us-west-2:111122223333:secret:aes128-1a2b3c",
-			wanted: Secret{from: aws.String("arn:aws:secretsmanager:us-west-2:111122223333:secret:aes128-1a2b3c")},
+		"should be able to unmarshal an imported SSM parameter name from other cloudformation stack": {
+			in: `from_cfn: "mygithubtoken"`,
+			wanted: Secret{from: stringOrFromCFN{
+				FromCFN: fromCFN{
+					Name: aws.String("mygithubtoken"),
+				},
+			},
```
Lou1415926: nit: to make it more realistic and show how this field is typically used
```suggestion
			in: `from_cfn: "stack-SSMGHTokenName"`,
			wanted: Secret{from: stringOrFromCFN{
				FromCFN: fromCFN{
					Name: aws.String("stack-SSMGHTokenName"),
				},
			},
```

internal/pkg/manifest/workload_ecs_test.go
```diff
@@ -257,12 +272,21 @@ func TestSecret_IsSecretsManagerName(t *testing.T) {
...
+			in: Secret{from: stringOrFromCFN{
+				Plain: aws.String("/github/token"),
+			},
+			},
 		},
 		"should return true if the secret refers to a SecretsManager secret name": {
 			in:     Secret{fromSecretsManager: secretsManagerSecret{Name: aws.String("aes128-1a2b3c")}},
 			wanted: true,
 		},
+		"should return false if the secret is imported": {
```
Lou1415926, line 287: noice!

internal/pkg/manifest/workload_ecs_test.go
```diff
@@ -257,12 +272,21 @@ func TestSecret_IsSecretsManagerName(t *testing.T) {
...
+			in: Secret{from: stringOrFromCFN{
+				FromCFN: fromCFN{aws.String("mygithubtoken")},
+			}},
+			wanted: false,
```
Lou1415926: nit:
```suggestion
```

internal/pkg/manifest/workload_ecs_test.go
```diff
@@ -271,15 +295,55 @@ func TestSecret_IsSecretsManagerName(t *testing.T) {
 	}
 }
 
+func TestSSMRequiresImport(t *testing.T) {
+	testCases := map[string]struct {
+		in     Secret
+		wanted bool
+	}{
+		"should return false if secret is plain": {
```
Lou1415926, line 307: Maybe one more test case for secret from `secretsmanager` name?

internal/pkg/manifest/workload_ecs.go
```diff
@@ -210,12 +210,20 @@ func (s *Secret) IsSecretsManagerName() bool {
 	return !s.fromSecretsManager.IsEmpty()
 }
 
+// SSMRequiresImport returns true if the SSM parameter name or secret ARN value is imported from CloudFormation stack.
+func (s *Secret) SSMRequiresImport() bool {
```
Lou1415926: nit: I feel like the `SSM` prefix is unnecessary and potentially misleading (since it works for secrets manager ARN as well)

```suggestion
// RequiresImport returns true if the SSM parameter name or secret ARN value is imported from CloudFormation stack.
func (s *Secret) RequiresImport() bool {
```

internal/pkg/manifest/workload_ecs.go
```diff
@@ -210,12 +210,20 @@ func (s *Secret) IsSecretsManagerName() bool {
...
+}
+
 // Value returns the secret value provided by clients.
 func (s *Secret) Value() string {
 	if !s.fromSecretsManager.IsEmpty() {
 		return aws.StringValue(s.fromSecretsManager.Name)
+	} else if !s.SSMRequiresImport() {
+		return aws.StringValue(s.from.Plain)
+	} else {
+		return aws.StringValue(s.from.FromCFN.Name)
```
Lou1415926: nit: can we flip the boolean so that the logic is more straightforward 
```suggestion
	} else if s.SSMRequiresImport() {
	        return aws.StringValue(s.from.FromCFN.Name)
	} else {
		return aws.StringValue(s.from.Plain)
```

internal/pkg/manifest/transform_test.go
```diff
@@ -1264,10 +1266,35 @@ func TestSecretTransformer_Transformer(t *testing.T) {
...
+				}
 			},
 			wanted: func(s *Secret) {
-				s.from = aws.String("/github/token")
+				s.from = stringOrFromCFN{
+					Plain: aws.String("/github/token"),
+				}
+			},
+		},
+		`"secretsmanager" set to empty when overriding with imported "from"`: {
```
Lou1415926, line 1279: noice!!

internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-manifest.yml
```diff
@@ -60,6 +60,11 @@ environments:
         path: /
         grace_period: 30s
       deregistration_delay: 30s
+    secrets:
```
Lou1415926: Can we add the secret to a sidecar and a logging's [secret options](https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#logging-secretOptions) just to make sure it works well for those two places as well?

internal/pkg/template/templates/workloads/partials/cf/secrets.yml
KollaAdithya: Changes are done!

internal/pkg/manifest/workload_ecs_test.go
KollaAdithya: added one more test case for secretsmanger

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -1050,11 +1050,15 @@ func convertSecrets(secrets map[string]manifest.Secret) map[string]template.Secr
 	if len(secrets) == 0 {
 		return nil
 	}
-	m := make(map[string]template.Secret)
+	m := make(map[string]template.Secret, len(secrets))
```
iamhopaul123, line 1054: Maybe add a comment to this `convertSecrets` function? Because it seems not that clear what the key is for both input and output.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -1050,11 +1050,15 @@ func convertSecrets(secrets map[string]manifest.Secret) map[string]template.Secr
...
 	for name, mftSecret := range secrets {
-		var tplSecret template.Secret = template.SecretFromSSMOrARN(mftSecret.Value())
+		var tplSecret template.Secret
 		if mftSecret.IsSecretsManagerName() {
```
iamhopaul123: Would it be better to use `case` here?

internal/pkg/deploy/cloudformation/stack/transformers.go
iamhopaul123: I agreed it makes little sense to have `from_cfn` for `secretsmanager` given the fact that "secretsmanager only accepts the name of a secret, not the arn". I wonder if it is a problem we need to keep in mind since ideally both name and arn should work. It opens a door for using any cross-account secret I guess and leverages the benefit for having `from_cfn` 🤔

iamhopaul123 (approved):

internal/pkg/manifest/workload_ecs.go
```diff
@@ -183,7 +183,7 @@ func (t TaskConfig) IsARM() bool {
 
 // Secret represents an identifier for sensitive data stored in either SSM or SecretsManager.
 type Secret struct {
-	from               *string              // SSM Parameter name or ARN to a secret.
+	from               stringOrFromCFN      // SSM Parameter name or ARN to a secret.
```
paragbhingre:
```suggestion
	from               stringOrFromCFN      // SSM Parameter name or ARN to a secret or secret imported from another stack.
```

internal/pkg/manifest/workload_ecs.go
```diff
@@ -211,12 +211,20 @@ func (s *Secret) IsSecretsManagerName() bool {
...
+
 // Value returns the secret value provided by clients.
 func (s *Secret) Value() string {
 	if !s.fromSecretsManager.IsEmpty() {
 		return aws.StringValue(s.fromSecretsManager.Name)
+	} else if s.RequiresImport() {
+		return aws.StringValue(s.from.FromCFN.Name)
+	} else {
+		return aws.StringValue(s.from.Plain)
 	}
```
paragbhingre: Tiny nit: 
```suggestion
	} else if s.RequiresImport() {
		return aws.StringValue(s.from.FromCFN.Name)
	}
	return aws.StringValue(s.from.Plain)
	
```

internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-manifest.yml
```diff
@@ -68,6 +68,12 @@ environments:
...
```
Lou1415926: qq: do we add this test to both `staging` and `prod` out of some specific concerns?

internal/pkg/manifest/workload_ecs.go
Lou1415926: +1 to Parag! A minor suggestion that I have is 
```
...// SSM Parameter name, plain secret ARN, or secret ARN imported from another stack.
```

internal/pkg/manifest/workload_ecs.go
```diff
@@ -211,12 +211,20 @@ func (s *Secret) IsSecretsManagerName() bool {
...
+}
+
 // Value returns the secret value provided by clients.
 func (s *Secret) Value() string {
 	if !s.fromSecretsManager.IsEmpty() {
 		return aws.StringValue(s.fromSecretsManager.Name)
+	} else if s.RequiresImport() {
+		return aws.StringValue(s.from.FromCFN.Name)
+	} else {
+		return aws.StringValue(s.from.Plain)
```
Lou1415926:
```suggestion
	} 
	if s.RequiresImport() {
		return aws.StringValue(s.from.FromCFN.Name)
	}
	return aws.StringValue(s.from.Plain)
```

internal/pkg/template/templates/workloads/partials/cf/secrets.yml
```diff
@@ -2,7 +2,12 @@
 Secrets:
 {{- range $name, $secret := .Secrets}}
 - Name: {{$name}}
-  ValueFrom: {{if not $secret.RequiresSub }}{{$secret.ValueFrom}}{{- else}} !Sub 'arn:${AWS::Partition}:{{$secret.Service}}:${AWS::Region}:${AWS::AccountId}:{{$secret.ValueFrom}}' {{- end }}
+{{- if $secret.RequiresImport}}
+  ValueFrom:
+    Fn::ImportValue: {{ quote $secret.ValueFrom -}}
```
paragbhingre: Do we need this before delimiters for some reason here? As far as I know they are used to truncate the white spaces. We can remove them unless they are doing some operation here. Similarly for other templates as well 
```suggestion
    Fn::ImportValue: {{ quote $secret.ValueFrom }}
```

internal/pkg/manifest/workload_ecs_test.go
```diff
@@ -218,13 +217,31 @@ func TestSecret_UnmarshalYAML(t *testing.T) {
 			in:        "key: value",
 			wantedErr: errors.New(`cannot marshal "secret" field to a string or "secretsmanager" object`),
 		},
-		"should be able to unmarshal an SSM parameter name": {
-			in:     "/github/token",
-			wanted: Secret{from: aws.String("/github/token")},
+		"should be able to unmarshal an plain SSM parameter name": {
```
Lou1415926: nit:
```suggestion
		"should be able to unmarshal a plain SSM parameter name": {
```

Lou1415926 (approved):
LGTM 🚢  Feel free to remove the label when you are ready!

internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-manifest.yml
KollaAdithya: When the manifest contains 
```yaml
logging:
  secrets:
    <key>: <value>
```
This will expose Secrets to the sidecar container, which I am testing in `prod` environment.

When the manifest contains
```yaml
secrets:
  <key>:  <value>
``` 
Secrets will be securely passed to your service as environment variables, which I am testing in `staging` environment.
Yes you are correct. I can test both of these in the same environment.

PR #4322, opened 2023-01-04
feat: use existing s3 bucket for cf

iamhopaul123:
Part of #3701

### DIFF
--- a/cf-custom-resources/lib/unique-json-values.js
+++ b/cf-custom-resources/lib/unique-json-values.js
@@ -93,7 +93,7 @@ const deadlineExpired = function () {
  *  "svc2": ["example.com"]
  * }
  *
- * The input event.ResourceProperties.AdditionalAlias is a string (e.g., "foobar.com").
+ * The input event.ResourceProperties.AdditionalStrings is a string slice (e.g., ["foobar.com"]).
  *
  * This function returns a sorted list of unique values found in the map and the additional string.
  * For this example, UniqueValues would be:
@@ -120,8 +120,12 @@ exports.handler = async function (event, context) {
           )
         );
         const unique = new Set(Object.values(filteredAliasesForService).flat());
-        if (event.ResourceProperties.AdditionalAlias) {
-          unique.add(event.ResourceProperties.AdditionalAlias);
+        if (event.ResourceProperties.AdditionalStrings) {
+          event.ResourceProperties.AdditionalStrings.forEach((element) => {
+            if (element) {
+              unique.add(element);
+            }
+          });
         }
         responseData.UniqueValues = Array.from(unique).sort();
         break;
--- a/cf-custom-resources/lib/unique-json-values.js
+++ b/cf-custom-resources/lib/unique-json-values.js
@@ -120,10 +120,7 @@ exports.handler = async function (event, context) {
           )
         );
         const unique = new Set(Object.values(filteredAliasesForService).flat());
-        if (
-          event.ResourceProperties.AdditionalAlias &&
-          event.ResourceProperties.AdditionalAlias !== ""
-        ) {
+        if (event.ResourceProperties.AdditionalAlias) {
           unique.add(event.ResourceProperties.AdditionalAlias);
         }
         responseData.UniqueValues = Array.from(unique).sort();
--- a/cf-custom-resources/lib/unique-json-values.js
+++ b/cf-custom-resources/lib/unique-json-values.js
@@ -67,7 +67,7 @@ const deadlineExpired = function () {
   return new Promise((resolve, reject) => {
     setTimeout(
       reject,
-      (9 * 60 * 1000) + (30 * 1000), /* 9.5 minutes */
+      9 * 60 * 1000 + 30 * 1000 /* 9.5 minutes */,
       new Error("Lambda took longer than 9.5 minutes")
     );
   });
@@ -75,7 +75,7 @@ const deadlineExpired = function () {
 
 /**
  * Main handler, invoked by Lambda
- * 
+ *
  * The input event.ResourceProperties.Aliases is a map of service name to
  * it's alises. For example, it might look like this:
  * {
@@ -83,7 +83,7 @@ const deadlineExpired = function () {
  *  "svc2": ["example.com"]
  *  "svc3": ["svc3.com"]
  * }
- * 
+ *
  * The input event.ResourceProperties.FilterFor is a comma delimated list
  * of keys in event.ResourceProperties.Aliases to filter for.
  * Taking the above map, and event.ResourceProperties.FilterFor = "svc1,svc2"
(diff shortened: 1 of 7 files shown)


cf-custom-resources/lib/unique-json-values.js
```diff
@@ -92,43 +92,62 @@ const deadlineExpired = function () {
...
-            ([key]) => filterFor.has(key)
+          Object.entries(aliasesForService).filter(([key]) =>
+            filterFor.has(key)
           )
         );
         const unique = new Set(Object.values(filteredAliasesForService).flat());
+        if (
+          event.ResourceProperties.AdditionalAlias &&
+          event.ResourceProperties.AdditionalAlias !== ""
+        ) {
```
Lou1415926: would this ⬇️  be enough?
```suggestion
        if (event.ResourceProperties.AdditionalAlias ) {
```

cf-custom-resources/test/unique-json-values-test.js
```diff
@@ -80,16 +80,19 @@ describe("Unique Aliases", () => {
       test(name, () => {
         const request = nock(responseURL)
           .put("/", (body) => {
-            return body.Status === "SUCCESS" &&
+            return (
+              body.Status === "SUCCESS" &&
               body.PhysicalResourceId === "mockID" &&
-              JSON.stringify(body.Data.UniqueValues) === JSON.stringify(expectedOutput);
+              JSON.stringify(body.Data.UniqueValues) ===
```
Lou1415926, line 86: (not that I'm asking boss to update the PR) the original line was under 120 chars though 🤔

cf-custom-resources/test/unique-json-values-test.js
```diff
@@ -108,86 +111,134 @@ describe("Unique Aliases", () => {
...
+        svc1: ["svc1.com"],
+        svc2: ["svc2.com"],
+      }),
+      FilterFor: "svc1,svc2",
+    },
+    ["svc1.com", "svc2.com"]
+  );
+
+  aliasTest(
+    "two services, one with multiple common aliases",
```
Lou1415926:
```suggestion
    "two services, one with multiple aliases",
```

cf-custom-resources/test/unique-json-values-test.js
```diff
@@ -108,86 +111,134 @@ describe("Unique Aliases", () => {
...
+      "four.lbws.com",
+      "lbws.com",
+      "one.lbws.com",
+      "three.lbws.com",
+      "two.lbws.com",
+    ]
+  );
+
+  aliasTest(
+    "bunch of services with single alias, some FilterFor services don't exist, and additional alias",
```
Lou1415926: nit: should we have a test for  "some FilterFor services don't exist" and a different test for "additional alias"?

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -44,9 +44,23 @@ UniqueAliasesAction:
...
+    {{- end}}
     FilterFor: !Ref ALBWorkloads
 {{- end}}{{/* endif or .CDNConfig.ImportedCertificate .DelegateDNS */}}
 
+{{- if .CDNConfig.Static}}
+CloudFrontOriginAccessControl:
+  Type: AWS::CloudFront::OriginAccessControl
+  Properties: 
+    OriginAccessControlConfig: 
+        Description: An optional description for the origin access control
```
Lou1415926: (I know this was unintentional, just throwing out an idea)
```suggestion
        Description: Access control for static s3 origin```

File: internal/pkg/template/templates/environment/partials/cdn-resources.yml [hunk 6]
```diff
@@ -78,6 +92,19 @@ CloudFrontDistribution:
             {{- else}}
             OriginProtocolPolicy: match-viewer
             {{- end}}
+      {{- if .CDNConfig.Static}}
+        - Id: !Sub 'copilot-${AppName}-${EnvironmentName}-origin-s3'
```
Lou1415926: nit: if "origin" is already implied, then -
```suggestion
        - Id: !Sub 'copilot-${AppName}-${EnvironmentName}-s3'
```

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -78,6 +92,19 @@ CloudFrontDistribution:
...
+          DomainName: {{.CDNConfig.Static.Location}}
+          OriginAccessControlId: !Ref CloudFrontOriginAccessControl
+          S3OriginConfig:
+            OriginAccessIdentity: ''
```
Lou1415926: Do you think it's worth commenting on why there is an empty string here? I'm fine either way.

Lou1415926 (commented):
LGTM just some nits!

cf-custom-resources/test/unique-json-values-test.js
iamhopaul123: it's automatically adjusted by the IDE prettify plugin. im assuming we can trust it 🤔

cf-custom-resources/lib/unique-json-values.js
```diff
@@ -92,43 +92,59 @@ const deadlineExpired = function () {
...
-          Object.entries(aliasesForService).filter(
-            ([key]) => filterFor.has(key)
+          Object.entries(aliasesForService).filter(([key]) =>
+            filterFor.has(key)
           )
         );
         const unique = new Set(Object.values(filteredAliasesForService).flat());
+        if (event.ResourceProperties.AdditionalAlias) {
+          unique.add(event.ResourceProperties.AdditionalAlias);
+        }
```
efekarakus: It's a bit odd that the function called `unique-json-values` is taking inputs called `Aliases` and `AdditionalAlias` 😅 

Maybe we shouldn't touch the old parameter name but what do you think of renaming `AdditionalAlias` to `AdditionalStrings` and making it a list of strings instead of a single string?

cf-custom-resources/test/unique-json-values-test.js
efekarakus: VSCode strikes again 😤

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-cloudfront-observability.yml
```diff
@@ -344,7 +344,7 @@ Resources:
         DefaultCacheBehavior:
           AllowedMethods: ["GET", "HEAD", "OPTIONS", "PUT", "PATCH", "POST", "DELETE"]
           CachePolicyId: 4135ea2d-6df8-44a3-9df3-4b5a84be39ad # See https://go.aws/3bJid3k
-          TargetOriginId: !Sub 'copilot-${AppName}-${EnvironmentName}-origin'
+          TargetOriginId: !Sub 'copilot-${AppName}-${EnvironmentName}'
```
Lou1415926: Will this introduce unexpected update for users with existing ALB+CDN setting?

Lou1415926 (approved):

internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-cloudfront-observability.yml
iamhopaul123: yeah i think so do you think we should keep the origin?

internal/pkg/template/env.go
```diff
@@ -178,8 +178,17 @@ func (elb *ELBAccessLogs) ShouldCreateBucket() bool {
 type CDNConfig struct {
 	ImportedCertificate *string
 	TerminateTLS        bool
+	Static              *CDNStatic
 }
 
+// CDNStatic represents static assets config for a Content Delivery Network.
+type CDNStatic struct {
```
efekarakus: nit:
```suggestion
type CDNStaticAssetConfig struct {
```

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -44,9 +44,23 @@ UniqueAliasesAction:
   Properties:
     ServiceToken: !GetAtt UniqueJSONValuesFunction.Arn
     Aliases: !Ref Aliases
+    {{- if .CDNConfig.Static}}
+    AdditionalAlias: {{.CDNConfig.Static.Alias}}
+    {{- end}}
```
efekarakus: like written in the previous comment, I wonder if instead of a single string it would be more flexible as a list of strings

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -44,9 +44,23 @@ UniqueAliasesAction:
...
     FilterFor: !Ref ALBWorkloads
 {{- end}}{{/* endif or .CDNConfig.ImportedCertificate .DelegateDNS */}}
 
+{{- if .CDNConfig.Static}}
+CloudFrontOriginAccessControl:
+  Type: AWS::CloudFront::OriginAccessControl
```
efekarakus, line 57: nit: should we include a 
```
Metadata:
    'aws:copilot:description': 'Access control to make the content in the S3 bucket only accessible through CloudFront
```

efekarakus (commented):
Looks great! tiny suggestions

internal/pkg/template/templates/environment/partials/cdn-resources.yml
iamhopaul123: Hmm I thought about this but then feel like we probably don't want to show this to the customer since they don't need to refer/use/care about this resource at all.

efekarakus: OK boss! the only part that sucks about the process tracker is that if an err occurs with this resource we don't display it in the progress tracker without the metadata :homer-disappearing:

iamhopaul123: Oh that makes sense i'll add it

efekarakus (approved):

PR #4329, opened 2023-01-06
chore: add messages delayed alarm for worker services

huanjani:
Based on the recommendation [here](https://docs.aws.amazon.com/AmazonECS/latest/userguide/deployment-alarm-failure.html), `ApproximateNumberOfMessagesDelayed` is a good metric for worker services. These changes enable  that alarm to be created for worker services.


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -830,6 +830,25 @@ func convertDeploymentConfig(in manifest.DeploymentConfiguration) template.Deplo
 	return out
 }
 
+func convertWorkerDeploymentConfig(in manifest.WorkerDeploymentConfig) template.DeploymentConfigurationOpts {
+	out := template.DeploymentConfigurationOpts{
+		MinHealthyPercent: minHealthyPercentDefault,
+		MaxPercent: maxPercentDefault,
```
iamhopaul123: Why is this `MaxPercent` instead of `MaxHealthyPercent`?

internal/pkg/manifest/workload_ecs.go
```diff
@@ -91,17 +91,33 @@ type AlarmArgs struct {
...
+type WorkerAlarmArgs struct {
+	AlarmArgs
+	MessagesDelayed *int `yaml:"messages_delayed"`
+}
+
 // DeploymentConfiguration represents the deployment strategies for a service.
 type DeploymentConfiguration struct {
 	Rolling        *string                    `yaml:"rolling"`
-	RollbackAlarms Union[[]string, AlarmArgs] // `yaml:"rollback_alarms"` 
+	RollbackAlarms Union[[]string, AlarmArgs] // `yaml:"rollback_alarms"`
```
iamhopaul123: Just double check: now that we have `Union` we don't have to worry about the env override mergo stuff for `transform.go` at all.

iamhopaul123 (approved):
LGTM awesome!

internal/pkg/deploy/cloudformation/stack/transformers.go
huanjani: I was just following the example of the existing code. @paragbhingre??

internal/pkg/manifest/workload_ecs.go
huanjani: Yeah, I don't think so. I manually tested for my last PR and it...just...worked! 🪄

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -830,6 +830,25 @@ func convertDeploymentConfig(in manifest.DeploymentConfiguration) template.Deplo
 	return out
 }
 
+func convertWorkerDeploymentConfig(in manifest.WorkerDeploymentConfig) template.DeploymentConfigurationOpts {
```
efekarakus: I wonder if we can get rid of the duplication by embedding the `DeploymentConfiguration` into a `WorkerDeploymentConfig` (here is an example: https://go.dev/play/p/IOO65jjWNIE)

```go
type DeploymentConfig struct { ... } // existing type

type WorkerDeploymentConfig struct {	
	DeploymentConfig
	RollbackAlarms Union[[]string, WorkerAlarmArgs] `yaml:"rollback_alarms"` // Same field name that needs to be replaced.
}

func convertWorkerDeploymentConfig(in manifest.WorkerDeploymentConfig) template.DeploymentConfigOpts {
	out := convertDeploymentConfig(in.DeploymentConfig)
	out.Rollback.MessagesDelayed = in.RollbackAlarms.Advanced.MessagesDelayed
	return out
}
```

internal/pkg/manifest/validate.go
```diff
@@ -131,10 +131,34 @@ func (d DeploymentConfiguration) validate() error {
 	return nil
 }
 
+func (d WorkerDeploymentConfig) validate() error {
```
efekarakus: samesies here, this would become:
```go
func (cfg WorkerDeploymentConfig) validate() error {
	if err := cfg.DeploymentConfig.validate(); err != nil { return err }
	return cfg.RollbackAlarms.validate()
}
```

internal/pkg/manifest/workload_ecs.go
```diff
@@ -91,17 +91,33 @@ type AlarmArgs struct {
 	MemoryUtilization *float64 `yaml:"memory_utilization"`
 }
 
+// WorkerAlarmArgs represents specs of CloudWatch alarms for Worker Service deployment rollbacks.
+type WorkerAlarmArgs struct {
+	AlarmArgs
+	MessagesDelayed *int `yaml:"messages_delayed"`
```
efekarakus, line 97: This is so strange that ECS recommends this metric 🤔 I'm struggling to understand how this gets used by customers today

internal/pkg/manifest/workload_ecs.go
```diff
@@ -91,17 +91,33 @@ type AlarmArgs struct {
...
 	Rolling        *string                    `yaml:"rolling"`
-	RollbackAlarms Union[[]string, AlarmArgs] // `yaml:"rollback_alarms"` 
+	RollbackAlarms Union[[]string, AlarmArgs] // `yaml:"rollback_alarms"`
 	// The rollback_alarms manifest field is a no-op until the EDS-CFN ABR bug is fixed.
 }
 
+// WorkerDeploymentConfig represents the deployment strategies for a worker service.
+type WorkerDeploymentConfig struct {
+	Rolling              *string                          `yaml:"rolling"`
+	WorkerRollbackAlarms Union[[]string, WorkerAlarmArgs] `yaml:"rollback_alarms"`
```
efekarakus: should we comment this?
```suggestion
	WorkerRollbackAlarms Union[[]string, WorkerAlarmArgs] // `yaml:"rollback_alarms"`
```

efekarakus (commented):
Looks great! just really a refactor comment

internal/pkg/deploy/cloudformation/stack/transformers.go
huanjani: He was following this: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-service-deploymentconfiguration.html

internal/pkg/deploy/cloudformation/stack/transformers.go
huanjani: Did a slight refactor after going in a big huge circle. 😸 
Talked offline; the version we settled on then didn't quite work.

internal/pkg/manifest/workload_ecs.go
```diff
@@ -91,15 +91,37 @@ type AlarmArgs struct {
...
+// WorkerDeploymentConfig represents the deployment strategies for a worker service.
+type WorkerDeploymentConfig struct {
+	DeploymentControllerConfig `yaml:",inline"`
+	WorkerRollbackAlarms       Union[[]string, WorkerAlarmArgs] // `yaml:"rollback_alarms"`
 	// The rollback_alarms manifest field is a no-op until the EDS-CFN ABR bug is fixed.
 }
 
 func (d *DeploymentConfiguration) isEmpty() bool {
-	return d == nil || (d.Rolling == nil && d.RollbackAlarms.IsZero())
+	return d == nil || (d.DeploymentControllerConfig.Rolling == nil && d.RollbackAlarms.IsZero())
```
Lou1415926: nit:
```suggestion
	return d == nil || (d.DeploymentControllerConfig.isEmpty() && d.RollbackAlarms.IsZero())
```
so that if we do have fields in addition to `rolling`, we won't have to remember to change the code here!

internal/pkg/manifest/validate.go
```diff
@@ -120,6 +120,26 @@ func (d DeploymentConfiguration) validate() error {
 	if err := d.RollbackAlarms.validate(); err != nil {
 		return fmt.Errorf(`validate "rollback_alarms": %w`, err)
 	}
+	if err := d.DeploymentControllerConfig.validate(); err != nil {
+		return fmt.Errorf(`validate "deployment controller strategy": %w`, err)
```
Lou1415926: What do you think of `validate "rolling"`? Just so that people know which field we are talking about.

internal/pkg/manifest/workload_ecs.go
```diff
@@ -91,15 +91,37 @@ type AlarmArgs struct {
...
+	Rolling *string `yaml:"rolling"`
+}
+
+// DeploymentConfiguration represents the deployment config for an ECS service.
 type DeploymentConfiguration struct {
-	Rolling        *string                    `yaml:"rolling"`
-	RollbackAlarms Union[[]string, AlarmArgs] // `yaml:"rollback_alarms"` 
+	DeploymentControllerConfig `yaml:",inline"`
+	RollbackAlarms             Union[[]string, AlarmArgs] // `yaml:"rollback_alarms"`
+	// The rollback_alarms manifest field is a no-op until the EDS-CFN ABR bug is fixed.
```
Lou1415926, line 109: What are we doing to prevent user from using this field?

Lou1415926 (commented):
LGTM!!

internal/pkg/manifest/workload_ecs.go
huanjani: commented out the yaml, so if they do use the field, nothing happens. users wouldn't know to try using it!

internal/pkg/manifest/workload_ecs.go
```diff
@@ -91,15 +91,41 @@ type AlarmArgs struct {
...
+	MessagesDelayed *int `yaml:"messages_delayed"`
+}
+
+// DeploymentControllerConfig represents deployment strategies for a service.
+type DeploymentControllerConfig struct {
+	Rolling *string `yaml:"rolling"`
+}
+
+// DeploymentConfiguration represents the deployment config for an ECS service.
 type DeploymentConfiguration struct {
```
efekarakus: nit: can we rename this to 🥺 to be consistent with `DeploymentControllerConfig` and `WorkerDeploymentConfig`
```suggestion
type DeploymentConfig struct {
```

efekarakus (approved):
🚀 just a tiny rename

PR #4331, opened 2023-01-06
chore(manifest): make cdn static path to be required

iamhopaul123:
Making cdn static path to be required because ALB is the default cache behavior and if we give `*` as a default value for path, it means all the traffic will be routed to the static pattern (a possible use case but not a sane default).


huanjani (approved):
I suppose making this more restrictive than before is okay, compatibility-wise. This is unlikely to break any scripts, too. LGTM!

efekarakus (approved):

PR #4346, opened 2023-01-10
chore(override): enable Lookup to detect YAML patch if dir has 1 file

efekarakus:
Related #4208 

### DIFF
--- a/internal/pkg/override/override.go
+++ b/internal/pkg/override/override.go
@@ -104,7 +104,7 @@ func (ext extension) isYAML() bool { return ext == ".yml" || ext == ".yaml" }
 
 func lookupYAMLPatch(path string, fs afero.Fs) (Info, error) {
 	if ext := extension(filepath.Ext(path)); !ext.isYAML() {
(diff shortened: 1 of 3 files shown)


internal/pkg/override/override.go
```diff
@@ -59,31 +104,31 @@ func (ext extension) isYAML() bool { return ext == ".yml" || ext == ".yaml" }
 
 func lookupYAMLPatch(path string, fs afero.Fs) (Info, error) {
 	if ext := extension(filepath.Ext(path)); !ext.isYAML() {
-		return unknownOverrider, fmt.Errorf("YAML patch documents require a .yml or .yaml extension: %q has a %q extension", path, ext)
+		return Info{}, fmt.Errorf("YAML patch documents require a .yml or .yaml extension: %q has a %q extension", path, ext)
```
paragbhingre: How about keeping extension values in double quotes? 
```suggestion
		return Info{}, fmt.Errorf(`YAML patch documents require a ".yml" or ".yaml" extension: %q has a %q extension`, path, ext)
```

efekarakus: Updated!

huanjani (approved):
lgtm!

iamhopaul123 (approved):

PR #4347, opened 2023-01-10
chore(sessions): add UserAgentExtras method to Provider

efekarakus:
I couldn't find a way to write a unit test that ensures the user-agent will have the extras on new requests.
However, I tested the method manually in `copilot app show` and I can see the new extras added on requests.

Related #4208 

### DIFF
--- a/internal/pkg/aws/sessions/sessions.go
+++ b/internal/pkg/aws/sessions/sessions.go
@@ -70,6 +70,14 @@ func UserAgentExtras(extras ...string) func(*Provider) {
 	}
 }
 
+// UserAgentExtras adds additional User-Agent extras to cached sessions and any new sessions.
+func (p *Provider) UserAgentExtras(extras ...string) {
+	p.userAgentExtras = append(p.userAgentExtras, extras...)
+	if p.defaultSess != nil {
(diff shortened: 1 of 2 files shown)


iamhopaul123 (approved):

internal/pkg/aws/sessions/sessions.go
```diff
@@ -70,6 +70,14 @@ func UserAgentExtras(extras ...string) func(*Provider) {
 	}
 }
 
+// UserAgentExtras adds additional User-Agent extras to cached sessions and any new sessions.
+func (p *Provider) UserAgentExtras(extras ...string) {
```
paragbhingre, line 74: What extra things happens here in addition to existing `UserAgentExtras` function?  Can't we use existing UserAgentExtras function for adding `overrides cdk` or `overrides cfnpatch` agents?

efekarakus: The `UserAgentExtras` function can only be used in the constructor, the method is so that you can add any extras at any point.

So we can start with a provider that's like this for a command:

```
provider := sessions.NewProvider(sessions.UserAgentExtras("svc package"))

....
if override.IsCDK() {
   provider.UserAgentExtras("cdk override")
}
```

paragbhingre (approved):

PR #4348, opened 2023-01-10
chore(override): add Noop Overrider

efekarakus:
Related #4208 

### DIFF
--- a/internal/pkg/override/noop.go
+++ b/internal/pkg/override/noop.go
@@ -1,0 +1,12 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package override
+
+// Noop represents an Overrider that does not do any transformations.
+type Noop struct{}
+
+// Override does nothing.
+func (no *Noop) Override(body []byte) ([]byte, error) {
+	return body, nil
+}
--- a/internal/pkg/override/noop_test.go
+++ b/internal/pkg/override/noop_test.go
@@ -1,0 +1,22 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package override
+
+import (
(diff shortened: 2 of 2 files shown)


internal/pkg/override/noop.go
```diff
@@ -0,0 +1,12 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package override
+
+// Noop represents an Overrider that does not do any transformations.
+type Noop struct{}
```
iamhopaul123, line 7: Why do we need this 🤔

efekarakus: The idea is to be able to do something like this from a client:

```go
info, err := override.Lookup(path/to/overrides)
if err != nil {
   return new(override.Noop)
}
if info.IsCDK() {
   return override.WithCDK(info.Path())
}
return override.WithYAMLPatch(info.Path())
```

This way the `deploy` package doesn't need to do any  `if override != nil` check and can just call `overrides.Override`

iamhopaul123: but why not return an error if `err != nil` instead of creating a `new(override.Noop)`?

efekarakus: Here is a sample scenario to illustrate:
```
copilot/
  frontend/
      manifest.yml
```
There is no `overrides/` directory defined by the customer, so we don't want to error.

My plan is to have a flow like this:
```go
info, err := override.Lookup("copilot/frontend/overrides")
if err != nil {
   if errors.Is(err, override.ErrNotExist) { return new(override.Noop), nil } // This is fine, have an overrider that doesn't do anythign.
   return nil, err
}
if info.IsCDK() {
  return override.WithCDK(info.Path()), nil
}
return override.WithYAMLPatch(info.Path()), nil
```

iamhopaul123 (approved):

huanjani (approved):
🚢

PR #4350, opened 2023-01-10
chore(override): enable passing down COPILOT env vars to CDK files

efekarakus:
Related #4208 

### DIFF
--- a/internal/pkg/override/cdk.go
+++ b/internal/pkg/override/cdk.go
@@ -16,30 +16,70 @@ import (
 	"github.com/spf13/afero"
 )
 
-// Executable is the interface to wrap os/exec calls.
-type Executable interface {
-	LookPath(file string) (string, error)
-	Command(name string, args ...string) *exec.Cmd // TODO(efe): Command should set cmd.Dir = rootAbsPath.
-}
-
 // CDK is an Overrider that can transform a CloudFormation template with the Cloud Development Kit.
 type CDK struct {
 	rootAbsPath string // Absolute path to the overrides/ directory.
 
-	out  io.Writer  // Writer for any os/exec calls.
-	fs   afero.Fs   // OS file system.
-	exec Executable // For testing os/exec calls.
+	out  io.Writer // Writer for any os/exec calls.
+	fs   afero.Fs  // OS file system.
+	exec struct {
+		LookPath func(file string) (string, error)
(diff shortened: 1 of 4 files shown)


internal/pkg/template/templates/overrides/cdk/stack.ts
```diff
@@ -4,14 +4,19 @@ import * as path from 'path';
...
 {{- end }}
 
+interface TransformedStackProps extends cdk.StackProps {
+    readonly appName?: string;
+    readonly envName?: string;
+}
+
 export class TransformedStack extends cdk.Stack {
-    constructor (scope: cdk.App, id: string, props?: cdk.StackProps) {
+    constructor (scope: cdk.App, id: string, props: TransformedStackProps) {
```
iamhopaul123: if `appName` and `envName` are optional, why do we make `props` required?

internal/pkg/override/cdk.go
```diff
@@ -16,30 +16,70 @@ import (
...
 	rootAbsPath string // Absolute path to the overrides/ directory.
 
-	out  io.Writer  // Writer for any os/exec calls.
-	fs   afero.Fs   // OS file system.
-	exec Executable // For testing os/exec calls.
+	out  io.Writer // Writer for any os/exec calls.
+	fs   afero.Fs  // OS file system.
+	exec struct {
+		LookPath func(file string) (string, error)
+		Command  func(name string, args ...string) *exec.Cmd
```
iamhopaul123, line 27: Will this be a problem for mock for unit test?

iamhopaul123 (approved):
LGTM overall :shipit:

internal/pkg/template/templates/overrides/cdk/stack.ts
efekarakus: good question 😂 changed to `appName` and `envName` being required

internal/pkg/override/cdk.go
efekarakus: Nope! In the unit tests we use the public constructor to define our own custom behavior:
```
WithCDK("some/path", CDKOpts{
   LookPathFn: func(string) (string, error) { return "", errors.New("some error") }
})
```

huanjani (approved):
LGTM!

PR #4357, opened 2023-01-12
refactor: deploy and template pkg to modify portmappings config

### LINKED ISSUE #4502: Taskdef override fails when trying to override portmappings for backend service
From gitter - I want to expose multiple ports on my backend service(in private vpc),I am trying to do so by adding port mapping to taskef_overrides but i am getting a "Fn::If requires a list argument with three elements" error.

Screenshot 2023-02-16 at 11 46 05 AM

paragbhingre:
This PR focuses on converting exposedPorts into portMappings in order to set right set of ports in the portmappings for primary as well as sidecar containers. 

This PR is a part of https://github.com/aws/copilot-cli/pull/4192

Also resolves #4502


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -112,6 +135,7 @@ func convertSidecar(s map[string]*manifest.SidecarConfig) ([]*template.SidecarOp
 			EntryPoint:   entrypoint,
 			HealthCheck:  convertContainerHealthCheck(config.HealthCheck),
 			Command:      command,
+			PortMappings: portMappings[name],
```
iamhopaul123: Why do we have both `Port` and `Portmappings` in `&template.SidecarOpts`

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -66,22 +66,45 @@ var (
...
+		}
+		portMapping[exposedPort.ContainerName] = append(portMapping[exposedPort.ContainerName], out)
+	}
+	return portMapping
+}
+
 // convertSidecar converts the manifest sidecar configuration into a format parsable by the templates pkg.
-func convertSidecar(s map[string]*manifest.SidecarConfig) ([]*template.SidecarOpts, error) {
-	if s == nil {
+func convertSidecar(sidecarsMap map[string]*manifest.SidecarConfig, portMappings map[string][]*template.PortMapping) ([]*template.SidecarOpts, error) {
```
iamhopaul123:
```suggestion
func convertSidecar(sidecarsMap map[string]*manifest.SidecarConfig, portMappings []*template.PortMapping) ([]*template.SidecarOpts, error) {
```

iamhopaul123 (commented):
I feel like instead of passing `exposedPorts []manifest.ExposedPort` around as a parameter in `template` package functions, what we should do is having a "parse" function that takes the manifest as input and return a struct containing all the info we want.

Right now the output of `s.manifest.ExposedPorts()` is not very useful in `template` package, so that we make too many assumptions in `template` package about how `manifest` should work.

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: Good question - sidecar `port` only defines the port that is being mentioned in the `port` field but portmappings hold all the extra ports as well which are being opened through either alb or nlb config. Does that make sense?

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: portmappings is in the form of `map[constianerName][]template.PortMapping` which helps us easily set particular container's portmappings to its config.

internal/pkg/deploy/cloudformation/stack/transformers.go
Lou1415926: Does this mean we don't need `Port` in `template.SidecarOpts` anymore?

internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
```diff
@@ -49,17 +49,16 @@
...
-{{- if $sidecar.Port}}
+{{include "image-overrides" . | indent 2}}
+{{- if .PortMappings}}
   PortMappings:
-    - ContainerPort: {{$sidecar.Port}}
-    {{/* Multiple exposed port is not supported yet. Therefore, if the target container is the sidecar, the target port must be the one and only one port that the container exposes. */}}
-    {{- if eq $.HTTPTargetContainer.Name $sidecar.Name}}
+    {{- range $portMapping := .PortMappings }}
+    - ContainerPort: {{ $portMapping.ContainerPort }}
+  {{- if and (eq $.HTTPTargetContainer.Name $sidecar.Name) (isEqual $.HTTPTargetContainer.Port $portMapping.ContainerPort)}}
```
Lou1415926: nit: can we align the `{{- end}}` after `Name: target` with this `{{- if}}`?

internal/pkg/deploy/cloudformation/stack/worker_svc_test.go
```diff
@@ -67,7 +67,7 @@ Outputs:
     Value: hello`,
 				}
 			},
-			wantedErr: fmt.Errorf("convert the sidecar configuration for service frontend: %w", errors.New("cannot parse port mapping from 80/80/80")),
+			wantedErr: fmt.Errorf("exposed ports configuration for service frontend: cannot parse port mapping from 80/80/80"),
```
Lou1415926: I think we should keep the original test case, and have a new test case for the "exposed port" error

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -271,7 +277,7 @@ func (s *LoadBalancedWebService) Template() (string, error) {
 	return string(overriddenTpl), nil
 }
 
-func (s *LoadBalancedWebService) httpLoadBalancerTarget() (targetContainer *string, targetPort *string) {
+func (s *LoadBalancedWebService) httpLoadBalancerTarget(exposedPorts []manifest.ExposedPort) (targetContainer *string, targetPort *string) {
```
Lou1415926: Would ⬇️ make sense:
```go
// In manifest pkg
func (s *LoadBalancedWebService) httpLoadBalancerTarget() (targetContainer *string, targetPort *string, error error)

// Here
targetContainer, targetPort, err := s.manifest.httpLoadBalancerTarget()
```

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -282,6 +288,15 @@ func (s *LoadBalancedWebService) httpLoadBalancerTarget() (targetContainer *stri
 		targetPort = s.manifest.Sidecars[aws.StringValue(targetContainer)].Port
 	}
 
+	// Route load balancer traffic to the target_port if mentioned.
```
Lou1415926: Should we have returned after L288 before L289?

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -87,7 +87,12 @@ func (s *BackendService) Template() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	sidecars, err := convertSidecar(s.manifest.Sidecars)
+	exposedPorts, err := s.manifest.ExposedPorts()
+	if err != nil {
+		return "", fmt.Errorf("exposed ports configuration for service %s: %w", s.name, err)
```
efekarakus: nit:
```suggestion
		return "", fmt.Errorf("parse exposed ports in service manifest %s: %w", s.name, err)
```

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -87,7 +87,12 @@ func (s *BackendService) Template() (string, error) {
...
+	}
+	portMappings := convertPortMapping(exposedPorts)
+	sidecars, err := convertSidecar(s.manifest.Sidecars, portMappings)
```
efekarakus: nit:
```suggestion
	portMappings := convertPortMappings(exposedPorts)
	sidecars, err := convertSidecars(s.manifest.Sidecars, portMappings)
```

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
efekarakus: 👍 that'd be nice! I like the idea of the `stack` package to be responsible only for conversion from `manifest` to `template` types. Since this method doesn't return a `template` type it hints to me that it's a good candidate to move to the `manifest` pkg.

I always forget why we are returning pointers here though, that's confusing to me :P

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
efekarakus: This method is also identical to `func (s *BackendService) httpLoadBalancerTarget` as far as I can tell. 
We could remove the duplication and not have to update this change in multiple locations like this:
```go
package manifest

func (svc *LoadBalancedWebService) HTTPLoadBalancerTarget() (ExposedPort, error)
func (svc *BackendService) HTTPLoadBalancerTarget() (ExposedPort, error)

func httpLoadBalancerTarget(albConfig RoutingRuleConfiguration, ports []ExposedPort) (ExposedPort, error)
```

internal/pkg/template/workload.go
```diff
@@ -234,6 +242,11 @@ func (tg HTTPTargetContainer) IsHTTPS() bool {
 	return tg.Port == "443"
 }
 
+// IsEqual returns true if httpContainerPort and port mapping port are equal.
+func IsEqual(s string, p uint16) bool {
```
efekarakus: The function name is a bit too generic for the given parameter types 💭 . For example, a client might think to use `isEqual struct1{} struct2{}`.

What do you think of instead adding the following function:
```go
func strconvUint16(val uint16) string { return strconv.FormatUint(uint64(val), 10) } 
```
and in the Go template we can do `{{eq port1 (strconvUint16 port2)}}`

internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
```diff
@@ -49,17 +49,16 @@
...
       Name: target
     {{- end}}
-    {{- if $sidecar.Protocol}}
-      Protocol: {{$sidecar.Protocol}}
-    {{- end}}
+      Protocol: {{if eq $portMapping.Protocol "udp"}}udp{{else}}tcp{{end}}
```
efekarakus: can we possibly move this logic into `converPortMappings` in `stack` so that this becomes:
```suggestion
      Protocol: {{$portMapping.Protocol}}
```

internal/pkg/template/workload.go
```diff
@@ -143,6 +143,14 @@ type SidecarOpts struct {
 	Command              []string
 	HealthCheck          *ContainerHealthCheck
 	EnvAddonsFeatureFlag bool
+	PortMappings         []*PortMapping
```
efekarakus: I wonder if the struct would be more readable if we grouped all the fields for the main container together. Both in the `stack` pkg here but also in `template.WorkloadOpts`.

For example:
```go
type WorkloadOpts struct {
//...
EnvVersion: s.rc.EnvVersion,

// Configuration for the main container.
PortMappings  []PortMapping
Variables         map[string]Variable
Secrets           map[string]Secret
Command      []string
Entrypoint    []string

// Other configurations...
}
```

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: Yeah I think it is safe to remove Port from `template.SidecarOpts`.

internal/pkg/deploy/cloudformation/stack/worker_svc_test.go
paragbhingre: So the thing here is that we send `exposedPorts` to the `convertSidecar` in `stack/lb-web-svc.go` . And in order `exposedPorts` we call `ExposedPorts()` method from the manifest which also parses the sidecar container ports in order to organize `[]exposedPort` for us. So in this test case before even `convertSidecar` fail for the same parsing reason, `ExposedPorts()` fails and responds with the new error. Does that make sense? So even if we create a new test case for the exposedports this test case will still give the changed error i.e. `exposed ports configuration ....`

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
paragbhingre: > Should we have returned after L288 before L289?

The reason why we do not return after 288 and before 289 is because, say we have the following manifest, where the customer wants to expose a new port `82` of the sidecar container, then he can write a manifest like the following  - 

```
name: lbservice

image:
  port: 80

http:
  path: '/'
  target_port: 82
  target_container: nginx

sidecars:
  nginx:
    image:
      port: 81

```
In that case, if we return after line 288, then we are making the assumption that the customer has not provided the target_port and wants to open a port that is mentioned in the `sidecar.image.port`. That is why we need to specifically check for `target_port` condition as well.

internal/pkg/deploy/cloudformation/stack/worker_svc_test.go
Lou1415926: I meant we want to change the manifest input for the original test case (to a valid manifest)  so that the old error message is testable :D With the reason you give, anything after the line of `ExposedPorts()` would be untestable. That would be weird and hard to justify I think 🤔

paragbhingre: No, I am not saying that anything after `ExposedPorts()` is untestable. The only thing is we are making `manifest.ParsePortMapping(config.Port)` call in the `convertSidecar` as well as `ExposedPorts()` but because `ExposedPorts()` is called before `convertSidecar` now we are getting that error from `ExposedPorts()`. 
I don't think we can modify the manifest in a way that would error for `manifest.ParsePortMapping(config.Port)` call in the `convertSidecar` but not in the `ExposedPort()`

Lou1415926: Does it mean `cannot parse port mapping` is dead code?

paragbhingre: Nope, it is not a dead code as we can see `cannot parse port mapping` in `wantedErr` in the above code snippet.

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: We will also remove Protocol from `template.SidecarOpts` as portmapping holds protocol info as well.

internal/pkg/manifest/svc.go
```diff
@@ -570,3 +575,79 @@ func sortExposedPorts(exposedPorts []ExposedPort) []ExposedPort {
...
+	if rrTarget != nil && *rrTarget != *targetContainer {
+		targetContainer = rrTarget
+		targetPort = s.Sidecars[aws.StringValue(targetContainer)].Port
+	}
+	// Route load balancer traffic to the target_port if mentioned.
+	if container, port := httpLoadBalancerTarget(exposedPorts, rrTargetPort); port != nil {
+		targetPort = port
+	} else if container != nil {
+		targetContainer = container
+	}
```
iamhopaul123: The logic seems confusing. I feel like we should

1. if target port and target container both are empty, return the main container/port
2. if target port is empty, return target container and it's only port
3. if target container is empty, find the container and return the container and target port
4. else (both are not empty), return target container/target port

Also both `rrTargetContainer` and `rrTargetPort` are from `manifest.RoutingRule` which we can access from `s`. Why do we make them parameters?

paragbhingre: I think I agree to the steps that you've mentioned, and that is what the current logic also does. May be to simplify I can follow your steps. 

> Also both rrTargetContainer and rrTargetPort are from manifest.RoutingRule which we can access from s. Why do we make them parameters?

This is a great question - we can not always take `rrTargetContainer` and `rrTargetPort` from `manifest.RoutingRule` because when `additional_rules` will come into picture then targetContainer and targetPort will be from `manifest.RoutingRule.additional_rules` and not from `manifest.RoutingRule`. Does that make sense?

internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
```diff
@@ -49,17 +49,16 @@
 {{- if $sidecar.Essential}}
   Essential: {{$sidecar.Essential}}
 {{- end}}
-{{include "image-overrides" . | indent 2}}  
-{{- if $sidecar.Port}}
+{{include "image-overrides" . | indent 2}}
+{{- if .PortMappings}}
```
Lou1415926: Was this intended to be `$sidecar.PortMappings`?

internal/pkg/manifest/svc.go
```diff
@@ -509,7 +510,9 @@ func (cfg NetworkLoadBalancerConfiguration) exposedPorts(exposedPorts []ExposedP
 	if err != nil {
 		return nil, err
 	}
-	if protocol == nil {
+	if strings.ToLower(aws.StringValue(protocol)) == "udp" {
```
Lou1415926: `nlb` does not accept `udp` protocol.

internal/pkg/manifest/svc.go
```diff
@@ -547,7 +550,9 @@ func (sidecar SidecarConfig) exposedPorts(sidecarName string) ([]ExposedPort, er
...
+		protocol = aws.String("udp")
+	} else {
 		protocol = aws.String("tcp")
 	}
```
Lou1415926: nit: to simplify the ptr/case conversion
```go
        sidecarPort, protocolPtr, err := ParsePortMapping(sidecar.Port)
        if err != nil { return nil, err }
        protocol := aws.StringValue(protocolPtr)  
        if protocolPtr == nil {
              protocol = "tcp"
        }       
         ...
        return []ExposedPort{
                Protocol: string.ToLower(protocol)
        }
```

internal/pkg/manifest/svc.go
```diff
@@ -570,3 +575,79 @@ func sortExposedPorts(exposedPorts []ExposedPort) []ExposedPort {
 	})
 	return exposedPorts
 }
+
+// HTTPLoadBalancerTarget returns target container and target port for the ALB configuration.
+func (s *LoadBalancedWebService) HTTPLoadBalancerTarget(exposedPorts []ExposedPort, rrTargetContainer *string, rrTargetPort *uint16) (targetContainer *string, targetPort *string) {
```
Lou1415926: Why do we prefer passing `rrTargetContainer` and `rrTargetPort` as parameters? I thought these information are already present in `s *LoadBalancedWebService`.

Lou1415926: Just some food for thought 💭 Not a request.

It seems like we are passing information from `manifest` to `stack` and then from `stack` back to `manifest`. The `exposedPort`, `rrTargetContainer` are examples.  In general, I found this pattern a bit difficult to follow.

From what I understand, for us to know:
1. The summary of port mappings (that is, `[]ExposedPort` of all containers)
2. For each container, what are the ports that it exposes (`[]ExposedPort` of one container)
3.  What is the target container, and target port

All of these information can be computed within the `manifest` package. 

Ideally, we would have a `manifest` package whose interface looks something like:
```go
type LoadBalancedWebService struct {
    cachedExposedPorts []ExposedPort
}

type ParsedSidecarConfig struct {
    Name string
    ...
    Ports []ExposedPort // Contains the ports that are exposed by the sidecar
}
func (s *LoadBalancedWebService) Sidecars() []ParsedSidecarConfig // Instead of figuring out what ports a sidecar container expose in `stack/transformer.go` L122, we do it in `manifest` pkg

func (s *LoadBalancedWebService) HTTPLoadBalancerTarget() (targetContainer, targetPort string)

```

Do you think this makes sense?

internal/pkg/manifest/svc.go
```diff
@@ -570,3 +575,87 @@ func sortExposedPorts(exposedPorts []ExposedPort) []ExposedPort {
...
+		targetPort = aws.String(template.StrconvUint16(aws.Uint16Value(rrTargetPort)))
+		if container := findContainerNameGivenPort(aws.Uint16Value(rrTargetPort), exposedPorts); container != "" {
+			targetContainer = aws.String(container)
+		}
+	}
+	return
+}
+
+// ContainerPort returns the main container port.
+func (s *LoadBalancedWebService) ContainerPort() string {
```
Lou1415926: nit:
```suggestion
func (s *LoadBalancedWebService) MainContainerPort() string {
```

internal/pkg/template/workload.go
```diff
@@ -142,6 +140,14 @@ type SidecarOpts struct {
 	EntryPoint   []string
 	Command      []string
 	HealthCheck  *ContainerHealthCheck
+	PortMappings []*PortMapping
+}
+
+// PortMapping holds container port mapping configuration.
+type PortMapping struct {
+	Protocol      string
+	ContainerPort uint16
```
Lou1415926, line 149: This might be my illusion but it seems like we are converting  `uint16`, `uint64` and `string`  a lot. If so can we simplify this?

internal/pkg/template/templates/workloads/partials/cf/sidecars.yml
paragbhingre: yeah right, but `.PortMappings` is also taking the port mappings related to that sidecar only.  I will make the change.

internal/pkg/template/workload.go
paragbhingre: You are right, but I have kept these values such that we will need to do less conversion while converting exposedPorts to PortMappings. Even if I change it ContainerPort to string we still need to make some conversions at other locations. Either ways conversions are required hence I at least tried making sure that  PortMapping and ExposedPort struct are similar.

internal/pkg/manifest/svc.go
paragbhingre: > Why do we prefer passing rrTargetContainer and rrTargetPort as parameters?

I have removed this from the code for now which is required in the next PRs when additional rules comes into the picture. I will add that part when I will release the PR for additional rules.

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -87,7 +87,11 @@ func (s *BackendService) Template() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	sidecars, err := convertSidecar(s.manifest.Sidecars)
+	_, err = s.manifest.ExposedPorts()
```
Lou1415926: Can `ExposedPorts` be a private function in of `manifest`?

paragbhingre: Made required changes as per the offline discussion.

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -87,7 +87,11 @@ func (s *BackendService) Template() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	sidecars, err := convertSidecar(s.manifest.Sidecars)
+	exposedPorts, err := s.manifest.ExposedPorts()
+	if err != nil {
+		return "", fmt.Errorf("parse exposed ports in service manifest %s: %w", s.name, err)
+	}
+	tmplSidecars, err := convertSidecars(s.manifest.Sidecars, exposedPorts.ContainerToPortsMapping)
```
efekarakus: nit: is there a reason why we renamed the variable? 
```suggestion
	sidecars, err := convertSidecars(s.manifest.Sidecars, exposedPorts.ContainerToPortsMapping)
```

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -140,56 +144,75 @@ func (s *BackendService) Template() (string, error) {
...
+
+		// Configuration for the main container.
+		EntryPoint:   entrypoint,
+		Command:      command,
+		HealthCheck:  convertContainerHealthCheck(s.manifest.BackendServiceConfig.ImageConfig.HealthCheck),
+		PortMappings: convertPortMappings(exposedPorts.ContainerToPortsMapping[s.name]),
+		Secrets:      convertSecrets(s.manifest.BackendServiceConfig.Secrets),
+		Variables:    convertEnvVars(s.manifest.BackendServiceConfig.Variables),
+
+		// Additional options that are common between **all** workload templates.
```
efekarakus, line 169: yay thank you for this!

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -66,8 +66,22 @@ var (
 	}
 )
 
-// convertSidecar converts the manifest sidecar configuration into a format parsable by the templates pkg.
-func convertSidecar(s map[string]*manifest.SidecarConfig) ([]*template.SidecarOpts, error) {
+func convertPortMappings(exposedPorts []manifest.ExposedPort) []*template.PortMapping {
+	var portMapping []*template.PortMapping
+	for _, exposedPort := range exposedPorts {
```
```suggestion
	out := make([]*template.PortMapping, len(exposedPorts))
	for i, exposedPort := range exposedPorts {
```

internal/pkg/manifest/backend_svc.go
```diff
@@ -18,9 +18,9 @@ type BackendService struct {
 	Workload             `yaml:",inline"`
 	BackendServiceConfig `yaml:",inline"`
 	// Use *BackendServiceConfig because of https://github.com/imdario/mergo/issues/146
-	Environments map[string]*BackendServiceConfig `yaml:",flow"`
-
-	parser template.Parser
+	Environments       map[string]*BackendServiceConfig `yaml:",flow"`
+	cachedExposedPorts ParsedContainerConfig
```
efekarakus: I understand that we are introducing caching because this method is invoked several times (once in `stack` and once in `manifest.HTTPLoadBalancerTarget`) but I think it add significant complexity due to the conditional and having to check both fields.

Can we remove this cached variable unless we can measure a significant performance difference without it?

internal/pkg/manifest/backend_svc.go
```diff
@@ -190,19 +190,23 @@ func newDefaultBackendService() *BackendService {
 }
 
 // ExposedPorts returns all the ports that are container ports available to receive traffic.
-func (b *BackendService) ExposedPorts() ([]ExposedPort, error) {
+func (b *BackendService) ExposedPorts() (ParsedContainerConfig, error) {
```
efekarakus: I was a big fan of the previous method signature 🥲 Atm it's not as clear as before from a client's perspective why a method that should return all the exposed ports is returning a container configuration.

From my understanding it looks like we're providing we're introducing `ParsedContainerConfig` because we want to be able to query `ContainerToPortsMapping` and `PortToContainerMapping` maps.

Instead what do you think of the following layout?
```go
func (b *BackendService) ExposedPorts() (ExposedPorts, error)

// ExposedPorts is a wrapper for a slice of ExposedPort to simplify querying data from the slice.
type ExposedPorts []ExposedPort

// ContainerNameIndex returns a map of exposed ports per container name.
func (p ExposedPorts) ContainerNameIndex() map[string][]ExposedPort

// PortToContainerNameIndex returns a map of  exposed port value in a task to container name.
func (p ExposedPorts) PortToContainerNameIndex() map[uint16]string
```

internal/pkg/manifest/svc.go
```diff
@@ -570,3 +575,114 @@ func sortExposedPorts(exposedPorts []ExposedPort) []ExposedPort {
 	})
 	return exposedPorts
 }
+
+// HTTPLoadBalancerTarget returns target container and target port for the ALB configuration.
```
efekarakus, line 579: Similar to the previous PR (#4438) can we highlight that these methods assume that the ALB config is not empty

internal/pkg/manifest/backend_svc.go
paragbhingre: If I am understanding this correctly, then we would need to call `ExposedPorts()` first every time we need one of `ContainerNameIndex` or `PortToContainerNameIndex`. Isn't that a lot of computation every time we need exposed ports?

internal/pkg/manifest/svc.go
paragbhingre: I have made changes as per your ask but we are not really calling this method only when ALB is empty reason being if we do so then [HTTPTargetContainer](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/deploy/cloudformation/stack/lb_web_svc.go#L223)'s  port and name will be nil if NLB is the one who is exposing the ports. And we have a known issue with NLB that we do not support Service connect when targetPort and targetContainers are exposed through NLB. We still look for the main container name and it's port as targetContainer and targetPort. 
We can fix this issue separately but for time being I have made changes you mentioned but not really calling this function when ALB is empty. 
Let me know if you have any more questions around this.

iamhopaul123 (approved):

internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/simple-template-without-port-config.yml
```diff
@@ -0,0 +1,387 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
Lou1415926, line 1: Is it possible not to add a new `testdata` template? It's a real pain maintaining all of them - for example, when all I've done is adding a new `Parameters` and I have to modify > 10 `testdata` templates just to pass the integration test. 

Maybe we can instead modify an existing `testdata` to test the case of "without-port-config`?

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -140,56 +144,75 @@ func (s *BackendService) Template() (string, error) {
 	if s.manifest.Network.Connect.Enabled() {
 		scConfig = convertServiceConnect(s.manifest.Network.Connect)
 	}
-	targetContainer, targetContainerPort := s.httpLoadBalancerTarget()
+	targetContainer, targetContainerPort, err := s.manifest.HTTPLoadBalancerTarget()
```
Lou1415926, line 147: I see `targetContainer` and `targetConatinerPort` are passed via go-template and as `Parameters` at the same time. Just out of curiosity (not a request) - do we plan to remove them from `Parameters` in the future?

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -140,56 +144,75 @@ func (s *BackendService) Template() (string, error) {
...
+		CustomResources: crs,
+
+		// Sidecar config.
+		Sidecars: sidecars,
+
+		// service connect and service discovery options.
 		ServiceConnect:           scConfig,
-		HTTPHealthCheck:          convertHTTPHealthCheck(&s.manifest.RoutingRule.HealthCheck),
-		DeregistrationDelay:      deregistrationDelay,
-		AllowedSourceIps:         allowedSourceIPs,
```
Lou1415926, line 172: Just double checking I am reading correctly about what's in `template.WorkloadOpts{}`
1. The only new changes are `HTTPTargetContainer` and `PortMappings`
2. The other changes are purely reordering.

Is this correct?

internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/simple-template-without-port-config.yml
paragbhingre: I thought about it but we really had no test case with the backend service without port configuration that is why I thought of adding this valid and important use case which we can use in the future for multiple different scenarios in case port is not mentioned. 

But if you insist then I can just merge this case in another test case.

internal/pkg/deploy/cloudformation/stack/scheduled_job.go
```diff
@@ -136,9 +136,13 @@ func (j *ScheduledJob) Template() (string, error) {
...
 	}
-	sidecars, err := convertSidecar(j.manifest.Sidecars)
+	exposedPorts, err := j.manifest.ExposedPorts()
 	if err != nil {
-		return "", fmt.Errorf("convert the sidecar configuration for job %s: %w", j.name, err)
+		return "", fmt.Errorf("parse exposed ports in service manifest %s: %w", j.name, err)
+	}
+	sidecars, err := convertSidecars(j.manifest.Sidecars, exposedPorts.PortsForContainer)
+	if err != nil {
+		return "", fmt.Errorf("convert the sidecar configuration for service %s: %w", j.name, err)
```
Lou1415926:
```suggestion
		return "", fmt.Errorf("convert the sidecar configuration for job %s: %w", j.name, err)
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -66,8 +66,22 @@ var (
...
+		out := &template.PortMapping{
+			ContainerPort: exposedPort.Port,
+			Protocol:      exposedPort.Protocol,
+			ContainerName: exposedPort.ContainerName,
+		}
+		portMapping = append(portMapping, out)
+	}
+	return portMapping
```
```suggestion
	portMapping  := male([]*template.PortMapping, len(exposedPorts))
	for idx, exposedPort := range exposedPorts {
		portMapping[idx] = &template.PortMapping{
			ContainerPort: exposedPort.Port,
			Protocol:      exposedPort.Protocol,
			ContainerName: exposedPort.ContainerName,
		}
	}
	return portMapping
```

internal/pkg/deploy/cloudformation/stack/transformers_test.go
```diff
@@ -34,18 +34,12 @@ func Test_convertSidecar(t *testing.T) {
...
-
-			wantedErr: fmt.Errorf("cannot parse port mapping from b/a/d/P/o/r/t"),
-		},
 		"good port without protocol": {
 			inPort:      aws.String("2000"),
 			inEssential: true,
 
 			wanted: &template.SidecarOpts{
 				Name:       "foo",
-				Port:       aws.String("2000"),
```
Lou1415926, line 48: It doesn't sound right to me that with `inPort:"2000"`, the expected outcome does not have any `PortMappings` 🤔. I know this is because the exposed ports are passed in as `nil` in the unit test, but I think we shouldn't leave the unit tests like this, that the `wanted` outcome looks wrong comparing with `in`.

I feel like we should either:
1. Remove `inPort` from these test cases. That is, we don't test at all if a sidecar's port is converted as expected in the unit test. We instead rely on integration test to make sure it's all right.
2. Or, pass in a mock `[]manifest.ExposedPort` in `got, err := convertSidecars(sidecar, mockExposedPorts)` at the current L225.

Personally, I am fine either way!

internal/pkg/manifest/backend_svc_test.go
```diff
@@ -1105,16 +1117,21 @@ func TestBackendService_ExposedPorts(t *testing.T) {
 					},
 				},
 			},
-			wantedExposedPorts: []ExposedPort{
-				{
-					Port:          80,
-					ContainerName: "frontend",
-					Protocol:      "tcp",
+
```
Lou1415926, line 1119: Can we remove the current  L1113 ⬇️ ?
```
//Port: aws.String("2000"),
```

internal/pkg/manifest/svc.go
```diff
@@ -571,6 +576,121 @@ func sortExposedPorts(exposedPorts []ExposedPort) []ExposedPort {
...
+		}
+		return
+	}
+
+	if rrTargetContainer == nil { // when target_container is nil
+		container, port := httpLoadBalancerTarget(exposedPorts, rrTargetPort)
+		targetPort = aws.StringValue(port)
+		if container != nil {
+			targetContainer = aws.StringValue(container)
+		}
```
Lou1415926, line 609: How should the caller interpret it when `targetContainer` is returned as `""` from `func (s *LoadBalancedWebService) HTTPLoadBalancerTarget()`?

internal/pkg/manifest/svc.go
```diff
@@ -571,6 +576,121 @@ func sortExposedPorts(exposedPorts []ExposedPort) []ExposedPort {
...
+		return
+	}
+
+	// when both target_port and target_container are not nil
+	targetContainer = aws.StringValue(rrTargetContainer)
+	targetPort = template.StrconvUint16(aws.Uint16Value(rrTargetPort))
+	return
```
Lou1415926, line 658: Are these exactly the same as `func (s *LoadBalancedWebService) HTTPLoadBalancerTarget()`?

Lou1415926 (commented):
I can approve the PR with the current code structure! but before that, just some small nits and questions 😂

internal/pkg/deploy/cloudformation/stack/backend_svc.go
paragbhingre: Yes we can remove them in the future but for time being to support places where TargetPort and TargetContainer are being used apart from the multiple port implementation we have kept it as is. One such example is [service connect.](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml#L71)

internal/pkg/deploy/cloudformation/stack/backend_svc.go
paragbhingre: Yes you are absolutely correct.

internal/pkg/manifest/svc.go
paragbhingre: I don't think that would the case because at the beginning of the function we set `targetContainer = aws.StringValue(s.Name)` So main container name would be `targetContainer`. Does that make sense?

internal/pkg/manifest/svc.go
paragbhingre: Yes it is, the only difference is how we are getting sidecars and primary container name from given `LoadBalancedWebService` or `BackendService`

Lou1415926 (approved):

PR #4358, opened 2023-01-12
feat: enable cdk overrides on svc package

efekarakus:
Support rendering `overrides/` dir with the CDK on `svc package`

Related #4208 


internal/pkg/cli/deploy/workload.go
```diff
@@ -262,12 +270,16 @@ func (d *workloadDeployer) generateCloudFormationTemplate(conf stackSerializer)
 	if err != nil {
 		return nil, fmt.Errorf("generate stack template: %w", err)
 	}
+	transformedTpl, err := d.overrider.Override([]byte(tpl))
```
iamhopaul123, line 268: Just curious why does `Override` take and return `[]byte` instead of `string`?

internal/pkg/cli/svc_package.go
```diff
@@ -150,6 +160,34 @@ func newWorkloadStackGenerator(o *packageSvcOpts) (workloadStackGenerator, error
...
+	if err != nil {
+		var errNotExist *override.ErrNotExist
+		if errors.As(err, &errNotExist) {
+			return new(override.Noop), nil
+		}
+		return nil, fmt.Errorf("look up overrider at %q: %w", pathToOverridesDir, err)
+	}
+	switch {
+	case info.IsCDK():
+		provider.UserAgentExtras("override cdk")
```
iamhopaul123: Maybe `cdk override` instead?

internal/pkg/cli/svc_package.go
```diff
@@ -150,6 +160,34 @@ func newWorkloadStackGenerator(o *packageSvcOpts) (workloadStackGenerator, error
...
+			Stdout: log.OutputWriter,
+			FS:     fs,
+			EnvVars: map[string]string{
+				"COPILOT_APPLICATION_NAME": app,
+				"COPILOT_ENVIRONMENT_NAME": env,
+			},
+		}), nil
+	case info.IsYAMLPatch():
+		provider.UserAgentExtras("override yamlpatch")
+		return new(override.Noop), nil
```
iamhopaul123: Just double check: this is a todo right?

internal/pkg/override/cdk.go
```diff
@@ -113,7 +113,7 @@ func (cdk *CDK) install() error {
 
 func (cdk *CDK) transform(body []byte) ([]byte, error) {
 	buildPath := filepath.Join(cdk.rootAbsPath, ".build")
-	if err := cdk.fs.MkdirAll(buildPath, fs.ModeDir); err != nil {
+	if err := cdk.fs.MkdirAll(buildPath, 0755); err != nil {
```
huanjani, line 116: What's the difference/which should we use when? (I see we have `0755` everywhere else in the codebase.)

internal/pkg/cli/svc_package.go
```diff
@@ -150,6 +160,34 @@ func newWorkloadStackGenerator(o *packageSvcOpts) (workloadStackGenerator, error
 	return deployer, nil
 }
 
+func newTemplateOverrider(pathToOverridesDir, app, env string, fs afero.Fs, provider *sessions.Provider) (clideploy.Overrider, error) {
+	info, err := override.Lookup(pathToOverridesDir, fs)
```
Lou1415926: Do we want to add a unit test for this?

Lou1415926 (commented):
LGTM!

internal/pkg/cli/svc_package.go
efekarakus: Done! I've moved the constructor to the `deploy/cli` pkg and added tests

internal/pkg/cli/deploy/workload.go
efekarakus: I was following the stdlib that usually takes `[]byte` but you're right that at least today we don't expect to override anything that's not a string 😂 .

I'll update the `override` pkg and interfaces in a separate PR to `Override(string) (string, error)` if that sounds ok?

internal/pkg/cli/svc_package.go
efekarakus: I kept it this is way to hint at a hierharchy "override <option>" but 🤷 I can change if we feel strongly against it

internal/pkg/cli/svc_package.go
efekarakus: yea that's a todo 😅

internal/pkg/override/cdk.go
efekarakus: > tl;dr: just use `0755`.


when i had `fs.ModeDir` I got an access denied permission because this is the permission that it sets apparently:
```
ls -l                                                                                                                                 
d---------  2 karakuse  staff     64 Jan 19 12:55 test
```
So it looks like it creates a file, marks it as a directory but with no other permissions 😅 so I got an access denied error when trying to write a file under it.

I then followed the permissions here for 0755 https://www.multacom.com/faq/password_protection/file_permissions.htm

The cool thing is that when using `fs.Mkdirall(path, 0755)` it automatically under the hood does: `fs.ModeDir | 0755` so you don't have to remember marking the file as a directory.
```
drwxr-xr-x  2 karakuse  staff    64B Jan 19 13:14 test
```

Lou1415926 (approved):
LGTM!

huanjani (approved):
🚢

PR #4361, opened 2023-01-12
chore: validate addons parameters before deployment

Lou1415926:
This PR validates the parameters in addon templates, as well as the customized parameters, before the CFN deployment. 

With the wrong addons parameter config, the CFN deployment fails nevertheless. With this PR, Copilot will error out earlier so that users don't have to wait until the deployment to know.

Note that this validation happens for both the workload addon, and the environment addon.

### DIFF
--- a/internal/pkg/addon/addons.go
+++ b/internal/pkg/addon/addons.go
@@ -66,7 +66,7 @@ type parser struct {
 	ws                 WorkspaceAddonsReader
 	addonsDirPath      func() string
 	addonsFilePath     func(fName string) string
-	validateParameters func(need, pass yaml.Node) error
+	validateParameters func(tplParams, customParams yaml.Node) error
 }
 
 // ParseFromWorkload parses the 'addon/' directory for the given workload
@@ -82,8 +82,8 @@ func ParseFromWorkload(workloadName string, ws WorkspaceAddonsReader) (*Workload
 		addonsFilePath: func(fName string) string {
 			return ws.WorkloadAddonFilePath(workloadName, fName)
 		},
-		validateParameters: func(need, pass yaml.Node) error {
-			return validateParameters(need, pass, wkldAddonsParameterReservedKeys)
+		validateParameters: func(tplParams, customParams yaml.Node) error {
+			return validateParameters(tplParams, customParams, wkldAddonsParameterReservedKeys)
 		},
 	}
 	stack, err := parser.stack()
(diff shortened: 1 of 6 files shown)


internal/pkg/addon/addons.go
```diff
@@ -66,7 +66,7 @@ type parser struct {
 	ws                 WorkspaceAddonsReader
 	addonsDirPath      func() string
 	addonsFilePath     func(fName string) string
-	validateParameters func(params yaml.Node) error
+	validateParameters func(need, pass yaml.Node) error
```
iamhopaul123: what do you think of renaming the param to be template and parameter or something like that? From the code it seems like `need` is the template file and `pass` is the parameter file

internal/pkg/addon/addons.go
```diff
@@ -230,22 +233,47 @@ func (p *parser) parseParameters(fNames []string) (yaml.Node, error) {
...
+func validateParameters(neededNode yaml.Node, passedNode yaml.Node, reservedKeys []string) error {
+	passed := make(map[string]yaml.Node)
+	if err := passedNode.Decode(passed); err != nil {
 		return fmt.Errorf("decode \"Parameters\" section of the parameters file: %w", err)
 	}
-	for _, key := range reservedKeys {
-		if _, ok := content[key]; ok {
+	needed := make(map[string]yaml.Node)
+	if err := neededNode.Decode(needed); err != nil {
+		return fmt.Errorf("decode \"Parameters\" section of the template file: %w", err)
```
iamhopaul123, line 246: Do we want to wrap the error msg differently for passed and needed?

internal/pkg/addon/addons.go
```diff
@@ -230,22 +233,47 @@ func (p *parser) parseParameters(fNames []string) (yaml.Node, error) {
...
+	}
+	// The reserved keys should present/be absent in the template/parameters file.
```
iamhopaul123:
```suggestion
	// The reserved keys should be present/absent in the template/parameters file.
```

iamhopaul123 (approved):
LGTM just nits

internal/pkg/addon/addons.go
Lou1415926: They are wrapped differently. For template, it ends with "...of the template file", whereas for paramters, it ends with "... of the parameters file". Or were you suggesting other changes?

internal/pkg/addon/addons.go
paragbhingre: +1

paragbhingre (approved):
I think naming is the only thing that confused me as Penghao mentioned. Otherwise, it looks good to me. Feel free to remove label once you decide on that.

PR #4365, opened 2023-01-12
feat: secrets enabled for rdws services

yohanthshetty:
## Overview:
In this PR, we are adding Secrets(ssm parameter store and secrets manager secrets) for rdws(app runner services).


internal/pkg/template/workload.go
```diff
@@ -751,12 +751,13 @@ type WorkloadOpts struct {
...
-	Private              bool
-	AppRunnerVPCEndpoint *string
-	Count                *string
+	StartCommand              *string
+	EnableHealthCheck         bool
+	Observability             ObservabilityOpts
+	Private                   bool
+	AppRunnerVPCEndpoint      *string
+	Count                     *string
+	RuntimeEnvironmentSecrets map[string]Secret
```
yohanthshetty: New variable RuntimeEnvironmentSecrets added.

internal/pkg/template/workload.go
```diff
@@ -849,21 +850,22 @@ func (t *Template) parseWkld(name, wkldDirName string, data interface{}, options
...
-			"jsonQueueURIs":        generateQueueURIJSON,
-			"envControllerParams":  envControllerParameters,
-			"logicalIDSafe":        StripNonAlphaNumFunc,
-			"wordSeries":           english.WordSeries,
-			"pluralWord":           english.PluralWord,
-			"contains":             contains,
-			"requiresVPCConnector": requiresVPCConnector,
+			"toSnakeCase":                  ToSnakeCaseFunc,
+			"hasSecrets":                   hasSecrets,
+			"hasRuntimeEnvironmentSecrets": hasRuntimeEnvironmentSecrets,
```
yohanthshetty: hasRuntimeEnvironmentSecrets parsing function to check if rdws service Manifest has secrets

internal/pkg/manifest/rd_web_svc.go
```diff
@@ -32,6 +32,7 @@ type RequestDrivenWebServiceConfig struct {
 	InstanceConfig                    AppRunnerInstanceConfig              `yaml:",inline"`
 	ImageConfig                       ImageWithPort                        `yaml:"image"`
 	Variables                         map[string]Variable                  `yaml:"variables"`
+	RuntimeEnvironmentSecrets         map[string]Secret                    `yaml:"secrets"`
```
iamhopaul123: Why not just 
```suggestion
	Secrets         map[string]Secret                    `yaml:"secrets"`
```

internal/pkg/aws/apprunner/service.go
```diff
@@ -33,6 +34,11 @@ type EnvironmentVariable struct {
 	Value string
 }
 
+type RuntimeEnvironmentSecret struct {
+	Name      string
+	ValueFrom string
```
iamhopaul123:
```suggestion
	Value string
```

internal/pkg/aws/apprunner/service.go
```diff
@@ -33,6 +34,11 @@ type EnvironmentVariable struct {
 	Value string
 }
 
+type RuntimeEnvironmentSecret struct {
```
iamhopaul123: Can we add doc string for public struct?
```suggestion
type EnvironmentSecret struct {
```

internal/pkg/template/workload.go
iamhopaul123: Can we remove `Runtime` throughout the PR? Or is there any reason why we want to add it?

Also why not reuse the existing `Secret` in `WorkloadOpts` for request driven web service (rdws)?

internal/pkg/template/templates/workloads/partials/cf/instancerole.yml
```diff
@@ -38,6 +38,25 @@ InstanceRole:
                 StringEquals:
                   'iam:ResourceTag/copilot-application': !Sub '${AppName}'
                   'iam:ResourceTag/copilot-environment': !Sub '${EnvName}'
+      - PolicyName: !Join ['', [!Ref AppName, '-', !Ref EnvName, '-', !Ref WorkloadName, SecretsPolicy]]
```
iamhopaul123: We might only want to include this additional policy if they have secrets enabled? aka `hasSecrets`

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
```diff
@@ -82,6 +82,20 @@ Resources:
           ImageRepositoryType: !Ref ImageRepositoryType
           ImageConfiguration:
             Port: !Ref ContainerPort
+            {{- if hasRuntimeEnvironmentSecrets .}}
+            RuntimeEnvironmentSecrets:
+            {{- range $name, $secret := .RuntimeEnvironmentSecrets}}
+            - Name: {{$name}}
+              Value: {{if not $secret.RequiresSub }}{{$secret.ValueFrom}}{{- else}} !Sub 'arn:${AWS::Partition}:{{$secret.Service}}:${AWS::Region}:${AWS::AccountId}:{{$secret.ValueFrom}}' {{- end }}
```
iamhopaul123: indentation needs to be consistent with variables section below and we don't need `-` because it doesn't create extra spaces as they are in the same line.
```suggestion
              - Name: {{$name}}
                Value: {{if not $secret.RequiresSub }}{{$secret.ValueFrom}}{{else}} !Sub 'arn:${AWS::Partition}:{{$secret.Service}}:${AWS::Region}:${AWS::AccountId}:{{$secret.ValueFrom}}' {{end }}
```

iamhopaul123 (commented):
Thank you for your contribution! Can we add unit tests and [local integ test](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/deploy/cloudformation/stack/rd_web_svc_integration_test.go) for this feature? Example integ test will be like [this](https://github.com/aws/copilot-cli/commit/41223e60243bd32460eb0ca32243201a34f3a8c4#diff-8a5e86d5e588a90aef0eb7fe24342476db01a0b639bb2a9cc7db23e9cc4505f1)

internal/pkg/template/templates/workloads/partials/cf/instancerole.yml
yohanthshetty: How do I include this? Withinn if statements?
{{- if hasSecrets .}}
    - PolicyName: !Join ['', [!Ref AppName, '-', !Ref EnvName, '-', !Ref WorkloadName, SecretsPolicy]]......
{{- end }}

iamhopaul123: yes something like that if you use `hasSecrets` for rdws as well.

internal/pkg/aws/apprunner/service.go
yohanthshetty: Added to new commit.

internal/pkg/aws/apprunner/service.go
yohanthshetty: Updated in recent commit

internal/pkg/manifest/rd_web_svc.go
yohanthshetty: Done

internal/pkg/template/templates/workloads/partials/cf/instancerole.yml
yohanthshetty: Done. Updated in next commit.

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
yohanthshetty: Done.

internal/pkg/template/workload.go
yohanthshetty: Ack. I have reused Secret

internal/pkg/aws/apprunner/service.go
```diff
@@ -33,6 +34,11 @@ type EnvironmentVariable struct {
 	Value string
 }
 
+type EnvironmentSecret struct {
```
yohanthshetty, line 38: // EnvironmentSecret contains the name and value of a Secret from SSM Parameter Store or Secrets Manager.
Will add the above doc string in next commit. Missed in this commit.

internal/pkg/template/templates/workloads/partials/cf/instancerole.yml
iamhopaul123: It's ok we'll squash and merge the PR after it gets approved, so don't worry about spamming commits.

iamhopaul123 (approved):
Awesome! This is much better :shipit:

internal/pkg/template/templates/workloads/services/rd-web/manifest.yml
```diff
@@ -65,6 +65,9 @@ network:
 #
 # tags:                         # Pass tags as key value pairs.
 #   project: project-name
+#
```
Lou1415926, line 68: Nice! Helpful for users to discover this feature 👍🏼

internal/pkg/template/templates/workloads/partials/cf/instancerole.yml
```diff
@@ -38,6 +38,27 @@ InstanceRole:
...
+            - Effect: 'Allow'
+              Action:
+                - 'ssm:GetParameters'
+              Resource:
+                - !Sub 'arn:${AWS::Partition}:ssm:${AWS::Region}:${AWS::AccountId}:parameter/*'
+            - Effect: 'Allow'
+              Action:
+                - 'secretsmanager:GetSecretValue'
+              Resource:
+                - !Sub 'arn:${AWS::Partition}:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:*'
```
Lou1415926, line 59: We probably want to add `Condition` to these policy statements like this: https://github.com/aws/copilot-cli/blob/4588b2227fc049a8364a19fcbc9f2c0ad604e697/internal/pkg/template/templates/workloads/partials/cf/executionrole.yml#L35

This way, Copilot will be authorized to fetch the SSM/secretsmanager value only if the SSM/secretsmanager is tagged with:
```
copilot-application: <appname>
copilot-environment: <envname>
```

So that Copilot is granted the minimal permissions that it needs

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
```diff
@@ -82,6 +82,25 @@ Resources:
           ImageRepositoryType: !Ref ImageRepositoryType
           ImageConfiguration:
             Port: !Ref ContainerPort
+            {{- if hasSecrets .}}
```
Lou1415926, line 85: Do you think we can reuse this [secrets partial file](https://github.com/aws/copilot-cli/blob/4588b2227fc049a8364a19fcbc9f2c0ad604e697/internal/pkg/template/templates/workloads/partials/cf/) by adding `{{include "secrets" . }}` here instead?  Or am I missing something  🤔 ?

internal/pkg/template/templates/workloads/services/rd-web/manifest.yml
```diff
@@ -65,6 +65,9 @@ network:
...
+# secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store and Secrets Manager.
+#  GITHUB_TOKEN: GITHUB_TOKEN  # The key is the name of the environment variable, the value is the ARN of the secret.
```
Lou1415926:
```suggestion
#  GITHUB_TOKEN: GITHUB_TOKEN  # The key is the name of the environment variable, the value is the name of an SSM parameter.
```

internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
```diff
@@ -56,6 +56,8 @@ environments:
       LOG_LEVEL: debug        # Log level for the "test" environment.
       DB_NAME:
         from_cfn: test-UserDB
+    secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store and Secrets Manager.
+      my-ssm-secret: "arn:aws:ssm:us-east-1:123456789123:parameter/jan11ssm"
```
Lou1415926, line 60: nit: Can we also do `from_cfn` and `secretsmanager` to see how it renders?
```suggestion
    secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store and Secrets Manager.
      my-ssm-secret: "arn:aws:ssm:us-east-1:123456789123:parameter/jan11ssm"
      MY_FROM_CFN_SECRET:
        from_cfn: my-other-stack-exported-secret
      MY_SECRETS_MANAGER_SECRET:
        secretsmanager: mysecretsmanagername
```

internal/pkg/describe/rd_web_service_test.go
```diff
@@ -143,6 +150,12 @@ func TestRDWebServiceDescriber_Describe(t *testing.T) {
 								Value: "test",
 							},
 						},
+						EnvironmentSecrets: []*apprunner.EnvironmentSecret{
+							{
+								Name:  "SOME_OTHER_SECRET",
+								Value: "arn:aws:ssm:us-east-1:111111111111:parameter/SHHHHH",
```
Lou1415926, line 156: 🤫

Lou1415926 (commented):
Thank you for your contribution! This looks great ❤️ I just have a few nits - after addressing them we can :shipit:

internal/pkg/deploy/cloudformation/stack/testdata/workloads/rdws-manifest.yml
yohanthshetty: Addressed in next commit. Renders fine!

internal/pkg/template/templates/workloads/partials/cf/instancerole.yml
yohanthshetty: I removed the condition for tags intentionally as we don't need these tags for secrets with App Runner service. Is it absolutely necessary to have them?

internal/pkg/template/templates/workloads/services/rd-web/cf.yml
yohanthshetty: Good call out. I have intentionally written this field definition in the cf.yml because for _RDWS_ service the CFN template should have **RuntimeEnvironmentSecrets** as the field heading instead of **Secrets**(which is the case for _ECS_ services). And also the Value field is **Value** for _RDWS_ service and **ValueFrom** for _ECS_ service.

internal/pkg/template/templates/workloads/services/rd-web/manifest.yml
yohanthshetty: Addressed in next commit.

internal/pkg/describe/rd_web_service_test.go
yohanthshetty: It's a separate test so I used two unique secrets for easier differentiation. Does not affect functionality in any way.

internal/pkg/describe/service.go
```diff
@@ -475,6 +475,26 @@ func (e containerEnvVars) humanString(w io.Writer) {
...
+	ValueFrom   string `json:"value"`
+}
+
+type envSecrets []*envSecret
+
+func (e envSecrets) humanString(w io.Writer) {
+	headers := []string{"Name", "Environment", "Value"}
+	var rows [][]string
+	sort.SliceStable(e, func(i, j int) bool { return e[i].Environment < e[j].Environment })
+	sort.SliceStable(e, func(i, j int) bool { return e[i].Name < e[j].Name })
```
Lou1415926: nit: perhaps calling `SliceStable` twice gives us the result that we want for now, but it might not in the future because go lib is free to change whatever algorithm they use under the hood. I think this ⬇️ should guarantee to get what's intended here
```suggestion
	sort.SliceStable(e, func(i, j int) bool { 
	  if e[i].Name == e[j].Name {
	     return  e[i].Environment < e[j].Environment
	  }
	return e[i].Name < e[j].Name 
	})
```

Lou1415926 (dismissed):
Looks great to me! I will approve the PR once the feedbacks are addressed in the next commit 👍🏼

internal/pkg/aws/apprunner/apprunner_test.go
```diff
@@ -77,6 +80,12 @@ func TestAppRunner_DescribeService(t *testing.T) {
 						Value: "info",
 					},
 				},
+				EnvironmentSecrets: []*EnvironmentSecret{
```
efekarakus, line 84: nit: can we add another value here to test the sorting? for example `parameter/zzz123` to make sure it appears after `jan11ssm`

internal/pkg/template/templates/workloads/partials/cf/instancerole.yml
```diff
@@ -38,6 +38,27 @@ InstanceRole:
                 StringEquals:
                   'iam:ResourceTag/copilot-application': !Sub '${AppName}'
                   'iam:ResourceTag/copilot-environment': !Sub '${EnvName}'
+      {{- if hasSecrets .}}
+      - PolicyName: !Join ['', [!Ref AppName, '-', !Ref EnvName, '-', !Ref WorkloadName, SecretsPolicy]]
```
efekarakus: I wonder if we are at risk of hitting the 128 char limit here for the policy name: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-policy.html#cfn-iam-policy-policyname

What do you think of simply:
```suggestion
        - PolicyName: 'AccessCopilotTaggedSecrets'
```

internal/pkg/template/templates/workloads/partials/cf/instancerole.yml
efekarakus: I can add context in the mean time and Wanxian can provide more data if needed.

It's necessary to have the tags otherwise we're giving v broad permissions to app runner services created with Copilot. Today, Copilot ensures that users opt-in to reading secrets by forcing them to add `copilot-application` and `copilot-environment` tags to their secrets (see https://aws.github.io/copilot-cli/docs/developing/secrets/). To make the process easy, the `copilot secret init` command automatically does that.

internal/pkg/describe/service.go
```diff
@@ -475,6 +475,26 @@ func (e containerEnvVars) humanString(w io.Writer) {
 	printTable(w, headers, rows)
 }
 
+// envSecret contains secrets for an rdws service.
+type envSecret struct {
```
efekarakus: Can we rename this struct to `rdwsSecret` since it's really similar to the existing `secret` type which also has an "environment" field:
https://github.com/aws/copilot-cli/blob/877347fa4f8f1112d7f5b7cb18c6726b3aa2d8e4/internal/pkg/describe/service.go#L478-L485

internal/pkg/aws/apprunner/apprunner_test.go
yohanthshetty: I added that, and checked. Test case Passed!

internal/pkg/template/templates/workloads/partials/cf/instancerole.yml
yohanthshetty: Ack.

internal/pkg/describe/service.go
yohanthshetty: Updated in next commit.

internal/pkg/template/templates/workloads/partials/cf/instancerole.yml
yohanthshetty: As discussed, I have added the Condition to check for tagged secrets.

internal/pkg/describe/service.go
yohanthshetty: Done!

efekarakus (approved):

paragbhingre:
#4506

PR #4390, opened 2023-01-20
fix: sc shouldn't affect gov cloud deployment

### LINKED ISSUE #4353: copilot-cli v1.24.0 causes AmazonECS ServerException when deploying
Running in *us-gov-west-1*, this command successfully deployed in copilot-cli v1.23.0:

`TAG=7dcd52058990a783e68906f0dba541fce373c2fc copilot deploy --env test --name web`

Now, with copilot-cli v1.24.0:

```
❯ TAG=7dcd52058990a783e68906f0dba541fce373c2fc copilot deploy --env test --name web
✔ Proposing infrastructure changes for stack [REDACTED]
- Updating the infrastructure for stack [REDACTED]                          [update rollback complete]  [26.6s]
  The following resource(s) failed to update: [Service].                                                         
  - An ECS service to run and maintain your tasks in the environment cluster         [update complete]           [1.8s]
    Resource handler returned message: "Service was unable to process this                                       
     request. (Service: AmazonECS; Status Code: 500; Error Code: ServerExc                                       
    eption; Request ID: 1157666c-74e1-4fe8-ba75-eb43fc9f0ff5; Proxy: null)                                       
    " (RequestToken: 6ca35636-d8cf-0fcc-13eb-b744abc4de41, HandlerErrorCod                                       
    e: GeneralServiceException)                                                                                  
    Deployments                                                                                                   
               Revision  Rollout      Desired  Running  Failed  Pending                                                   
      PRIMARY  11        [completed]  1        1        0       0                                                         
  - An ECS task definition to group your containers and run them on ECS              [delete complete]           [0.0s]
```

It's interesting that with v1.24.0 it doesn't even appear to advance the ECS task revision... it just shows the current version (in this case, 11, is already running).

web/manifest.yml
```
name: web

http:
  path: '/'
  healthcheck: '/health'

image:
  location: [REDACTED]:${TAG}
  port: 8000

cpu: 256
memory: 1024
platform: linux/x86_64
count:
  spot: 1
exec: true
entrypoint: ["gunicorn", "-b", "0.0.0.0:8000", "--workers", "4", [REDACTED], "--timeout", "120"]

environments:
  test:
    http:
      alias:
      [REDACTED]
    variables:
      [REDACTED]

    secrets:
      [REDACTED]
network:
  vpc:
    placement: 'private'
```

environments/test/manifest.yml
```
name: test
type: Environment

network:
  vpc:
    id: [REDACTED]
    subnets:
      public:
        - id: [REDACTED]
        - id: [REDACTED]
      private:
        - id: [REDACTED]
        - id: [REDACTED]

http:
  public:
    certificates:
      [REDACTED]

observability:
  container_insights: false
```

iamhopaul123:
fixes #4353

### DIFF
--- a/internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
+++ b/internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
@@ -76,7 +76,10 @@ ServiceConnectConfiguration:
           {{- end}}
   {{- end}}
   {{- else}}
-  Enabled: False
+  !If
+    - IsGovCloud
+    - !Ref AWS::NoValue
+    - Enabled: False
   {{- end}}
 NetworkConfiguration:
   AwsvpcConfiguration:
--- a/internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/http-autoscaling-template.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/http-autoscaling-template.yml
@@ -43,6 +43,7 @@ Parameters:
     Type: String
     Default: false
 Conditions:
+  IsGovCloud: !Equals [!Ref "AWS::Partition", "aws-us-gov"]
   HasAddons: !Not [!Equals [!Ref AddonsTemplateURL, ""]]
   HasEnvFile: !Not [!Equals [!Ref EnvFileARN, ""]]
   ExposePort: !Not [!Equals [!Ref TargetPort, -1]]
@@ -411,8 +412,10 @@ Resources:
       PropagateTags: SERVICE
       EnableExecuteCommand: true
       LaunchType: FARGATE
-      ServiceConnectConfiguration:
-        Enabled: False
+      ServiceConnectConfiguration: !If
+        - IsGovCloud
+        - !Ref AWS::NoValue
+        - Enabled: False
       NetworkConfiguration:
         AwsvpcConfiguration:
           AssignPublicIp: ENABLED
--- a/internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/http-full-config-template.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/http-full-config-template.yml
@@ -43,6 +43,7 @@ Parameters:
     Type: String
     Default: false
 Conditions:
+  IsGovCloud: !Equals [!Ref "AWS::Partition", "aws-us-gov"]
   HasAddons: !Not [!Equals [!Ref AddonsTemplateURL, ""]]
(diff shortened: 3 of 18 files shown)


Lou1415926 (approved):
Thanks boss for the quick turnaround!

Lou1415926 (commented):
Oh anything we need to change in those slow integ tests tagged with `//go:build integration`?

paragbhingre (approved):
🚢
Feel free to remove label after addressing Wanxian's comment.

efekarakus (approved):
Thank you boss!

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
```diff
@@ -76,7 +76,10 @@ ServiceConnectConfiguration:
           {{- end}}
   {{- end}}
   {{- else}}
-  Enabled: False
+  !If
+    - IsGovCloud
+    - !Ref AWS::NoValue
+    - Enabled: False
```
efekarakus, line 82: dumb question but instead of a condition with `!Ref AWS::NoValue` why not do:
```go
{{if .ServiceConnect}}
ServiceConnectConfiguration:

{{- end}}
```
This way we would get rid of an extra CFN condition?

iamhopaul123: Why is it related to the gov-cloud 🤔 Like we need to determine if they are using gov-cloud or not.

efekarakus: I think that's my point, wouldn't moving the if-conditional above the `ServiceConnectConfiguration` help not have to worry about any partitions?

For all partitions there would be the same behavior of No:Value if the user doesn't have `connect: true` in their manifest

iamhopaul123: oooh but getting rid of `ServiceConnectConfiguration` wouldn't help you switch from enabling service connect to disabling is the reason behind, unless you explicitly specify `Enabled: False`.

efekarakus: ooh what! that's surprising to me 🤔 my guess would have been that service connect gets disabled if the config is deleted

iamhopaul123: Yeah that's how ecs update-service api works :)

efekarakus: 🤦 that makes sense

PR #4393, opened 2023-01-20
chore(templates): write ingress as an RDWS workload addon for environment-level aurora

Lou1415926:
For an RDWS to access an environment-level aurora serverless cluster, we need to attach to `ServiceSecurityGroup` with an ingress for access to the cluster.

This ingress is created as a hook resource in the RDWS's workload-level addon.


internal/pkg/template/templates/addons/aurora/env/rdws/ingress.yml
```diff
@@ -0,0 +1,28 @@
+Parameters:
+  App:
+    Type: String
+    Description: Your application's name.
+  Env:
+    Type: String
+    Description: The environment name your service, job, or workflow is being deployed to.
```
huanjani: Q: should this wording be different for RDWS?

internal/pkg/template/templates/addons/aurora/env/rdws/ingress.yml
```diff
@@ -0,0 +1,28 @@
...
+      ToPort: {{- if eq .Engine "MySQL" }} 3306 {{- else}} 5432 {{- end}}
+      FromPort: {{- if eq .Engine "MySQL" }} 3306 {{- else}} 5432 {{- end}}
+      GroupId:
+        Fn::ImportValue: !Sub "${App}-${Env}-{{logicalIDSafe .ClusterName}}DBClusterSecurityGroup"
+      IpProtocol: tcp
+      SourceSecurityGroupId: !Ref ServiceSecurityGroupId
+Outputs:
+  {{logicalIDSafe .ClusterName}}AuroraServerlessSecret: # injected as {{ logicalIDSafe .ClusterName | printf "%sAuroraServerlessSecret" | envVarSecret | toSnakeCase}} environment variable by Copilot.
+    Description: "The JSON secret that holds the database username and password. Fields are 'host', 'port', 'dbname', 'username', 'password', 'dbClusterIdentifier' and 'engine'"
+    Value: { Fn::ImportValue: { Fn::Sub: "${App}-${Env}-{{logicalIDSafe .ClusterName}}AuroraSecret" }}
```
efekarakus: Interesting 🤔 

Shouldn't the customer use:
```yaml
secrets:
   DB_SECRET:
      secretsmanager:
         from_cfn: ${COPILOT_APP_NAME}-${COPILOT_ENV_NAME}-ClusterNameAuroraSecret
```

If I'm understanding correctly, the output in this PR will ony inject the ARN of the secretsmanager secret as an env var and the user will have to fetch the secret within the container

internal/pkg/template/templates/addons/aurora/env/rdws/serverlessv2.yml
```diff
@@ -0,0 +1,154 @@
...
+    Type: String
+    Description: The name of the initial database to be created in the Aurora Serverless v2 cluster.
+    Default: {{.InitialDBName}}
+    # Cannot have special characters
+    # Naming constraints: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Limits.html#RDS_Limits.Constraints
+  VPCID:
+    Type: String
+    Description: The ID of the VPC in which to create the Aurora Serverless v2 cluster.
+    Default: ""
+  PrivateSubnets:
```
efekarakus, line 19: Do we need the accompanying `addons.parameters.yml` file or is that already added in a previous PR?

internal/pkg/template/templates/addons/aurora/env/rdws/ingress.yml
Lou1415926: I think with @yohanthshetty's recent change, the injection is possible now 🤔 https://github.com/aws/copilot-cli/blob/a60cf0fbac6a339dc6e98019258fa70159a9359b/internal/pkg/template/templates/workloads/services/rd-web/cf.yml#L98

internal/pkg/template/templates/addons/aurora/env/rdws/serverlessv2.yml
Lou1415926: Both `env/serverlessv2.yml` and `env/rdws/serverlessv2/yml` share the same `env/addons.parameters.yml` file:
https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/addons/aurora/env/addons.parameters.yml

internal/pkg/template/templates/addons/aurora/env/rdws/ingress.yml
Lou1415926: done!

internal/pkg/template/templates/addons/aurora/env/serverlessv2.yml
```diff
@@ -51,7 +48,7 @@ Resources:
       VpcId: !Ref VPCID
       Tags:
         - Key: Name
-          Value: !Sub 'copilot-${App}-${Env}-AuroraCluster'
```
iamhopaul123, line 54: Why do we remove `copilot`?

iamhopaul123 (approved):

internal/pkg/template/templates/addons/aurora/env/serverlessv2.yml
Lou1415926: Because I thought they were unnecessary. However the other secuirty groups (e.g. EnvironmentSecurityGroup)'s tags also have the `copilot-` prefix. So I added them back.

huanjani (approved):
LGTM! Added the DNM label in case @efekarakus had any more questions/comments.

internal/pkg/template/templates/addons/aurora/env/rdws/ingress.yml
efekarakus: My brain might not be working, iirc `NestedStack.SecretOutputs` are discovered as SecretsManager secrets by searching for the logical ID and checking the Type of the resource. Since this is an import, we won't be able to find that 🤔 

Let me know if I'm missing something!

Lou1415926: good call boss it's my brain that is not working 😂 

I am going to remove this `Output` for this PR. In one of the next PR that handles `RecommendedActions`, I will recommend a step to inject the secret from manifest:
```
Recommended follow-up actions:
- Run `copilot env deploy` to deploy your environment addons, along with the environment. 
- Update the manifest for your `api` service:

secrets:
  USER_DB_SECRET:
    from_cfn: ${COPILOT_APPLICATION_NAME}-${COPILOT_ENVIRONMENT_NAME}-UserDBSecret

```

efekarakus (approved):

PR #4402, opened 2023-01-23
fix: render env addons additional params with correct indentation

Lou1415926:
Previously, when there are more than one additional params, the indentation is off. This fixes that.


huanjani (approved):

efekarakus (approved):
:mildpanic:

PR #4403, opened 2023-01-23
chore: add `Conditions` and `Parameters` for imported VPC

Lou1415926:
The `Conditions` and `Parameters` can be used in `addons.parameters.yml`. 

The scenario that this unblocks: e.g. users has `test` and `prod`, where `test` has managed VPC and `prod` has imported VPC. Just having `VPCID: !Ref VPC` in `addons.parameters.yml` would lead to CFN validation error in `prod`.

Without this PR, there are only two ways for users to work
1. By having both `test.addons.parameters.yml` and `prod.addons.parameter.yml`, and copy the file's content to `addons.parameters.yml` when deploying to the respective environment.
2. By having a `Mappings` section in the addon YAML, hard-coding all VPC IDs.

Neither is ideal.


internal/pkg/template/templates/addons/aurora/env/addons.parameters.yml
```diff
@@ -1,9 +1,9 @@
 Parameters:
   VPCID: 
-    - !If ImportedVPC
+    - !If IsVPCImported
     - !Ref ImportedVPCID
```
paragbhingre, line 4: qq: Previously, from where do we used to pull the value of this `ImportedVPCID` param?

Lou1415926: Here: https://github.com/aws/copilot-cli/blob/710bc41fb8a0a8ad1bab8986d8e75c8b8557c617/internal/pkg/template/templates/addons/aurora/env/addons.parameters.yml

It is not actually used in any recent release because the template isn't used yet.

huanjani (approved):
LGTM

paragbhingre (approved):
LGTM 🚢

internal/pkg/template/templates/environment/cf.yml
```diff
@@ -38,7 +38,12 @@ Parameters:
     AllowedValues: [true, false]
   ServiceDiscoveryEndpoint:
     Type: String
+  ImportedVPCID:
+    Type: String
+  ImportedPrivateSubnets:
+    Type: String
```
efekarakus, line 44: Hmmm, what do you think of a different alternative where we do option2. but generate it for the user.

i.e. when a user creates an Addon, we can create a `mappings.yml` file that contains a Mappings like:
```yaml
Mappings:
  VpcId:
      test: vpc-12345
      prod: vpc-78912
```

My intuition is that we want stack parameters to be generic Parameters rather than specific like `ImportedVPCID`.
For example, we can have just a `VpcId` parameter where:
- If empty, we create the VPC.
- Otherwise, we use the value.


So I'd be OK with either generating the `mappings` file or using flexible Parameters.

Lou1415926:
```
Mappings:
  VpcId:
      test: vpc-12345
      prod: vpc-78912
```
People would need to manage an imported VPC IDs in both the env manifests and ⬆️.

> flexible Parameters

Just making sure I'm on the same page with ya - 

In addition to renaming `ImportedVPCID` to `VpcId`, we want to change the creation of VPC-related resources from using `go-template` rendering to using CFN `Conditions` - is this what this proposal implies? I'm down to do so

efekarakus: >  is this what this proposal implies?

Yeah that's what I meant, but to be honest I am more aligned with @iamhopaul123 's recommendation of not doing anything until a customer asks for it because there is a mitigation path

iamhopaul123: Yeah after a second thought i feel like it's probably fine to have users who import VPC/subnets to hardcode the ID. It's not a very popular use case to import VPC/subnets and use them in addon stack at the same time, but adding conditions and params is a one-way door 🤔

PR #4404, opened 2023-01-24
chore(describe): add rollback alarms to `svc status`

huanjani:
This implementation was hindered by the fact that [there's no way to tag alarms via CFN](https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/64). So rather than fetching alarms by tags, here we fetch with the Copilot-generated alarm name prefix ("`appName`-`envName`-`svcName`-CopilotRollback"). 

The new "Type" column in the describe output specifies if an alarm is an Auto Scaling alarm or a Rollback Alarm. Adding a column seemed more clear than creating two subsections, but if it's too wide, we can try something else. The preexisting "Type" options ("Metric" and "Composite") were not output. 

Screen Shot 2023-01-20 at 12 33 54 PM

 


internal/pkg/describe/status_describe.go
```diff
@@ -197,6 +207,21 @@ func (s *ecsStatusDescriber) ecsServiceAutoscalingAlarms(cluster, service string
 	if err != nil {
 		return nil, fmt.Errorf("get auto scaling CloudWatch alarms: %w", err)
 	}
+	for i := range alarms {
+		alarms[i].Type = autoscalingAlarmType
```
paragbhingre, line 224: Why are we setting all the alarm types to be `autoscalingAlarmType`?

huanjani: So it's easier to distinguish between Auto Scaling alarms and alarms for deployment rollbacks.

internal/pkg/aws/cloudwatch/cloudwatch.go
```diff
@@ -101,6 +101,31 @@ func (cw *CloudWatch) AlarmStatus(alarms []string) ([]AlarmStatus, error) {
...
+		alarmResp, err = cw.client.DescribeAlarms(&cloudwatch.DescribeAlarmsInput{
+			AlarmNamePrefix: aws.String(prefix),
+			NextToken:       alarmResp.NextToken,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("describe CloudWatch alarms: %w", err)
+		}
+		if alarmResp == nil {
+			return nil, nil
+		}
```
iamhopaul123: Should we `break` instead?

internal/pkg/aws/cloudwatch/cloudwatch.go
```diff
@@ -101,6 +101,31 @@ func (cw *CloudWatch) AlarmStatus(alarms []string) ([]AlarmStatus, error) {
 	return alarmStatus, nil
 }
 
+// AlarmStatusesFromNamePrefix returns the status of all alarms whose names begin with the given prefix.
+func (cw *CloudWatch) AlarmStatusesFromNamePrefix(prefix string) ([]AlarmStatus, error) {
```
iamhopaul123: Can we remove duplication with `AlarmStatus` by defining something like

```go
func (cw *CloudWatch) AlarmStatus(options AlarmStatusOption) ([]AlarmStatus, error)
```

and caller can call either
```go
cw.AlarmStatus(WithNames(names))
```
or
```go
cw.AlarmStatus(WithPrefix(fmt.Sprintf("%s-%s-%s-CopilotRollback", app, env, svc)))
```
Under the hood both` WithPrefix` and `WithNames` update `DescribeAlarmsInput`.

internal/pkg/aws/cloudwatch/cloudwatch.go
huanjani: Is there any benefit to using `break` instead? I figured `return` exits the entire method earlier....

iamhopaul123: Break returns `alarmStatuses` and `nil`, which is the same as returning `alarmStatuses` and `nil` here. One scenario is `alarmResp` was not `nil` before and `alarmStatuses` got some values already, and in one page loop however, we get a `nil` `alarmStatuses` and in your current logic, it will return `nil` and `nil`.

huanjani: I didn't think that scenario was possible, but sure, I can change it just in case!

internal/pkg/aws/cloudwatch/cloudwatch.go
```diff
@@ -73,32 +73,57 @@ func (cw *CloudWatch) AlarmsWithTags(tags map[string]string) ([]AlarmStatus, err
...
+func (cw *CloudWatch) AlarmStatuses(opts ...DescribeAlarmOpts) ([]AlarmStatus, error) {
+	if len(opts) == 0 {
 		return nil, nil
 	}
-	var alarmStatus []AlarmStatus
+	var alarmStatuses []AlarmStatus
 	var err error
 	alarmResp := &cloudwatch.DescribeAlarmsOutput{}
+	in := &cloudwatch.DescribeAlarmsInput{
+		NextToken: alarmResp.NextToken,
```
Lou1415926: Why do we need to assign `NextToken` here?

internal/pkg/aws/cloudwatch/cloudwatch.go
```diff
@@ -73,32 +73,57 @@ func (cw *CloudWatch) AlarmsWithTags(tags map[string]string) ([]AlarmStatus, err
...
-			AlarmNames: aws.StringSlice(alarms),
-			NextToken:  alarmResp.NextToken,
-		})
+		alarmResp, err = cw.client.DescribeAlarms(in)
 		if err != nil {
 			return nil, fmt.Errorf("describe CloudWatch alarms: %w", err)
 		}
-		alarmStatus = append(alarmStatus, cw.compositeAlarmsStatus(alarmResp.CompositeAlarms)...)
-		alarmStatus = append(alarmStatus, cw.metricAlarmsStatus(alarmResp.MetricAlarms)...)
+		if alarmResp == nil {
```
Lou1415926, line 110: Is it possible for both `alarmResp` and `err` to be `nil`?

internal/pkg/aws/cloudwatch/cloudwatch_test.go
```diff
@@ -353,3 +354,159 @@ func TestCloudWatch_AlarmsWithTags(t *testing.T) {
 
 	}
 }
+
+func TestCloudWatch_AlarmStatusesFromNamePrefix(t *testing.T) {
```
Lou1415926: nit: since we are not adding a new exported func now, we can probably just add some of the unit test cases to `TestCloudWatch_AlarmStatuses` instead of having a new test suite!

internal/pkg/describe/status_describe.go
```diff
@@ -18,15 +18,19 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/ecs"
 )
 
-const fmtAppRunnerSvcLogGroupName = "/aws/apprunner/%s/%s/service"
+const (
+	fmtAppRunnerSvcLogGroupName = "/aws/apprunner/%s/%s/service"
+	autoscalingAlarmType        = "Auto Scaling"
```
Lou1415926, line 23: What about "Scaling"? Just an idea to make the column less wide 😂

internal/pkg/describe/status_describe.go
```diff
@@ -156,13 +160,29 @@ func (s *ecsStatusDescriber) Describe() (HumanJSONStringer, error) {
...
+	}
+	rollbackAlarms, err := s.ecsServiceRollbackAlarms(s.app, s.env, s.svc)
+	if err != nil {
+		return nil, err
+	}
+	for _, alarm := range rollbackAlarms {
+		alarms[alarm.Name] = alarm
+	}
+	alarmList := make([]cloudwatch.AlarmStatus, len(alarms))
+	i := 0
```
Lou1415926: nit: ⬇️ should initialize `i` as `0`
```suggestion
	var i int
```

internal/pkg/describe/status_describe.go
```diff
@@ -193,10 +213,25 @@ func (s *ecsStatusDescriber) ecsServiceAutoscalingAlarms(cluster, service string
...
 		return nil, fmt.Errorf("get auto scaling CloudWatch alarms: %w", err)
 	}
+	for i := range alarms {
+		alarms[i].Type = autoscalingAlarmType
+	}
+	return alarms, nil
+}
+
+func (s *ecsStatusDescriber) ecsServiceRollbackAlarms(app, env, svc string) ([]cloudwatch.AlarmStatus, error) {
+	// This will not fetch imported alarms, as we filter by the Copilot-generated prefix of alarm names. This will also not fetch Copilot-generated alarms with names exceeding 255 characters, due to the balanced truncating of `TruncateAlarmName`.
```
Lou1415926, line 230: [`TruncateAlarmName`](https://github.com/aws/copilot-cli/blob/dffeb8d52b66659e7eb81bfb194846af233ca6c7/internal/pkg/template/workload.go#L526) seems to produce deterministic results. Is is possible for us to fetch the managed rollback alarms by their exact names?

internal/pkg/aws/cloudwatch/cloudwatch.go
huanjani: I had added it at some point bc it seemed necessary ... but whatever behavior it was preventing is no longer happening without it! Deleted.

internal/pkg/aws/cloudwatch/cloudwatch.go
huanjani: If there are no alarms, I'd expect `DescribeAlarms` to return `nil, nil`.

internal/pkg/aws/cloudwatch/cloudwatch_test.go
huanjani: Good idea. There isn't currently a `TestCloudWatch_AlarmStatuses` test suite-- they're bundled into `AlarmsWithTags`, so I'll separate out and combine.

internal/pkg/describe/status_describe.go
huanjani: Part of the problem is that the type of alarm is included in the alarm name, so if we're fetching without a specific alarm type, we can't truncate in the same way as `TruncateAlarmName`. I assume very few alarm names will actually require truncation, as 255 characters is quite long.

internal/pkg/describe/status_describe.go
huanjani: I understand your concern! But I don't think "Scaling" is clear enough, unfortunately.

internal/pkg/aws/cloudwatch/cloudwatch.go
```diff
@@ -73,32 +73,53 @@ func (cw *CloudWatch) AlarmsWithTags(tags map[string]string) ([]AlarmStatus, err
...
+// WithPrefix sets DescribeAlarms to filter on a name prefix.
+func WithPrefix(prefix string) DescribeAlarmOpts {
+	return func(in *cloudwatch.DescribeAlarmsInput) {
+		in.AlarmNamePrefix = aws.String(prefix)
+	}
+}
+
+// AlarmStatuses returns the status of each given alarm name.
+func (cw *CloudWatch) AlarmStatuses(opts ...DescribeAlarmOpts) ([]AlarmStatus, error) {
+	if len(opts) == 0 {
```
iamhopaul123: If so why not make required ones in param? For example if name is required

```golang
AlarmStatuses(name string, opts ...DescribeAlarmOpts)
```

huanjani: Because we aren't searching for specific alarms by name; does that make sense?

iamhopaul123: ah i see so this function takes at least one param. Should we return an error then? Or i wonder if there is a valid use case for this function to call without any param and return all the alarm statuses?

huanjani: In theory there could be a future use case to return all statuses! 🤷🏼

iamhopaul123: Yeah I think that'll make more sense to me too. Can we remove this condition for now?

iamhopaul123 (approved):

iamhopaul123 (approved):

internal/pkg/aws/cloudwatch/cloudwatch_test.go
```diff
@@ -81,23 +75,129 @@ func TestCloudWatch_AlarmsWithTags(t *testing.T) {
...
+					NextToken:       nil,
+				}).Return(nil, nil)
+			},
+		},
+		"success with prefix": {
+			in: WithPrefix(mockPrefix),
+			setupMocks: func(m cloudWatchMocks) {
+				m.cw.EXPECT().DescribeAlarms(&cloudwatch.DescribeAlarmsInput{
+					AlarmNamePrefix: aws.String(mockPrefix),
+					NextToken:       nil,
```
Lou1415926: nit (and other places as well where assigning `nil` is redundant)
```suggestion
```

Lou1415926 (approved):
feel free to remove the label whenver!

PR #4405, opened 2023-01-24
feat: enable cdk overrides on `svc deploy`

efekarakus:


internal/pkg/deploy/cloudformation/cloudformation_test.go
```diff
@@ -23,6 +23,43 @@ import (
...
+		// GIVEN
+		var stack StackConfiguration = &mockStackConfig{template: "hello"}
+		ovrdr := &mockOverrider{err: errors.New("some error")}
+
+		// WHEN
+		stack = WrapWithTemplateOverrider(stack, ovrdr)
+		_, err := stack.Template()
+
+		// THEN
+		require.ErrorContains(t, err, "override template:")
```
Lou1415926: tiniest nit: what do you think of  `require.ErrorContains(t, err, "override template: some error")` since we mocked the error at L52?

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -55,6 +55,38 @@ type StackConfiguration interface {
...
+}
+
+// overridableStack is a StackConfiguration with overrides applied.
+type overridableStack struct {
+	StackConfiguration
+	overrider Overrider
+}
+
+// Template returns the overriden CloudFormation stack template.
+func (s *overridableStack) Template() (string, error) {
```
Lou1415926, line 70: I feel like there is a long-due refactor that we've been wanting to do... that is, `copilot svc deploy` should under the hood go through the same call stack as `copilot svc package` to generate a template. The only difference should be that `svc deploy` go one step ahead and deploy the template.

If, theoretically, that the refactor has been done, then we probably'd end up doing this in `cli/deploy`
```go
func (d *wkldDeployer) DeployWorkload() {
  tmpl, err := d.generateTemplate()
  err := d.deployer.DeployWorkload(tmpl)
}

func (d *wkldDeployer) GenerateCloudFormationTemplate() {
   tmpl, err := d.generateTemplate()
}

func (d *wkldDeployer) generateTemplate(conf stackSerializer) (string, error) {
   tpl, err := conf.Template()
   if err != nil {
      return nil, fmt.Errorf("generate stack template: %w", err)
   }
   transformedTpl, err := d.overrider.Override([]byte(tpl))
   if err != nil {
	   return nil, fmt.Errorf("override template: %w", err)
   }
   return transformedTpl, nil
}
```


I think it makes sense to pass overrider to `cloudformation`, but I also wonder if we wouldn't have done so if we've had the refactor.

Lou1415926 (commented):
LGTM!! Just a tiny nit and a random thought that doesn't need to be addressed rn.

internal/pkg/deploy/cloudformation/cloudformation_test.go
efekarakus: Done!

internal/pkg/deploy/cloudformation/cloudformation.go
efekarakus: Totally agree 😍 although `d.deployer.DeployWorkload(tmpl)` would require a bunch of additional parameters like the CFN Parameters (all the other methods in `StackConfig`) but I agree that direction would be awesome.

Like `deploy/cloudformation` could take an `io.Writer` for the progress tracker and these params but I don't think I'll be able to get to it part of this task :(

Lou1415926 (approved):
🚢

iamhopaul123 (approved):

PR #4408, opened 2023-01-24
chore: write `addons.parameters.yml` for env aurora that works for managed VPC

Lou1415926:
If people have some envs with imported VPC, we recommend them to use `Mappings`.

### DIFF
--- a/internal/pkg/template/templates/addons/aurora/env/addons.parameters.yml
+++ b/internal/pkg/template/templates/addons/aurora/env/addons.parameters.yml
@@ -1,9 +1,3 @@
 Parameters:
-  VPCID: 
-    - !If ImportedVPC
-    - !Ref ImportedVPCID
-    - !Ref VPC
-  PrivateSubnets:
-    - !If ImportedVPC
-    - !Join [ ',', [ !Ref ImportedPrivateSubnet1, !Ref ImportedPrivateSubnet2 ] ]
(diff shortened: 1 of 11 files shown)


iamhopaul123 (approved):
🙇‍♂️

internal/pkg/template/templates/addons/aurora/env/addons.parameters.yml
```diff
@@ -1,9 +1,3 @@
 Parameters:
-  VPCID: 
-    - !If IsVPCImported
-    - !Ref ImportedVPCID
-    - !Ref VPC
-  PrivateSubnets:
-    - !If IsVPCImported
-    - !Ref ImportedPrivateSubnets
-    - !Join [ ',', [ !Ref PrivateSubnet1, !Ref PrivateSubnet2 ] ]
+  VPCID: !Ref VPC
```
efekarakus, line 2: dumb question, do we need to update the CF templates as well in `aurora/env/serverlessv2.yml` with these parameters?

Lou1415926: No need! Those templates just make use of the `VPCID` parameter like this
 https://github.com/aws/copilot-cli/blob/b69393aa9e504ac3b398c53adef360e3f1ab0fb0/internal/pkg/template/templates/addons/aurora/env/serverlessv2.yml#L51

efekarakus (approved):

PR #4409, opened 2023-01-24
fix(custom-domain-action): ignore A-record deletion if not found

### LINKED ISSUE #4401: HTTP Alias Issue left HTTPSCert and CustomDomainAction stuck in UPDATE_FAILED
Hello, and thank you for your work on copilot!

I've ben experimenting with running some workloads via copilot, and have run into an issue that I seem stuck on...

For a load balanced web service, I attempted to add another http.alias value. The deploy failed as there was already a dns record for that subdomain created previously. I attempted two things:
1. changing the existing route53 record to a different subdomain
2. deleting the existing route53 record

Neither of those worked, and the only way to get it to redeploy succesfully was removing the new alias.

What I noticed, is that some components of the parent env stack are now stuck at UPDATE_FAILED:

image

I tried to use the "continue update rollback" feature in cloudformation, which seemed to resolve for the HTTPSCert, but the CustomDomainAction is still continually failing (even though the env seems to be all deployed and in a healthy state).

I see that the custom domain action is failing because it is trying to delete the dns record that it failed to create in the first place, but it doesn't exist. `Received response status [FAILED] from custom resource. Message returned: [Tried to delete resource record set [name='subdomain.example.com.', type='A'] but it was not found]` 

Is there some way around this? Also, is there any particular procedure I should follow for adding an alias where there might be a recently deleted route53 record? Perhaps I need to delete some other dns related resources for that subdomain, wait for the dns ttl, etc.?

efekarakus:
Fixes #4401

If the record fails to insert, then the custom resource fails. On failure, it tries to delete the record that was *not* inserted. Hence the Delete workflow fails resulting in the custom resource to fail to delete.

This change ensures that we ignore the error if the A-record that we're trying to delete doesn't exist in the first place.


cf-custom-resources/lib/custom-domain.js
```diff
@@ -158,17 +163,27 @@ const writeARecord = async function (
...
+  } catch (err) {
+    if (action === changeRecordAction.Delete && isRecordSetNotFoundErr(err)) {
+      console.log(`${err.message}; Copilot is ignoring this record.`);
+      return;
+    }
+    throw err;
+  }
 };
 
+const isRecordSetNotFoundErr = (err) => err.message.includes("Tried to delete resource record set") && err.message.includes("but it was not found")
```
iamhopaul123, line 186: Maybe in the comment include an example err msg?

iamhopaul123 (approved):

cf-custom-resources/lib/custom-domain.js
```diff
@@ -158,17 +163,28 @@ const writeARecord = async function (
...
+      hostedZoneId,
+      action,
+      alias,
+      accessDNS,
+      accessHostedZone
+    );
+    await waitForRecordChange(route53, changeBatch.ChangeInfo.Id);
+  } catch (err) {
+    if (action === changeRecordAction.Delete && isRecordSetNotFoundErr(err)) {
+      console.log(`${err.message}; Copilot is ignoring this record.`);
```
paragbhingre, line 178: How about keeping `Copilot is ignoring this record.` at the beginning so that customers see our message first before seeing long less user-friendly error?

paragbhingre: Ahh I see the exact error message below which is also not bad. It is pretty clear. You can skip the above comment.

cf-custom-resources/test/custom-domain-test.js
```diff
@@ -406,4 +406,46 @@ describe("DNS Validated Certificate Handler", () => {
...
+          DomainName: testDomainName,
+          Aliases: testAliases,
+          Region: "us-east-1",
+          PublicAccessDNS: testAccessDNS,
+          PublicAccessHostedZone: testLBHostedZone,
+          AppDNSRole: testRootDNSRole,
+        },
+      })
+      .expectResolve(() => {
+        expect(request.isDone()).toBe(true);
```
paragbhingre, line 448: Can we actually see somewhere here if our custom error message is being logged?

paragbhingre (approved):
Looks good to me. Feel free to remove the label if there is nothing to change.

cf-custom-resources/test/custom-domain-test.js
efekarakus: It's logged in CloudWatch ! so in case this issue happens, customers will be able to navigate to the logs of the lambda function and see the message there :)

PR #4416, opened 2023-01-25
chore(addon): construct marshalers for env-level addons

Lou1415926:

### DIFF
--- a/internal/pkg/addon/storage.go
+++ b/internal/pkg/addon/storage.go
@@ -23,8 +23,10 @@ const (
 	rdsRDWSV2TemplatePath = "addons/aurora/rdws/serverlessv2.yml"
 	rdsRDWSParamsPath     = "addons/aurora/rdws/addons.parameters.yml"
 
-	envS3TemplatePath             = "addons/s3/env/cf.yml"
-	envS3AccessPolicyTemplatePath = "addons/s3/env/access_policy.yml"
+	envS3TemplatePath                   = "addons/s3/env/cf.yml"
+	envS3AccessPolicyTemplatePath       = "addons/s3/env/access_policy.yml"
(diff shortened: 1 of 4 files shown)


iamhopaul123 (approved):

internal/pkg/addon/storage.go
```diff
@@ -45,43 +55,62 @@ var storageTemplateFunctions = map[string]interface{}{
...
-// S3Props contains S3-specific properties for addon.NewS3Template().
+// S3Props contains S3-specific properties.
 type S3Props struct {
 	*StorageProps
 }
 
-// NewS3Template creates a new S3 marshaler which can be used to write CF via addonWriter.
-func NewS3Template(input *S3Props) *S3Template {
+// NewWorkloadS3Template creates a marshaler for a workload-level S3 addon.
+func NewWorkloadS3Template(input *S3Props) *S3Template {
```
efekarakus: nit: what do you think of dropping the `New` prefix from all these functions so that it's just `addon.WorkloadS3Template`, `addon.EnvS3Template`

internal/pkg/addon/storage.go
```diff
@@ -259,18 +327,37 @@ func (r *RDSTemplate) MarshalBinary() ([]byte, error) {
 // NewRDSParams creates a new RDS parameters marshaler.
 func NewRDSParams() *RDSParams {
```
efekarakus: nit: maybe this one too can be `RDWSParamsForRDS`, `EnvParamsForRDS`, `RDWSParamsForEnvRDS`

efekarakus (approved):
just a small refactor suggestion to drop the `New` 

feel free to remove the label if you disagree or address the feedback

internal/pkg/addon/storage.go
Lou1415926: done!

PR #4418, opened 2023-01-25
test(e2e): Add secrets to App Runner test suite

yohanthshetty:


e2e/apprunner/copilot/back-end/manifest.yml
```diff
@@ -0,0 +1,61 @@
+# The manifest for the "back-end" service.
```
iamhopaul123: Why do we add `back-end` and `test` env manifest?

e2e/apprunner/copilot/front-end/manifest.yml
```diff
@@ -28,4 +28,9 @@ network:
 cpu: 1024
 # Amount of memory in MiB used by the task.
 memory: 2048
-# Number of tasks that should be running in your service.
\ No newline at end of file
+# Number of tasks that should be running in your service.
+
+secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store and Secrets Manager.
+  my-ssm-secret: "arn:aws:ssm:us-east-1:067305307211:parameter/test-ssm-secret"
```
iamhopaul123, line 34: this secret only exist in 067305307211 account which means the e2e test only works for that account. What I would expect from the e2e test is:

1. Create a secret using AWS CLI like https://github.com/aws/copilot-cli/blob/316aa07072b0b47baef5ce5efbaadd72710fcc21/e2e/sidecars/sidecars_suite_test.go#L47 in `BeforeSuite`
2. Use the secret in any of `back-end` or `front-end`
3. In `apprunner_test.go` do `svc show` and see if the secret is correctly  shown in the response https://github.com/aws/copilot-cli/blob/316aa07072b0b47baef5ce5efbaadd72710fcc21/e2e/sidecars/sidecars_test.go#L237
4. In `AfterSuite` remember to delete the secret.
5.

iamhopaul123: this secret only exist in 067305307211 account which means the e2e test only works for that account. What I would expect from the e2e test is:

1. Create a secret using AWS CLI like https://github.com/aws/copilot-cli/blob/316aa07072b0b47baef5ce5efbaadd72710fcc21/e2e/sidecars/sidecars_suite_test.go#L47 in `BeforeSuite`
2. Use the secret in any of `back-end` or `front-end`
3. In `apprunner_test.go` do `svc show` and see if the secret is correctly  shown in the response https://github.com/aws/copilot-cli/blob/316aa07072b0b47baef5ce5efbaadd72710fcc21/e2e/sidecars/sidecars_test.go#L237
4. In `AfterSuite` remember to delete the secret.

e2e/apprunner/apprunner_suite_test.go
```diff
@@ -36,10 +38,15 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 	Expect(err).NotTo(HaveOccurred())
 	appName = fmt.Sprintf("e2e-apprunner-%d", time.Now().Unix())
+	// err = command.Run("aws", []string{"secretsmanager", "create-secret", "--name", "MyTestSecret", "--secret-string", "{\"user\":\"yohanthr\",\"password\":\"EXAMPLE-PASSWORD\"}", "--tags", "Key=copilot-application,Value=", appName, "Key=copilot-environment,Value=test"})
```
yohanthshetty: Will remove this in the next commit. Uploaded by mistake.

e2e/apprunner/copilot/back-end/manifest.yml
yohanthshetty: Removed!

e2e/apprunner/copilot/front-end/manifest.yml
yohanthshetty: Updated with the changes.

e2e/apprunner/apprunner_suite_test.go
yohanthshetty: Removed

e2e/apprunner/apprunner_suite_test.go
```diff
@@ -20,7 +21,8 @@ const feSvcName = "front-end"
 const beSvcName = "back-end"
 const envName = "test"
 
-/**
+/*
```
iamhopaul123: Can we revert this?

e2e/apprunner/apprunner_suite_test.go
```diff
@@ -36,10 +38,14 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 	Expect(err).NotTo(HaveOccurred())
 	appName = fmt.Sprintf("e2e-apprunner-%d", time.Now().Unix())
+	err = command.Run("aws", []string{"ssm", "put-parameter", "--name", "my-ssm-param", "--value", "abcd1234", "--type", "String", "--tags", "[{\"Key\":\"copilot-application\",\"Value\":\"" + appName + "\"},{\"Key\":\"copilot-environment\", \"Value\":\"" + envName + "\"}]"})
```
iamhopaul123:
```suggestion
	err = command.Run("aws", []string{"ssm", "put-parameter", "--name", "e2e-apprunner-" + appName, "--value", "abcd1234", "--type", "String", "--tags", "[{\"Key\":\"copilot-application\",\"Value\":\"" + appName + "\"},{\"Key\":\"copilot-environment\", \"Value\":\"" + envName + "\"}]"})
```
to make sure we have different secrets created for each e2e test.

e2e/apprunner/copilot/back-end/manifest.yml
iamhopaul123: doesn't seem to be removed tho.

iamhopaul123 (commented):
Can we add test for secret manager as well? maybe add it to backend?

e2e/apprunner/apprunner_suite_test.go
Lou1415926: I guess it's probably fine keeping a deterministic name 🤔  it'd be difficult to inject this value into `front-end/manifest.yml` otherwise. What do you and @iamhopaul123 think about naming it `e2e-apprunner-ssm-param`?

e2e/apprunner/apprunner_test.go
```diff
@@ -165,6 +165,18 @@ var _ = Describe("App Runner", Ordered, func() {
 			}
 		})
 
+		It("should return correct secrets", func() {
+			fmt.Printf("\n\nsecrets: %+v\n\n", svc.Secrets)
+			expectedSecrets := map[string]string{
+				"COPILOT_ENVIRONMENT_NAME": envName,
+				"COPILOT_SERVICE_NAME":     feSvcName,
```
Lou1415926: I might have missed something from one of the previous PRs - why are `COPILOT_ENVIRONMENT_NAME` and `COPILOT_SERVICE_NAME`  also in the `Secrets` section of `copilot svc show`?

yohanthshetty: Will remove this. They are redundant fields.

e2e/apprunner/apprunner_suite_test.go
yohanthshetty: +1 to `e2e-apprunner-ssm-param`, yes it would be extra effort to inject that value into `front-end/manifest.yml`

iamhopaul123: OK let's keep it for now.

e2e/apprunner/copilot/back-end/manifest.yml
yohanthshetty: removed

yohanthshetty: Added `e2e-apprunner-MyTestSecret` (secrets manager secret) to backend service.

e2e/apprunner/apprunner_test.go
```diff
@@ -165,6 +165,26 @@ var _ = Describe("App Runner", Ordered, func() {
...
+			expectedSecrets := map[string]string{
+				"my-ssm-param": "e2e-apprunner-ssm-param",
+			}
+			for _, secret := range svc.Secrets {
+				Expect(secret.Value).To(Equal(expectedSecrets[secret.Name]))
+			}
+		})
+
+		It("should return correct secrets for backendSvc", func() {
+			fmt.Printf("\n\nsecrets: %+v\n\n", backendSvc.Secrets)
```
iamhopaul123:
```suggestion
```

e2e/apprunner/apprunner_test.go
```diff
@@ -165,6 +165,26 @@ var _ = Describe("App Runner", Ordered, func() {
...
+			expectedSecrets := map[string]string{
+				"USER_CREDS": "arn:aws:secretsmanager:us-east-1:067305307211:secret:e2e-apprunner-MyTestSecret",
```
iamhopaul123: We should just compare if the show result secret ARN contains `e2e-apprunner-MyTestSecret`, since the region and account could be different depending on your local environment.

e2e/apprunner/copilot/back-end/manifest.yml
```diff
@@ -0,0 +1,57 @@
...
+#
+# variables:                    # Pass environment variables as key value pairs.
+#   LOG_LEVEL: info
+#
+# tags:                         # Pass tags as key value pairs.
+#   project: project-name
+#
+secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store and Secrets Manager.
+  USER_CREDS: 
+    secretsmanager: 'e2e-apprunner-MyTestSecret'  # The key is the name of the environment variable, the value is the name of an SSM parameter.
```
Lou1415926: Is there any reason why you wanted to add this secret to the backend manifest, instead of frontend manifest?

If not, can we:
1. delete `back-end/manifest.yml`
2.  move this to secret `front-end/manifest.yml` so that it has:
```
secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store and Secrets Manager.
  my-ssm-param: e2e-apprunner-ssm-param
  USER_CREDS:
    secretsmanager: 'e2e-apprunner-MyTestSecret'
```


Unless necessary, I'd love us to maintain less manifests for e2e tests.

e2e/apprunner/apprunner_test.go
yohanthshetty: True, I thought so too! Will do that.

Lou1415926 (commented):
sorry for the iterations! +1 to what PH has with an additional tiny nit

e2e/apprunner/copilot/back-end/manifest.yml
yohanthshetty: @iamhopaul123 suggested I do that. I can add them both to just one manifest too.
Doest that work @iamhopaul123 ?

iamhopaul123: Yeah i agreed with @Lou1415926 's idea. Let's do it

e2e/apprunner/apprunner_test.go
yohanthshetty: Done

e2e/apprunner/copilot/back-end/manifest.yml
yohanthshetty: DOne

Lou1415926 (approved):

iamhopaul123 (approved):
:shipit:

PR #4423, opened 2023-01-26
chore(cli): add and validate a lifecycle flag for `storage init`

Lou1415926:
Note that the flag is optional defaulting to "workload" so we don't prompt for it.

### DIFF
--- a/internal/pkg/cli/storage_init.go
+++ b/internal/pkg/cli/storage_init.go
@@ -12,6 +12,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ssm"
 	"github.com/aws/copilot-cli/internal/pkg/aws/identity"
 	"github.com/aws/copilot-cli/internal/pkg/aws/sessions"
+	"github.com/dustin/go-humanize/english"
 
 	"github.com/aws/copilot-cli/internal/pkg/addon"
 	"github.com/aws/copilot-cli/internal/pkg/config"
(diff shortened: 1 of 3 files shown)


internal/pkg/cli/storage_init.go
```diff
@@ -882,6 +903,7 @@ Resource names are injected into your containers as environment variables for ea
 	cmd.Flags().StringVarP(&vars.storageName, nameFlag, nameFlagShort, "", storageFlagDescription)
 	cmd.Flags().StringVarP(&vars.storageType, storageTypeFlag, typeFlagShort, "", storageTypeFlagDescription)
 	cmd.Flags().StringVarP(&vars.workloadName, workloadFlag, workloadFlagShort, "", storageWorkloadFlagDescription)
+	cmd.Flags().StringVarP(&vars.lifecycle, storageLifecycleFlag, "", lifecycleWorkloadLevel, storageLifecycleFlagDescription)
```
efekarakus, line 906: gotcha, so we will default to "workload" and never prompt?

efekarakus (approved):

internal/pkg/cli/storage_init.go
huanjani: Is 'workload' is the default and we don't prompt bc of backwards compatibility concerns?

huanjani (approved):

internal/pkg/cli/storage_init.go
Lou1415926: yeah that was sort of the thinking! To be specific, I personally think adding new required flags isn't quite a breaking behavior, because the only ones it breaks for are the script users. However, in the mean time I wasn't sure if it was worth breaking script users, potentially upsetting some of them.

I am on the fence honestly. I am happy to convert!

efekarakus: > To be specific, I personally think adding new required flags isn't quite a breaking behavior

👍 

My approach from now is:
1. If we find a config file for the given name, like manifest or templates, we skip.
2. Otherwise, it's fine to prompt

So the way for users not to break their script is to drop a file where the resources will be created so that copilot skips the write

PR #4425, opened 2023-01-27
chore(deploy): inject Overrider to `EnvDeployer` 

efekarakus:

### DIFF
--- a/internal/pkg/cli/deploy/env.go
+++ b/internal/pkg/cli/deploy/env.go
@@ -13,8 +13,6 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/aws/copilot-cli/internal/pkg/override"
-
 	awscfn "github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/copilot-cli/internal/pkg/addon"
 	"github.com/aws/copilot-cli/internal/pkg/aws/cloudformation"
@@ -33,6 +31,7 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/describe"
 	"github.com/aws/copilot-cli/internal/pkg/describe/stack"
 	"github.com/aws/copilot-cli/internal/pkg/manifest"
+	"github.com/aws/copilot-cli/internal/pkg/override"
 	"github.com/aws/copilot-cli/internal/pkg/template"
 	"github.com/aws/copilot-cli/internal/pkg/template/artifactpath"
 	"github.com/aws/copilot-cli/internal/pkg/term/log"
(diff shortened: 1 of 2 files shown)


huanjani (approved):
lgtm

internal/pkg/cli/deploy/env.go
```diff
@@ -156,9 +163,11 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
 			TemplatePatcher: cfnClient,
 			Env:             in.Env,
 		},
-		newStackSerializer: func(in *cfnstack.EnvConfig, lastForceUpdateID string, oldParams []*awscfn.Parameter) stackSerializer {
-			return cfnstack.NewEnvConfigFromExistingStack(in, lastForceUpdateID, oldParams)
+		newStack: func(in *cfnstack.EnvConfig, lastForceUpdateID string, oldParams []*awscfn.Parameter) deploycfn.StackConfiguration {
+			conf := cfnstack.NewEnvConfigFromExistingStack(in, lastForceUpdateID, oldParams)
+			return deploycfn.WrapWithTemplateOverrider(conf, overrider)
 		},
+		overrider:    overrider,
```
paragbhingre: Just a quick question, do we need overrider separately here? Or we can just use it from the `newStack`?
Otherwise looks good to me.

efekarakus: 😍 thanks for pointing this out! I added this field earlier before the `newStack` refactor and forgot to remove 🤦 Fixed!

paragbhingre (approved):

internal/pkg/cli/deploy/env.go
```diff
@@ -140,6 +142,10 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
 	if err != nil {
 		return nil, fmt.Errorf("initialize env describer: %w", err)
 	}
+	overrider := in.Overrider
+	if overrider == nil {
+		overrider = new(override.Noop)
+	}
```
Lou1415926, line 148: sorry late to the party! Is there a reason why this is treated differently than workload deployer? https://github.com/aws/copilot-cli/blob/feb5084f180e8a0fe3dc173aa37be39944ca9234/internal/pkg/cli/deploy/workload.go#L240

efekarakus: Oh! I was supposed to remove the if-statement in the follow-up https://github.com/aws/copilot-cli/pull/4428 🤦 

I had it here so that the code didn't panic until I added the code that initialized the overrider, does that make sense?
I'll remove it in my next PR!

PR #4427, opened 2023-01-27
refactor(workspace): decouple write and addon path computation

Lou1415926:

### DIFF
--- a/internal/pkg/cli/storage_init.go
+++ b/internal/pkg/cli/storage_init.go
@@ -639,7 +639,7 @@ func (o *initStorageOpts) Execute() error {
 		return err
 	}
 	for _, addon := range addonBlobs {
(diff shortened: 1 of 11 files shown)


internal/pkg/cli/storage_init.go
```diff
@@ -686,6 +687,7 @@ func (o *initStorageOpts) addonBlobs() ([]addonBlob, error) {
 			name:        o.storageName,
```
efekarakus: nit: why not have the name be `fmt.Sprintf("%s.yml", o.storageName)`?

efekarakus (approved):
lgtm, just a tiny suggestion

huanjani (approved):
LGTM!

internal/pkg/cli/storage_init.go
Lou1415926: The intention was that perhaps we will support non-CFN addons in the future. Then the addon itself needs to decide what extension it's supposed to be. But yea, this is probably a premature generalization. Changed it to `".yml"`!

PR #4429, opened 2023-01-27
chore(override): write os/exec output to stderr

efekarakus:


huanjani (approved):
:shipit:

internal/pkg/override/cdk.go
```diff
@@ -20,7 +20,7 @@ import (
 type CDK struct {
 	rootAbsPath string // Absolute path to the overrides/ directory.
 
-	out  io.Writer // Writer for any os/exec calls.
+	w    io.Writer // Writer to pipe stdout and stderr content from os/exec calls.
```
Lou1415926: nit:  when reading ⬇️  i needed to look at the comment here in order to make sure `cdk.w` is used as intended.  What about renaming `w` to something like `execOut`
```
cmd.Stdout = cdk.w
cmd.Stderr = cdk.w
```

Lou1415926 (approved):

Issue #4254, opened 2022-12-06
Native image building and uploading for sidecar container


iamhopaul123:
As a Copilot user, I have a Rails/React app which is currently using Puma as the app server and Nginx as the web server. Nginx is currently installed in the app Dockerfile and launched in the Docker entrypoint. I'd like the run Nginx as a sidecar instead. However, since I need to also have the React and Active Admin assets compiled and included in the Nginx image, it means that I would be updating that image somewhat frequently, and it's somewhat of a pain to do that as Copilot doesn't handle updating sidecar images automatically.

carlosaherrera:
Similarly, I have a Django app that has two sidecars configured: one for a Celery worker and another for a Celery beat. It appears Copilot only allows you to configure a sidecar by pointing to an image URL. But I would like to have the option to point to a Dockerfile instead so that Copilot can build, upload, and deploy it for me automatically, like it currently does for the main Django app.

efekarakus:
Hi folks!!

I'm really excited to say that AWS Copilot v1.27 is out and you can now build and push sidecar container images! 🚀 

* Guide: https://aws.github.io/copilot-cli/docs/developing/sidecars/#image-build
* Blog post: https://aws.github.io/copilot-cli/blogs/release-v127/#build-and-push-container-images-for-sidecar-containers
* Release notes: https://github.com/aws/copilot-cli/releases/tag/v1.27.0

I'll close the issue now! If you have any questions, feedback, or suggestions please feel free to open new GitHub issues 🥳

PR #4430, opened 2023-01-28
chore(manifest): add sidecar image fields

KollaAdithya:
This PR adds the manifest fields for sidecar. 
Related to #4254 
Here is the proposed manifest Ux:
```yaml
sidecars:
  nginx:
    image: 'url'
```
  or
```yaml
sidecars:
  nginx:
    image:
      location: 'url'
```
or
```yaml
sidecars:
  nginx:
    image:
      build: 'path/to/Dockerfile'
```
or
```yaml
sidecars:
  nginx:
    image:
      build:
        dockerfile: path/to/Dockerfile
        context: context/dir
        target: build-stage
        cache_from:
          - image:tag
        args:
          key: value
```


internal/pkg/manifest/workload_ecs.go
```diff
@@ -299,19 +299,56 @@ func (lc *Logging) GetEnableMetadata() *string {
...
+
+// UnmarshalYAML overrides the default YAML unmarshaling logic for the Image
+// struct, allowing it to perform more complex unmarshaling behavior.
+// This method implements the yaml.Unmarshaler (v3) interface.
+func (s *SidecarImageConfig) UnmarshalYAML(value *yaml.Node) error {
+	type scImageConfig SidecarImageConfig
+	if err := value.Decode((*scImageConfig)(s)); err != nil {
+		return err
+	}
+	if !s.Build.IsZero() && s.Location != nil {
```
paragbhingre: What I think is this validation should be inside manifest/validation.go and not here in the `UnmarshalYAML`.

internal/pkg/manifest/workload_ecs_test.go
```diff
@@ -478,3 +509,39 @@ func TestLogging_GetEnableMetadata(t *testing.T) {
...
 	}
 }
+
+func Test_SidecarImageURI(t *testing.T) {
+	testCases := map[string]struct {
+		in     SidecarConfig
+		wanted *string
+	}{
+		"empty SidecarConfig": {},
+		"should return URI is provided directly through `image` ": {
```
paragbhingre:
```suggestion
		"should return URI if provided directly through `image` ": {
```

internal/pkg/manifest/validate.go
```diff
@@ -2042,3 +2045,8 @@ func contains(name string, names []string) bool {
 	}
 	return false
 }
+
+// validate is a no-op for SidecarImageConfig.
+func (s SidecarImageConfig) validate() error {
```
paragbhingre: This is the place where you should be testing what we are testing inside `UnmarshalYAML`

internal/pkg/manifest/validate.go
```diff
@@ -1326,6 +1326,9 @@ func (s SidecarConfig) validate() error {
 	if err := s.DependsOn.validate(); err != nil {
 		return fmt.Errorf(`validate "depends_on": %w`, err)
 	}
+	if err := s.Image.Advanced.validate(); err != nil {
+		return fmt.Errorf(`validate "sidecar_imageconfig: %w`, err)
```
paragbhingre: We always showcase the exact manifest field to the customer that requires fixing. 

```suggestion
		return fmt.Errorf(`validate "build: %w`, err)
```

internal/pkg/manifest/backend_svc_test.go
```diff
@@ -1098,8 +1108,10 @@ func TestBackendService_ExposedPorts(t *testing.T) {
 					},
 					Sidecars: map[string]*SidecarConfig{
 						"xray": {
-							//Port:       aws.String("2000"),
-							Image:      aws.String("123456789012.dkr.ecr.us-east-2.amazonaws.com/xray-daemon"),
+							//Port: aws.String("2000"),
+							Image: Union[*string, SidecarImageConfig]{
+								Basic: aws.String("123456789012.dkr.ecr.us-east-2.amazonaws.com/xray-daemon"),
```
paragbhingre: For one of the test we can also try putting this URI inside `image.location` instead of `image`

paragbhingre (commented):
I just have a few nits and a couple of requests. Overall, it looks good to me.

internal/pkg/manifest/backend_svc_test.go
efekarakus: Yeah that'd be awesome! Perhaps not in `_ExposedPorts` tests but within the `transform_test`?

internal/pkg/manifest/workload_ecs.go
```diff
@@ -299,19 +299,56 @@ func (lc *Logging) GetEnableMetadata() *string {
...
+			mustExist:   true,
+		}
+	}
+	return nil
+}
+
+// SidecarImageURI returns the URI of the image.
+func (s *SidecarConfig) SidecarImageURI() *string {
+	// Prefer to use the "Image" string in SidecarConfig. Otherwise,
+	// "Location" in SidecarImageConfig. If no image URI is specified, return nil.
```
efekarakus: This sort of info would be useful to the client! Therefore, it'd be neat to move it higher up into the method doc comments

internal/pkg/manifest/workload_ecs.go
```diff
@@ -299,19 +299,56 @@ func (lc *Logging) GetEnableMetadata() *string {
...
+			firstField:  "build",
+			secondField: "location",
+			mustExist:   true,
+		}
+	}
+	return nil
+}
+
+// SidecarImageURI returns the URI of the image.
+func (s *SidecarConfig) SidecarImageURI() *string {
```
efekarakus: What do you think of this method signature:
```
// ImageURI returns the location of the image if one is set. 
// If the image needs to be build, return "" and false.
func (cfg *SidecareConfig) ImageURI() (string, bool)
```
It's like looking up a value from a hashmap `v, ok := map[key]`

internal/pkg/manifest/svc_test.go
```diff
@@ -147,7 +147,7 @@ environments:
 						Sidecars: map[string]*SidecarConfig{
 							"xray": {
 								Port:       aws.String("2000/udp"),
-								Image:      aws.String("123456789012.dkr.ecr.us-east-2.amazonaws.com/xray-daemon"),
+								Image:      BasicToUnion[*string, SidecarImageConfig](aws.String("123456789012.dkr.ecr.us-east-2.amazonaws.com/xray-daemon")),
```
efekarakus, line 160: It'd be amazing if we can add a test case here where the `inContent` has the advanced configuration for the sidecar and we ensure that we're unmarshaling properly into `SidecarConfig`

efekarakus (commented):
Looks great!

internal/pkg/manifest/workload_ecs.go
```diff
@@ -299,19 +299,56 @@ func (lc *Logging) GetEnableMetadata() *string {
...
-	Essential     *bool                `yaml:"essential"`
-	CredsParam    *string              `yaml:"credentialsParameter"`
-	Variables     map[string]Variable  `yaml:"variables"`
-	Secrets       map[string]Secret    `yaml:"secrets"`
-	MountPoints   []SidecarMountPoint  `yaml:"mount_points"`
-	DockerLabels  map[string]string    `yaml:"labels"`
-	DependsOn     DependsOn            `yaml:"depends_on"`
-	HealthCheck   ContainerHealthCheck `yaml:"healthcheck"`
+	Port          *string                            `yaml:"port"`
+	Image         Union[*string, SidecarImageConfig] `yaml:"image"`
```
huanjani, line 303: nice!

internal/pkg/manifest/backend_svc_test.go
KollaAdithya: Do we still need custom Transformer for mergo in `transform.go` and `transform_test` 🤔

internal/pkg/manifest/backend_svc_test.go
```diff
@@ -1098,8 +1108,12 @@ func TestBackendService_ExposedPorts(t *testing.T) {
 					},
 					Sidecars: map[string]*SidecarConfig{
 						"xray": {
-							//Port:       aws.String("2000"),
-							Image:      aws.String("123456789012.dkr.ecr.us-east-2.amazonaws.com/xray-daemon"),
+							//Port: aws.String("2000"),
```
huanjani: Should this still be commented out? 🤔

huanjani (approved):
Just one tiny question; otherwise, LGTM!

internal/pkg/manifest/backend_svc_test.go
KollaAdithya: removed the commented line

internal/pkg/manifest/workload_ecs.go
```diff
@@ -299,19 +299,37 @@ func (lc *Logging) GetEnableMetadata() *string {
...
 
+// SidecarImageConfig represents the docker build arguments and location of the existing image.
+type SidecarImageConfig struct {
+	Build    Union[*string, DockerBuildArgs] // `yaml:"build"`
+	Location *string                         // `yaml:"location"`
+}
+
+// ImageURI returns the location of the image if one is set.
+// If the image needs to be build, return "" and false.
+func (cfg *SidecarConfig) ImageURI() (string, bool) {
```
iamhopaul123, line 323: Are we using the `bool` from output anywhere 🤔

KollaAdithya: No we are not using `bool` anywhere. As per Efe's [recommendation ](https://github.com/aws/copilot-cli/pull/4430#discussion_r1091369434) I changed to this method signature. 
I will change it back to 
`func (cfg *SidecarConfig) ImageURI() *string` if you insist.

iamhopaul123 (approved):

internal/pkg/manifest/workload_ecs.go
```diff
@@ -299,19 +299,37 @@ func (lc *Logging) GetEnableMetadata() *string {
...
+	Secrets       map[string]Secret                  `yaml:"secrets"`
+	MountPoints   []SidecarMountPoint                `yaml:"mount_points"`
+	DockerLabels  map[string]string                  `yaml:"labels"`
+	DependsOn     DependsOn                          `yaml:"depends_on"`
+	HealthCheck   ContainerHealthCheck               `yaml:"healthcheck"`
 	ImageOverride `yaml:",inline"`
 }
 
+// SidecarImageConfig represents the docker build arguments and location of the existing image.
+type SidecarImageConfig struct {
```
Lou1415926, line 316: sorry I am late to the party 😂 

This is not a request to change the implementation but just an idea. What do you think of doing this:
```go
type ImageLocationOrBuild struct {
	Build    Union[*string, DockerBuildArgs] `yaml:"build"`
	Location *string                                         `yaml:"location"`
}
type Image struct {
      ImageLocationOrBuild `yaml:",inline"`
}

type SidecarConfig struct {
     Image         Union[*string, ImageLocationOrBuild] `yaml:"image"`
}
```

This struct here is named `SidecarImageConfig` but there is actually nothing specific about `Sidecar` in it. It looks like a general purpose struct that can be used for main container as well.

PR #4437, opened 2023-01-30
chore:  execute `storage init` with environment-level storages

Lou1415926:


internal/pkg/cli/storage_init.go
```diff
@@ -692,65 +691,63 @@ func (o *initStorageOpts) readWorkloadType() error {
...
-	name        string
 	description string
 	blob        encoding.BinaryMarshaler
+	path        string
 }
 
 func (o *initStorageOpts) addonBlobs() ([]addonBlob, error) {
-	templateBlob, err := o.newAddonTemplate()
+	switch {
+	case o.lifecycle == lifecycleWorkloadLevel && o.storageType == s3StorageType:
```
efekarakus: nit: I was experimenting with this model of writing so that there is no conditional, what do you think? 
https://go.dev/play/p/hMJ3YcffM0x

It's really a nit

internal/pkg/cli/storage_init.go
```diff
@@ -692,65 +691,63 @@ func (o *initStorageOpts) readWorkloadType() error {
...
+	case o.lifecycle == lifecycleWorkloadLevel && o.storageType == rdsStorageType:
+		return o.wkldRDSAddonBlobs()
+	case o.lifecycle == lifecycleEnvironmentLevel && o.storageType == s3StorageType:
+		return o.envS3AddonBlobs()
+	case o.lifecycle == lifecycleEnvironmentLevel && o.storageType == dynamoDBStorageType:
+		return o.envDDBAddonBlobs()
+	case o.lifecycle == lifecycleEnvironmentLevel && o.storageType == rdsStorageType:
+		return o.envRDSAddonBlobs()
+	}
+	return nil, fmt.Errorf("storage type %s doesn't have a CF template", o.storageType)
```
efekarakus:
```suggestion
	return nil, fmt.Errorf("storage type %s doesn't have a CloudFormation template", o.storageType)
```
To avoid acronyms that users might not be familiar with

internal/pkg/cli/storage_init.go
```diff
@@ -692,65 +691,63 @@ func (o *initStorageOpts) readWorkloadType() error {
...
-	}
-	return addon.RDWSParamsForRDS(), nil
+	return []addonBlob{
+		{
+			path:        o.ws.EnvAddonFilePath(fmt.Sprintf("%s.yml", o.storageName)),
+			description: "template",
+			blob:        addon.EnvDDBTemplate(props),
+		},
+		{
+			path:        o.ws.WorkloadAddonFilePath(o.workloadName, fmt.Sprintf("%s-access-policy.yml", o.storageName)),
```
efekarakus: Wouldn't it be possible to define an env addon for DDB in a Git repository that contains only environment manifests?

For example, what if my git repository is just this:
```
copilot/
  environments/
      test/manifest.yml
      prod/manifest.yml
```
Should we still let users create this template and instead provide this addon blob as a recommended action?

internal/pkg/cli/storage_init.go
```diff
@@ -93,7 +93,7 @@ Aurora Serverless is an on-demand autoscaling configuration for Amazon Aurora, a
 	fmtStorageInitNamePrompt = "What would you like to " + color.Emphasize("name") + " this %s?"
 	storageInitNameHelp      = "The name of this storage resource. You can use the following characters: a-zA-Z0-9-_"
 
-	storageInitSvcPrompt = "Which " + color.Emphasize("workload") + " would you like to associate with this storage resource?"
+	storageInitSvcPrompt = "Which " + color.Emphasize("workload") + " need access to the storage?"
```
huanjani:
```suggestion
	storageInitSvcPrompt = "Which " + color.Emphasize("workload") + " needs access to the storage?"
```

internal/pkg/cli/storage_init.go
Lou1415926: The `switch selection` kind of writing seems to end up being as efficient as the other, if not better 🤔 It is more readable as well. I will update the code.

Lou1415926: done!

internal/pkg/cli/storage_init.go
Lou1415926: reiterating the offline discussion here for visibility - in the ⬆️  case, we will:
1. not write the addon template to `copilot/<svc>/addons/ingress.yml
2. will provide a recommended actions to add the ingress

This logic is updated in the current iteration!

internal/pkg/cli/storage_init.go
```diff
@@ -654,13 +652,15 @@ func (o *initStorageOpts) Execute() error {
...
 		return err
 	}
-
 	addonBlobs, err := o.addonBlobs()
 	if err != nil {
 		return err
 	}
 	for _, addon := range addonBlobs {
-		path, err := o.ws.Write(addon.blob, o.ws.WorkloadAddonFilePath(o.workloadName, fmt.Sprintf("%s.yml", addon.name)))
+		if isWorkloadLevel := !addon.isEnvLevel; isWorkloadLevel && !o.workloadExists {
```
efekarakus: I wonder if this conditional should be a method on `addonBlob`?
```go
func (blob addonBlob) isWritable() bool
```

internal/pkg/cli/storage_init.go
```diff
@@ -692,65 +692,70 @@ func (o *initStorageOpts) readWorkloadType() error {
...
-	}
-	return addon.RDWSParamsForRDS(), nil
+	return []addonBlob{
+		{
+			path:        o.ws.EnvAddonFilePath(fmt.Sprintf("%s.yml", o.storageName)),
+			description: "template",
+			blob:        addon.EnvDDBTemplate(props),
+			isEnvLevel:  true,
+		},
+		{
```
efekarakus: Since in the recommended action we will suggest for them to run `storage init` with a different flag in their workspace,

I wonder if we can simplify and remove the `isWritable` method if we just don't append the workload blob if `!workloadExists`

Lou1415926: done!

efekarakus (approved):

iamhopaul123 (approved):

PR #4438, opened 2023-01-30
refactor: Network Load Balancer Targets and Health Checks

paragbhingre:


internal/pkg/manifest/svc.go
```diff
@@ -570,3 +572,46 @@ func sortExposedPorts(exposedPorts []ExposedPort) []ExposedPort {
 	})
 	return exposedPorts
 }
+
+// ParseNLBPortMapping returns parsed NLB port and protocol.
+func ParseNLBPortMapping(nlbPort *string) (port *string, protocol *string, err error) {
```
efekarakus: Can we add unit tests to the new public  functions/methods?

internal/pkg/manifest/svc.go
```diff
@@ -570,3 +572,46 @@ func sortExposedPorts(exposedPorts []ExposedPort) []ExposedPort {
...
+		return nil, nil, err
+	}
+	if protocol == nil {
+		protocol = aws.String(defaultNLBProtocol)
+	}
+	return
+}
+
+// NetworkLoadBalancerTarget returns target container and target port for the NLB configuration.
+func (s *LoadBalancedWebService) NetworkLoadBalancerTarget() (targetContainer *string, targetPort *string, err error) {
```
efekarakus: Is there a scenario here where we return `nil, nil, nil` if there is no `nlb` setting set in the manifest?

Otherwise, it looks like we are always setting a value for `targetContainer` and `targetPort` so we could return `string` instead of `*string`.

internal/pkg/manifest/svc.go
efekarakus: Given that the [default when `nil` is TCP](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_PortMapping.html), I wonder if we even need to have a `ParseNLBPortMapping` function or if we can re-use `ParsePortMapping`?

internal/pkg/manifest/svc.go
paragbhingre: We do not send `nil, nil, nil` but definitely send `nil, nil, err` in case we get any parsing error in nlb port. 
"no nlb setting set in the manifest" case is already being handled even before coming to `NetworkLoadBalancerTarget()` method. We exit in case of no nlb from [convertNetworkLoadBalancer](https://github.com/aws/copilot-cli/pull/4438/files#diff-8852a201e2b451d8c315d78bf675f068153e17c99506a4dc6ecc692a25cb33cbL438) itself.

internal/pkg/manifest/svc.go
paragbhingre: I'd agree to what you are saying, but to avoid the duplication of that extra step of checking for protocol is nil and if yes then set as TCP I thought of creating this new method. I can very well use the existing  `ParsePortMapping` and write the if protocol == nil then set it as TCP logic wherever it is required. Does that sound good to you?

internal/pkg/manifest/svc.go
efekarakus: >"no nlb setting set in the manifest" case is already being handled even before coming to NetworkLoadBalancerTarget() method. We exit in case of no nlb from convertNetworkLoadBalancer itself.

But this means that there is a _temporal_-coupling between the `nlConfig.IsEmpty` method and the `NetworkLoadBalancerTarget` method. In other words, a client needs to know ahead of time that (1) they should be calling `!IsEmpty` first and then (2) this method. Otherwise, they will get incorrect return values `(&s.Name, nil, nil)`.

There is nothing in the method signature today that indicates this method depends on `IsEmpty`.

My recommendation is to either:

1. Keep `*string, *string, error` as return only if we do the following in the method body:
   ```go
   svc.NLBConfig.IsEmpty() {
      return nil, nil, nil
   }
   ```
   This way you can remove the check from the `convertNLBHealthCheck` function as well.

2. Or we (1) update the doc comments saying this method should only be called when the NLB config is not empty and (2) update the return to `(string, string, error)` to indicate that if there is no error with parsing there will always be a valid returned target container and port

paragbhingre: I understand what you're saying here. I think I will choose option 2.1 because
1. Option 1 says, we can remove  `svc.NLBConfig.IsEmpty()` check from `convertNetworkLoadBalancer()` by placing that check inside `NetworkLoadBalancerTarget()`. But that will break the code as `convertNetworkLoadBalancer()` shouldn't be executed if there is empty nlb config which we are now checking inside `NetworkLoadBalancerTarget()` and returning `nil, nil, nil`. 
2. Even if we implement option 1 and get returned value as `nil, nil, nil` in `convertNetworkLoadBalancer()` then also using that condition to exit `convertNetworkLoadBalancer()` method creates `temporal-coupling` between `convertNetworkLoadBalancer()` and `NetworkLoadBalancerTarget()`

efekarakus (approved):

iamhopaul123 (approved):

PR #4441, opened 2023-01-31
chore: add asset pkg and filter for upload

iamhopaul123:


internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -0,0 +1,117 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package customresource provides functionality to upload Copilot custom resources.
```
efekarakus: wrong package comment

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -0,0 +1,117 @@
...
+type UploadFunc func(dest string, contents io.Reader) (url string, err error)
+
+// UploadInput is the input of Upload.
+type UploadInput struct {
+	Source      string
+	Destination string
+	Includes    []string
+	Excludes    []string
+	Upload      UploadFunc
+	Reader      *afero.Afero
```
efekarakus: Can we depend on an interface instead of a concrete implementation (e.g. `afero.Fs`) and rename the field to `FS`

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// Upload uploads static assets to Cloud Storage.
+func Upload(in *UploadInput) ([]string, error) {
```
efekarakus: one pattern that I find kinda neat is separating _required_ parameters from _optional_ ones. For example, here it would be:
```go
type UploadOpts struct {
   Includes []string // Relative path under source to include in the upload.
   Excludes []string // Relative path under source to include in the upload.
   UploadFn UploadFunc // Custom implementation on how to upload the contents under a file. Defaults to S3UploadFn.
}

func Upload(fs afero.Fs, source, destination string, opts UploadOpts)
```

What do you think boss? We do something similar in `override/cdk.go`

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -0,0 +1,117 @@
...
+// UploadFunc is the function signature to upload contents to a destination.
+type UploadFunc func(dest string, contents io.Reader) (url string, err error)
+
+// UploadInput is the input of Upload.
+type UploadInput struct {
+	Source      string
+	Destination string
+	Includes    []string
+	Excludes    []string
+	Upload      UploadFunc
```
efekarakus: nit: Can we include doc comments for these fields 🥺 I assume that source is an absolute path but destination is a location in S3?

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// Upload uploads static assets to Cloud Storage.
+func Upload(in *UploadInput) ([]string, error) {
+	files, err := listFiles(in.Reader, in.Source)
+	if err != nil {
+		return nil, err
+	}
+	filter := filter{buildPatterns(in.Includes, in.Excludes)}
+	filteredFiles, err := filter.apply(files)
```
efekarakus: lol this whole package is so similar to my interview question for SDE2s 😂 

What do you think of a layout like this overall:
```go
type filepathMatcher interface {
   match(path string) (bool, error)
}

type includeMatcher string
func (f includeMatcher) match(path string) (bool, error)

type excludeMatcher string
func (f excludeMatcher) match(path string) (bool, error)

type andMatcher []filepathMatcher
func (f andMatcher) match(path string) (bool, error)
```

and in `Upload` we get something like:
```go
includeMatchers := buildincludeMatchers(opts.Include...)
excludeMatchers := buildExcludeMatchers(opts.Exclude...)
matcher := buildAndMatcher(append(includeMatchers, excludeMatchers...))

var fpaths []string
err := afero.Walk(fs, source, walkFnWithMatcher(fpaths, matcher))
```

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -0,0 +1,117 @@
...
+	for _, syntax := range includes {
+		filterPatterns = append(filterPatterns, pattern{
+			action: include,
+			syntax: syntax,
+		})
+	}
+	return filterPatterns
+}
+
+func listFiles(fs *afero.Afero, path string) ([]string, error) {
```
efekarakus: can we leverage https://pkg.go.dev/github.com/spf13/afero#Walk here? 

I'm thinking that we can create a fn that returns a https://pkg.go.dev/path/filepath#WalkFunc:
```go
func walkFnWithMatcher(fpaths []string, matcher filepathMatcher) filepath.WalkFunc {
   return func(path string, info fs.FileInfo, err error) error {
      if info.IsDir() {  return filepath.SkipDir }
      ok, err := matcher.match(path)
      if err != nil { return err }
      if ok { fpaths = append(fpaths, path) }
      return nil
   }
}
```

Something like this, my only concern is that I'm not sure if the append will work

internal/pkg/deploy/upload/asset/asset_test.go
```diff
@@ -0,0 +1,83 @@
...
+			expectedFiles: []string{"test/copilot/.workspace"},
+		},
+		"success with include only": {
+			inSource:   "./test",
+			inIncludes: []string{"copilot/prod/manifest.yaml"},
+			mockFileSystem: func(fs afero.Fs) {
+				afero.WriteFile(fs, "test/copilot/.workspace", []byte(mockContent), 0644)
+				afero.WriteFile(fs, "copilot/prod/manifest.yaml", []byte(mockContent), 0644)
+			},
+			expectedFiles: []string{"test/copilot/.workspace"},
```
Lou1415926, line 39: +1 to having comment doc for `UploadInput` fields especially `Includes` and `Excludes`. I thought the `expectedFiles` here should be `[]string{"copilot/prod/manifest.yaml"}` because I didn't understand what `Includes` was.

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -0,0 +1,117 @@
...
+func buildExcludeMatchers(excludes []string) []filepathMatcher {
+	var matchers []filepathMatcher
+	for _, exclude := range excludes {
+		matchers = append(matchers, excludeMatcher(exclude))
+	}
+	return matchers
+}
+
+func (m excludeMatcher) match(path string) (bool, error) {
+	return match(string(m), path)
```
efekarakus, line 85: should this be 👇? otherwise includeMatcher and excludeMatcher look the same to me
```suggestion
	return !match(string(m), path)
```
It feels counter intuitive but we want `match` to be `false` instead of `true` to indicate the file shouldn't be added if the exclusion pattern matches 😅

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -0,0 +1,117 @@
...
+
+// compositeMatcher is a composite matcher consisting of include matchers and exclude matchers.
+// Note that exclude matchers will be applied before include matchers.
+type compositeMatcher []filepathMatcher
+
+func buildCompositeMatchers(includeMatchers, excludeMatchers []filepathMatcher) compositeMatcher {
+	return append(excludeMatchers, includeMatchers...)
+}
+
+func (m compositeMatcher) match(path string) (bool, error) {
```
efekarakus: a slightly different way of writing perhaps like this? This way we don't depend on `matcher.(includeMatcher)` 

_I think_ this should pass the unit tests.

```go
type sequentialMatcher struct {
  matchers []filepathMatcher
}

func (m *sequentialMatcher) match(path string) (bool, error) {
   ok := true
   for _, matcher := range m.matchers {
       isMatch, err := matcher.match(path)
       if err != nil {
           return false, err
       }
       ok = isMatch
   }
   return ok, nil
}

/**
Example usage: buildSequentialMatcher(
    buildAndMatcher(excludeMatchers), // AND because if any exclude is "false", then the file should be excluded.
    buildOrMatcher(includeMatchers),    //  OR because if any include is "true", then the file should be included.
 )
 */
func buildSequentialMatcher(matchers ...filepathMatcher) filepathMatcher

```

iamhopaul123: I don't think it will pass this
```golang
		"success with include only": {
			inSource:   "./test",
			inIncludes: []string{"copilot/prod/manifest.yaml"},
			mockFileSystem: func(fs afero.Fs) {
				afero.WriteFile(fs, "test/copilot/.workspace", []byte(mockContent), 0644)
				afero.WriteFile(fs, "copilot/prod/manifest.yaml", []byte(mockContent), 0644)
			},
			expectedFiles: []string{"test/copilot/.workspace"},
		},
```
It'll return an empty list `[]{}` instead.

efekarakus (approved):
ty boss! looks great

Lou1415926 (approved):

PR #4452, opened 2023-02-02
test: add tests for GenerateCloudFormationTemplate

efekarakus:

### DIFF
--- a/internal/pkg/cli/deploy/backend_test.go
+++ b/internal/pkg/cli/deploy/backend_test.go
@@ -6,6 +6,10 @@ package deploy
 import (
 	"errors"
 	"testing"
+	"time"
+
+	"github.com/aws/copilot-cli/internal/pkg/override"
+	"gopkg.in/yaml.v3"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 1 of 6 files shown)


internal/pkg/cli/deploy/rdws_test.go
```diff
@@ -18,6 +22,36 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestRdwsDeployer_GenerateCloudFormationTemplate(t *testing.T) {
+	t.Run("ensure resulting CloudFormation template custom resource paths are empty", func(t *testing.T) {
+		// GIVEN
+		backend := mockBackendServiceDeployer()
```
iamhopaul123: 🤔

iamhopaul123 (commented):
lgtm overall just one nit 🚢

internal/pkg/cli/deploy/rdws_test.go
efekarakus: lol i noticed that too, just fixed it!

internal/pkg/cli/deploy/backend_test.go
```diff
@@ -16,6 +20,40 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestBackendSvcDeployer_GenerateCloudFormationTemplate(t *testing.T) {
+	t.Run("ensure resulting CloudFormation template custom resource paths are empty", func(t *testing.T) {
```
Lou1415926, line 24: qq: what about paths like addons path? do we want to keep  them empty as well ?

Lou1415926 (approved):

internal/pkg/cli/deploy/backend_test.go
efekarakus: ooh good thinking, let me test that with overrides and see if CDK complains. If it does i'll fix that too

iamhopaul123 (approved):

PR #4453, opened 2023-02-03
chore: populate lambda fn Code values without having to upload

efekarakus:

### DIFF
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -491,6 +491,26 @@ func (d *workloadDeployer) pushAddonsTemplateToS3Bucket() (string, error) {
 }
 
 func (d *workloadDeployer) runtimeConfig(in *StackRuntimeConfiguration) (*stack.RuntimeConfig, error) {
+	out, err := d.convertRuntimeConfig(in)
+	if err != nil {
+		return nil, err
+	}
+	// Assign defaults if not set.
+	if len(out.CustomResourcesURL) == 0 {
+		// Populate the custom resource paths that copilot should upload the resources to.
+		crs, err := d.customResources(d.templateFS)
+		if err != nil {
+			return nil, err
+		}
+		out.CustomResourcesURL = make(map[string]string)
+		for _, cr := range crs {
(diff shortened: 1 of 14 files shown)


internal/pkg/cli/deploy/workload.go
```diff
@@ -486,6 +491,26 @@ func (d *workloadDeployer) pushAddonsTemplateToS3Bucket() (string, error) {
...
+		return nil, err
+	}
+	// Assign defaults if not set.
+	if len(out.CustomResourcesURL) == 0 {
+		// Populate the custom resource paths that copilot should upload the resources to.
+		crs, err := d.customResources(d.templateFS)
+		if err != nil {
+			return nil, err
+		}
+		out.CustomResourcesURL = make(map[string]string)
```
iamhopaul123, line 505: It feels odd to me to check if `len(out.CustomResourcesURL) == 0` and if so initialize `out.CustomResourcesURL` again. Isn't it empty already?

efekarakus: ooh yea, the problem is that if `out.CustomResourcesURL` is nil then we can't write `out.CustomResourcesURL[cr.Name()] = ...`

Like I have to initialize the map if it's nil if that makes sense.

iamhopaul123 (approved):

internal/pkg/cli/deploy/workload.go
```diff
@@ -486,6 +491,26 @@ func (d *workloadDeployer) pushAddonsTemplateToS3Bucket() (string, error) {
...
+		for _, cr := range crs {
+			out.CustomResourcesURL[cr.Name()] = s3.URL(d.resources.Region, d.resources.S3Bucket, cr.ArtifactPath())
```
Lou1415926, line 507: What are your thoughts on handling all these in logic in `Template()` of `stack` package?

efekarakus: That's a great idea, I like that!  Now I need to somehow find the time to do the refactor that moves the `customresources` from the `deploy` pkg to stack  :rip:

efekarakus:
I'll close this PR in favor of Wanxian's recommendation which is a better approach: https://github.com/aws/copilot-cli/pull/4453/files#r1096275545

PR #4455, opened 2023-02-03
chore(prompt): prompt with friendly text in place of raw values

Lou1415926:

### DIFF
--- a/internal/pkg/term/prompt/select.go
+++ b/internal/pkg/term/prompt/select.go
@@ -40,16 +40,21 @@ var regexpSGR = regexp.MustCompile(sgr)
 
 // Option represents a choice with a hint for clarification.
 type Option struct {
-	Value string
-	Hint  string
+	Value        string // The actual value represented by the option.
+	FriendlyText string // An optional FriendlyText displayed in place of Value.
+	Hint         string // An optional Hint displayed alongside the Value or FriendlyText.
 }
 
 // String implements the fmt.Stringer interface.
 func (o Option) String() string {
-	if o.Hint == "" {
-		return fmt.Sprintf("%s\t", o.Value)
+	text := o.Value
+	if o.FriendlyText != "" {
+		text = o.FriendlyText
 	}
-	return fmt.Sprintf("%s\t%s", o.Value, color.Faint.Sprintf("(%s)", o.Hint))
+	if o.Hint == "" {
+		return fmt.Sprintf("%s\t", text)
+	}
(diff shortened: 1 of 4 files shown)


internal/pkg/term/prompt/select.go
```diff
@@ -58,15 +63,15 @@ func (p Prompt) SelectOption(message, help string, opts []Option, promptCfgs ...
...
+	choices, choice2Value, err := stringifyOptions(opts)
 	if err != nil {
 		return "", err
 	}
 	result, err := p.SelectOne(message, help, choices, promptCfgs...)
 	if err != nil {
 		return "", err
 	}
-	return parseValueFromOptionFmt(result), nil
+	return choice2Value[result], nil
```
efekarakus: nit: I think this is a preference thing, so feel free to ignore, but I don't find it common to return more than 2 return values in Go. The only way that I've seen it is for things like databases:
```go
sql, close, err := sql.Open("db://")
if err != nil { ... }
defer close()
```
Where you want to explicitly highlight to the client "hey, don't forget to call close"

----

What do you think of these alternatives:
1. Keep `stringifyOptions` as is but move the map creation to a custom type
```go
type options []Option
func (opts options) friendlyTextIndex() map[string]string
```
2. Or, instead build a composite type and return two values.
```go
type prettyOptions struct { ... }
func (opts prettyOptions) Choices() []string
func (opts prettyOptions) ValueFrom(friendlyText string) string

func prettifyOptions(opts []Option) (prettyOptions, error)
```

efekarakus (approved):
Approved with a suggestion - since the function is private, the feedback isn't blocking

internal/pkg/term/prompt/select.go
Lou1415926: Makes sense to me!!

I ended up doing a modified version of alternative 2.
```go
type prettyOptions struct { 
  choices []string
  choice2Value map[string]string
} 
func prettifyOptions(opts []Option) (prettyOptions, error)

func (p Prompt) SelectOption(message, help string, opts []Option, promptCfgs ...PromptConfig) (value string, err error) {
        prettified, err := prettifyOptions(opts)
        choices := prettified.choices
        ....
	return prettified.choice2Value[result], nil
}
```

That is, instead of having `Choices() []string` and `ValueFrom(friendlyText string) string`, the caller directly accesses `prettyOptions`'s member fields. Because I thought `prettyOptions` do not need to be abstracted from `SelectOptions` since they are in the same package.

iamhopaul123 (approved):

PR #4462, opened 2023-02-06
chore(term): enable selecting a workload from config store

Lou1415926:

### DIFF
--- a/internal/pkg/term/selector/errors.go
+++ b/internal/pkg/term/selector/errors.go
@@ -19,49 +19,49 @@ var (
 	ErrSubnetsNotFound = errors.New("no existing subnets found")
 )
 
-// ErrNoWorkloadInApp occurs when there is no workload in an application.
-type ErrNoWorkloadInApp struct {
+// errNoWorkloadInApp occurs when there is no workload in an application.
+type errNoWorkloadInApp struct {
 	appName string
 }
 
-func (e *ErrNoWorkloadInApp) Error() string {
+func (e *errNoWorkloadInApp) Error() string {
 	return fmt.Sprintf("no workloads found in app %s", e.appName)
 }
 
 // RecommendActions gives suggestions to fix the error.
-func (e *ErrNoWorkloadInApp) RecommendActions() string {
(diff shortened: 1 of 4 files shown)


internal/pkg/term/selector/selector.go
```diff
@@ -960,6 +954,31 @@ func (s *ConfigSelector) Job(msg, help, app string) (string, error) {
 	return selectedJobName, nil
 }
 
+func (s *ConfigSelector) Workload(msg, help, app string) (string, error) {
```
paragbhingre, line 958: Can we put comment to this public function? 
And where are we calling this function from?

internal/pkg/term/selector/errors.go
```diff
@@ -13,3 +18,51 @@ var (
 	// ErrSubnetsNotFound is returned when no existing subnets are found.
 	ErrSubnetsNotFound = errors.New("no existing subnets found")
 )
+
+// ErrNoWorkloadInApp occurs when there is no workload in an application.
+type ErrNoWorkloadInApp struct {
```
efekarakus: can we add unit tests for these errors 🥺

Lou1415926: done!

internal/pkg/term/selector/selector.go
Lou1415926: > Can we put comment to this public function?

done!

> And where are we calling this function from?

we will need it in `storage init`

internal/pkg/term/selector/errors_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+package selector
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestErrNoWorkloadInApp_RecommendActions(t *testing.T) {
```
efekarakus, line 19: oh we can have testcases for `_Error` as well

efekarakus (approved):

iamhopaul123 (approved):

PR #4463, opened 2023-02-06
chore: inject S3 bucket name down to stack from deploy pkg

efekarakus:
The first two commits are the same as #4453, I've only added https://github.com/aws/copilot-cli/commit/a665b6240e132c47d4bc44b716fa72cb29474ac9 to start injecting the artifact bucket name down to stack to address Wanxian's comment.

### DIFF
--- a/internal/pkg/cli/deploy/backend.go
+++ b/internal/pkg/cli/deploy/backend.go
@@ -21,12 +21,12 @@ import (
 type backendSvcDeployer struct {
 	*svcDeployer
 	backendMft         *manifest.BackendService
-	customResources    customResourcesFunc
 	aliasCertValidator aliasCertValidator
 }
 
 // NewBackendDeployer is the constructor for backendSvcDeployer.
 func NewBackendDeployer(in *WorkloadDeployerInput) (*backendSvcDeployer, error) {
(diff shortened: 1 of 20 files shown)


iamhopaul123 (approved):

internal/pkg/cli/deploy/backend.go
```diff
@@ -91,12 +91,13 @@ func (d *backendSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (
...
-		EnvManifest:   d.envConfig,
-		Manifest:      d.backendMft,
-		RawManifest:   d.rawMft,
-		RuntimeConfig: *rc,
-		Addons:        d.addons,
+		App:                d.app,
+		EnvManifest:        d.envConfig,
+		Manifest:           d.backendMft,
+		RawManifest:        d.rawMft,
+		ArtifactBucketName: d.resources.S3Bucket,
```
paragbhingre, line 98: Is bucketName itself is enough to pass down to cfn or do we need S3Key as well?

efekarakus: S3 key will computed in the follow-up PR :)

paragbhingre (approved):

PR #4465, opened 2023-02-07
feat: prompt for storage "lifecycle" in `storage init` 

Lou1415926:
Screen Shot 2023-02-07 at 11 53 32 AM

Screen Shot 2023-02-07 at 11 53 44 AM

### DIFF
--- a/internal/pkg/cli/storage_init.go
+++ b/internal/pkg/cli/storage_init.go
@@ -57,8 +57,8 @@ const (
 	lifecycleEnvironmentLevel = "environment"
 	lifecycleWorkloadLevel    = "workload"
 
-	lifecycleEnvironmentFriendlyText = "No, I want the storage to be created and deleted at the environment level"
-	fmtLifecycleWorkloadFriendlyText = "Yes, the table should be create and deleted at the same time as %s"
+	lifecycleEnvironmentFriendlyText = "No, the storage should be created and deleted at the environment level"
+	fmtLifecycleWorkloadFriendlyText = "Yes, the storage should be created and deleted at the same time as %s"
 )
 
 var validLifecycleOptions = []string{lifecycleWorkloadLevel, lifecycleEnvironmentLevel}
@@ -454,7 +454,7 @@ func (o *initStorageOpts) validateOrAskLifecycle() error {
 	options := []prompt.Option{
 		{
 			Value:        lifecycleWorkloadLevel,
-			FriendlyText: fmt.Sprintf(fmtLifecycleWorkloadFriendlyText, o.workloadName),
+			FriendlyText: fmt.Sprintf(fmtLifecycleWorkloadFriendlyText, color.HighlightUserInput(o.workloadName)),
 		},
 		{
 			Value:        lifecycleEnvironmentLevel,
--- a/internal/pkg/cli/storage_init.go
+++ b/internal/pkg/cli/storage_init.go
@@ -428,6 +428,11 @@ func (o *initStorageOpts) validateOrAskLifecycle() error {
 	if err == nil {
 		o.lifecycle = lifecycleWorkloadLevel
 		o.workloadExists = true
+		log.Infof("%s %s %s\n",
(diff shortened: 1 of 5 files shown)


internal/pkg/cli/storage_init.go
```diff
@@ -395,25 +395,105 @@ func (o *initStorageOpts) askStorageNameWithDefault(friendlyText, defaultName st
...
+		prompt.WithFinalMessage("Lifecycle: "))
+	if err != nil {
+		return fmt.Errorf("ask for lifecycle: %w", err)
+	}
+	o.lifecycle = lifecycle
 	return nil
 }
 
-func (o *initStorageOpts) validateWorkloadName() error {
+func (o *initStorageOpts) validateStorageLifecycle() error {
```
Lou1415926, line 476: This is just moving the func that was previously at L231 to here

internal/pkg/cli/storage_init.go
```diff
@@ -60,6 +53,16 @@ const (
...
 
+const (
+	lifecycleEnvironmentLevel = "environment"
+	lifecycleWorkloadLevel    = "workload"
+
+	lifecycleEnvironmentFriendlyText = "No, the storage should be created and deleted at the environment level"
+	fmtLifecycleWorkloadFriendlyText = "Yes, the storage should be created and deleted at the same time as %s"
+)
+
+var validLifecycleOptions = []string{lifecycleWorkloadLevel, lifecycleEnvironmentLevel}
```
iamhopaul123, line 64: nit: does `workloadLevelLifecycle` make more sense?

internal/pkg/cli/storage_init.go
```diff
@@ -395,25 +395,105 @@ func (o *initStorageOpts) askStorageNameWithDefault(friendlyText, defaultName st
...
+			return nil
+		}
+	}
+	return fmt.Errorf("invalid lifecycle; must be one of %s", english.OxfordWordSeries(quoteStringSlice(validLifecycleOptions), "or"))
+}
+
+// validateWorkloadNameWithLifecycle requires the workload to be in the workspace if the storage lifecycle is on workload-level.
+// Otherwise, it only caches whether the workload is present.  
+func (o *initStorageOpts) validateWorkloadNameWithLifecycle() error {
+	if o.workloadExists {
```
iamhopaul123, line 488: Can we possibly remove `o.workloadExists` and `o.cycle` and just use `o.workloadName`, `o.envName` to determine if the cycle is env level or workload level? If they pass workload name or env name through flags, in `Validate` or `Ask` we'll check if it exists and error out if both workload and env flags are set. Otherwise, we ask if they want env level or workload level and if so which workload/env they want to use.

Lou1415926: I don't follow 🤔 . There is no `o.envName`.  Environment-level storage is not env-specific. Besides, it is entirely possible that people run
```
copilot storage int --workload api --lifecycle environment
```
in a workspace that does not have `api`.

iamhopaul123: Ah I see.

iamhopaul123 (approved):

internal/pkg/cli/storage_init.go
```diff
@@ -636,6 +716,10 @@ func (o *initStorageOpts) Execute() error {
 }
 
 func (o *initStorageOpts) readWorkloadType() error {
+	if !o.workloadExists {
+		o.workloadType = workloadTypeNonLocal
```
efekarakus, line 720: what's the point of type non-local? is it going to be used in a follow-up PR?

efekarakus (approved):

internal/pkg/cli/storage_init.go
Lou1415926: Not really!

I thought of `o.workloadType == ""` to mean that "the workload type is not read yet", which is different than that "the workload type does not need to be read". Therefore, I thought keeping `workloadType` as `""` in this case may potentially communicate the wrong message.

PR #4466, opened 2023-02-07
refactor: convert NLB listener and target group to a list

paragbhingre:
#4170 
This PR refactors the stack pks and cfn templates for NLB to modify listener rules and target groups into a list. We will still accept only one listener and target group at the moment. After the addition of `additional_rules` this list will be populated with the more listener rules and target groups. 


internal/pkg/template/workload.go
```diff
@@ -282,6 +282,30 @@ func ImportedVariable(name string) Variable {
...
+func uniqueAliasesForARecords(aliases []string, uniqueMap map[string]bool) []string {
+	var list []string
+	for _, entry := range aliases {
+		if _, value := uniqueMap[entry]; !value {
+			uniqueMap[entry] = true
+			list = append(list, entry)
+		}
+	}
+	return list
+}
```
iamhopaul123: We should probably avoid using `get` in golang functions. How about `UniqueAliases`? Also it doesn't seem to add too much value for `uniqueAliasesForARecords`. Maybe consider using a nested `for` loop.

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -11,57 +11,58 @@ PublicNetworkLoadBalancer:
             !Sub '${AppName}-${EnvName}-PublicSubnets'
     Type: network
 
-NLBListener:
+{{- range $listener := .NLB.Listener}}
+NLBListenerFor{{ $listener.Port }}:
```
iamhopaul123: Why do we need so many listeners...can we not have one listener with multiple target groups?

paragbhingre: In NLB, we have a one-to-one mapping of nlb port vs. container port. Due to that reason, we have to create an NLBListener for each routing rule. Also, as opposed to the public load balancer, which has multiple listener rules per ALB listener, NLB doesn't have that facility.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -489,17 +494,20 @@ func (s *LoadBalancedWebService) convertNetworkLoadBalancer() (networkLoadBalanc
...
+					TargetPort:      targetPort,
+					SSLPolicy:       nlbConfig.SSLPolicy,
+					Aliases:         aliases,
+					HealthCheck:     hc,
+					Stickiness:      nlbConfig.Stickiness,
+				},
 			},
-			MainContainerPort: s.manifest.MainContainerPort(),
+			MainContainerPort:   s.manifest.MainContainerPort(),
+			CertificateRequired: certRequired,
```
efekarakus: nit: we can get rid of the temp variable
```suggestion
			CertificateRequired: strings.ToUpper(aws.StringValue(protocol)) == manifest.TLS,
```

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
efekarakus: Is this backwards compatible?

If someone has already deployed an NLB, by changing the LogicalID it will result in the listener to be deleted and then recreated. My guess is that it will lead to downtime 💭 

:thinkspin: methinks that we would need something like:

```
# Keep existing resource
NLBListener: 
  ...

{{- if gt (len .NLB.Listener) 1 }}
{{- range $i, $listener := .NLB.Listener }}
NLBListener{{$i + 1}}:
   # Remaining properties...
{{- end }}  
```

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -130,7 +133,7 @@ NLBCustomDomainAction:
     ServiceName: !Ref WorkloadName
     RootDNSRole: {{ .AppDNSDelegationRole }}
     DomainName:  {{ .AppDNSName }}
-    Aliases: !Split [",", !Ref NLBAliases]
+    Aliases: {{ if .NLB.GetAliases }} {{ fmtSlice .NLB.GetAliases }} {{ else }} [] {{ end }}
```
efekarakus: Interesting, why this change? is it some sort of bug fix

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -193,7 +196,7 @@ NLBCustomDomainRole:
     ManagedPolicyArns:
       - !Sub arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
 {{- end}}
-{{- if eq .NLB.Listener.Protocol "TLS" }}
+{{- if .NLB.CertificateRequired }}
```
efekarakus, line 198: Out of curiosity why this change? isn't it equivalent to before.

Or is the idea that in the next PRs we will support multiple listeners and we need a single field to check across all listeners if there is `tls` as protocol?

internal/pkg/template/workload.go
```diff
@@ -282,6 +282,30 @@ func ImportedVariable(name string) Variable {
 	return importedEnvVar(name)
 }
 
+// GetAliases return all the unique aliases specified across all the routing rules in NLB.
+// Currently, we only have primary routing rule, but we will be getting additional routing rule soon.
+func (cfg *NetworkLoadBalancer) GetAliases() []string {
```
efekarakus: nit: 👀 Check out the first bullet point here: https://google.github.io/styleguide/go/best-practices#naming-conventions

internal/pkg/template/workload.go
```diff
@@ -282,6 +282,30 @@ func ImportedVariable(name string) Variable {
...
+	var uniqueAliases []string
+	uniqueAliasMap := make(map[string]bool)
```
```suggestion
	seen := make(map[string]bool)
```

efekarakus:
Looks good! Mostly a question on backwards compatibility

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
paragbhingre: So previously, we had only one listener and `NLBAliases` used to get value from that single listener. Now when we will have multiple listeners then instead of computing `NLBAliases` prior to coming to this  point we are building it from `GetAliases` method which also makes sure all the aliases that we get as response are unique no matter how customer has mentioned them in the listeners. For example, customer can specify 2 same aliases in 2 different listener rules.

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
paragbhingre: Yes, you are right. For now there is only one listener so we can easily pick a value from `.NLB.Listener.Protocol` but going forward we can't keep spanning all the listeners. If we have even one listener with TLS then we will set this flag to true and call cert create and validation function.

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
paragbhingre: That's a great call out. I am not sure if this will cause a downtown but definitely block customers with the error saying Listener for particular port already exist. I will make the required changes.

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -11,57 +11,113 @@ PublicNetworkLoadBalancer:
             !Sub '${AppName}-${EnvName}-PublicSubnets'
     Type: network
 
+{{- if eq (len .NLB.Listener) 1 }}
+{{- range $listener := .NLB.Listener }}
```
efekarakus: oooh I wonder if instead of a big if-else statement we could write it this way:
```
{{- range $i, $listener := .NLB.Listener }}
{{if eq $i 0}}
NLBListener:
{{- else }}
NLBListener{{$i + 1}}:
{{- end }}
```

This way if there is a bug fix we don't need to remember to update the properties of the same resource in two different places

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
```diff
@@ -111,7 +111,13 @@ Resources:
     {{- end}}
     {{- end}}
     {{- if .NLB}}
+    {{- if eq (len .NLB.Listener) 1 }}
       - NLBListener
+    {{- else if gt (len .NLB.Listener) 1 }}
+    {{- range $listener := .NLB.Listener }}
+      - NLBListenerFor{{$listener.Port}}
+    {{- end }}
+    {{- end }}
```
efekarakus: nit: I think we can perhaps simplify this slightly we adopt a different naming pattern:
```
{{- range $i, $listener := .NLB.Listener }}
- NLBListener{{- if ne $i 0}}{{$i + 1}}{{- end}}
{{- end }}
```

efekarakus (commented):
Looks great, just some suggestions to simplify the templating

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
paragbhingre: Yeah you are right, I thought you are suggesting me to keep existing logic as is. But I agree we can just have one block doing both the conditions.

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
paragbhingre: This naming pattern is appealing to me. I'm just curious if there's a reason for adding `1` to every `$i` in the loop. Because we are anyway keeping the first resource as `NLBListener` or `NLBTargetGroup` and then appending an index to it, which would start at 1. In the event that we have two listeners, we will have `NLBListener` and `NLBListener1` which seems fine to me. Or you want `NLBListener` and `NLBListener2`  for some reason.

efekarakus: oh I like your suggestion even more! starting from 2 :shipit: but yea that was why I was doing +1

internal/pkg/deploy/cloudformation/stack/transformers.go
efekarakus: can we get rid of the temp variable? the conversation was marked as resolved but the change wasn't updated :)

internal/pkg/template/templates/workloads/partials/cf/nlb.yml
```diff
@@ -10,58 +10,66 @@ PublicNetworkLoadBalancer:
         - Fn::ImportValue:
             !Sub '${AppName}-${EnvName}-PublicSubnets'
     Type: network
-
+{{- range $i, $listener := .NLB.Listener }}
+{{- if eq $i 0 }}
 NLBListener:
+{{- else }}
+NLBListener{{$i}}:
```
efekarakus: we can do the same thing here now 👍 
```suggestion
NLBListener{{- if ne $i 0 }}{{$i +2}}{{- end}}:
```

efekarakus (approved):
Sweet, approving with the `do-not-merge` label feel free to remove it once the +2 and temp variable feedback is addressed

iamhopaul123 (approved):

PR #4467, opened 2023-02-07
chore(manifest): Populate Sidecar docker build arguments using BuildConfig function

KollaAdithya:
Related to #4254 
This PR will populate the docker build arguments for the Sidecar container images using `BuildConfig` function

### DIFF
--- a/internal/pkg/manifest/workload_ecs.go
+++ b/internal/pkg/manifest/workload_ecs.go
@@ -334,7 +334,7 @@ func (cfg *SidecarConfig) ImageURI() (string, bool) {
 func (cfg *SidecarImageConfig) BuildConfig(rootDirectory string) *DockerBuildArgs {
 	df := cfg.dockerfile()
 	ctx := cfg.context()
-	dockerfile, context := getDockerfileAndContext(rootDirectory, df, ctx)
+	dockerfile, context := dockerfileAndContext(rootDirectory, df, ctx)
 	return &DockerBuildArgs{
 		Dockerfile: dockerfile,
 		Context:    context,
--- a/internal/pkg/manifest/workload_ecs.go
+++ b/internal/pkg/manifest/workload_ecs.go
@@ -330,6 +330,55 @@ func (cfg *SidecarConfig) ImageURI() (string, bool) {
 	return "", false
 }
 
+// BuildConfig populates a docker.BuildArguments struct from the sidecar image fields available in the manifest.
+func (cfg *SidecarImageConfig) BuildConfig(rootDirectory string) *DockerBuildArgs {
+	df := cfg.dockerfile()
+	ctx := cfg.context()
+	dockerfile, context := getDockerfileAndContext(rootDirectory, df, ctx)
+	return &DockerBuildArgs{
+		Dockerfile: dockerfile,
+		Context:    context,
+		Args:       cfg.args(),
+		Target:     cfg.target(),
+		CacheFrom:  cfg.cacheFrom(),
+	}
+}
+
+// dockerfile returns the path to the sidecar container's Dockerfile if one is set.
+// If no dockerfile is specified, returns "".
+func (cfg *SidecarImageConfig) dockerfile() string {
+	if cfg.Build.Advanced.Dockerfile != nil {
+		return aws.StringValue(cfg.Build.Advanced.Dockerfile)
+	}
+	var dfPath string
+	if cfg.Build.Basic != nil {
+		dfPath = aws.StringValue(cfg.Build.Basic)
+	}
+	return dfPath
+}
+
+// context returns the build context directory of sidecar container's if it exists, otherwise an empty string.
+func (cfg *SidecarImageConfig) context() string {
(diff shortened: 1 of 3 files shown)


internal/pkg/manifest/workload_ecs.go
```diff
@@ -330,6 +330,55 @@ func (cfg *SidecarConfig) ImageURI() (string, bool) {
...
+		Args:       cfg.args(),
+		Target:     cfg.target(),
+		CacheFrom:  cfg.cacheFrom(),
+	}
+}
+
+// dockerfile returns the path to the sidecar container's Dockerfile if one is set.
+// If no dockerfile is specified, returns "".
+func (cfg *SidecarImageConfig) dockerfile() string {
+	if cfg.Build.Advanced.Dockerfile != nil {
```
iamhopaul123: Maybe consider using `case`?

iamhopaul123 (approved):

internal/pkg/manifest/workload_ecs.go
```diff
@@ -330,6 +330,56 @@ func (cfg *SidecarConfig) ImageURI() (string, bool) {
 	return "", false
 }
 
+// BuildConfig populates a docker.BuildArguments struct from the sidecar image fields available in the manifest.
+func (cfg *SidecarImageConfig) BuildConfig(rootDirectory string) *DockerBuildArgs {
```
Lou1415926, line 334: If you think https://github.com/aws/copilot-cli/pull/4430/files#r1101811523 sounds good, then we probably won't need to repeat these all these logic. If I am not mistaken, these seem to be almost exactly the same as these https://github.com/aws/copilot-cli/blob/764b324c2b29be013d74dc47a91d02dd18c4130c/internal/pkg/manifest/workload.go#L200 
```go
// dockerfile returns the path to the workload's Dockerfile. If no dockerfile is specified,
// returns "".
func (i *Image) dockerfile() string {
	// Prefer to use the "Dockerfile" string in BuildArgs. Otherwise,
	// "BuildString". If no dockerfile specified, return "".
	if i.Build.BuildArgs.Dockerfile != nil {
		return aws.StringValue(i.Build.BuildArgs.Dockerfile)
	}

	var dfPath string
	if i.Build.BuildString != nil {
		dfPath = aws.StringValue(i.Build.BuildString)
	}

	return dfPath
}

// context returns the build context directory if it exists, otherwise an empty string.
func (i *Image) context() string {
	return aws.StringValue(i.Build.BuildArgs.Context)
}

// args returns the args section, if it exists, to override args in the dockerfile.
// Otherwise it returns an empty map.
func (i *Image) args() map[string]string {
	return i.Build.BuildArgs.Args
}

// target returns the build target stage if it exists, otherwise nil.
func (i *Image) target() *string {
	return i.Build.BuildArgs.Target
}

// cacheFrom returns the cache from build section, if it exists.
// Otherwise it returns nil.
func (i *Image) cacheFrom() []string {
	return i.Build.BuildArgs.CacheFrom
}

```

KollaAdithya: yes, These functions look exactly similar to [copilot-cli/internal/pkg/manifest/workload.go](https://github.com/aws/copilot-cli/blob/764b324c2b29be013d74dc47a91d02dd18c4130c/internal/pkg/manifest/workload.go#L200)

correct me if am dumb, Even after implementing this https://github.com/aws/copilot-cli/pull/4430/files#r1101811523

If I want to loop over each sidecar and return the `DockerBuildArgs` like this below
```
func (s *LoadBalancedWebService) SidecarBuildArgs(wsRoot string) map[string]*DockerBuildArgs {
	sc := s.LoadBalancedWebServiceConfig.Sidecars
	sidecarBuildArgs := make(map[string]*DockerBuildArgs, len(sc))
	for k, v := range sc {
		sidecarBuildArgs[k] = v.Image.Advanced.BuildConfig(wsRoot)
	}
	return sidecarBuildArgs
}
```
then I would still need to write all these functions 
```
// BuildConfig populates a docker.BuildArguments struct from the sidecar image fields available in the manifest.
func (cfg *ImageLocationOrBuild) BuildConfig(rootDirectory string) *DockerBuildArgs {
	df := cfg.dockerfile()
	ctx := cfg.context()
	dockerfile, context := dockerfileAndContext(rootDirectory, df, ctx)
	return &DockerBuildArgs{
		Dockerfile: dockerfile,
		Context:    context,
		Args:       cfg.args(),
		Target:     cfg.target(),
		CacheFrom:  cfg.cacheFrom(),
	}
}

// dockerfile returns the path to the sidecar container's Dockerfile if one is set.
// If no dockerfile is specified, returns "".
func (cfg *ImageLocationOrBuild) dockerfile() string {
	switch {
	case cfg.Build.Advanced.Dockerfile != nil:
		return aws.StringValue(cfg.Build.Advanced.Dockerfile)
	case cfg.Build.Basic != nil:
		return aws.StringValue(cfg.Build.Basic)
	default:
		return ""
	}

}
```

PR #4469, opened 2023-02-07
chore: add static site manifest and init

iamhopaul123:

### DIFF
--- a/internal/pkg/cli/svc_init.go
+++ b/internal/pkg/cli/svc_init.go
@@ -107,7 +107,7 @@ var serviceTypeHints = map[string]string{
 	manifest.LoadBalancedWebServiceType:  "Internet to ECS on Fargate",
 	manifest.BackendServiceType:          "ECS on Fargate",
 	manifest.WorkerServiceType:           "Events to SQS to ECS on Fargate",
-	manifest.StaticSiteType:              "S3 Bucket for static assets",
+	manifest.StaticSiteType:              "Internet to CDN to S3 bucket",
 }
 
 type initWkldVars struct {
--- a/internal/pkg/cli/svc_init.go
+++ b/internal/pkg/cli/svc_init.go
@@ -303,25 +303,7 @@ func (o *initSvcOpts) Ask() error {
 	if shouldSkipAsking {
 		return nil
 	}
-	if o.wkldType == manifest.StaticSiteType {
-		return o.askStaticSite()
-	}
-	err = o.askDockerfile()
-	if err != nil {
-		return err
-	}
-	if o.dockerfilePath == "" {
-		if err := o.askImage(); err != nil {
-			return err
-		}
-	}
-	if err := o.askSvcPort(); err != nil {
-		return err
-	}
-	if err := o.askSvcPublishers(); err != nil {
-		return err
-	}
-	return nil
+	return o.askSvcDetails()
 }
 
 // Execute writes the service's manifest file and stores the service in SSM.
@@ -383,6 +365,25 @@ func (o *initSvcOpts) RecommendActions() error {
 			defaultEnvironmentName),
 	})
 	return nil
+}
+
+func (o *initSvcOpts) askSvcDetails() error {
+	if o.wkldType == manifest.StaticSiteType {
+		return o.askStaticSite()
+	}
+	err := o.askDockerfile()
+	if err != nil {
+		return err
+	}
+	if o.dockerfilePath == "" {
+		if err := o.askImage(); err != nil {
+			return err
+		}
+	}
+	if err := o.askSvcPort(); err != nil {
+		return err
+	}
+	return o.askSvcPublishers()
 }
 
 func (o *initSvcOpts) askSvcType() error {
--- a/internal/pkg/cli/svc_init.go
+++ b/internal/pkg/cli/svc_init.go
@@ -107,6 +107,7 @@ var serviceTypeHints = map[string]string{
 	manifest.LoadBalancedWebServiceType:  "Internet to ECS on Fargate",
 	manifest.BackendServiceType:          "ECS on Fargate",
 	manifest.WorkerServiceType:           "Events to SQS to ECS on Fargate",
+	manifest.StaticSiteType:              "S3 Bucket for static assets",
 }
 
 type initWkldVars struct {
@@ -302,6 +303,9 @@ func (o *initSvcOpts) Ask() error {
 	if shouldSkipAsking {
 		return nil
 	}
+	if o.wkldType == manifest.StaticSiteType {
+		return o.askStaticSite()
+	}
 	err = o.askDockerfile()
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


internal/pkg/config/workload.go
```diff
@@ -186,7 +186,7 @@ func (s *Store) ListServices(appName string) ([]*Workload, error) {
 func (s *Store) ListJobs(appName string) ([]*Workload, error) {
 	wklds, err := s.listWorkloads(appName)
 	if err != nil {
-		return nil, fmt.Errorf("read service configuration for application %s: %w", appName, err)
+		return nil, fmt.Errorf("read job configuration for application %s: %w", appName, err)
```
Lou1415926, line 189: do we need this error wrapping?

internal/pkg/template/templates/workloads/services/static-site/manifest.yml
```diff
@@ -0,0 +1,43 @@
...
+type: {{.Type}}
+
+
+{{- if not .FileUploads}}
+# files:
+#   - source: ./static
+#     destination: static
+{{- else}}
+files:
+{{- range $file := .FileUploads}}
```
paragbhingre, line 18: Sorry I am dumb, I am not able to understand why we are looping over FileUploads and setting source, destination and so on same in every loop?

Lou1415926 (approved):

internal/pkg/cli/svc_init.go
```diff
@@ -107,6 +107,7 @@ var serviceTypeHints = map[string]string{
 	manifest.LoadBalancedWebServiceType:  "Internet to ECS on Fargate",
 	manifest.BackendServiceType:          "ECS on Fargate",
 	manifest.WorkerServiceType:           "Events to SQS to ECS on Fargate",
+	manifest.StaticSiteType:              "S3 Bucket for static assets",
```
efekarakus: What do you think of this boss?
```suggestion
	manifest.StaticSiteType:              "Internet to CDN to S3 bucket",
```

internal/pkg/cli/svc_init.go
```diff
@@ -302,6 +303,9 @@ func (o *initSvcOpts) Ask() error {
 	if shouldSkipAsking {
 		return nil
 	}
+	if o.wkldType == manifest.StaticSiteType {
+		return o.askStaticSite()
+	}
 	err = o.askDockerfile()
```
efekarakus: nit: I wonder if there is a broader pattern here? Something like:
```go
return o.askServiceDetails()

func askServiceDetails() {
   if o.workloadType == manifest.StaticSiteType {
      return o.askStaticSiteDetails()
   }
   askDockerfile()
   if isExposedService(o.workloadType) {
      o.askSvcPort()
   }
   if o.workloadType == workerService {
      o.askSvcPublishers()
   }
}
```

internal/pkg/manifest/svc.go
```diff
@@ -27,7 +28,7 @@ const (
 	// WorkerServiceType is a worker service that manages the consumption of messages.
 	WorkerServiceType = "Worker Service"
 	// StaticSiteType is a static site service that manages static assets.
-	StaticSiteType = "Static Site"
+	StaticSiteType = "Static Site Service"
```
efekarakus: Hmmmmm :thinkspin: I kinda liked the previous version. Can you check with @genbit about our preference?

internal/pkg/template/templates/workloads/services/static-site/manifest.yml
```diff
@@ -0,0 +1,43 @@
...
+# Your service name will be used in naming your resources like S3 buckets, etc.
+name: {{.Name}}
+type: {{.Type}}
+
+
+{{- if not .FileUploads}}
+# files:
+#   - source: ./static
+#     destination: static
+{{- else}}
```
efekarakus: what if we made our default:
```suggestion
{{- if not .FileUploads}}
files:
   - source: ./
     destination: assets/
     exclude: 
     - copilot/
{{- else}}
```

internal/pkg/template/templates/workloads/services/static-site/manifest.yml
```diff
@@ -0,0 +1,43 @@
...
+files:
+{{- range $file := .FileUploads}}
+  - {{- if $.Source}}
+    source: {{$.Source}}
+    {{- end}}
+    {{- if $.Destination}}
+    destination: {{$.Destination}}
+    {{- end}}
```
efekarakus: is there no mandatory field here?

internal/pkg/template/templates/workloads/services/static-site/manifest.yml
```diff
@@ -0,0 +1,43 @@
...
+    context: {{$.Context}}
+    {{- end}}
+    {{- if $.Recursive}}
+    recursive: {{$.Recursive}}
+    {{- end}}
+    {{- if $.Reinclude.StringSlice}}
+    Reinclude: {{$.Reinclude.StringSlice}}
+    {{- end}}
+    {{- if $.Exclude.StringSlice}}
+    Exclude: {{$.Exclude.StringSlice}}
```
efekarakus:
```suggestion
    reinclude: {{$.Reinclude.StringSlice}}
    {{- end}}
    {{- if $.Exclude.StringSlice}}
    exclude: {{$.Exclude.StringSlice}}
```

efekarakus (commented):
LGTM some suggestions below!

internal/pkg/config/workload.go
iamhopaul123: I think it's still worth because it is to avoid using "workload" term in error msg. For example: `read service configuration for application chicken: invalid character 'o' looking for beginning of value`

internal/pkg/manifest/svc.go
iamhopaul123: We check if it ends with "Service" to determine if it is a service or job in `config` pkg. Do you think we can make `Static Site` an exception?

internal/pkg/cli/svc_init.go
iamhopaul123: 🤔 it's not only for cdn and possibly for ALB right

internal/pkg/manifest/svc.go
efekarakus: 🤦 I feel like we could have something like this:
```go
package manifest

type Type string
const (
   StaticSiteType = "Static Site"
)

var serviceTypes = []Type{StaticSiteType, WorkerServiceType, ...}

func (t Type) IsService() bool {
  for _, valid range := serviceTypes {
     if valid == t {
        return true
     }
  }
  return false
}
```
what do you think?

iamhopaul123: next pr i'll move types to a new pkg

internal/pkg/template/templates/workloads/services/static-site/manifest.yml
efekarakus: Don't say that ! it's not at all dumb

My understanding is that `.FileUploads` is more like "directories". For example, if we wanted to upload 2 different directories this would translate to:
1. First, upload `source: 'src/static/images'` to the S3 bucket under `destination: 'assets/'`
2. Then, upload `source: 'src/static/js'` to the S3 bucket under `destination: 'assets/'`

Hope that helps understands what the field represents!

efekarakus (approved):

PR #4472, opened 2023-02-08
refactor(deploy): use test double in tests instead of init a stack

efekarakus:
Today, we're creating a real `stack` object in our `cli/deploy` unit tests. When modifying the constructors in the `stack` pkg that results in all the tests in `cli/deploy` to fail because some dependency goes missing.

This change allows the unit tests in `cli/deploy` to pass in a stub CFN stack instead of initializing a real one.

### DIFF
--- a/internal/pkg/cli/deploy/backend.go
+++ b/internal/pkg/cli/deploy/backend.go
@@ -20,8 +20,11 @@ import (
 
 type backendSvcDeployer struct {
 	*svcDeployer
-	backendMft         *manifest.BackendService
+	backendMft *manifest.BackendService
+
+	// Overriden in tests.
 	aliasCertValidator aliasCertValidator
+	newStack           func() cloudformation.StackConfiguration
 }
 
 // NewBackendDeployer is the constructor for backendSvcDeployer.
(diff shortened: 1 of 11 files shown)


iamhopaul123 (approved):

internal/pkg/cli/deploy/backend.go
```diff
@@ -20,8 +20,11 @@ import (
 
 type backendSvcDeployer struct {
 	*svcDeployer
-	backendMft         *manifest.BackendService
+	backendMft *manifest.BackendService
+
+	// Overriden in tests.
 	aliasCertValidator aliasCertValidator
+	newStack           func() cloudformation.StackConfiguration
```
Lou1415926, line 28: What do you think of doing something similar with `cli/deploy/env.go`? So that we can use `NewMockStackConfiguration` instead of writing our own stub.
```go
// In cli/deploy/backend.go
type  backendSvcDeployer struct {
    newStack func() cloudformation.StackConfiguration
}

func NewBackendDeployer(in *WorkloadDeployerInput) (*backendSvcDeployer, error) {
    return &backendSvcDeployer{
		newStack: func() cloudformation.StackConfiguration {
                       return stack.NewBackendService(stack.BackendServiceConfig{
			App:                d.app,
			EnvManifest:        d.envConfig,
			Manifest:           d.backendMft,
			RawManifest:        d.rawMft,
			ArtifactBucketName: d.resources.S3Bucket,
			RuntimeConfig:      *rc,
			Addons:             d.addons,
		     })
                 }
	}, nil 
}

// In cli/deploy/backend_test.go
func TestBackendServiceDeployer_stackConfiguration(t *testing.T) {
   ....
   stackSerializer: cfnmocks.NewMockStackConfiguration(ctrl),
   deployer := &backendSvcDeployer{...}
   deployer.newStack = func () cloudformation.StackConfiguration{ return stackSerializer }
}

```

Lou1415926 (approved):

internal/pkg/cli/deploy/backend.go
efekarakus: I wish :cool-crying: if you look at the lbws instead of backend you can see that it's not possible :'(

I think the real solution is to use dependency injection properly in both `cli/deploy` and `stack`. Then I could have just passed down some stubs and we wouldn't need to write a `newStack` field in the first place

PR #4473, opened 2023-02-08
refactor(storage): create single purpose constructors

Lou1415926:
Two major changes:
1. Previously, `WorkloadType`  are passed in as props to RDS constructors. Depending on whether `WorkloadType` is RDWS or not, the constructor decides what the template path is. This PR removes `WorkloadType` and instead create different constructors designed for RDWS vs. non-RDWS. 
2. Previously, AccessPolicy and Ingress uses the same `Template` struct as 

### DIFF
--- a/internal/pkg/addon/storage.go
+++ b/internal/pkg/addon/storage.go
@@ -81,16 +81,6 @@ func EnvS3Template(input *S3Props) *S3Template {
 	}
 }
 
-// EnvS3AccessPolicyTemplate creates a new marshaler for the access policy attached to a workload
-// for permissions into an environment-level S3 addon.
-func EnvS3AccessPolicyTemplate(input *S3Props) *S3Template {
-	return &S3Template{
-		S3Props:  *input,
-		parser:   template.New(),
-		tmplPath: envS3AccessPolicyTemplatePath,
-	}
-}
-
 // S3Template contains configuration options which fully describe an S3 bucket.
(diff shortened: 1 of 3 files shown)


iamhopaul123 (approved):

internal/pkg/addon/storage.go
```diff
@@ -248,9 +226,47 @@ type DDBLocalSecondaryIndex struct {
 	Name         *string
 }
 
+// AccessPolicyProps TODO
```
efekarakus: ? 😂  (same for `AccessPolicyTemplate`, and `RDSIngressProps` comment)

internal/pkg/addon/storage.go
```diff
@@ -259,53 +276,91 @@ type RDSProps struct {
...
 	}
+}
+
+// RDWSServerlessV1Template creates a marshaler for an Aurora Serverless v1 addon attached on an RDWS.
+func RDWSServerlessV1Template(input RDSProps) *RDSTemplate {
 	return &RDSTemplate{
 		RDSProps: input,
 		parser:   template.New(),
-		tmplPath: tmplPath,
+		tmplPath: rdsRDWSTemplatePath,
```
efekarakus, line 292: completely useless feedback: but I keep reading this as `rdsRDS` lol

efekarakus (approved):

PR #4478, opened 2023-02-09
refactor: move manifest type to a new pkg

iamhopaul123:

### DIFF
--- a/internal/pkg/manifest/manifestinfo/type.go
+++ b/internal/pkg/manifest/manifestinfo/type.go
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-package manifesttype
+package manifestinfo
 
 const (
 	// LoadBalancedWebServiceType is a web service with a load balancer and Fargate as compute.
--- a/internal/pkg/manifest/manifestinfo/type.go
+++ b/internal/pkg/manifest/manifestinfo/type.go
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+// Package manifestinfo provides access to information embedded in a manifest.
 package manifestinfo
 
 const (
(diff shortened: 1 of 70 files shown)


internal/pkg/manifest/manifesttype/type.go
```diff
@@ -0,0 +1,61 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
Lou1415926: I wonder what's the cons of just having `internal/pkg/manifest/type.go` ? instead of having a new package

efekarakus (commented):
How much would you hate me if we rename the package to `manifestinfo`? (to avoid the double "t"s)

I was looking to see if there is anything similar in the stdlib and ran into this package: https://pkg.go.dev/debug/buildinfo@go1.20

internal/pkg/manifest/manifesttype/type.go
iamhopaul123: it's a dependency problem. Mainly for mft depending on `config`

internal/pkg/manifest/manifestinfo/type.go
```diff
@@ -0,0 +1,61 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package manifestinfo
```
efekarakus, line 5: Package docs boss 🥺

internal/pkg/manifest/manifestinfo/type.go
```diff
@@ -0,0 +1,61 @@
...
+	return []string{
+		RequestDrivenWebServiceType,
+		LoadBalancedWebServiceType,
+		BackendServiceType,
+		WorkerServiceType,
+	}
+}
+
+// JobTypes returns the list of supported job manifest types.
+func JobTypes() []string {
```
efekarakus, line 33: unit tests?

efekarakus (approved):
Approving with label assuming the comments will be addressed!

Lou1415926 (approved):

Lou1415926:
should have merged this after mine 😤 got a million conflicts now!

PR #4479, opened 2023-02-09
chore: execute `--add-ingress-from` for `storage init`

Lou1415926:
Will rebase on #4473


internal/pkg/cli/storage_init.go
```diff
@@ -926,61 +942,77 @@ func (o *initStorageOpts) wkldRDSAddonBlobs() ([]addonBlob, error) {
...
+	rdwsIngressTmplBlob := addonBlob{
+		path:        o.ws.WorkloadAddonFilePath(o.workloadName, fmt.Sprintf("%s-ingress.yml", o.storageName)),
+		description: "template",
+		blob: addon.EnvServerlessRDWSIngressTemplate(addon.RDSIngressProps{
+			ClusterName: o.storageName,
+			Engine:      o.rdsEngine,
+		}),
+	}
+	rdwsIngressParamBlob := addonBlob{
+		path:        o.ws.WorkloadAddonFilePath(o.workloadName, "addons.parameters.yml"),
```
iamhopaul123: do we need to make `addons.parameters.yml` a const?

internal/pkg/cli/storage_init.go
```diff
@@ -926,61 +942,77 @@ func (o *initStorageOpts) wkldRDSAddonBlobs() ([]addonBlob, error) {
...
-		},
+	tmplBlob := addonBlob{
+		path:        o.ws.EnvAddonFilePath(fmt.Sprintf("%s.yml", o.storageName)),
+		description: "template",
+		blob:        addon.EnvServerlessTemplate(props),
 	}
-	if o.workloadType != manifestinfo.RequestDrivenWebServiceType || !o.workloadExists {
-		return blobs, nil
+	paramBlob := addonBlob{
+		path:        o.ws.EnvAddonFilePath("addons.parameters.yml"),
```
iamhopaul123: same for this

iamhopaul123 (approved):

internal/pkg/cli/storage_init.go
```diff
@@ -724,6 +724,13 @@ func (o *initStorageOpts) validateOrAskAuroraInitialDBName() error {
 
 // Execute deploys a new environment with CloudFormation and adds it to SSM.
 func (o *initStorageOpts) Execute() error {
+	if o.addIngressFrom != "" {
```
efekarakus: I wonder if it's worthwhile encapsulating this in a separate method, something like:
```go
func (o *initStorageOpts) Execute() error {
    o.consumeFlags()
}
```

efekarakus: As a larger question, I assume our flow should be like this:
```
Validate() // Never sets any internal state and just checks if flags are compatible.
Ask() // Sets only flag values in the `*Vars` struct
Execute() // Sets all internal state
```

efekarakus (approved):

internal/pkg/cli/storage_init.go
Lou1415926: There is one internal state that gets set during `Validate` and `Ask`: `o.workloadExists`. The intention was to reduce function calls.

However, the cost of that function call isn't large though (one more fs read), so the pros brought by caching internal state in `Validate`/`Ask` (thus going against the pattern) is marginal. So I changed the implementation to ⬇️ 

https://github.com/aws/copilot-cli/pull/4479/commits/3a0bde20c82f5d5ae7943deb65087ae7b51799c7

PR #4481, opened 2023-02-10
chore: implement `ImageLocationOrBuild` general struct useful for both main and sidecar images

KollaAdithya:
In the previous PR(https://github.com/aws/copilot-cli/pull/4430) (already merged ) we added sidecar image build fields using `SidecarImageConfig` struct specific to sidecars.
As per Wanxian's feedback https://github.com/aws/copilot-cli/pull/4430#discussion_r1101811523, 
Instead of implementing specific struct to Sidecar images, with the help of this PR will create a  generic `ImageLocationOrBuild` struct that can be used by both main container and Sidecar container images.

Advantage of creating a generic struct `ImageLocationOrBuild` is that we do not need to repeat the [same logic]( https://github.com/aws/copilot-cli/pull/4467#discussion_r1101814180) of populating the docker build args for sidecar images separately and we can use the same logic for both sidecar and main container images.

So I will close PR as we  no longer  need this functions in  https://github.com/aws/copilot-cli/pull/4467


internal/pkg/manifest/workload.go
```diff
@@ -129,11 +129,16 @@ type Workload struct {
...
-	DockerLabels map[string]string `yaml:"labels,flow"`     // Apply Docker labels to the container at runtime.
-	DependsOn    DependsOn         `yaml:"depends_on,flow"` // Add any sidecar dependencies.
+	ImageLocationOrBuild `yaml:",inline"`
+	Credentials          *string           `yaml:"credentials"`     // ARN of the secret containing the private repository credentials.
+	DockerLabels         map[string]string `yaml:"labels,flow"`     // Apply Docker labels to the container at runtime.
+	DependsOn            DependsOn         `yaml:"depends_on,flow"` // Add any sidecar dependencies.
+}
+
+// ImageLocationOrBuild represents the docker build arguments and location of the existing image.
+type ImageLocationOrBuild struct {
```
KollaAdithya, line 135: This is generic struct that can be used by both Main container and sidecar images.

internal/pkg/manifest/validate.go
```diff
@@ -2067,9 +2067,9 @@ func contains(name string, names []string) bool {
 	return false
 }
 
-// validate returns nil if SidecarImageConfig is configured correctly.
-func (cfg SidecarImageConfig) validate() error {
-	if !cfg.Build.IsZero() && cfg.Location != nil {
+// validate returns nil if ImageLocationOrBuild is configured correctly.
+func (i ImageLocationOrBuild) validate() error {
```
iamhopaul123, line 2072: Is this function invoked for the main container?

iamhopaul123 (approved):

internal/pkg/manifest/validate.go
KollaAdithya: No, this is just to validate sidecar containers

iamhopaul123: yeah but main container also includes `ImageLocationOrBuild`. Shall we invoke it in `Image.validate()` to make sure users specifying the correct `image` section in their manifests?

KollaAdithya: yup adding to the `Image` section as well

KollaAdithya: Both cases are handled now well in validation
1. Both `Build` and `location` are empty
2. Both  `Build` and `location` are not empty

internal/pkg/manifest/validate.go
```diff
@@ -653,6 +653,9 @@ func (i ImageWithPort) validate() error {
 // validate returns nil if Image is configured correctly.
 func (i Image) validate() error {
 	var err error
+	if err := i.ImageLocationOrBuild.validate(); err != nil {
+		return err
+	}
 	if err = i.Build.validate(); err != nil {
 		return fmt.Errorf(`validate "build": %w`, err)
```
paragbhingre: I was just wondering if we should keep this code inside `ImageLocationOrBuild.validate()` as this validate belongs to `ImageLocationOrBuild` and not `Image`?

KollaAdithya: L 659 - 660 can be added to `ImageLocationOrBuild.validate()` and the remaining logic L661-668 where both `build` and `location` are empty should be inside the `Image` (main container) only as this case is not applicable for sidecar containers where both `build` and `location` can be empty

KollaAdithya: Done 
Added these below to `ImageLocationOrBuild` as this case is applicable to both sidecars and main container.
```
if err = i.Build.validate(); err != nil {
    return fmt.Errorf(`validate "build": %w`, err)
}

File: internal/pkg/manifest/validate.go [hunk 4]
```diff
@@ -2072,6 +2069,10 @@ func contains(name string, names []string) bool {
 
 // validate returns nil if ImageLocationOrBuild is configured correctly.
 func (i ImageLocationOrBuild) validate() error {
+	var err error
+	if err = i.Build.validate(); err != nil {
```
paragbhingre: Just tiniest nit:
```suggestion
        if err := i.Build.validate(); err != nil {
```

paragbhingre (approved):
Looks good to me apart from the tiniest nit. Good job Adi!
Feel free to remove the label after addressing the nit.

PR #4482, opened 2023-02-10
chore(stack): initialize custom resources in constructors

efekarakus:


efekarakus:
The PR itself isn't actually that big, it's just that a lot of mock files got deleted/changed

internal/pkg/template/templatetest/templatetest.go
```diff
@@ -0,0 +1,77 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package templatetest provides test doubles for embedded templates.
```
iamhopaul123, line 4: what is "test doubles" 🤔

iamhopaul123 (approved):

internal/pkg/template/templatetest/templatetest.go
efekarakus: https://testing.googleblog.com/2013/07/testing-on-toilet-know-your-test-doubles.html

It's the general term for giving a "test object" instead of the real one.
In our code, we typically have:
- Stubs: the custom structs that we ourselves write in the `_test.go` files that return some dummy data.
- Mocks: we use the gomock library to test if an object was called with the right input and how many times.
- Fakes: very rarely we use fakes, the best example is `afero.NewMemMapFS`, but they are really nice for high fidely cheap integ tests

internal/pkg/deploy/cloudformation/stack/workload.go
```diff
@@ -111,6 +111,12 @@ type location interface {
 	GetLocation() string
 }
 
+// uploadable is the interface for an object that can be uploaded to an S3 bucket.
```
Lou1415926, line 114: Is the plan to do something similar to ⬇️  in `Template()`:
```go
func (svc *BackendService) Template() (string, error) {
   for _, uploadable := range svc.localCRs {
      url, err := s3.upload(uploadable.ArtifactPath(), uploadable.Content())
      crURLs[uploadable.Name()] = url
   }
  opt := &template.WorkloadOpts{
      customResources: crURLs
  }
}
```

Lou1415926 (approved):

internal/pkg/deploy/cloudformation/stack/workload.go
efekarakus: yup!

PR #4483, opened 2023-02-10
chore(alarms): quick fix-- avoid calling `AlarmStatuses` with empty opts

huanjani:
The `AlarmStatuses` method (and `DescribeAlarms` API call) allow no arguments, in which case all alarms in the account are returned, without being filtered on name or whatever else. This prevents those calls where we don't have the filtering criteria we need.

### DIFF
--- a/internal/pkg/aws/cloudwatch/cloudwatch.go
+++ b/internal/pkg/aws/cloudwatch/cloudwatch.go
@@ -73,6 +73,9 @@ func (cw *CloudWatch) AlarmsWithTags(tags map[string]string) ([]AlarmStatus, err
 		}
 		alarmNames = append(alarmNames, name)
 	}
+	if len(alarmNames) == 0 {
+		return nil, nil
+	}
 	return cw.AlarmStatuses(WithNames(alarmNames))
 }
 
@@ -94,6 +97,8 @@ func WithPrefix(prefix string) DescribeAlarmOpts {
 }
 
 // AlarmStatuses returns the statuses of alarms optionally filtered (by name, prefix, etc.).
+// If the optional parameter is passed in but is nil, the statuses of ALL alarms in the
+// account will be returned!
 func (cw *CloudWatch) AlarmStatuses(opts ...DescribeAlarmOpts) ([]AlarmStatus, error) {
 	var alarmStatuses []AlarmStatus
 	in := &cloudwatch.DescribeAlarmsInput{}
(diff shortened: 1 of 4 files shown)


internal/pkg/aws/cloudwatch/cloudwatch.go
```diff
@@ -73,6 +73,9 @@ func (cw *CloudWatch) AlarmsWithTags(tags map[string]string) ([]AlarmStatus, err
 		}
 		alarmNames = append(alarmNames, name)
 	}
+	if len(alarmNames) == 0 {
```
efekarakus, line 76: great find! 👏

internal/pkg/aws/cloudwatch/cloudwatch_test.go
```diff
@@ -74,12 +75,45 @@ func TestCloudWatch_AlarmsWithTags(t *testing.T) {
 		"return if no alarms found": {
 			setupMocks: func(m cloudWatchMocks) {
 				m.rg.EXPECT().GetResourcesByTags(cloudwatchResourceType, gomock.Eq(testTags)).Return([]*rg.Resource{}, nil)
-				m.cw.EXPECT().DescribeAlarms(gomock.Any()).Return(nil, nil)
 			},
 
 			wantAlarmStatus: nil,
 		},
-	}
+		"success": {
```
efekarakus: 1. should we add a test case for the scenario where there are no alarms found with `GetResourcesByTags`?
2. Can we make the test case string more verbose, for example:
```suggestion
		"should invoke DescribeAlarms on alarms that have matching tags": {
```

internal/pkg/describe/status_describe.go
```diff
@@ -216,6 +216,9 @@ func (s *ecsStatusDescriber) ecsServiceAutoscalingAlarms(cluster, service string
 	if err != nil {
 		return nil, fmt.Errorf("retrieve auto scaling alarm names for ECS service %s/%s: %w", cluster, service, err)
 	}
+	if alarmNames == nil {
```
efekarakus: nit: like the previous one to catch the empty slice condition
```suggestion
	if len(alarmNames) == 0 {
```

efekarakus (approved):
Looks great, a tiny test suggestion and then feel free to remove the label.

Thanks for figuring this out Janice!

internal/pkg/aws/cloudwatch/cloudwatch_test.go
huanjani: I added a "return if no alarms found" test case. Did you mean something else/more?

huanjani: Oh wait, haha, I didn't add it. I just changed it so that if `GetResourcesByTags` doesn't return anything, `DescribeAlarms` isn't called.

iamhopaul123 (approved):
:tada:

PR #4490, opened 2023-02-14
chore(cli): add deployment type and alarm names to `svc show`

huanjani:
without ABR:
Screen Shot 2023-02-13 at 5 08 55 PM

with ABR:
Screen Shot 2023-02-13 at 5 09 18 PM


internal/pkg/describe/service.go
```diff
@@ -296,6 +300,32 @@ func (d *ecsServiceDescriber) ServiceConnectDNSNames() ([]string, error) {
...
+	}
+	if service.DeploymentConfiguration.Alarms == nil {
+		return nil, nil
+	}
+	return aws.StringValueSlice(service.DeploymentConfiguration.Alarms.AlarmNames), nil
+}
+
+// DeploymentType returns 'Alarm-Based Rollback' or 'Default'.
+func (d *ecsServiceDescriber) DeploymentType() (string, error) {
+	deploymentType := "Default"
```
huanjani: I'm not wed to this term. How do people feel about "Rolling"? My one concern is that folks may not realize that is the default deployment type.

internal/pkg/describe/backend_service.go
```diff
@@ -146,20 +147,30 @@ func (d *BackendServiceDescriber) Describe() (HumanJSONStringer, error) {
 		if err != nil {
 			return nil, fmt.Errorf("retrieve platform: %w", err)
 		}
-		backendSvcEnvVars, err := svcDescr.EnvVars()
+		deploymentType, err := svcDescr.DeploymentType()
 		if err != nil {
-			return nil, fmt.Errorf("retrieve environment variables: %w", err)
+			return nil, fmt.Errorf("retrieve rollback alarm names: %w", err)
```
iamhopaul123: Should it be "Get deployment type" instead?

internal/pkg/describe/service.go
```diff
@@ -296,6 +300,32 @@ func (d *ecsServiceDescriber) ServiceConnectDNSNames() ([]string, error) {
...
+	var err error
+	alarmNames, err := d.DeploymentConfigAlarmNames()
+	if err != nil {
+		return "", fmt.Errorf("retrieve alarm names: %w", err)
+	}
+	if alarmNames != nil {
```
iamhopaul123: It doesn't make a lot of sense to me that if the deployment alarm name is specified and the deployment type becomes "Alarm-Based Rollback" 🤔  "Default" is also ambiguous term if it's not defined by ECS. Maybe we don't need to make changes to this section?

internal/pkg/describe/backend_service_test.go
```diff
@@ -112,8 +119,35 @@ func TestBackendServiceDescriber_Describe(t *testing.T) {
 			},
 			wantedError: fmt.Errorf("retrieve platform: some error"),
 		},
+		"return error if fail to retrieve rollback alarm names": {
```
Lou1415926, line 122: Would we want one more unit test that tests "return error if fail to retrieve deployment type" ?

internal/pkg/describe/service.go
```diff
@@ -296,6 +300,32 @@ func (d *ecsServiceDescriber) ServiceConnectDNSNames() ([]string, error) {
 	return service.ServiceConnectAliases(), nil
 }
 
+// DeploymentConfigAlarmNames returns the rollback alarm names of a service.
+func (d *ecsServiceDescriber) DeploymentConfigAlarmNames() ([]string, error) {
```
Lou1415926: what about 
```suggestion
func (d *ecsServiceDescriber) RollbackAlarmNames() ([]string, error) {
```

internal/pkg/describe/service.go
Lou1415926: Is alarm based rollback exclusive with "Rolling"? 

From my understanding, alarm-based rollback is an addition on top of "Rolling":
1. All services in Copilot is [rolling](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/deployment-type-ecs.html)
2. We have "default" rolling and "recreate" rolling
3. Alarm-based rollback is not exclusive with any of the above. That is, we can have "default rolling + alarm-based rollback", "default rolling w/o alarm-based rollback", "recreate rolling + abr", "recreate rolling w/o abr"

If this is the case, I agree that returning either "Rolling" or "Alarm-Based Rollback" might be a little counterintuitive here.

efekarakus: Agreed to both of these feedbacks!
ABR can be enabled with various deployment _strategies_  such as "rolling", "blue/green", "canary".

If you'd like we can rename this method `IsAlarmBasedRollbackEnabled` and then in the `svc show` table we can display:
| Environment | Tasks | vCPU | Mem | Rollback on Alarm |
|-------------|-------|-------|-------|-------------------|
| test| 1 | 0.25 | 512 | ✔️ |

internal/pkg/describe/backend_service.go
```diff
@@ -237,6 +249,17 @@ func (w *backendSvcDesc) HumanString() string {
...
+		fmt.Fprint(writer, color.Bold.Sprint("\nRollback Alarms\n\n"))
+		writer.Flush()
+		headers := []string{"Name"}
+		fmt.Fprintf(writer, "  %s\n", strings.Join(headers, "\t"))
+		fmt.Fprintf(writer, "  %s\n", strings.Join(underline(headers), "\t"))
+		for _, alarm := range w.Alarms {
+			fmt.Fprintf(writer, "  %s\n", alarm)
+		}
+		fmt.Fprint(writer, "\n")
+	}
```
efekarakus: what do you think of creating a custom type and moving this over to a method?

```go
type rollbackAlarms []string
func (abr rollbackAlarms) humanString(w writeFlusher)
```
This way if we ever want to update how rollback alarms are displayed we won't have to remember to update across `backendSvcDesc` and `LBWebServiceDescriber`

internal/pkg/describe/service.go
```diff
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
```
efekarakus: nit: :O did the IDE put the import statement here 🤔 I thought it would have gotten lumped bellow with the rest of the github.com/aws/ imports

internal/pkg/describe/service.go
efekarakus: 👍 also does it need to be public?

internal/pkg/describe/service.go
efekarakus: In the long run I think we need a better visual representation for all these settings:
|Environment | test | prod|
|:-----------|:----|:-----|
| Count | 1 | Autoscaling 1-10|
| vCPU | 0.25 | 0.25 |
| Mem | 512 | 512 |
| Deployment Strategy | recreate | rolling with min 100% and max 200% |
| Rollback on alarm | ❌ | ✔️ |

internal/pkg/describe/worker_service.go
```diff
@@ -145,6 +156,7 @@ type workerSvcDesc struct {
 	Type           string               `json:"type"`
 	App            string               `json:"application"`
 	Configurations ecsConfigurations    `json:"configurations"`
+	Alarms         []string             `json:"rollbackAlarms,omitempty"`
```
efekarakus, line 154: do we not want to include the "Rollback Alarms" in `HumanString` for worker services?

huanjani: 🤦🏼‍♀️

internal/pkg/describe/service.go
huanjani: Okay, for now, we'll let the listed Rollback Alarms indicate that that's the deployment type. I'll remove the column.

huanjani: Screen Shot 2023-02-15 at 4 33 19 PM

internal/pkg/describe/backend_service.go
huanjani: Done. But used just Writer, not WriteFlusher to match the other `humanString` methods; we can swap them all together later.

internal/pkg/describe/service.go
huanjani: moved it!

internal/pkg/describe/service.go
huanjani: changed the name but left it as public for now, so we can reexamine all of the `svcDescr` methods together because they're so similar.

Lou1415926 (approved):
LGTM :shipit:

internal/pkg/describe/backend_service.go
```diff
@@ -237,6 +244,12 @@ func (w *backendSvcDesc) HumanString() string {
 	fmt.Fprint(writer, color.Bold.Sprint("\nConfigurations\n\n"))
 	writer.Flush()
 	w.Configurations.humanString(writer)
+	if len(w.Alarms) > 0 {
+		fmt.Fprint(writer, color.Bold.Sprint("\nRollback Alarms\n\n"))
+		writer.Flush()
+		alarms := rollbackAlarms(w.Alarms)
+		alarms.humanString(writer)
```
efekarakus:
```suggestion
		rollbackAlarms(w.Alarms).humanString(writer)
```

internal/pkg/describe/service.go
```diff
@@ -296,6 +299,17 @@ func (d *ecsServiceDescriber) ServiceConnectDNSNames() ([]string, error) {
 	return service.ServiceConnectAliases(), nil
 }
 
+func (d *ecsServiceDescriber) RollbackAlarmNames() ([]string, error) {
```
efekarakus, line 303: doc-comments 🥺

efekarakus (approved):
Approved with do-not-merge with 2 tiny refactors

internal/pkg/describe/service.go
huanjani: Ha! 🤦🏼 I deleted them when I started making it private. Will reinstate! Thanks for catching.

PR #4492, opened 2023-02-14
chore: Implementation of Sidecar build and push in Parallel

KollaAdithya:
This PR relates to #4254 
With this PR we build and push multiple sidecar images in parallel to AWS ECR.

But the UX of the Command line looks messy because of the go routines. Still working on it to make UX prettier.

### DIFF
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -357,7 +357,6 @@ func (d *workloadDeployer) uploadContainerImage(imgBuilderPusher imageBuilderPus
 	if !required {
 		return nil, scDigests, nil
 	}
-	log.Infoln(scDigests)
 	return aws.String(digest), scDigests, nil
 }
 
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -337,7 +337,7 @@ func (d *workloadDeployer) uploadContainerImage(imgBuilderPusher imageBuilderPus
 			return nil, nil, fmt.Errorf("build and push main container image: %w", err)
 		}
 	}
-	args := scbuildArgs(d.name, d.gitTag, UUIDTag, d.workspacePath, d.mft)
+	args := scBuildArgs(d.name, d.gitTag, UUIDTag, d.workspacePath, d.mft)
 	scdigests := make(map[string]string, len(args))
 	ctx, cancelWait := context.WithTimeout(context.Background(), waitForImageBuildAndPush)
 	defer cancelWait()
@@ -419,8 +419,8 @@ func userOrGitTag(userTag, gitTag string) string {
 	return gitTag
 }
 
-// scbuildArgs returns the map of dockerengine build arguments for the each sidecar that requires a native build from dockerfile.
-func scbuildArgs(name, gitTag, uuid, workspacePath string, unmarshaledManifest interface{}) map[string]*dockerengine.BuildArguments {
+// scBuildArgs returns the map of dockerengine build arguments for the each sidecar that requires a native build from dockerfile.
+func scBuildArgs(name, gitTag, uuid, workspacePath string, unmarshaledManifest interface{}) map[string]*dockerengine.BuildArguments {
 	mf, ok := unmarshaledManifest.(dfArgs)
 	var args map[string]*manifest.DockerBuildArgs
 	if ok {
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -357,6 +357,7 @@ func (d *workloadDeployer) uploadContainerImage(imgBuilderPusher imageBuilderPus
 	if !required {
 		return nil, scDigests, nil
 	}
+	log.Infoln(scDigests)
 	return aws.String(digest), scDigests, nil
 }
 
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -5,12 +5,14 @@ package deploy
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
@@ -39,12 +41,14 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/workspace"
 	"github.com/google/uuid"
 	"github.com/spf13/afero"
+	"golang.org/x/sync/errgroup"
 )
 
 const (
 	fmtForceUpdateSvcStart    = "Forcing an update for service %s from environment %s"
 	fmtForceUpdateSvcFailed   = "Failed to force an update for service %s from environment %s: %v.\n"
 	fmtForceUpdateSvcComplete = "Forced an update for service %s from environment %s.\n"
+	waitForImageBuildAndPush  = 100 * time.Second
 )
 
 // ActionRecommender contains methods that output action recommendation.
@@ -317,32 +321,43 @@ func (d *workloadDeployer) uploadContainerImage(imgBuilderPusher imageBuilderPus
 	if err != nil {
 		return nil, nil, err
 	}
-	if !required {
-		return nil, nil, nil
-	}
 	UUIDTag, err := generateUUID()
 	if err != nil {
 		return nil, nil, err
 	}
-	// If it is built from local Dockerfile, build and push to the ECR repo.
-	buildArg, err := buildArgs(d.name, d.imageTag, d.gitTag, UUIDTag, d.workspacePath, d.mft)
-	if err != nil {
-		return nil, nil, err
-	}
-	digest, err := imgBuilderPusher.BuildAndPush(dockerengine.New(exec.NewCmd()), buildArg)
-	if err != nil {
-		return nil, nil, fmt.Errorf("build and push main container image: %w", err)
+	var digest string
+	if required {
+		// If it is built from local Dockerfile, build and push to the ECR repo.
+		buildArg, err := buildArgs(d.name, d.imageTag, d.gitTag, UUIDTag, d.workspacePath, d.mft)
+		if err != nil {
+			return nil, nil, err
+		}
+		digest, err = imgBuilderPusher.BuildAndPush(dockerengine.New(exec.NewCmd()), buildArg)
+		if err != nil {
+			return nil, nil, fmt.Errorf("build and push main container image: %w", err)
+		}
 	}
 	args := scbuildArgs(d.name, d.gitTag, UUIDTag, d.workspacePath, d.mft)
 	if args == nil {
 		return aws.String(digest), nil, nil
 	}
 	scdigests := make(map[string]string, len(args))
+	ctx, cancelWait := context.WithTimeout(context.Background(), waitForImageBuildAndPush)
+	defer cancelWait()
+	g, _ := errgroup.WithContext(ctx)
 	for k, v := range args {
-		scdigests[k], err = imgBuilderPusher.BuildAndPush(dockerengine.New(exec.NewCmd()), v)
-		if err != nil {
-			return nil, nil, fmt.Errorf("build and push sidecar container images: %w", err)
-		}
+		scName := k
+		dArgs := v
+		g.Go(func() error {
+			scdigests[scName], err = imgBuilderPusher.BuildAndPush(dockerengine.New(exec.NewCmd()), dArgs)
+			if err != nil {
+				return err
+			}
+			return nil
+		})
+	}
+	if err := g.Wait(); err != nil {
+		return nil, nil, err
 	}
 	return aws.String(digest), scdigests, nil
 }
@@ -511,6 +526,7 @@ func (d *workloadDeployer) uploadArtifacts() (*UploadArtifactsOutput, error) {
 	if err != nil {
 		return nil, fmt.Errorf("upload custom resources for %q: %w", d.name, err)
 	}
+	log.Infoln(urls)
 	out.CustomResourceURLs = urls
 	return out, nil
 }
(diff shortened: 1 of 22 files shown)


internal/pkg/cli/deploy/workload.go
```diff
@@ -343,10 +389,76 @@ func buildArgs(name, imageTag, workspacePath string, unmarshaledManifest interfa
...
+// Otherwise, return git commit id.
+func userOrGitTag(userTag, gitTag string) string {
+	if userTag != "" {
+		return userTag
+	}
+	return gitTag
+}
+
+// scbuildArgs returns the map of dockerengine build arguments for the each sidecar that requires a native build from dockerfile.
+func scbuildArgs(name, gitTag, uuid, workspacePath string, unmarshaledManifest interface{}) map[string]*dockerengine.BuildArguments {
```
huanjani: nit: can this be `scBuildArgs`? 🥺

internal/pkg/cli/deploy/workload.go
```diff
@@ -37,13 +39,16 @@ import (
...
+	"github.com/google/uuid"
 	"github.com/spf13/afero"
+	"golang.org/x/sync/errgroup"
 )
 
 const (
 	fmtForceUpdateSvcStart    = "Forcing an update for service %s from environment %s"
 	fmtForceUpdateSvcFailed   = "Failed to force an update for service %s from environment %s: %v.\n"
 	fmtForceUpdateSvcComplete = "Forced an update for service %s from environment %s.\n"
+	waitForImageBuildAndPush  = 100 * time.Second
```
huanjani: I'm just curious how you decided on this amount of time. 🤷🏼

huanjani:
Just a couple random thoughts: I assume there isn't a situation in which build and push order matters and it's okay for them to default build and push in parallel.... 
Also, if someone has Container Dependencies with `depends_on`, that overrules/doesn't affect this, right?

internal/pkg/cli/deploy/workload.go
KollaAdithya: I initially thought given an X amount of time will be sufficient for building and pushing images. But the build process depends on the size of dockerfile, so removed the timeout set with `context`.

internal/pkg/cli/deploy/workload.go
```diff
@@ -302,27 +314,62 @@ func (d *workloadDeployer) forceDeploy(in *forceDeployInput) error {
...
+
+// generateUUID generates a random UUID to tag main and sidecar containers.
+func generateUUID() (string, error) {
+	id, err := uuid.NewRandom()
 	if err != nil {
-		return nil, fmt.Errorf("build and push image: %w", err)
+		return "", fmt.Errorf("generate random uuid to tag main and sidecar images: %w", err)
 	}
-	return aws.String(digest), nil
+	return id.String(), err
```
Lou1415926, line 369:
```suggestion
	return id.String(), nil
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -302,27 +314,62 @@ func (d *workloadDeployer) forceDeploy(in *forceDeployInput) error {
 	return nil
 }
 
-func (d *workloadDeployer) uploadContainerImage(imgBuilderPusher imageBuilderPusher) (*string, error) {
+func (d *workloadDeployer) uploadContainerImage(imgBuilderPusher imageBuilderPusher) (*string, map[string]string, error) {
```
Lou1415926, line 317: What about ⬇️  skeleton?
```go
// In manifest package
func DockerfileBuildRequired() (bool, error) {
   // Return true if >= 1 container (main + sidecars) needs docker build.
   // Return false otherwise.
}

// here
func ((d *workloadDeployer) uploadContainerImage(imgBuilderPusher imageBuilderPusher) (*string, map[string]string, error) {
  required, err := manifest.DockerBuildRequired(d.mft)
  if !required { return nil, nil }
  ...
}
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -87,12 +90,18 @@ type fileReader interface {
...
+	ContainerPlatform() string
+}
+
 // StackRuntimeConfiguration contains runtime configuration for a workload CloudFormation stack.
 type StackRuntimeConfiguration struct {
 	// Use *string for three states (see https://github.com/aws/copilot-cli/pull/3268#discussion_r806060230)
 	// This is mainly to keep the `workload package` behavior backward-compatible, otherwise our old pipeline buildspec would break,
 	// since previously we parsed the env region from a mock ECR URL that we generated from `workload package``.
 	ImageDigest        *string
+	ScImageDigests     map[string]string
```
efekarakus, line 104: nit:
```suggestion
	ImageDigestFor     map[string]string  // Mapping of container name to to image digest.
```

internal/pkg/cli/deploy/workload.go
efekarakus: Can we simplify and have have the following signature?
```suggestion
func (d *workloadDeployer) uploadContainerImage(imgBuilderPusher imageBuilderPusher) (map[string]string, error) {
```
So that the image digest for the main container is also included in the map?

internal/pkg/cli/deploy/workload.go
```diff
@@ -302,27 +314,62 @@ func (d *workloadDeployer) forceDeploy(in *forceDeployInput) error {
...
 	if err != nil {
-		return nil, err
-	}
-	if !required {
-		return nil, nil
+		return nil, nil, err
 	}
-	// If it is built from local Dockerfile, build and push to the ECR repo.
-	buildArg, err := buildArgs(d.name, d.imageTag, d.workspacePath, d.mft)
+	UUIDTag, err := generateUUID()
```
efekarakus, line 322:
```suggestion
	uuidTag, err := generateUUID()
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -302,27 +314,62 @@ func (d *workloadDeployer) forceDeploy(in *forceDeployInput) error {
...
+		return nil, nil, err
+	}
+	var digest string
+	if required {
+		// If it is built from local Dockerfile, build and push to the ECR repo.
+		buildArg, err := buildArgs(d.name, d.imageTag, d.gitTag, UUIDTag, d.workspacePath, d.mft)
+		if err != nil {
+			return nil, nil, err
+		}
+		digest, err = imgBuilderPusher.BuildAndPush(dockerengine.New(exec.NewCmd()), buildArg)
```
efekarakus, line 333: if I'm understanding correctly we're only building and pushing sidecar images concurrently?

This method is essentially split into two: doing the same work once for the main container and then again for sidecars within goroutines. I think this is a hint that there is duplication and we can simplify our flow 🤔 

Can we get to this type of end desired state:
```go
for name, imgCfg  := range d.mft.ImageConfigs() { // ⬅️ Includes main container and sidecar. This is what we need.
    if !imgCfg.RequiresBuild() { continue }
    // the Go routine stuff...
}
```

The goal is that there should be no separate codepaths for the main container vs. sidecar containers. They're all just containers.

internal/pkg/cli/deploy/workload.go
```diff
@@ -302,27 +314,62 @@ func (d *workloadDeployer) forceDeploy(in *forceDeployInput) error {
...
-	digest, err := imgBuilderPusher.BuildAndPush(dockerengine.New(exec.NewCmd()), buildArg)
+	args := scBuildArgs(d.name, d.gitTag, UUIDTag, d.workspacePath, d.mft)
+	scDigests := make(map[string]string, len(args))
+	ctx := context.Background()
+	g, _ := errgroup.WithContext(ctx)
+	for k, v := range args {
+		scName := k
+		dArgs := v
+		g.Go(func() error {
+			scDigests[scName], err = imgBuilderPusher.BuildAndPush(dockerengine.New(exec.NewCmd()), dArgs)
```
efekarakus, line 346: I think this map might not be safe for concurrent usage 💭 we'll have multiple go routines accessing and writing at the same time and I'm not sure if that behavior is ok.

Instead, consider using channels or a lock to safely write values

internal/pkg/cli/deploy/workload.go
```diff
@@ -302,27 +314,62 @@ func (d *workloadDeployer) forceDeploy(in *forceDeployInput) error {
...
+		}
 	}
-	digest, err := imgBuilderPusher.BuildAndPush(dockerengine.New(exec.NewCmd()), buildArg)
+	args := scBuildArgs(d.name, d.gitTag, UUIDTag, d.workspacePath, d.mft)
+	scDigests := make(map[string]string, len(args))
+	ctx := context.Background()
+	g, _ := errgroup.WithContext(ctx)
+	for k, v := range args {
+		scName := k
+		dArgs := v
```
efekarakus, line 344: Can we include a comment to https://go.dev/doc/effective_go#goroutines to explain why it's done this way:
```suggestion
	for name, imgCfg := range imageConfigs {
		name := name // See https://go.dev/doc/effective_go#goroutines 
		imgCfg := imgCfg
```

internal/pkg/manifest/backend_svc.go
```diff
@@ -124,6 +124,20 @@ func (s *BackendService) BuildArgs(wsRoot string) *DockerBuildArgs {
 	return s.ImageConfig.Image.BuildConfig(wsRoot)
 }
 
+// SidecarBuildArgs returns a map of docker build arguments for sidecar container images.
+func (s *BackendService) SidecarBuildArgs(wsRoot string) map[string]*DockerBuildArgs {
```
efekarakus, line 128: As mentioned above, consider a more generic method:
```suggestion
// ImageConfigs returns a map of container image configurations for all containers in the manifest.
func (s *BackendService) ImageConfigs(contextDir string) map[string]*DockerBuildArgs {
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -509,6 +620,11 @@ func (d *workloadDeployer) runtimeConfig(in *StackRuntimeConfiguration) (*stack.
 			Region:                   d.env.Region,
 			CustomResourcesURL:       in.CustomResourceURLs,
 			EnvVersion:               envVersion,
+			ScECRImage: stack.SidecarECRImage{
```
Lou1415926, line 623: This is just a super rough sketch so the details are totally not accurate. What do you think of ⬇️ ? The idea is that we don't need to have `ImageConfig` and `SidecarImageConfig` very where, because the main container is just another container, not so different than a sidecar. They should ideally be treated as equal.
```go
// In cli/deploy/workload.gos
func (d *workloadDeployer) uploadContainerImage(imgBuilderPusher imageBuilderPusher) (map[string]RuntimeImage, error) {
   // Return a map that looks like 
   // {svcName: imageDigestForMainContainer, sidecar1: imageDigestForSidecar1, sidecar1: imageDigestForSidecar1}
}

type RuntimeImage struct {
   Digest string
   CustomTag string
   GitShortCommit string
   UUID string
}

// ReferenceTag returns the tag that should be used to reference the image.
func (i RuntimeImage) ReferenceTag() string {
   if i.CustomTag != "" { return i.CustomTag }
   if i.GitShortCommit != "" { return i.GitShortCommit }
   return i.UUID
}

type UploadArtifactsOutput struct {
   ImageDigests     map[string]RuntimeImage
   EnvFileARN         string
   AddonsURL          string
   CustomResourceURLs map[string]string
}

func (d *workloadDeployer) uploadArtifacts() (*UploadArtifactsOutput, error) {
   imageDigests, err := d.uploadContainerImage(d.imageBuilderPusher)
   out := &UploadArtifactsOutput{
      ImageDigest:    imageDigests,
   }
}

type StackRuntimeConfiguration struct {
   ImageDigests map[string]RuntimeImage
}

func (d *workloadDeployer) runtimeConfig(in *StackRuntimeConfiguration) (*stack.RuntimeConfig, error) {
  images := make(map[string]stack.ECRImage)
  for container, img := range in.ImageDigests {
     images[container] = &stack.ECRImage{
      Repourl: d.resources.RepositoryURLs[d.name],
      ImageTag: img.ReferenceTag(),
      Digest: img.Digest,
     }
  }
  return &stack.RuntimeConfig{
   Images: images
  }
}

// In deploy/cloudformation/stack/workload_ecs.go
type RuntimeConfig struct {
   Images map[string]ECRImage // Optional. Image location in an ECR repository.
}

// In deploy/cloudformation/stack/transformer.go
func convertSidecars(s map[string]*manifest.SidecarConfig, exposedPorts map[string][]manifest.ExposedPort, rc RuntimeConfig) ([]*template.SidecarOpts, error) {
   for _, name := range keys {
         var imageURI string
         if uri, buildRequired := config.ImageURI(); buildRequired {
            imageURI = uri
         } else {
            imageURI = rc.Images[key].GetLocation()
         }
   }
}


```

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -133,10 +140,13 @@ func (c CmdClient) Build(in *BuildArguments) error {
...
-		[]string{"login", "-u", username, "--password-stdin", uri},
-		exec.Stdin(strings.NewReader(password)))
-
+	var err error
+	loginOnce.Do(
+		func() {
+			err = c.runner.Run("docker",
+				[]string{"login", "-u", username, "--password-stdin", uri},
+				exec.Stdin(strings.NewReader(password)))
+		})
```
efekarakus, line 149: Instead of a global `sync.Once` to me this is hinting that we need a different constructor that enables parallel build and push for example:

```go
func NewWithLogin(exec Cmd, registryURI string, username, passwd string) (*CmdClient, error) {
   client := New(exec)
   if err := client.Login(registryURI, username, passwd); err != nil { return nil, err }
   return client, nil
}
```

And then all the clients can safely use this client. Would that work? 💭 


Or alternatively, we return a specific error here that catches "the specified item already exists" if the client receives that error they ignore it on purpose 💭

efekarakus (commented):
Let me know what you think of the comments below!

My impression is that the current PR is trying to do too much, and we need to send several refactor PRs before a parallel build and push of container images.

1. A PR to introduce `ImageConfigs` method.
2. A PR to change how we display image builds such that multiple builds can be displayed on the terminal.
3. A PR that first builds and push sequentially.
4. A PR that then refactors and builds and push concurrently.

PR #4500, opened 2023-02-15
chore: make proper suggestions for environment storage in `storage init`

Lou1415926:


internal/pkg/cli/cli.go
```diff
@@ -128,6 +128,15 @@ func indentListItem(multiline string) string {
 	return strings.Join(prefixedLines, "\n")
 }
 
+func indentBy(multiline string, indentCount int) string {
```
iamhopaul123, line 131:
```suggestion
func indentedBy(multiline string, indentCount int) string {
```

internal/pkg/workspace/workspace.go
```diff
@@ -188,6 +188,16 @@ func (ws *Workspace) WorkloadExists(name string) (bool, error) {
 	return exists, nil
 }
 
+// EnvironmentsExist returns true if any environment is managed in the workspace.
+func (ws *Workspace) EnvironmentsExist() (bool, error) {
```
iamhopaul123: Maybe `AnyEnvironmentExist` or `HasEnvironment` by reading the comment?

internal/pkg/cli/storage_init.go
```diff
@@ -517,12 +522,24 @@ func (o *initStorageOpts) validateStorageLifecycle() error {
 // validateWorkloadNameWithLifecycle requires the workload to be in the workspace if the storage lifecycle is on workload-level.
 // Otherwise, it only caches whether the workload is present.
 func (o *initStorageOpts) validateWorkloadNameWithLifecycle() error {
+	if o.lifecycle == lifecycleEnvironmentLevel {
+		environmentsExist, err := o.ws.EnvironmentsExist()
+		if err != nil {
+			return fmt.Errorf("check if environments are managed in the workspace: %w", err)
+		}
+		if !environmentsExist {
+			return &errEnvironmentsNotInWorkspace{}
```
iamhopaul123:
```suggestion
			return &errNoEnvironmentInWorkspace{}
```

internal/pkg/cli/storage_init.go
```diff
@@ -1072,13 +1112,89 @@ For example, in JavaScript you can write:
...
+		return []string{envDeployAction, addIngressAction}
+	}
+	if onlyWkld := o.addIngressFrom != ""; onlyWkld {
+		return []string{svcMftAction, svcDeployAction}
+	}
 	return nil
 }
 
+func (o *initStorageOpts) manifestSuggestion() string {
+	if o.storageType == s3StorageType {
```
iamhopaul123: Maybe using switch here?

iamhopaul123 (approved):

internal/pkg/cli/cli.go
Lou1415926: I meant for `indent` to be a verb, as in `indent something by` 🤔 If this sounds good to you I will keep it as `indentBy`

internal/pkg/cli/storage_init.go
Lou1415926: I thought about it but `if o.storageType == manifestinfo.RDWS` would be nested in `case rdsStorageType`. Doesn't seem like it's helping too much.

Lou1415926: see if you like the current version better!

internal/pkg/workspace/workspace.go
Lou1415926: done!

Lou1415926:
An example:
Screenshot 2023-02-16 at 9 34 47 AM

internal/pkg/cli/storage_init.go
```diff
@@ -517,12 +522,24 @@ func (o *initStorageOpts) validateStorageLifecycle() error {
 // validateWorkloadNameWithLifecycle requires the workload to be in the workspace if the storage lifecycle is on workload-level.
 // Otherwise, it only caches whether the workload is present.
 func (o *initStorageOpts) validateWorkloadNameWithLifecycle() error {
+	if o.lifecycle == lifecycleEnvironmentLevel {
+		hasEnv, err := o.ws.HasEnvironment()
+		if err != nil {
+			return fmt.Errorf("check if environments are managed in the workspace: %w", err)
```
efekarakus: nit: I wonder if "manage" could create confusion for folks
```suggestion
			return fmt.Errorf("check if environment manifests exist in the workspace: %w", err)
```

internal/pkg/cli/storage_init.go
```diff
@@ -730,18 +747,23 @@ func (o *initStorageOpts) Execute() error {
...
+			log.Successf("Wrote CloudFormation %s at %s\n",
+				addon.description,
+				color.HighlightResource(displayPath(path)),
+			)
+			continue
 		}
-		path = displayPath(path)
-		log.Successf("Wrote CloudFormation %s at %s\n",
+		_, ok := err.(*workspace.ErrFileExists)
+		if !ok {
```
efekarakus: nit: I think it might be safer for us to use the `errors.As` function in case the workspace methods ever wrap this error.

```go
var ferr *workspace.ErrFileExists
if !errors.As(err, ferr) { return err }
```

internal/pkg/cli/storage_init.go
```diff
@@ -1072,13 +1112,93 @@ For example, in JavaScript you can write:
...
+    from_cfn: ${COPILOT_APPLICATION_NAME}-${COPILOT_ENVIRONMENT_NAME}-%sAuroraSecret`,
+			logicalIDSafeStorageName, logicalIDSafeStorageName)
+
+	}
+	return ""
+}
+
+func (o *initStorageOpts) addIngressSuggestion() string {
+	return fmt.Sprintf(`copilot storage init -n %s \
+--storage-type %s \
```
efekarakus, line 1176: Just to double check my understnading, we are not writing `--lifecycle environment` here on purpose because this addon will be a workload addon that grants access to the workload ecs tasks to the environment level s3 bucket right?

internal/pkg/workspace/workspace.go
```diff
@@ -188,6 +188,16 @@ func (ws *Workspace) WorkloadExists(name string) (bool, error) {
 	return exists, nil
 }
 
+// HasEnvironment returns true if the workspace manages environments.
+func (ws *Workspace) HasEnvironment() (bool, error) {
```
efekarakus: nit:
```suggestion
// HasEnvironments returns true if the workspace manages environments.
func (ws *Workspace) HasEnvironments() (bool, error) {
```

internal/pkg/workspace/workspace_test.go
```diff
@@ -344,6 +344,36 @@ func TestWorkspace_WorkloadExists(t *testing.T) {
...
+		}
+		got, err := ws.HasEnvironment()
+		require.NoError(t, err)
+		require.True(t, got)
+	})
+
+	t.Run("returns false if environments are not managed in the workspace", func(t *testing.T) {
+		fs := afero.NewMemMapFs()
+		_, _ = fs.Create("a/copilot/environments")
+		ws := &Workspace{
```
efekarakus: Check out the `TestWorkspace_EnvOverridesPath`, I wonder if we can get to a situation where our unit tests don't rely on setting _any_ private internal fields. I think that makes our unit tests more fragile because whenever we change the internal representation of a type we have to also update our unit tests.

Ideally, we'd be able to write a test like this:
```go
ws, err := Use(fakeFS)
require.NoError(t, err)
ok, err := ws.HasEnvironment()
require.NoError(t, err)
require.True(ok)
```

efekarakus (commented):
Looks good! Mostly a recommendation that I want to share around unit testing whenever we can, sometimes it's too difficult but I think in this situation it's doable.

internal/pkg/cli/storage_init.go
Lou1415926: > this addon will be a workload addon that grants access to the workload ecs tasks to the environment level s3

The addon produced by `--add-ingress-from` is indeed always a workload-level addon that grants the workload access to the environment storage (s3/ddb/rds)

However, the reasons why we are not writing `--lifecycle` are not only that. `--lifecycle` specifies the storage's lifecycle, not any addon's lifecycle. Since `--add-ingress-from` is not a storage addon, but an ingress addon, `--lifecycle` does not apply. 

> Since `--lifecycle` is useless when using `--add-ingress-from`, then why we don't error out if `--add-ingress-from` and `--lifecycle` are specified together? 

We totally could. In fact, we error out for `copilot storage init --add-ingress-from xx --lifecycle workload` - Copilot does not support adding additional ingress to a workload storage because of best practice. 

The only "compatible" value for `--lifecycle` is "environment" in this case, because `--add-ingress-from` is used to add an ingress to an "environment-level storage".

internal/pkg/workspace/workspace_test.go
Lou1415926: This concerns make sense to me!

My only concern with using `Use`/`Create` in unit test is that it assumes `Use`/`Create` is correctly implemented. The unit test here is then more of an "integration test" that tests whether an exported function works nicely with `Use`  / `Create`. 

I guess this can be thought of a nice perk 🤔

Lou1415926: done!

efekarakus (approved):

internal/pkg/workspace/workspace_test.go
efekarakus: That's true! It becomes closer to an "integ" test rather than an isolated unit-test 💭 

If it's any consolation the stdlib seems to do the same thing: https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/crypto/sha256/sha256_test.go;l=107

But I agree that it'd be extra nice if only  `Test_Use` were to fail instead of all these other ones in case a change to `Use` breaks a particular expected behavior.

PR #4501, opened 2023-02-15
refactor: convert ALB listener and target group to a list

paragbhingre:
This PR refactors ALB config to have list of alb listeners and target groups for both LBWS as well as Backend Service.


internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -174,33 +174,17 @@ func (s *LoadBalancedWebService) Template() (string, error) {
...
-		}
-	}
-
-	aliasesFor, err := convertHostedZone(s.manifest.RoutingRule.RoutingRuleConfiguration)
-	if err != nil {
-		return "", err
-	}
 	var deregistrationDelay *int64 = aws.Int64(60)
 	if s.manifest.RoutingRule.DeregistrationDelay != nil {
 		deregistrationDelay = aws.Int64(int64(s.manifest.RoutingRule.DeregistrationDelay.Seconds()))
```
iamhopaul123, line 176: Do we need to move this logic to `convertApplicationLoadBalancer` too?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -395,6 +416,10 @@ type networkLoadBalancerConfig struct {
 	appDNSName           *string
 }
 
+type applicationLoadBalancerConfig struct {
+	settings *template.ApplicationLoadBalancer
+}
```
iamhopaul123: Why do we need this instead of using `*template.ApplicationLoadBalancer` directly?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -141,16 +141,21 @@ func convertContainerHealthCheck(hc manifest.ContainerHealthCheck) *template.Con
...
-func convertHostedZone(m manifest.RoutingRuleConfiguration) (template.AliasesForHostedZone, error) {
+func convertHostedZone(alias manifest.Alias, defaultHostedZone *string) (template.AliasesForHostedZone, error) {
 	aliasesFor := make(map[string][]string)
-	defaultHostedZone := m.HostedZone
-	if len(m.Alias.AdvancedAliases) != 0 {
-		for _, alias := range m.Alias.AdvancedAliases {
+	if len(alias.AdvancedAliases) != 0 {
+		for _, alias := range alias.AdvancedAliases {
 			if alias.HostedZone != nil {
+				if isDuplicateAliasEntry(aliasesFor[*alias.HostedZone], aws.StringValue(alias.Alias)) {
```
iamhopaul123, line 152: Why do we need this 🤔

internal/pkg/template/templates/workloads/partials/cf/alb.yml
```diff
@@ -1,50 +1,51 @@
...
     {{- end }}
-    {{- if .HTTPVersion}}
-    ProtocolVersion: {{.HTTPVersion}}
+    {{- if $listener.HTTPVersion}}
+    ProtocolVersion: {{$listener.HTTPVersion}}
     {{- end}}
     TargetGroupAttributes:
       - Key: deregistration_delay.timeout_seconds
-        Value: {{.DeregistrationDelay}} # ECS Default is 300; Copilot default is 60.
+        Value: {{$.DeregistrationDelay}} # ECS Default is 300; Copilot default is 60.
```
iamhopaul123, line 41: nice!

internal/pkg/template/workload.go
```diff
@@ -802,13 +849,13 @@ type WorkloadOpts struct {
 // or an empty string if it shouldn't be configured, defaulting to the
 // target protocol. (which is what happens, even if it isn't documented as such :))
 // https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-targetgroup.html#cfn-elasticloadbalancingv2-targetgroup-healthcheckprotocol
-func (w WorkloadOpts) HealthCheckProtocol() string {
+func (albl ApplicationLoadBalancerRoutineRule) HealthCheckProtocol() string {
```
iamhopaul123:
```suggestion
func (alb ApplicationLoadBalancerRoutineRule) HealthCheckProtocol() string {
```

internal/pkg/template/workload.go
```diff
@@ -770,6 +816,7 @@ type WorkloadOpts struct {
 	DeregistrationDelay     *int64
 	AllowedSourceIps        []string
 	NLB                     *NetworkLoadBalancer
+	ALB                     *ApplicationLoadBalancer
```
iamhopaul123: Do we need to remove `HTTPTargetContainer` etc. from the top layer?

internal/pkg/template/workload.go
```diff
@@ -290,14 +284,39 @@ func (cfg *NetworkLoadBalancer) Aliases() []string {
 	for _, listener := range cfg.Listener {
 		for _, entry := range listener.Aliases {
 			if _, value := seen[entry]; !value {
-				seen[entry] = true
-				uniqueAliases = append(uniqueAliases, entry)
+				uniqueAliases = append(uniqueAliases, uniqueEntriesFromList(listener.Aliases, seen)...)
```
iamhopaul123: Maybe I misunderstood something but I am not following why we are doing this. It seems like we are not appending `entry` which is a unique alias. Instead, we do `append(uniqueAliases, uniqueEntriesFromList(listener.Aliases, seen)...)` multiple times whenever we see any alias that is not in `seen` map.

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
paragbhingre: The reason I did not move this logic inside `convertApplicationLoadBalancer` because we use this value for NLB as well, which is a bug because if we have config such as `http: false` and if customer wants to modify `deregistrationDelay` for NLB, then there is no way. This bug can be fixed separately, but time being I have kept this as is.

internal/pkg/template/workload.go
paragbhingre: Great callout @iamhopaul123 ❤️ 
I will remove parameters those are now part of ALB(ApplicationLoadBalancer).  Specifically `HTTPTargetContainer` we will still need because we use this config in service connect and we are touching service connect as a part of this project. So keeping it as is.

internal/pkg/template/workload.go
paragbhingre: So there are couple of things happening:
1. In the previous PR, we decided to remove `uniqueEntriesFromList()` method and add that logic inside `Aliases()` method itself. But now for ALB also we need the similar kind of logic so to remove duplication of code I removed common code out in the `uniqueEntriesFromList()` method which nothing but just gives the non-duplicate entries by checking the map that we fill as we get aliases from different listener rules. 
2. Consider following example with multiple additional rules
```
Listener: []ApplicationLoadBalancerRoutineRule{
					{
						Aliases: []string{
							"testAlias1",
							"testAlias2",
						},
					},
					{
						Aliases: []string{
							"testAlias1",
							"testAlias3",
						},
					},
				},
```
From the example you can see that listener1 has aliases as "testAlias1" and "testAlias2" whereas listener2 has "testAlias1" and "testAlias3". So in order to create certificate for all the unique aliases across all the listeners we need to write logic that traverses through all the listeners which is happening in `Aliases()` method and we are figure our all the aliases that are unique across all the listeners which happens in `uniqueEntriesFromList()`. 
Ultimately we are doing the same thing as you mentioned here  --  `It seems like we are not appending entry which is a unique alias`.  But we are doing this inside `uniqueEntriesFromList()` to reduce the duplication of code.
Does that make sense?
I am also adding a test for `Aliases()` which will help us understand the scenario.

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: let's consider an example manifest. 

```
name: lbservice

image:
  port: 80

http:
  path: '/'
  target_port: 82
  alias:
    - name: example.com
      hosted_zone: mockHostedZone2
    - name: foobar.com
    - name: "*.foobar.com"
  hosted_zone: mockHostedZone1
  additional_rules:
    - path: "/admin"
       target_port: 81   
       alias:
         - name: adminExample.com
            hosted_zone: mockHostedZone3
          - name: foobar.com
       hosted_zone: mockHostedZone1
```
As shown in the preceding example, the alias "foobar.com" will be added to "mockHostedZone1" in both the primary and additional rules. Because customers can specify the same alias but different paths to reach different container ports, So in order to make sure that `mockHostedZone1` doesn't have a duplicate entry of `foobar.com` we need to make sure it is not already in the []string of `mockHostedZone1`

iamhopaul123: That makes sense but i am not convinced it is the most efficient way of doing it especially we need to loop through to see if the alias is already in the list. Maybe consider using a map?

internal/pkg/template/workload.go
iamhopaul123: I am still not quite following. I feel like the original version is doing exactly the work and will return `[]string{testAlias1, testAlias2, testAlias3}` 🤔 Your modification will pass the unit tests but seems redundant or can you give me an example that this function previously won't work?

paragbhingre: `uniqueEntriesFromList(listener.Aliases, seen)...)` is just added to remove the duplication of code from `Aliases()` method that is there for BackendService and LoadBalancedService. If you want I can keep the code as is, but there will code duplication for both LBWS and BS.

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: I just tried keeping it consistent with NLB struct type. Its like we can keep some high level i.e. app level configs here like NLB does. NLB has appDNSName in here in it's own setting. May be in the future we will also need this place to add some high level resources.

internal/pkg/template/workload.go
```diff
@@ -456,6 +491,17 @@ type NetworkLoadBalancer struct {
...
 }
 
+// ApplicationLoadBalancer holds configuration that's needed for an Application Load Balancer.
+type ApplicationLoadBalancer struct {
+	PublicSubnetCIDRs []string
+	Listener          []ApplicationLoadBalancerRoutineRule
+	HostedZoneAliases AliasesForHostedZone
+	HTTPRedirect      bool
+	HTTPSListener     bool
+	ALBEnabled        bool
```
efekarakus: Why do we need this boolean? would the client set `ALB` to `nil` if it's not enabled?

internal/pkg/template/workload.go
```diff
@@ -456,6 +491,17 @@ type NetworkLoadBalancer struct {
 	CertificateRequired bool
 }
 
+// ApplicationLoadBalancer holds configuration that's needed for an Application Load Balancer.
+type ApplicationLoadBalancer struct {
+	PublicSubnetCIDRs []string
```
efekarakus: Is this field used anywhere? Can we delete it

internal/pkg/template/workload.go
```diff
@@ -456,6 +491,17 @@ type NetworkLoadBalancer struct {
 	CertificateRequired bool
 }
 
+// ApplicationLoadBalancer holds configuration that's needed for an Application Load Balancer.
+type ApplicationLoadBalancer struct {
```
efekarakus: My impression from reading the templates that this corresponds more to:
```go
type ALBListener struct // or alternatively "HTTPListener"
```

The struct contains fields that are relevant to all the listener rules under a listener. ALB level config such as subnets and security groups are not relevant going to be in this struct.

internal/pkg/template/workload.go
```diff
@@ -456,6 +491,17 @@ type NetworkLoadBalancer struct {
...
+	PublicSubnetCIDRs []string
+	Listener          []ApplicationLoadBalancerRoutineRule
+	HostedZoneAliases AliasesForHostedZone
+	HTTPRedirect      bool
+	HTTPSListener     bool
```
efekarakus: nit: consider renaming to
```suggestion
	RedirectToHTTPS      bool // Only relevant if IsHTTPS is true.
	IsHTTPS     bool  // True if the listener listening on port 443.
```
so that it reads in the template `{{$listener.IsHTTPS}}` or `{{$listener.RedirectToHTTPS}}`

internal/pkg/template/workload.go
```diff
@@ -456,6 +491,17 @@ type NetworkLoadBalancer struct {
 	CertificateRequired bool
 }
 
+// ApplicationLoadBalancer holds configuration that's needed for an Application Load Balancer.
+type ApplicationLoadBalancer struct {
+	PublicSubnetCIDRs []string
+	Listener          []ApplicationLoadBalancerRoutineRule
```
efekarakus: I believe these are more equivalent to listener rules:
```suggestion
	Rules          []ALBListenerRule
```

internal/pkg/template/workload.go
```diff
@@ -422,6 +441,22 @@ func SecretFromSecretsManager(value string) secretsManagerName {
 	}
 }
 
+// ApplicationLoadBalancerRoutineRule holds configuration that's needed for an Application Load Balancer listener.
+type ApplicationLoadBalancerRoutineRule struct {
+	// The path and protocol that the Application Load Balancer listens to.
+	Path     string
+	Protocol string
```
efekarakus: do we need this field? I don't see it used in the template

internal/pkg/template/workload.go
```diff
@@ -290,14 +284,39 @@ func (cfg *NetworkLoadBalancer) Aliases() []string {
...
+		for _, entry := range listener.Aliases {
+			if _, value := seen[entry]; !value {
+				uniqueAliases = append(uniqueAliases, uniqueEntriesFromList(listener.Aliases, seen)...)
 			}
 		}
 	}
 	return uniqueAliases
 }
 
+func uniqueEntriesFromList(aliases []string, uniqueMap map[string]bool) []string {
```
efekarakus:
```suggestion
func uniqueEntriesFromList(aliases []string, seen map[string]bool) []string {
```

efekarakus (commented):
I haven't finished reviewing, but just publishing early my existing feedback.

internal/pkg/template/workload.go
paragbhingre: I think I was trying to make this parameter as a part of `ALB` but later realized it is used on much higher context such as in `envControllerParameters` and so on. So that is why decided to keep it high level param inside `template.WorkloadOpts` and forgot to remove from this place. I will remove it.

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
efekarakus: Ack, sounds good! But I agree that this is an important bug that we must fix, thanks for catching it 🙏

internal/pkg/template/workload.go
```diff
@@ -802,13 +840,13 @@ type WorkloadOpts struct {
 // or an empty string if it shouldn't be configured, defaulting to the
 // target protocol. (which is what happens, even if it isn't documented as such :))
 // https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-targetgroup.html#cfn-elasticloadbalancingv2-targetgroup-healthcheckprotocol
-func (w WorkloadOpts) HealthCheckProtocol() string {
+func (alb ALBListenerRule) HealthCheckProtocol() string {
```
```suggestion
func (lr ALBListenerRule) HealthCheckProtocol() string {
```

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
```diff
@@ -123,9 +125,11 @@ Resources:
       {{- end }}
       LoadBalancers:
   {{- if .ALBEnabled}}
```
efekarakus: Do we still need an `ALBEnabled` field? or can we do `{{- if .ALB }}`

(samesies in `backend/cf.yml`)

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
```diff
@@ -45,43 +45,44 @@ LoadBalancerDNSAlias{{$hostedZoneID}}:
 {{- end}}
 {{- end}}
 
-HTTPSRulePriorityAction:
+{{- range $i, $listener := .ALB.Rules}}
+HTTPSRulePriorityAction{{ if ne $i 0 }}{{ $i }}{{ end }}:
```
efekarakus: OOHHH I wonder if this is going to be problematic to create in parallel!
https://github.com/aws/copilot-cli/blob/3c212fb45cf46ea497f4bb31955ac1c2e183625a/cf-custom-resources/lib/alb-rule-priority-generator.js#L77-L104

@iamhopaul123 I'd be curious to hear your input too.

My guess is that if we create all these actions in parallel then they will return the same rule priority and cause a stack failure on deployment.

Possible mitigations that I can think of are:

1. To add a `DependsOn` for each priority action, something like:
   ```yaml
   HTTPSRulePriorityAction{{$i}}:
     DependsOn:
        - HTTPSRulePriorityAction{{$i - 1}} # Force rule priority actions to be created sequentially.
   ```
   But that's costly for customers and I don't know if I trust that 
   
2. We should consider updating the `RulePriorityFunction` custom resource to return a map instead of a single value, so that we can do something like:
   ```yaml
   Priority: !GetAtt RulePriorityAction.Rule1Priority
   ---
   Priority: !GetAtt RulePriorityAction.Rule2Priority
   ```

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
```diff
@@ -45,43 +45,44 @@ LoadBalancerDNSAlias{{$hostedZoneID}}:
...
-    {{- if eq .WorkloadType "Backend Service"}}
+    RulePath: {{$listener.Path}}
+    {{- if eq $.WorkloadType "Backend Service" }}
     ListenerArn: !GetAtt EnvControllerAction.InternalHTTPSListenerArn
     {{- else}}
     ListenerArn: !GetAtt EnvControllerAction.HTTPSListenerArn
     {{- end}}
 
-HTTPRuleWithDomainPriorityAction:
+HTTPRuleWithDomainPriorityAction{{ if ne $i 0 }}{{ $i }}{{ end }}:
```
efekarakus: (same concern for all `RulePriorityFunction` actions)

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
```diff
@@ -120,41 +121,40 @@ HTTPListenerRuleWithDomain:
       - Field: 'path-pattern'
         PathPatternConfig:
           Values:
-            !If
-              - IsDefaultRootPath
```
efekarakus, line 124: can we remove this condition from the template now?

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
efekarakus: Let's discuss offline, but my recommendation is to get this PR unblocked since no customer can leverage multi-rules yet.

We can start with approach (1) that should work but is costly and do a quick follow-up with a long-term proper fix.

internal/pkg/template/templates/workloads/partials/cf/alb.yml
```diff
@@ -1,50 +1,51 @@
...
+    {{- if $rule.HTTPVersion}}
+    ProtocolVersion: {{$rule.HTTPVersion}}
     {{- end}}
     TargetGroupAttributes:
       - Key: deregistration_delay.timeout_seconds
-        Value: {{.DeregistrationDelay}} # ECS Default is 300; Copilot default is 60.
+        Value: {{$.DeregistrationDelay}} # ECS Default is 300; Copilot default is 60.
       - Key: stickiness.enabled
-        Value: !Ref Stickiness
+        Value: {{$rule.Stickiness}}
```
efekarakus, line 43: does this render properly in CFN? because stickiness is a bool and not a string in the Go struct

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -460,6 +485,107 @@ func convertEnvSecurityGroupCfg(mft *manifest.Environment) (*template.SecurityGr
...
+			RedirectToHTTPS:   httpRedirect,
+			IsHTTPS:           s.httpsEnabled,
+			HostedZoneAliases: aliasesFor,
+		},
+	}
+
+	return config, nil
+}
+
+func (s *BackendService) convertApplicationLoadBalancer() (applicationLoadBalancerConfig, error) {
```
efekarakus: I don't quite follow the difference between `*LoadBalancedWebService` method and `*BackendService` they seem pretty similar, can you help me understand why we need to have two separate ones?

efekarakus: I wonder if a flow like this would work and reduce duplication:
```go
type loadBalancerTargeter interface {
   HTTPLoadBalancerTarget() (string, string, error)
   MainContainerPort string
}

type routingRuleConfigConverter struct {
   rule manifest.RoutingRuleConfiguration
   manifest loadBalancerTargeter
}   
func (conv routingRuleConfigConverter) convert() (albConfig, error)

type RoutingRuleConfigOrBoolConverter struct {
   rule manifest.RoutingRuleConfigOrBool
   manifest loadBalancerTargeter
}
func (conv routingRuleConfigConverter) convert() (albConfig, error) {
   if conf.rule.Disabled() { return albConfig{}, nil }
   return routingRuleConfigConverter{
        rule: conf.rule.RoutingRuleConfiguration,
        manifest: conv.manifest,
   }.convert()
}
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -395,6 +416,10 @@ type networkLoadBalancerConfig struct {
 	appDNSName           *string
 }
 
+type applicationLoadBalancerConfig struct {
+	settings *template.ALBListener
+}
```
efekarakus: I agree with @iamhopaul123's earlier feedback that we can follow YAGNI for now and get rid of this struct in favor of `*template.ALBListener`. We can introduce it in the future when we need it.

internal/pkg/template/templates/workloads/partials/cf/alb.yml
paragbhingre: yes it does work fine because [Stickiness](https://github.com/aws/copilot-cli/pull/4501/files#diff-53f2b158bad1d10014e86bdd2b5b6ccc88acb52a641ccfce4a65c9fad47517cdR454) is a string in the Go struct.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -460,6 +481,132 @@ func convertEnvSecurityGroupCfg(mft *manifest.Environment) (*template.SecurityGr
...
+	var rules []template.ALBListenerRule
+	// build listener rule config from primary rule config from manifest.
+	rule, err := RoutingRuleConfigOrBoolConverter{
+		rule:         albConfig,
+		manifest:     s.manifest,
+		httpsEnabled: s.httpsEnabled,
+	}.convert()
+	if err != nil {
+		return nil, nil
+	}
```
iamhopaul123: Should we return the error?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -460,6 +481,132 @@ func convertEnvSecurityGroupCfg(mft *manifest.Environment) (*template.SecurityGr
...
+	aliasesFor, err := convertHostedZone(albConfig.Alias, albConfig.HostedZone)
+	if err != nil {
+		return nil, fmt.Errorf(`convert "http.alias" to string slice: %w`, err)
```
iamhopaul123: we probably don't need to wrap this error

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -460,6 +481,132 @@ func convertEnvSecurityGroupCfg(mft *manifest.Environment) (*template.SecurityGr
...
+	}
+	rules = append(rules, *rule)
```
iamhopaul123, line 506: Maybe do this right after we get `rule`?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -460,6 +481,132 @@ func convertEnvSecurityGroupCfg(mft *manifest.Environment) (*template.SecurityGr
...
+}
+
+type routingRuleConfigConverter struct {
+	rule         manifest.RoutingRuleConfiguration
+	manifest     loadBalancerTargeter
+	httpsEnabled bool
+}
+
+// RoutingRuleConfigOrBoolConverter holds routing rule config for LBWS.
+type RoutingRuleConfigOrBoolConverter struct {
```
iamhopaul123: Why do we even need this...the only convert method is just calling `routingRuleConfigConverter.convert()`

internal/pkg/template/workload.go
iamhopaul123:
```golang
uniqueAliases = append(uniqueAliases, entry)
seen[entry] = true
```
does this not work?

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: In case if error we will be appending nil value of the `rule` to `rules`. I don't think that operation is even required in case of error.

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: So `routingRuleConfigConverter.convert()` this method needs one of the parameters as `RoutingRuleConfiguration` and for LBWS we have `RoutingRuleConfigOrBool`, so we need a way to just send `RoutingRuleConfiguration` to `routingRuleConfigConverter.convert()` method which is our common code between BS and LBWS. does that make sense? To reduce duplicate code we have `routingRuleConfigConverter.convert()` and that accepts only `RoutingRuleConfiguration`.

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
paragbhingre: I think we can do `{{- if .ALB }}` but `ALBEnabled` has been used on a very broader scope that's why I kept it as is. It has been used in the template parameters as well.

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
paragbhingre: We are taking care of this issue in the fix [here](https://github.com/aws/copilot-cli/pull/4568).

internal/pkg/deploy/cloudformation/stack/transformers.go
iamhopaul123: `func (conv RoutingRuleConfigOrBoolConverter) convert() (*template.ALBListenerRule, error)` does not use anything unique about `RoutingRuleConfigOrBool`, which doesn't make sense to me. Why do we not use the `OrBool` part at all?

iamhopaul123: if we don't need `OrBool` logic I still think we can just use `routingRuleConfigConverter` because when we initialize the struct we can use `rule.RoutingRuleConfiguration` (rule is of type `RoutingRuleConfigOrBool`)

paragbhingre: I do agree to your point, I think we can just have `convert` method on `RoutingRuleConfiguration` and send rule as `rule: albConfig.RoutingRuleConfiguration` from LBWS because we anyway check for following condition 
```
if albConfig.Disabled() || albConfig.IsEmpty() {
	return nil, nil
}
```
Which indeed takes care of the `OrBool` part of the `RoutingRuleConfigOrBool`. So we are only left with `RoutingRuleConfiguration` part now and we can use `routingRuleConfigConverter` directly and it's `convert` method of rest of the work that we were doing through `convert` method of the `RoutingRuleConfigOrBoolConverter`

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
```diff
@@ -51,8 +51,8 @@ HTTPSRulePriorityAction:
   Type: Custom::RulePriorityFunction
   Properties:
     ServiceToken: !GetAtt RulePriorityFunction.Arn
-    RulePath: !Ref RulePathSlice
-    {{- if eq .WorkloadType "Backend Service"}}
+    RulePath: {{ fmtSlice (quoteSlice .ALB.RulePaths) }}
```
Lou1415926: Is the CFN parameter `RulePath` still used at all?

internal/pkg/describe/uri.go
```diff
@@ -76,7 +76,7 @@ func (d *LBWebServiceDescriber) URI(envName string) (URI, error) {
 		return URI{}, fmt.Errorf("get stack resources for service %s: %w", d.svc, err)
 	}
 	for _, resource := range resources {
-		if resource.LogicalID == svcStackResourceALBTargetGroupLogicalID {
+		if strings.HasPrefix(resource.LogicalID, svcStackResourceALBTargetGroupLogicalID) {
```
iamhopaul123, line 79: Just a reminder that we might need to modify this file for multiple ports  to show multiple urls.

internal/pkg/template/templates/workloads/partials/cf/alb.yml
```diff
@@ -89,7 +90,7 @@ RulePriorityFunctionRole:
                 - elasticloadbalancing:DescribeRules
               Resource: "*"
 
-{{- if .HTTPSListener}}
+{{- if .ALB.IsHTTPS}}
```
iamhopaul123: Are we sure `ALB` is not nil here? We usually only define concrete struct for CFN template to avoid this problem since there's no linter available to know if it is pointer value or concrete value. I would suggest we only use concrete ALB here and use other variable (like `ALB.Enabled` to determine if ALB is enabled or not.

internal/pkg/describe/uri.go
paragbhingre: So I had this conversation with Efe and Wanxian in the past that if we should be showing all the urls here or just the main url and all three of us had the same choice of showing only main url reason being it looks cluttered if we show all the urls.

internal/pkg/template/templates/workloads/partials/cf/alb.yml
paragbhingre: So before even calling `alb` cfn template we check for if `.ALB` is not `nil` [here](https://github.com/aws/copilot-cli/pull/4501/files#diff-89efe4a9579f819c8940b629c45537331bac2a4350e29eb00a1b6fac7b076855R141).

internal/pkg/template/workload.go
```diff
@@ -275,6 +264,11 @@ type importableValue interface {
 	Value() string
 }
 
+// IsHTTPS returns true if the target container's port is 443.
+func (lr ALBListenerRule) IsHTTPS() bool {
```
Lou1415926: This function signature says "whether the listener rule is an HTTPS rule", which does not seem to be accurate. What do you think of either of the following:
```go
type TargetPort string
func (p TargetPort) IsHTTPS() { return string(p) == "443" }
```
or
```go
	switch {
	case lr.HTTPHealthCheck.Port == "443":
		return "HTTPS"
	case lr.TargetPort == "443" && lr.HTTPHealthCheck.Port == "":
		return "HTTPS"
	case lr.TargetPort == "443" && lr.HTTPHealthCheck.Port != "443":
```

internal/pkg/template/workload.go
```diff
@@ -288,17 +282,42 @@ func ImportedVariable(name string) Variable {
 func (cfg *NetworkLoadBalancer) Aliases() []string {
 	var uniqueAliases []string
 	seen := make(map[string]bool)
-	for _, listener := range cfg.Listener {
-		for _, entry := range listener.Aliases {
+	for _, rule := range cfg.Listener {
```
Lou1415926: NLB's `cfg.Listener` contains a list of listeners, not a list of rules. NLB's listener doesn't have rules. Can we change the variable name back to `listener`?

internal/pkg/template/workload_test.go
```diff
@@ -314,40 +314,32 @@ func TestSecretsManagerName_ValueFrom(t *testing.T) {
 
 func TestWorkload_HealthCheckProtocol(t *testing.T) {
 	testCases := map[string]struct {
-		opts     WorkloadOpts
+		opts     ALBListenerRule
```
Lou1415926, line 311: Can we rename `func TestWorkload_HealthCheckProtocol(t *testing.T)` to `func TestALBListenerRule_HealthCheckProtocol(t *testing.T)`

internal/pkg/template/workload_test.go
```diff
@@ -479,3 +471,36 @@ func TestRollingUpdateRollbackConfig_TruncateAlarmName(t *testing.T) {
 		})
 	}
 }
+
+func TestApplicationLoadBalancer_Aliases(t *testing.T) {
```
Lou1415926, line 469: Since we now also have the `Aliases` method for NLB, do we need a unit test suite for it as well?

internal/pkg/template/workload_test.go
```diff
@@ -479,3 +471,36 @@ func TestRollingUpdateRollbackConfig_TruncateAlarmName(t *testing.T) {
...
+	tests := map[string]struct {
+		opts     ALBListener
+		expected []string
+	}{
+		"LBWS with multiple listener rules having multiple aliases each": {
```
Lou1415926, line 474: Do we need a test case that says `"LBWS with no aliases"`

Lou1415926 (commented):
A few questions - haven't finished reviewing the whole PR (haven't read the `cloudformation/stack` part, but I will come back to it.

internal/pkg/template/templates/workloads/partials/cf/https-listener.yml
paragbhingre: I will clean that from the CFN. But we need that as a parameter as we use that param to build url in uri.go.

internal/pkg/template/workload_test.go
paragbhingre: Both the functions are same that's why I just kept tests for one. But happy to add for nlb as well.

internal/pkg/template/workload_test.go
paragbhingre: The real motto of this function to find out unique aliases across all the listener rules(ALB) or listeners(NLB) and return those aliases. That is why I didn't find a need for the test with no Aliases as it returns empty []string. But I have added that test for safety.

internal/pkg/template/templates/workloads/partials/cf/alb.yml
```diff
@@ -1,50 +1,51 @@
...
-    ProtocolVersion: {{.HTTPVersion}}
+    {{- if $rule.HTTPVersion}}
+    ProtocolVersion: {{$rule.HTTPVersion}}
     {{- end}}
     TargetGroupAttributes:
       - Key: deregistration_delay.timeout_seconds
-        Value: {{.DeregistrationDelay}} # ECS Default is 300; Copilot default is 60.
+        Value: {{$.DeregistrationDelay}} # ECS Default is 300; Copilot default is 60.
       - Key: stickiness.enabled
-        Value: !Ref Stickiness
```
iamhopaul123, line 42: Do we still need `Stickiness` the parameter?

paragbhingre: removed.

paragbhingre: removed.

internal/pkg/template/workload.go
```diff
@@ -291,14 +280,39 @@ func (cfg *NetworkLoadBalancer) Aliases() []string {
 	for _, listener := range cfg.Listener {
 		for _, entry := range listener.Aliases {
 			if _, value := seen[entry]; !value {
```
Lou1415926:
```suggestion
			if _, ok := seen[entry]; !ok {
```

internal/pkg/template/workload.go
```diff
@@ -291,14 +280,39 @@ func (cfg *NetworkLoadBalancer) Aliases() []string {
...
+	seen := make(map[string]bool)
+	for _, rule := range cfg.Rules {
+		for _, entry := range rule.Aliases {
+			if _, value := seen[entry]; !value {
 				uniqueAliases = append(uniqueAliases, entry)
+				seen[entry] = true
 			}
 		}
 	}
 	return uniqueAliases
```

There are two pros of using `map[string]struct{}` instead of `map[string]bool`:
1. A `struct{}{}` takes zero byte
2. It is ambiguous what `seen["something"] = false` means

```suggestion
	var uniqueAliases []string
	seen := make(map[string]struct{}{})
	exists := struct{}{}
	for _, rule := range cfg.Rules {
		for _, entry := range rule.Aliases {
			if _, ok := seen[entry]; !ok {
				uniqueAliases = append(uniqueAliases, entry)
				seen[entry] = exists
			}
		}
	}
	return uniqueAliases
```

internal/pkg/template/workload.go
```diff
@@ -291,14 +280,39 @@ func (cfg *NetworkLoadBalancer) Aliases() []string {
...
 				seen[entry] = true
+			}
+		}
+	}
+	return uniqueAliases
+}
+
+// Aliases return all the unique aliases specified across all the routing rules in ALB.
+// Currently, we only have primary routing rule, but we will be getting additional routing rule soon.
+func (cfg *ALBListener) Aliases() []string {
```
Lou1415926: Can we move these functions to below the corresponding struct definition?

internal/pkg/template/workload_test.go
```diff
@@ -479,3 +465,93 @@ func TestRollingUpdateRollbackConfig_TruncateAlarmName(t *testing.T) {
...
+		"LBWS having no aliases": {
+			opts: ALBListener{
+				Rules: []ALBListenerRule{
+					{
+						Aliases: []string{},
+					},
+					{
+						Aliases: []string{},
+					},
+				},
```
Lou1415926: nit:
```suggestion
				Rules: []ALBListenerRule{{},{}},
```

internal/pkg/template/workload_test.go
```diff
@@ -479,3 +465,93 @@ func TestRollingUpdateRollbackConfig_TruncateAlarmName(t *testing.T) {
...
+			opts: NetworkLoadBalancer{
+				Listener: []NetworkLoadBalancerListener{
+					{
+						Aliases: []string{},
+					},
+					{
+						Aliases: []string{},
+					},
+				},
+			},
```
Lou1415926: nit:
```suggestion
				Rules: []ALBListenerRule{{},{}},
```

internal/pkg/template/workload_test.go
Lou1415926: Yeah I think in this case we should. Ideally the tests are agnostic of the implementation: it doesn't know if the implementations are actually the same or not.

internal/pkg/template/templates/workloads/partials/cf/alb.yml
```diff
@@ -1,50 +1,51 @@
...
+        Value: {{$rule.Stickiness}}
     TargetType: ip
     VpcId:
       Fn::ImportValue:
         !Sub "${AppName}-${EnvName}-VpcId"
-
+    {{- end}}
```
Lou1415926: nit: Does this end `{{- range $i, $rule := .ALB.Rules}}`? If so, can we fix the indentation so that they are aligned?

internal/pkg/template/templates/workloads/services/backend/cf.yml
```diff
@@ -99,25 +88,30 @@ Resources:
     Type: AWS::ECS::Service
     DependsOn:
       - EnvControllerAction
-      {{- if and .ALBEnabled .HTTPSListener}}
-      - HTTPListenerRuleWithDomain
-      - HTTPSListenerRule
-      {{ else if .ALBEnabled}}
-      - HTTPListenerRule
+      {{- if .ALB }}
+      {{- range $i, $rule := .ALB.Rules }}
```
Lou1415926: is `rule` used?

internal/pkg/template/templates/workloads/services/backend/cf.yml
```diff
@@ -99,25 +88,30 @@ Resources:
...
       {{- end}}
+      {{- end }}
+      {{- end }}
     Properties:
       {{- "\n"}}{{ include "service-base-properties" . | indent 6 }}
       ServiceRegistries: !If [ExposePort, [{RegistryArn: !GetAtt DiscoveryService.Arn, Port: !Ref TargetPort}], !Ref "AWS::NoValue"]
-      {{- if .ALBEnabled}}
+      {{- if .ALB}}
       {{- if .HTTPHealthCheck.GracePeriod }}
       HealthCheckGracePeriodSeconds: {{.HTTPHealthCheck.GracePeriod}}
```
Lou1415926, line 128: I wonder if `HTTPHealthCheck` should have been under the `ALB` struct 🤔? If so, I'm okay if we add it in a separate PR as well - this one is already biggg

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -263,10 +249,6 @@ func (s *BackendService) Parameters() ([]*cloudformation.Parameter, error) {
 				ParameterKey:   aws.String(WorkloadRulePathParamKey),
 				ParameterValue: s.manifest.RoutingRule.Path,
 			},
-			{
-				ParameterKey:   aws.String(WorkloadRulePathSliceParamKey),
-				ParameterValue: aws.String(strings.Join([]string{aws.StringValue(s.manifest.RoutingRule.Path)}, ",")),
-			},
 			{
 				ParameterKey:   aws.String(WorkloadStickinessParamKey),
```
Lou1415926: I want to make sure: I guess `s.manifest.RoutingRule.IsEmpty()` is equivalent to `.ALB` in go template?

internal/pkg/deploy/cloudformation/stack/testdata/stacklocal/cf.params.json
```diff
@@ -10,7 +10,6 @@
     "HTTPSEnabled": "true",
     "LogRetention": "30",
     "RulePath": "my-svc",
```
Lou1415926, line 12: I thought in this PR we are removing these parameters (`RulePath`, `Stickiness`) - why are they still in this integration test 🤔 ? Is this test passing?

internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/http-autoscaling-params.json
```diff
@@ -14,7 +14,6 @@
 		"TargetContainer": "http-autoscaling",
 		"TargetPort": "8080",
 		"RulePath": "http-autoscaling-path",
```
Lou1415926, line 16: Same question here (why are the parameters still there) 🤔

internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-nlb-dev.params.json
```diff
@@ -12,7 +12,6 @@
     "NLBAliases": "",
```
Lou1415926, line 12: Same question

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
```diff
@@ -41,32 +41,14 @@ Parameters:
     Type: String
   TargetPort:
     Type: Number
-{{- if .NLB }}
-  NLBAliases:
```
Lou1415926: https://github.com/aws/copilot-cli/pull/3189 The reason why we added these parameters is to render the uri. Are the URI still showing up correctly after you removed the params?

internal/pkg/template/templates/workloads/partials/cf/alb.yml
iamhopaul123: +1 and we could add comment like 

```
{{- end }}{{/* range $topic := .Subscribe.Topics */}}
```

internal/pkg/deploy/cloudformation/stack/testdata/stacklocal/cf.params.json
iamhopaul123: +1 i think the tests pass because we don't remove them from the param list (which we might want to)

paragbhingre: I had understood the comment differently at the 1st and that's why I only removed parameters from CFN templates and not param list which indeed fails the deployment and tests doesn't catch that for a reason that penghao mentioned above. So We need `RulePath`, `NLBPort` and `NLBAliases` in `uri.go` so we cannot get rid off them. But we can get rid off `Stickiness` so I am removing that one parameter only from the param list and cfn templates.

internal/pkg/template/templates/workloads/services/backend/cf.yml
paragbhingre: Ohh good question, the reason why I did not add it under `.ALB` was because if use observe L106 then we use this value on a service level here. I think we need to fix this scenario as a part of another PR. Because if we try to get value of the `GracePeriod ` from `.ALB` then we need to get that value from primary listener rule only because every listener rule can have their own specific value. So We need to make decision for this particular `HealthCheckGracePeriodSeconds` param what value should be assigned. That is why kept this param as is for now.

internal/pkg/deploy/cloudformation/stack/backend_svc.go
paragbhingre: yes, you are right.

internal/pkg/template/templates/workloads/services/backend/cf.yml
Lou1415926: > we need to get that value from primary listener rule only because every listener rule can have their own specific value.

Why do we assume that `GracePeriod` is a property of a listener rule?

Lou1415926: I think it's okay that we leave if outside of `.ALB`, but for a different reason.

This [field](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-service.html#cfn-ecs-service-healthcheckgraceperiodseconds) is used for services with ELB, which can be an ALB or an NLB.

It's fine that we keep it as is ntil we have a manifest field to set the grace period for NLB - for example ⬇️  - 
```
nlb:
  grace_periods: 1s
```

paragbhingre: > It's fine that we keep it as is ntil we have a manifest field to set the grace period for NLB 

this is another bug I think I have already discussed with Efe and we can fix that separately.  [Here](https://github.com/aws/copilot-cli/pull/4501#discussion_r1108939008) is another field with the similar problem.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -463,6 +484,117 @@ func convertEnvSecurityGroupCfg(mft *manifest.Environment) (*template.SecurityGr
 	}, nil
 }
 
+func (s *LoadBalancedWebService) convertApplicationLoadBalancer() (*template.ALBListener, error) {
```
iamhopaul123: Should we rename this function as it seems to return `ALBListener`. `convertApplicationLoadBalancer` I would assume it returns ALB config but it actually returns listener config.

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -191,18 +182,13 @@ func (s *BackendService) Template() (string, error) {
...
-		AllowedSourceIps:    allowedSourceIPs,
 		DeregistrationDelay: deregistrationDelay,
-		HostedZoneAliases:   hostedZoneAliases,
-		HTTPSListener:       s.httpsEnabled,
-		HTTPRedirect:        s.httpsEnabled,
 		HTTPTargetContainer: template.HTTPTargetContainer{
 			Port: targetContainerPort,
 			Name: targetContainer,
 		},
 		HTTPHealthCheck: convertHTTPHealthCheck(&s.manifest.RoutingRule.HealthCheck),
```
iamhopaul123, line 190: Shall we move `HTTPTargetContainer` and `HTTPHealthCheck` to `ALB` as well? They seem to be unique per listener too?

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -191,18 +182,13 @@ func (s *BackendService) Template() (string, error) {
...
-		HTTPVersion:     convertHTTPVersion(s.manifest.RoutingRule.ProtocolVersion),
+		ALB:             albConfig,
```
iamhopaul123: Maybe consider renaming `ALB`. It's ALB listener config instead.

internal/pkg/template/workload.go
```diff
@@ -756,21 +794,18 @@ type WorkloadOpts struct {
...
 
 	// Additional options for service templates.
 	WorkloadType            string
 	HealthCheck             *ContainerHealthCheck
 	HTTPTargetContainer     HTTPTargetContainer
 	HTTPHealthCheck         HTTPHealthCheckOpts
 	DeregistrationDelay     *int64
-	AllowedSourceIps        []string
 	NLB                     *NetworkLoadBalancer
+	ALB                     *ALBListener
```
iamhopaul123: Same for this `ALB`. `NLB` is `NetworkLoadBalancer` whereas `ALB` is  `ALBListener`

For ALB there are ALB level config and listener level config. I think the struct should be like this
```golang
type ApplicationLoadBalancer struct {
   LBSetting
   Listeners []listenerconfig
}

type listenerconfig struct {
   ...
}
```

paragbhingre: I agree to your point of renaming `ALB` to `ALBListener` as it holds listener configurations. And for the structure that you are suggesting above:
1. We have only one listener and multiple listener rules so I think there is no need to have `Listeners []listenerconfig`.  
2. The reason why we got rid of the separate `LBSetting` and `Listeners` because of this [comment](https://github.com/aws/copilot-cli/pull/4501#discussion_r1113481634) that we discussed earlier. 

So I think just renaming `ALB` to `ALBListeners` would do as per our expectations. Let me know what do you feel about it.

internal/pkg/deploy/cloudformation/stack/backend_svc.go
paragbhingre: So the reason why we can not move this configuration inside ALB is: 
1. `HTTPTargetContainer` this is being used with service connect and if we put it inside ALB then it would be more complicated to figure out which configuration to be used for Service Connect from the list of listener rules having `HTTPSTargetContainer`. So I kept this config as is until we decide to support multiple ports through SC. 
2. For `HTTPHealthCheck` Wanxian also asked the same question [here](https://github.com/aws/copilot-cli/pull/4501#discussion_r1126997091). So we have a bug where we use `GracePeriod` param from HTTPSHealthCheck for ALB as well as NLB. So we will fix this bug separately.

internal/pkg/template/workload.go
```diff
@@ -423,6 +396,21 @@ func SecretFromSecretsManager(value string) secretsManagerName {
 	}
 }
 
+// ALBListenerRule holds configuration that's needed for an Application Load Balancer listener rules.
```
```suggestion
// ALBListenerRule holds configuration that's needed for an Application Load Balancer listener rule.
```

internal/pkg/template/workload.go
```diff
@@ -423,6 +396,21 @@ func SecretFromSecretsManager(value string) secretsManagerName {
...
+type ALBListenerRule struct {
```
Lou1415926: nit: can we move this definition to just above `type ALBListener struct` so that ALB-related structs are close to each other?

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -330,18 +309,10 @@ func (s *LoadBalancedWebService) Parameters() ([]*cloudformation.Parameter, erro
 				ParameterKey:   aws.String(WorkloadRulePathParamKey),
 				ParameterValue: s.manifest.RoutingRule.Path,
```
Lou1415926, line 310: After we support multiple rules, I assume we will just pick one paths from all the rules, is this correct? Then when showing the URI, we will just show that one path.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -144,16 +144,21 @@ func convertContainerHealthCheck(hc manifest.ContainerHealthCheck) *template.Con
...
+				if isDuplicateAliasEntry(aliasesFor[*alias.HostedZone], aws.StringValue(alias.Alias)) {
+					continue
+				}
 				aliasesFor[*alias.HostedZone] = append(aliasesFor[*alias.HostedZone], *alias.Alias)
 				continue
 			}
 			if defaultHostedZone != nil {
+				if isDuplicateAliasEntry(aliasesFor[*defaultHostedZone], aws.StringValue(alias.Alias)) {
+					continue
+				}
```
Lou1415926, line 161: I agree that this is not the most efficient. We can make a refactor here in the future and hopefully get rid of the `isDuplicateAliasEntry`. It's fine with me that we leave if as is now - the number of aliases can't be large, so the computational cost can't be too bad. I'd say the biggest reason why I'd wanted a refactor here is to increase the readability.

We don't need to refactor this in this PR because there are already a lot going on now. This is just something that we should do in the future 👍🏼

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -463,6 +484,117 @@ func convertEnvSecurityGroupCfg(mft *manifest.Environment) (*template.SecurityGr
...
+	}
+
+	config := &template.ALBListener{
+		Rules:             rules,
+		RedirectToHTTPS:   httpRedirect,
+		IsHTTPS:           s.httpsEnabled,
+		HostedZoneAliases: aliasesFor,
+	}
+
+	return config, nil
```
```suggestion
	return &template.ALBListener{
		Rules:             rules,
		RedirectToHTTPS:   httpRedirect,
		IsHTTPS:           s.httpsEnabled,
		HostedZoneAliases: aliasesFor,
	}, nil
```

iamhopaul123 (approved):
lgtm and +1 to wx's fb

internal/pkg/deploy/cloudformation/stack/testdata/stacklocal/cf.params.json
iamhopaul123: We can probably get rid of these params by calling API to get the rule path, port, host header etc. But I think we can do that later.

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
paragbhingre: yes we are just going to showcase primary path that they configure in `http.path`. Whatever they configure in `additional_rules` they should be able to access the the same link with different path. And if configured different aliases then they should just use that `alias/path` to reach out to that particular service.

Lou1415926 (approved):
:shipit:

PR #4509, opened 2023-02-16
chore: always render Custom Resource object location on `package`

efekarakus:

### DIFF
--- a/internal/pkg/aws/s3/s3.go
+++ b/internal/pkg/aws/s3/s3.go
@@ -130,7 +130,7 @@ func (s *S3) EmptyBucket(bucket string) error {
 
 // ParseURL parses Object URLs or s3 URIs and returns the bucket name and the key.
 // For example, the object URL: "https://stackset-myapp-infrastru-pipelinebuiltartifactbuc-1nk5t9zkymh8r.s3-us-west-2.amazonaws.com/scripts/dns-cert-validator/dd2278811c3"
-// or alternativey, the s3 URI: "s3://stackset-myapp-infrastru-pipelinebuiltartifactbuc-1nk5t9zkymh8r/scripts/dns-cert-validator/dd2278811c3"
+// or alternatively, the s3 URI: "s3://stackset-myapp-infrastru-pipelinebuiltartifactbuc-1nk5t9zkymh8r/scripts/dns-cert-validator/dd2278811c3"
 // Returns "stackset-myapp-infrastru-pipelinebuiltartifactbuc-1nk5t9zkymh8r" and  "scripts/dns-cert-validator/dd2278811c3"
 func ParseURL(url string) (bucket string, key string, err error) {
 	if strings.HasPrefix(url, s3URIPrefix) {
@@ -208,6 +208,10 @@ func (s *S3) upload(bucket, key string, buf io.Reader) (string, error) {
 	return resp.Location, nil
 }
 
+// parseS3URI parses the bucket name and object key from a [s3:// URI].
+// For example: s3://mybucket/puppy.jpg
+//
+// [s3:// URI]: https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-bucket-intro.html#accessing-a-bucket-using-S3-format
 func parseS3URI(uri string) (bucket, key string, err error) {
 	parsed := strings.SplitN(strings.TrimPrefix(uri, s3URIPrefix), "/", 2)
 	if len(parsed) != 2 {
@@ -216,6 +220,10 @@ func parseS3URI(uri string) (bucket, key string, err error) {
 	return parsed[0], parsed[1], nil
 }
 
+// parseObjectURL parses the bucket name and object key from a [virtual-hosted-style access URL].
+// For example: https://DOC-EXAMPLE-BUCKET1.s3.us-west-2.amazonaws.com/puppy.png
+//
+// [virtual-hosted-style access URL]: https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-bucket-intro.html#virtual-host-style-url-ex
 func parseObjectURL(url string) (bucket, key string, err error) {
 	parsedURL := strings.SplitN(strings.TrimPrefix(url, "https://"), "/", 2)
 	if len(parsedURL) != 2 {
--- a/internal/pkg/aws/s3/s3.go
+++ b/internal/pkg/aws/s3/s3.go
@@ -9,6 +9,8 @@ import (
 	"io"
 	"strings"
 
+	awsarn "github.com/aws/aws-sdk-go/aws/arn"
+
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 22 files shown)


internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -96,14 +98,25 @@ type EnvConfig struct {
...
 
+func (cfg *EnvConfig) loadCustomResourceURLs(crs []uploadable) error {
+	if len(cfg.CustomResourcesURLs) != 0 {
+		return nil
+	}
+	bucket, _, err := s3.ParseARN(cfg.ArtifactBucketARN)
+	if err != nil {
+		return fmt.Errorf("parse artifact bucket ARN: %w", err)
+	}
+	cfg.CustomResourcesURLs = customResourceURLs("dummy" /* region is ignored when parsing URLs */, bucket, crs)
```
iamhopaul123: Remind me why we ignore region and does this create invalid urls 🤔

efekarakus: 👍 OK boss I changed things up so that this question is no longer relevant.

`s3.ParseURL` can now accept both S3 and an HTTPS URI so I don't have to write "dummy". Check out the last commit.

internal/pkg/aws/s3/s3.go
```diff
@@ -120,29 +125,30 @@ func (s *S3) EmptyBucket(bucket string) error {
 	}
 }
 
-// ParseURL parses S3 object URL and returns the bucket name and the key.
-// For example: https://stackset-myapp-infrastru-pipelinebuiltartifactbuc-1nk5t9zkymh8r.s3-us-west-2.amazonaws.com/scripts/dns-cert-validator/dd2278811c3
-// returns "stackset-myapp-infrastru-pipelinebuiltartifactbuc-1nk5t9zkymh8r" and
-// "scripts/dns-cert-validator/dd2278811c3"
+// ParseURL parses Object URLs or s3 URIs and returns the bucket name and the key.
+// For example, the object URL: "https://stackset-myapp-infrastru-pipelinebuiltartifactbuc-1nk5t9zkymh8r.s3-us-west-2.amazonaws.com/scripts/dns-cert-validator/dd2278811c3"
+// or alternativey, the s3 URI: "s3://stackset-myapp-infrastru-pipelinebuiltartifactbuc-1nk5t9zkymh8r/scripts/dns-cert-validator/dd2278811c3"
```
iamhopaul123:
```suggestion
// or alternatively, the s3 URI: "s3://stackset-myapp-infrastru-pipelinebuiltartifactbuc-1nk5t9zkymh8r/scripts/dns-cert-validator/dd2278811c3"
```

internal/pkg/aws/s3/s3.go
```diff
@@ -198,3 +204,32 @@ func (s *S3) upload(bucket, key string, buf io.Reader) (string, error) {
 	}
 	return resp.Location, nil
 }
+
+func parseS3URI(uri string) (bucket, key string, err error) {
```
iamhopaul123, line 215: Can we add examples for both `parseS3URI` and `parseObjectURL`? In case we'll use them in the other functions. Also we could use regex which is easier to read and less error prone 🤔

internal/pkg/aws/s3/s3.go
```diff
@@ -198,3 +204,32 @@ func (s *S3) upload(bucket, key string, buf io.Reader) (string, error) {
...
+	parsed := strings.SplitN(strings.TrimPrefix(uri, s3URIPrefix), "/", 2)
+	if len(parsed) != 2 {
+		return "", "", fmt.Errorf("cannot parse S3 URI %s into bucket name and key", uri)
+	}
+	return parsed[0], parsed[1], nil
+}
+
+func parseObjectURL(url string) (bucket, key string, err error) {
+	parsedURL := strings.SplitN(strings.TrimPrefix(url, "https://"), "/", 2)
```
iamhopaul123, line 228: Just double check: is it always `https`?

efekarakus: oh! perhaps it wasn't clear `parseObjectURL` is a refactor of the previous `ParseURL` func. I renamed the function to make it private, there is no other change.

And yea, there are [two protocols](https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-bucket-intro.html) to access the object (`https://` and `s3://`)

internal/pkg/aws/s3/s3.go
efekarakus: I can add the examples, but they are already there in the exported function.

> Also we could use regex which is easier to read and less error prone 

I wrote this in the other comment, I don't think I made it clear but the logic here is not new :) I renamed `ParseURL` to `parseObjectURL`.

Only `parseS3URI` is new! I'm not convinced that a regular expression is easier to read than a `strings.Split("/")` 💭

iamhopaul123 (approved):

internal/pkg/aws/s3/s3.go
```diff
@@ -176,7 +182,7 @@ func (s *S3) isBucketExists(bucket string) (bool, error) {
 	}
 	_, err := s.s3Client.HeadBucket(input)
 	if err != nil {
-		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == notFound {
+		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == errCodeNotFound {
```
Lou1415926, line 188: Is `errCodeNotFound` defined in some aws sdk package? That is, would ⬇️ be possible
```go
import "github.com/aws/aws-sdk-go/aws"

const errCodeNotFound = aws.ErrCodeNotFound
```

internal/pkg/deploy/cloudformation/stack/env.go
```diff
@@ -96,14 +98,28 @@ type EnvConfig struct {
 	ForceUpdate         bool
 }
 
+func (cfg *EnvConfig) loadCustomResourceURLs(crs []uploadable) error {
```
Lou1415926, line 101: Update: scratch this comment ! misread the code


---
what do you think of this signature?
```go
func (cfg *EnvConfig) loadCustomResourceURLs(crs []uploadable) (map[string]string, error) 
```

This way, in `NewEnvStackConfig(input *EnvConfig) (*Env, error)`, we can do
```go
urls, err := input.loadCustomResourceURLs(uploadableCRs(crs).convert());
return &Env{
		in:                in,
		prevParams:        prevParams,
		lastForceUpdateID: lastForceUpdateID,
		parser:            fs,
                customResourceURLs: urls
	}, nil
```

It's slightly more obvious what fields are set and what fields are not this way

Lou1415926 (commented):
LGTM! Just two little nits

internal/pkg/aws/s3/s3.go
efekarakus: I coudn't find that error code exposed in the AWS SDK :( https://github.com/search?q=repo%3Aaws%2Faws-sdk-go+NotFound+language%3AGo&type=code&p=2&l=Go

internal/pkg/deploy/cloudformation/stack/env.go
efekarakus: We discussed offline! the `load` method's receive is the runtimeConfig or the inputConfig for an environment and not `Env` itself

Lou1415926 (approved):

PR #4529, opened 2023-02-20
chore(refactor): Convert current buildargs into `map`

KollaAdithya:
This PR will refactor current docker `buildArgs`  into `map` so that the same function will be used for both sidecar and main containers.


internal/pkg/cli/deploy/workload.go
```diff
@@ -311,40 +353,52 @@ func (d *workloadDeployer) uploadContainerImage(imgBuilderPusher imageBuilderPus
...
+		digests[k] = RuntimeImage{
+			Digest:         digest,
+			CustomTag:      d.runtimeImage.CustomTag,
+			GitShortCommit: d.runtimeImage.GitShortCommit,
+			UUID:           d.runtimeImage.UUID,
+		}
 	}
-	return aws.String(digest), nil
+	image := digests[d.name]
+	return &image.Digest, nil
```
KollaAdithya, line 366: In this PR I am passing only main container `Digest`. In the next refactor the return type will be changed to `map[string]RuntimeImage`.

internal/pkg/cli/deploy/workload.go
```diff
@@ -37,13 +37,15 @@ import (
...
 	"github.com/aws/copilot-cli/internal/pkg/workspace"
+	"github.com/google/uuid"
 	"github.com/spf13/afero"
 )
 
 const (
 	fmtForceUpdateSvcStart    = "Forcing an update for service %s from environment %s"
 	fmtForceUpdateSvcFailed   = "Failed to force an update for service %s from environment %s: %v.\n"
 	fmtForceUpdateSvcComplete = "Forced an update for service %s from environment %s.\n"
+	imageTagLatest            = "latest"
```
efekarakus: nit: can we move this to a separate `const`block? The ones above seem to be releated to a spinner whereas this one is for container images.

internal/pkg/cli/deploy/workload.go
```diff
@@ -160,7 +162,7 @@ type WorkloadDeployerInput struct {
 	Name             string
 	App              *config.Application
 	Env              *config.Environment
-	ImageTag         string
+	RuntimeImage     RuntimeImage
```
efekarakus: What do you think of renaming the struct and field like this?
```suggestion
	Image     ContainerImageIdentifier
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -170,6 +172,14 @@ type WorkloadDeployerInput struct {
 	customResources customResourcesFunc
 }
 
+// RuntimeImage is the configuration of image digest and tags of an ECR image.
+type RuntimeImage struct {
+	Digest         string
+	CustomTag      string
+	GitShortCommit string
+	UUID           string
```
efekarakus: nit: to highlight that all of these are tags, what do you think of adding the suffix:
```suggestion
	CustomTag      string
	GitShortCommitTag string
	UUIDTag           string  // If not specified, a random UUID is generated.
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -224,12 +234,17 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 	if err != nil {
 		return nil, fmt.Errorf("unmarshal the manifest used to deploy environment %s: %w", in.Env.Name, err)
 	}
+	uuidTag, err := generateUUID()
```
efekarakus: should this be set only if the client hasn't specified one? or if the client shouldn't specify one can we make the `UUUIDTag` field private ? (i.e. `uuidTag`)

internal/pkg/cli/deploy/workload.go
```diff
@@ -302,6 +317,33 @@ func (d *workloadDeployer) forceDeploy(in *forceDeployInput) error {
 	return nil
 }
 
+// generateUUID generates a random UUID to tag main and sidecar container images.
+func generateUUID() (string, error) {
+	id, err := uuid.NewRandom()
+	if err != nil {
+		return "", fmt.Errorf("generate random uuid to tag workload images: %w", err)
```
efekarakus: the function name seems generic enough ("generate uuid") whereas the wrapped error text is too specific. My recommendation is to get rid of the function and just move it in the constructor of `newWorkloadDeployer` for you to keep the specific error message

internal/pkg/cli/deploy/workload.go
```diff
@@ -302,6 +317,33 @@ func (d *workloadDeployer) forceDeploy(in *forceDeployInput) error {
...
+	}
+	return id.String(), nil
+}
+
+// ReferenceTag returns the tag that should be used to reference the image.
+func (i RuntimeImage) ReferenceTag() string {
+	cgTag := i.customOrGitTag()
+	if cgTag != "" {
```
```suggestion
func (img RuntimeImage) ReferenceTag() string {
	if tag := img.customOrGitTag(); tag != "" {
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -311,40 +353,52 @@ func (d *workloadDeployer) uploadContainerImage(imgBuilderPusher imageBuilderPus
 		return nil, nil
 	}
 	// If it is built from local Dockerfile, build and push to the ECR repo.
-	buildArg, err := buildArgs(d.name, d.imageTag, d.workspacePath, d.mft)
+	buildArg, err := buildArgs(d.name, d.workspacePath, d.runtimeImage, d.mft)
```
efekarakus:
```suggestion
	buildArgsPerContainer, err := buildArgs(d.name, d.workspacePath, d.runtimeImage, d.mft)
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -311,40 +353,52 @@ func (d *workloadDeployer) uploadContainerImage(imgBuilderPusher imageBuilderPus
...
 	if err != nil {
 		return nil, err
 	}
-	digest, err := imgBuilderPusher.BuildAndPush(dockerengine.New(exec.NewCmd()), buildArg)
-	if err != nil {
-		return nil, fmt.Errorf("build and push image: %w", err)
+	images := make(map[string]RuntimeImage, len(buildArg))
+	for k, v := range buildArg {
```
efekarakus:
```suggestion
	for name, buildArgs := range buildArgsPerContainer {
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -311,40 +353,52 @@ func (d *workloadDeployer) uploadContainerImage(imgBuilderPusher imageBuilderPus
...
+			UUID:           d.runtimeImage.UUID,
+		}
 	}
-	return aws.String(digest), nil
+	image := images[d.name]
+	return &image.Digest, nil
 }
 
-func buildArgs(name, imageTag, workspacePath string, unmarshaledManifest interface{}) (*dockerengine.BuildArguments, error) {
+func buildArgs(name, workspacePath string, i RuntimeImage, unmarshaledManifest interface{}) (map[string]*dockerengine.BuildArguments, error) {
```
efekarakus: There is a fine balance with naming variables, you want to be as short as possible while not not losing any info (https://a-nickels-worth.dev/papers/varnaming/#headline-12).

This is a matter of taste so feel free to disagree, but `i` in my mind is always associated with "index" and for-loops. Instead, `img` is 2 characters extra but gets rid of that ambiguity when I am reading the body of this function:

```suggestion
func buildArgsPerContainer(name, workspacePath string, img RuntimeImage, unmarshaledManifest interface{}) (map[string]*dockerengine.BuildArguments, error) {
```

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -187,8 +187,10 @@ func (s *LoadBalancedWebService) BuildRequired() (bool, error) {
 }
 
 // BuildArgs returns a docker.BuildArguments object given a ws root directory.
-func (s *LoadBalancedWebService) BuildArgs(wsRoot string) *DockerBuildArgs {
-	return s.ImageConfig.Image.BuildConfig(wsRoot)
+func (s *LoadBalancedWebService) BuildArgs(wsRoot string) map[string]*DockerBuildArgs {
```
efekarakus: nit: I know this was there before 🙏 , but ideally the `manifest` package is completely unaware of concepts like "workspace". 

Instead, can we rename this parameter to and update the comment:
```suggestion
func (s *LoadBalancedWebService) BuildArgs(contextDir string) map[string]*DockerBuildArgs {
```
The clients can decide to pass the workspace root as the default context directory, but the manifest is much more flexible than that. By updating the parameter name, we are hinting at other clients that it doesn't need to be always the path to the workspace

efekarakus:
Also awesome! The size of the PR was perfect 👌 thank you so much Adi, it was really nice to review this - v enjoyable 🚀 !

internal/pkg/cli/deploy/workload.go
KollaAdithya: yeah I will make this field private `uuidTag` so that the client can not specify.

internal/pkg/cli/deploy/workload.go
```diff
@@ -172,12 +174,12 @@ type WorkloadDeployerInput struct {
...
-	Digest         string
-	CustomTag      string
-	GitShortCommit string
-	UUID           string
+// ContainerImageIdentifier is the configuration of image digest and tags of an ECR image.
+type ContainerImageIdentifier struct {
+	Digest            string
+	CustomTag         string
+	GitShortCommitTag string
+	uuidTag           string // If not specified, a random UUID is generated.
```
efekarakus: nit: now that the field is private the comment is no longer relevant :)

internal/pkg/manifest/rd_web_svc.go
```diff
@@ -160,9 +160,9 @@ func (s *RequestDrivenWebService) ContainerPlatform() string {
 }
 
 // BuildArgs returns a docker.BuildArguments object given a ws root directory.
```
efekarakus: can we also update the comments to not say "ws root" anymore 🥺

efekarakus (approved):
Looks good, approving w the label. Feel free to remove once addressed

internal/pkg/cli/deploy/workload.go
```diff
@@ -224,12 +236,17 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 	if err != nil {
 		return nil, fmt.Errorf("unmarshal the manifest used to deploy environment %s: %w", in.Env.Name, err)
 	}
+	id, err := uuid.NewRandom()
+	if err != nil {
+		return nil, fmt.Errorf("generate random uuid to tag workload images: %w", err)
```
paragbhingre:
```suggestion
		return nil, fmt.Errorf("generate a random UUID to tag workload images: %w", err)
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -170,6 +174,14 @@ type WorkloadDeployerInput struct {
 	customResources customResourcesFunc
 }
 
+// ContainerImageIdentifier is the configuration of image digest and tags of an ECR image.
```
paragbhingre:
```suggestion
// ContainerImageIdentifier is the configuration of the image digest and tags of an ECR image.
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -311,40 +345,52 @@ func (d *workloadDeployer) uploadContainerImage(imgBuilderPusher imageBuilderPus
 		return nil, nil
 	}
 	// If it is built from local Dockerfile, build and push to the ECR repo.
-	buildArg, err := buildArgs(d.name, d.imageTag, d.workspacePath, d.mft)
+	buildArgsPerContainer, err := buildArgsPerContainer(d.name, d.workspacePath, d.image, d.mft)
```
paragbhingre, line 348: dumb question: is `buildArgsPerContainer` is just building docker args for the main container for now?

paragbhingre, line 348: nit: its just my preference which looks short and sweet. 
```suggestion
	buildArgsFor, err := buildArgsPerContainer(d.name, d.workspacePath, d.image, d.mft)
```

paragbhingre (commented):
Looks good to me, I just have a question and few nits.

internal/pkg/cli/deploy/workload.go
KollaAdithya: yes. currently we have only build args for main container, in the next PR I will add the sidecar build args as well.

iamhopaul123 (approved):

PR #4533, opened 2023-02-22
build: upgrade to go1.20

Lou1415926:
Release note for go v1.20: https://tip.golang.org/doc/go1.20

This PR does a bunch of things:
1. Upgrade go v1.20
2. Bump the GitHub action `setup-go` from `@v2` to `@v3`
3. This also requires us to bump the `golangci-lint` version in the GitHub action from `1.46` to `1.52`
4. `rand.Seed` is deprecated in v1.20 so this is fixed.
5. The lint at `1.52` treats linter issue that is *not* in the build tag as fatal.  Therefore the linter issue in the e2e tests need to be fixed now. So they are fixed. 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.20
+          go-version: '1.20'
 
       - name: Set up Node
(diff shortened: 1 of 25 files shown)


efekarakus (approved):
aww yeah 🚀

iamhopaul123 (approved):

PR #4536, opened 2023-02-22
chore: add static site to cfn pkg

iamhopaul123:

### DIFF
--- a/internal/pkg/cli/deploy/static_site.go
+++ b/internal/pkg/cli/deploy/static_site.go
@@ -82,7 +82,7 @@ func (d *staticSiteDeployer) DeployWorkload(in *DeployWorkloadInput) (ActionReco
 	}); err != nil {
 		return nil, err
 	}
-	return NoopActionRecommender{}, nil
+	return noopActionRecommender{}, nil
 }
 
 // UploadArtifacts uploads static assets to the app stackset bucket.
--- a/internal/pkg/cli/deploy/static_site.go
+++ b/internal/pkg/cli/deploy/static_site.go
@@ -64,7 +64,7 @@ func (*staticSiteDeployer) IsServiceAvailableInRegion(region string) (bool, erro
 // GenerateCloudFormationTemplate generates a CloudFormation template and parameters for a workload.
 func (d *staticSiteDeployer) GenerateCloudFormationTemplate(in *GenerateCloudFormationTemplateInput) (
 	*GenerateCloudFormationTemplateOutput, error) {
-	conf, err := d.stackConfiguration(in.StackRuntimeConfiguration)
+	conf, err := d.stackConfiguration(&in.StackRuntimeConfiguration)
 	if err != nil {
 		return nil, err
 	}
@@ -73,13 +73,16 @@ func (d *staticSiteDeployer) GenerateCloudFormationTemplate(in *GenerateCloudFor
 
 // DeployWorkload deploys a static site service using CloudFormation.
 func (d *staticSiteDeployer) DeployWorkload(in *DeployWorkloadInput) (ActionRecommender, error) {
-	conf, err := d.stackConfiguration(in.StackRuntimeConfiguration)
+	conf, err := d.stackConfiguration(&in.StackRuntimeConfiguration)
 	if err != nil {
 		return nil, err
 	}
-	return nil, d.deploy(in.Options, svcStackConfigurationOutput{
+	if err := d.deploy(in.Options, svcStackConfigurationOutput{
 		conf: cloudformation.WrapWithTemplateOverrider(conf, d.overrider),
-	})
+	}); err != nil {
+		return nil, err
+	}
+	return nil, nil
 }
 
 // UploadArtifacts uploads static assets to the app stackset bucket.
@@ -100,12 +103,12 @@ func (d *staticSiteDeployer) UploadArtifacts() (*UploadArtifactsOutput, error) {
 	return d.uploadArtifacts()
 }
 
-func (d *staticSiteDeployer) stackConfiguration(in StackRuntimeConfiguration) (cloudformation.StackConfiguration, error) {
-	rc, err := d.runtimeConfig(&in)
+func (d *staticSiteDeployer) stackConfiguration(in *StackRuntimeConfiguration) (cloudformation.StackConfiguration, error) {
+	rc, err := d.runtimeConfig(in)
 	if err != nil {
 		return nil, err
 	}
-	conf, err := stack.NewStaticSite(stack.StaticSiteConfig{
+	conf, err := stack.NewStaticSite(&stack.StaticSiteConfig{
 		App:                d.app,
 		EnvManifest:        d.envConfig,
 		Manifest:           d.staticSiteMft,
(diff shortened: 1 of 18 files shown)


internal/pkg/cli/deploy/static_site.go
```diff
@@ -43,14 +48,38 @@ func NewStaticSiteDeployer(in *WorkloadDeployerInput) (*staticSiteDeployer, erro
...
+}
+
 // DeployWorkload deploys a static site service using CloudFormation.
 func (d *staticSiteDeployer) DeployWorkload(in *DeployWorkloadInput) (ActionRecommender, error) {
-	return nil, nil
+	conf, err := d.stackConfiguration(in.StackRuntimeConfiguration)
+	if err != nil {
+		return nil, err
+	}
+	return nil, d.deploy(in.Options, svcStackConfigurationOutput{
```
efekarakus: huh I just noticed this strange pattern that we have ! 

I'd have expected something like this normally:
```go
if err := d.deploy(...); err != nil {
   return nil, err
}
return new(noopActionRecommender), nil
```

internal/pkg/cli/deploy/static_site.go
```diff
@@ -68,5 +97,26 @@ func (d *staticSiteDeployer) UploadArtifacts() (*UploadArtifactsOutput, error) {
 			return nil, err
 		}
 	}
-	return &UploadArtifactsOutput{}, nil
+	return d.uploadArtifacts()
+}
+
+func (d *staticSiteDeployer) stackConfiguration(in StackRuntimeConfiguration) (cloudformation.StackConfiguration, error) {
```
efekarakus: nit: to be consistent w the rest of the deployers
```suggestion
func (d *staticSiteDeployer) stackConfiguration(in *StackRuntimeConfiguration) (cloudformation.StackConfiguration, error) {
```

internal/pkg/deploy/cloudformation/stack/static_site.go
```diff
@@ -0,0 +1,77 @@
...
+	Manifest           *manifest.StaticSite
+	RawManifest        []byte // Content of the manifest file without any transformations.
+	RuntimeConfig      RuntimeConfig
+	RootUserARN        string
+	ArtifactBucketName string
+	Addons             NestedStackConfigurer
+}
+
+// NewStaticSite creates a new CFN stack from a manifest file, given the options.
+func NewStaticSite(conf StaticSiteConfig) (*StaticSite, error) {
```
efekarakus: nit: to save some mem space
```suggestion
func NewStaticSite(conf *StaticSiteConfig) (*StaticSite, error) {
```

internal/pkg/deploy/cloudformation/stack/static_site_test.go
```diff
@@ -0,0 +1,134 @@
...
+			require.Equal(t, tc.wantedStack.image, stack.image)
+			require.Equal(t, tc.wantedStack.manifest, stack.manifest)
+			require.NotNil(t, stack.addons)
+		})
+	}
+}
+
+func TestStaticSite_SerializedParameters(t *testing.T) {
+	c := &StaticSite{
+		wkld: &wkld{
```
efekarakus: instead of relying on the internal private fields can we do like the above and call the constructor: `NewStaticSite`

internal/pkg/deploy/upload/customresource/customresource_test.go
```diff
@@ -246,7 +246,61 @@ func TestBackend(t *testing.T) {
...
+		"EnvControllerFunction":       "manual/scripts/custom-resources/envcontrollerfunction/72297cacaeab3a267e371c17ea3f0235905b0da51410eb31c10f7c66ba944044.zip",
+		"RulePriorityFunction":        "manual/scripts/custom-resources/rulepriorityfunction/1385d258950a50faf4b5cd7deeecbc4bcc79a0d41d631e3977cffa0332e6f0c6.zip",
+	}
+
+	// WHEN
+	crs, err := StaticSite(fakeFS)
+
+	// THEN
+	require.NoError(t, err)
+	require.Equal(t, fakeFS.matchCount, 0, "expected path calls do not match")
```
efekarakus, line 279: nice!

internal/pkg/deploy/upload/customresource/customresource_test.go
```diff
@@ -246,7 +246,61 @@ func TestBackend(t *testing.T) {
...
+		r, err := zip.NewReader(bytes.NewReader(buf.Bytes()), size)
+		require.NoError(t, err)
+
+		_, err = r.Open("index.js")
+		require.NoError(t, err, "an index.js file must be present in all custom resources")
+	}
+
+	// ensure artifact paths match.
+	for _, cr := range crs {
+		require.Equal(t, fakePaths[cr.Name()], cr.ArtifactPath())
```
efekarakus, line 303: What do  you think of deleting this block since we have a match count of 0 and no custom resources for static sites?

internal/pkg/manifest/workload.go
```diff
@@ -836,7 +836,7 @@ func DockerfileBuildRequired(svc interface{}) (bool, error) {
 	}
 	mf, ok := svc.(manifest)
 	if !ok {
-		return false, fmt.Errorf("manifest does not have required methods BuildRequired()")
+		return false, nil
```
efekarakus, line 839: Is this ok to change, do we have any clients that rely on the error instead of checking the boolean?

efekarakus (commented):
Looks great, just some minor requests!

internal/pkg/manifest/workload.go
iamhopaul123: I tried to look at it and did not find any. And I think it's ok to combine "no need to build" and "do not apply" for now as the both results are "we can just skip".

internal/pkg/deploy/upload/customresource/customresource_test.go
iamhopaul123: I keep it here is because the static pattern will have a custom resource which is the step function trigger. Maybe it's ok to keep 🤔

efekarakus (approved):

internal/pkg/cli/deploy/static_site.go
```diff
@@ -43,13 +48,40 @@ func NewStaticSiteDeployer(in *WorkloadDeployerInput) (*staticSiteDeployer, erro
...
+	conf, err := d.stackConfiguration(&in.StackRuntimeConfiguration)
+	if err != nil {
+		return nil, err
+	}
+	if err := d.deploy(in.Options, svcStackConfigurationOutput{
+		conf: cloudformation.WrapWithTemplateOverrider(conf, d.overrider),
+	}); err != nil {
+		return nil, err
+	}
 	return nil, nil
```
Lou1415926: The first param is always a `nil`. I guess this is to satisfy an interface 🤔 
What about returning a no-op `ActionRecommender` instead of `nil`?

internal/pkg/cli/deploy/workload.go
```diff
@@ -55,6 +55,14 @@ type ActionRecommender interface {
 	RecommendedActions() []string
 }
 
+// NoopActionRecommender represents an ActionRecommender that does not recommend anything.
+type NoopActionRecommender struct{}
```
efekarakus: nit: doesn't need to be exported

Lou1415926 (approved):
noice!

PR #4540, opened 2023-02-22
chore(prompt): add MultiSelectOptions method

efekarakus:

### DIFF
--- a/internal/pkg/term/prompt/select.go
+++ b/internal/pkg/term/prompt/select.go
@@ -74,6 +74,27 @@ func (p Prompt) SelectOption(message, help string, opts []Option, promptCfgs ...
 	return prettified.choice2Value[result], nil
 }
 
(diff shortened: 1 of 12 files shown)


internal/pkg/term/prompt/select.go
```diff
@@ -101,6 +122,38 @@ func (p Prompt) SelectOne(message, help string, options []string, promptCfgs ...
 	return result, err
 }
 
+// MultiSelect prompts the user with a list of options to choose from with the arrow keys and enter key.
+func (p Prompt) MultiSelect(message, help string, options []string, validator ValidatorFunc, promptCfgs ...PromptConfig) ([]string, error) {
```
efekarakus, line 126: This method is the same, I just moved it from `prompt.go` to `select.go`

iamhopaul123 (approved):

KollaAdithya (approved):
🚢

PR #4541, opened 2023-02-23
chore(selector): add CFNSelector

efekarakus:
This PR  should be reviewed after #4540.
The commit that has the change is d6b8265f91a9ced608ee2fd06075ccca85c70292.

### DIFF
--- a/internal/pkg/term/selector/cfn.go
+++ b/internal/pkg/term/selector/cfn.go
@@ -1,0 +1,64 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package selector
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/aws/copilot-cli/internal/pkg/template"
+	"github.com/aws/copilot-cli/internal/pkg/term/prompt"
+	"gopkg.in/yaml.v3"
+)
+
+// CFNSelector represents a selector for a CloudFormation template.
+type CFNSelector struct {
+	prompt Prompter
+}
+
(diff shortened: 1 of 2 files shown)


internal/pkg/term/selector/cfn.go
```diff
@@ -0,0 +1,64 @@
...
+		} `yaml:"Resources"`
+	}{}
+	if err := yaml.Unmarshal([]byte(body), &tpl); err != nil {
+		return nil, fmt.Errorf("unmarshal CloudFormation template: %v", err)
+	}
+
+	// Prompt for a selection.
+	var options []prompt.Option
+	for name, resource := range tpl.Resources {
+		if resource.Type == "AWS::Lambda::Function" || strings.HasPrefix(resource.Type, "Custom::") {
```
iamhopaul123, line 42: Maybe it makes more sense to filter out from a specific list but I guess it's ok to be more restrictive at first.

iamhopaul123 (approved):

internal/pkg/term/selector/cfn.go
efekarakus: I kept debating if this method should take as input something like:
```go
func Resources(body string, filterFn ...func(template.CFNResource) bool) ([]template.CFNResource, error)
```
Happy to change it to that boss, if you prefer it to! It's definitely more flexible 👍 at the end I just kept it a bit simpler for now

iamhopaul123: Yeah i think we can keep things simple for now

paragbhingre (approved):
It looks good to me. Putting DNM just in case you decide to change the resource function definition But for me, it looks simple and easy to understand.

PR #4547, opened 2023-02-24
chore: Build and push sidecar images sequentially

KollaAdithya:
1. Removes `BuildRequired()` from all workloads.
2. This PR builds and pushes all sidecar images sequentially.


internal/pkg/cli/deploy/workload.go
```diff
@@ -339,19 +339,27 @@ func (img ContainerImageIdentifier) customOrGitTag() string {
...
+// sidecarReferenceTag returns the tag that should be used to reference the image.
+func (img ContainerImageIdentifier) sidecarReferenceTag() string {
+	if img.GitShortCommitTag != "" {
+		return img.GitShortCommitTag
+	}
+	return img.uuidTag
+}
+
 func (d *workloadDeployer) uploadContainerImages(imgBuilderPusher imageBuilderPusher) (map[string]ContainerImageIdentifier, error) {
 	required, err := manifest.DockerfileBuildRequired(d.mft)
```
KollaAdithya: I am thinking  of removing `func (s *LoadBalancedWebService) BuildRequired() (bool, error)` from all the workloads as we can just depend on just  length of `BuildArgs()` to determine docker build  required or not. what you folks think of it 🤔 .

internal/pkg/cli/deploy/workload.go
```diff
@@ -339,19 +339,27 @@ func (img ContainerImageIdentifier) customOrGitTag() string {
 	return img.GitShortCommitTag
 }
 
+// sidecarReferenceTag returns the tag that should be used to reference the image.
+func (img ContainerImageIdentifier) sidecarReferenceTag() string {
```
efekarakus: Can you elaborate why we need this? 💭 I wonder if we can have the sidecars also use `ReferenceTag` (and hence the custom tag if the user has provided one)

efekarakus: as a side-note, what do you think of renaming the `ReferenceTag` method to just `Tag`?

internal/pkg/cli/deploy/workload.go
efekarakus: sounds good to me!

internal/pkg/cli/deploy/workload.go
```diff
@@ -339,19 +339,27 @@ func (img ContainerImageIdentifier) customOrGitTag() string {
...
 	}
-	if !required {
-		return nil, nil
-	}
 	// If it is built from local Dockerfile, build and push to the ECR repo.
 	buildArgsPerContainer, err := buildArgsPerContainer(d.name, d.workspacePath, d.image, d.mft)
 	if err != nil {
 		return nil, err
 	}
+	if !required && len(buildArgsPerContainer) == 0 {
```
efekarakus: yeah you're right this definitely reads strange, I'd rather just have here:
```suggestion
	if len(buildArgsPerContainer) == 0 {
```
which makes things pretty obvious

internal/pkg/cli/deploy/workload.go
```diff
@@ -370,24 +378,32 @@ func (d *workloadDeployer) uploadContainerImages(imgBuilderPusher imageBuilderPu
...
-			Context:    aws.StringValue(v.Context),
-			Args:       v.Args,
-			CacheFrom:  v.CacheFrom,
-			Target:     aws.StringValue(v.Target),
+	argsPerContainer, err := mf.BuildArgs(workspacePath)
+	if err != nil {
+		return nil, fmt.Errorf("check if manifest requires building from local Dockerfile: %w", err)
+	}
+	dArgs := make(map[string]*dockerengine.BuildArguments, len(argsPerContainer))
+	for container, buildArgs := range argsPerContainer {
```
efekarakus, line 366: is it possible for `buildArgs` to be `nil`? or is the key-value pair not populated if the container has a `location` instead of `build`

internal/pkg/cli/deploy/workload.go
```diff
@@ -370,24 +378,32 @@ func (d *workloadDeployer) uploadContainerImages(imgBuilderPusher imageBuilderPu
...
+		var tags []string
+		if container == name {
+			tags = append(tags, imageTagLatest, img.ReferenceTag())
+		} else {
+			defaultScImgTag := fmt.Sprintf("%s-%s", container, imageTagLatest)
+			tags = append(tags, defaultScImgTag, fmt.Sprintf("%s-%s", container, img.sidecarReferenceTag()))
+		}
```
efekarakus: a slightly different way of writing this that avoids the `else` statement:
```go
tags := []string{imageTagLatest, img.ReferenceTag())
if container != mainContainer {
   tags := []string{fmt.Sprintf("%s-%s", container, imageTagLatest), fmt.Sprintf("%s-%s", container, img.ReferenceTag())
}
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -561,9 +577,15 @@ func (d *workloadDeployer) runtimeConfig(in *StackRuntimeConfiguration) (*stack.
...
+		var imageTag string
+		if container == d.name {
+			imageTag = img.customOrGitTag()
+		} else {
+			imageTag = img.GitShortCommitTag
+		}
 		images[container] = stack.ECRImage{
 			RepoURL:  d.resources.RepositoryURLs[d.name],
-			ImageTag: img.customOrGitTag(),
+			ImageTag: imageTag,
```
efekarakus: why not just `img.ReferenceTag()`?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -95,7 +95,12 @@ func convertSidecars(s map[string]*manifest.SidecarConfig, exposedPorts map[stri
 	sort.Strings(keys)
 	for _, name := range keys {
 		config := s[name]
-		imageURI, _ := config.ImageURI()
+		var imageURI string
+		if uri, buildRequired := config.ImageURI(); buildRequired {
+			imageURI = uri
```
efekarakus: Wut I'm so confused what's happening here 😅 
reading the doc comments for `ImageURI` it says:
```
// ImageURI returns the location of the image if one is set.
// If the image needs to be build, return "" and false.
```

So should this be something like (the opposite of a "build being required"):
```suggestion
		if uri, hasLocation := config.ImageURI(); hasLocation {
			imageURI = uri
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -95,7 +95,12 @@ func convertSidecars(s map[string]*manifest.SidecarConfig, exposedPorts map[stri
...
+		} else {
+			imageURI = rc.Images[name].GetSidecarLocation(name)
```
efekarakus: what do you think of renaming the field to:
```suggestion
			imageURI = rc.PushedImages[name].URL(name)
```
I feel like it conveys it better that these images were the ones that got uploaded to ECR by copilot 💭

internal/pkg/deploy/cloudformation/stack/workload.go
```diff
@@ -113,6 +113,20 @@ func (i ECRImage) GetLocation() string {
 	return fmt.Sprintf("%s:%s", i.RepoURL, "latest")
 }
 
+// GetSidecarLocation returns the sidecar container ECR image URI.
+// Prefer referring the image using git short commit id as <sidecarname>-<gitshortcommitid>.
+// Otherwise, each image after a push to ECR will get a digest and we refer to the image via the digest.
+// Finally, if no digest or tag is present, this occurs with the "package" commands, we default to the <sidecarname>-<latest> tag.
+func (i ECRImage) GetSidecarLocation(name string) string {
```
efekarakus: This method is really similar to `GetLocation` above. 

I wonder if we could have removed the two different methods if `ECRImage` like this:
```go
type ECRImage struct {
	RepoURL  string // RepoURL is the ECR repository URL the container image should be pushed to.
	Tag string // Tag is the container image's unique tag.
	Digest   string // The image digest.
	ContainerName string // The container name. 
	MainContainerName string // The workload's container name.
}
```
The client would pass in both of the names to `ECRImage` in `workloadDeployer.runtimeConfig` and insider `GetLocation` we will check if the two names are the same or not and set things accordingly.

efekarakus: Also if you don't mind renaming the methods to just `URL` 🥺

internal/pkg/manifest/backend_svc.go
```diff
@@ -120,10 +120,21 @@ func (s *BackendService) BuildRequired() (bool, error) {
 }
 
 // BuildArgs returns a docker.BuildArguments object for the service given a context directory.
-func (s *BackendService) BuildArgs(contextDir string) map[string]*DockerBuildArgs {
+func (s *BackendService) BuildArgs(contextDir string) (map[string]*DockerBuildArgs, error) {
```
efekarakus, line 118: aww yea, nice!

internal/pkg/cli/deploy/workload.go
KollaAdithya: As per the design proposal we decided not to tag Sidecar images with `custom tag` provided by user, so that is the reason for separate method to tag sidecar images instead of using `ReferenceTag()`.

internal/pkg/deploy/cloudformation/stack/transformers.go
KollaAdithya: sorry for the confusion 😅. yes,this should be renamed.

internal/pkg/cli/deploy/workload.go
KollaAdithya: `key-value` pair will not at all populate when the container has a `location` instead of `build`. i.e `len(argsPerContainer)` will be zero.

internal/pkg/cli/deploy/workload.go
```diff
@@ -323,7 +323,7 @@ func (d *workloadDeployer) forceDeploy(in *forceDeployInput) error {
 }
 
 // ReferenceTag returns the tag that should be used to reference the image.
```
efekarakus: doc comment needs to be updated :)

internal/pkg/cli/deploy/workload.go
```diff
@@ -370,24 +376,29 @@ func (d *workloadDeployer) uploadContainerImages(imgBuilderPusher imageBuilderPu
...
-			Context:    aws.StringValue(v.Context),
-			Args:       v.Args,
-			CacheFrom:  v.CacheFrom,
-			Target:     aws.StringValue(v.Target),
+	argsPerContainer, err := mf.BuildArgs(workspacePath)
+	if err != nil {
+		return nil, fmt.Errorf("check if manifest requires building from local Dockerfile: %w", err)
+	}
+	dArgs := make(map[string]*dockerengine.BuildArguments, len(argsPerContainer))
+	tags := []string{imageTagLatest, img.Tag()}
```
efekarakus: nit: can we define `tags` inside of the for-loop to make it clear that each `tags` is at a container lvl

internal/pkg/cli/deploy/workload.go
```diff
@@ -370,24 +376,29 @@ func (d *workloadDeployer) uploadContainerImages(imgBuilderPusher imageBuilderPu
...
+	for container, buildArgs := range argsPerContainer {
+		if container != name {
+			tags = []string{fmt.Sprintf("%s-%s", container, imageTagLatest), fmt.Sprintf("%s-%s", container, img.nonCustomTag())}
```
efekarakus: Thinking about this, every new run of `copilot svc package --upload-assets` will result in a new UUID in the constructor.

Wouldn't that mean each time we run the command if the user has a `<sidecar>.image.build` then we will end up pushing a new tag for the same sidecar image even if nothing changed? That seems problematic 💭

internal/pkg/cli/deploy/workload.go
```diff
@@ -561,17 +572,23 @@ func (d *workloadDeployer) runtimeConfig(in *StackRuntimeConfiguration) (*stack.
 	}
 	images := make(map[string]stack.ECRImage, len(in.ImageDigests))
 	for container, img := range in.ImageDigests {
+		imageTag := img.customOrGitTag()
+		if container != d.name {
```
efekarakus, line 560: Can we add a comment here explaining _why_ the logic is different for sidecars? 

If we come back to read this code in the future, I'd be curious why this is necessary in the first place 👀

internal/pkg/deploy/cloudformation/stack/workload.go
```diff
@@ -94,23 +94,36 @@ func (cfg *RuntimeConfig) loadCustomResourceURLs(bucket string, crs []uploadable
...
+func (i ECRImage) URI() string {
+	if i.ContainerName == i.MainContainerName {
+		if i.ImageTag != "" {
+			return fmt.Sprintf("%s:%s", i.RepoURL, i.ImageTag)
+		}
+		if i.Digest != "" {
+			return fmt.Sprintf("%s@%s", i.RepoURL, i.Digest)
+		}
+		return fmt.Sprintf("%s:%s", i.RepoURL, "latest")
+	} else {
```
efekarakus: no need for the else statement

efekarakus (commented):
Looks good! One main issue to investigate, let us know your thoughts!

internal/pkg/cli/deploy/workload.go
KollaAdithya: This also happens  when user runs `copilot svc deploy` and also same uses same image, everytime a new random UUID tag will be attached. What could be potential solution for this 🤔

efekarakus (approved):

internal/pkg/manifest/job.go
```diff
@@ -137,15 +139,21 @@ func (j *ScheduledJob) Publish() []Topic {
 }
 
 // BuildArgs returns a docker.BuildArguments object for the job given a context directory.
-func (j *ScheduledJob) BuildArgs(contextDir string) map[string]*DockerBuildArgs {
+func (j *ScheduledJob) BuildArgs(contextDir string) (map[string]*DockerBuildArgs, error) {
+	required, err := requiresBuild(j.ImageConfig.Image)
+	if err != nil {
+		return nil, fmt.Errorf("check if manifest requires building from local Dockerfile: %w", err)
```
paragbhingre: I think we don't need to wrap this error. AFAICT, I see "check if manifest requires building from local Dockerfile:" already being wrapped inside `buildArgsPerContainer`.

internal/pkg/manifest/lb_web_svc_test.go
```diff
@@ -1608,70 +1607,6 @@ func TestLoadBalancedWebService_Publish(t *testing.T) {
 	}
 }
 
-func TestLoadBalancedWebService_BuildRequired(t *testing.T) {
```
paragbhingre, line 1611: As we remove `BuildRequired`, can we now have unit tests for  `BuildArgs`?

internal/pkg/manifest/backend_svc.go
```diff
@@ -114,16 +114,22 @@ func (s *BackendService) Publish() []Topic {
...
+	if err != nil {
+		return nil, err
+	}
 	buildArgs := make(map[string]*DockerBuildArgs, len(s.Sidecars)+1)
-	buildArgs[aws.StringValue(s.Name)] = s.ImageConfig.Image.BuildConfig(contextDir)
-	return buildArgs
+	if required {
+		buildArgs[aws.StringValue(s.Name)] = s.ImageConfig.Image.BuildConfig(contextDir)
+	}
+	for name, config := range s.Sidecars {
```
paragbhingre: I'm sure it is difficult to write only one function to do what we are doing in this `BuildArgs` throughout all our service types. I think we can separate out this side loop and put it in one function so that we reduce duplication. Let me know what do you think about it.

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -59,7 +59,7 @@ func New(cmd Cmd) CmdClient {
 // BuildArguments holds the arguments that can be passed while building a container.
 type BuildArguments struct {
 	URI        string            // Required. Location of ECR Repo. Used to generate image name in conjunction with tag.
-	Tags       []string          // Optional. List of tags to apply to the image besides "latest".
```
paragbhingre, line 62: Why did it change from optional to required? We still use git short commit id if user doesn't provide tag right?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -95,7 +95,12 @@ func convertSidecars(s map[string]*manifest.SidecarConfig, exposedPorts map[stri
 	sort.Strings(keys)
 	for _, name := range keys {
 		config := s[name]
-		imageURI, _ := config.ImageURI()
+		var imageURI string
+		if uri, hasLocation := config.ImageURI(); hasLocation {
+			imageURI = uri
+		} else {
+			imageURI = rc.PushedImages[name].URI()
+		}
```
paragbhingre: How about this?
```suggestion
		imageURI := rc.PushedImages[name].URI()
		if uri, hasLocation := config.ImageURI(); hasLocation {
			imageURI = uri
		} 
```

paragbhingre (commented):
I just have few nits and a question.

internal/pkg/manifest/backend_svc.go
KollaAdithya: Done. created a function `buildArgs` to reduce code duplication.

internal/pkg/docker/dockerengine/dockerengine.go
KollaAdithya: Previously if the client does not provide any `Tags` then docker build defaults to `latest`.  But now the logic is changed because we do not need to have`latest tag for sidecars`, so client should explicitly provide  `latest` tag or any custom tags for building and pushing image. This makes the `Tags` Required if that makes sense.

paragbhingre: If client doesn't provide tag then what happens? Do we error out in that situation?

internal/pkg/manifest/rd_web_svc.go
```diff
@@ -160,10 +155,8 @@ func (s *RequestDrivenWebService) ContainerPlatform() string {
 }
 
 // BuildArgs returns a docker.BuildArguments object given a context directory.
-func (s *RequestDrivenWebService) BuildArgs(contextDir string) map[string]*DockerBuildArgs {
-	buildArgs := make(map[string]*DockerBuildArgs, 1)
-	buildArgs[aws.StringValue(s.Name)] = s.ImageConfig.Image.BuildConfig(contextDir)
-	return buildArgs
+func (s *RequestDrivenWebService) BuildArgs(contextDir string) (map[string]*DockerBuildArgs, error) {
+	return buildArgs(contextDir, s.Name, s.ImageConfig.Image, nil)
```
paragbhingre: I like the way you wrote a separate function to reduce duplication, but the worry I have is sending nil values to that function. And another concern is that this function requires four parameters as input. That is why I think it is difficult to reduce duplication in these kinds of functions. One thing we can do is create a struct with all these input values and write the buildArgs method on it. Or just de-duplicate the sidecar loop and leave the rest as is.

internal/pkg/docker/dockerengine/dockerengine.go
KollaAdithya: If the tags are empty the current behavior is building all the images with default `latest` tag and not pushing any images to the ECR repository which is not desired and will result in a deployment failure as we do not any images pushed into ECR.

Good call 👍  to error out so that we do not even build and push images if the tags are empty.

internal/pkg/manifest/rd_web_svc.go
KollaAdithya: My concern is that if we de-duplicate only sidecar loop then we have two functions exactly doing the same thing one for main container which is `BuildArgs` and other function say `scBuildArgs` for all the workloads. Is that a good approach ?

KollaAdithya: updated my PR accordingly after offline discussion

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -290,3 +297,11 @@ func userHomeDirectory() string {
 
 	return home
 }
+
+type errEmptyImageTags struct {
+	uri string
+}
+
+func (e *errEmptyImageTags) Error() string {
+	return fmt.Sprintf("tags to reference an image should not be empty for building and pushing into ECR repository %s", e.uri)
```
paragbhingre:
```suggestion
	return fmt.Sprintf("tags to reference an image should not be empty for building and pushing into the ECR repository %s", e.uri)
```

paragbhingre (approved):
Looks good to me. Feel free to remove the label after addressing tiniest nit. 
Great Job Adi 🎉

PR #4561, opened 2023-02-28
chore(cli): implement Validate for `svc override`

efekarakus:

### DIFF
--- a/internal/pkg/cli/svc_override.go
+++ b/internal/pkg/cli/svc_override.go
@@ -102,8 +102,9 @@ func (o *overrideSvcOpts) validateAppName() error {
 	if o.appName == "" {
 		return errNoAppInWorkspace
 	}
-	if _, err := o.appConfig(); err != nil {
-		return err
+	_, err := o.cfgStore.GetApplication(o.appName)
+	if err != nil {
+		return fmt.Errorf("get application %q configuration: %w", o.appName, err)
 	}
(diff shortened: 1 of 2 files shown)


Lou1415926 (approved):
LGTM :shipit:

internal/pkg/cli/svc_override.go
```diff
@@ -73,6 +98,35 @@ func (o *overrideSvcOpts) RecommendActions() error {
...
+		if o.cdkLang == valid {
+			return nil
+		}
+	}
+	return fmt.Errorf("%q is not a valid CDK language: must be one of: %s",
+		o.cdkLang,
+		strings.Join(applyAll(validCDKLangs, strconv.Quote), ", "))
+}
+
+func (o *overrideSvcOpts) appConfig() (*config.Application, error) {
```
paragbhingre: Are we going to use this method somewhere else if not then just keeping return type as error would be fine I think. We do not need to return `config.Application` if we aren't planning to use it.

paragbhingre (commented):
Just one question, otherwise looks good to me.

internal/pkg/cli/svc_override.go
efekarakus: Sure, I removed the `appConfig` method and inlined it instead

paragbhingre (approved):

PR #4563, opened 2023-02-28
chore(progtrack): add rollback alarm names & states

huanjani:
These changes add a table of the rollback alarms associated with the deploying service and their states.

This is what it looks like mid-deployment:
Screenshot 2023-02-28 at 10 38 46 AM

This is what it looks like after deployment:
Screenshot 2023-02-28 at 1 58 19 PM

With multiple alarms:
Screenshot 2023-02-28 at 3 13 00 PM


internal/pkg/ecs/ecs.go
```diff
@@ -63,6 +68,7 @@ type ServiceDesc struct {
 type Client struct {
 	rgGetter       resourceGetter
 	ecsClient      ecsClient
+	cwClient       cwClient
```
Lou1415926: How is this client used?

internal/pkg/term/progress/status.go
```diff
@@ -71,6 +73,17 @@ func prettifyElapsedTime(sw *stopWatch) string {
...
 
+func prettifyAlarmState(state string) string {
+	pretty := fmt.Sprintf("[%s]", state)
+	switch state {
+	case alarmOKState:
+		pretty = color.Green.Sprintf("[%s]", state)
+	case inAlarmState:
+		pretty = color.Red.Sprintf("[%s]", state)
+	}
+	return pretty
```
Lou1415926: nit:
```suggestion
	switch state {
	case alarmOKState:
		return color.Green.Sprintf("[%s]", state)
	case inAlarmState:
		return color.Red.Sprintf("[%s]", state)
	default: 
	         return fmt.Sprintf("[%s]", state)
	}
```

Lou1415926 (commented):
One small question, otherwise LGTM!

internal/pkg/term/progress/status.go
huanjani: 😆 I forgot about default! 🤦🏼

internal/pkg/ecs/ecs.go
huanjani: removed!

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/aws/copilot-cli/internal/pkg/aws/cloudwatch"
```
efekarakus: nit: do you mind re-arranging all these imports, so that the `github.com/` ones appear together?

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -404,7 +411,7 @@ func (cf CloudFormation) changeRenderers(in changeRenderersInput) ([]progress.Re
 			}
 			renderer = r
 		case aws.StringValue(change.ResourceChange.ResourceType) == ecsServiceResourceType:
-			renderer = progress.ListeningECSServiceResourceRenderer(in.stackStreamer, cf.ecsClient, logicalID, description, progress.ECSServiceRendererOpts{
+			renderer = progress.ListeningECSServiceResourceRenderer(in.stackStreamer, cf.ecsClient, cf.cwClient, logicalID, description, progress.ECSServiceRendererOpts{
```
efekarakus: What do you think of refactoring to have two different types as input to the constructor?
```go
// ECSServiceRendererCfg holds configuration required to initialize an ECS service renderer.
type ECSServiceRendererCfg struct 

// ECSServiceRendererOpts holds optional configuration to stylize an ECS service renderer. 
type ECSServiceRendererOpts struct

func ListeningECSServiceResourceRenderer(cfg ECSServiceRendererCfg, opts ECSServiceRendererOpts)
```

internal/pkg/stream/ecs.go
```diff
@@ -5,6 +5,7 @@ package stream
 
 import (
 	"fmt"
+	"github.com/aws/copilot-cli/internal/pkg/aws/cloudwatch"
```
efekarakus: nit: same with the imports here, it's weird that the IDE doesn't do this properly 💭

internal/pkg/stream/ecs.go
```diff
@@ -123,6 +132,18 @@ func (s *ECSDeploymentStreamer) Fetch() (next time.Time, done bool, err error) {
 		return next, false, fmt.Errorf("fetch service description: %w", err)
 	}
 	s.retries = 0
```
efekarakus: hmm I wonder if the retries for `ecs` vs `cw` should be separate? (e.g. `ecsRetries` and `cwRetries`)

> DescribeServices:  Bucket refill rate (or Sustained rate) 20
> https://docs.aws.amazon.com/AmazonECS/latest/APIReference/request-throttling.html

> DescribeAlarms: 9 transactions per second (TPS) per Region. The maximum number of operation requests you can make per second without being throttled.
https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_limits.html

With the existing implementation, say the following sequence of events happens:
1. We start with `retries: 0`, ECS call succeeds but CW gives a throttling error.
2. We are now at `retries: 1`, ECS call succeeds so we **reset** retries and call CW which gives throttling error again.
3. Now `retries` is back to 1 instead of having 2 for CW.

efekarakus: The gist is right now, `nextFetchDate` might be returning an earlier date and we should be penalizing the client more

internal/pkg/stream/ecs.go
```diff
@@ -123,6 +132,18 @@ func (s *ECSDeploymentStreamer) Fetch() (next time.Time, done bool, err error) {
...
+	var alarms []cloudwatch.AlarmStatus
+	if out.DeploymentConfiguration != nil && out.DeploymentConfiguration.Alarms != nil {
```
efekarakus: Should we also be checking ` aws.BoolValue(out.DeploymentConfiguration.Alarms.Enable)`?

internal/pkg/term/progress/ecs.go
```diff
@@ -149,6 +158,27 @@ func (c *rollingUpdateComponent) renderFailureMsgs(out io.Writer) (numLines int,
...
+			prettifyAlarmState(a.Status),
+		})
+	}
+	table := newTableComponent(color.Faint.Sprintf("Alarms"), header, rows)
+	table.Padding = c.padding
+	nl, err := table.Render(out)
+	if err != nil {
+		return 0, fmt.Errorf("render alarms table: %w", err)
+	}
+	return nl, err
```
efekarakus: nit: can we do like the above method where we have create an empty line before the `Alarms` table?
```go
components := []Renderer {
  &singleLineCompoenent{}, // Add an empty line before renderering alarms.
  table,
}
return renderComponents(out, components)
```

efekarakus (commented):
aww yea 🌈

internal/pkg/stream/ecs.go
```diff
@@ -157,11 +168,27 @@ func (s *ECSDeploymentStreamer) Fetch() (next time.Time, done bool, err error) {
...
+		s.cwRetries = 0
+	}
+	
 	s.eventsToFlush = append(s.eventsToFlush, ECSService{
 		Deployments:         deployments,
 		LatestFailureEvents: failureMsgs,
+		Alarms:              alarms,
 	})
-	return nextFetchDate(s.clock, s.rand, s.retries), done, nil
+	return nextFetchDate(s.clock, s.rand, s.ecsRetries), done, nil
```
efekarakus: nit: This way folks might not be confused why this is only `ecsRetries` and not `cwRetries`
```suggestion
	return nextFetchDate(s.clock, s.rand, 0), done, nil
```

efekarakus (approved):

Lou1415926 (approved):

PR #4565, opened 2023-03-01
fix(template): disassociate abr when delete

iamhopaul123:
This PR fixes a bug that when users remove ABR from their manifests and run `svc deploy`, the alarms will not be disassociated from the services. Also, `svc deploy` to remove ABR will fail if the alarm is currently in `ALARM` state.

### DIFF
--- a/internal/pkg/aws/cloudformation/testdata/parse/lb-web-svc.yaml
+++ b/internal/pkg/aws/cloudformation/testdata/parse/lb-web-svc.yaml
@@ -140,6 +140,10 @@ Resources:
       DeploymentConfiguration:
         MinimumHealthyPercent: 100
         MaximumPercent: 200
+        Alarms:
+          AlarmNames: []
+          Enable: false
+          Rollback: true
       # This may need to be adjusted if the container takes a while to start up
       HealthCheckGracePeriodSeconds: 60
       LoadBalancers:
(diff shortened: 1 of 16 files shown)


internal/pkg/aws/cloudformation/testdata/parse/lb-web-svc.yaml
```diff
@@ -140,6 +140,10 @@ Resources:
       DeploymentConfiguration:
         MinimumHealthyPercent: 100
         MaximumPercent: 200
+        Alarms:
+          AlarmNames: []
+          Enable: false
+          Rollback: true
```
iamhopaul123, line 146: Note that all three are required field, and `UpdateService` won't pick up any changes if they are not explicitly specified.

huanjani (approved):
Thank you!

efekarakus (approved):
> Also, svc deploy to remove ABR will fail if the alarm is currently in ALARM state.

Is this an ECS API behavior? I'm surprised it works this way 💭 
What if you had a poorly configured alarm and you want to bypass it, how would the customer do that?

iamhopaul123:
> Is this an ECS API behavior? I'm surprised it works this way 💭
What if you had a poorly configured alarm and you want to bypass it, how would the customer do that?

No it's just because right now we don't disassociate the alarm, so when we trigger a new deployment and the alarm is in "ALARM", the stack update fails

efekarakus:
🤦 The comment in the PR body was meant to describe _today's_ behavior, not the behavior once the PR is merged :)

PR #4566, opened 2023-03-01
chore(cli): implement Ask for svc override

efekarakus:
Sample screenshot of the resources prompt:
Screenshot 2023-02-28 at 6 12 11 PM

### DIFF
--- a/internal/pkg/cli/svc_override.go
+++ b/internal/pkg/cli/svc_override.go
@@ -5,6 +5,7 @@ package cli
 
 import (
 	"fmt"
+	"io"
 	"strconv"
 	"strings"
 
@@ -13,6 +14,9 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/aws/identity"
 	"github.com/aws/copilot-cli/internal/pkg/aws/sessions"
 	"github.com/aws/copilot-cli/internal/pkg/config"
+	"github.com/aws/copilot-cli/internal/pkg/template"
+	"github.com/aws/copilot-cli/internal/pkg/term/prompt"
+	"github.com/aws/copilot-cli/internal/pkg/term/selector"
 	"github.com/aws/copilot-cli/internal/pkg/workspace"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
@@ -20,25 +24,39 @@ import (
 
 const (
 	// IaC options for overrides.
-	cdkIacToolkit = "cdk"
+	cdkIaCTool = "cdk"
 
 	// IaC toolkit configuration.
 	typescriptCDKLang = "typescript"
 )
 
-var iacToolkits = []string{
-	cdkIacToolkit,
+var validIaCTools = []string{
+	cdkIaCTool,
 }
 
 var validCDKLangs = []string{
 	typescriptCDKLang,
 }
 
+type stringWriteCloser interface {
+	fmt.Stringer
+	io.WriteCloser
+}
+
+type closableStringBuilder struct {
+	*strings.Builder
+}
+
+// Close implements the io.Closer interface for a strings.Builder and is a no-op.
+func (sb *closableStringBuilder) Close() error {
+	return nil
+}
+
 type overrideVars struct {
-	name    string
-	envName string
-	appName string
-	iacTool string
+	name      string
+	appName   string
+	iacTool   string
+	resources []template.CFNResource
 
 	// CDK override engine flags.
 	cdkLang string
@@ -48,9 +66,13 @@ type overrideSvcOpts struct {
 	overrideVars
 
 	// Interfaces to interact with dependencies.
-	ws       wsWlDirReader
-	fs       afero.Fs
-	cfgStore store
+	ws         wsWlDirReader
+	fs         afero.Fs
+	cfgStore   store
+	prompt     prompter
+	wsPrompt   wsSelector
+	cfnPrompt  cfnSelector
+	packageCmd func(w stringWriteCloser) (executor, error)
 }
 
 func newOverrideSvcOpts(vars overrideVars) (*overrideSvcOpts, error) {
@@ -67,12 +89,18 @@ func newOverrideSvcOpts(vars overrideVars) (*overrideSvcOpts, error) {
 	}
 	cfgStore := config.NewSSMStore(identity.New(defaultSess), ssm.New(defaultSess), aws.StringValue(defaultSess.Config.Region))
 
-	return &overrideSvcOpts{
+	prompt := prompt.New()
+	cmd := &overrideSvcOpts{
 		overrideVars: vars,
 		ws:           ws,
 		fs:           fs,
 		cfgStore:     cfgStore,
-	}, nil
+		prompt:       prompt,
+		wsPrompt:     selector.NewLocalWorkloadSelector(prompt, cfgStore, ws),
+		cfnPrompt:    selector.NewCFNSelector(prompt),
+	}
+	cmd.packageCmd = cmd.newSvcPackageCmd
+	return cmd, nil
 }
 
 // Validate returns an error for any invalid optional flags.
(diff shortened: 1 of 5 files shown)


internal/pkg/cli/svc_override.go
```diff
@@ -120,6 +154,109 @@ func (o *overrideSvcOpts) validateCDKLang() error {
...
+		appName: o.appName,
+	})
+	if err != nil {
+		return nil, err
+	}
+	cmd.templateWriter = tplBuf
+	return cmd, nil
+}
+
+func (o *overrideSvcOpts) askIaCTool() error {
```
huanjani, line 246: Design/UX-wise, is it better to acknowledge that there is currently only one valid IaCTool? (Like how we surface that we found only one env/workload/etc.) In other words, is it better to do this-- lay the groundwork for eventual multiple choices and ask them to select from one, or is it better to avoid seeming like it could be an oversight/robotic automation?

internal/pkg/cli/svc_override.go
```diff
@@ -120,6 +154,109 @@ func (o *overrideSvcOpts) validateCDKLang() error {
...
+	}
+	pkgCmd, err := o.packageCmd(buf)
+	if err != nil {
+		return err
+	}
+	if err := pkgCmd.Execute(); err != nil {
+		return fmt.Errorf("generate CloudFormation template for service %q: %v", o.name, err)
+	}
+	msg := fmt.Sprintf("Which resources in %q would you like to override?", o.name)
+	resources, err := o.cfnPrompt.Resources(msg, "Resources:", "", buf.String())
```
huanjani, line 218: This is very cool. The 'help' text is great.

internal/pkg/cli/svc_override.go
efekarakus: Yea that totally makes sense, we should just default to the one option.
Honestly, I thought that `prompt.MultiSelect` and `prompt.SelectOne` already did that 😂 

What do you think if I update in a separate PR the `prompt` pkg's functions to immediately return the one option if the slice contains one element?

huanjani (approved):
:shipit:

internal/pkg/cli/svc_override.go
```diff
@@ -13,32 +14,49 @@ import (
...
+}
+
+// Close implements the io.Closer interface for a strings.Builder and is a no-op.
+func (sb *closableStringBuilder) Close() error {
+	return nil
+}
+
 type overrideVars struct {
-	name    string
-	envName string
```
bvtujo, line 39: We don't need envName anymore?

internal/pkg/cli/svc_override.go
```diff
@@ -120,6 +154,109 @@ func (o *overrideSvcOpts) validateCDKLang() error {
...
+	envs, err := o.cfgStore.ListEnvironments(o.appName)
+	if err != nil {
+		return nil, fmt.Errorf("list environments in application %q: %v", o.appName, err)
+	}
+	if len(envs) == 0 {
+		return nil, fmt.Errorf("no environments found in application %q", o.appName)
+	}
+	cmd, err := newPackageSvcOpts(packageSvcVars{
+		name:    o.name,
+		envName: envs[0].Name, // Use any environment to get a sample CloudFormation template for the service.
```
bvtujo, line 236: What happens if there are resources included in one environment but not another (env overrides)? for example, if managed EFS is defined for a service in one environment but not the other, the CF template will be different: we have a conditional `{{- if .Storage.ManagedVolumeInfo}}` inclusion for the Access Point. There are probably other resources like this, but this is the first one that comes to mind.

efekarakus: oh yea great question!

`svc override` is "best effort" so the idea is we'll just scaffold some resources for you to make it easy to get started but you could skip this command altogether and write the files yourself manually.

For example, to be able to do the example you wrote above, in the CDK app's `stack.ts` the client would write:

```ts
if props.envName === "prod" {
   this.transformEFS();
}
```

internal/pkg/cli/svc_override.go
efekarakus: Yeah, I guess this is related to your other comment :) 
I wanted to provide a simpler experience if possible with less prompts. The goal is to show _some_ useful resources so that the user can get started with writing CDK code.

If for some reason the resource isn't available in other environments, the `copilot deploy` or the `Build` stage in the pipeline will fail 👍 so the user should get an early signal that they have to update the CDK code with an if-statement

internal/pkg/cli/svc_override.go
iamhopaul123: +1 to Austin's question. Maybe we should take env as an input?

bvtujo: Yeah, I think I like adding env as an input here. That lets the customer get the correct, expected template for the given inputs and there's no guesswork. If there's only one environment, then there will be no selector.

internal/pkg/cli/svc_override.go
bvtujo: hmm, I'm not sure I agree that a command completing successfully locally with one fewer prompt, then failing in the build stage, is a better experience than having one more prompt for the command and reducing the chance of build failure. I think the extra choice here is better than the confusion the failure would create 🤔

internal/pkg/cli/svc_override.go
efekarakus: But isn't the end outcome the same? For example, as a user I choose "prod" and the resource is available there but then missing in the "test" environment. That doesn't really help me, I have to write the if-statement either way when running `copilot deploy --env test`.

I am down to add the `--env` flag as optional, but I'm still not in favor of prompting here.

iamhopaul123: Isn't the final manifest itself determined by the environment after applying the env specific setting? If we don't know what environment they are trying to override the service against, how do we generate the correct CFN template and get all resources 🤔

bvtujo: hmm, I think a lot depends on the mechanics of how overrides are implemented. I wonder if there is a way to help customers understand the differences in templates between their environments so they can write correct CDK code. Visibility into this would be useful. Maybe providing a oneliner to "svc package > a.yml && svc package > b.yml && diff a.yml b.yml". Or a guide. Otherwise it's a little like a shot in the dark which resources are present in which envs.

bvtujo: Fact is this is a high context command which requires a lot of understanding, so perhaps we can provide plenty of education and insight into how someone might use it. That's for the site and the blog, though, and shouldn't block us here.

efekarakus: If the customer is indeed doing CloudFormation's best practice: 
> Reuse templates to replicate stacks in multiple environments
> https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/best-practices.html#reuse

Then the environment does not matter. This is why I'm not prompting for the environment.

If the user is using `environments` in the manifest to toggle resources, then even if we were to prompt for the environment to show the "right" resources the `package` command will fail while deploying in other environments.
So is it really important to show the appropriate environment resources, I'm still not convinced?

I'd rather optimize for the case where `copilot package` would always output the same resources across environments.

bvtujo: 🧐 I see, so this edge case is _possible_ but not _likely_, nor is it best practice. I wonder if there's a natural place in the docs to advise against toggling resources with overrides, or if this is a niche enough problem to just disregard it. 

I'm convinced by this; otherwise the pr looks good! 

:shipit:

bvtujo (approved):
🚢

internal/pkg/cli/svc_override.go
efekarakus: I'll add the flag as optional, so that folks have a way of seeing the resource that they want :) but I just don't think that different templates/env is a common enough scenario to warrant making `--env` _required_.

efekarakus: Created the PR here to add the optional flag: https://github.com/aws/copilot-cli/pull/4573 !

Issue #4447, opened 2023-02-02
Add env_file support to sidecars


carlosaherrera:
This configuration works as expected. The variables declared in both the `test` and `sidecars` sections get passed to the containers.
```yml
# You can override any of the values defined above by environment.
environments:
  # Although not mentioned in documentation, configuring sidecar environment-specific variables is supported.
  # ref: https://github.com/aws/copilot-cli/issues/3674#issuecomment-1157084196
  test:
    variables:
      SOME_VARIABLE1: 'somevalue1'
      SOME_VARIABLE2: 'somevalue2'
    sidecars:
      mysidecar1:
        variables:
          SOME_VARIABLE1: 'somevalue1'
          SOME_VARIABLE2: 'somevalue2'
```
But it's rather redundant. Especially when I need to pass the same variables to all containers. I thought I could be more succinct by utilizing the `env_file` param like so
```yml
# You can override any of the values defined above by environment.
environments:
  # Although not mentioned in documentation, configuring sidecar environment-specific variables is supported.
  # ref: https://github.com/aws/copilot-cli/issues/3674#issuecomment-1157084196
  test:
    env_file: env/.test.env
    sidecars:
      mysidecar1:
        env_file: env/.test.env
```
The main service container gets passed the environment variables just fine. Good. But the sidecars don't. So it looks like I have to resort to the former approach.

KollaAdithya:
Hello @carlosaherrera !
Your feature request makes sense to me.

efekarakus:
Hi folks!!

I'm really excited to say that AWS Copilot v1.27 is out and you can now build and push local `.env` files! 🚀 

* Guide: https://aws.github.io/copilot-cli/docs/developing/sidecars/#envFile
* Blog post: https://aws.github.io/copilot-cli/blogs/release-v127/#upload-local-environment-files-for-sidecar-containers
* Release notes: https://github.com/aws/copilot-cli/releases/tag/v1.27.0

I'll close the issue now! If you have any questions, feedback, or suggestions please feel free to open new GitHub issues 🥳

PR #4567, opened 2023-03-01
feat: add env file support for all sidecar containers

bvtujo:
- chore: add Env File field to manifest and add mft validation
- chore: change how env files are handled in parameters
- chore: change how manifest package handles env files
- chore: update templates to include required policies and Env File keys
- chore: add unit tests for cli/deploy package
- chore: fix cli deploy fields
- test: fix stack integ tests

This PR adds support for env files in all customer-defined sidecars, as well as the managed 
Logging sidecar pattern. 

It does this by creating a new parameter for each sidecar, ${sidecarName}EnvFileARN, and 
the corresponding condition. It uses that condition to modify the execution role policy 
so that the ECS service has access to all env files used in the task. 

The feature uploads each env file only once, allowing customers to specify the same env file
across multiple containers and not waste any space in S3 or bandwidth uploading duplicates.

Addresses #4447.


internal/pkg/manifest/backend_svc.go
```diff
@@ -126,9 +126,20 @@ func (s *BackendService) BuildArgs(contextDir string) map[string]*DockerBuildArg
...
 }
 
-// EnvFile returns the location of the env file against the ws root directory.
-func (s *BackendService) EnvFile() string {
-	return aws.StringValue(s.TaskConfig.EnvFile)
+// EnvFiles returns the locations of all env files against the ws root directory.
+// This method returns a map[string]string where the keys are container names
+// and the values are either env file paths or empty strings.
+func (s *BackendService) EnvFiles() map[string]string {
+	envFiles := make(map[string]string)
```
paragbhingre: How about this?
plus 2 for main container as well as FirelensContainer. 
```suggestion
	envFiles := make(map[string]string, len(s.Sidecars)+2)
```

internal/pkg/manifest/job.go
```diff
@@ -148,9 +148,20 @@ func (j *ScheduledJob) BuildRequired() (bool, error) {
...
 }
 
-// EnvFile returns the location of the env file against the ws root directory.
-func (j *ScheduledJob) EnvFile() string {
-	return aws.StringValue(j.TaskConfig.EnvFile)
+// EnvFiles returns the locations of all env files against the ws root directory.
+// This method returns a map[string]string where the keys are container names
+// and the values are either env file paths or empty strings.
+func (j *ScheduledJob) EnvFiles() map[string]string {
+	envFiles := make(map[string]string)
```
paragbhingre: Samesies as above

paragbhingre (commented):
Just a nit, otherwise first pass at it looks good to me.

internal/pkg/manifest/backend_svc.go
bvtujo: What's the behavior of a map initialized with a larger length if it's not actually filled out? 

Like, if I'm iterating over it later am I going to have two iterations of my for loop even if there's not actually an env file for the main container or logging sidecar?

huanjani: The length is "approximate" anyway, and more space will be created if necessary. No, it shouldn't iterate over empty spaces. https://go.dev/ref/spec#Making_slices_maps_and_channels

bvtujo: Shipped.

bvtujo: Upon further review, this change represented a magic number with little documentation in the code, and has been removed in favor of readability. The ruling on the field is overturned.

internal/pkg/manifest/backend_svc.go
```diff
@@ -126,9 +126,20 @@ func (s *BackendService) BuildArgs(contextDir string) map[string]*DockerBuildArg
 	return buildArgs
 }
 
-// EnvFile returns the location of the env file against the ws root directory.
-func (s *BackendService) EnvFile() string {
-	return aws.StringValue(s.TaskConfig.EnvFile)
+// EnvFiles returns the locations of all env files against the ws root directory.
+// This method returns a map[string]string where the keys are container names
+// and the values are either env file paths or empty strings.
+func (s *BackendService) EnvFiles() map[string]string {
```
iamhopaul123, line 134: Looks like the logic for `EnvFiles` funcs are almost identical. Can we possibly reduce the duplication?

e2e/sidecars/sidecars_test.go
```diff
@@ -54,7 +55,9 @@ sidecars:
     image: %s    # Image URL for sidecar container.
     variables:
       NGINX_PORT: %s
+    env_file: ./magic.env
 logging:
+  env_file: ./magic.env
```
iamhopaul123, line 60: Do we need to validate the env files are indeed added to the service?

bvtujo: yeah, good call

bvtujo: Done. I settled for writing some logic that checks that there is in fact an env file included with each container. Taken along with the fact that the service won't run if the permissions are wrong or the env file can't be found or uploaded, this should be sufficient.

huanjani (approved):
:shipit:

iamhopaul123 (approved):
LGTM. Maybe next time make a smaller PR so that it's safer and easier to review 😅

bvtujo:
Adding a label until I can verify that the reason the sidecars e2e is failing locally at the last step (checking for svc logs) has nothing to do with the changes here.

PR #4568, opened 2023-03-01
chore: assign priority to multiple listener rules

paragbhingre:
This PR fixes assigning the priority to multiple listener rules in the service. But it does not solve the problem of assigning priorities to two different services being deployed at once. That case has to be taken care in the future. 


cf-custom-resources/lib/alb-rule-priority-generator.js
```diff
@@ -158,22 +158,41 @@ const getListenerRules = async function (listenerArn) {
...
+          } else {
+            responseData.Priority = await calculateNextRulePriority(
+                event.ResourceProperties.ListenerArn
+            );
+          }
         } else {
-          responseData.Priority = await calculateNextRulePriority(
-            event.ResourceProperties.ListenerArn
-          );
+          for (let i=0; i < event.ResourceProperties.RulePath.length; i++){
```
huanjani, line 168: Is it typical to increment/decrement by 1? If more listener rules are added later, is it possible for the priority to be between existing ones?

cf-custom-resources/lib/alb-rule-priority-generator.js
```diff
@@ -158,22 +158,41 @@ const getListenerRules = async function (listenerArn) {
...
+      event.PhysicalResourceId || `alb-rule-priority-${event.LogicalResourceId}`;
   try {
     switch (event.RequestType) {
       case "Create":
       case "Update":
-        if (isRootPath) {
-          responseData.Priority = await calculateNextRootRulePriority(
-            event.ResourceProperties.ListenerArn
-          );
+        if (event.ResourceProperties.RulePath.length == 1){
```
iamhopaul123: Why do we take `event.ResourceProperties.RulePath.length == 1` as a special case? Or will we combine them later?

iamhopaul123:
Also can we change the title and commit to chore? It is not fixing a bug i guess

cf-custom-resources/lib/alb-rule-priority-generator.js
paragbhingre: I was thinking to keep one port vs multiple port implementation separate. I thought of combining them into one and then we will have this following logic -

```

for (let i=0; i < event.ResourceProperties.RulePath.length; i++){
            if (event.ResourceProperties.RulePath[i] === "/") {
              if (nextRootRuleNumber == null){
                nextRootRuleNumber = await calculateNextRootRulePriority(
                    event.ResourceProperties.ListenerArn
                );
              }
              responseData["Priority"+i]  = nextRootRuleNumber--;
            } else {
              if (nextNonRootRuleNumber == null) {
                nextNonRootRuleNumber = await calculateNextRulePriority(
                    event.ResourceProperties.ListenerArn
                );
              }
              responseData["Priority"+i]  = nextNonRootRuleNumber++;
            }
          }
```

In this case we return `Prioirty0`, `Prioirty1`, `Prioirty2` and so on. But currently we only return `Priority` without any postfix.  But I think as per our offline discussion with Wx also whenever customer will upgrade to new version cfn template will also expect `Priority{{ $i }}` instead of just `Priority` and that shouldn't replace listener rule all together. 


But as you said for this PR I think we need to keep it as I have implemented because we do not have Priority{{ $i }} logic implemented yet. So removing event.ResourceProperties.RulePath.length == 1 will result into errors.

paragbhingre: But as you said for this PR I think we need to keep it as I have implemented because we do not have `Priority{{ $i }}` logic implemented yet. So removing `event.ResourceProperties.RulePath.length == 1` will result into errors.

cf-custom-resources/lib/alb-rule-priority-generator.js
paragbhingre: Good question - if you look at this existing [test case](https://github.com/aws/copilot-cli/blob/5c0124b09eeb99e7d75b8fa6032b06653ddbe57c/cf-custom-resources/test/alb-rule-priority-generator-test.js#L232) which has priority 3 and 5 assigned. But whenever new listener rule comes in, it assigns priority as 6 and not 4 which is between 3-5 and also available. I think we do it because additional rules are for more specific route paths which should get the most highest priority hence we do what we are doing here.

cf-custom-resources/lib/alb-rule-priority-generator.js
iamhopaul123: Can we still return Prioirty, Prioirty1, Prioirty2 if we add an "if statement" before `responseData["Priority"+i]  = nextNonRootRuleNumber++;`?

paragbhingre: I can do that but I think I need to do the same for `responseData["Priority"+i]  = nextRootRuleNumber--;` as well.

huanjani (approved):
lgtm

internal/pkg/deploy/cloudformation/stack/testdata/workloads/svc-staging.stack.yml
```diff
@@ -427,7 +429,7 @@ Resources: # If a bucket URL is specified, that means the template exists.
     Type: Custom::RulePriorityFunction
     Properties:
       ServiceToken: !GetAtt RulePriorityFunction.Arn
-      RulePath: !Ref RulePath
```
iamhopaul123, line 434: Can we not use `!Select [ 0, !Ref RulePath ]` to replace `RulePath` entirely?

paragbhingre: We can not because `RulePath` is not a list at the moment. And if we make it a list then there are multiple places where we will need to make change which is anyway going away with the upcoming multiple port PR.

iamhopaul123 (approved):

PR #4569, opened 2023-03-01
chore: Add labels to docker build call

KollaAdithya:
This PR should be reviewed and  merged only after #4547 
This PR adds labels to set metadata of an image. Labels added are
1. `com.aws.copilot.image.builder=copilot-cli`
2. `com.aws.copilot.image.container.name=Name Of the container`
3. `com.aws.copilot.image.version= copilot version`

This PR also address

I missed this scenario to test in PR #4547 where manifest contains both main and sidecar containers `image.build` which is giving following error when performing docker inspect 

`Error: No such object: 535307839156.dkr.ecr.us-west-2.amazonaws.com/gang/jagg`

With this PR, we correctly reference the image with tag when performing `docker inspect`. when performing docker inspect we use `uri:tag`.

For Main container we call ` docker inspect --format '{{json (index .RepoDigests 0)}}' uri:latest`
For Sidecar container images we call `docker inspect --format '{{json (index .RepoDigests 0)}}' uri:<sidecarname>-latest`

### DIFF
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -50,6 +50,12 @@ const (
 	imageTagLatest = "latest"
 )
 
+const (
+	labelForBuilder       = "com.aws.copilot.image.builder"
+	labelForVersion       = "com.aws.copilot.image.version"
+	labelForContainerName = "com.aws.copilot.image.container.name"
+)
+
 // ActionRecommender contains methods that output action recommendation.
 type ActionRecommender interface {
 	RecommendedActions() []string
@@ -363,13 +369,6 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
 	if err != nil {
 		return nil, fmt.Errorf("check if manifest requires building from local Dockerfile: %w", err)
 	}
-	// Add labels to ECR images with docker build call
-	labels := map[string]string{
-		"builder": "copilot-cli",
-	}
-	if version.Version != "" {
-		labels["version"] = version.Version
-	}
 	dArgs := make(map[string]*dockerengine.BuildArguments, len(argsPerContainer))
 	for container, buildArgs := range argsPerContainer {
 		tags := []string{imageTagLatest}
@@ -382,16 +381,21 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
 				tags = append(tags, fmt.Sprintf("%s-%s", container, img.GitShortCommitTag))
 			}
 		}
+		labels := make(map[string]string, 3)
+		labels[labelForBuilder] = "copilot-cli"
+		if version.Version != "" {
+			labels[labelForVersion] = version.Version
+		}
+		labels[labelForContainerName] = container
 		dArgs[container] = &dockerengine.BuildArguments{
-			Dockerfile:    aws.StringValue(buildArgs.Dockerfile),
-			Context:       aws.StringValue(buildArgs.Context),
-			Args:          buildArgs.Args,
-			CacheFrom:     buildArgs.CacheFrom,
-			Target:        aws.StringValue(buildArgs.Target),
-			Platform:      mf.ContainerPlatform(),
-			Tags:          tags,
-			ContainerName: container,
-			Labels:        labels,
+			Dockerfile: aws.StringValue(buildArgs.Dockerfile),
+			Context:    aws.StringValue(buildArgs.Context),
+			Args:       buildArgs.Args,
+			CacheFrom:  buildArgs.CacheFrom,
+			Target:     aws.StringValue(buildArgs.Target),
+			Platform:   mf.ContainerPlatform(),
+			Tags:       tags,
+			Labels:     labels,
 		}
 	}
 	return dArgs, nil
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -373,6 +373,7 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
 		return nil, fmt.Errorf("check if manifest requires building from local Dockerfile: %w", err)
 	}
 	dArgs := make(map[string]*dockerengine.BuildArguments, len(argsPerContainer))
+	tags := []string{imageTagLatest, img.Tag()}
 	for container, buildArgs := range argsPerContainer {
 		tags := []string{imageTagLatest}
 		if img.Tag() != "" {
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -36,6 +36,7 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/term/color"
 	"github.com/aws/copilot-cli/internal/pkg/term/log"
 	termprogress "github.com/aws/copilot-cli/internal/pkg/term/progress"
+	"github.com/aws/copilot-cli/internal/pkg/version"
 	"github.com/aws/copilot-cli/internal/pkg/workspace"
 	"github.com/spf13/afero"
 )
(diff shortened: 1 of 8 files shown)


internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -119,6 +121,14 @@ func (c CmdClient) Build(in *BuildArguments) error {
 		args = append(args, "--build-arg", fmt.Sprintf("%s=%s", k, in.Args[k]))
 	}
 
+	if in.ContainerName != "" {
+		in.Labels["container"] = in.ContainerName
+	}
+	// Add Labels to docker build call.
+	for k, v := range in.Labels {
+		args = append(args, "--label", fmt.Sprintf("%s=%s", k, v))
```
huanjani: Do you think we need to show the user the `docker build --label` command that is being run, like we do if we are using the `--platform` flag? I don't have a preference, just a thought.

KollaAdithya: we do show labels to user in the `docker build command` . 
Here is what command line looks like
`Building your container image: docker build -t 1234679.dkr.ecr.us-west-2.amazonaws.com/vote/results -t 12345679.dkr.ecr.us-west-2.amazonaws.com/vote/results  --platform linux/x86_64 /Users/kolladit/demo/backend -f /Users/kolladit/demo/backend/Dockerfile`

KollaAdithya: we do show labels to user in the docker build command .

Here is what command line looks like
`Building your container image: docker build -t 1234679.dkr.ecr.us-west-2.amazonaws.com/vote/results:nginx-latest  --platform linux/x86_64 --label builder=copilot-cli --label version=v1.26.0 --label container=nginx-latest /Users/kolladit/demo/backend -f /Users/kolladit/demo/backend/Dockerfile`

huanjani: Isn't that only `if in.Platform != ""`? ((New) line 134 below)

KollaAdithya: yes that's right. So do we need to show the user even if the platform =="" ?

huanjani: I don't have a strong opinion. Do you think it would be useful for the user?

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -125,6 +127,14 @@ func (c CmdClient) Build(in *BuildArguments) error {
 		args = append(args, "--build-arg", fmt.Sprintf("%s=%s", k, in.Args[k]))
 	}
 
+	if in.ContainerName != "" {
```
Lou1415926: Is there a reason why you prefer setting the container label here instead if in `cli/deploy/workload.go`?

KollaAdithya: The reason for setting the `container label` in `dockerengine.go` because 

I declared labels as `map[string]string`, but in [`cli/deploy/workload.go`](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/cli/deploy/workload.go#L366) we have a `for loop` to populate the `dockerengine.BuildArguments` 

if I do `labels["container"]= containerName` then the labels["container"] will always have the name of last Container. 
Is there any alternative that I can use?

Lou1415926: Maybe move the declaration of `labels` inside of the for loop?

internal/pkg/cli/deploy/workload.go
```diff
@@ -362,6 +363,13 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
 	if err != nil {
 		return nil, fmt.Errorf("check if manifest requires building from local Dockerfile: %w", err)
 	}
+	// Add labels to ECR images with docker build call
+	labels := map[string]string{
+		"builder": "copilot-cli",
```
efekarakus: see "Docker label / OCI image annotation metadata types" in https://github.com/opencontainers/image-spec/blob/main/annotations.md

Perhaps, we should consider our labels to be of the format: `aws.copilot.image.version`, `aws.copilot.image.builder`, `aws.copilot.image.container.name`.
I wonder if @butlerj would have an opinion on this. Would you mind reaching out to him?

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -125,6 +127,14 @@ func (c CmdClient) Build(in *BuildArguments) error {
 		args = append(args, "--build-arg", fmt.Sprintf("%s=%s", k, in.Args[k]))
 	}
```
efekarakus, line 128: I wonder if we should be creating a copy of `in.Labels` here so that we don't touch the set of labels given by a consumer package:
```go
labels := make(map[string]string, len(in.Labels))
for k, v := range in.Labels {
   labels[k] = v
}
```

internal/pkg/cli/deploy/workload.go
jlbutler: Agreed on the reverse domain naming. You might consider using `com.aws.copilot` even.

On the labels vs annotations, you can do either. I might suggest that using annotations (stored in the manifest) might be a bit more portable across tools. But I think any tool supporting the use of annotations could potentially support labels as well (which are stored in the config blob).

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -125,6 +126,17 @@ func (c CmdClient) Build(in *BuildArguments) error {
...
 
+	// copy of Labels from BuildArguments.
+	labels := make(map[string]string, len(in.Labels))
+	for k, v := range in.Labels {
+		labels[k] = v
+	}
```
efekarakus: no need to copy now that you're no longer modifying the map :)

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -167,7 +179,7 @@ func (c CmdClient) Push(uri string, tags ...string) (digest string, err error) {
 		}
 	}
 	buf := new(strings.Builder)
-	if err := c.runner.Run("docker", []string{"inspect", "--format", "'{{json (index .RepoDigests 0)}}'", uri}, exec.Stdout(buf)); err != nil {
+	if err := c.runner.Run("docker", []string{"inspect", "--format", "'{{json (index .RepoDigests 0)}}'", imageName(uri, tags[0])}, exec.Stdout(buf)); err != nil {
```
efekarakus, line 186: can you add a comment explaining why we're indexing to 0? For example:
```
// The container image will have the same digest regardless of the associated tag.
// Pick the first tag and get the image's digest.
```

efekarakus (approved):
looks good! :shipit: with 2 small nits. 

Feel free to remove the label after addressing them.

internal/pkg/docker/dockerengine/dockerengine.go
iamhopaul123: Why do we add the first tag to the URI instead of previous `uri`?

KollaAdithya: Previously even the sidecar images are also build with `latest` tag but we do not `push the latest tag` of sidecar images to ECR. In this way we get the Image digest with out referencing with the `tag`. i.e just using uri.  This is the change where I removed the `building latest  tag for all images` [link](https://github.com/aws/copilot-cli/pull/4547/files#diff-245a109ace8ab38c02069b4557e8256a2fcf479ab23323d871df8357ecce6b51L86)

After addressing some of the feedback from Parag in the previous PR #4547  that we do not need to build the sidecar images with `latest` tag. I changed the Implementation and it is fault from my end that I did not test the functionality of sidecar build.

With this change we know exactly reference the image with the help of `uri:tag`   to get image digest. Using the first tag because in case of main container `tgas[0]  will always be latest` and for sidecar images `tags[0] will be <sidecarName>-latest`.

iamhopaul123: > Using the first tag because in case of main container tgas[0]  will always be latest and for sidecar images tags[0] will be <sidecarName>-latest.

I see gotcha. maybe add this to the comment as well? Sorry for not reading the description of the PR

```
For Main container we call  docker inspect --format '{{json (index .RepoDigests 0)}}' uri:latest
For Sidecar container images we call docker inspect --format '{{json (index .RepoDigests 0)}}' uri:<sidecarname>-latest
```

iamhopaul123 (approved):

PR #4572, opened 2023-03-03
chore(cli): implement Execute for `svc override`

efekarakus:

### DIFF
--- a/internal/pkg/cli/flag.go
+++ b/internal/pkg/cli/flag.go
@@ -129,8 +129,9 @@ const (
 	inputFilePathFlag = "cli-input-yaml"
 
 	// Flags for overriding templates.
-	iacToolFlag     = "tool"
-	cdkLanguageFlag = "cdk-language"
+	iacToolFlag       = "tool"
+	cdkLanguageFlag   = "cdk-language"
+	skipResourcesFlag = "skip-resources"
 
 	// Other.
 	svcPortFlag             = "port"
@@ -228,8 +229,8 @@ Must be one of: %s.`, strings.Join(applyAll(validIaCTools, strconv.Quote), ", ")
 	cdkLanguageFlagDescription = `Optional. The Cloud Development Kit language.`
 	overrideEnvFlagDescription = `Optional. Name of the environment to use when retrieving resources in a template.
 Defaults to a random environment.`
-
-	repoURLFlagDescription = fmt.Sprintf(`The repository URL to trigger your pipeline.
+	skipResourcesFlagDescription = `Optional. Skip asking for which resources to override and generate empty patch files.`
+	repoURLFlagDescription       = fmt.Sprintf(`The repository URL to trigger your pipeline.
 Supported providers are: %s.`, strings.Join(manifest.PipelineProviders, ", "))
 
 	ingressTypeFlagDescription = fmt.Sprintf(`Required for a Request-Driven Web Service. Allowed source of traffic to your service.
--- a/internal/pkg/cli/flag.go
+++ b/internal/pkg/cli/flag.go
@@ -229,8 +229,9 @@ Must be one of: %s.`, strings.Join(applyAll(validIaCTools, strconv.Quote), ", ")
 	cdkLanguageFlagDescription = `Optional. The Cloud Development Kit language.`
 	overrideEnvFlagDescription = `Optional. Name of the environment to use when retrieving resources in a template.
 Defaults to a random environment.`
-	skipResourcesFlagDescription = `Optional. Skip asking for which resources to override and generate empty patch files.`
-	repoURLFlagDescription       = fmt.Sprintf(`The repository URL to trigger your pipeline.
+	skipResourcesFlagDescription = `Optional. Skip asking for which resources to override and generate empty IaC extension files.`
+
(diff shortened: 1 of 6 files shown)


internal/pkg/cli/flag.go
```diff
@@ -228,8 +229,8 @@ Must be one of: %s.`, strings.Join(applyAll(validIaCTools, strconv.Quote), ", ")
 	cdkLanguageFlagDescription = `Optional. The Cloud Development Kit language.`
 	overrideEnvFlagDescription = `Optional. Name of the environment to use when retrieving resources in a template.
 Defaults to a random environment.`
-
-	repoURLFlagDescription = fmt.Sprintf(`The repository URL to trigger your pipeline.
+	skipResourcesFlagDescription = `Optional. Skip asking for which resources to override and generate empty patch files.`
```
paragbhingre: just a dumb question: do you think after we introduce yaml patch feature this `patch` word here would be little confusing? I am not sure about it, but in first glance I got confused if I am reviewing override feature or yaml patch. It could be just me so feel free to ignore if patch is the right word that should be used here.

internal/pkg/override/cdk.go
```diff
@@ -140,11 +139,20 @@ func (cdk *CDK) cleanUp(body []byte) ([]byte, error) {
...
 // ScaffoldWithCDK bootstraps a CDK application under dir/ to override the seed CloudFormation resources.
+// If the directory is not empty, then returns an error.
 func ScaffoldWithCDK(fs afero.Fs, dir string, seeds []template.CFNResource) error {
+	// Ignore errors on purpose. IsEmpty returns false if the directory does not exist, hence we should only check
+	// if the dir is empty if it also exists.
+	exists, _ := afero.Exists(fs, dir)
+	isEmpty, _ := afero.IsEmpty(fs, dir)
+	if exists && !isEmpty {
+		return fmt.Errorf("directory %q is not empty", dir)
+	}
```
dannyrandall, line 150:
```suggestion
	if isEmpty, _ := afero.IsEmpty(fs, dir); !isEmpty {
		return fmt.Errorf("directory %q is not empty", dir)
	}
```
does `exists` even matter?

internal/pkg/override/cdk.go
```diff
@@ -140,11 +139,20 @@ func (cdk *CDK) cleanUp(body []byte) ([]byte, error) {
 }
 
 // ScaffoldWithCDK bootstraps a CDK application under dir/ to override the seed CloudFormation resources.
+// If the directory is not empty, then returns an error.
 func ScaffoldWithCDK(fs afero.Fs, dir string, seeds []template.CFNResource) error {
+	// Ignore errors on purpose. IsEmpty returns false if the directory does not exist, hence we should only check
+	// if the dir is empty if it also exists.
+	exists, _ := afero.Exists(fs, dir)
+	isEmpty, _ := afero.IsEmpty(fs, dir)
```
paragbhingre, line 147: I agree to the condition that we are trying to test here but is ignoring errors here is safe? Are we sure that there wouldn't be a case where some genuine error or panic won't occur or won't cause us to fail?

internal/pkg/override/cdk.go
dannyrandall: actually maybe i'm misunderstanding because of the error. should there be a different error when the directory doesn't exist? like `directory doesn't exist`? or should that not even be an error case?

dannyrandall: oh nvm. i understand the comment now lol. good work 👍

dannyrandall (approved):

internal/pkg/override/cdk.go
efekarakus: - If the directory does not exist -> 👌 it's all good . We will create the directory for them and write files.
- If the directory exists and is empty -> 👌 it's all good. We will write the files to the directory.
- If the directory exists but is not empty -> 💥bad. We don't want to overwrite any files that they have in the directory.

Sadly, the following does not work:
```
isEmpty, _ := afero.IsEmpty(fs, dir); !isEmpty
```
That's what I had initially, but you can see the implementation here: https://github.com/spf13/afero/blob/d92c300a1b6eaa1690b7aecab2ef7fd0d64a94e1/util.go#L285
if the directory does not exist, then `afero.IsEmpty` returns false.

internal/pkg/override/cdk.go
efekarakus: I've replied here: https://github.com/aws/copilot-cli/pull/4572/files#r1127021722

Does that help?

internal/pkg/cli/flag.go
efekarakus: 👍 Good thinking. Let me rename all "patch" occurrences to "extensions".

YAMLPatch is just one way of extending, but an alternative is CDK extensions

internal/pkg/override/cdk.go
efekarakus: I've updated the comment slightly, hopefully that makes it a bit clearer :) I was surprised that `IsEmpty` returned false when the path did not exist, but I can also see why afero maintainers implement it this way.

internal/pkg/cli/flag.go
efekarakus: Done! Is that better?

paragbhingre (approved):
LGTM!

PR #4573, opened 2023-03-03
chore(cli): add optional --env flag to `svc override`

efekarakus:
Address comment feedback from #4566 


internal/pkg/cli/svc_override.go
```diff
@@ -243,6 +255,22 @@ func (o *overrideSvcOpts) newSvcPackageCmd(tplBuf stringWriteCloser) (executor,
...
+		return o.envName, nil
+	}
+	envs, err := o.cfgStore.ListEnvironments(o.appName)
+	if err != nil {
+		return "", fmt.Errorf("list environments in application %q: %v", o.appName, err)
+	}
+	if len(envs) == 0 {
+		return "", fmt.Errorf("no environments found in application %q", o.appName)
+	}
+	return envs[0].Name, nil
```
dannyrandall, line 271: is this "random"?🤔 is the order from `ListEnvironments` deterministic?

dannyrandall (commented):
lg, just a small question

internal/pkg/cli/svc_override.go
efekarakus: You're right it's not "random", I just didn't want customers to rely on a particular environment for it. We should be able to change which env we default to if we want to.

It's always the first environment that they've created for now. Would you like me to update the flag description?

dannyrandall: i was thinking that we don't want people to rely on the undocumented, deterministic behavior hiding potential errors - but with the context that this happens in `svc override`, which is just a best-effort scaffolding that's likely only run a single time, i think the behavior/description is perfectly fine 👌

dannyrandall (approved):

paragbhingre (approved):

PR #4575, opened 2023-03-03
chore: add skeleton code for diff package

Lou1415926:

### DIFF
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -13,8 +13,8 @@ type Node struct {
 	key      string
 	children map[string]*Node // A list of non-empty pointers to the children nodes.
 
-	oldValue *yaml.Node // Only populated for a leaf node.
-	newValue *yaml.Node // Only populated for a leaf node.
+	oldValue *yaml.Node // Only populated for a leaf node (i.e. that has no child node).
+	newValue *yaml.Node // Only populated for a leaf node (i.e. that has no child node).
 }
 
 // Parse constructs a diff tree that represent the differences between two YAML documents.
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -8,7 +8,7 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-// Node represents a difference between two YAML documents.
+// Node represents a segment on a difference between two YAML documents.
 type Node struct {
 	key      string
 	children map[string]*Node // A list of non-empty pointers to the children nodes.
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -1,0 +1,27 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package diff
+
+import (
+	"gopkg.in/yaml.v3"
+)
+
+// DiffNode represents a difference between two YAML documents.
+type DiffNode struct {
+	key      string
+	children map[string]*DiffNode // A list of non-empty pointers to the children nodes.
+
+	oldValue *yaml.Node // Only populated for a leaf node.
+	newValue *yaml.Node // Only populated for a leaf node.
+}
+
+// ConstructDiffTree constructs a diff tree that represent the differences between two YAML documents.
+func ConstructDiffTree(curr, old []byte) (*DiffNode, error) {
+	return &DiffNode{}, nil
+}
+
+// String returns the string representation of the tree stemmed from the diffNode n.
+func (n *DiffNode) String() string {
+	return ""
+}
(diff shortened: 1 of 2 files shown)


internal/pkg/template/diff/diff.go
```diff
@@ -0,0 +1,27 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package diff
+
+import (
+	"gopkg.in/yaml.v3"
+)
+
+// DiffNode represents a difference between two YAML documents.
+type DiffNode struct {
```
efekarakus: nit: consider renaming to just node so that it reads `diff.Node` instead of `diff.DiffNode`

internal/pkg/template/diff/diff.go
```diff
@@ -0,0 +1,27 @@
...
+	key      string
+	children map[string]*DiffNode // A list of non-empty pointers to the children nodes.
+
+	oldValue *yaml.Node // Only populated for a leaf node.
+	newValue *yaml.Node // Only populated for a leaf node.
+}
+
+// ConstructDiffTree constructs a diff tree that represent the differences between two YAML documents.
+func ConstructDiffTree(curr, old []byte) (*DiffNode, error) {
```
efekarakus: What do you think of renaming this function to `Parse` similar to https://pkg.go.dev/text/template/parse@go1.20.1#Parse (that pkg might be a good inspiration for names)

internal/pkg/template/diff/diff.go
```diff
@@ -0,0 +1,27 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package diff
```
efekarakus, line 5: package comment plz 🥺

internal/pkg/template/diff/diff.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"gopkg.in/yaml.v3"
+)
+
+// DiffNode represents a difference between two YAML documents.
```
efekarakus: nit:
```suggestion
// DiffNode holds the contents of two YAML documents under a particular key.
```

efekarakus: out of curiosity: Interesting I wonder how this is going to behave if a key exists in one document but not the other? I am guessing one of `oldValue` or `newValue` will be nil?

internal/pkg/template/diff/diff.go
```diff
@@ -0,0 +1,27 @@
...
+type DiffNode struct {
+	key      string
+	children map[string]*DiffNode // A list of non-empty pointers to the children nodes.
+
+	oldValue *yaml.Node // Only populated for a leaf node.
```
efekarakus: out of curiosity: how do we define a leaf node?

efekarakus (commented):
Looks good, just a bunch of clarification questions and nits

internal/pkg/template/diff/diff.go
Lou1415926: By `len(children) == 0`. I can add this to the field comment.

internal/pkg/template/diff/diff.go
Lou1415926: > I am guessing one of oldValue or newValue will be nil?

This is correct 👍🏼 

> // DiffNode holds the contents of two YAML documents under a particular key.

This is not super duper accurate though - I want to clarify a little to make sure we are on the same page. A node is added to the diff tree iff somewhere under the key, there is a diff:

### Example 1
```
# document 1
Earth:
  Asia: "yo"
  NorthAmerica: "hey"

# document 2
Earth:
  Asia: "yo
  NorthAmerica: "hey"
```
This should yield an empty diff tree with only the sentinel node `&diffNode{}`, because there is no diff

### Example 1
```
# document 1
Earth:
  Asia: "yo"
  NorthAmerica: "hey"

# document 2
Earth:
  NorthAmerica: "hey"
```
This should yield a tree like this:
Untitled-2023-03-03-1011


`NorthAmerica` is not in the tree, because there is no diff.

internal/pkg/template/diff/diff.go
efekarakus: derp that makes sense!

efekarakus (approved):

internal/pkg/template/diff/diff.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"gopkg.in/yaml.v3"
+)
+
+// Node represents a segment on a difference between two YAML documents.
+type Node struct {
+	key      string
+	children map[string]*Node // A list of non-empty pointers to the children nodes.
+
+	oldValue *yaml.Node // Only populated for a leaf node (i.e. that has no child node).
```
paragbhingre, line 16: Dumb question: Why do we only check values for only child nodes?

Lou1415926: It is not necessary to store the yaml content for a branch node. The (diffed) YAML content under the branch node can be reconstructed using its `key` and all of its `children`.

This probably will be clearer if you see the implementation code. The diagram that I showed in https://github.com/aws/copilot-cli/pull/4575#discussion_r1124852003 might be helpful for understanding as well.

huanjani (approved):
I love that you did a skeleton for the unit tests, too!

PR #4576, opened 2023-03-03
chore: parse the diff of the `map` portions of two YAML documents

Lou1415926:


internal/pkg/template/diff/diff.go
```diff
@@ -19,10 +21,107 @@ type Node struct {
 
 // Parse constructs a diff tree that represent the differences between two YAML documents.
 func Parse(curr, old []byte) (*Node, error) {
```
efekarakus: now that I think about it, what do you think of this API format instead: https://dave.cheney.net/2019/09/24/be-wary-of-functions-which-take-several-parameters-of-the-same-type

```go
type From []byte

// Diff constructs a tree of *Node that represents the differences between the from and to YAML documents.
func (from From) Diff(to []byte) (*Node, error)
```

internal/pkg/template/diff/diff.go
```diff
@@ -19,10 +21,107 @@ type Node struct {
...
+	var err error
+	switch {
+	case curr.Kind == yaml.SequenceNode && old.Kind == yaml.SequenceNode:
+		children, err = parseSequence(curr, old)
+	case curr.Kind == yaml.DocumentNode && old.Kind == yaml.DocumentNode:
+		fallthrough
+	case curr.Kind == yaml.MappingNode && old.Kind == yaml.MappingNode:
+		children, err = parseMap(curr, old)
+	default:
+		return nil, fmt.Errorf("unknown combination of node kinds: %v, %v", curr.Kind, old.Kind)
```
efekarakus: Ah I am guessing this is why you've asked a while back if it's guaranteed that the API with CloudFormation never supports composite types.

Now that I'm thinking about it, I believe an IAM policy is possible counter example:
```yaml
resources: '*' # same for actions

# ----  new.yaml
resources:
  - arn1
  - arn2
```
In this scenario, I wonder if we can showcase a diff like this 💭 ?
```
~ resources: "*" -> [arn1, arn2] 
```

efekarakus (commented):
The rest of the PR looks good to me, let me know what you think of the comment!

internal/pkg/template/diff/diff.go
Lou1415926: Good call! 

It is possible that the item in the list is more complex than just a string.
```
Resources:
  - !Join ["", ["arn::aws::s3", "fancy"]]
  - !Join ["", ["arn::aws::s3", "extra-fancy"]]
```

With the proposal, it'd become
```
~ resources: "*" -> [!Join ["", ["arn::aws::s3", "fancy"]], !Join ["", ["arn::aws::s3", "extra-fancy"]]]
```

which can rendered to long and unreadable if the list item becomes more complex.

What do you think of this:

```
- Resources: "*" 
+ Resources:
  + - !Join ["", ["arn::aws::s3", "fancy"]]
  + - !Join ["", ["arn::aws::s3", "extra-fancy"]]
```

efekarakus: Looks great, :shipit:

Lou1415926: Similarly, for a field that can be a scalar or map:
```
- Resources: "*" 
+ Resources:
  + Condition: condition1
  + ExceptFor: this
```

I don't know if there is any actual CFN field that is a string-or-map, but I'm planning to add the logic just in case 😂

Lou1415926: I plan to add this logic to a separate PR if that's alright with you!

efekarakus: sounds good! I'll leave the comment as unresolved so that others can also see.

internal/pkg/template/diff/diff.go
Lou1415926: done!

internal/pkg/template/diff/diff.go
```diff
@@ -17,12 +19,115 @@ type Node struct {
...
+func (from From) Parse(to []byte) (*Node, error) {
+	var toNode, fromNode yaml.Node
+	if err := yaml.Unmarshal(to, &toNode); err != nil {
+		return nil, fmt.Errorf("unmarshal current template: %w", err)
+	}
+	if err := yaml.Unmarshal(from, &fromNode); err != nil {
+		return nil, fmt.Errorf("unmarshal old template: %w", err)
+	}
+
+	return (*fromYAML)(&fromNode).parse(&toNode, "")
```
efekarakus: I'm okay if you want to keep the _private_ function `parse(from, to, key)` and get rid of the type casting.

efekarakus (approved):

iamhopaul123 (approved):

PR #4595, opened 2023-03-06
chore(manifest): support additional_listeners for NLB

paragbhingre:
This PR adds `additional_listeners` manifest field to support multiple ports through NLB.  


internal/pkg/cli/deploy/lbws.go
```diff
@@ -151,7 +151,7 @@ func (d *lbWebSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (*s
 		return nil, err
 	}
 	var opts []stack.LoadBalancedWebServiceOption
-	if !d.lbMft.NLBConfig.IsEmpty() {
+	if !d.lbMft.NLBConfig.PrimaryRoutingRule.IsEmpty() {
```
iamhopaul123: Why do we differentiate primary routing route and additional routing rules?

paragbhingre: Because we have same routing rule details in both primary as well as additional routing rules, so it is efficient to separate out the common parameters between these two out and call them primary and additional.

iamhopaul123: I meant why can't we combine primary routing rule and additional routing rule?

paragbhingre: I did not understand your question. How do we combine them? Today we define routing rule config inside `nlb` and after multiple port feature we will define more routing rules under  `nlb.additional_rules`. I am not sure combining them means what?

internal/pkg/cli/deploy/lbws.go
```diff
@@ -262,7 +262,15 @@ func (d *lbWebSvcDeployer) validateNLBRuntime() error {
 		logAppVersionOutdatedError(aws.StringValue(d.lbMft.Name))
 		return err
 	}
-	return validateLBWSAlias(d.lbMft.NLBConfig.Aliases, d.app, d.env.Name)
+	if err := validateLBWSAlias(d.lbMft.NLBConfig.PrimaryRoutingRule.Aliases, d.app, d.env.Name); err != nil {
```
iamhopaul123: Maybe `d.lbMft.NLBConfig.Aliases()`?

internal/pkg/cli/deploy/lbws.go
```diff
@@ -246,7 +246,7 @@ func (d *lbWebSvcDeployer) validateALBRuntime() error {
 }
 
 func (d *lbWebSvcDeployer) validateNLBRuntime() error {
-	if d.lbMft.NLBConfig.Aliases.IsEmpty() {
+	if d.lbMft.NLBConfig.PrimaryRoutingRule.Aliases.IsEmpty() {
```
iamhopaul123: shall we check additional routing rules too? Or they can't exist without primary routing rule

internal/pkg/manifest/validate.go
```diff
@@ -90,11 +90,21 @@ func (l LoadBalancedWebService) validate() error {
...
 		return fmt.Errorf("validate network load balancer target: %w", err)
 	}
+	for _, additionalRules := range l.NLBConfig.AdditionalRoutingRules {
+		if err = validateTargetContainer(validateTargetContainerOpts{
+			mainContainerName: aws.StringValue(l.Name),
+			mainContainerPort: l.ImageConfig.Port,
+			targetContainer:   additionalRules.TargetContainer,
+			sidecarConfig:     l.Sidecars,
+		}); err != nil {
+			return fmt.Errorf("validate HTTP load balancer target: %w", err)
```
iamhopaul123: is it network load balancer target?

internal/pkg/manifest/validate.go
```diff
@@ -1937,36 +1959,56 @@ func populateNLBPortsAndValidate(containerNameFor map[uint16]string, opts valida
 		return nil
 	}
 	nlb := opts.nlb
-	if nlb.Port == nil {
+	if nlb.PrimaryRoutingRule.Port == nil {
 		return nil
 	}
-	nlbPort, _, err := ParsePortMapping(nlb.Port)
+
+	if err := populateAndValidateNLBPorts(nlb.PrimaryRoutingRule, containerNameFor, opts); err != nil {
```
iamhopaul123: is it a valid use case to specify only additional rules without specifying primary rule for target container etc.?

internal/pkg/cli/deploy/lbws.go
paragbhingre: Additional rules won't exist before primary routing rule. As we discussed offline as well there are multiple features such as SD and SC depends on the input provided in primary routing rule.

iamhopaul123: Sorry do we error out right now if they define manifest like
```
nlb:
  additional_rules:
     target_port: xxx
     target_container:xxx
```

paragbhingre: Yes, we error out because we check for `nlb.port` required parameter is defined or not? If it is not defined then we error out saying define required `nlb.port` parameter.

iamhopaul123: gotcha sounds good then

internal/pkg/manifest/validate.go
paragbhingre: No it is not a valid use case. The reason being what we discussed in the earlier [comment](https://github.com/aws/copilot-cli/pull/4595#discussion_r1129977121).

internal/pkg/cli/deploy/lbws.go
paragbhingre: I think we need to write a `Aliases()` on `NLBConfig` struct in order to do what you are suggesting. I have changed the implementation to have list of `NetworkLoadBalancerRoutingRule` instead of calling  validateLBWSAlias  for both Primary as well as Additional rules. Let me know if new commits solve your concern.

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -232,6 +232,11 @@ func (s LoadBalancedWebService) applyEnv(envName string) (workloadManifest, erro
 
 // NetworkLoadBalancerConfiguration holds options for a network load balancer
 type NetworkLoadBalancerConfiguration struct {
+	PrimaryRoutingRule     NetworkLoadBalancerRoutingRule   `yaml:",inline"`
```
Lou1415926, line 235: what do you think of just `RoutingRule`?

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -250,23 +255,42 @@ func (c *NetworkLoadBalancerConfiguration) IsEmpty() bool {
...
-	exposedPorts = append(exposedPorts, out...)
 	portsForContainer, containerForPort := prepareParsedExposedPortsMap(sortExposedPorts(exposedPorts))
 	return ExposedPortsIndex{
 		PortsForContainer: portsForContainer,
 		ContainerForPort:  containerForPort,
 	}, nil
 }
+
+// NLBRoutingRules returns primary as well as additional rules as a list of NetworkLoadBalancerRoutingRule.
+func (cfg NetworkLoadBalancerConfiguration) NLBRoutingRules() []NetworkLoadBalancerRoutingRule {
```
Lou1415926: Can we move the function close to the definition of its receiver?

internal/pkg/manifest/lb_web_svc.go
Lou1415926: Is this valid? iirc there is no "rule" in NLB

internal/pkg/cli/deploy/lbws.go
efekarakus: My tiny nit is that sometimes we use the word "Main" and sometimes "Primary" to refer to similar concepts. Can we standardize to perhaps `Main`?

https://github.com/aws/copilot-cli/blame/499bb1ca37b11a9955377594ac8c8826bb7d7ce8/internal/pkg/manifest/svc.go#L642

internal/pkg/manifest/lb_web_svc.go
paragbhingre: Oops you are right, we can name it as listener though. Good catch!

internal/pkg/cli/deploy/lbws.go
```diff
@@ -271,7 +277,7 @@ func validateLBWSAlias(aliases manifest.Alias, app *config.Application, envName
 	}
 	aliasList, err := aliases.ToStringSlice()
 	if err != nil {
-		return fmt.Errorf(`convert 'http.alias' to string slice: %w`, err)
+		return fmt.Errorf(`convert 'nlb.alias' to string slice: %w`, err)
```
efekarakus: This function appears to be also called by `validateALBRuntime` as well 💭 Wouldn't that result in a confusing error message if the error happens under `http`?

If the above is true, my suggestion is just to `return err` here and then each client should wrap the error:
```go
return fmt.Errorf("validate http.alias: %v", err)
return fmt.Errorf("validate nlb.alias: %v", err)
```

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
```diff
@@ -344,15 +344,15 @@ func (s *LoadBalancedWebService) Parameters() ([]*cloudformation.Parameter, erro
 			},
 		}...)
 	}
-	if !s.manifest.NLBConfig.IsEmpty() {
-		port, _, err := manifest.ParsePortMapping(s.manifest.NLBConfig.Port)
+	if !s.manifest.NLBConfig.PrimaryRoutingRule.IsEmpty() {
+		port, _, err := manifest.ParsePortMapping(s.manifest.NLBConfig.PrimaryRoutingRule.Port)
```
efekarakus: Do you think we could get rid of this parameter completely? If I recall correctly, you moved all the data to be passed through `template` types.

I'm asking because we don't seem to pass any parameters for the additional rules so that made me think that maybe it's not needed for the main one either.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -465,7 +465,7 @@ func convertEnvSecurityGroupCfg(mft *manifest.Environment) (*template.SecurityGr
 
 func (s *LoadBalancedWebService) convertNetworkLoadBalancer() (networkLoadBalancerConfig, error) {
```
efekarakus: The PR title starts with `feat` but as far as I can tell we don't pass multiple `Listeners` yet for NLB right? should the PR be a `chore:` instead

internal/pkg/manifest/lb_web_svc.go
efekarakus: oh yea this is awkward 😂 

It looks like `additional_rules` makes sense for ALB but for NLB this would be more like `additional_listeners` 

I think we might need to bite the bullet and renaming this field for NLB 😭 Perhaps, since this is a more advanced feature we can deem having two separate field names OK.


(imo we also need to rename the fields and types)

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -241,7 +246,7 @@ type NetworkLoadBalancerConfiguration struct {
 	Aliases         Alias              `yaml:"alias"`
 }
 
-func (c *NetworkLoadBalancerConfiguration) IsEmpty() bool {
+func (c *NetworkLoadBalancerRoutingRule) IsEmpty() bool {
```
efekarakus: doc comment 🥺

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -232,6 +232,11 @@ func (s LoadBalancedWebService) applyEnv(envName string) (workloadManifest, erro
 
 // NetworkLoadBalancerConfiguration holds options for a network load balancer
 type NetworkLoadBalancerConfiguration struct {
+	PrimaryRoutingRule     NetworkLoadBalancerRoutingRule   `yaml:",inline"`
+	AdditionalRoutingRules []NetworkLoadBalancerRoutingRule `yaml:"additional_rules"`
+}
+
+type NetworkLoadBalancerRoutingRule struct {
```
efekarakus: doc comment

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -250,23 +255,42 @@ func (c *NetworkLoadBalancerConfiguration) IsEmpty() bool {
...
 	}
+	// port from http.target_port.
 	exposedPorts = append(exposedPorts, lbws.RoutingRule.exposedPorts(exposedPorts, workloadName)...)
-	out, err := lbws.NLBConfig.exposedPorts(exposedPorts, workloadName)
-	if err != nil {
-		return ExposedPortsIndex{}, err
+
+	nlbRoutingRules := lbws.NLBConfig.NLBRoutingRules()
+	// port from nlb.target_port and nlb.additional_rules[x].target_port
+	for _, rule := range nlbRoutingRules {
```
efekarakus:
```suggestion
	// port from nlb.target_port and nlb.additional_rules[x].target_port
	for _, listener := range  lbws.NLBConfig.Listeners() {
```

internal/pkg/manifest/lb_web_svc.go
efekarakus: Is there a unit test for this exposed method?

internal/pkg/manifest/validate.go
```diff
@@ -87,13 +87,16 @@ func (l LoadBalancedWebService) validate() error {
...
+	nlbRoutingRules := l.NLBConfig.NLBRoutingRules()
+	for _, rule := range nlbRoutingRules {
+		if err = validateTargetContainer(validateTargetContainerOpts{
+			mainContainerName: aws.StringValue(l.Name),
+			mainContainerPort: l.ImageConfig.Port,
+			targetContainer:   rule.TargetContainer,
+			sidecarConfig:     l.Sidecars,
+		}); err != nil {
+			return fmt.Errorf("validate network load balancer target: %w", err)
+		}
```
efekarakus: what do you think of including the index for which listener's target container we are validating?

For example:
```go
return fmt.Errorf("validate target for nlb.listener[%i]: %w", i, err)
```

internal/pkg/manifest/validate.go
```diff
@@ -843,6 +846,16 @@ func (ip IPNet) validate() error {
 
 // validate returns nil if NetworkLoadBalancerConfiguration is configured correctly.
 func (c NetworkLoadBalancerConfiguration) validate() error {
+	nlbRoutingRules := c.NLBRoutingRules()
+	for _, rule := range nlbRoutingRules {
+		if err := rule.validate(); err != nil {
+			return fmt.Errorf("validate nlb routing rule; %s", err.Error())
```
efekarakus: 😮 do we do this anywhere? this seems so strange
```suggestion
			return fmt.Errorf("validate nlb.listener[%i]: %v", i, err)
```

internal/pkg/manifest/validate.go
```diff
@@ -1937,36 +1950,52 @@ func populateNLBPortsAndValidate(containerNameFor map[uint16]string, opts valida
...
+		targetContainer = aws.StringValue(rr.TargetContainer)
 	}
-	if exposed, ok := containerNameFor[containerPort]; ok {
-		if nlb.TargetContainer != nil && aws.StringValue(nlb.TargetContainer) != exposed {
+	containerNameFor[targetPort] = targetContainer
+	return nil
+}
+
+func errOnContainersExposingSamePort(containerNameFor map[uint16]string, targetPort uint16, targetContainer *string) error {
+	if exposed, ok := containerNameFor[targetPort]; ok {
```
```suggestion
container, exists := containerNameFor[targetPort]
if !exists {
   return nil
}

```

internal/pkg/deploy/cloudformation/stack/lb_web_svc.go
paragbhingre: If you are talking about `NLBAliases` and `NLBPort` template params then we do need them in uri.go to build our NLB links that customer will use to send their request. I know it's not the most perfect way to do it, but that's how we have done in the past. Me and Penghao agreed on the point that we should get rid of that logic to use template params to describe NLB links. But we will do it separately in general refactoring instead of multiple port feature. Then we can remove these params.  Does that sound good to you?

internal/pkg/manifest/validate.go
paragbhingre: I think that just happened by mistake 🙈

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -230,8 +230,14 @@ func (s LoadBalancedWebService) applyEnv(envName string) (workloadManifest, erro
 	return &s, nil
 }
 
-// NetworkLoadBalancerConfiguration holds options for a network load balancer
+// NetworkLoadBalancerConfiguration holds options for a network load balancer.
 type NetworkLoadBalancerConfiguration struct {
+	MainListener        NetworkLoadBalancerListener   `yaml:",inline"`
```
Lou1415926: nit: I wonder if the `Main` prefix is redundant. I think the fact that this listener is not an "additional listener" already implies that it is the main. I don't have a strong opinion on this naming though.

internal/pkg/manifest/lb_web_svc_test.go
```diff
@@ -2397,3 +2526,68 @@ func TestLoadBalancedWebService_BuildArgs(t *testing.T) {
...
 }
+
+func TestNetworkLoadBalancerConfiguration_NLBListeners(t *testing.T) {
+	testCases := map[string]struct {
+		in     NetworkLoadBalancerConfiguration
+		wanted []NetworkLoadBalancerListener
+	}{
+		"return empty list if there are no Listeners provided": {
+			in: NetworkLoadBalancerConfiguration{},
+		},
```
Lou1415926: nit:
```suggestion
		"return empty list if there are no listener provided": {		},
```

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -250,23 +257,42 @@ func (c *NetworkLoadBalancerConfiguration) IsEmpty() bool {
...
 	}, nil
 }
+
+// NLBListeners returns main as well as additional listeners as a list of NetworkLoadBalancerListener.
+func (cfg NetworkLoadBalancerConfiguration) NLBListeners() []NetworkLoadBalancerListener {
+	var nlbListeners []NetworkLoadBalancerListener
+	if !cfg.MainListener.IsEmpty() {
+		nlbListeners = append(nlbListeners, cfg.MainListener)
+	}
+	nlbListeners = append(nlbListeners, cfg.AdditionalListeners...)
```
Lou1415926: is this valid:
```
nlb:
   additional_listeners:
      - ...
```

I suppose not?

internal/pkg/manifest/validate.go
```diff
@@ -843,6 +845,15 @@ func (ip IPNet) validate() error {
 
 // validate returns nil if NetworkLoadBalancerConfiguration is configured correctly.
 func (c NetworkLoadBalancerConfiguration) validate() error {
+	for idx, listener := range c.NLBListeners() {
```
Lou1415926: Do we want to validate here that `AdditionalListeners` are there only if `MiainListener` is there?

internal/pkg/manifest/validate.go
```diff
@@ -843,6 +845,15 @@ func (ip IPNet) validate() error {
...
+		if err := listener.validate(); err != nil {
+			return fmt.Errorf("validate nlb.listener[%d]; %s", idx+1, err)
```
Lou1415926: Why `idx+1` 🤔 ? Can we validate the main listener first, and then loop through the additional listeners? This way we can just use `idx` when wrapping errors from the additional listeners.

internal/pkg/manifest/validate.go
```diff
@@ -1937,36 +1948,54 @@ func populateNLBPortsAndValidate(containerNameFor map[uint16]string, opts valida
 		return nil
 	}
 	nlb := opts.nlb
-	if nlb.Port == nil {
+	if nlb.MainListener.Port == nil {
```
Lou1415926: Do we still need this if block 🤔

internal/pkg/manifest/validate.go
```diff
@@ -1937,36 +1948,54 @@ func populateNLBPortsAndValidate(containerNameFor map[uint16]string, opts valida
...
+	for _, listener := range nlb.NLBListeners() {
+		if err := populateAndValidateNLBPorts(listener, containerNameFor, opts); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func populateAndValidateNLBPorts(listener NetworkLoadBalancerListener, containerNameFor map[uint16]string, opts validateExposedPortsOpts) error {
+	nlbPort, _, err := ParsePortMapping(listener.Port)
```
Lou1415926, line 1975: qq: is it possible for `listener.Port` to be `nil` here? I'm assuming we would have errored out if a listener has a `nil` port, is this correct?

internal/pkg/manifest/validate.go
```diff
@@ -1937,36 +1948,54 @@ func populateNLBPortsAndValidate(containerNameFor map[uint16]string, opts valida
...
+
+	for _, listener := range nlb.NLBListeners() {
+		if err := populateAndValidateNLBPorts(listener, containerNameFor, opts); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func populateAndValidateNLBPorts(listener NetworkLoadBalancerListener, containerNameFor map[uint16]string, opts validateExposedPortsOpts) error {
```
Lou1415926: Do we still need the `opts validateExposedPortsOpts` parameter?

internal/pkg/manifest/validate.go
```diff
@@ -1937,36 +1948,54 @@ func populateNLBPortsAndValidate(containerNameFor map[uint16]string, opts valida
...
+	targetContainer := opts.mainContainerName
+	if listener.TargetContainer != nil {
+		targetContainer = aws.StringValue(listener.TargetContainer)
 	}
-	containerNameFor[containerPort] = targetContainerName
+	containerNameFor[targetPort] = targetContainer
+	return nil
+}
 
+func errOnContainersExposingSamePort(containerNameFor map[uint16]string, targetPort uint16, targetContainer *string) error {
```
Lou1415926: nit: maybe something along the line
```suggestion
func validatePortExposedOnce(containerNameFor map[uint16]string, targetPort uint16, targetContainer *string) error {
```

internal/pkg/manifest/lb_web_svc.go
paragbhingre: yeah it is not valid, but that is taken care by the validation already so we are safe here. But if you want to be extra safe then I can put this with MainListener as well.

paragbhingre: Yeah I think I should keep additional listeners with the main listener to make sure if main listener exists then only additional listeners are considered. This will also take care of your next [comment](https://github.com/aws/copilot-cli/pull/4595#discussion_r1131608657).

Lou1415926: 👍🏼  In this case we can just return when `!cfg.MainListener.IsEmpty()`

internal/pkg/manifest/validate.go
paragbhingre: Do you mean 1st validate the 0th index separately and then loop over remaining items? 
I kept idx+1 so that customer won't see listener[0] and get confused. Is it right practice to show 0th index to the customer?

Lou1415926: I meant ⬇️ 
```
if err := validate(c.MainListener); err != nil {
}

for idx, listener := range c.AdditionalListeners {
}
```

Lou1415926: I think all of our customers are programmers, starting with `0` is probably fine.. This is also consistent with our other error messages taht involves an index

internal/pkg/manifest/validate.go
paragbhingre: removed.
Edit - I think now we have 2 params under `NetworkLoadBalancerConfiguration` i.e. `MainListener` and `AdditionalListeners` so it's required to return if `MainListener` is not defined. 
So I have changed the existing entry condition to `if opts.nlb == nil || opts.nlb.MainListener.IsEmpty()` to make sure in order to proceed we need nlb is not null and main listener is provided before any other additional listeners.

internal/pkg/manifest/validate.go
paragbhingre: I think we can just need `mainContainerName`, so we can just send that param instead. Made the changes.

internal/pkg/manifest/validate.go
paragbhingre: yes we error out if port is empty even before coming here. So port being nil here is not possible.

internal/pkg/manifest/lb_web_svc.go
paragbhingre: Let's see if other's have any opinion about this. I don't have a strong opinion either.

internal/pkg/cli/deploy/lbws.go
```diff
@@ -262,7 +265,12 @@ func (d *lbWebSvcDeployer) validateNLBRuntime() error {
 		logAppVersionOutdatedError(aws.StringValue(d.lbMft.Name))
 		return err
 	}
-	return validateLBWSAlias(d.lbMft.NLBConfig.Aliases, d.app, d.env.Name)
+	for _, listener := range d.lbMft.NLBConfig.NLBListeners() {
+		if err := validateLBWSAlias(listener.Aliases, d.app, d.env.Name); err != nil {
+			return fmt.Errorf(`validate 'nlb.alias': %w`, err)
```
efekarakus: should this be something like:
```go
for i, listener := range d.lbMft.NLBConfig.NLBListeners() {
   err := fmt.Errorf(`validate "nlb.alias": %w`, err)
   if i > 0 {
       err = fmt.Errorf(`validate "nlb.additional_listeners[%d].alias": %w`, i-1, err)
   }
   return err
}
```

internal/pkg/manifest/lb_web_svc.go
efekarakus: sounds like a good idea to me too! It's a `nit` for me as well so it's not blocking.

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -250,23 +257,42 @@ func (c *NetworkLoadBalancerConfiguration) IsEmpty() bool {
...
+// NLBListeners returns main as well as additional listeners as a list of NetworkLoadBalancerListener.
+func (cfg NetworkLoadBalancerConfiguration) NLBListeners() []NetworkLoadBalancerListener {
+	if cfg.MainListener.IsEmpty() {
+		return nil
+	}
+	var nlbListeners []NetworkLoadBalancerListener
+	nlbListeners = append(nlbListeners, cfg.MainListener)
+	nlbListeners = append(nlbListeners, cfg.AdditionalListeners...)
+
+	return nlbListeners
```
efekarakus: nit:
```suggestion
	return append([]NetworkLoadBalancerListener{cfg.MainListener}, cfg.AdditionalListeners...)
```

internal/pkg/manifest/validate.go
```diff
@@ -90,10 +90,20 @@ func (l LoadBalancedWebService) validate() error {
 	if err = validateTargetContainer(validateTargetContainerOpts{
 		mainContainerName: aws.StringValue(l.Name),
 		mainContainerPort: l.ImageConfig.Port,
-		targetContainer:   l.NLBConfig.TargetContainer,
+		targetContainer:   l.NLBConfig.MainListener.TargetContainer,
 		sidecarConfig:     l.Sidecars,
 	}); err != nil {
-		return fmt.Errorf("validate network load balancer target: %w", err)
+		return fmt.Errorf("validate target for nlb.listener: %w", err)
```
```suggestion
		return fmt.Errorf("validate target for nlb: %w", err)
```

internal/pkg/manifest/validate.go
```diff
@@ -90,10 +90,20 @@ func (l LoadBalancedWebService) validate() error {
...
+	}
+	for idx, listener := range l.NLBConfig.AdditionalListeners {
+		if err = validateTargetContainer(validateTargetContainerOpts{
+			mainContainerName: aws.StringValue(l.Name),
+			mainContainerPort: l.ImageConfig.Port,
+			targetContainer:   listener.TargetContainer,
+			sidecarConfig:     l.Sidecars,
+		}); err != nil {
+			return fmt.Errorf("validate target for nlb.listener[%d]: %w", idx, err)
```
```suggestion
			return fmt.Errorf("validate target for nlb.additional_listeners[%d]: %w", idx, err)
```

internal/pkg/manifest/validate.go
```diff
@@ -843,6 +853,18 @@ func (ip IPNet) validate() error {
 
 // validate returns nil if NetworkLoadBalancerConfiguration is configured correctly.
 func (c NetworkLoadBalancerConfiguration) validate() error {
+	if err := c.MainListener.validate(); err != nil {
+		return fmt.Errorf("validate nlb.listener; %s", err)
```
efekarakus: the `; %s` is here as well :) we need to fix them all to:
```suggestion
		return fmt.Errorf("validate nlb: %w", err)
```

internal/pkg/manifest/validate.go
```diff
@@ -843,6 +853,18 @@ func (ip IPNet) validate() error {
...
+	}
+	for idx, listener := range c.AdditionalListeners {
+		if err := listener.validate(); err != nil {
+			return fmt.Errorf("validate nlb.listener[%d]; %s", idx, err)
```
```suggestion
			return fmt.Errorf("validate nlb.additional_listeners[%d]: %w", idx, err)
```

internal/pkg/manifest/validate.go
```diff
@@ -1933,40 +1955,58 @@ func populateALBPortsAndValidate(containerNameFor map[uint16]string, opts valida
...
 func populateNLBPortsAndValidate(containerNameFor map[uint16]string, opts validateExposedPortsOpts) error {
-	if opts.nlb == nil {
+	if opts.nlb == nil || opts.nlb.MainListener.IsEmpty() {
 		return nil
 	}
 	nlb := opts.nlb
-	if nlb.Port == nil {
-		return nil
+	if err := populateAndValidateNLBPorts(nlb.MainListener, containerNameFor, opts.mainContainerName); err != nil {
+		return fmt.Errorf(`validate nlb.listener: %w`, err)
```
efekarakus: (same here)

efekarakus (commented):
LGTM! Some small error message corrections and an NLB alias thought but after that :shipit:

internal/pkg/cli/deploy/lbws.go
efekarakus: Thinking about this some more, what does it to have a different alias per listener for NLB?

wouldn't all the alias names point to the same NLB endpoint 💭 I wonder if `alias` should remain a property only at the top level and not part of any listener 🤔 i.e. something like:

```go
type NetworkLoadBalancerConfiguration struct {
   Aliases Alias  `yaml:"alias"`
   MainListener NetworkLoadBalancerListener `yaml:",inline"`
   AdditionalListeners []NetworkLoadBalancerListener `yaml:"additional_listeners"`
}
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -465,7 +465,7 @@ func convertEnvSecurityGroupCfg(mft *manifest.Environment) (*template.SecurityGr
...
 	nlbConfig := s.manifest.NLBConfig
-	if nlbConfig.IsEmpty() {
```
Lou1415926: I think we should implement `IsEmpty` for `manifest.NLB` as well, and calls `nlbConfg.IsEmpty()` here, instead if `nlbConfig.MainListener.IsEmpty()`.  

It is true that for a valid NLB config, `nlbConfig.MainListener.IsEmpty()` implies that  `nlbConfig.IsEmpty()`, but I think this is an assumption that client package should not be aware of, but rather hidden within `manifest` package. LMK if this makes sense!

internal/pkg/cli/deploy/lbws.go
```diff
@@ -151,7 +151,7 @@ func (d *lbWebSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (*s
 		return nil, err
 	}
 	var opts []stack.LoadBalancedWebServiceOption
-	if !d.lbMft.NLBConfig.IsEmpty() {
+	if !d.lbMft.NLBConfig.MainListener.IsEmpty() {
```
Lou1415926: Same here about calling `NLBConfig.IsEmpty()`

internal/pkg/manifest/lb_web_svc.go
Lou1415926: nit:
```suggestion
	return append(nlbListeners, cfg.AdditionalListeners...)
```

internal/pkg/manifest/validate.go
```diff
@@ -166,7 +176,7 @@ func (w WorkerAlarmArgs) validate() error {
 // validate returns nil if LoadBalancedWebServiceConfig is configured correctly.
 func (l LoadBalancedWebServiceConfig) validate() error {
 	var err error
-	if l.RoutingRule.Disabled() && l.NLBConfig.IsEmpty() {
+	if l.RoutingRule.Disabled() && l.NLBConfig.MainListener.IsEmpty() {
```
Lou1415926: Same here. We should call `l.NLBConfig.IsEmpty()`

internal/pkg/manifest/validate.go
```diff
@@ -1933,40 +1955,58 @@ func populateALBPortsAndValidate(containerNameFor map[uint16]string, opts valida
 }
 
 func populateNLBPortsAndValidate(containerNameFor map[uint16]string, opts validateExposedPortsOpts) error {
-	if opts.nlb == nil {
+	if opts.nlb == nil || opts.nlb.MainListener.IsEmpty() {
```
Lou1415926: sameee

internal/pkg/manifest/validate.go
```diff
@@ -1933,40 +1955,58 @@ func populateALBPortsAndValidate(containerNameFor map[uint16]string, opts valida
...
-	if nlb.Port == nil {
-		return nil
+	if err := populateAndValidateNLBPorts(nlb.MainListener, containerNameFor, opts.mainContainerName); err != nil {
+		return fmt.Errorf(`validate nlb.listener: %w`, err)
 	}
-	nlbPort, _, err := ParsePortMapping(nlb.Port)
+
+	for idx, listener := range nlb.AdditionalListeners {
+		if err := populateAndValidateNLBPorts(listener, containerNameFor, opts.mainContainerName); err != nil {
+			return fmt.Errorf(`validate nlb.listener[%d]: %w`, idx, err)
```
Lou1415926: just in case: same as Efe's comment about wrapping the error with `nlb.additional_listener[%d]` instead of `nlb.listener[%d]`!

internal/pkg/manifest/validate_test.go
```diff
@@ -178,12 +178,41 @@ func TestLoadBalancedWebService_validate(t *testing.T) {
...
+						MainListener: NetworkLoadBalancerListener{
+							Port:            aws.String("443"),
+							TargetContainer: aws.String("foo"),
+						},
+					},
+				},
+			},
+			wantedErrorMsgPrefix: `validate target for nlb.listener: `,
+		},
+		"error if fail to validate network load balancer target for additional listener": {
```
Lou1415926: Can you remind me why these configurations are problemetic 😰 Are any containers exposing the same ports?

Lou1415926 (commented):
+1 to efe's!

internal/pkg/manifest/validate_test.go
paragbhingre: Ohh no, if you look at `additional_listeners` it is trying to expose port `444` of sidecar container `foo` which does not even exist that is why we are getting error - `validate target for nlb.additional_listeners[0]: target container "foo" doesn't exist`

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: Yeah I think it does a lot of sense. I totally agree that client shouldn't make any assumption that `nlbConfig.MainListener.IsEmpty()` implies that `nlbConfig.IsEmpty()`. I have added this method which is really useful for validations as well.

internal/pkg/manifest/validate.go
```diff
@@ -846,6 +856,28 @@ func (c NetworkLoadBalancerConfiguration) validate() error {
 	if c.IsEmpty() {
 		return nil
 	}
+	if err := c.Listener.validate(); err != nil {
+		return fmt.Errorf("validate nlb: %w", err)
```
iamhopaul123: Should "validate nlb" be prefixed by functions that call `NetworkLoadBalancerConfiguration.validate()` 🤔

paragbhingre: yes actually I was thinking the same how we can have a different error message in here because else we will end up having `validate "nlb": validate nlb` for main listener and `validate "nlb": validate nlb.additional_listeners[x]` for additional listeners. 

How about we only return error in case of main listener which would make error something like - `validate "nlb": port must be specified...` and for additional listener it will be `validate nlb: validate nlb.additional_listeners[x]: port must be specified..`?

internal/pkg/manifest/svc.go
```diff
@@ -479,7 +479,7 @@ func (rr RoutingRuleConfiguration) exposedPorts(exposedPorts []ExposedPort, work
 // exportPorts returns any new ports that should be exposed given the network load balancer
 // configuration that's not part of the existing containerPorts.
 func (cfg NetworkLoadBalancerListener) exposedPorts(exposedPorts []ExposedPort, workloadName string) ([]ExposedPort, error) {
-	if cfg.IsEmpty() {
+	if cfg.Port == nil {
```
Lou1415926: Just curious - any reason why we changed `cfg.IsEmpty()` to `cfg.Port == nil`? After validation, it's guaranteed that `Port != nil` for all listeners; therefore, `cfg.IsEmpty()` is eventually equivalent to `cfg.Port == nil` - is this the reason why?

If so, do we guarantee that `validate` always happen before `exposedPorts`?

internal/pkg/manifest/validate.go
Lou1415926: The `validate "nlb"` wrapping already happens at https://github.com/aws/copilot-cli/blob/7c00353d9c743d55b1490479e134e3c892083e5e/internal/pkg/manifest/validate.go#L234. I think we can remove all "validate nlb" wrapping in this function. E.g.

```go
if err := c.Aliases.validate(); err != nil {
		return fmt.Errorf(`validate "alias": %w`, err)
	}
```

internal/pkg/manifest/validate.go
```diff
@@ -853,21 +853,31 @@ func (ip IPNet) validate() error {
...
+		for _, advancedAlias := range c.Aliases.AdvancedAliases {
+			if advancedAlias.HostedZone != nil {
+				return fmt.Errorf(`validate nlb: "hosted_zone" is not supported for Network Load Balancer`)
+			}
+		}
 	}
 	for idx, listener := range c.AdditionalListeners {
 		if err := listener.validate(); err != nil {
-			return fmt.Errorf("validate nlb.listener[%d]; %s", idx, err)
+			return fmt.Errorf("validate nlb.additional_listeners[%d]: %w", idx, err)
```
Lou1415926: If we remove all `validate "nlb"` wrapping in this `validate()`, then
```suggestion
			return fmt.Errorf(`validate "additional_listeners[%d]": %w`, idx, err)
```

Lou1415926 (commented):
LGTM! Just a couple of very minor nits

internal/pkg/manifest/svc.go
paragbhingre: The reason I changed it to `Port == nil` is that it is consistent with all the other exposedPorts functions. Also, another important reason why changed is because Main listener is comprised of `NetworkLoadBalancerListener` properties and `Aliases` which makes it important for us to make sure customer has not provided any property to declare Main listener as empty and `cfg.IsEmpty()` gives answer as Main listener is not empty even if customer has just provided `Aliases` and `cfg.Port == nil` perfectly does its job for every NLB listener, irrespective of `exopsedPorts()` called before or after validation. Does that answer your questions?

Lou1415926: > it is consistent with all the other exposedPorts functions.

Do you mean for example L458?
```
if rr.TargetPort == nil {
   return nil
}
```

`rr.TargetPort == nil` makes sense to me, because `exposedPort` returns of a list of container ports exposed by `targetPort`.

However, `cfg.Port == nil` is a little counterintuitive, because `cfg.Port` is the listener port, not the target port. Although `exportPorts` indeed uses  `cfg.Port`, but it's only when `cfg.TargetPort` is `nil`. 

I know that in the real world,  we require `cfg.Port` to be `non-nil`, i.e. there won't be a case when `cfg.TargetPort != nil && cfg.Port == nil`. However, this is not a logic that `exposedPorts` handles.

> Main listener is comprised of `NetworkLoadBalancerListener` properties and `Aliases`

I am confused here. I thought the "main listener" **is** a `NetworkLoadBalancerListener`, and does not contain `Aliases`. `Aliases` is a property of `NetworkLoadBalancerConfiguration`, but this function's receiver is `NetworkLoadBalancerListener`.

paragbhingre: Ohh you are right, I got confused with the naming. Yeah "main listener" is a NetworkLoadBalancerListener, and does not contain Aliases. So the thing is if customer defines just `nlb.Stickiness: true` then  `if cfg.IsEmpty()`is  false and we execute rest of the logic for `exposedPorts()` though ideally we should catch this logic in validate itself. For now we definitely execute validate before we execute `exposedPorts()` so we are safe. But `cfg.Port == nil` keeps us safe in all the conditions. I am fine either way.

iamhopaul123 (approved):

Lou1415926 (approved):

PR #4596, opened 2023-03-06
chore: parse the diff of two YAML files for properties with same names but different kinds

Lou1415926:
Address https://github.com/aws/copilot-cli/pull/4576#discussion_r1125105443


efekarakus (approved):
:shipit:

internal/pkg/template/diff/diff_test.go
```diff
@@ -127,6 +127,96 @@ func TestFrom_Parse(t *testing.T) {
...
+							key: "Mary",
+							children: map[string]*Node{
+								"Dialogue": leafDialogue,
+							},
+						},
+					},
+				}
+			},
+		},
+		"change a map to list": {
```
huanjani, line 186: The mock text is AMAZING. ❤️

internal/pkg/template/diff/diff.go
```diff
@@ -42,11 +42,12 @@ func (from From) Parse(to []byte) (*Node, error) {
 }
 
 func parse(from, to *yaml.Node, key string) (*Node, error) {
-	if to == nil || from == nil {
+	// Handle base cases.
+	if (to == nil || from == nil) || to.Kind != from.Kind {
```
iamhopaul123:
```suggestion
	if to == nil || from == nil || to.Kind != from.Kind {
```

internal/pkg/template/diff/diff_test.go
```diff
@@ -127,6 +127,96 @@ func TestFrom_Parse(t *testing.T) {
...
+  Dialogue:
+    - Bear: "I know I'm supposed to keep an eye on you"
+      Tone: disappointed
+    - Dog: "ikr"
+      Tone: pleased`,
+			old: `Mary:
+  Dialogue:
+    Bear: (disappointed) "I know I'm supposed to keep an eye on you"
+    Dog: (pleased) "ikr"`,
+			wanted: func() *Node {
```
iamhopaul123, line 197: Why do we use `func()`?

iamhopaul123 (approved):
LGTM just one nit and a question

internal/pkg/template/diff/diff_test.go
Lou1415926: Without `func()` it'd have to be super deeply nested, whereas now I can define `leaf` var and reduce the nesting. It's just a matter of readability.

Issue #4259, opened 2022-12-09
Preview diff on deploy/package


dannyrandall:
This issue is a design proposal for adding a flag to `copilot [noun] deploy` and `copilot [noun] package` that previews changes that will be made to the existing cloudformation stack. This will show what changes Copilot will make to existing and new resources, which will become especially useful when testing/deploying extensions via https://github.com/aws/copilot-cli/issues/4208 or https://github.com/aws/copilot-cli/issues/4209.

## Proposal
A new flag, `--diff`, will be added to `deploy` and `package` commands that calculates and prints the diff between the deployed stack (including the addons stack!) and the stack that _will be_ deployed or generated.

After displaying the diff, `copilot [noun] deploy --diff` will prompt with a confirmation like:
```
Continue with deployment? [y/N]:
```
before deploying the proposed changes.

`copilot [noun] package --diff` will print the diff and exit. The exit code will be as follows:
0 = no diffs found
1 = diffs found
2 = error producing diff

This should allow for a script to skip deploying a stack if no changes will occur.

### UX Examples
Some examples of how a diff will look in different scenarios:

#### Adding a lambda function
```
~ Resources:
  + EnvControllerFunction:
    + Type: AWS::Lambda::Function
    + Properties:
      + Code:
        + S3Bucket: "s3://location"
        + S3Key: "s3://location"
      + Timeout: 900
      + MemorySize: 512
```

#### Adding a statement to an IAM Role `AssumeRolePolicyDocument`:
```
~ Resources:
  ~ TaskRole:
    ~ Properties:
      ~ AssumeRolePolicyDocument:
        ~ Statement:
          # (1 unchanged list entry)
          + - Action: "*"
              Effect: "Allow"
              Principal:
                Service: "ecs-tasks.amazonaws.com"
          # (2 unchanged list entries)
```

#### Updating an attribute of an IAM Role `AssumeRolePolicyDocument`:
```
~ Resources:
  ~ TaskRole:
    ~ Properties:
      ~ AssumeRolePolicyDocument:
        ~ Statement:
          # (1 unchanged list entry)
          ~ - Effect: "Allow" -> "Deny"
          # (2 unchanged list entries)
```

#### Changing the `ContainerPort` parameter
```
~ Parameters:
  ~ ContainerPort:
    ~ Value: 80 -> 443
```

efekarakus:
Hi folks!!

[...]

PR #4597, opened 2023-03-07
chore: parse the diff of the `seq` portions of two YAML documents 

Lou1415926:

### DIFF
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -6,6 +6,7 @@ package diff
 
 import (
 	"fmt"
+	"strconv"
 
 	"gopkg.in/yaml.v3"
 )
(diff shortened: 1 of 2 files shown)


internal/pkg/template/diff/diff.go
```diff
@@ -88,7 +89,80 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+
+func compare(from, to []yaml.Node) (map[string]*nodeOccurrence, error) {
+	seen := make(map[string]*nodeOccurrence)
+	for _, node := range from {
+		b, err := yaml.Marshal(node)
+		if err != nil {
+			return nil, err
+		}
+		content := string(b)
+		if v, ok := seen[content]; ok {
```
efekarakus, line 141: do we need make sure that the [`Style`](https://github.com/go-yaml/yaml/blob/f6f7691b1fdeb513f56608cd2c32c51f8194bf51/yaml.go#L379) field is consistent before serializing into a string for comparison of nodes?
Similarly, we might need to set the `Comment` fields to `""` recursively before marshaling.

Otherwise, they might be not be the same 🤔

internal/pkg/template/diff/diff.go
```diff
@@ -88,7 +89,80 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+		return nil, err
+	}
+	seen, err := compare(fromSeq, toSeq)
+	if err != nil {
+		return nil, err
+	}
+	var index int
+	children := make(map[string]*Node)
+	for _, occurrence := range seen {
+		if occurrence.inTo == occurrence.inFrom {
```
efekarakus, line 106: If I'm following correctly, we're assuming that that Sequences behave as `sets` is that correct? (i.e. the order in which a node appears is not relevant)

I wonder if this is always a safe assumption 💭 Can we think of any counter examples? If not perhaps we can drop a comment explaining that here :)

internal/pkg/template/diff/diff.go
```diff
@@ -88,7 +89,80 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+		}
+		for i := 0; i < occurrence.inTo-occurrence.inFrom; i++ {
+			children[strconv.Itoa(index)] = &Node{
+				key:      strconv.Itoa(index),
+				newValue: &(occurrence.node),
+			}
+			index++
+		}
+	}
+	return children, nil
```
efekarakus, line 124: In Danny's UX: https://github.com/aws/copilot-cli/issues/4259 (which is inspired by Terraform https://www.hashicorp.com/blog/terraform-0-14-adds-a-new-concise-diff-format-to-terraform-plans) there is also some metadata around the number of elements that do not change:
```
# (2 unchanged elements hidden)
```

I wonder how we can capture that info 💭  Perhaps it warrants being a field in `Node`

internal/pkg/template/diff/diff.go
```diff
@@ -88,7 +89,80 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+}
+
+type nodeOccurrence struct {
+	node   yaml.Node
+	inTo   int
+	inFrom int
+}
+
+func compare(from, to []yaml.Node) (map[string]*nodeOccurrence, error) {
```
efekarakus, line 133: What do you think of renaming the function and changing the return type as follows:
```suggestion
// nodeCounter holds the number of times a yaml.Node appears in a sequence.
type nodeCounter struct {
    node yaml.Node
    counts []int // Number of times node appeared in each sequence.
}

// countDistinct returns a slice of each unique yaml.Node across sequences and the number of times the node appeared in each sequence.
func countDistinct(sequences ...[]yaml.Node) ([]*nodeCounter, error) {
```

My concerns were that `compare` is a bit too generic as a name for what this function is doing (it's really just counting).
Similarly, I wasn't sure if a `map[string]` is needed because clients don't care about the key (the serialized yaml.Node) 

Let me know what you think!

internal/pkg/template/diff/diff.go
Lou1415926: You are right 😭  Some intrinsic functions like `!Sub` takes an ordered list. These are all I've found so far after I search through our codebase. I can't think of any counter examples so far for "real" CFN fields (will update this comment if any one comes up). Good catch!

Lou1415926: To be safe though, I'm fine with taking the approach that assumes the list is ordered. This works better with the proposal in #4259 where
```yaml
~ Resources:
  ~ TaskRole:
    ~ Properties:
      ~ AssumeRolePolicyDocument:
        ~ Statement:
          # (1 unchanged list entry)
          + - Action: "*"
              Effect: "Allow"
              Principal:
                Service: "ecs-tasks.amazonaws.com"
          # (2 unchanged list entries)
```

There are ` # (1 unchanged list entry)` at the beginning and `# (2 unchanged list entries)` at the end. That means the design here assumes that the list is ordered.

iamhopaul123 (approved):
LGTM awesome +1 to efe's comment

Lou1415926:
The PR is based on the assumption that "the list aren't ordered". This is a wrong assumption in YAML. I will close this one in favor of a new PR that makes the correct assumption.

New PR: #4611

PR #4602, opened 2023-03-07
chore(describe): add env and alarm description to `svc show`

huanjani:
These changes provide more information about rollback alarms associated with a given service.
Screenshot 2023-03-07 at 2 33 42 PM


internal/pkg/aws/cloudwatch/cloudwatch.go
```diff
@@ -125,6 +133,58 @@ func (cw *CloudWatch) AlarmStatuses(opts ...DescribeAlarmOpts) ([]AlarmStatus, e
...
+		alarmDescriptions = append(alarmDescriptions, cw.metricAlarmsDescriptions(alarmResp.MetricAlarms)...)
+		if alarmResp.NextToken == nil {
+			break
+		}
+		in.NextToken = alarmResp.NextToken
+	}
+	return alarmDescriptions, nil
+}
+
+func (cw *CloudWatch) compositeAlarmsDescriptions(alarms []*cloudwatch.CompositeAlarm) []*AlarmDescription {
```
iamhopaul123, line 160: how about
```suggestion
func (cloudwatch.CompositeAlarm) description() *AlarmDescription {
```
and move the for loop outside 🤔

iamhopaul123 (approved):
lgtm overall just one nit

internal/pkg/aws/cloudwatch/cloudwatch.go
```diff
@@ -125,6 +133,58 @@ func (cw *CloudWatch) AlarmStatuses(opts ...DescribeAlarmOpts) ([]AlarmStatus, e
...
 }
 
+// AlarmDescriptions returns the config of alarms filtered by name.
+func (cw *CloudWatch) AlarmDescriptions(alarmNames []string) ([]*AlarmDescription, error) {
+	var alarmDescriptions []*AlarmDescription
+	in := &cloudwatch.DescribeAlarmsInput{
+		AlarmNames: aws.StringSlice(alarmNames),
+	}
+	for {
+		alarmResp, err := cw.client.DescribeAlarms(in)
```
paragbhingre, line 143: Should this be outside of for loop? Or am I misunderstanding something here? Are we describing all the alarms in every loop?

internal/pkg/aws/cloudwatch/cloudwatch.go
```diff
@@ -125,6 +133,58 @@ func (cw *CloudWatch) AlarmStatuses(opts ...DescribeAlarmOpts) ([]AlarmStatus, e
...
+		if err != nil {
+			return nil, fmt.Errorf("describe CloudWatch alarms: %w", err)
+		}
+		if alarmResp == nil {
+			break
+		}
+		alarmDescriptions = append(alarmDescriptions, cw.compositeAlarmsDescriptions(alarmResp.CompositeAlarms)...)
```
paragbhingre, line 150: What is the difference between composite and Metric alarm Description and which onr do we choose to display if one alarm has both the descriptions available?

paragbhingre (commented):
Description is going to be very helpful for customer 🤩
I just have a couple questions.

internal/pkg/aws/cloudwatch/cloudwatch.go
huanjani: Good question! It's because of the `NextToken`-- if there are multiple pages of output, then we have to paginate through it.

internal/pkg/aws/cloudwatch/cloudwatch.go
huanjani: [Composite alarms](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Create_Composite_Alarm.html) are overarching alarms made up of metric or other composite alarms. Both types work for ABR, so we need to display descriptions of both types.

internal/pkg/aws/cloudwatch/cloudwatch.go
huanjani: I think it's actually a little more streamlined and clear as is, and also avoids nested for-loops. 🤷🏼

paragbhingre (approved):

PR #4605, opened 2023-03-08
chore: compute the longest common subseq given two lists

Lou1415926:


internal/pkg/template/diff/list.go
```diff
@@ -0,0 +1,51 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
efekarakus: nit: should we rename the file to `lcs.go`

internal/pkg/template/diff/list.go
```diff
@@ -0,0 +1,51 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package diff
+
+func longestCommonSubsequence[T comparable](a []T, b []T) []T {
```
efekarakus: How are we going to use this with `yaml.Node`s? 😅 

Do you think that having a signature like this makes sense:
```suggestion
// Returns the longest common sequence between a and b, as well as the indices where each sequence items appeared in a and b.
func longestCommonSubsequence[T comparable](a []T, b []T) ([]T, []int, []int) {
```

My thought was that we can pass in the string representation of each `yaml.Node` in `a` and `b` and since we get the indices we would have a way of telling if the cursor should be deleted or added

efekarakus (commented):
Awesomeee, just curious to hear your thoughts on one suggestion

internal/pkg/template/diff/list.go
Lou1415926: > My thought was that we can pass in the string representation of each

Exactly this 

> since we get the indices we would have a way of telling if the cursor should be deleted or added

SGTM

Lou1415926: > My thought was that we can pass in the string representation of each yaml.Node in a and b

Change of mind 🚑 🚑 ! This is not the best way to compare two list items. Consider this:
```yaml
- Sid: power
  Action: '*'
  Resources: '*'

# Compare with

- Sid: power
  Resources: '*'
  Action: '*'
```

The only thing that changed was the order of the properties, but this will produce a diff. CloudFormation does not create a change set in this case either. 

This has something to do with the `Content []*yaml.Node` field in a `yaml.Node`. I guess YAML marshals the sub fields with the order that a child appears in `Content`. We could of course sort the `Content` before marshaling... but it looks fragile and untidy.


My plan now is to do something like:
```go
// var fromSeq, toSeq []yaml.Node
lcsInFrom, lcsInTo := longestCommonSubsequence(fromSeq, toSeq, func(inA, inB int) bool {
   diff, err := parse(fromSeq[inA], toSeq[inB], "") // A recursive call to "parse". 
   if err != nil {
      return false
   }
   return diff == nil
})

```

I've also added more unit tests that target the case where `[]yaml.Node`, instead of `[]string`, are passed in.

efekarakus (approved):
:shipit:

internal/pkg/template/diff/lcs.go
```diff
@@ -0,0 +1,53 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package diff
+
+type eqFunc func(inA, inB int) bool
```
iamhopaul123, line 6: Why do we need this 🤔

internal/pkg/template/diff/lcs.go
```diff
@@ -0,0 +1,53 @@
...
+
+func longestCommonSubsequence[T any](a []T, b []T, eq eqFunc) ([]int, []int) {
```
iamhopaul123, line 16: maybe an example input and output would be more helpful? from the function name i would assume it output the longest common sequence.

iamhopaul123 (approved):

internal/pkg/template/diff/lcs.go
Lou1415926: It doesn't matter, just for easier refactoring (potentially). For example, if we want to change the `eq` parameter to `func(inA, inB) (bool, error)`, we won't need to change the function signature everywhere - we just need to change the code around where it's invoked.

PR #4609, opened 2023-03-09
chore: keep `http.path: '/'` for LBWS in apps with domain

efekarakus:
Address UX feedback from https://app.gitter.im/#/room/#aws_copilot-cli:gitter.im/$BrcLblwPZotiiRhfC6TOSgtLApSNnZBto3Puky4BSzk to stop writing the service name in LBWS manifest if the application is initialized with a domain

### DIFF
--- a/internal/pkg/initialize/workload.go
+++ b/internal/pkg/initialize/workload.go
@@ -307,39 +307,41 @@ func (w *WorkloadInitializer) newServiceManifest(i *ServiceProps) (encoding.Bina
 	}
 }
 
-func (w *WorkloadInitializer) newLoadBalancedWebServiceManifest(i *ServiceProps) (*manifest.LoadBalancedWebService, error) {
+func (w *WorkloadInitializer) newLoadBalancedWebServiceManifest(inProps *ServiceProps) (*manifest.LoadBalancedWebService, error) {
 	var httpVersion string
-	if i.Port == commonGRPCPort {
+	if inProps.Port == commonGRPCPort {
 		log.Infof("Detected port %s, setting HTTP protocol version to %s in the manifest.\n",
-			color.HighlightUserInput(strconv.Itoa(int(i.Port))), color.HighlightCode(manifest.GRPCProtocol))
+			color.HighlightUserInput(strconv.Itoa(int(inProps.Port))), color.HighlightCode(manifest.GRPCProtocol))
 		httpVersion = manifest.GRPCProtocol
 	}
-	props := &manifest.LoadBalancedWebServiceProps{
+	outProps := &manifest.LoadBalancedWebServiceProps{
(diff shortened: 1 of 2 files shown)


internal/pkg/initialize/workload.go
```diff
@@ -307,39 +307,41 @@ func (w *WorkloadInitializer) newServiceManifest(i *ServiceProps) (encoding.Bina
...
+		HealthCheck: inProps.HealthCheck,
+		Platform:    inProps.Platform,
 	}
-	existingSvcs, err := w.Store.ListServices(i.App)
+	existingSvcs, err := w.Store.ListServices(inProps.App)
 	if err != nil {
 		return nil, err
 	}
-	// We default to "/" for the first service, but if there's another
+	// We default to "/" for the first service or if the application is initialized with a domain, but if there's another
```
KollaAdithya, line 334: may be a dumb question: If an application is initialized with domain, does this mean for all the services in that application will have `http.path: '/'`

KollaAdithya (approved):

internal/pkg/initialize/workload.go
efekarakus: When an app is initialized with a domain, say `example.com`, then we can use host-header based routing for LoadBalancedWebServices.

So for example, `frontend.test.example.com/` will route to the `frontend` service; whereas, `backend.test.example.com/` will route to the `backend` service.

Whereas if there is no domain that's associated with the application, then we have to use path based routing.
So for example, `http://alb-demo-1234567890.us-west-2.elb.amazonaws.com/` will go to the `frontend` service; and since we cannot use the same URL to navigate to a different service we give instead the default path: `http://alb-demo-1234567890.us-west-2.elb.amazonaws.com/backend` to go to `backend`.

Hope that clarified how things worked behind the scenes :)

internal/pkg/initialize/workload_test.go
```diff
@@ -297,7 +298,7 @@ func TestAppInitOpts_createLoadBalancedAppManifest(t *testing.T) {
 
 			wantedPath: "/",
 		},
-		"creates manifest with {app name} as the path if there's another LBWebApp": {
```
paragbhingre, line 300: Just curious, did we ever have this implementation where app name was the path for LBWS?

paragbhingre (commented):
LGTM! Thanks for fixing some naming.

internal/pkg/initialize/workload_test.go
efekarakus: this is legacy code (before v1.0) back when copilot used to call "apps" -> "projects" and "services" -> "apps". I just updated the testcase name to reflect the terminology we've been using

paragbhingre (approved):

PR #4611, opened 2023-03-10
chore: parse the diff of the `seq` portions of two YAML documents

Lou1415926:

### DIFF
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -113,7 +113,7 @@ func parseSequence(fromNode, toNode *yaml.Node) (map[string]*Node, error) {
 		}
 		return err == nil && diff == nil
 	})
-	childKey, children, inspector := seqChildKeyFunc(), make(map[string]*Node), newInspector(fromSeq, toSeq, lcsIndices)
+	nextChildKey, children, inspector := seqChildKeyFunc(), make(map[string]*Node), newInspector(fromSeq, toSeq, lcsIndices)
 	for action := inspector.inspect(); action != done; action = inspector.inspect() {
 		switch action {
 		case match:
@@ -124,13 +124,13 @@ func parseSequence(fromNode, toNode *yaml.Node) (map[string]*Node, error) {
 			if diff.err != nil {
 				return nil, diff.err
 			}
-			children[childKey()] = diff.node
+			children[nextChildKey()] = diff.node
 		case deletion:
 			item := inspector.fromItem()
-			children[childKey()] = &Node{oldValue: &item}
+			children[nextChildKey()] = &Node{oldValue: &item}
 		case insertion:
 			item := inspector.toItem()
-			children[childKey()] = &Node{newValue: &item}
+			children[nextChildKey()] = &Node{newValue: &item}
 		}
 		inspector.proceed()
 	}
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -91,12 +91,12 @@ func isYAMLLeaf(node *yaml.Node) bool {
 }
 
 func parseSequence(fromNode, toNode *yaml.Node) (map[string]*Node, error) {
-	var fromSeq, toSeq []yaml.Node
-	if err := toNode.Decode(&toSeq); err != nil {
-		return nil, err
+	fromSeq, toSeq := make([]yaml.Node, len(fromNode.Content)), make([]yaml.Node, len(toNode.Content)) // NOTE: should be the same as calling `Decode`.
+	for idx, v := range fromNode.Content {
+		fromSeq[idx] = *v
 	}
-	if err := fromNode.Decode(&fromSeq); err != nil {
-		return nil, err
+	for idx, v := range toNode.Content {
+		toSeq[idx] = *v
 	}
 	type cachedEntry struct {
 		node *Node
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -160,9 +160,6 @@ func parseSequence(from, to *yaml.Node) (map[string]*Node, error) {
 		f++
 		t++
 	}
-	for k, v := range children {
-		fmt.Printf("%v:%v\n", k, v)
-	}
 	return children, nil
 }
 
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -6,6 +6,7 @@ package diff
 
 import (
 	"fmt"
+	"strconv"
 
 	"gopkg.in/yaml.v3"
 )
@@ -101,11 +102,12 @@ func parseSequence(from, to *yaml.Node) (map[string]*Node, error) {
 	lcsIdxInFrom, lcsIdxInTo := longestCommonSubsequence(fromSeq, toSeq, func(inA, inB int) bool {
 		diff, err := parse(&(fromSeq[inA]), &(toSeq[inB]), "")
 		if diff != nil { // NOTE: cache the diff only if a modification could have happened at this position.
-			cachedDiff[fmt.Sprintf("%v,%v", inA, inB)] = diff
-			cachedErr[fmt.Sprintf("%v,%v", inA, inB)] = err
+			cachedDiff[cachedKey(inA, inB)] = diff
+			cachedErr[cachedKey(inA, inB)] = err
 		}
 		return err == nil && diff == nil // NOTE: we swallow the error for now.
 	})
+	childKey := seqChildKeyFunc()
 	children := make(map[string]*Node)
 	var f, t, i int
 	for {
@@ -121,24 +123,21 @@ func parseSequence(from, to *yaml.Node) (map[string]*Node, error) {
 			// TODO(lou1415926): (x unchanged items)
 		case f != lcsF && t != lcsT: // Modification.
 			// TODO(lou1415926): handle list of maps modification
-			diff, err := cachedDiff[fmt.Sprintf("%v,%v", f, t)], cachedErr[fmt.Sprintf("%v,%v", f, t)]
+			diff, err := cachedDiff[cachedKey(f, t)], cachedErr[cachedKey(f, t)]
 			if err != nil {
 				return nil, err
 			}
-			children[fmt.Sprintf("%d,%d", f, t)] = diff
+			children[childKey()] = diff
 			f++
 			t++
-
 		case f != lcsF: // Deletion.
-			children[fmt.Sprintf("%d,%d", f, t)] = &Node{oldValue: &(fromSeq[f])}
+			children[childKey()] = &Node{oldValue: &(fromSeq[f])}
 			f++
-
 		case t != lcsT: // Insertion.
-			children[fmt.Sprintf("%d,%d", f, t)] = &Node{newValue: &(toSeq[t])}
+			children[childKey()] = &Node{newValue: &(toSeq[t])}
 			t++
 		}
 	}
-
 	for {
 		if f >= len(fromSeq) && t >= len(toSeq) {
 			break
@@ -151,11 +150,11 @@ func parseSequence(from, to *yaml.Node) (map[string]*Node, error) {
 			if err != nil {
 				return nil, err
 			}
-			children[fmt.Sprintf("%d,%d", f, t)] = diff
+			children[childKey()] = diff
 		case f < len(fromSeq): // Deletion.
-			children[fmt.Sprintf("%d,%d", f, t)] = &Node{oldValue: &(fromSeq[f])}
+			children[childKey()] = &Node{oldValue: &(fromSeq[f])}
 		case t < len(toSeq): // Insertion.
-			children[fmt.Sprintf("%d,%d", f, t)] = &Node{newValue: &(toSeq[t])}
+			children[childKey()] = &Node{newValue: &(toSeq[t])}
 		}
 		f++
 		t++
@@ -200,4 +199,17 @@ func unionOfKeys[T any](a, b map[string]T) map[string]struct{} {
 		keys[k] = exists
 	}
 	return keys
+}
+
+func cachedKey(inFrom, inTo int) string {
+	return fmt.Sprintf("%d,%d", inFrom, inTo)
+}
+
+// TODO(lou1415926): use a more meaningful key for a seq child.
+func seqChildKeyFunc() func() string {
+	idx := -1
+	return func() string {
+		idx++
+		return strconv.Itoa(idx)
+	}
 }
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -113,7 +113,7 @@ func parseSequence(from, to *yaml.Node) (map[string]*Node, error) {
 			f++
 			t++
 			i++
-			// TODO: (x unchanged items)
+			// TODO(lou1415926): (x unchanged items)
 		case f != lcsF && t != lcsT: // Modification.
 			// TODO: cache the diff from LCS?
 			diff, err := parse(&(fromSeq[f]), &(toSeq[t]), "")
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -113,7 +113,7 @@ func parseSequence(fromNode, toNode *yaml.Node) (map[string]*Node, error) {
 		}
 		return err == nil && diff == nil
 	})
-	nextKey, children, inspector := seqChildKeyFunc(), make(map[string]*Node), newInspector(fromSeq, toSeq, lcsIndices)
+	nextKey, children, inspector := seqChildKeyFunc(), make(map[string]*Node), newLCSStateMachine(fromSeq, toSeq, lcsIndices)
 	for action := inspector.inspect(); action != actonDone; action = inspector.inspect() {
 		switch action {
 		case actionMatch:
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -140,10 +140,7 @@ func parseSequence(fromNode, toNode *yaml.Node) (map[string]*Node, error) {
 			to.proceed()
 		}
 	}
-	for {
-		if from.index() >= len(fromSeq) && to.index() >= len(toSeq) {
-			break
-		}
+	for from.index() < len(fromSeq) || to.index() < len(toSeq) {
 		switch {
 		case from.index() < len(fromSeq) && to.index() < len(toSeq): // Modification.
 			diff, err := parse(&(fromSeq[from.index()]), &(toSeq[to.index()]), "")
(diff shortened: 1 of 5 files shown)


internal/pkg/template/diff/diff.go
```diff
@@ -90,7 +90,77 @@ func isYAMLLeaf(node *yaml.Node) bool {
 }
 
 func parseSequence(from, to *yaml.Node) (map[string]*Node, error) {
-	return nil, nil
+	fromSeq, toSeq := make([]yaml.Node, len(from.Content)), make([]yaml.Node, len(to.Content))
+	if err := to.Decode(&toSeq); err != nil {
```
dannyrandall: do you have to pass in a slice with the capacity already allocated for this to work? or does `Decode` already call `make()`?

internal/pkg/template/diff/diff.go
```diff
@@ -90,7 +90,77 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+		return nil, err
+	}
+	if err := from.Decode(&fromSeq); err != nil {
+		return nil, err
+	}
+	cachedDiff, cachedErr := make(map[string]*Node), make(map[string]error)
```
dannyrandall:
```suggestion
    type cacheEntry struct {
    	  node *Node
	      error error
    }
	cache := make(map[string]cacheEntry)
```
if the keys are the same and entries are always present in both?

internal/pkg/template/diff/diff.go
```diff
@@ -90,7 +90,77 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+	lcsIdxInFrom, lcsIdxInTo := longestCommonSubsequence(fromSeq, toSeq, func(inA, inB int) bool {
```
dannyrandall:
```suggestion
	lcsIdxInFrom, lcsIdxInTo := longestCommonSubsequence(fromSeq, toSeq, func(idxFrom, idxTo int) bool {
```
nit

dannyrandall (commented):
couple small things from a first glance

internal/pkg/template/diff/diff.go
Lou1415926: I *think* `Decode` calls `make()` if it's a `nil` slice, and does not do so otherwise. The motivation was to allocate just enough memory to prevent resizing the slice.

internal/pkg/template/diff/diff.go
Lou1415926: 9Ehh since the size of `struct` is usually larger than the sum of the sizes of its fields, I tried to avoid doing this since there might potentially be a lot of diff. But after a quick look, it seems like in the case of pointer + error, the struct size isn't larger than the sum (guess it sort of depend on CPU arch though).

internal/pkg/template/diff/diff.go
efekarakus: Wut 🤔  it's friday afternoon and my brain stopped working but what is the point of the `Decode` 😅 

Isn't `from.Content` already a `[]yaml.Node` so can we avoid calling `Decode` altogether?

internal/pkg/template/diff/diff.go
```diff
@@ -90,7 +91,75 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+		diff, err := parse(&(fromSeq[inA]), &(toSeq[inB]), "")
+		if diff != nil { // NOTE: cache the diff only if a modification could have happened at this position.
+			cachedDiff[cachedKey(inA, inB)] = diff
+			cachedErr[cachedKey(inA, inB)] = err
+		}
+		return err == nil && diff == nil // NOTE: we swallow the error for now.
```
efekarakus: nit: is this "for now going: to change 😝 ? I wonder if we should remove these comments

internal/pkg/template/diff/diff.go
```diff
@@ -90,7 +91,75 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+	})
+	childKey := seqChildKeyFunc()
+	children := make(map[string]*Node)
+	var f, t, i int
```
dannyrandall: could we add a comment on what they are? 🥺 i think i've figured it out but took reading the rest to decide that `f= pointer in from, t = pointer in to, i = ?? (lcs pointer?)`

internal/pkg/template/diff/diff.go
```diff
@@ -131,3 +200,16 @@ func unionOfKeys[T any](a, b map[string]T) map[string]struct{} {
 	}
 	return keys
 }
+
+func cachedKey(inFrom, inTo int) string {
```
dannyrandall: nit
```suggestion
func cacheKey(inFrom, inTo int) string {
```

internal/pkg/template/diff/diff.go
```diff
@@ -90,7 +91,75 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+		}
+		return err == nil && diff == nil // NOTE: we swallow the error for now.
+	})
+	childKey := seqChildKeyFunc()
+	children := make(map[string]*Node)
+	var f, t, i int
+	for {
+		if i >= len(lcsIdxInFrom) {
+			break
+		}
```
dannyrandall: say from is `a,b,c` and to is `a,b,c,d` - will we miss adding the `d` to `children` on line 136? (does that even matter? :joy:)

efekarakus: Maybe drop a comment saying that `lcsIndxInFrom` and `lcsIdxInTo` are equal to each other. Or maybe this would be clearer:
```go
if lcsLen := len(lcsIdxInFrom); i >= lcsLen
```

internal/pkg/template/diff/diff.go
dannyrandall: also maybe a description of each for loop above it could be nice

internal/pkg/template/diff/diff.go
dannyrandall: oh maybe it gets caught in the lower for loop but i'll leave this comment here to be sure

internal/pkg/template/diff/diff.go
```diff
@@ -90,7 +91,75 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+			t++
+		case f != lcsF: // Deletion.
+			children[childKey()] = &Node{oldValue: &(fromSeq[f])}
+			f++
+		case t != lcsT: // Insertion.
+			children[childKey()] = &Node{newValue: &(toSeq[t])}
+			t++
+		}
+	}
+	for {
```
dannyrandall: yeah some doc here that this for loop handles insertions/deletions/modifications _beyond_ the lcs would be helpful 👍

dannyrandall (commented):
mostly just asking for more doc to help make the algo more understandable, but looks solid otherwise!!! great work 👏👏👏👏👏

internal/pkg/template/diff/diff.go
dannyrandall: i wonder if it'd be nice to just rely on `Decode` to do that bc i don't think perf/space is that big of an issue and i'd like to simplify where we can 🥺...at least until we have evidence that resizing is causing noticeable perf issues😊

internal/pkg/template/diff/diff.go
```diff
@@ -90,7 +91,75 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+		}
+		lcsF, lcsT := lcsIdxInFrom[i], lcsIdxInTo[i]
+		switch {
+		case f == lcsF && t == lcsT: // Match.
+			f++
+			t++
+			i++
+			// TODO(lou1415926): (x unchanged items)
+		case f != lcsF && t != lcsT: // Modification.
+			// TODO(lou1415926): handle list of maps modification
```
efekarakus, line 122: What does this mean 😅 wouldn't `diff` contain what we want?

internal/pkg/template/diff/diff.go
```diff
@@ -90,7 +91,75 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+			f++
+		case t != lcsT: // Insertion.
+			children[childKey()] = &Node{newValue: &(toSeq[t])}
+			t++
+		}
+	}
+	for {
+		if f >= len(fromSeq) && t >= len(toSeq) {
+			break
+		}
```
efekarakus: should we delete this if-statement given the first switch case?

internal/pkg/template/diff/diff.go
```diff
@@ -90,7 +91,75 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+		switch {
+		case f >= len(fromSeq) && t >= len(toSeq):
+			break
+		case f < len(fromSeq) && t < len(toSeq): // Modification.
+			diff, err := parse(&(fromSeq[f]), &(toSeq[t]), "")
```
efekarakus: I see -- the `cachedDiff` can only be used until the LCS is over. After we exceed an LCS index, we have to find the new diffs.

internal/pkg/template/diff/diff.go
```diff
@@ -90,7 +91,75 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+	cachedDiff, cachedErr := make(map[string]*Node), make(map[string]error)
+	lcsIdxInFrom, lcsIdxInTo := longestCommonSubsequence(fromSeq, toSeq, func(inA, inB int) bool {
+		diff, err := parse(&(fromSeq[inA]), &(toSeq[inB]), "")
+		if diff != nil { // NOTE: cache the diff only if a modification could have happened at this position.
+			cachedDiff[cachedKey(inA, inB)] = diff
+			cachedErr[cachedKey(inA, inB)] = err
+		}
+		return err == nil && diff == nil // NOTE: we swallow the error for now.
+	})
+	childKey := seqChildKeyFunc()
```
efekarakus: nit:
```suggestion
	next := seqChildKeyFunc()
```

internal/pkg/template/diff/diff.go
```diff
@@ -90,7 +91,75 @@ func isYAMLLeaf(node *yaml.Node) bool {
...
+	children := make(map[string]*Node)
+	var f, t, i int
+	for {
+		if i >= len(lcsIdxInFrom) {
+			break
+		}
+		lcsF, lcsT := lcsIdxInFrom[i], lcsIdxInTo[i]
+		switch {
+		case f == lcsF && t == lcsT: // Match.
```
efekarakus: nit: I wonder if this body would be easier to grok if things were laid out like:
| current name | proposed |
|:--------|:------|
| `f` | `fromIdx` |
| `t` | `toIdx` |
| `i` | `lcsIdx` |


and we change the function signature of:
```go
// before
longestCommonSubsequence[T any](a []T, b []T, eq eqFunc) ([]int, []int)

// after
type lcsIndex struct {
  from int
  to int
}
longestCommonSubsequence[T any](a []T, b []T, eq eqFunc) []lcsIndex
```

then the whole thing can read:
```go
lcs := longestCommonSubsequence(fromSeq, toSeq, ...)

// other stuff...

lcsIdx := lcs[lcsIndex]
switch {
case fromIdx == lcsIdx.from && toIdx == lcsIdx.to: // and so on...
}
```

Just a thought 💭 I'd be curious to hear if others find that easier to follow or if they have a different suggestion.

efekarakus (commented):
The logic looks good 🚀

iamhopaul123 (approved):
awesome! LGTM

internal/pkg/template/diff/diff.go
Lou1415926: It's not! I was thinking "for now" as "at this point of execution" 😂  because if `err != nil`, it's most likely an un/marshaling error, and it's gonna surface sooner or later. This `eq` func just doesn't bother handling it! We can remove the comment!

internal/pkg/template/diff/diff.go
Lou1415926: `diff` contains a part of what we want!

I think of "list of maps" as having two scenarios: 1. the maps are "implicitly" keyed; 2. otherwise

Consider this example for "implicit" key
```yaml
- Name: Efe
  Boss: True
- Name: Boss # You can't have another person named "Boss" here! i.e. implicitly keyed.
  Boss: True

# to ⬇️ 

- Name: Efe
  Boss: True
- Name: Boss
  Boss: So True
```

I want this diff to be rendered as 

```yaml
~ - Name: Boss
  ~ Boss: True -> So True
```
That is, the "implicit key" here is `Name`. The current implementation does not handle this.

An example of "unkeyed" items:
```yaml
- Shape: round
  Taste: sweet
- Shape: round
  Taste: sweet
  Color: orange

⬇️ 
- Shape: round
 Taste: sweet
- Shape: round
  Taste: bitter
  Color: orange
```

The `diff` should be:
```yaml
~ - Shape: round
  ~ Taste: sweet -> bitter
  ~ Color: orange
```

instead of 
```yaml
~ - Taste: sweet -> bitter
```

That means there are extra information we need to encode into `Node`, for a map in a list.

internal/pkg/template/diff/diff.go
Lou1415926: ooo the `cachedDiff` can be used here. I just forgot to change it 😓 

The `longestCommonSubsequence` is guaranteed to have called `parse` on every single combination of `f` and `t`. So `cachedDiff` works for the whole `fromSeq x toSeq` matrix, as long as that position has as diff.

I should add this fact to `longestCommonSubsequence`'s comment as well I think.

internal/pkg/template/diff/diff.go
Lou1415926: done!

internal/pkg/template/diff/diff.go
Lou1415926: Good idea 👍🏼 lmk how you like the current `iter` implementation & if it smells like over-engineering 😂

internal/pkg/template/diff/diff.go
Lou1415926: Refactored the code a little, added the comment here! https://github.com/aws/copilot-cli/blob/eda0f814832fd689ab7e4448f09655a6eeddc3b8/internal/pkg/template/diff/iter.go#L50

internal/pkg/template/diff/diff.go
Lou1415926: renamed to `nextKey()` 👀

internal/pkg/template/diff/iter.go
```diff
@@ -0,0 +1,108 @@
...
+import "gopkg.in/yaml.v3"
+
+type action int
+
+const (
+	match        action = iota
+	modification        = iota
+	deletion            = iota
+	insertion           = iota
+	done                = iota
```
dannyrandall:
```suggestion
	match        action = iota
	modification
	deletion
	insertion
	done
```

internal/pkg/template/diff/iter.go
```diff
@@ -0,0 +1,108 @@
...
+	from       tracker[yaml.Node]
+	to         tracker[yaml.Node]
+	lcsIndices tracker[lcsIndex]
+	currAction action
+}
+
+func (i *inspector) inspect() action {
+	var action action
+	var (
+		commonDone = i.lcsIndices.index >= len(i.lcsIndices.data)
```
dannyrandall:
```suggestion
		lcsDone = i.lcsIndices.index >= len(i.lcsIndices.data)
```

internal/pkg/template/diff/iter.go
```diff
@@ -0,0 +1,108 @@
...
+			// Ex: "a" -> "a,d,e". When the inspector moves to "d" in "to", both common and "from" are done, and "d,e" are considered to be inserted.
+			action = insertion
+		default:
+			// Ex: "a,b" -> "a,c". When the inspector moves to index 1 of both list, common is done, and b is modified into c.
+			action = modification
+		}
+		i.currAction = action
+		return action
+	}
+	commonIdx := i.lcsIndices.data[i.lcsIndices.index]
```
dannyrandall, line 62:
```suggestion
	lcsIdx := i.lcsIndices.data[i.lcsIndices.index]
```

internal/pkg/template/diff/iter.go
```diff
@@ -0,0 +1,108 @@
...
+		case fromDone && toDone:
+			action = done
+		case toDone:
+			// Ex: "a,d,e" -> "a". When the inspector moves to "d" in "from", both common and "to" are done, and "d,e" are considered deleted.
+			action = deletion
+		case fromDone:
+			// Ex: "a" -> "a,d,e". When the inspector moves to "d" in "to", both common and "from" are done, and "d,e" are considered to be inserted.
+			action = insertion
+		default:
+			// Ex: "a,b" -> "a,c". When the inspector moves to index 1 of both list, common is done, and b is modified into c.
```
dannyrandall:
```suggestion
			// Ex: "a,b" -> "a,c". When the inspector moves by 1 index in both lists, common is done, and b is modified into c.
```
is this more accurate?

dannyrandall (approved):
this is great! good job @Lou1415926 🎉

internal/pkg/template/diff/iter.go
```diff
@@ -0,0 +1,108 @@
...
+
+func newInspector(fromSeq, toSeq []yaml.Node, lcsIndices []lcsIndex) inspector {
+	return inspector{
+		from:       tracker[yaml.Node]{data: fromSeq},
+		to:         tracker[yaml.Node]{data: toSeq},
+		lcsIndices: tracker[lcsIndex]{data: lcsIndices},
+	}
+}
+
+type inspector struct {
```
efekarakus: nit: is this type name too generic given that it's taking `lcsIndices` as an input. 

What do you think of  `lcsStateMachine` as the type name?

internal/pkg/template/diff/iter.go
```diff
@@ -0,0 +1,108 @@
...
+	from       tracker[yaml.Node]
+	to         tracker[yaml.Node]
+	lcsIndices tracker[lcsIndex]
+	currAction action
+}
+
+func (i *inspector) inspect() action {
```
efekarakus:
```suggestion
func (i *inspector) action() action {
```

internal/pkg/template/diff/iter.go
efekarakus: nit: can we prefix the actions with `action` e.g.
```
actionMatch action = iota
actionMod
actionDel
acitionInsert
actionDone
```

Your code reminded me of this talk: https://go.dev/talks/2011/lex.slide#10

internal/pkg/template/diff/iter.go
```diff
@@ -0,0 +1,108 @@
...
+		action = deletion
+	default:
+		// Ex: "a,b,c" -> "1,2,a". When the inspector is at (a,1) and (a,2), only "a" is common, "1,2" are considered inserted.
+		action = insertion
+	}
+	i.currAction = action
+	return action
+}
+
+func (i *inspector) proceed() {
```
efekarakus: nit:
```suggestion
func (i *inspector) next() {
```

efekarakus (approved):
looks good to me too, similar to danny I just have some naming recommendations

internal/pkg/template/diff/iter.go
Lou1415926: changed it to `// Ex: "a,b" -> "a,c". When the lcsStateMachine moves to (b,c), lcs is done, and b is modified into c.`

PR #4612, opened 2023-03-10
chore: stop displaying hints in answer of MultiSelectOption

efekarakus:

### DIFF
--- a/internal/pkg/term/prompt/select.go
+++ b/internal/pkg/term/prompt/select.go
@@ -190,3 +190,12 @@ func parseValueFromOptionFmt(formatted string) string {
 	}
 	return strings.TrimSpace(formatted)
 }
+
+func parseValuesFromOptions(formatted string) string {
+	options := strings.Split(formatted, ", ")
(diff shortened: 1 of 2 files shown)


iamhopaul123 (approved):

internal/pkg/term/prompt/select.go
```diff
@@ -190,3 +190,12 @@ func parseValueFromOptionFmt(formatted string) string {
 	}
 	return strings.TrimSpace(formatted)
 }
+
+func parseValuesFromOptions(formatted string) string {
```
Lou1415926, line 194: This works well as well with [`FriendlyText`](https://github.com/aws/copilot-cli/blob/f6940a431e98b2992bcb0a2fdbfde901580f4c9b/internal/pkg/term/prompt/select.go#L44) right? I think it does but I just wanted to double check.

We use `FriendlyText` in `copilot storage init` at the "lifecycle" prompt.

Lou1415926 (approved):

PR #4622, opened 2023-03-13
chore(manifest): support additional_rules for ALB

### LINKED ISSUE #4589: Do I need target_port with target_container?
Until now, all I had to do was specify target_container and it would flow to the sidecar port beyond it.
```
name: example
type: 'Load Balanced Web Service'

image:
  build:
    dockerfile: example/docker/php/Dockerfile
    target: dev
    context: example
  port: 9000

http:
  path: '/'
  alias: example.com
  targetContainer: 'nginx'

sidecars:
  nginx:
    port: 80
    image: xxxxxxxxxxxxx
```
I think it is from the timing of raising copilot-cli from 1.24.0 to 1.26.0, but the main container (php-fpm:9000port) is now registered in the target group.

Fix
```
http:
  path: '/'
  alias: example.com
  targetContainer: 'nginx'
  target_port: 80    ------> Add
```
So, we resolved the issue by specifying target_port.
In the documentation, target_port is only introduced in the NLB section, is this a correct setting? 
https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#nlb-target-port

paragbhingre:
This PR adds `additional_rules` manifest field to support multiple ports through ALB.

This PR resolves https://github.com/aws/copilot-cli/issues/4589


internal/pkg/manifest/lb_web_svc.go
```diff
@@ -272,8 +274,11 @@ func (lbws *LoadBalancedWebService) ExposedPorts() (ExposedPortsIndex, error) {
 		}
 		exposedPorts = append(exposedPorts, out...)
 	}
-	// port from http.target_port.
-	exposedPorts = append(exposedPorts, lbws.RoutingRule.exposedPorts(exposedPorts, workloadName)...)
+	// port from http.target_port and http.additional_rules[x].target_port
+	for _, rule := range lbws.RoutingRule.ALBRoutingRules() {
+		out := rule.exposedPorts(exposedPorts, workloadName)
+		exposedPorts = append(exposedPorts, out...)
```
Lou1415926: nit:
```suggestion
		exposedPorts = append(exposedPorts, rule.exposedPorts(exposedPorts, workloadName)...)
```

internal/pkg/manifest/svc.go
```diff
@@ -557,8 +557,8 @@ func (s *LoadBalancedWebService) HTTPLoadBalancerTarget() (targetContainer strin
 	targetContainer = aws.StringValue(s.Name)
 	targetPort = s.MainContainerPort()
 
-	rrTargetContainer := s.RoutingRule.TargetContainer
-	rrTargetPort := s.RoutingRule.TargetPort
+	rrTargetContainer := s.RoutingRule.MainRoutingRule.TargetContainer
```
Lou1415926: What do you think of renaming `MainRoutingRule` to `Main` ?

internal/pkg/manifest/validate.go
```diff
@@ -745,36 +765,54 @@ func (CommandOverride) validate() error {
...
+func (r RoutingRuleConfiguration) validate() error {
+	if r.IsEmpty() {
+		return nil
+	}
+	// we consider the fact that primary routing rule is mandatory before you write any additional routing rules.
+	if err := r.MainRoutingRule.validate(); err != nil {
+		return err
+	}
+	// checking this condition here as validate method on ALBRoutingRule only validates common parameters between
+	// MainRoutingRule and AdditionalRoutingRules
```
Lou1415926: nit: not to represent everyone so plz correct me if others think otherwise😂 because `r. TargetContainerCamelCase` is involved, I feel like it should be fairly obvious to readers why this validation does not happen in `func (r ALBRoutingRule) validate() error {`,
```suggestion
```

internal/pkg/manifest/validate.go
```diff
@@ -745,36 +765,54 @@ func (CommandOverride) validate() error {
...
+	if r.MainRoutingRule.TargetContainer != nil && r.TargetContainerCamelCase != nil {
+		return &errFieldMutualExclusive{
+			firstField:  "target_container",
+			secondField: "targetContainer",
+		}
+	}
+
```
Lou1415926, line 783: Do you think we can do something like this to motivate users to move away from the deprecated field: if targetContainer is specified with additional_rules, we error out and ask users to move to  target_container

internal/pkg/manifest/validate.go
```diff
@@ -1929,28 +1962,39 @@ func populateALBPortsAndValidate(containerNameFor map[uint16]string, opts valida
...
 	// is trying to expose the port that is already being exposed by container y, so error out.
-	if opts.alb == nil {
+	if opts.alb == nil || opts.alb.IsEmpty() {
 		return nil
 	}
 	alb := opts.alb
-	if alb.TargetPort == nil {
+	for _, rule := range alb.ALBRoutingRules() {
+		if rule.TargetPort == nil {
+			return nil
```
Lou1415926: Was this meant to be `continue`?

internal/pkg/manifest/validate_test.go
```diff
@@ -1491,7 +1516,135 @@ func TestRoutingRule_validate(t *testing.T) {
...
+			RoutingRuleConfiguration: RoutingRuleConfiguration{
+				MainRoutingRule: ALBRoutingRule{
+					Path:            stringP("/"),
+					TargetContainer: aws.String("mockContainer"),
+				},
+				TargetContainerCamelCase: aws.String("mockContainer"),
+			},
+			wantedError: fmt.Errorf(`must specify one, not both, of "target_container" and "targetContainer"`),
+		},
+		"error if one of allowed_source_ips is not valid for additional routing rules": {
```
Lou1415926: Waht do you think of this instead:
1. Write unit test cases inside of `func TestALBRoutingRule_validate(t *testing.T)` to test if fields like `allowed_source_ips` and `protocol` are errored out
2.  Here, in addition to the first two test cases (confi is nil, targetContainer), just write two more unit test cases here: "error if the main routing rule is invalid", "error it an additional routing rule is invalid".

Lou1415926 (commented):
Looks good!

internal/pkg/manifest/validate.go
paragbhingre: Ohh right. I will make the change.

internal/pkg/initialize/workload_test.go
```diff
@@ -369,7 +369,7 @@ func TestAppInitOpts_createLoadBalancedAppManifest(t *testing.T) {
 				require.Equal(t, tc.inSvcName, aws.StringValue(manifest.Workload.Name))
 				require.Equal(t, tc.inSvcPort, aws.Uint16Value(manifest.ImageConfig.Port))
 				require.Contains(t, tc.inDockerfilePath, aws.StringValue(manifest.ImageConfig.Image.Build.BuildArgs.Dockerfile))
-				require.Equal(t, tc.wantedPath, aws.StringValue(manifest.RoutingRule.Path))
+				require.Equal(t, tc.wantedPath, aws.StringValue(manifest.RoutingRule.MainRoutingRule.Path))
```
efekarakus: what do you think of renaming the `RoutingRule` field to `HTTP` in `manifest.LoadBalancedWebServiceConfig`?

(I'm happy to approve this PR and then submitting a separate one for that) I feel like it's more aligned with the actual YAML fields and would make the code easier to read

internal/pkg/manifest/http.go
```diff
@@ -38,6 +38,11 @@ func (r *RoutingRuleConfigOrBool) UnmarshalYAML(value *yaml.Node) error {
 	if !r.RoutingRuleConfiguration.IsEmpty() {
 		// Unmarshalled successfully to r.RoutingRuleConfiguration, unset r.Enabled, and return.
 		r.Enabled = nil
+		// this assignment lets us treat the main listener rule and additional listener rules equally
+		// because we eliminate the need for TargetContainerCamelCase by assigning its value to TargetContainer.
+		if r.TargetContainerCamelCase != nil {
```
efekarakus: noice, we tested this right with something like `svc package` to make sure it works as expected?

If so maybe we can add a comment to the body of the PR saying that it resolves #4589

internal/pkg/manifest/http.go
```diff
@@ -47,37 +52,47 @@ func (r *RoutingRuleConfigOrBool) UnmarshalYAML(value *yaml.Node) error {
...
-// RoutingRuleConfiguration holds the path to route requests to the service.
+// RoutingRuleConfiguration holds options for application load balancer.
 type RoutingRuleConfiguration struct {
+	MainRoutingRule          ALBRoutingRule   `yaml:",inline"`
+	TargetContainerCamelCase *string          `yaml:"targetContainer"` // "targetContainerCamelCase" for backwards compatibility
+	AdditionalRoutingRules   []ALBRoutingRule `yaml:"additional_rules"`
+}
+
+// ALBRoutingRules returns main as well as additional routing rules as a list of ALBRoutingRule.
+func (cfg RoutingRuleConfiguration) ALBRoutingRules() []ALBRoutingRule {
```
efekarakus: What do you think of renaming to just:
```suggestion
func (cfg RoutingRuleConfiguration) RoutingRules() []ALBRoutingRule {
```
I'm just saying that given that a listener rule is only applicable for an ALB, prefixing it with "ALB" seems redundant 💭

efekarakus (commented):
Looks great! just some tiny naming suggestions 🚀

internal/pkg/manifest/http.go
paragbhingre: Actually I tested it separately if it assigns the value of `TargetContainerCamelCase` to `targetContainer` and that worked well. But now when I tried running with svc package then we are running into error that cannot define both `target_container` and `targetContainer` which we can solve by setting the value of `TargetContainerCamelCase` to nil but it makes me think if we should modify the customer input or not? What do you think about it?

internal/pkg/manifest/http.go
paragbhingre: Do you recommend changing just the name of the function or also changing the name of the struct?

internal/pkg/manifest/http.go
efekarakus: oh go for it! It's okay to change it to `nil` (we don't change the manifest file's value, just our internal representation so it's ok in my book)

internal/pkg/manifest/http.go
efekarakus: both would be great! sorry I should have been more explicit :lolsob:

internal/pkg/initialize/workload_test.go
paragbhingre: Happy to create another PR for it as this PR is already big.

paragbhingre:
@Lou1415926 and @efekarakus I have addressed your feedback except [one](https://github.com/aws/copilot-cli/pull/4622#discussion_r1137416817) from Wanxian, as I am still figuring out few things there.

internal/pkg/manifest/validate.go
paragbhingre: As per discussion with Wanxian, I will address this comment separately in another PR. Feel free to review and leave your comments. 

cc: @Lou1415926 @efekarakus

internal/pkg/manifest/validate_test.go
```diff
@@ -1541,105 +1541,35 @@ func TestRoutingRuleConfiguration_validate(t *testing.T) {
...
-				MainRoutingRule: ALBRoutingRule{
+				Main: RoutingRule{
 					Path:            stringP("/"),
 					TargetContainer: aws.String("mockContainer"),
 				},
 				TargetContainerCamelCase: aws.String("mockContainer"),
 			},
 			wantedError: fmt.Errorf(`must specify one, not both, of "target_container" and "targetContainer"`),
 		},
-		"error if one of allowed_source_ips is not valid for additional routing rules": {
```
Lou1415926: I see these test cases were deleted because of https://github.com/aws/copilot-cli/pull/4622#discussion_r1137435060. Just wanted to make sure I didn't miss anything: were they moved to `func TestALBRoutingRule_validate(t *testing.T)` or were they simply deleted?

internal/pkg/manifest/http.go
```diff
@@ -47,37 +53,47 @@ func (r *RoutingRuleConfigOrBool) UnmarshalYAML(value *yaml.Node) error {
 	return nil
 }
 
-// RoutingRuleConfiguration holds the path to route requests to the service.
+// RoutingRuleConfiguration holds options for application load balancer.
 type RoutingRuleConfiguration struct {
+	Main                     RoutingRule   `yaml:",inline"`
+	TargetContainerCamelCase *string       `yaml:"targetContainer"` // "targetContainerCamelCase" for backwards compatibility
```
Lou1415926: Do we need to mark this as deprecated?
```suggestion
	TargetContainerCamelCase *string       `yaml:"targetContainer"` // Deprecated. kept for backwards compatibility
```

internal/pkg/manifest/http.go
```diff
@@ -38,6 +38,12 @@ func (r *RoutingRuleConfigOrBool) UnmarshalYAML(value *yaml.Node) error {
 	if !r.RoutingRuleConfiguration.IsEmpty() {
 		// Unmarshalled successfully to r.RoutingRuleConfiguration, unset r.Enabled, and return.
 		r.Enabled = nil
+		// this assignment lets us treat the main listener rule and additional listener rules equally
+		// because we eliminate the need for TargetContainerCamelCase by assigning its value to TargetContainer.
+		if r.TargetContainerCamelCase != nil && r.Main.TargetContainer == nil {
+			r.Main.TargetContainer = r.TargetContainerCamelCase
+			r.TargetContainerCamelCase = nil
```
Lou1415926, line 45: If we don't set it as `nil`, it would be easier  to handle https://github.com/aws/copilot-cli/pull/4622#discussion_r1137416817

Lou1415926 (approved):

internal/pkg/manifest/validate_test.go
paragbhingre: Okay so test cases that I removed are already present in the `TestALBRoutingRule_validate`. I added similar test here just to make sure that `additional_rules` also gets tested which it does through `TestALBRoutingRule_validate`. But just to make sure we get proper error message with prefix `validate "additional_rules[0]"` I added similar tests with additional_rules here. Does that make sense? So just removing them is fine.

internal/pkg/manifest/http.go
paragbhingre: If we do not set r.TargetContainerCamelCase to nil then we fall into this [error](https://github.com/aws/copilot-cli/pull/4622/files/9a430d43cc5d7ba8063a3c67d41f0de3aaf7cc0a..1e6437744bcf47652a65e9111b5c89f11ba511be#diff-5fc13795122fe030799a09ef956a8b66cec3d40bf1a81266323edbf3626b5685R777) even when the use case is valid that customer has only provided `targetContainer` and not `target_container`. Does that make sense?

internal/pkg/manifest/backend_svc.go
```diff
@@ -208,7 +208,10 @@ func (b *BackendService) ExposedPorts() (ExposedPortsIndex, error) {
 		}
 		exposedPorts = append(exposedPorts, out...)
 	}
-	exposedPorts = append(exposedPorts, b.RoutingRule.exposedPorts(exposedPorts, workloadName)...)
+	for _, rule := range b.RoutingRule.RoutingRules() {
+		out := rule.exposedPorts(exposedPorts, workloadName)
+		exposedPorts = append(exposedPorts, out...)
```
efekarakus: nit:
```suggestion
		exposedPorts = append(exposedPorts, rule.exposedPorts(exposedPorts, workloadName)...)
```

internal/pkg/manifest/http.go
```diff
@@ -47,37 +53,47 @@ func (r *RoutingRuleConfigOrBool) UnmarshalYAML(value *yaml.Node) error {
 	return nil
 }
 
-// RoutingRuleConfiguration holds the path to route requests to the service.
+// RoutingRuleConfiguration holds options for application load balancer.
 type RoutingRuleConfiguration struct {
+	Main                     RoutingRule   `yaml:",inline"`
+	TargetContainerCamelCase *string       `yaml:"targetContainer"` // Deprecated. "targetContainerCamelCase" for backwards compatibility
```
efekarakus: Wut is this comment malformed? what is "targetContainerCamelCase" should this be:
```suggestion
	TargetContainerCamelCase *string       `yaml:"targetContainer"` // Deprecated. Maintained for backwards compatibility, use [RoutingRule.TargetContainer] instead.
```

efekarakus (approved):

PR #4630, opened 2023-03-15
chore: allow artifact upload without building image

dannyrandall:
Also fixes upload for single files with no configured destination.

### DIFF
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -332,28 +332,28 @@ func (img ContainerImageIdentifier) Tag() string {
 	return img.GitShortCommitTag
 }
 
-func (d *workloadDeployer) uploadContainerImages(imgBuilderPusher imageBuilderPusher) (map[string]ContainerImageIdentifier, error) {
+func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) error {
 	// If it is built from local Dockerfile, build and push to the ECR repo.
 	buildArgsPerContainer, err := buildArgsPerContainer(d.name, d.workspacePath, d.image, d.mft)
 	if err != nil {
-		return nil, err
+		return err
 	}
 	if len(buildArgsPerContainer) == 0 {
-		return nil, nil
+		return nil
 	}
-	images := make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
+	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
 	for name, buildArgs := range buildArgsPerContainer {
-		digest, err := imgBuilderPusher.BuildAndPush(dockerengine.New(exec.NewCmd()), buildArgs)
+		digest, err := d.imageBuilderPusher.BuildAndPush(dockerengine.New(exec.NewCmd()), buildArgs)
 		if err != nil {
-			return nil, fmt.Errorf("build and push image: %w", err)
+			return fmt.Errorf("build and push image: %w", err)
 		}
-		images[name] = ContainerImageIdentifier{
+		out.ImageDigests[name] = ContainerImageIdentifier{
 			Digest:            digest,
 			CustomTag:         d.image.CustomTag,
 			GitShortCommitTag: d.image.GitShortCommitTag,
 		}
 	}
-	return images, nil
+	return nil
 }
 
 func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentifier, unmarshaledManifest interface{}) (map[string]*dockerengine.BuildArguments, error) {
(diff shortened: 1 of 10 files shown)


internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -41,6 +41,8 @@ func Upload(fs afero.Fs, source, destination string, opts *UploadOpts) ([]string
 		for i, f := range files {
 			paths[i] = filepath.Join(source, f.Name())
 		}
+	} else if destination == "" { // only applies to files, not folders
```
huanjani, line 44: Isn't the default destination `assets`?

internal/pkg/cli/deploy/workload.go
```diff
@@ -439,36 +427,36 @@ type UploadArtifactsOutput struct {
...
+// with any relevant information to be returned by uploadArtifacts.
+type uploadArtifactFunc func(out *UploadArtifactsOutput) error
 
-	out := &UploadArtifactsOutput{
-		ImageDigests: imageDigests,
-		EnvFileARNs:  s3Artifacts.envFileARNs,
-		AddonsURL:    s3Artifacts.addonsURL,
+// uploadArtifacts runs each of the uploadArtifact functions sequentially and returns
+// the output built by each of those functions. It short-circuts and returns
+// the error if one of steps returns an error.
```
huanjani, line 436: Would the order ever matter? If an error IS returned, does the user need to know what got uploaded and what didn't?

internal/pkg/cli/deploy/workload.go
```diff
@@ -439,36 +427,36 @@ type UploadArtifactsOutput struct {
...
+func (d *workloadDeployer) uploadArtifacts(steps ...uploadArtifactFunc) (*UploadArtifactsOutput, error) {
+	out := &UploadArtifactsOutput{}
+	for _, step := range steps {
+		if err := step(out); err != nil {
```
huanjani, line 440: this is cool.

internal/pkg/deploy/upload/asset/asset.go
huanjani: oh nvm. that's with no FileUploads.
I am making the default `/` with the selector....

internal/pkg/cli/deploy/workload.go
dannyrandall: right now, i don't think it matters what order each of the uploads happens in? theoretically, we could probably do the uploads in parallel if we wanted. 

as for the uploads - i think it could be nice for the static site to let them know what gets uploaded. however - that actually happens somewhere else lol: https://github.com/aws/copilot-cli/blob/12d5060828892efd017d69fdf5a1885aacfdaf6e/internal/pkg/cli/deploy/static_site.go#L90-L102

internal/pkg/deploy/upload/asset/asset.go
dannyrandall: the default destination in their bucket, for just a file - will be `/<filepath>`. (that's what this is doing)

huanjani: If it's set as `/`, then destination won't be `""` though, right?

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -0,0 +1,31 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+AWSTemplateFormatVersion: 2010-09-09
+Description: CloudFormation template that represents a static site stored in S3.
```
Lou1415926: An alternative is `Description: CloudFormation template that represents a static site backed by S3.`. I don't have preference though!

internal/pkg/deploy/upload/asset/asset.go
Lou1415926: is the `source` an absolute path (it should be right?), if so, would we be uploading to a path that looks like `/Users/admin/Projects/report-site/static/asset/meme.jpg`

dannyrandall: no -- it's actually just a path relative to the `copilot` directory. So given the folder `<repo>/assets/pic.jpg`, `source: assets/pic.jpg` with no destination set would upload to their s3 bucket with the key `/assets/pic.jpg`.

huanjani (approved):
lgtm

internal/pkg/deploy/upload/asset/asset.go
huanjani: "`LocalPath`: represents the path of a local file or directory. It can be written as an absolute path or relative path." https://docs.aws.amazon.com/cli/latest/reference/s3/index.html
And why would it be relative to the `copilot` dir and now the `cwd`? 
Also just a note to keep Windows users in mind as we work on this.

dannyrandall: i think the `cwd` doesn't really make sense for copilot - we're not picky about _where_ in a repo the `copilot` command is ran. i.e., running `cd /repo/cmd/main && copilot deploy` should deploy the same thing as `cd /repo/cmd && copilot deploy` - if we make it relative to `cwd`, the manifest source configuration will only be correct for a single location in the repository. what do you think?

Lou1415926 (approved):

internal/pkg/deploy/upload/asset/asset.go
huanjani: Agree!

PR #4631, opened 2023-03-15
feat(stack): add support for additional_listeners for NLB

paragbhingre:
This PR adds ability to support `additional_listeners` for NLB end to end.  


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -591,52 +591,64 @@ func (conv routingRuleConfigConverter) convert() (*template.ALBListenerRule, err
 	return config, nil
 }
 
+// isCertRequired returns true if any of the NLB listeners have protocol as TLS set.
+func isCertRequired(listeners []template.NetworkLoadBalancerListener) bool {
```
efekarakus: nit: The name of the function is a bit too generic given the parameter types, one might think that the input could be also ALB listeners.
One little trick that I like doing is instead creating methods on type definitions so that the function name is no longer global:
```go
type nlbListeners []template.NetworkLoadBlanacerListener
func (ls nlbListeners) isCertRequired() bool
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -591,52 +591,64 @@ func (conv routingRuleConfigConverter) convert() (*template.ALBListenerRule, err
...
-	if protocol == nil {
-		protocol = aws.String(defaultNLBProtocol)
+		hc := convertNLBHealthCheck(&listener.HealthCheck)
+		listeners = append(listeners, template.NetworkLoadBalancerListener{
+			Port:            aws.StringValue(port),
+			Protocol:        strings.ToUpper(aws.StringValue(protocol)),
+			TargetContainer: targetContainer,
+			TargetPort:      targetPort,
+			SSLPolicy:       listener.SSLPolicy,
+			HealthCheck:     hc,
```
efekarakus:
```suggestion
			HealthCheck:     convertNLBHealthCheck(&listener.HealthCheck),
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -591,52 +591,64 @@ func (conv routingRuleConfigConverter) convert() (*template.ALBListenerRule, err
...
-					HealthCheck:     hc,
-					Stickiness:      nlbConfig.Listener.Stickiness,
-				},
-			},
+			PublicSubnetCIDRs:   s.publicSubnetCIDRBlocks,
+			Listener:            listeners,
+			Aliases:             aliases,
 			MainContainerPort:   s.manifest.MainContainerPort(),
-			CertificateRequired: strings.ToUpper(aws.StringValue(protocol)) == manifest.TLS,
+			CertificateRequired: isCertRequired(listeners),
```
efekarakus: wouldn't the certificate required properly be at a listener level?

Like each listener would only get a `CertificateArn:` if the target_port has a `/tls` protocol 💭

internal/pkg/manifest/svc.go
```diff
@@ -664,32 +664,45 @@ func prepareParsedExposedPortsMap(exposedPorts []ExposedPort) (map[string][]Expo
 
 // NetworkLoadBalancerTarget returns target container and target port for the NLB configuration.
 // This method should be called only when NLB config is not empty.
-func (s *LoadBalancedWebService) NetworkLoadBalancerTarget() (targetContainer string, targetPort string, err error) {
-	// Configure target container and port.
-	targetContainer = aws.StringValue(s.Name)
-	if s.NLBConfig.Listener.TargetContainer != nil {
-		targetContainer = aws.StringValue(s.NLBConfig.Listener.TargetContainer)
+func (s *LoadBalancedWebService) NetworkLoadBalancerTarget(lContainer string, lPort *int, nlbPort string) (targetContainer string, targetPort string, err error) {
```
efekarakus: If I'm following correctly, we are passing the listener properties that are already in the `s *LoadBalancedWebService` to this method. 

Can we instead have this method at the listener level? and have as input the map of ports/container pairs 
```go
func (l NetworkLoadBalancerListener) Target(exposedPorts map[uint16]string) (containerName string, containerPort string, err error)
```

internal/pkg/manifest/svc.go
```diff
@@ -664,32 +664,45 @@ func prepareParsedExposedPortsMap(exposedPorts []ExposedPort) (map[string][]Expo
...
 		}
-		targetPort = aws.StringValue(port)
+		return
 	}
-	// Finally, if a target port is explicitly specified, use that value.
-	if s.NLBConfig.Listener.TargetPort != nil {
-		targetPort = strconv.Itoa(aws.IntValue(s.NLBConfig.Listener.TargetPort))
+
+	if lContainer == "" { // when target_container is nil
+		container, port := httpLoadBalancerTarget(exposedPorts, uint16P(uint16(aws.IntValue(lPort))))
```
efekarakus: do we need to rename this function? given that NLBs don't know about HTTP

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -591,52 +591,64 @@ func (conv routingRuleConfigConverter) convert() (*template.ALBListenerRule, err
...
 	if nlbConfig.IsEmpty() {
 		return networkLoadBalancerConfig{}, nil
 	}
 
-	// Parse targetContainer and targetPort for the Network Load Balancer targets.
-	targetContainer, targetPort, err := s.manifest.NetworkLoadBalancerTarget()
-	if err != nil {
-		return networkLoadBalancerConfig{}, err
-	}
+	var listeners []template.NetworkLoadBalancerListener
```
Lou1415926: tiny nit:
```suggestion
  listeners := make([]template.NetworkLoadBalancerListener, len(nlbConfig.NLBListeners()))
```

internal/pkg/deploy/cloudformation/stack/transformers.go
Lou1415926: Out of curiosity, do we assume all aliases under `nlb.alias` need to be certified? Do all TLS listeners use the same certificates?

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: Okay so the philosophy behind this logic is we check for all the listeners if any of them have `tls` mentioned in the `port`. If we get at least one `port` with `tls` on then we set this flag to `true`. Once this flag is on then we create `NLBCertValidatorAction`.  The listeners still decide if they want cert or not based on if they have `tls` mentioned in the `port`. So `CertificateRequired` only defines if we should trigger `NLBCertValidatorAction`? and `tls` value in the `port` decides that listener needs that certificate. Does that make sense?

Just to be sure we are on the same page @efekarakus, did you mean `port` instead of target_port has `/tls`?

internal/pkg/manifest/svc.go
paragbhingre: We could have it but there are multiple different things that we need here to make a final decision about `targetContainer` and `targetPort`.  Firstly we need  `exposedPorts` that we are planning to pass as a method input. secondly, we need `workload` because we assign it as default `targetContainer`. Thirdly, we need `s.Sidecars` in case we only have `targetContainer` mentioned and if it is sidecar container then we need to pull port of the sidecar container to set it as `targetPort`. So if we decide to set this method at listener level then we will anyway need to send these parameters as input to this method. Do you have any thoughts on this?

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: AFAICT, we all the aliases to be certified then only customers will be redirected to the NLB from those aliases. And all the users will use the same certificate as I do not see the need of having all the listeners their own certificate with all the aliases certified as a part of it. Do let me know if you have any use case in mind that contradicts my understanding.

internal/pkg/manifest/svc.go
```diff
@@ -628,7 +632,8 @@ func (s *BackendService) HTTPLoadBalancerTarget() (targetContainer string, targe
 	return
 }
 
-func httpLoadBalancerTarget(exposedPorts ExposedPortsIndex, rrTargetPort *uint16) (targetContainer *string, targetPort *string) {
+// targetContainerFromTargetPort returns target container and target port from the given target_port input.
+func targetContainerFromTargetPort(exposedPorts ExposedPortsIndex, rrTargetPort *uint16) (targetContainer *string, targetPort *string) {
```
Lou1415926: nit: I supposed `rr` standed for routing rule? 
```suggestion
func targetContainerFromTargetPort(exposedPorts ExposedPortsIndex, targetPort *uint16) (targetContainer *string, targetPort *string) {
```

internal/pkg/manifest/svc.go
```diff
@@ -662,34 +667,46 @@ func prepareParsedExposedPortsMap(exposedPorts []ExposedPort) (map[string][]Expo
 	return parsedContainerMap, parsedExposedPortMap
 }
 
-// NetworkLoadBalancerTarget returns target container and target port for the NLB configuration.
+// Target returns target container and target port for the NLB Listener configuration.
 // This method should be called only when NLB config is not empty.
-func (s *LoadBalancedWebService) NetworkLoadBalancerTarget() (targetContainer string, targetPort string, err error) {
+func (l NetworkLoadBalancerListener) Target(exposedPorts ExposedPortsIndex) (targetContainer string, targetPort string, err error) {
```
Lou1415926: nit:
```suggestion
func (listener NetworkLoadBalancerListener) Target(exposedPorts ExposedPortsIndex) (targetContainer string, targetPort string, err error) {
```

internal/pkg/manifest/svc.go
```diff
@@ -662,34 +667,46 @@ func prepareParsedExposedPortsMap(exposedPorts []ExposedPort) (map[string][]Expo
...
 	// Configure target container and port.
-	targetContainer = aws.StringValue(s.Name)
-	if s.NLBConfig.Listener.TargetContainer != nil {
-		targetContainer = aws.StringValue(s.NLBConfig.Listener.TargetContainer)
-	}
-
+	targetContainer = exposedPorts.WorkloadName
```
Lou1415926: Can we move this to L682?

internal/pkg/manifest/workload_ecs.go
```diff
@@ -131,6 +131,7 @@ type ExposedPort struct {
 	ContainerName string // The name of the container that exposes this port.
 	Port          uint16 // The port number.
 	Protocol      string // Either "tcp" or "udp", empty means the default value that the underlying service provides.
+	Implicit      bool   // Defines if the port is the main port of the container. defaults to false.
```
Lou1415926: Why is this named `Implicit`?

`Explicit` seemed to make more sense to me 😂  That "this port is explicitly exposed via 'image.port' , 'sidecar.port', etc."

internal/pkg/manifest/svc.go
```diff
@@ -662,34 +667,46 @@ func prepareParsedExposedPortsMap(exposedPorts []ExposedPort) (map[string][]Expo
...
+	if l.TargetContainer == nil && l.TargetPort == nil { // both targetPort and targetContainer are nil.
+		return
+	}
+
+	if l.TargetPort == nil { // when target_port is nil
+		if aws.StringValue(l.TargetContainer) != exposedPorts.WorkloadName {
+			targetContainer = aws.StringValue(l.TargetContainer)
+			for _, portConfig := range exposedPorts.PortsForContainer[targetContainer] {
+				if portConfig.Implicit {
+					targetPort = strconv.Itoa(int(portConfig.Port))
```
Lou1415926, line 690: We probably need a comment here.
```suggestion
					targetPort = strconv.Itoa(int(portConfig.Port)) 
					/* NOTE: When the `target_port` is empty, the intented target port should be the port that is explictly exposed by the container. Consider the following example
					```
  					http:
  					   target_container: nginx
 					   target_port: 83 # Implicitly exposed by the nginx container
					nlb:
					  port: 80/tcp
					  target_container: nginx
					 sidecars:
					   nginx:
					     port: 81 # Explicitly exposed by the nginx container.
					 ```					 
 					 In this example, the target port for the NLB listener should be 81 
					*/
					
```

internal/pkg/manifest/svc.go
```diff
@@ -662,34 +667,46 @@ func prepareParsedExposedPortsMap(exposedPorts []ExposedPort) (map[string][]Expo
...
+		return
 	}
-	// Finally, if a target port is explicitly specified, use that value.
-	if s.NLBConfig.Listener.TargetPort != nil {
-		targetPort = strconv.Itoa(aws.IntValue(s.NLBConfig.Listener.TargetPort))
+
+	if l.TargetContainer == nil { // when target_container is nil
+		container, port := targetContainerFromTargetPort(exposedPorts, uint16P(uint16(aws.IntValue(l.TargetPort))))
+		targetPort = aws.StringValue(port)
+		if container != nil {
```
Lou1415926, line 715: This normally isn't expected to happen  right? Can we add a comment that says so?

internal/pkg/manifest/workload_ecs.go
paragbhingre: Yeah, I changed it. Also me and Efe decided to call this as `IsContainerDefinedPort`. So updating changes in the upcoming commit.

internal/pkg/manifest/svc.go
```diff
@@ -534,10 +534,10 @@ func (sidecar SidecarConfig) exposedPorts(sidecarName string) ([]ExposedPort, er
...
 	return []ExposedPort{
 		{
-			Port:          uint16(port),
-			Protocol:      strings.ToLower(protocol),
-			ContainerName: sidecarName,
-			Implicit:      true,
+			Port:                   uint16(port),
+			Protocol:               strings.ToLower(protocol),
+			ContainerName:          sidecarName,
+			IsContainerDefinedPort: true,
```
Lou1415926: 😂 don't hate me but what about `isDefinedByContainer`, so that it reads:
```go
for _, portConfig := range exposedPorts.PortsForContainer[targetContainer] {
  if portConfig.isDefinedByContainer
```

also, it seems like this can be a private field, is that correct?

internal/pkg/manifest/svc.go
```diff
@@ -669,44 +669,57 @@ func prepareParsedExposedPortsMap(exposedPorts []ExposedPort) (map[string][]Expo
...
-			targetContainer = aws.StringValue(l.TargetContainer)
+	if listener.TargetPort == nil { // when target_port is nil
+		if aws.StringValue(listener.TargetContainer) != exposedPorts.WorkloadName {
+			targetContainer = aws.StringValue(listener.TargetContainer)
 			for _, portConfig := range exposedPorts.PortsForContainer[targetContainer] {
-				if portConfig.Implicit {
+				if portConfig.IsContainerDefinedPort {
 					targetPort = strconv.Itoa(int(portConfig.Port))
+					/* NOTE: When the `target_port` is empty, the intended target port should be the port that is explicitly exposed by the container. Consider the following example
+										```
```
Lou1415926: Can we un-indent these examples to align with `NOTE`?

internal/pkg/manifest/svc.go
```diff
@@ -662,34 +667,59 @@ func prepareParsedExposedPortsMap(exposedPorts []ExposedPort) (map[string][]Expo
 	return parsedContainerMap, parsedExposedPortMap
 }
 
-// NetworkLoadBalancerTarget returns target container and target port for the NLB configuration.
+// Target returns target container and target port for the NLB Listener configuration.
```
Lou1415926: nit:
```suggestion
// Target returns target container and target port for a NLB listener configuration.
```

Lou1415926 (commented):
I will review the `stack` package tomorrow but the rest LGTM!

Lou1415926 (approved):

efekarakus (approved):

PR #4632, opened 2023-03-15
refactor: differentiate diff node types

Lou1415926:

### DIFF
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -12,24 +12,54 @@ import (
 )
 
 // Node represents a segment on a difference between two YAML documents.
-type Node struct {
-	key      string
-	children map[string]*Node // A list of non-empty pointers to the children nodes.
+type Node interface {
+	String() string
 
-	oldValue *yaml.Node // Only populated for a leaf node (i.e. that has no child node).
-	newValue *yaml.Node // Only populated for a leaf node (i.e. that has no child node).
+	key() string
+	newValue() *yaml.Node
+	oldValue() *yaml.Node
+	children() map[string]Node
+}
+
+type basicNode struct {
+	keyV       string
+	childNodes map[string]Node // A list of non-empty pointers to the children nodes.
+
+	oldV *yaml.Node // Only populated for a leaf node (i.e. that has no child node).
+	newV *yaml.Node // Only populated for a leaf node (i.e. that has no child node).
+}
+
+func (n *basicNode) key() string {
+	return n.keyV
+}
+
+func (n *basicNode) newValue() *yaml.Node {
+	return n.newV
+}
+
+func (n *basicNode) oldValue() *yaml.Node {
+	return n.oldV
 }
 
 // String returns the string representation of the tree stemmed from the diffNode n.
-func (n *Node) String() string {
+func (n *basicNode) children() map[string]Node {
+	return n.childNodes
+}
+
+// String returns the string representation of the tree stemmed from the diffNode n.
+func (n *basicNode) String() string {
 	return ""
+}
+
+type seqNode struct {
+	basicNode
 }
 
 // From is the YAML document that another YAML document is compared against.
 type From []byte
 
 // Parse constructs a diff tree that represent the differences of a YAML document against the From document.
-func (from From) Parse(to []byte) (*Node, error) {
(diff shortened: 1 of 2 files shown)


internal/pkg/template/diff/diff.go
```diff
@@ -6,63 +6,83 @@ package diff
...
-	"strconv"
 
 	"gopkg.in/yaml.v3"
 )
 
 // Node represents a segment on a difference between two YAML documents.
-type Node struct {
-	key      string
-	children map[string]*Node // A list of non-empty pointers to the children nodes.
+type Node interface {
```
efekarakus: It's a bit odd to have a public interface that has only private methods 🤔 Have we seen something like this in the std lib?
What do you think of an API like this:
```go
// Tree represents a difference tree between two YAML documents.
type Tree struct {
   root diffNode
}

// comparableNode is the interface to represents the difference between two *yaml.Node.
type comparableNode interface {
   key() string
   // your other methods...
}

// node is a concrete implementation of a comparableNode.
type node struct {
   keyValue string
   // your other fields...
} 

func (from From) Parse(to []byte) (*Tree, error)
```

internal/pkg/template/diff/diff.go
```diff
@@ -113,7 +133,10 @@ func parseSequence(fromNode, toNode *yaml.Node) (map[string]*Node, error) {
 		}
 		return err == nil && diff == nil
 	})
-	nextKey, children, inspector := seqChildKeyFunc(), make(map[string]*Node), newLCSStateMachine(fromSeq, toSeq, lcsIndices)
+	if len(fromSeq) == len(toSeq) && len(lcsIndices) == len(fromSeq) {
```
efekarakus, line 152: Can we add a comment here 🥺 
```suggestion
	// No difference if the two sequences have the same size and the LCS is the entire sequence.
	if len(fromSeq) == len(toSeq) && len(lcsIndices) == len(fromSeq) {
```

internal/pkg/template/diff/diff.go
```diff
@@ -124,28 +147,43 @@ func parseSequence(fromNode, toNode *yaml.Node) (map[string]*Node, error) {
 			if diff.err != nil {
 				return nil, diff.err
 			}
-			children[nextKey()] = diff.node
+			children = append(children, &seqItemNode{
+				node: node{
```
efekarakus: nit: why is the syntax different here than the rest of the case statements?
```suggestion
				node{
```

internal/pkg/template/diff/diff.go
```diff
@@ -124,28 +147,43 @@ func parseSequence(fromNode, toNode *yaml.Node) (map[string]*Node, error) {
...
+func parseMap(from, to *yaml.Node) ([]Node, error) {
 	currMap, oldMap := make(map[string]yaml.Node), make(map[string]yaml.Node)
 	if err := to.Decode(currMap); err != nil {
 		return nil, err
 	}
 	if err := from.Decode(oldMap); err != nil {
 		return nil, err
 	}
-	children := make(map[string]*Node)
+	children := make([]Node, 0)
```
efekarakus:
```suggestion
	var children []Node
```

efekarakus (commented):
looks good just some tiny recommendations

internal/pkg/template/diff/diff.go
Lou1415926: I was experimenting with having a `context *yaml.Node` field for a `seqItemNode` to address https://github.com/aws/copilot-cli/pull/4611#discussion_r1132963408. Since `node` wouldn't be the only field in `seqItemNode`, the `node: ` was needed. I guess that change was somehow picked into this branch lol

internal/pkg/template/diff/diff.go
```diff
@@ -6,63 +6,98 @@ package diff
...
+
+// diffNode is the interface to represents the difference between two *yaml.Node.
+type diffNode interface {
+	key() string
+	newValue() *yaml.Node
+	oldValue() *yaml.Node
+	children() []diffNode
+}
+
+// node is a concrete implementation of a comparableNode.
```
efekarakus:
```suggestion
// node is a concrete implementation of a diffNode.
```

efekarakus (approved):

dannyrandall (approved):

PR #4633, opened 2023-03-15
chore: fix deduping imports in "stack.ts" for cdk overrides

efekarakus:

### DIFF
--- a/internal/pkg/template/overrides.go
+++ b/internal/pkg/template/overrides.go
@@ -5,6 +5,7 @@ package template
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 	"unicode"
 )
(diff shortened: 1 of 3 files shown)


internal/pkg/template/overrides.go
```diff
@@ -43,19 +44,37 @@ type CFNResource struct {
...
-			uniqueTypes = append(uniqueTypes, r.Type)
+		serviceName := strings.Split(strings.ToLower(string(r.Type)), "::")[1]
+		if _, ok := seen[serviceName]; ok {
+			continue
 		}
-		seen[r.Type] = struct{}{}
+		seen[serviceName] = r.Type
+	}
+
+	imports := make([]CDKImport, len(seen))
```
paragbhingre, line 68: How about adding this before for loop and inside for loop just add `imports = append(imports, r.Type)`. This will save extra for logic to create our resultant slice.

paragbhingre (approved):
Looks good to me! feel free to remove label after addressing small request.

internal/pkg/template/overrides.go
```diff
@@ -43,19 +44,37 @@ type CFNResource struct {
...
-	seen := make(map[CFNType]struct{})
+// Imports returns a list of CDK imports for a given list of CloudFormation resources.
+func (rs cfnResources) Imports() []CDKImport {
+	// Find a unique CFN type per service.
+	// For example, given "AWS::ECS::Service" and "AWS::ECS::TaskDef" we want to retain only one of them.
+	seen := make(map[string]CFNType)
 	for _, r := range rs {
-		if _, ok := seen[r.Type]; !ok {
-			uniqueTypes = append(uniqueTypes, r.Type)
+		serviceName := strings.Split(strings.ToLower(string(r.Type)), "::")[1]
```
dannyrandall, line 61: panic if invalid type? should we check length of split?

dannyrandall (approved):

internal/pkg/template/overrides.go
efekarakus: :rip: sadly if I want to use `append` I need to use `var imports []CDKImport` rather than initializing it with a size ahead of time

https://go.dev/play/p/q6LfsDBih3y

internal/pkg/template/overrides.go
efekarakus: For visibility, I made this assumption everything in this file :lolsob: because of this:
> Resource type identifiers **always** take the following form:
> `service-provider::service-name::data-type-name`
> https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html

I know our `Custom::` types are not of that format, but we never pass them down here (although yea it makes the code a bit more fragile)

PR #4634, opened 2023-03-15
chore: implements YAML diff writer for `map` and scalar `seq`

Lou1415926:
That writer works for `map` node, `scalar` node and  `seq` nodes whose content are not `map`.

The writer is on parity with the differ in `diff.go`. The to do for both are the same:
1. Handle "(x unchanged items)"
2. Handle `seq` with `map` content


internal/pkg/template/diff/write.go
Lou1415926, line 1: Although this PR needs to be rebased after #4632 is merged, and has to be merged after that one - the logics in `formatter.go` and `write.go` should be reviewable already!

Edit: rebased1!!!

internal/pkg/template/diff/diff.go
```diff
@@ -6,63 +6,98 @@ package diff
...
+	key() string
+	newValue() *yaml.Node
+	oldValue() *yaml.Node
+	children() []diffNode
+}
+
+// node is a concrete implementation of a comparableNode.
+type node struct {
+	keyValue   string
+	childNodes []diffNode // A list of non-empty pointers to the children nodes.
```
dannyrandall: just kinda laughing at these names as i was reading through the PR... too bad we can't use the function names (`key`/`children`)

internal/pkg/template/diff/formatter.go
dannyrandall, line 1: i'm assuming tests will come for this file so i won't hold up this PR for that - just will help in seeing what the formats look like :)

internal/pkg/template/diff/write.go
```diff
@@ -0,0 +1,105 @@
...
+			return err
+		}
+		return s.writeInsert(node, indent, formatter)
+	}
+	content := processMultiline(formatter.formatMod(node), prefixBy(prefixMod), indentBy(indent))
+	_, err := s.writer.Write([]byte(content + "\n"))
+	return err
+}
+
+func (s *Writer) writeDel(node diffNode, indent int, formatter formatter) error {
```
dannyrandall: any chance we could use `color` to highlight delete's red and insert's green (or something?)

dannyrandall (approved):

internal/pkg/template/diff/formatter.go
Lou1415926: do you think we need one 👀  i was hoping `write_test.go` is enough to cover everything that we care about 😂

internal/pkg/template/diff/formatter.go
```diff
@@ -0,0 +1,91 @@
...
+
+type formatter interface {
+	formateYAML(*yaml.Node) ([]byte, error)
+	formatMod(node diffNode) string
+	formatPath(node diffNode) string
+}
+
+type seqItemFormatter struct{}
+
+func (f *seqItemFormatter) formateYAML(node *yaml.Node) ([]byte, error) {
```
efekarakus: 😂 
```suggestion
func (f *seqItemFormatter) formatYAML(node *yaml.Node) ([]byte, error) {
```

internal/pkg/template/diff/formatter.go
```diff
@@ -0,0 +1,91 @@
...
+				Value: f.key,
+			},
+			node,
+		},
+	}
+	return yaml.Marshal(wrapped)
+}
+
+func (f *keyedFormatter) formatMod(node diffNode) string {
+	return fmt.Sprintf("%s: %s -> %s", node.key(), node.oldValue().Value, node.newValue().Value)
```
efekarakus: nit: This makes me think that maybe we should rename the methods to this? 💭 
```suggestion
	return fmt.Sprintf("%s: %s -> %s", node.key(), node.oldYAML().Value, node.newYAML().Value)
```

internal/pkg/template/diff/formatter.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (f *keyedFormatter) formatPath(node diffNode) string {
+	return node.key() + ":"
+}
+
+func prefixBy(prefix string) func(line string) string {
```
efekarakus: nit:
```suggestion
func prefixByFn(prefix string) func(line string) string {
```

internal/pkg/template/diff/formatter.go
```diff
@@ -0,0 +1,91 @@
...
+	return func(line string) string {
+		return fmt.Sprintf("%s %s", prefix, line)
+	}
+}
+
+func indentBy(count int) func(line string) string {
```
efekarakus:
```suggestion
func indentByFn(count int) func(line string) string {
```

internal/pkg/template/diff/write.go
```diff
@@ -0,0 +1,105 @@
...
+	prefixMod = "~"
+)
+
+const indentInc = 4
+
+// Writer writes the string representation of a diff tree.
+type Writer struct {
+	tree   Tree
+	writer io.Writer
+}
```
efekarakus: Out of curiosity, was there a reason why you didn't prefer `Write` to be a method on `Tree`?
e.g.
```go
// Write writes a visual representation of the diff tree to w.
func (t *Tree) Write(w io.Writer) error
```

efekarakus (commented):
I haven't finished reviewing but I'm publishing what I have so far!

internal/pkg/template/diff/formatter.go
dannyrandall: oh good point lol. prob fine without!

internal/pkg/template/diff/write.go
Lou1415926: I thought of `Tree` as the data storage structure, and how to render the data is a different responsibility. However, it could on the other hand be argued that rendering is also a part of the spec that the tree has to provide. I feel like there isn't a clear winner which is better, and I don't have a super concrete reason for going the way I chose. The real thought process was that when I realized I'd need to pass `w io.Writer` around in all these private functions, I thought maybe it's a sign that it should have been a struct on its own 😂

efekarakus: oh it's fine, I was just trying to think if we can reduce the API surface of the package. We can still have a struct like this:
```

func (t *Tree) Write(w io.Writer) error {
  tw := &treeWriter{t, w}
  return tw.Write()
}

type treeWriter struct {
   tree *Tree
   out   io.Writer
}
```

internal/pkg/template/diff/write.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Write uses the writer to write the string representation of the diff tree stemmed from the root.
+func (s *Writer) Write() error {
+	if s.tree.root == nil {
+		_, err := s.writer.Write([]byte("No changes.\n"))
```
efekarakus: I wonder if we should instead just no write anything? And let the client decide how to display info back.

For example, here is what I'm imagining happens in the `cli` pkg:
```go
sb := new(strings.Builder)
tree, err := diff.Parse(deployedTemplate, pkgTemplate)
if err != nil { return err }
if err := tree.Write(sb); err != nil { return err }

out := sb.String()
if len(out) == "" {
   log.Info("There are no changes between the deployed stack and the local manifest for %q", o.name)
}
fmt.Println(out)
```

internal/pkg/template/diff/write.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+func (s *Writer) writeMod(node diffNode, indent int, formatter formatter) error {
+	if node.oldYAML().Kind != node.newYAML().Kind {
+		if err := s.writeDel(node, indent, formatter); err != nil {
+			return err
+		}
+		return s.writeInsert(node, indent, formatter)
+	}
```
efekarakus, line 82: this is so nice 🤩

internal/pkg/template/diff/write_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+func Test_Integration_Parse_Write(t *testing.T) {
+	testCases := map[string]struct {
+		curr        string
+		old         string
+		wanted      string
+		wantedError error
+	}{
+		"add a map": {
+			curr: `Mary:
```
efekarakus: nit: would you mind starting the multiline strings with a new line, to make it easier to see the end outcome e.g.
```suggestion
			curr: `
Mary:
```

internal/pkg/template/diff/write_test.go
```diff
@@ -0,0 +1,211 @@
...
+        ~ cm: 190 -> 168
+`,
+		},
+		"list does not change": {
+			old:    `Alphabet: [a,b,c,d]`,
+			curr:   `Alphabet: [a,b,c,d]`,
+			wanted: "No changes.\n",
+		},
+		"list reordered": {
+			// TODO(lou1425926): complete the test.
```
efekarakus, line 75: Isn't this testcase captured by "list with a scalar value changed" below?

internal/pkg/template/diff/write_test.go
```diff
@@ -0,0 +1,211 @@
...
+			curr: `DanceCompetition: [dog,bear,mouse,cat]`,
+			wanted: `~ DanceCompetition:
+    + - mouse
+`,
+		},
+		"list with deletion": {
+			old:  `PotatoChipCommittee: [dog,bear,cat,mouse]`,
+			curr: `PotatoChipCommittee: [dog,bear,mouse]`,
+			wanted: `~ PotatoChipCommittee:
+    - - cat
```
efekarakus, line 92: we'll have the "# of unmodified items" later right?

efekarakus (approved):

internal/pkg/template/diff/write_test.go
Lou1415926: yeahhh I left it here just in case I have the time to implement the logic that writes ⬇️  😂 I think of this as a nice to have though 😂 
```
bear(moved from 2 to 3), dog (moved from 3 to 2)
```

(not important but instead of "scalar value changed", it'd actually be an insertion (at the new position)  + a deletion (from the old position))

internal/pkg/template/diff/write_test.go
Lou1415926: yessssss

PR #4640, opened 2023-03-17
chore: parse and write the diff of `map` that are items in a `seq`

Lou1415926:

### DIFF
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -11,6 +11,11 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+// For a property whose value is a list of map, map its name to the identifying property of the items.
+var mapItemsIdentifiedBy = map[string]string{
+	"ContainerDefinitions": "Name",
+}
+
 // Tree represents a difference tree between two YAML documents.
 type Tree struct {
 	root diffNode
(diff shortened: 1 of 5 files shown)


internal/pkg/template/diff/diff.go
```diff
@@ -11,6 +11,11 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+// For a property whose value is a list of map, map its name to the identifying property of the items.
+var mapItemsIdentifiedBy = map[string]string{
+	"ContainerDefinitions": "Name",
```
dannyrandall, line 16: niceeeeee

internal/pkg/template/diff/diff_test.go
```diff
@@ -336,6 +363,13 @@ Dog: (pleased) "ikr"`, t),
 	}
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
+			ogItemsIdentifiedBy := mapItemsIdentifiedBy
+			defer func() {
+				mapItemsIdentifiedBy = ogItemsIdentifiedBy
+			}()
+			mapItemsIdentifiedBy = map[string]string{
+				"StrawberryPopularitySurvey": "Name",
```
dannyrandall, line 371: so random lol :joy:

internal/pkg/template/diff/diff.go
dannyrandall: though i wonder (nit) if something like `seqObjectIdentifiers` indicates better that it's only used to identify sequence objects? i don't feel too strongly abt it

dannyrandall (approved):
for the survey: i like strawberries

Lou1415926:
Closing this PR for now. Rendering "implicit key" is just a bonus. The `(x unchanged items)` line in the diff is already enough for users to identify the items that have changed.


This would be a nice to have though. I will(might) reopen the PR in the future if I find a better implementation solution.

PR #4641, opened 2023-03-17
feat: yaml patch override

dannyrandall:


internal/pkg/cli/deploy/rdws.go
```diff
@@ -176,7 +176,7 @@ func (d *rdwsDeployer) stackConfiguration(in *StackRuntimeConfiguration) (*rdwsS
 	}
 	return &rdwsStackConfigurationOutput{
 		svcStackConfigurationOutput: svcStackConfigurationOutput{
-			conf: conf,
+			conf: cloudformation.WrapWithTemplateOverrider(conf, d.overrider),
```
efekarakus, line 179: :doge: was this a miss from my end?

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,253 @@
...
+		var err error
+		switch patch.Operation {
+		case "add":
+			err = patch.doAdd(&root)
+		case "remove":
+			err = patch.doRemove(&root)
+		case "replace":
+			err = patch.applyReplace(&root)
+		default:
+			return nil, fmt.Errorf("unsupported operation %q", patch.Operation)
```
efekarakus: nit: should we including the valid set of operations as hint here? perhaps with `RecommendActions` 💭

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,253 @@
...
+)
+
+// Patch applies overrides configured as JSON Patches,
+// as defined in https://www.rfc-editor.org/rfc/rfc6902.
+type Patch struct {
+	rootAbsPath string   // Absolute path to the overrides/ directory.
+	fs          afero.Fs // OS file system.
+}
+
+type PatchOpts struct {
```
efekarakus, line 25: Doc comments missing for `PatchOpts`, `WithPatch` and `Override` :)

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,253 @@
...
+		return nil, fmt.Errorf("invalid template: %w", err)
+	}
+
+	for _, patch := range patches {
+		var err error
+		switch patch.Operation {
+		case "add":
+			err = patch.doAdd(&root)
+		case "remove":
+			err = patch.doRemove(&root)
```
efekarakus: nit: what do you think of renaming this like 👇  to be consistent w 
```go
	err = patch.applyAdd(&root)
case "remove":
	err = patch.applyRemove(&root)
```
Alternatively, we can pull out the `visit` functions and do something like:
```go
	err = patch.apply(&root, addLeafFn)
case "remove":
	err = patch.apply(&root, removeLeafFn)
```
where 
```go
func addLeadFn(node *yaml.Node, pointer []string) error { ... }
```

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,253 @@
...
+		}
+		return err
+	} else if len(pointer) == 0 {
+		return nil
+	}
+
+	switch node.Kind {
+	case yaml.DocumentNode:
+		if len(node.Content) != 1 {
+			return fmt.Errorf("don't support multi-doc yaml")
```
efekarakus: What does this mean 😅

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,253 @@
...
+	return errStopFollowingPointer
+}
+
+var errStopFollowingPointer = errors.New("stop following pointer")
+
+func followJSONPointer(root *yaml.Node, pointer string, visit func(node *yaml.Node, pointer []string) error) error {
+	split := strings.Split(pointer, "/")
+	for i := range split {
+		split[i] = strings.ReplaceAll(split[i], "~0", "~")
+		split[i] = strings.ReplaceAll(split[i], "~1", "/")
```
efekarakus: nit: Would you mind adding a comment to section 3. of https://www.rfc-editor.org/rfc/rfc6901 to explain these characters :)

They also recommend doing the opposite sequence 😂 
> This is performed by first transforming any occurrence of the sequence '\~1' to '/', and then transforming any occurrence of the sequence '\~0' to '\~'.

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,253 @@
...
+	return followJSONPointer(root, y.Path, func(node *yaml.Node, pointer []string) error {
+		if len(pointer) != 1 {
+			return nil
+		}
+
+		switch node.Kind {
+		case yaml.MappingNode:
+			for i := 0; i < len(node.Content); i += 2 {
+				if node.Content[i].Value == pointer[0] {
+					node.Content = append(node.Content[:i], node.Content[i+2:]...)
```
efekarakus: will this work if we try to remove the last key in a map? (i.e. can we do `node.Content[i+2:]`?)

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,253 @@
...
+		}
+
+		return nil
+	})
+}
+
+func (y *yamlPatch) doRemove(root *yaml.Node) error {
+	return followJSONPointer(root, y.Path, func(node *yaml.Node, pointer []string) error {
+		if len(pointer) != 1 {
+			return nil
```
efekarakus: I wonder if the `nil` and `errStopFollowingPointer` pointer relationship should be reversed. For example,

```go
if len(pointer) > 1 {
   return errSkip
}
// ...
return nil  // if successfully added, removed, or replaced an element.
```
(inspired by https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/io/fs/walk.go;l=15)

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,253 @@
...
+	split := strings.Split(pointer, "/")
+	for i := range split {
+		split[i] = strings.ReplaceAll(split[i], "~0", "~")
+		split[i] = strings.ReplaceAll(split[i], "~1", "/")
+	}
+
+	return followJSONPointerHelper(root, split, visit)
+}
+
+func followJSONPointerHelper(node *yaml.Node, pointer []string, visit func(node *yaml.Node, pointer []string) error) error {
```
efekarakus: I wonder if the `visit` function should take as input the subpath traveled so far that in our errors we can provide exactly where the issue occurs?
```suggestion
func followJSONPointerHelper(node *yaml.Node, pointer []string, visit func(node *yaml.Node, traversed, remaining []string) error) error {
```
so that we can do something like:
```go
if err := visit(node, pointer); err != nil && !errors.Is(err, errSkip) {
   return fmt.Errorf("visit %q: %v", strings.Join(traversed, jsonPathSeparator), err)
}

// and further down
return fmt.Errorf("visit %q: key %q not found in map", strings.Join(traversed, jsonPathSeparator),pointer[0])
```

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,253 @@
...
+
+func (p *Patch) unmarshalPatches() ([]yamlPatch, error) {
+	var patches []yamlPatch
+
+	files, err := afero.ReadDir(p.fs, p.rootAbsPath)
+	if err != nil {
+		return nil, fmt.Errorf("read directory %q: %w", p.rootAbsPath, err)
+	}
+
+	for _, file := range files {
```
efekarakus: What do you think of just expecting a _single file_ for now under the directory that conforms to the YAML patch format (otherwise returning an error)?
I think in the ideal world, we will have `copilot [noun] override --tool yamlpatch` generate two files under the `overrides` directory: `README.md` and `cfn.patches.yaml`.

We might need to also update to no longer expect a single file but if there is no `cdk.json` then assume it's YAML patch.
https://github.com/aws/copilot-cli/blob/caa6aca43b9ab4f0df4b80430fd36eebc532d78a/internal/pkg/override/override.go#L87-L91

efekarakus (commented):
Looks really good, it was really easy to follow 🤩 

Mostly minor recommendations 🚀

internal/pkg/cli/deploy/rdws.go
dannyrandall: maybe 🙄

internal/pkg/override/patch.go
dannyrandall: yeah - the last key in the map should be at index `len(node.Content)-2`, so the index `i+2` would be `len(node.Content)` - which, when slicing a slice, produces a slice of length 0. (also see [this test](https://github.com/aws/copilot-cli/pull/4641/files#diff-db45db57a8c5fa7983bfd1bd51464cb707a656831be2aa65c72bf3f758cf296cR172-R187))

```
func main() {
  list := []string{"a", "b", "c"}
  fmt.Printf("list: %+v, len(): %d\n", list, len(list))
  fmt.Printf("list[3:] %+v\n", list[3:])
}

// Output:
// list: [a b c], len(): 3
// list[3:] []
```

internal/pkg/override/patch.go
dannyrandall: sure, a single file sounds good to me - people can use comments in that file to help organize their overrides. i'll make that change!

internal/pkg/override/patch.go
dannyrandall: i think the relationship is the same here as `io/fs` - `Walk` continues calling the `WalkDirFunction` on every file until `SkipDir`, `SkipAll`, or a non-nil error is returned... We keep calling `visit()` until `errStopFollowingPointer` or a non-nil error is returned. do you think the naming is potentially confusing...?

dannyrandall:
I've been looking into reusing/merging some code from `taskdef_overrides` - there is one major difference in how patches work vs. taskdef overrides that complicates that: taskdef overrides recursively creates nodes from the input _path_, whereas yaml patches require a pointer that points to an existing (or new leaf, in the case of "add") node in the tree.

For example, given the input:
```
map:
  key: value
```
This works in taskdef overrides:
```
path: map.newKey.subKey
value: subKey_value
```
This not supported in a YAML Patch per [RFC6902](https://www.rfc-editor.org/rfc/rfc6902). Specifically:
> However, the object itself or an array containing it does need to exist, and it remains an error for that not to be the case.  For example, an "add" with a target location of "/a/b" starting with this document:
>
>   { "a": { "foo": 1 } }
>
>  is not an error, because "a" exists, and "b" will be added to its value.  It is an error in this document:
>
>   { "q": { "bar": 2 } }
>
>   because "a" does not exist.

To accomplish the same thing from my example in yaml patches, the operation would look like:
```
op: add
path: map/newKey
value:
  subKey: subKey_value
```
---
At this point, I'm hesitant to merge the two logical paths (one creating nodes as it follows a path, and one that fails) in favor of ensuring compatibility for existing taskdef override users. I think in a future release we can instead help customers migrate from taskdef overrides to yaml patches with some logic to convert an existing override to a patch - and as part of that logic, merge the two patterns by applying the conversion before applying the patches.

internal/pkg/override/patch.go
dannyrandall: you can [embed two yaml documents in a single file](https://yaml.org/spec/1.2.2/#22-structures). since multi document files don't exist in JSON (and therefore JSON Pointer syntax), there is no way to target a specific document with a JSON Pointer. So we just assume the target is the first document, and error if there are multiple documents in the file.

dannyrandall: update: https://pkg.go.dev/gopkg.in/yaml.v3#Unmarshal 🤦 
> Unmarshal decodes the first document

Since we only call unmarshal once, I think it's ok if we just ignore this case.

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,268 @@
...
+		var err error
+		switch patch.Operation {
+		case "add":
+			err = patch.applyAdd(&root)
+		case "remove":
+			err = patch.applyRemove(&root)
+		case "replace":
+			err = patch.applyReplace(&root)
+		default:
+			return nil, fmt.Errorf("unsupported operation %q. supported operations are %q, %q, and %q.", patch.Operation, "add", "remove", "replace")
```
efekarakus: nit:
```suggestion
			return nil, fmt.Errorf("unsupported operation %q: supported operations are %q, %q, and %q.", patch.Operation, "add", "remove", "replace")
```

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,268 @@
...
+			if node.Content[i].Value == remaining[0] {
+				return followJSONPointerHelper(node.Content[i+1], append(traversed, remaining[0]), remaining[1:], visit)
+			}
+		}
+
+		return fmt.Errorf("key %q: %q not found in map", strings.Join(traversed, jsonPointerSeparator), remaining[0])
+	case yaml.SequenceNode:
+		if remaining[0] == "-" {
+			return followJSONPointerHelper(node.Content[len(node.Content)-1], append(traversed, remaining[0]), remaining[1:], visit)
+		}
```
efekarakus: ooh would you mind adding a comment here saying that removing elements using "-" is not allowed according to the spec:
> Note that the use of the "-" character to index an array will always
   result in such an error condition because by definition it refers to
   a nonexistent array element.  

I initially thought that doing the following was a valid option:
```go
- op: remove
  path: /Resources/TaskDef/List/-
```

internal/pkg/override/patch_test.go
```diff
@@ -0,0 +1,452 @@
...
+			yaml: `
+a:
+  b:
+    - c
+    - d`,
+			overrides: `
+- op: add
+  path: /a/b/1/e
+  value: val`,
+			expectedErr: `unable to apply "add" patch: key "/a/b/1": invalid node type 0x8`,
```
efekarakus: 😅  the node type `0x8` won't make sense to users, can we add a `nodeKindStringer` type to make the error more friendly?
```go
type nodeKindStringer yaml.Kind

// String implements fmt.Stringer.
func (k nodeKindStringer) String() string {
   switch k {
   case yaml.DocumentNode:
      return "document"
   case yaml.MappingNode:
      return "mapping"
   // ...
   }
}
```

and in `followJSONPointerHelper`:
```go
return fmt.Errorf("key %q: invalid node type %q", strings.Join(traversed, jsonPointerSeparator), nodeKindStringer(node.Kind))
```

efekarakus (approved):
Some error message suggestions, feel free to remove the label afterwards

internal/pkg/override/patch.go
dannyrandall: hopefully the updated logic is sufficient; `-` logic only exists in `add` now

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,283 @@
...
+		fs = opts.FS
+	}
+
+	return &Patch{
+		filePath: filePath,
+		fs:       fs,
+	}
+}
+
+type yamlPatch struct {
```
Lou1415926: nit: can we move this type declaration to just before `func (p *yamlPatch) applyRemove(root *yaml.Node) error` ?

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,283 @@
...
+			err = patch.applyAdd(&root)
+		case "remove":
+			err = patch.applyRemove(&root)
+		case "replace":
+			err = patch.applyReplace(&root)
+		default:
+			return nil, fmt.Errorf("unsupported operation %q: supported operations are %q, %q, and %q.", patch.Operation, "add", "remove", "replace")
+		}
+		if err != nil {
+			return nil, fmt.Errorf("unable to apply %q patch: %w", patch.Operation, err)
```
Lou1415926: what do you think of including the index information? So that people see for example `unable to apply the patch at index 0 with "add": value required`

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,283 @@
...
+
+func followJSONPointerHelper(node *yaml.Node, traversed, remaining []string, visit visitNodeFunc) error {
+	if err := visit(node, remaining); err != nil {
+		if errors.Is(err, errStopFollowingPointer) {
+			return nil
+		}
+		return fmt.Errorf("key %q: %w", strings.Join(traversed, jsonPointerSeparator), err)
+	} else if len(remaining) == 0 {
+		return nil
+	}
```
Lou1415926: nit: this seems to be the same?
```suggestion
	}
	if len(remaining) == 0 {
		return nil
	}
```

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,283 @@
...
+
+	switch node.Kind {
+	case yaml.DocumentNode:
+		if len(node.Content) == 0 {
+			return nil // weird, but ok ¯\_(ツ)_/¯
```
Lou1415926: When would this happen? Maybe only when the first document is empty? What about something like ⬇️  😂 

```
// NOTE: happens only when xxx; realistically wouldn't happen (weird, but ok ¯\_(ツ)_/¯)
```

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,283 @@
...
+	if p.Value.IsZero() {
+		return fmt.Errorf("value required")
+	}
+
+	return followJSONPointer(root, p.Path, func(node *yaml.Node, pointer []string) error {
+		if len(pointer) != 1 {
+			return nil
+		}
+
+		switch node.Kind {
```
Lou1415926: How does the `ScalarNode` case play with the this? I guess it'd error out in `followHelper` because if the JSON pointer goes to a scalar node, it doesn't make sense for "add"?

internal/pkg/override/patch_test.go
```diff
@@ -0,0 +1,483 @@
...
+import (
+	"strings"
+	"testing"
+
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+)
+
+func TestPatch(t *testing.T) {
```
Lou1415926, line 15: noice  test cases👍🏼

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+func (p *yamlPatch) applyRemove(root *yaml.Node) error {
+	return followJSONPointer(root, p.Path, func(node *yaml.Node, pointer []string) error {
+		if len(pointer) != 1 {
+			return nil
+		}
+
+		switch node.Kind {
+		case yaml.DocumentNode:
```
Lou1415926: How would we hit this case?

internal/pkg/override/patch.go
dannyrandall: good idea!!

internal/pkg/override/patch.go
dannyrandall: code is gone in refactor!

internal/pkg/override/patch.go
dannyrandall: it's a fairly...useless scenario for our use case, but with the following patch:
```
op: remove
path: ""
```
(see this test https://github.com/aws/copilot-cli/pull/4641/files#diff-db45db57a8c5fa7983bfd1bd51464cb707a656831be2aa65c72bf3f758cf296cR344-R350)

internal/pkg/override/patch.go
dannyrandall: yeah previously this was caught by the followHelper, but in the refactor i now have a case for scalar nodes here!

internal/pkg/override/patch.go
dannyrandall: i've played with it for a while and can't actually find a yaml document that would make this occur - i'll add a note that it's just there to avoid panics. In theory it would be something like 
```
---
key: value
```
but the parser still creates a document node in this case, with a `!!null` node inside.

dannyrandall: i'll switch it to spit out an error as well

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,318 @@
...
+			Kind:  yaml.ScalarNode,
+			Tag:   "!!str",
+			Value: pointer.FinalKey(),
+		})
+		parent.Content = append(parent.Content, &p.Value)
+	case yaml.SequenceNode:
+		if pointer.FinalKey() == "-" {
+			// add to end of sequence
+			parent.Content = append(parent.Content, &p.Value)
+			return nil
```
iamhopaul123: Is this possible to make this whole thing `parent.add(p.Value)`? sorry i can't include more lines but it should reach to the end.

and `applyAdd(root *yaml.Node)` will be clearer: we find the target node to add/replace/delete, and then apply the action.

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+func (p pointer) FinalKey() string {
+	if len(p) == 0 {
+		return ""
+	}
+	return p[len(p)-1]
+}
+
+func (y yamlPatch) Pointer() pointer {
```
iamhopaul123: Why do we make them public?

internal/pkg/override/patch.go
```diff
@@ -0,0 +1,318 @@
...
+	for i := 0; i < len(node.Content); i += 2 {
+		if node.Content[i].Value == key {
+			return i, nil
+		}
+	}
+
+	return 0, fmt.Errorf("key %q: %q not found in map", strings.Join(traversed, jsonPointerSeparator), key)
+}
+
+func getNode(node *yaml.Node, remaining, traversed pointer) (*yaml.Node, error) {
```
iamhopaul123: maybe consider making the name more meaningful since it is a crucial one. Like `findTargetNode` or sth similar.

iamhopaul123 (commented):
LGTM overall! just few nits

internal/pkg/override/patch.go
dannyrandall: problem is that replace doesn't use the parent node, it uses the "node" node - so depending on the action, we need a different node

dannyrandall: though i guess the action for replace could be - take the parent, get the correct child node using `getNode()` and replace that?

iamhopaul123: no i just meant we can wrap this up to make the logic clearer. It can be different for replace doesn't necessarily need to be `parent.`

dannyrandall: i think all of the logic for each operation (add/remove/replace) is totally separate, so i'm not sure there's much benefit in breaking out the find parent and "do operation" logic, since it can't get reused in any way...

iamhopaul123 (approved):

PR #4648, opened 2023-03-20
chore: get the template body of a deployed stack and an addon stack

Lou1415926:


internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -213,6 +214,33 @@ func New(sess *session.Session, opts ...OptFn) CloudFormation {
 	return client
 }
 
+// Template returns a deployed stack's template.
+func (cf CloudFormation) Template(stackName string) (string, error) {
+	return cf.cfnClient.TemplateBody(stackName)
+}
+
+// AddonsTemplate returns the environment stack's addons template, if there is any.
+func (cf CloudFormation) AddonsTemplate(stackName string) (string, error) {
```
efekarakus: Out of curiosity why do we want the addons template body?

I feel like the diff for addons isn't particularly interesting since it's not abstracted by us, customers can already run `git diff` for example to view the changes. Do we need to display the diff for addons 🤔 ?

efekarakus (commented):
the code looks good, just a question

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -213,6 +214,33 @@ func New(sess *session.Session, opts ...OptFn) CloudFormation {
 	return client
 }
 
+// Template returns a deployed stack's template.
+func (cf CloudFormation) Template(stackName string) (string, error) {
+	return cf.cfnClient.TemplateBody(stackName)
+}
+
+// AddonsTemplate returns the environment stack's addons template, if there is any.
```
dannyrandall: is this specific to env stack?

dannyrandall (approved):

internal/pkg/deploy/cloudformation/cloudformation.go
dannyrandall: i think it makes sense to show - their last commit/edit may or may not have been deployed, so our diff gives them a diff against what is really deployed.

Lou1415926: The design in #4259 included the addons stack diff as a part of the feature. I guess one benefit is for users to have one place to view the diff, instead of having to run `git diff` as well as `copilot --diff` 🤔 ?

I think it makes sense though if we don't include the addons diff, at least for MVP - since we are not sure yet if it's beneficial enough. Pinging @dannyrandall to see if you have anything to add! 👀

efekarakus (approved):
Gotcha, I'll approve with `do-not-merge` and up to y'all if you want the addons diff. I feel like just showing the URL of addons changing is good enough given that they have edited the addon themselves (there is no abstraction) but I'm totally OK if you'd like to keep it too. If you think it's a better UX!

PR #4650, opened 2023-03-21
feat(stack): add support for additional_rules for ALB

paragbhingre:
This PR adds ability to support `additional_rules` for ALB end to end.

### DIFF
--- a/internal/pkg/cli/deploy/backend.go
+++ b/internal/pkg/cli/deploy/backend.go
@@ -126,18 +126,18 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
 	if d.backendMft.RoutingRule.IsEmpty() {
 		return nil
 	}
-	if err := d.validateALBRuntimeFor(d.backendMft.RoutingRule.Main); err != nil {
+	if err := d.validateRuntimeRoutingRule(d.backendMft.RoutingRule.Main); err != nil {
 		return fmt.Errorf(`validate ALB runtime configuration for "http": %w`, err)
 	}
 	for idx, rule := range d.backendMft.RoutingRule.AdditionalRoutingRules {
-		if err := d.validateALBRuntimeFor(rule); err != nil {
+		if err := d.validateRuntimeRoutingRule(rule); err != nil {
 			return fmt.Errorf(`validate ALB runtime configuration for "http.additional_rules[%d]": %w`, idx, err)
 		}
 	}
 	return nil
 }
 
-func (d *backendSvcDeployer) validateALBRuntimeFor(rule manifest.RoutingRule) error {
+func (d *backendSvcDeployer) validateRuntimeRoutingRule(rule manifest.RoutingRule) error {
 	if rule.IsEmpty() {
 		return nil
 	}
--- a/internal/pkg/cli/deploy/backend.go
+++ b/internal/pkg/cli/deploy/backend.go
@@ -127,11 +127,11 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
 		return nil
 	}
 	if err := d.validateALBRuntimeFor(d.backendMft.RoutingRule.Main); err != nil {
-		return fmt.Errorf("validate ALB runtime configuration for http: %w", err)
+		return fmt.Errorf(`validate ALB runtime configuration for "http": %w`, err)
 	}
 	for idx, rule := range d.backendMft.RoutingRule.AdditionalRoutingRules {
 		if err := d.validateALBRuntimeFor(rule); err != nil {
-			return fmt.Errorf("validate ALB runtime configuration for http.additional_rules[%d]: %w", idx, err)
+			return fmt.Errorf(`validate ALB runtime configuration for "http.additional_rules[%d]": %w`, idx, err)
 		}
 	}
 	return nil
--- a/internal/pkg/cli/deploy/backend.go
+++ b/internal/pkg/cli/deploy/backend.go
@@ -126,20 +126,35 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
 	if d.backendMft.RoutingRule.IsEmpty() {
 		return nil
 	}
+	if err := d.validateALBRuntimeFor(d.backendMft.RoutingRule.Main); err != nil {
+		return fmt.Errorf("validate ALB runtime configuration for http: %w", err)
+	}
+	for idx, rule := range d.backendMft.RoutingRule.AdditionalRoutingRules {
+		if err := d.validateALBRuntimeFor(rule); err != nil {
+			return fmt.Errorf("validate ALB runtime configuration for http.additional_rules[%d]: %w", idx, err)
+		}
+	}
+	return nil
+}
+
+func (d *backendSvcDeployer) validateALBRuntimeFor(rule manifest.RoutingRule) error {
+	if rule.IsEmpty() {
+		return nil
+	}
 	hasImportedCerts := len(d.envConfig.HTTPConfig.Private.Certificates) != 0
 	switch {
-	case d.backendMft.RoutingRule.Main.Alias.IsEmpty() && hasImportedCerts:
+	case rule.Alias.IsEmpty() && hasImportedCerts:
 		return &errSvcWithNoALBAliasDeployingToEnvWithImportedCerts{
 			name:    d.name,
 			envName: d.env.Name,
 		}
-	case d.backendMft.RoutingRule.Main.Alias.IsEmpty():
+	case rule.Alias.IsEmpty():
 		return nil
 	case !hasImportedCerts:
 		return fmt.Errorf(`cannot specify "alias" in an environment without imported certs`)
 	}
 
-	aliases, err := d.backendMft.RoutingRule.Main.Alias.ToStringSlice()
+	aliases, err := rule.Alias.ToStringSlice()
 	if err != nil {
 		return fmt.Errorf("convert aliases to string slice: %w", err)
 	}
--- a/internal/pkg/cli/deploy/lbws.go
+++ b/internal/pkg/cli/deploy/lbws.go
@@ -254,7 +254,7 @@ func (d *lbWebSvcDeployer) validateALBRuntimeFor(rule manifest.RoutingRule) erro
 			return err
 		}
 		if err := validateLBWSAlias(rule.Alias, d.app, d.env.Name); err != nil {
-			return fmt.Errorf(`validate 'http.alias': %w`, err)
+			return fmt.Errorf(`validate 'alias': %w`, err)
 		}
 		return nil
 	}
--- a/internal/pkg/cli/deploy/lbws.go
+++ b/internal/pkg/cli/deploy/lbws.go
@@ -189,19 +189,19 @@ func (d *lbWebSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (*s
 
 func (d *lbWebSvcDeployer) validateALBRuntime() error {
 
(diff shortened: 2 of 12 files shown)


internal/pkg/cli/deploy/backend.go
```diff
@@ -126,20 +126,35 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
 	if d.backendMft.RoutingRule.IsEmpty() {
 		return nil
 	}
+	if err := d.validateALBRuntimeFor(d.backendMft.RoutingRule.Main); err != nil {
+		return fmt.Errorf("validate ALB runtime configuration for http: %w", err)
```
iamhopaul123:
```suggestion
		return fmt.Errorf(`validate ALB runtime configuration for "http": %w`, err)
```

internal/pkg/cli/deploy/backend.go
```diff
@@ -126,20 +126,35 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
...
+	}
+	for idx, rule := range d.backendMft.RoutingRule.AdditionalRoutingRules {
+		if err := d.validateALBRuntimeFor(rule); err != nil {
+			return fmt.Errorf("validate ALB runtime configuration for http.additional_rules[%d]: %w", idx, err)
```
iamhopaul123:
```suggestion
			return fmt.Errorf(`validate ALB runtime configuration for "http.additional_rules[%d]": %w`, idx, err)
```
also can we have a test case covering this error msg?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -526,58 +524,64 @@ func (s *BackendService) convertALBListener() (*template.ALBListener, error) {
...
+	manifest        loadBalancerTargeter
+	httpsEnabled    bool
+	redirectToHTTPS bool
 }
 
 func (conv routingRuleConfigConverter) convert() (*template.ALBListenerRule, error) {
 	var aliases []string
 	var err error
+
+	exposedPorts, err := conv.manifest.ExposedPorts()
```
iamhopaul123, line 569: why do we call `ExposedPorts` outside rather than within `Target()`?

paragbhingre: Because we are calling `Target()` on listener rule level and not manifest level as target is listener rule level entity. Does that make sense?

iamhopaul123: In that case would it make more sense to revert rule and exposedPorts? Like `exposedPorts.Target(conv.rule)`? As `exposedPorts` is actually the one with more info.

paragbhingre: As per the offline discussion keeping it as is for readability and consistency purpose with NLB.

internal/pkg/cli/deploy/backend.go
paragbhingre: Yeah we have it [here](https://github.com/aws/copilot-cli/pull/4650/files#diff-abcea5573166990296031a09372d242d7a1a7ff625e48ff67526725f37525614R219)

internal/pkg/cli/deploy/backend.go
```diff
@@ -126,20 +126,35 @@ func (d *backendSvcDeployer) validateALBRuntime() error {
 	if d.backendMft.RoutingRule.IsEmpty() {
 		return nil
 	}
+	if err := d.validateALBRuntimeFor(d.backendMft.RoutingRule.Main); err != nil {
```
bvtujo: Can we rename this to `validateRuntimeALBRoutingRule` or something a little more descriptive? It's clever to continue the sentence in the signature but it is slightly jarring to see it in context here

internal/pkg/cli/deploy/lbws.go
```diff
@@ -188,13 +188,27 @@ func (d *lbWebSvcDeployer) stackConfiguration(in *StackRuntimeConfiguration) (*s
 }
 
 func (d *lbWebSvcDeployer) validateALBRuntime() error {
```
bvtujo, line 190: Is there any way we can reduce the code duplication in this function? It seems to appear in LBWS and Backend service and I wonder if there's an opportunity to extract this to a function that can be called from both BackendService.validateALBRuntime and LBWS.validateALBRuntime. Maybe there are too many variables though.

bvtujo (approved):
LGTM, just a few nits. And can we add this feature to an e2e test?

internal/pkg/manifest/svc.go
```diff
@@ -550,76 +550,49 @@ func sortExposedPorts(exposedPorts []ExposedPort) []ExposedPort {
...
 
 	if rrTargetPort == nil { // when target_port is nil
-		if rrTargetContainer != s.Name {
+		if aws.StringValue(rrTargetContainer) != exposedPorts.WorkloadName {
 			targetContainer = aws.StringValue(rrTargetContainer)
-			targetPort = aws.StringValue(s.Sidecars[aws.StringValue(rrTargetContainer)].Port)
+			for _, portConfig := range exposedPorts.PortsForContainer[targetContainer] {
+				if portConfig.isDefinedByContainer {
+					targetPort = strconv.Itoa(int(portConfig.Port))
+					/* NOTE: When the `target_port` is empty, the intended target port should be the port that is explicitly exposed by the container. Consider the following example
```
bvtujo, line 576: Sorry, sorry, more comments. Can we make sure to document this super clearly on the site?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -490,31 +490,29 @@ func (s *LoadBalancedWebService) convertALBListener() (*template.ALBListener, er
...
+		httpRedirect := true
+		if routingRule.RedirectToHTTPS != nil {
+			httpRedirect = aws.BoolValue(routingRule.RedirectToHTTPS)
+		}
+		rule, err := routingRuleConfigConverter{
+			rule:            routingRule,
+			manifest:        s.manifest,
+			httpsEnabled:    s.httpsEnabled,
+			redirectToHTTPS: httpRedirect,
+		}.convert()
```
bvtujo, line 504: clever, i like this pattern. haven't seen it in go much.

internal/pkg/cli/deploy/backend.go
paragbhingre: Sure I will rename this to `validateRuntimeRoutingRule`. In `ALBRoutingRule` the word `ALB` is repetitive because routing rule belongs to ALB itself.

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: @efekarakus's mind ❤️

internal/pkg/cli/deploy/lbws.go
paragbhingre: Yeah, the problem with making these two functions into one is the number of parameters we need to send and the different runtime requirements that these two services hold.

iamhopaul123 (approved):

internal/pkg/deploy/cloudformation/stack/backend_svc.go
```diff
@@ -221,7 +221,14 @@ func (s *BackendService) Parameters() ([]*cloudformation.Parameter, error) {
...
 	}
-	targetContainer, targetPort, err := s.manifest.HTTPLoadBalancerTarget()
+	exposedPorts, err := s.manifest.ExposedPorts()
+	if err != nil {
+		return nil, fmt.Errorf("parse exposed ports in service manifest %s: %w", s.name, err)
+	}
+	targetContainer, targetPort, err := s.manifest.RoutingRule.Main.Target(exposedPorts)
+	if targetPort == "" {
+		targetPort = s.manifest.MainContainerPort()
+	}
```
efekarakus, line 231: nit: can we move the setting of the variable after handling the error so that we don't do extra computation if the `Target` call returned an error?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -526,58 +524,64 @@ func (s *BackendService) convertALBListener() (*template.ALBListener, error) {
...
-	hostedZoneAliases, err := convertHostedZone(rrConfig.Main.Alias, rrConfig.Main.HostedZone)
-	if err != nil {
-		return nil, err
+	var hostedZoneAliases template.AliasesForHostedZone
+	for _, routingRule := range rrConfig.RoutingRules() {
+		rule, err := routingRuleConfigConverter{
+			rule:            routingRule,
+			manifest:        s.manifest,
+			httpsEnabled:    s.httpsEnabled,
+			redirectToHTTPS: s.httpsEnabled,
```
efekarakus, line 533: Oh should this be something similar to LBWS here?
```go
httpRedirect := s.httpsEnabled
if routingRule.RedirectToHTTPS != nil {
	httpRedirect = aws.BoolValue(routingRule.RedirectToHTTPS)
}
```

internal/pkg/manifest/svc.go
```diff
@@ -550,76 +550,49 @@ func sortExposedPorts(exposedPorts []ExposedPort) []ExposedPort {
...
-
-	if rrTargetContainer == nil { // when target_container is nil
-		container, port := targetContainerFromTargetPort(exposedPorts, rrTargetPort)
-		targetPort = aws.StringValue(port)
-		if container != nil {
-			targetContainer = aws.StringValue(container)
+	targetContainer = exposedPorts.WorkloadName
+	for _, portConfig := range exposedPorts.PortsForContainer[targetContainer] {
+		if portConfig.isDefinedByContainer {
+			targetPort = strconv.Itoa(int(portConfig.Port))
```
efekarakus, line 560: nit: looks like this could be a neat method on the `ExposedPortsIndex` so that we can remove some duplication, for example:

```go
targetPort = exposedPorts.findDefaultContainerPort()
// later down.
targetPort = exposedPorts.definedContainerPort(aws.StringValue(rrTargetContainer))
```

and:

```go
func (idx ExposedPortsIndex) findDefaultContainerPort() string {
   return idx.definedContainerPort(idx.WorkloadName)
}

// definedContainerPort returns the explicitly defined container port, if there is no port exposed for the container then returns the empty string "".
func (idx ExposedPortsIndex) definedContainerPort(container string) string {
  // This implementation..
}
```

efekarakus (commented):
Looks great! Just leaving some feedback (sorry it's post-merge 😢) They are not blocking except a potential issue with `redirectToHttps`

internal/pkg/deploy/cloudformation/stack/backend_svc.go
paragbhingre: Took a note of this and next nit comment as well. I will fix them as separate PR when I will address other pending items as well from other PRs. Is that fine with you?

efekarakus: yup yup 🚀

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: Okay so looks like we made this change as a part of [http.redirect_to_https](https://github.com/aws/copilot-cli/pull/4003) feature where we did have a [requirement](https://github.com/aws/copilot-cli/pull/4003#discussion_r975670081) for LBWS to disable http redirect to https  even when domains are present but at the same time CDN is enabled. The idea being we wanted CDN to terminate the TLS and everything behind it uses HTTP. There wasn't any supported use cases for backend service and adding this filed to boolean was just so we could reuse the template for both lbws/backend.

PR #4653, opened 2023-03-21
refactor: Login to docker client only once for all images

KollaAdithya:
This PR provides logic to perform docker login if ECR credentials not enabled. After login is succeeded we build and push all the container images.

### DIFF
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -366,11 +366,12 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 		return nil
 	}
 	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
-	if err := loginToDockerClient(d.repository); err != nil {
-		return err
+	dockerCmdClient := dockerengine.New(exec.NewCmd())
+	if err := d.repository.Login(dockerCmdClient); err != nil {
+		return fmt.Errorf("login to docker: %w", err)
 	}
 	for name, buildArgs := range buildArgsPerContainer {
-		digest, err := d.repository.BuildAndPush(dockerengine.New(exec.NewCmd()), buildArgs)
+		digest, err := d.repository.BuildAndPush(dockerCmdClient, buildArgs)
 		if err != nil {
 			return fmt.Errorf("build and push image: %w", err)
 		}
@@ -379,15 +380,6 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 			CustomTag:         d.image.CustomTag,
 			GitShortCommitTag: d.image.GitShortCommitTag,
 		}
-	}
-	return nil
-}
-
-// loginToDockerClient logs in to the Docker client using the provided RepositoryService.
-// Returns an error if any error occurs during the login process.
-func loginToDockerClient(rs repositoryService) error {
-	if err := rs.Login(dockerengine.New(exec.NewCmd())); err != nil {
-		return fmt.Errorf("login to docker: %w", err)
 	}
 	return nil
 }
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -68,8 +68,7 @@ func (noopActionRecommender) RecommendedActions() []string {
 	return nil
 }
 
-type RepositoryService interface {
-	URI() (string, error)
+type repositoryService interface {
 	Login(docker repository.ContainerLoginBuildPusher) error
 	BuildAndPush(docker repository.ContainerLoginBuildPusher, args *dockerengine.BuildArguments) (string, error)
 }
@@ -157,7 +156,7 @@ type workloadDeployer struct {
 	fs               fileReader
 	s3Client         uploader
 	addons           stackBuilder
-	repository       RepositoryService
+	repository       repositoryService
 	deployer         serviceDeployer
 	tmplGetter       deployedTemplateGetter
 	endpointGetter   endpointGetter
@@ -367,7 +366,7 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 		return nil
 	}
 	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
-	if err := LoginToDockerClient(d.repository); err != nil {
+	if err := loginToDockerClient(d.repository); err != nil {
 		return err
 	}
 	for name, buildArgs := range buildArgsPerContainer {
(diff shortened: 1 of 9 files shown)


internal/pkg/cli/deploy/workload.go
```diff
@@ -67,7 +67,9 @@ func (noopActionRecommender) RecommendedActions() []string {
 	return nil
 }
 
-type imageBuilderPusher interface {
+type RepositoryService interface {
+	URI() (string, error)
```
iamhopaul123: it doesn't make sense to me when a function seems to only be used within the other pkg, but the interface is declared in this pkg.

internal/pkg/cli/deploy/workload.go
```diff
@@ -356,6 +361,15 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	return nil
 }
 
+// LoginToDockerClient logs in to the Docker client using the provided RepositoryService.
+// Returns an error if any error occurs during the login process.
+func LoginToDockerClient(rs RepositoryService) error {
+	if err := rs.Login(dockerengine.New(exec.NewCmd())); err != nil {
```
iamhopaul123: I feel like the login should be handled by this pkg only (this function can be private) and we don't need to update `interface.go` in `cli` pkg. Plus, we can remove `URI()` from `RepositoryService` and make `RepositoryService` to be a private interface.

internal/pkg/cli/deploy/workload.go
KollaAdithya: Done!

internal/pkg/cli/deploy/workload.go
```diff
@@ -356,6 +360,15 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	return nil
 }
 
+// loginToDockerClient logs in to the Docker client using the provided RepositoryService.
+// Returns an error if any error occurs during the login process.
+func loginToDockerClient(rs repositoryService) error {
```
dannyrandall: i feel like this function doesn't need to exist - it's not really hiding any logic from L346 - we should just change up there to:
```
if err := d.repository.Login(dockerengine.New(exec.NewCmd())); err != nil {
  return fmt.Errorf("login to docker: %w", err)
}
```

iamhopaul123: This function doesn't bring too much benefit 🤔 and we can probably reuse `dockerengine.New(exec.NewCmd()` too.

internal/pkg/cli/task_run.go
```diff
@@ -959,6 +960,9 @@ func (o *runTaskOpts) buildAndPushImage() error {
 	if o.dockerfileContextPath != "" {
 		ctx = o.dockerfileContextPath
 	}
+	if err := o.repository.Login(dockerengine.New(exec.NewCmd())); err != nil {
```
iamhopaul123: can we reuse `dockerengine.New(exec.NewCmd()`?

internal/pkg/repository/repository.go
```diff
@@ -91,3 +79,25 @@ func (r *Repository) URI() (string, error) {
 	}
 	return uri, nil
 }
+
+// Login authenticates with a ECR registry by performing a Docker login,
+// but only if the `credStore` attribute value is not set to `ecr-login`.
+// If the `credStore` value is `ecr-login`, no login is performed.
+// Returns an error if any error occurs during the login process.
+func (r *Repository) Login(docker ContainerLoginBuildPusher) error {
+	uri, err := r.URI()
```
dannyrandall, line 88: do you have any idea what the impact of this change will be if someone has set `image.build.uri`? i'm not actually sure what a valid configuration looks like using that field, but wonder if it still works if someone set URI to an ECR repo in a different account 🤔 (or if that even works today!). Just thinking that [previously we logged in to `args.URI`](https://github.com/aws/copilot-cli/pull/4653/files#diff-c3341fa47b7006ee5fa20ac85575594903aa23154c1427d2daa666e074d87e75L53-L59), but we won't anymore if that's set.

bvtujo: Cross account ECR pull is technically possible but I'm not sure about the required permissions. Copilot might enable it depending on which account the regional resources stackset instance gets created in. 

Possible validations might be 
a) to just warn if the customer is using a custom URI that push might not work
b) to check whether we actually need to call login; currently it doesn't look like push destination is configurable unless there's some fancy logic I'm missing where you can specify a destination in `image.build.args`

bvtujo (commented):
+1 to other feedback others have left; will approve when addressed

internal/pkg/repository/repository.go
KollaAdithya: Discussed this with @dannyrandall offline
 `args.URI` is never set expect 
 https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/repository/repository.go#L54-L58
 which basically means `args.URI` will always be `uri` returned from `r.URI()`

internal/pkg/cli/deploy/workload.go
```diff
@@ -342,8 +343,12 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 		return nil
 	}
 	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
+	dockerCmdClient := dockerengine.New(exec.NewCmd())
+	if err := d.repository.Login(dockerCmdClient); err != nil {
```
dannyrandall: i wonder if `Login` should return the URI and set it in `buildArgs`? otherwise i think we'll make the call to ECR for every single container...

KollaAdithya: `Login` will be called only once right for all the images. `Login` function is not inside any `loop` then how the login will be called for every container 🤔

dannyrandall: but we call `repository.URI()` every time we call `BuildAndPush()`! and `repository.URI()` makes an API call to ECR.

KollaAdithya: Just made the changes so that we pass the `URI` to `dockerBuildArgs`.

internal/pkg/repository/repository.go
```diff
@@ -83,21 +83,21 @@ func (r *Repository) URI() (string, error) {
 // Login authenticates with a ECR registry by performing a Docker login,
 // but only if the `credStore` attribute value is not set to `ecr-login`.
 // If the `credStore` value is `ecr-login`, no login is performed.
-// Returns an error if any error occurs during the login process.
-func (r *Repository) Login(docker ContainerLoginBuildPusher) error {
+// Returns an uri and if any error occurs during the login process.
```
bvtujo:
```suggestion
// Returns a URI or an error, if any occurs during the login process.
```

bvtujo (approved):
LGTM, just one doc nit!

dannyrandall (approved):

PR #4656, opened 2023-03-21
chore: let svc/env deployer generate the diff

Lou1415926:


internal/pkg/cli/deploy/env.go
```diff
@@ -157,6 +159,7 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
 
 		appCFN:      deploycfn.New(defaultSession, deploycfn.WithProgressTracker(os.Stderr)),
 		envDeployer: cfnClient,
+		tmplGetter:  cfnClient,
```
paragbhingre, line 162: Can we not use `envDeployer: cfnClient` instead of creating new here?

paragbhingre (approved):
Looks good to me! Just a question.

internal/pkg/cli/deploy/env.go
Lou1415926: My intuition is that the template getter should not necessarily = env deployer, because their purposes are different. In this case, they just happen to be the same instance under the hood, partially because of how our codebase is currently structure. 

A counter argument would be that "it makes sense if getting a deployed template is also a responsibility of a deployer". If we prefer this line of thinking, I'm fine as well. By default, I try to separate responsibilities as fine-grind as possible, under the constraint of time & effort

paragbhingre: I think agree with the counter argument, also it is better to have separate responsibilities as fine-grind as possible. Let's keep as is.

internal/pkg/cli/deploy/env.go
```diff
@@ -220,6 +223,23 @@ func (d *envDeployer) UploadArtifacts() (*UploadEnvArtifactsOutput, error) {
...
 
+// DeployDiff returns the stringified diff of the template against the deployed template of the environment.
+func (d *envDeployer) DeployDiff(template string) (string, error) {
+	tmpl, err := d.tmplGetter.Template(cfnstack.NameForEnv(d.app.Name, d.env.Name))
+	if err != nil {
+		return "", fmt.Errorf("retrieve the deployed template for %q: %w", d.env.Name, err)
+	}
+	diffTree, err := diff.From(tmpl).Parse([]byte(template))
+	if err != nil {
+		return "", fmt.Errorf("parse the diff of the template to be deployed against the deployed template: %w", err)
```
```suggestion
		return "", fmt.Errorf("parse the diff against the deployed env stack: %w", err)
```
maybe this is enough?

iamhopaul123 (approved):

efekarakus (approved):

PR #4657, opened 2023-03-21
chore: add `--diff` flag in svc/job/env deploy

Lou1415926:
Looks like a large PR but the logic is super similar across env/svc/job deploy. Most of the additions are because of unit tests! (and the unit tests are similar across env/svc/job deploy as well)

### DIFF
--- a/internal/pkg/cli/env_deploy.go
+++ b/internal/pkg/cli/env_deploy.go
@@ -32,6 +32,7 @@ type deployEnvVars struct {
 	name            string
 	forceNewUpdate  bool
 	disableRollback bool
+	showDiff        bool
 }
 
 type deployEnvOpts struct {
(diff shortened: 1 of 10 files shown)


internal/pkg/cli/env_deploy.go
```diff
@@ -151,6 +157,30 @@ func (o *deployEnvOpts) Execute() error {
 	if err != nil {
 		return fmt.Errorf("upload artifacts for environment %s: %w", o.name, err)
 	}
+	if o.showDiff {
+		output, err := deployer.GenerateCloudFormationTemplate(&deploy.DeployEnvironmentInput{
```
iamhopaul123: What do you think if we move logics within the if statement to another method so that i'll be easier to follow the logic in `Execute`?

iamhopaul123 (approved):

internal/pkg/cli/flag.go
```diff
@@ -249,6 +250,7 @@ const (
 	yesFlagDescription          = "Skips confirmation prompt."
 	resourceTagsFlagDescription = `Optional. Labels with a key and value separated by commas.
 Allows you to categorize resources.`
+	diffFlagDescription = "Show the diff of the to-be-deployed CFN template against the deployed one."
```
efekarakus, line 253: I guess one of these PRs will have a merge conflict 😅 https://github.com/aws/copilot-cli/pull/4662

efekarakus (approved):
looks good, +1 to boss's feedback

internal/pkg/cli/flag.go
Lou1415926: ya i came prepared. The parts that likely will be in config (the `flag.go` file are on their own separate commits). I will just need to rebase and drop that commit in the other PR

PR #4662, opened 2023-03-22
chore: add `--diff` flag in svc/job/env package 

Lou1415926:
Large PR but the real logic is repetitive across the three commands. So consider #additions as #additions/3 additions.

### DIFF
--- a/internal/pkg/cli/env_package.go
+++ b/internal/pkg/cli/env_package.go
@@ -66,6 +66,7 @@ type packageEnvOpts struct {
 	tplWriter    io.WriteCloser
 	paramsWriter io.WriteCloser
 	addonsWriter io.WriteCloser
+	diffWriter   io.Writer
 
 	newInterpolator func(appName, envName string) interpolator
 	newEnvPackager  func() (envPackager, error)
@@ -100,6 +101,7 @@ func newPackageEnvOpts(vars packageEnvVars) (*packageEnvOpts, error) {
 		tplWriter:    os.Stdout,
 		paramsWriter: discardFile{},
 		addonsWriter: discardFile{},
+		diffWriter:   os.Stdout,
 
 		newInterpolator: func(appName, envName string) interpolator {
 			return manifest.NewInterpolator(appName, envName)
@@ -196,6 +198,9 @@ func (o *packageEnvOpts) Execute() error {
 	})
 	if err != nil {
 		return fmt.Errorf("generate CloudFormation template from environment %q manifest: %v", o.envName, err)
+	}
+	if o.showDiff {
+		return diff(packager, res.Template, o.diffWriter)
 	}
 	addonsTemplate, err := packager.AddonsTemplate()
 	if err != nil {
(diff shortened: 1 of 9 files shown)


internal/pkg/cli/env_package.go
```diff
@@ -131,6 +134,14 @@ func newPackageEnvOpts(vars packageEnvVars) (*packageEnvOpts, error) {
 
 // Validate returns an error for any invalid optional flags.
 func (o *packageEnvOpts) Validate() error {
+	if o.showDiff {
+		if o.outputDir != "" {
+			return fmt.Errorf("`--%s` cannot be specified together with `--%s`", diffFlag, stackOutputDirFlag)
+		}
```
efekarakus: can we use `cmd.MarkFlagsMutuallyExclusive` instead to get rid of the functionality? (samesies in other commands)

internal/pkg/cli/flag.go
```diff
@@ -249,6 +250,7 @@ const (
 	yesFlagDescription          = "Skips confirmation prompt."
 	resourceTagsFlagDescription = `Optional. Labels with a key and value separated by commas.
 Allows you to categorize resources.`
+	diffFlagDescription = "Show the diff of the to-be-deployed CFN template against the deployed one."
```
efekarakus: To go with a similar wording to `cdk diff`, perhaps:
> Compares the specified stack with the deployed stack or a local template file
```suggestion
	diffFlagDescription = "Compares the generated CloudFormation template to the deployed stack."
```

efekarakus (commented):
A nice thing that I've read in the CDK diff command is this:
> and returns with status 1 if any difference is found

So I assume it exits with 0 if there are no changes. We should consider also supporting hte same behavior. But the changes look great, just two small suggestions.

internal/pkg/cli/svc_package.go
```diff
@@ -470,5 +488,9 @@ func buildSvcPackageCmd() *cobra.Command {
 	cmd.Flags().StringVar(&vars.tag, imageTagFlag, "", imageTagFlagDescription)
 	cmd.Flags().StringVar(&vars.outputDir, stackOutputDirFlag, "", stackOutputDirFlagDescription)
 	cmd.Flags().BoolVar(&vars.uploadAssets, uploadAssetsFlag, false, uploadAssetsFlagDescription)
+	cmd.Flags().BoolVar(&vars.showDiff, diffFlag, false, diffFlagDescription)
+
+	cmd.MarkFlagsMutuallyExclusive(diffFlag, stackOutputDirFlag)
+	cmd.MarkFlagsMutuallyExclusive(diffFlag, uploadAssetsFlag)
```
iamhopaul123, line 482: Why do we not allow them to specify upload assets flag?

efekarakus (approved):

dannyrandall (approved):

internal/pkg/cli/svc_package.go
Lou1415926: `copilot svc package --diff --upload-assets`  - not sure if it makes sense to upload assets while they just want to show diff 🤔  

We upload env file, custom resource and addons with `--upload-assets`. Env files URLs are passed via. `Parameters` iirc, custom resources/addons URLs are deterministic and are rendered into the template regardless whether they are uploaded. Therefore, with/without `--upload-assets` should render the same template. We can definitely relax this constraint now or later, I am not feeling strong on this.

PR #4663, opened 2023-03-22
chore: ignore `Metadata.Manifest` in diff tree

Lou1415926:

### DIFF
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -207,7 +207,6 @@ func parseSequence(fromNode, toNode *yaml.Node) ([]diffNode, error) {
 				matchCount = 0
 			}
 		case actionMod:
-			// TODO(lou1415926): handle list of maps modification
 			diff := cachedDiff[cacheKey(inspector.fromIndex(), inspector.toIndex())]
 			if diff.err != nil {
 				return nil, diff.err
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -86,8 +86,17 @@ type seqItemNode struct {
 // From is the YAML document that another YAML document is compared against.
 type From []byte
 
+// ParseWithIgnorer constructs a diff tree that represent the differences of a YAML document against the From document, ignoring certain paths.
+func (from From) ParseWithIgnorer(to []byte, ignorer *Ignorer) (Tree, error) {
+	return from.parseRoot(to, ignorer)
+}
+
 // Parse constructs a diff tree that represent the differences of a YAML document against the From document.
-func (from From) Parse(to []byte, overrider Overrider) (Tree, error) {
+func (from From) Parse(to []byte) (Tree, error) {
+	return from.parseRoot(to, &noneOverrider{})
+}
+
+func (from From) parseRoot(to []byte, overrider overrider) (Tree, error) {
 	var toNode, fromNode yaml.Node
 	if err := yaml.Unmarshal(to, &toNode); err != nil {
 		return Tree{}, fmt.Errorf("unmarshal current template: %w", err)
@@ -117,12 +126,11 @@ func (from From) Parse(to []byte, overrider Overrider) (Tree, error) {
 	return Tree{
 		root: root,
 	}, nil
-
 }
 
-func parse(from, to *yaml.Node, key string, overrider Overrider) (diffNode, error) {
-	if overrider.Match(from, to, key) {
-		return overrider.Parse(from, to, key)
+func parse(from, to *yaml.Node, key string, overrider overrider) (diffNode, error) {
+	if overrider.match(from, to, key) {
+		return overrider.parse(from, to, key)
 	}
 	// Handle base cases.
 	if to == nil || from == nil || to.Kind != from.Kind {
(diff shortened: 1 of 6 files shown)


internal/pkg/template/diff/diff.go
```diff
@@ -87,7 +87,7 @@ type seqItemNode struct {
 type From []byte
 
 // Parse constructs a diff tree that represent the differences of a YAML document against the From document.
-func (from From) Parse(to []byte) (Tree, error) {
+func (from From) Parse(to []byte, overrider Overrider) (Tree, error) {
```
efekarakus: What do you think of having two different methods for a clearer API?

```go
func (from From) Parse(to []byte) (Tree, error)
func (from From) ParseWithOverrider(to []byte, ovrdr Overrider) (Tree, error)
```
both of these can call the same underlying `parse` function but the first one would use a `noopOverrider` vs. the other one would pass down the parameter

internal/pkg/template/diff/override.go
```diff
@@ -0,0 +1,62 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package diff
+
+import "gopkg.in/yaml.v3"
+
+type Overrider interface {
+	Match(from, to *yaml.Node, key string) bool
+	Parse(from, to *yaml.Node, key string) (diffNode, error)
```
efekarakus: Can we get away with these methods being private? A client wouldn't be able to implement a test double Overrider themselves anyway since `diffNode` isn't exported

internal/pkg/template/diff/override.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// CFNOverrider returns an Overrider that handles special behaviors when parsing the diff tree of two CFN documents written in YAML.
+func CFNOverrider() Overrider {
+	// return &noneOverrider{}
```
efekarakus:
```suggestion
```

internal/pkg/template/diff/diff.go
efekarakus: Actually, thinking some more maybe we can start by being super specific rather than the more generic Overrider term:

```go
func (from From) Parse(to []byte) (Tree, error)
func (from From) ParseWithIgnorer(to []byte, ignorer Ignorer) (Tree, error)
```
and clients can write

```go
from.ParseWithIgnorer(to, diff.IgnoreCFNMetadata())
```

internal/pkg/template/diff/override.go
```diff
@@ -0,0 +1,62 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package diff
+
+import "gopkg.in/yaml.v3"
+
+type Overrider interface {
```
efekarakus: doc comments 🥲

efekarakus (commented):
Looks good! I'm mostly just recommending some API name changes

internal/pkg/template/diff/diff.go
Lou1415926: I ended up just did this to be super duper specific

```
func (from From) Parse(to []byte) (Tree, error)
func (from From) ParseWithCFNIgnorer(to []byte) (Tree, error)
```

and not exposing anything from `override` - not the interface `overrider` nor any of the specific overrider.

internal/pkg/template/diff/override.go
```diff
@@ -0,0 +1,51 @@
...
+	m.curr = m.curr.next
+	return false
+}
+
+// Parse is a no-op for an ignorer.
+func (m *ignorer) parse(_, _ *yaml.Node, _ string) (diffNode, error) {
+	return nil, nil
+}
+
+type noneOverrider struct{}
```
efekarakus: nit: 
```suggestion
type noopOverrider struct{}
```

efekarakus (approved):

internal/pkg/template/diff/diff_test.go
```diff
@@ -353,8 +337,65 @@ Dog: (pleased) "ikr"`, t),
...
+			} else {
+				require.NoError(t, err)
+				require.True(t, equalTree(got, Tree{tc.wanted()}, t), "should get the expected tree")
 			}
-			if tc.wanted != nil {
+		})
+	}
+}
+
+func TestFrom_Parse_CFN(t *testing.T) {
```
paragbhingre: Just for consistency does it have to be?
```suggestion
func TestFrom_ParseWithCFNIgnorer(t *testing.T) {
```

internal/pkg/template/diff/override.go
```diff
@@ -0,0 +1,51 @@
...
+
+// Parse is a no-op for an ignorer.
+func (m *ignorer) parse(_, _ *yaml.Node, _ string) (diffNode, error) {
+	return nil, nil
+}
+
+type noopOverrider struct{}
+
+// Match always returns false for a noopOverrider.
+func (_ *noopOverrider) match(_, _ *yaml.Node, _ string) bool {
```
dannyrandall:
```suggestion
func (*noopOverrider) match(_, _ *yaml.Node, _ string) bool {
```
iirc you don't need `_`

internal/pkg/template/diff/override.go
```diff
@@ -0,0 +1,51 @@
...
+	return false
+}
+
+// Parse is a no-op for a noopOverrider.
+func (m *noopOverrider) parse(_, _ *yaml.Node, _ string) (diffNode, error) {
```
dannyrandall:
```suggestion
func (*noopOverrider) parse(_, _ *yaml.Node, _ string) (diffNode, error) {
```

dannyrandall (approved):

PR #4666, opened 2023-03-22
feat: upgrade pipeline CodeBuild project deps to std4 versions

efekarakus:
Update CodeBuild buildspec dependency versions and default project image, so that Copilot overrides with the CDK can be synthesized.

Related #4208 

### DIFF
--- a/internal/pkg/deploy/pipeline_test.go
+++ b/internal/pkg/deploy/pipeline_test.go
@@ -5,8 +5,9 @@ package deploy
 
 import (
 	"errors"
-	"gopkg.in/yaml.v3"
 	"testing"
+
+	"gopkg.in/yaml.v3"
 
 	"github.com/aws/copilot-cli/internal/pkg/config"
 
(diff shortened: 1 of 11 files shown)


internal/pkg/deploy/pipeline_test.go
```diff
@@ -201,7 +202,7 @@ func TestPipelineSourceFromManifest(t *testing.T) {
 
 func TestPipelineBuild_Init(t *testing.T) {
 	const (
-		defaultImage   = "aws/codebuild/amazonlinux2-x86_64-standard:3.0"
```
Lou1415926, line 204: Are we required to update this var if we update the real default image version? (I guess not?)

If not, what do you think of we  just use a value like "mock/image:1.0" so that it doesn't appear in search results 🤔

internal/pkg/template/templates/cicd/buildspec.yml
```diff
@@ -3,8 +3,9 @@ version: 0.2
 phases:
   install:
     runtime-versions:
-      docker: 19
-      ruby: 2.6
+      docker: 20
+      ruby: 3.1
```
Lou1415926, line 7: noice thank you!

Lou1415926 (approved):

internal/pkg/deploy/pipeline_test.go
efekarakus: :rip: sadly it looks like we need to use the real default image because this test checks that the `image` used by default is indeed "4.0"

internal/pkg/template/templates/cicd/buildspec.yml
```diff
@@ -3,8 +3,9 @@ version: 0.2
...
+      nodejs: 16
```
dannyrandall, line 8: [rip](https://nodejs.org/en/blog/announcements/nodejs16-eol)

dannyrandall (approved):

PR #4667, opened 2023-03-23
chore(describe): fix `svc show` to list only rollback alarms

huanjani:
Previously, we erroneously checked for rollback names to be not nil, then fetched the descriptions of the (non-existent) names. The `DescribeAlarms` call fetches all alarms in an account if no alarm names are passed in as a filter. So if no rollback alarms were present, all non-rollback alarm descriptions would be displayed!

Instead, the condition should check for a non-zero slice length because the empty slice is not nil.

### DIFF
--- a/internal/pkg/describe/backend_service.go
+++ b/internal/pkg/describe/backend_service.go
@@ -178,7 +178,7 @@ func (d *BackendServiceDescriber) Describe() (HumanJSONStringer, error) {
 		if err != nil {
 			return nil, fmt.Errorf("retrieve rollback alarm names: %w", err)
 		}
-		if alarmNames != nil {
+		if len(alarmNames) != 0 {
 			cwAlarmDescr, err := d.initCWDescriber(env)
 			if err != nil {
 				return nil, err
--- a/internal/pkg/describe/backend_service_test.go
+++ b/internal/pkg/describe/backend_service_test.go
@@ -199,7 +199,7 @@ func TestBackendServiceDescriber_Describe(t *testing.T) {
 						OperatingSystem: "LINUX",
 						Architecture:    "X86_64",
 					}, nil),
-					m.ecsDescriber.EXPECT().RollbackAlarmNames().Return(nil, nil),
+					m.ecsDescriber.EXPECT().RollbackAlarmNames().Return([]string{}, nil),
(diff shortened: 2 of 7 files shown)


internal/pkg/describe/backend_service_test.go
```diff
@@ -516,7 +516,7 @@ func TestBackendServiceDescriber_Describe(t *testing.T) {
 						OperatingSystem: "LINUX",
 						Architecture:    "X86_64",
 					}, nil),
-					m.ecsDescriber.EXPECT().RollbackAlarmNames().Return(nil, nil),
+					m.ecsDescriber.EXPECT().RollbackAlarmNames().Return([]string{}, nil),
```
dannyrandall: why did these need to change? i think it's pretty normal to return nil instead of an empty slice...

internal/pkg/describe/backend_service.go
```diff
@@ -178,7 +178,7 @@ func (d *BackendServiceDescriber) Describe() (HumanJSONStringer, error) {
 		if err != nil {
 			return nil, fmt.Errorf("retrieve rollback alarm names: %w", err)
 		}
-		if alarmNames != nil {
+		if len(alarmNames) != 0 {
```
iamhopaul123, line 181: We should probably fix `AlarmDescriptions` too. We don't list everything unless in the future we need this functionality.

internal/pkg/describe/backend_service_test.go
iamhopaul123: +1 maybe we should fix `AlarmDescriptions`

huanjani: > why did these need to change? i think it's pretty normal to return nil instead of an empty slice...

I had changed it bc that was the cause of tests passing previously when they shouldn't have-- `RollbackAlarmNames()` was returning an empty slice, so the unit test met the condition of `alarmNames` being nil or not, but in reality it was only ever an empty slice or a non-empty slice. 

I'll change `RollbackAlarmNames` to return nil, rather than changing the condition.

bvtujo (approved):
LGTM, great looking tests too.

iamhopaul123 (approved):

PR #4675, opened 2023-03-23
chore: handle exit code when there is a diff

Lou1415926:
https://github.com/aws/copilot-cli/pull/4662#pullrequestreview-1353462792 

### DIFF
--- a/internal/pkg/cli/env_package.go
+++ b/internal/pkg/cli/env_package.go
@@ -194,8 +194,8 @@ func (o *packageEnvOpts) Execute() error {
 	}
 	if o.showDiff {
 		if err := diff(packager, res.Template, o.diffWriter); err != nil {
-			var errNonEmptyDiff *errNonEmptyDiff
-			if errors.As(err, &errNonEmptyDiff) {
+			var errHasDiff *errHasDiff
+			if errors.As(err, &errHasDiff) {
 				return err
 			}
 			return &errDiffNotAvailable{
--- a/internal/pkg/cli/env_package.go
+++ b/internal/pkg/cli/env_package.go
@@ -4,6 +4,7 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -192,7 +193,15 @@ func (o *packageEnvOpts) Execute() error {
 		return fmt.Errorf("generate CloudFormation template from environment %q manifest: %v", o.envName, err)
 	}
 	if o.showDiff {
-		return diff(packager, res.Template, o.diffWriter)
+		if err := diff(packager, res.Template, o.diffWriter); err != nil {
+			var errNonEmptyDiff errNonEmptyDiff
+			if errors.Is(err, &errNonEmptyDiff) {
+				return err
+			}
+			return &errDiffNotAvailable{
+				parentErr: err,
+			}
+		}
 	}
 	addonsTemplate, err := packager.AddonsTemplate()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


KollaAdithya (approved):

internal/pkg/cli/env_package.go
```diff
@@ -192,7 +193,15 @@ func (o *packageEnvOpts) Execute() error {
 		return fmt.Errorf("generate CloudFormation template from environment %q manifest: %v", o.envName, err)
 	}
 	if o.showDiff {
-		return diff(packager, res.Template, o.diffWriter)
+		if err := diff(packager, res.Template, o.diffWriter); err != nil {
+			var errNonEmptyDiff errNonEmptyDiff
+			if errors.Is(err, &errNonEmptyDiff) {
```
dannyrandall: i think this should be `errors.As`

internal/pkg/cli/svc_deploy.go
```diff
@@ -582,16 +586,35 @@ func (e *errFeatureIncompatibleWithEnvironment) RecommendActions() string {
 
 }
 
+type errNonEmptyDiff struct{}
+
+func (e *errNonEmptyDiff) Error() string {
+	return "diff detected "
```
dannyrandall:
```suggestion
	return "diff detected"
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -582,16 +586,35 @@ func (e *errFeatureIncompatibleWithEnvironment) RecommendActions() string {
...
+func (e *errNonEmptyDiff) ExitCode() int {
+	return 1
+}
+
 func diff(differ templateDiffer, tmpl string, writer io.Writer) error {
 	out, err := differ.DeployDiff(tmpl)
 	if err != nil {
 		return err
 	}
+	var nonDiff bool
```
dannyrandall:
```suggestion
	var noDiff bool
```

internal/pkg/cli/svc_deploy.go
```diff
@@ -582,16 +586,35 @@ func (e *errFeatureIncompatibleWithEnvironment) RecommendActions() string {
...
 	}
 	_, err = writer.Write([]byte(out))
-	return err
+	if err != nil {
+		return err
+	}
+	if nonDiff {
+		return nil
+	}
+	return &errNonEmptyDiff{}
```
dannyrandall: non empty diff confuses me haha...could we change the wording/logic to `noDiff`?

Lou1415926:
```
out, err := differ.DeployDiff(tmpl)
	if err != nil {
		return err
	}
	if out != "" {
		if _, err := writer.Write([]byte(out)); err != nil {
			return err
		}
		return &errHasDiff{}
	}
	if _, err := writer.Write([]byte("No changes.\n")); err != nil {
		return err
	}
	return nil
```

Better?

internal/pkg/cli/svc_deploy.go
```diff
@@ -582,16 +586,32 @@ func (e *errFeatureIncompatibleWithEnvironment) RecommendActions() string {
...
 		return err
 	}
-	if out == "" {
-		out = "No changes.\n"
+	if out != "" {
+		if _, err := writer.Write([]byte(out)); err != nil {
+			return err
+		}
+		return &errHasDiff{}
+	}
```
paragbhingre: How about ?

```
if out, err := differ.DeployDiff(tmpl); err != nil {
	return err
} else if out != "" {
	if _, err := writer.Write([]byte(out)); err != nil {
		return err
	}
	return &errHasDiff{}
}
```

paragbhingre (approved):
LGTM! DNM just in case any comment left to be addressed from Danny. Feel free to remove it.

PR #4688, opened 2023-03-24
chore: render YAML tags such as `!Ref` in `diff`

Lou1415926:

### DIFF
--- a/internal/pkg/template/diff/formatter.go
+++ b/internal/pkg/template/diff/formatter.go
@@ -30,7 +30,14 @@ func (f *seqItemFormatter) formatYAML(node *yaml.Node) ([]byte, error) {
 }
 
 func (f *seqItemFormatter) formatMod(node diffNode) string {
-	return fmt.Sprintf("- %s -> %s", node.oldYAML().Value, node.newYAML().Value)
+	oldValue, newValue := node.oldYAML().Value, node.newYAML().Value
+	if v, err := yaml.Marshal(node.oldYAML()); err == nil { // NOTE: Marshal handles YAML tags such as `!Ref` and `!Sub`.
+		oldValue = strings.TrimSuffix(string(v), "\n")
+	}
+	if v, err := yaml.Marshal(node.newYAML()); err == nil {
+		newValue = strings.TrimSuffix(string(v), "\n")
+	}
+	return fmt.Sprintf("- %s -> %s", oldValue, newValue)
 }
 
 func (f *seqItemFormatter) formatPath(_ diffNode) string {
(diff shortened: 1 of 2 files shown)


dannyrandall (approved):

internal/pkg/template/diff/formatter.go
```diff
@@ -30,7 +30,14 @@ func (f *seqItemFormatter) formatYAML(node *yaml.Node) ([]byte, error) {
 }
 
 func (f *seqItemFormatter) formatMod(node diffNode) string {
-	return fmt.Sprintf("- %s -> %s", node.oldYAML().Value, node.newYAML().Value)
+	oldValue, newValue := node.oldYAML().Value, node.newYAML().Value
+	if v, err := yaml.Marshal(node.oldYAML()); err == nil { // NOTE: Marshal handles YAML tags such as `!Ref` and `!Sub`.
+		oldValue = strings.TrimSuffix(string(v), "\n")
```
iamhopaul123, line 35: I feel like this is always the value we want instead of `node.oldYAML().Value`. Does it make more sense to error out if we fail to `yaml.Marshal(node.oldYAML())`? In case they use `! Ref` or similar and the yaml marshal fails at the same time, and we'll just swallow the error and use `node.oldYAML().Value` which might not be expected.

Lou1415926: Realistically, I think it's unlikely that this will error out, because:
1. The yaml node was unmarshalled from a working YAML file
2. The yaml node isn't modified

If for whatever reason there is indeed a marshal error, I thought `node.oldYAML().Value` should already be enough for users to pinpoint the location of the change, in most cases. Therefore, surfacing `node.oldYAML().Value` might be moroe helpful than surfacing the error (again, if there is any 😂 ) in this case 🤔

iamhopaul123 (approved):

internal/pkg/template/diff/formatter.go
```diff
@@ -68,7 +75,14 @@ func (f *keyedFormatter) formatYAML(node *yaml.Node) ([]byte, error) {
 }
 
 func (f *keyedFormatter) formatMod(node diffNode) string {
-	return fmt.Sprintf("%s: %s -> %s", node.key(), node.oldYAML().Value, node.newYAML().Value)
+	oldValue, newValue := node.oldYAML().Value, node.newYAML().Value
```
efekarakus, line 78: can we dedupe the code ? 

like a 

```go
type yamlNode yaml.Node
func (cur yamlNode) modString(other yamlNode) (string, error)
```

PR #4690, opened 2023-03-27
chore: add YAML patch metrics on override

efekarakus:

### DIFF
--- a/internal/pkg/override/patch.go
+++ b/internal/pkg/override/patch.go
@@ -118,7 +118,7 @@ func unmarshalPatches(path string, fs afero.Fs) ([]yamlPatch, error) {
 // addYAMLPatchDescription updates the Description field of a CloudFormation
 // to indicate it has been overriden with YAML patches for us to keep track of usage metrics.
 func addYAMLPatchDescription(body *yaml.Node) {
-	if body.Kind != yaml.DocumentNode {
+	if body.Kind != yaml.DocumentNode || len(body.Content) == 0 {
 		return
 	}
 	body = body.Content[0] // Move inside the document.
--- a/internal/pkg/override/patch.go
+++ b/internal/pkg/override/patch.go
@@ -18,7 +18,7 @@ const (
 	yamlPatchFile        = "cfn.patches.yml"
 )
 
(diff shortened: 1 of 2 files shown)


internal/pkg/override/patch.go
```diff
@@ -114,6 +115,23 @@ func unmarshalPatches(path string, fs afero.Fs) ([]yamlPatch, error) {
 	return patches, nil
 }
 
+// addYAMLPatchDescription updates the Description field of a CloudFormation
+// to indicate it has been overriden with YAML patches for us to keep track of usage metrics.
+func addYAMLPatchDescription(body *yaml.Node) {
+	if body.Kind != yaml.DocumentNode {
```
dannyrandall:
```suggestion
	if body.Kind != yaml.DocumentNode || len(body.Content) == 0 {
```

internal/pkg/override/patch.go
```diff
@@ -114,6 +115,23 @@ func unmarshalPatches(path string, fs afero.Fs) ([]yamlPatch, error) {
...
+		return
+	}
+	body = body.Content[0] // Move inside the document.
+	for i := 0; i < len(body.Content); i += 2 {
```
dannyrandall, line 125: i'm fine to leave it this way too
```suggestion
    i, err := findInMap(body, "Description", nil)
	if err != nil {
    }
```
https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/override/patch.go#L259

dannyrandall (approved):

internal/pkg/override/patch.go
```diff
@@ -114,6 +115,23 @@ func unmarshalPatches(path string, fs afero.Fs) ([]yamlPatch, error) {
 	return patches, nil
 }
 
+// addYAMLPatchDescription updates the Description field of a CloudFormation
+// to indicate it has been overriden with YAML patches for us to keep track of usage metrics.
+func addYAMLPatchDescription(body *yaml.Node) {
```
Lou1415926, line 120: Do we want to exclude those `Description`s from `Parameters`?

Lou1415926 (approved):

PR #4697, opened 2023-03-28
chore: perform `docker login` only once for all the images

KollaAdithya:
In the previous PR #4653 (which is reverted) is printing the below messages even when we run `copilot svc override` as shown below.
image

This PR will fix that issue by removing the  `docker login` process  in `workloadDeployer` constructor and  instead perform `docker login` just before the Build and Push of the images occurs.  In this way we ensure that we print `Login Succeeded` only if we build and push any images.

Summary of the changes.

1. Refactor: move `dockerCmdClient` to Repository struct.
2. Perform `docker Login` just before `BuildAndPush` of images occurs for all `workloads` and `task`.
3. Remove [r.URI()](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/cli/task_run.go#L730) in `task_run.go` and instead use the `uri` from `repository.Login()`.


internal/pkg/cli/deploy/workload.go
```diff
@@ -68,7 +68,8 @@ func (noopActionRecommender) RecommendedActions() []string {
 	return nil
 }
 
-type imageBuilderPusher interface {
+type repositoryService interface {
+	Login(docker repository.ContainerLoginBuildPusher) (string, string, error)
```
dannyrandall: how much do we care about showing the output of the `docker login` command if it succeeds? if we don't care, we could remove the second string output and only include the output in the case of an error 🤔. just a thought - i'm happy either way!

internal/pkg/cli/deploy/workload.go
```diff
@@ -230,8 +234,13 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 	}
 
 	repoName := fmt.Sprintf("%s/%s", in.App.Name, in.Name)
-	imageBuilderPusher := repository.NewWithURI(
+	repository := repository.NewWithURI(
 		ecr.New(defaultSessEnvRegion), repoName, resources.RepositoryURLs[in.Name])
+	dockerCmdClient := dockerengine.New(exec.NewCmd())
+	uri, loginOut, err := repository.Login(dockerCmdClient)
+	if err != nil {
+		return nil, fmt.Errorf("login to docker: %w", err)
```
dannyrandall: should probably include `loginOut` on error, otherwise it won't be shown

internal/pkg/cli/deploy/workload.go
```diff
@@ -253,30 +262,32 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
...
+		templateFS:               template.New(),
+		envVersionGetter:         in.EnvVersionGetter,
+		overrider:                in.Overrider,
+		customResources:          in.customResources,
 		defaultSess:              defaultSession,
 		defaultSessWithEnvRegion: defaultSessEnvRegion,
 		envSess:                  envSession,
 		store:                    store,
 		envConfig:                envConfig,
+		dockerCmdClient:          dockerCmdClient,
```
dannyrandall: what do you think about embedding this in `repository` instead? it seems like it's only used when calling functions on `repository` anyways

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,7 +378,8 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	}
 	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
 	for name, buildArgs := range buildArgsPerContainer {
-		digest, err := d.imageBuilderPusher.BuildAndPush(dockerengine.New(exec.NewCmd()), buildArgs)
+		fmt.Print(d.dockerLoginOut)
```
dannyrandall: probably don't need to print out the login information on every container build - maybe just once before? (if at all; re: my prev comment!)

internal/pkg/cli/deploy/workload.go
KollaAdithya: I thought it would be good to display that the `docker login` is successful. I am happy to change to change if we do not need to display `login Succeeded`.

internal/pkg/cli/deploy/workload.go
KollaAdithya: Done. moved to `repository` pkg.

internal/pkg/cli/deploy/workload.go
```diff
@@ -230,8 +232,12 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 	}
 
 	repoName := fmt.Sprintf("%s/%s", in.App.Name, in.Name)
-	imageBuilderPusher := repository.NewWithURI(
+	repository := repository.NewWithURI(
 		ecr.New(defaultSessEnvRegion), repoName, resources.RepositoryURLs[in.Name])
+	uri, loginOut, err := repository.Login()
+	if err != nil {
+		return nil, fmt.Errorf("login to docker %s: %w", loginOut, err)
```
```suggestion
		return nil, fmt.Errorf("login to docker: %w. `docker login` output: %s", err, loginOut)
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -142,28 +142,30 @@ type GenerateCloudFormationTemplateOutput struct {
...
+	name           string
+	app            *config.Application
+	env            *config.Environment
+	image          ContainerImageIdentifier
+	resources      *stack.AppRegionalResources
+	mft            interface{}
+	rawMft         []byte
+	workspacePath  string
+	uri            string
+	dockerLoginOut string
```
```suggestion
	dockerLoginOutput string
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -358,16 +365,17 @@ func (img ContainerImageIdentifier) Tag() string {
...
-	buildArgsPerContainer, err := buildArgsPerContainer(d.name, d.workspacePath, d.image, d.mft)
+	buildArgsPerContainer, err := buildArgsPerContainer(d.name, d.workspacePath, d.uri, d.image, d.mft)
 	if err != nil {
 		return err
 	}
 	if len(buildArgsPerContainer) == 0 {
 		return nil
 	}
 	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
+	fmt.Print(d.dockerLoginOut)
```
dannyrandall: `Println`?

internal/pkg/cli/deploy/workload.go
dannyrandall: could we move the `Login()` call to `uploadContainerImages()`? then we don't have to save the docker login output on `workloadDeployer`!

KollaAdithya: So if we move `Login()` to the `uploadContainerImages()`  function then we perform `Login` even when we upload custom resources, EnvFiles. Is this fine?

dannyrandall: i think `uploadContainerImages()` only gets called once? [sequentially when we upload env files and custom resources](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/cli/deploy/backend.go#L63-L65)? where are you seeing that it'll get called multiple times?

KollaAdithya: I mean if there are no container images to upload but we have env files to upload then also we perform `Login()`.

KollaAdithya: So I have the below code

 ```
	uri, dockerLoginOutput, err := d.repository.Login()
	if err != nil {
		return fmt.Errorf("login to docker: %w. `docker login` output: %s", err, dockerLoginOutput)
	}
	buildArgsPerContainer, err := buildArgsPerContainer(d.name, d.workspacePath, uri, d.image, d.mft)
```
I am performing Login just before the `buildArgsPerContainer()` to pass `URI ` as an argument to set `buildArgs`. As a result, the `Login` method is executed even when there are no container images to upload.

Comment (dannyrandall) [reply internal/pkg/cli/deploy/workload.go]: ooooh i see what you're saying. i actually think we want the new behavior of only logging in if we're going to upload containers. The `Login()` function logs in the local `docker` engine in to ECR, which we only need if we're pushing container images there!

File: internal/pkg/cli/task_run_test.go [hunk 8]
```diff
@@ -844,11 +843,10 @@ func TestTaskRunOpts_Execute(t *testing.T) {
 					Command:    []string{},
 					EntryPoint: []string{},
 				}).Return(nil)
-				m.repository.EXPECT().BuildAndPush(gomock.Any(), gomock.Eq(&defaultBuildArguments))
-				m.repository.EXPECT().URI().Return(mockRepoURI, errors.New("some error"))
+				m.repository.EXPECT().Login().Return(mockRepoURI, errors.New("some error"))
 				mockHasDefaultCluster(m)
 			},
-			wantedError: errors.New("get ECR repository URI: some error"),
+			wantedError: errors.New("login to docker: some error"),
```
dannyrandall, line 849: nit: rename this test

internal/pkg/repository/repository.go
```diff
@@ -26,54 +28,45 @@ type Registry interface {
 
 // Repository builds and pushes images to a repository.
 type Repository struct {
-	name     string
-	registry Registry
-	uri      string
+	name            string
+	registry        Registry
+	uri             string
+	dockerCmdClient ContainerLoginBuildPusher
```
dannyrandall: nit
```suggestion
	docker ContainerLoginBuildPusher
```

internal/pkg/cli/deploy/workload.go
KollaAdithya: Done! moved the `Login` to `uploadContainerImages`

dannyrandall (approved):
looks great! thanks @KollaAdithya 🚀

internal/pkg/cli/deploy/workload.go
```diff
@@ -365,9 +364,14 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	if len(buildArgsPerContainer) == 0 {
 		return nil
 	}
+	uri, err := d.repository.Login()
+	if err != nil {
+		return fmt.Errorf("login to docker: %w", err)
```
Lou1415926: nit: since it is not necessarily docker hub!
```suggestion
		return fmt.Errorf("login to image repository: %w", err)
```

internal/pkg/repository/repository.go
```diff
@@ -45,35 +49,24 @@ func NewWithURI(registry Registry, name, uri string) *Repository {
...
 // BuildAndPush builds the image from Dockerfile and pushes it to the repository with tags.
-func (r *Repository) BuildAndPush(docker ContainerLoginBuildPusher, args *dockerengine.BuildArguments) (digest string, err error) {
+func (r *Repository) BuildAndPush(args *dockerengine.BuildArguments) (digest string, err error) {
 	if args.URI == "" {
 		uri, err := r.URI()
 		if err != nil {
 			return "", err
 		}
 		args.URI = uri
 	}
```
Lou1415926, line 64: do we still need this chunk?

internal/pkg/repository/repository.go
```diff
@@ -91,3 +84,25 @@ func (r *Repository) URI() (string, error) {
...
+		return "", fmt.Errorf("retrieve URI for repository: %w", err)
+	}
+	if !r.docker.IsEcrCredentialHelperEnabled(uri) {
+		username, password, err := r.registry.Auth()
+		if err != nil {
+			return "", fmt.Errorf("get auth: %w", err)
+		}
+
+		if err := r.docker.Login(uri, username, password); err != nil {
+			return "", fmt.Errorf("login to repo %s: %w", uri, err)
```
Lou1415926:
```suggestion
			return "", fmt.Errorf("docker login %s: %w", uri, err)
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -365,9 +364,14 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+	}
 	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
 	for name, buildArgs := range buildArgsPerContainer {
-		digest, err := d.imageBuilderPusher.BuildAndPush(dockerengine.New(exec.NewCmd()), buildArgs)
+		buildArgs.URI = uri
```
Lou1415926: I might be missing something, but why do we need to pass `URI` as a build args 🤔 ? It looks like `URI` is stored as an internal state of `repository`

Lou1415926 (commented):
LG! I have just a few nits and questions

internal/pkg/repository/repository.go
KollaAdithya: This can be removed. we can just use `args.URI` that is set in `workload.go`

internal/pkg/cli/deploy/workload.go
KollaAdithya: So when `Login` is called we already making an [api call](https://github.com/aws/copilot-cli/pull/4697/files#diff-c3341fa47b7006ee5fa20ac85575594903aa23154c1427d2daa666e074d87e75R93) to fetch the `URI` of the repository. if we set `URI` in `buildArgs` then we do not have to again make an [api call](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/repository/repository.go#L54) to fetch the URI.

Lou1415926: But we won't make the additional API call because `uri` is stored in `repository`, right? https://github.com/aws/copilot-cli/blob/67db7a2d4f94c69c836586f714341e587b1b8220/internal/pkg/repository/repository.go#L84

```go
// URI returns the uri of the repository.
func (r *Repository) URI() (string, error) {
	if r.uri != "" {
		return r.uri, nil
	}
	uri, err := r.registry.RepositoryURI(r.name)
	if err != nil {
		return "", fmt.Errorf("get repository URI: %w", err)
	}
	return uri, nil
}
```

KollaAdithya: Yeah that is true in case of `workload.go` as URI is in repository. But in case of [taskrun.go](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/cli/task_run.go#L224), URI is not set in Repository.

Lou1415926:
```go
// URI returns the uri of the repository.
func (r *Repository) URI() (string, error) {
	if r.uri != "" {
		return r.uri, nil
	}
	uri, err := r.registry.RepositoryURI(r.name)
	if err != nil {
		return "", fmt.Errorf("get repository URI: %w", err)
	}
        r.uri = uri // <-----This
	return uri, nil
}
```
Maybe we should set `r.uri = uri` before returning 🤔 

It just feels a bit weird that we are passing `uri` around, both from `repository` to `workload`/ `task_run`, and then again from `workload` / `task_run` back to `repository` 😂

internal/pkg/repository/repository.go
KollaAdithya: I removed the logic to set URI in BuildArgs for `workload/task_run`. Then I think we need to need the above chunk.

internal/pkg/cli/deploy/workload.go
KollaAdithya: Done. changed the logic to set uri in `repository`.

internal/pkg/cli/interfaces.go
```diff
@@ -166,15 +165,19 @@ type secretDeleter interface {
...
 type imageBuilderPusher interface {
-	BuildAndPush(docker repository.ContainerLoginBuildPusher, args *dockerengine.BuildArguments) (string, error)
+	BuildAndPush(args *dockerengine.BuildArguments) (string, error)
 }
-
 type repositoryURIGetter interface {
 	URI() (string, error)
 }
 
+type dockerLogin interface {
```
Lou1415926: nit: the interface function seems docker-agnostic!

```suggestion
type repositoryLogin interface {
```

internal/pkg/cli/task_run.go
```diff
@@ -719,12 +719,16 @@ func (o *runTaskOpts) Execute() error {
 
 	// NOTE: if image is not provided, then we build the image and push to ECR repo
 	if o.image == "" {
-		uri, err := o.repository.Login()
+		uri, err := o.repository.URI()
+		if err != nil {
+			return fmt.Errorf("get ECR repository URI: %w", err)
+		}
```
Lou1415926: nit: we can maybe move ethis line to just before `o.image = fmt.Sprintf(fmtImageURI, uri, tag)`, if that is the only place we need `uri` for.

Lou1415926 (approved):

Issue #4681, opened 2023-03-24
Failed create 'copilot' folder in current dir because an account named 'copilot' exists on Mac


idawnwon:
This thing really makes me laugh and cry 😂.

I just follow [the starter guide](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/AWS_Copilot.html), very careful, step by step. It keeps saying `couldn't find an application associated with this workspace`.

Yes, there is no folder named 'copilot' in my current work directory. 

According to #3262, I tried `find / -name "copilot"`, on my Mac, there is an account named "Copilot", the home folder of it is 'copilot', so its folder is searched out. By that time, I believe this couldn't be the problem. My working folder is in my account, very deep, very far from the "Copilot" account.

According to #4195, I tried to find some where above `../../../../`, OMG, here comes the 'copilot' folder of that account. I `cd` into it, no manifest file is there.  I believe this couldn't be the problem, again.

I tried to uninstall copilot-cli, reinstall by another way. Restart computer ....

3 hours later, doesn't work. What pissed me off is, even I wanna give up with copilot-cli, I found no way to delete those already created apps. Deleted CloudFormation stacks in AWS Console, `copilot app ls`, they are still listed there.

Finally, I deleted the "Copilot" account. Problem gone. `copilot init`, a blink 'copilot' folder is created in my current work directory.

I know that this is really one edge case. But it reflects copilot-cli really has to give more explicit warning or errors to prevent user guessing where was the 'copilot' folder gone.

Whatever, I still wanna thank you guys for creating Copilot, really handy, after all. ❤️

huanjani:
Hi @idawnwon! 

Thank you for telling us about your experience. If we had a warning in the docs, maybe on [this page](https://aws.github.io/copilot-cli/docs/getting-started/first-app-tutorial/), do you think that would prevent this from happening to someone else? Did you use the docs at https://aws.github.io/copilot-cli/, or just the starter guide you mentioned above? I'm just curious about the user experience. 

Also, for future reference, you can delete the traces of apps so they're not listed with `copilot app ls` by deleting them from your Systems Manager Parameter Store.

Thanks and sorry
Related: [#3294](https://github.com/aws/copilot-cli/issues/3294)

idawnwon:
@huanjani, thank you for your timely reply! ♥️ 

[...]

Issue #3294, opened 2022-02-17
Copilot workspace/manifest files are written to an unrelated directory coincidentally named copilot.


cristobalmackenzie:
Hi,

I was following the quickstart from the docs, and initialized an app with `copilot init`.

After completing the quickstart, I had trouble deleting the `.workspace` file and couldn't figure out why. I later found that the .workspace and manifest files where created in another directory next to the one I was on, that happens to be called copilot also but is unrelated.

Running `myuser@/repos/example-directory $ copilot init` created all copilot files in a previously existing `/repos/copilot` directory, instead of in `/repos/example-directory/copilot`.

The copilot cli seems perfectly able to find these files which end up in an unrelated folder, except for deleting them after running `copilot app delete`.

huanjani:
Hi, @cristobalmackenzie! 

Eek, sorry about that! Thanks for bringing this to our attention. We wanted to give users the flexibility to run commands from different places in their filesystems, but this is an unintended consequence. We will look into preventing cases of mistaken 🧑‍✈️ identity from happening again!

PR #4699, opened 2023-03-28
chore(docs): add warning re: existing dirs named "copilot"

huanjani:
Please help with the wording of this! 🥺 
Addresses: #3294, #4681.


Lou1415926 (commented):
I have two little thoughts to share 💭 

1. Looking at https://github.com/aws/copilot-cli/issues/4681, I wonder if searching up to 5 levels is too deep? Maybe we just need to search for 3 levels?
2. Instead of suggesting that users rename their existing `copilot` folder, I wonder if we should instead suggest that users create a `copilot` folder at the correct place, so that we won't look for the wrong `copilot` further deep down.

huanjani:
> 1. Looking at #4681, I wonder if searching up to 5 levels is too deep? Maybe we just need to search for 3 levels?

It could be possibly backwards-incompatible to change it though, right?

> 2. Instead of suggesting that users rename their existing `copilot` folder, I wonder if we should instead suggest that users create a `copilot` folder at the correct place, so that we won't look for the wrong `copilot` further deep down.

Yeah, I felt a little uncomfortable asking people to rename their dir. I like your solution!

site/content/docs/concepts/applications.en.md
```diff
@@ -10,6 +10,9 @@ Your application configuration (which services and environments belong to it) is
 
 ## Creating an App
 
+!!! Attention
+If you have an existing directory named `copilot/` in your local directory that was not created by AWS Copilot, please ,make an empty directory also named `copilot` closer to your working directory.
```
Lou1415926:
```suggestion
    If you have an existing directory named `copilot/` in your local directory that was not created by AWS Copilot, please make an empty directory also named `copilot` closer to your working directory.
```

Lou1415926: In addition, what do you think if we describe the potential problem first? Something like:
> If you have an existing `copilot/` directory that you created for other purposes, you might find Copilot creating files in that directory. If this happens, you can create an empty directory named `copilot/` closer to your working directory. Copilot will instead use this empty directory.

Lou1415926 (approved):

dannyrandall (approved):

PR #4716, opened 2023-03-30
chore: add `CommandContext` to `exec` package for context cancellation

KollaAdithya:
This PR adds `Command Context` to exec package that helps to interrupt the process  if the context becomes done before the command completes on its own.


internal/pkg/exec/exec.go
Lou1415926, line 1: what do you think of this 👁️ 
```go
// Cmd runs external commands, it wraps the exec.Command function from the stdlib so that
// running external commands can be unit tested.
type Cmd struct {
	command func(ctx context.Context, name string, args []string, opts ...CmdOption) cmdRunner
}

func NewCmd() *Cmd {
	return &Cmd{
		command: func(ctx context.Context, name string, args []string, opts ...CmdOption) cmdRunner {
			cmd := exec.CommandContext(ctx, name, args...)
	                cmd.Stdout = os.Stderr
	                cmd.Stderr = os.Stderr
	                for _, opt := range opts {
		                opt(cmd)
	                }
	                return cmd
		 }
	}
}

// Run starts the named command and waits until it finishes.
func (c *Cmd) Run(name string, args []string, opts ...CmdOption) error {
	cmd := c.command(context.Background(), name, args, opts...)
	return cmd.Run()
}

// RunWithContext starts the named command with the given context.
// Command execution process will be killed if the context becomes done before the command completes on its own.
func (c *Cmd) RunWithContext(ctx context.Context, name string, args []string, opts ...CmdOption) error {
	cmd := c.command(ctx, name, args, opts...)
	return cmd.Run()
}
```

Lou1415926 (commented):
great job splitting the codes into small bites!

efekarakus (approved):

Lou1415926 (approved):

PR #4718, opened 2023-03-30
chore: perform parallel build for all container images without context cancellation

KollaAdithya:
This PR adds the logic to parallel build and push images with out context cancellation i.e if one of the go routine exits with an error, the other will still perform build and push of images.

The logic for context Cancellation will be added in the next PR.

Below are the screenshots of how the UI looks in different scenarios.

1. All the images are successfully build and pushed
Screenshot 2023-03-29 at 9 50 02 PM

2. If path of Dockerfile is incorrect.
Screenshot 2023-03-29 at 9 45 27 PM

3. If one of the docker build gives an error.
Screenshot 2023-03-29 at 9 53 58 PM

### DIFF
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -234,10 +234,6 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 	repoName := fmt.Sprintf("%s/%s", in.App.Name, in.Name)
 	repository := repository.NewWithURI(
 		ecr.New(defaultSessEnvRegion), repoName, resources.RepositoryURLs[in.Name])
-	uri, loginOut, err := repository.Login()
-	if err != nil {
-		return nil, fmt.Errorf("login to docker %s: %w", loginOut, err)
-	}
 	store := config.NewSSMStore(identity.New(defaultSession), ssm.New(defaultSession), aws.StringValue(defaultSession.Config.Region))
 	envDescriber, err := describe.NewEnvDescriber(describe.NewEnvDescriberConfig{
 		App:         in.App.Name,
@@ -282,8 +278,6 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 		envSess:                  envSession,
 		store:                    store,
 		envConfig:                envConfig,
-		uri:                      uri,
-		dockerLoginOut:           loginOut,
 
 		mft:    in.Mft,
 		rawMft: in.RawMft,
@@ -372,13 +366,19 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	if len(buildArgsPerContainer) == 0 {
 		return nil
 	}
+<<<<<<< HEAD
 	err = d.repository.Login()
 	if err != nil {
 		return fmt.Errorf("login to image repository: %w", err)
+=======
+	uri, err := d.repository.Login()
+	if err != nil {
+		return fmt.Errorf("login to docker: %w", err)
+>>>>>>> cafd5b3f (remove Login() from workload constructor and perform in uploadcontainerImages())
 	}
 	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
-	fmt.Print(d.dockerLoginOut)
 	for name, buildArgs := range buildArgsPerContainer {
+		buildArgs.URI = uri
 		digest, err := d.repository.BuildAndPush(buildArgs)
 		if err != nil {
 			return fmt.Errorf("build and push image: %w", err)
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -5,6 +5,7 @@ package deploy
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -43,6 +44,7 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/version"
 	"github.com/aws/copilot-cli/internal/pkg/workspace"
 	"github.com/spf13/afero"
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/term"
 )
 
@@ -421,28 +423,72 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	if len(buildArgsPerContainer) == 0 {
 		return nil
 	}
-<<<<<<< HEAD
 	err = d.repository.Login()
 	if err != nil {
 		return fmt.Errorf("login to image repository: %w", err)
-=======
-	uri, err := d.repository.Login()
-	if err != nil {
-		return fmt.Errorf("login to docker: %w", err)
->>>>>>> cafd5b3f (remove Login() from workload constructor and perform in uploadcontainerImages())
 	}
 	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
+
+	// create an array of output buffers, one for each container.
+	buffers := make([]*buildPushOutputBuffer, len(buildArgsPerContainer))
+
+	// create a context and an error group.
+	ctx := context.Background()
+	g, _ := errgroup.WithContext(ctx)
+
+	// counter for indexing the output buffers.
+	count := 0
+
+	// mutex for synchronizing access to the output map.
+	var mux sync.Mutex
 	for name, buildArgs := range buildArgsPerContainer {
 		buildArgs.URI = uri
-		digest, err := d.repository.BuildAndPush(buildArgs)
-		if err != nil {
-			return fmt.Errorf("build and push image: %w", err)
+		// create a copy of loop variables to avoid data race.
+		name := name
+		buildArgs := buildArgs
+
+		// create a pipe for streaming the build and push output.
+		pr, pw := io.Pipe()
+
+		buffer := &buildPushOutputBuffer{
+			ContainerName: name,
 		}
-		out.ImageDigests[name] = ContainerImageIdentifier{
-			Digest:            digest,
-			CustomTag:         d.image.CustomTag,
-			GitShortCommitTag: d.image.GitShortCommitTag,
-		}
+		buffers[count] = buffer
+		count++
+
+		// create a new cursor and hide it.
+		curs := cursor.New()
+		curs.Hide()
+		defer curs.Show()
+
+		g.Go(func() error {
+			defer pw.Close()
+			digest, err := d.repository.BuildAndPush(buildArgs, pw)
+			if err != nil {
+				return fmt.Errorf("build and push image: %w", err)
+			}
+			mux.Lock()
+			defer mux.Unlock()
+			out.ImageDigests[name] = ContainerImageIdentifier{
+				Digest:            digest,
+				CustomTag:         d.image.CustomTag,
+				GitShortCommitTag: d.image.GitShortCommitTag,
+			}
+			return nil
+		})
+
+		g.Go(func() error {
+			return copyOutputToBuffer(pr, buffer)
+		})
+	}
+
+	g.Go(func() error {
+		return printOutputFromBuffers(buffers)
+	})
+
+	// wait for all goroutines to complete and return any errors.
+	if err := g.Wait(); err != nil {
+		return err
 	}
 	return nil
 }
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -28,6 +28,7 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/deploy/upload/customresource"
 	"github.com/aws/copilot-cli/internal/pkg/describe"
 	"github.com/aws/copilot-cli/internal/pkg/docker/dockerengine"
+	"github.com/aws/copilot-cli/internal/pkg/exec"
 	"github.com/aws/copilot-cli/internal/pkg/manifest"
 	"github.com/aws/copilot-cli/internal/pkg/repository"
 	"github.com/aws/copilot-cli/internal/pkg/template"
@@ -142,14 +143,16 @@ type GenerateCloudFormationTemplateOutput struct {
 }
 
 type workloadDeployer struct {
-	name          string
-	app           *config.Application
-	env           *config.Environment
-	image         ContainerImageIdentifier
-	resources     *stack.AppRegionalResources
-	mft           interface{}
-	rawMft        []byte
-	workspacePath string
+	name           string
+	app            *config.Application
+	env            *config.Environment
+	image          ContainerImageIdentifier
+	resources      *stack.AppRegionalResources
+	mft            interface{}
+	rawMft         []byte
+	workspacePath  string
+	uri            string
+	dockerLoginOut string
 
 	// Dependencies.
 	fs               fileReader
@@ -171,6 +174,7 @@ type workloadDeployer struct {
 	envSess                  *session.Session
 	store                    *config.Store
 	envConfig                *manifest.Environment
+	dockerCmdClient          dockerengine.CmdClient
 }
 
 // WorkloadDeployerInput is the input to for workloadDeployer constructor.
@@ -232,6 +236,11 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 	repoName := fmt.Sprintf("%s/%s", in.App.Name, in.Name)
 	repository := repository.NewWithURI(
 		ecr.New(defaultSessEnvRegion), repoName, resources.RepositoryURLs[in.Name])
+	dockerCmdClient := dockerengine.New(exec.NewCmd())
+	uri, loginOut, err := repository.Login(dockerCmdClient)
+	if err != nil {
+		return nil, fmt.Errorf("login to docker: %w", err)
+	}
 	store := config.NewSSMStore(identity.New(defaultSession), ssm.New(defaultSession), aws.StringValue(defaultSession.Config.Region))
 	envDescriber, err := describe.NewEnvDescriber(describe.NewEnvDescriberConfig{
 		App:         in.App.Name,
@@ -276,6 +285,9 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 		envSess:                  envSession,
 		store:                    store,
 		envConfig:                envConfig,
+		dockerCmdClient:          dockerCmdClient,
+		uri:                      uri,
+		dockerLoginOut:           loginOut,
 
 		mft:    in.Mft,
 		rawMft: in.RawMft,
@@ -357,7 +369,7 @@ func (img ContainerImageIdentifier) Tag() string {
 
 func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) error {
 	// If it is built from local Dockerfile, build and push to the ECR repo.
-	buildArgsPerContainer, err := buildArgsPerContainer(d.name, d.workspacePath, d.image, d.mft)
+	buildArgsPerContainer, err := buildArgsPerContainer(d.name, d.workspacePath, d.uri, d.image, d.mft)
 	if err != nil {
 		return err
 	}
@@ -383,7 +395,7 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	return nil
 }
 
-func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentifier, unmarshaledManifest interface{}) (map[string]*dockerengine.BuildArguments, error) {
+func buildArgsPerContainer(name, workspacePath, uri string, img ContainerImageIdentifier, unmarshaledManifest interface{}) (map[string]*dockerengine.BuildArguments, error) {
 	type dfArgs interface {
 		BuildArgs(rootDirectory string) (map[string]*manifest.DockerBuildArgs, error)
 		ContainerPlatform() string
@@ -415,6 +427,7 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
 		}
 		labels[labelForContainerName] = container
 		dArgs[container] = &dockerengine.BuildArguments{
+			URI:        uri,
 			Dockerfile: aws.StringValue(buildArgs.Dockerfile),
 			Context:    aws.StringValue(buildArgs.Context),
 			Args:       buildArgs.Args,
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -82,7 +82,7 @@ func (noopActionRecommender) RecommendedActions() []string {
 
 type repositoryService interface {
 	Login() error
-	BuildAndPush(args *dockerengine.BuildArguments) (string, error)
+	BuildAndPush(args *dockerengine.BuildArguments, w io.Writer) (string, error)
 }
 
 type templater interface {
@@ -442,7 +442,6 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	// mutex for synchronizing access to the output map.
 	var mux sync.Mutex
 	for name, buildArgs := range buildArgsPerContainer {
-		buildArgs.URI = uri
 		// create a copy of loop variables to avoid data race.
 		name := name
 		buildArgs := buildArgs
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -532,7 +532,7 @@ func copyOutputToBuffer(pr io.Reader, buffer *syncbuffer.TermPrinter, name strin
 	return nil
 }
 
-// printAndErase prints the build and push output from the list of buildPushOutputBuffer buffers.
+// printAndErase prints the build and push output from the list of TermPrinter buffers.
 // It polls each buffer until all build and push calls are completed,
 // and erases the previous output after sleeping for a short duration.
 func printAndErase(t terminalWidthGetter, termPrinters []*syncbuffer.TermPrinter) error {
@@ -568,7 +568,7 @@ func printAndErase(t terminalWidthGetter, termPrinters []*syncbuffer.TermPrinter
 	return nil
 }
 
-// PrintAllprints all contents from given TermPrinter buffers once they have finished writing.
+// PrintAll prints all contents from given TermPrinter buffers once they have finished writing.
 func printAll(printers []*syncbuffer.TermPrinter) error {
 	count := 0
 	for {
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -9,7 +9,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math"
 	"os"
 	"path/filepath"
 	"strings"
@@ -41,11 +40,11 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/term/cursor"
 	"github.com/aws/copilot-cli/internal/pkg/term/log"
 	termprogress "github.com/aws/copilot-cli/internal/pkg/term/progress"
+	"github.com/aws/copilot-cli/internal/pkg/term/syncbuffer"
 	"github.com/aws/copilot-cli/internal/pkg/version"
 	"github.com/aws/copilot-cli/internal/pkg/workspace"
 	"github.com/spf13/afero"
 	"golang.org/x/sync/errgroup"
-	"golang.org/x/term"
 )
 
 const (
@@ -64,9 +63,9 @@ const (
 )
 
 const (
-	// pollInterval is the time Interval to wait between checking whether the output buffers are done.
-	pollInterval = 60 * time.Millisecond
-	maxLogLines  = 5
+	// pollIntervalForBuildPush is the time Interval to wait between checking whether the output buffers are done.
+	pollIntervalForBuildPush = 60 * time.Millisecond
+	maxLogLines              = 5
 )
 
 // ActionRecommender contains methods that output action recommendation.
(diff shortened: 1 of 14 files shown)


internal/pkg/cli/deploy/workload.go
```diff
@@ -56,6 +63,12 @@ const (
 	labelForContainerName = "com.aws.copilot.image.container.name"
 )
 
+const (
+	// pollInterval is the time Interval to wait between checking whether the output buffers are done.
+	pollInterval = 60 * time.Millisecond
```
dannyrandall: nit: since this is a big package, these should probably be named more specific to what their usage is - `pollInterval` sounds fairly generic

internal/pkg/cli/deploy/workload.go
```diff
@@ -355,6 +368,52 @@ func (img ContainerImageIdentifier) Tag() string {
 	return img.GitShortCommitTag
 }
 
+// buildPushOutputBuffer contains the output configuration for docker build and push.
+type buildPushOutputBuffer struct {
+	ContainerName string // ContainerName is the name of the container.
+
```
```suggestion
	ContainerName string // ContainerName is the name of the container.
```
nit

internal/pkg/cli/deploy/workload.go
```diff
@@ -355,6 +368,52 @@ func (img ContainerImageIdentifier) Tag() string {
...
+	// Extract the last 5 lines and store them in a fixed-size array
+	var logLines [maxLogLines]string
+	idx := 0
+	for start < end {
+		logLines[idx] = strings.TrimSpace(lines[start])
+		start++
+		idx++
+	}
+	return lines[0], logLines
```
dannyrandall: if `len(lines) < maxLogLines`, won't `logLines` contain `lines[0]`? i wonder if the label line should just be a fixed string on `buildPushOutputBuffer`?

internal/pkg/cli/deploy/workload.go
```diff
@@ -369,16 +428,66 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 		return fmt.Errorf("login to image repository: %w", err)
 	}
 	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
+
+	// create an array of output buffers, one for each container.
+	buffers := make([]*buildPushOutputBuffer, len(buildArgsPerContainer))
+
+	// create a context and an error group.
+	ctx := context.Background()
+	g, _ := errgroup.WithContext(ctx)
```
dannyrandall, line 400:
```suggestion
	g, ctx := errgroup.WithContext(context.Background())
```
the `ctx` returned by `errgroup.WithContext()` is the one that gets cancelled when `g` gets an error! we should use that context so that one failure will cancel the others.

internal/pkg/cli/deploy/workload.go
```diff
@@ -369,16 +428,66 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+
+	// counter for indexing the output buffers.
+	count := 0
+
+	// mutex for synchronizing access to the output map.
+	var mux sync.Mutex
```
```suggestion
	var mu sync.Mutex
```
nit - and declare it next to the variable it protects (in this case, `out.ImageDigests` on L430)

internal/pkg/cli/deploy/workload.go
```diff
@@ -369,16 +428,66 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+		buffer := &buildPushOutputBuffer{
+			ContainerName: name,
 		}
+		buffers[count] = buffer
+		count++
+
+		// create a new cursor and hide it.
+		curs := cursor.New()
+		curs.Hide()
+		defer curs.Show()
```
dannyrandall: do we need to create a cursor for each container?

internal/pkg/cli/deploy/workload.go
dannyrandall: oh is context going to be used in a future PR?

internal/pkg/cli/deploy/workload.go
```diff
@@ -428,6 +537,90 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
...
+	defer func() {
+		buffer.doneMu.Lock()
+		buffer.done = true
+		buffer.doneMu.Unlock()
+	}()
+
+	// copy the build and push output to the output buffer
+	_, err := io.Copy(buffer, pr)
+	if err == io.EOF {
+		return nil
```
```suggestion
	if err == io.EOF || err == nil {
		return nil
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -428,6 +537,90 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
...
+		// check whether all build and push calls are completed.
+		allDone := true
+		for _, buf := range buffers {
+			buf.doneMu.Lock()
+			if !buf.done {
+				allDone = false
+			}
+			buf.doneMu.Unlock()
+			label, logs := buf.logs()
+			outputLogs := removeEmptyStrings(logs)
```
dannyrandall: why do we want to do this? don't we want a fixed height of logs so the labels don't bounce around?

internal/pkg/cli/deploy/workload.go
```diff
@@ -428,6 +537,90 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
...
+			if len(outputLogs) > 0 {
+				fmt.Fprintln(os.Stderr, label)
+				for _, logLine := range outputLogs {
+					fmt.Fprintln(os.Stderr, "\t", logLine)
```
```suggestion
					fmt.Fprintln(os.Stderr, "  ", logLine)
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -428,6 +537,90 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
...
+				}
+				labelLength, err := dockerBuildLabelLength(label)
+				if err != nil {
+					return fmt.Errorf("get terminal size %w", err)
+				}
+				writtenLines = writtenLines + labelLength + len(outputLogs)
```
dannyrandall: i wonder if we need this for _every_ line? in case of a docker build line longer than the term. width?

internal/pkg/cli/deploy/workload.go
```diff
@@ -428,6 +537,90 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
...
+// and returns the number of lines required to display it as an integer.
+// If there is an error getting the terminal size, the function returns an error.
+func dockerBuildLabelLength(label string) (int, error) {
+	width, _, err := term.GetSize(int(os.Stderr.Fd()))
+	if err != nil {
+		return 0, err
+	}
+	numLines := math.Ceil(float64(len(label)) / float64(width))
+	return int(numLines), nil
+}
```
dannyrandall: this is awesome! i think we need to integrate this with our deployment tracker, i sometimes see weirdness with a thin terminal.

internal/pkg/cli/deploy/workload.go
dannyrandall: ah i see why it's not a fixed string, though the double-printing issue still stands 👍

dannyrandall (commented):
great work! this is going to be super cool.

internal/pkg/cli/deploy/workload.go
```diff
@@ -355,6 +368,52 @@ func (img ContainerImageIdentifier) Tag() string {
 	return img.GitShortCommitTag
 }
 
+// buildPushOutputBuffer contains the output configuration for docker build and push.
+type buildPushOutputBuffer struct {
+	ContainerName string // ContainerName is the name of the container.
```
efekarakus: nit: is there a reason why this field is exported while the other ones aren't?

internal/pkg/cli/deploy/workload.go
```diff
@@ -355,6 +368,52 @@ func (img ContainerImageIdentifier) Tag() string {
...
+
+	bufMu sync.Mutex // bufMu is a mutex to protect access to the buffer.
+
+	buf bytes.Buffer // buf is the buffer containing the output of build and push.
+
+	doneMu sync.Mutex // doneMu is a mutex to protect access to the done field.
+
+	done bool // done is a flag indicating whether the build and push is completed.
```
```suggestion
	bufMu sync.Mutex // bufMu is a mutex to protect access to the buffer.
	buf bytes.Buffer // buf is the buffer containing the output of build and push.

	doneMu sync.Mutex // doneMu is a mutex to protect access to the done field.
	done bool // done is a flag indicating whether the build and push is completed.
```

internal/pkg/cli/deploy/workload.go
efekarakus: Alternatively,
```go
if err != nil && !errors.Is(err, io.EOF) {
    return fmt.Errorf("copying build and push output for container %s: %w", buffer.ContainerName, err)
}
return nil
```

internal/pkg/cli/deploy/workload.go
efekarakus: I wonder if there should be a different UX when the environment variable `CI=true` is set so that we don't fill their logs with a bunch of cursor movement strings 💭

internal/pkg/cli/deploy/workload.go
KollaAdithya: In case if the wrong path of the Dockerfile is provided in the manifest then the buffer contain only one line of the error message  and the remaining 4 lines will be empty. If it is ok to have empty lines then i am happy to remove the logic of `removeEmptyStrings`.
Screenshot 2023-03-29 at 9 45 27 PM

internal/pkg/cli/deploy/workload.go
```diff
@@ -355,6 +368,52 @@ func (img ContainerImageIdentifier) Tag() string {
...
+
+// Write appends the given bytes to the buffer.
+func (b *buildPushOutputBuffer) Write(p []byte) (n int, err error) {
+	b.bufMu.Lock()
+	defer b.bufMu.Unlock()
+	return b.buf.Write(p)
+}
+
+// logs returns the label (i.e., the first line of the buffer) and the last five lines of the buffer.
+func (b *buildPushOutputBuffer) logs() (string, [maxLogLines]string) {
```
efekarakus: Do you think it's possible to have an API like this:

```go
type buildPushOutputBuffer struct
// Write implements the io.Writer interface.
func (buf *buildPushOutputBuffer) Write(p []byte) (n int, err error)

// Strings returns an empty slice if there are no build or push output.
// Otherwise, it returns a slice of all the lines printed while building and pushing a container.
func (buf *buildPushOutputBuffer) Strings() []string

// fileWriter is the interface to write to a file.
type fileWriter interface {
	io.Writer
	Fd() uintptr
}

// terminalBuildPushPrinter is responsible for printing a *buildPushOutputBuffer to a terminal.
type terminalBuildPushPrinter struct {
   term fileWriter  // Should be log.DiagnosticWriter.
   buf *buildPushOutputBuffer

   prevWrittenNumLines int
}

// Print renders the build push output to t.term and erases any prevWrittenNumLines.
func (t *terminalBuildPushPrinter) Print() error
// numLines returns the actual number of written lines given the terminal width.
func (t *terminalBuildPushPrinter) numLines(lines []string) int

type ciBuildPushPrinter struct {
   out io.Writer
   buf *buildPushOutputBuffer
}
// Print renders the enter content of buf to out, once the buf is done.
func (ci *ciBuildPushPrinter) Print() error
```

internal/pkg/cli/deploy/workload.go
KollaAdithya: yeah, I will include the `context cancellation` in the next PR

internal/pkg/cli/deploy/workload.go
```diff
@@ -428,6 +537,90 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
 	return dArgs, nil
 }
 
+// copyOutputToBuffer copies the build and push output from the given io.Reader to the buildPushOutputBuffer buffer.
+// return an error if copying fails or if the reader returns an unexpected error.
+func copyOutputToBuffer(pr io.Reader, buffer *buildPushOutputBuffer) error {
+	defer func() {
+		buffer.doneMu.Lock()
+		buffer.done = true
+		buffer.doneMu.Unlock()
```
efekarakus: 1. I don't quite follow why we are marking the buffer as done. Shouldn't it be only `done = true` the build and push itself is done? I might be misunderstanding the purpose of `done`.
2. What do you think of moving this to a method on the buffer `func (buf *buildPushOutputBuffer) MarkDone()` ?

internal/pkg/cli/deploy/workload.go
```diff
@@ -428,6 +537,90 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
...
+
+// printOutputFromBuffers prints the build and push output from the list of buildPushOutputBuffer buffers to stderr.
+// It polls each buffer until all build and push calls are completed,
+// and erases the previous output after sleeping for a short duration.
+func printOutputFromBuffers(buffers []*buildPushOutputBuffer) error {
+	for {
+		writtenLines := 0
+		// check whether all build and push calls are completed.
+		allDone := true
+		for _, buf := range buffers {
```
efekarakus: If we do the proposed API above, the bulk of the logic in this for-loop would move to `func (t *terminalBuildPushPrinter) Print() error`

internal/pkg/cli/deploy/workload.go
```diff
@@ -428,6 +537,90 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
...
+			buf.doneMu.Lock()
+			if !buf.done {
+				allDone = false
+			}
+			buf.doneMu.Unlock()
```
efekarakus: nit: consider moving this to a method `Done() bool`.

Perhaps after the next round of refactors and seeing the code we can consider using a `Done() <-chan struc{}` here (but just a nit)

internal/pkg/cli/deploy/workload.go
```diff
@@ -428,6 +537,90 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
...
+			label, logs := buf.logs()
+			outputLogs := removeEmptyStrings(logs)
+			if len(outputLogs) > 0 {
+				fmt.Fprintln(os.Stderr, label)
```
efekarakus: instead of `os.Stderr` can we use `log.DiagnosticWriter`

internal/pkg/cli/deploy/workload_test.go
```diff
@@ -211,7 +211,7 @@ func TestWorkloadDeployer_UploadArtifacts(t *testing.T) {
 						"com.aws.copilot.image.builder":        "copilot-cli",
 						"com.aws.copilot.image.container.name": "mockWkld",
 					},
-				}).Return("", mockError)
+				}, gomock.Any()).Return("", mockError)
```
efekarakus, line 215: I wonder if the difficulty in writing tests for this logic is a hint that we should be extracting our `buildPushOutputBuffer`, `terminalBuildPushPrinter`, `ciBuildPushPrinter` into its own package that would allow us to unit test it as well as mock the printers 💭 

Perhaps, having a separate package under `term/` would help something like:
```go
package term/syncbuffer  // A new syncbuffer package

type SyncBuffer struct // replaces buildPushOutputBuffer as a more generic option.
type TermBufferPrinter struct // replaces terminalBuildPushPrinter.
type CIBufferPrinter struct // replaces ciBuildPushPrinter.

func PrintAll(printers []SyncBufferPrinter) error

func (tb *TermBuffer) TermPrinter(fw FileWriter) *TermBufferPrinter
func (tb *TermBuffer) CIPrinter(w io.writer) *CIPrinter
```

efekarakus (commented):
This is awesome!!!

I know it's a lot of comments, but the PR itself is great. I was mostly brainstorming on how we can ease testing and  use well-known interfaces.

Let me know what you think of the suggestions!!

internal/pkg/cli/deploy/workload.go
KollaAdithya: `done`  is used to indicate the completion of copying the build and push output from `pipe reader` to the output buffer.

internal/pkg/cli/deploy/workload_test.go
KollaAdithya: Created a `syncbuffer` package for all the Buffers and `TermPrinter`.

internal/pkg/cli/deploy/workload.go
KollaAdithya: changed the logic to have fixed height of logs.

internal/pkg/cli/deploy/workload.go
```diff
@@ -56,6 +62,12 @@ const (
 	labelForContainerName = "com.aws.copilot.image.container.name"
 )
 
+const (
+	// pollIntervalForBuildPush is the time Interval to wait between checking whether the output buffers are done.
+	pollIntervalForBuildPush = 60 * time.Millisecond
+	maxLogLines              = 5
```
efekarakus, line 68: is this needed?

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+type fileWriter interface {
+	io.Writer
+}
+
+// SyncBuffer is a synchronized buffer that can be used to store output data and coordinate between multiple goroutines.
+type SyncBuffer struct {
+	BufMu sync.Mutex   // BufMu is a mutex that can be used to protect access to the buffer.
+	Buf   bytes.Buffer // Buf is the buffer that stores the data.
+
+	Done chan struct{} // Done is a channel that can be used to signal when the operations are complete.
```
efekarakus, line 32: Can we have all of these fields to be private?

We want to [_encapsulate_](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)) these decisions. We should be able to change for example the type of buffer or whether or not we need a lock in the future.

Instead, if any of these need to be initiliazed consider providing a constructor:
```go
func New()  *SyncBuffer {
   return &SyncBuffer{
      done: make(chan struct{})
   }
}
```

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+// Strings returns an empty slice if the buffer is empty.
+// Otherwise, it returns a slice of all the lines stored in the buffer.
+func (b *SyncBuffer) strings() []string {
+	b.BufMu.Lock()
+	defer b.BufMu.Unlock()
+
+	// Split the buffer bytes into lines.
+	lines := strings.Split(strings.TrimSpace(b.Buf.String()), "\n")
+
+	return lines
```
efekarakus, line 51: nit:
```suggestion
	return strings.Split(strings.TrimSpace(b.Buf.String()), "\n")
```

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+func (b *SyncBuffer) MarkDone() {
+	close(b.Done)
+}
+
+// TermPrinter is a printer to display logs in the terminal.
+type TermPrinter struct {
+	Term             fileWriter
+	Buf              *SyncBuffer
+	PrevWrittenLines int
+	TermWidth        int
```
efekarakus, line 74: samesies here, we should aim to have all of our fields as private

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// NewTermPrinter returns a new instance of TermPrinter that writes logs to the given file writer and reads logs from a new synchronized buffer.
+func NewTermPrinter(fw fileWriter) *TermPrinter {
```
efekarakus, line 78: - Can we expose `fileWriter` since it's an input to the constructor?
- To make the struct flexible, consider taking as input the "number of lines" to print from the buffer. This way we don't have to limit ourselves to 5 
- Should the `termWidth` field be calculated here in the constructor? this way clients don't have to call the `TerminalWidth` method and pass the value down to the object

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+	return &TermPrinter{
+		Term: fw,
+		Buf: &SyncBuffer{
+			Done: make(chan struct{}),
+		},
+	}
+}
+
+// PrintLastFiveLines prints the label and the last five lines of logs to the termPrinter fileWriter.
+func (t *TermPrinter) PrintLastFiveLines() {
```
efekarakus, line 88: 1. If we take the number of lines as input to the constructor, then this becomes:
```suggestion
func (tp *TermPrinter) Print() {
```
    
   Also imo the receiver should have been named "tp" 🧻 😜

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+	logs := t.Buf.strings()
+	outputLogs := t.lastFiveLogLines(logs)
+	if len(outputLogs) > 0 {
```
efekarakus, line 91: nit: reverse the conditional to reduce nesting:
```go
if len(outputLogs) == 0 { return }
```

internal/pkg/term/syncbuffer/syncbuffer.go
efekarakus: 2. Since this is a `TermPrinter` for a `syncbuffer`, consider doing the printing and erasing logic all in this one method to simplify the life of a client.

The overall flow should be something like:
```go
if len(outputLogs) == 0 {
   return
}
tp.eraseLines() // the method only erases lines if numPrevWrittenLines > 0
tp.writeLines()
```

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+		writtenLines := t.numLines(append(outputLogs[:], logs[0]))
+		t.PrevWrittenLines = writtenLines
+	}
+}
+
+// lastFiveLogLines returns the last five lines of the given logs, or all the logs if there are less than five lines.
+func (t *TermPrinter) lastFiveLogLines(logs []string) [maxLogLines]string {
+	// Determine the start and end index to extract last 5 lines
+	start := 1
```
efekarakus, line 104: consider passing `logs[1:]` as input to this method so that we don't have to do this unusual starting position for the `lastNLines`

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+	return logLines
+}
+
+// numLines calculates and returns the actual number of lines needed to print the given string slice based on the terminal width.
+func (t *TermPrinter) numLines(lines []string) int {
+	var numLines float64
+	for _, line := range lines {
+		// Empty line should be considered as a new line
+		if line == "" {
+			numLines = numLines + 1
```
efekarakus, line 127: nit:
```suggestion
			numLines += 1

File: internal/pkg/term/syncbuffer/syncbuffer.go [hunk 29]
```diff
@@ -0,0 +1,149 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package syncbuffer provides a synchronized buffer to store and access logs from multiple goroutines.
```
efekarakus, line 4:
```suggestion
// Package syncbuffer provides a goroutine safe bytes.Buffer as well printing functionality to the terminal.
```

efekarakus (commented):
I've only reviewed the `syncbuffer` but I think that should result in some changes to `deploy` as well :)

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// Write appends the given bytes to the buffer.
+func (b *SyncBuffer) Write(p []byte) (n int, err error) {
+	b.BufMu.Lock()
+	defer b.BufMu.Unlock()
+	return b.Buf.Write(p)
+}
+
+// Strings returns an empty slice if the buffer is empty.
```
Lou1415926, line 42: It seems like in this case `strings.Split` would return a slice of length 1 https://pkg.go.dev/strings#Split 👁️ .

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+		},
+	}
+}
+
+// PrintLastFiveLines prints the label and the last five lines of logs to the termPrinter fileWriter.
+func (t *TermPrinter) PrintLastFiveLines() {
+	logs := t.Buf.strings()
+	outputLogs := t.lastFiveLogLines(logs)
+	if len(outputLogs) > 0 {
+		fmt.Fprintln(t.Term, logs[0])
```
Lou1415926, line 92: oh I might have missed something - why is the first element treated differently here and in `lastFiveLogLines`?

Edit: I see why now - the first element is always the build label is it? Is it possible to make `syncbuffer` agnostic of how the input logs are arranged?

internal/pkg/cli/task_run.go
```diff
@@ -970,7 +970,7 @@ func (o *runTaskOpts) buildAndPushImage() error {
 		Dockerfile: o.dockerfilePath,
 		Context:    ctx,
 		Tags:       append([]string{imageTagLatest}, additionalTags...),
-	}); err != nil {
+	}, os.Stderr); err != nil {
```
Lou1415926, line 973: what about `log.DiagnosticWriter` (the writer that was used by `log.Infoln`)

internal/pkg/cli/deploy/workload.go
```diff
@@ -368,17 +387,73 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+			if err != nil {
+				return fmt.Errorf("build and push image: %w", err)
+			}
+			mu.Lock()
+			defer mu.Unlock()
+			out.ImageDigests[name] = ContainerImageIdentifier{
+				Digest:            digest,
+				CustomTag:         d.image.CustomTag,
+				GitShortCommitTag: d.image.GitShortCommitTag,
+			}
```
Lou1415926, line 434: nit: not a huge difference, but I feel like it's probably safe to unlock the map before `return nil` !
```suggestion
			mu.Lock()
			out.ImageDigests[name] = ContainerImageIdentifier{
				Digest:            digest,
				CustomTag:         d.image.CustomTag,
				GitShortCommitTag: d.image.GitShortCommitTag,
			}
			mu.Unlock()
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -428,6 +503,82 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
...
+			printer.PrintLastFiveLines()
+			totalWrittenLines = totalWrittenLines + printer.PrevWrittenLines
+		}
+		if allDone {
+			break
+		}
+		// sleep for a short time and erase the previous output.
+		time.Sleep(pollIntervalForBuildPush)
+		// erase all the written lines to the terminal.
+		cursor.EraseLinesAbove(os.Stderr, totalWrittenLines)
```
paragbhingre, line 560: Why is it `os.Stderr`? Should it be `os.Stdout`?

internal/pkg/cli/deploy/workload.go
```diff
@@ -368,17 +387,73 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+			return nil
+		})
+
+		g.Go(func() error {
+			return copyOutputToBuffer(pr, termPrinter, name)
+		})
+	}
+
+	// check if CI is true or not.
+	if isCIEnvironment() {
```
Lou1415926, line 444: niceee

internal/pkg/cli/deploy/workload.go
```diff
@@ -428,6 +503,82 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
...
+		}
+		// sleep for a short time and erase the previous output.
+		time.Sleep(pollIntervalForBuildPush)
+		// erase all the written lines to the terminal.
+		cursor.EraseLinesAbove(os.Stderr, totalWrittenLines)
+	}
+	return nil
+}
+
+// PrintAll prints all contents from given TermPrinter buffers once they have finished writing.
```
paragbhingre, line 565:
```suggestion
// printAll prints all contents from given TermPrinter buffers once they have finished writing.
```

Lou1415926 (commented):
This is super exciting!! +1 to Efe's comments. i just have a few additional nits.

internal/pkg/term/syncbuffer/syncbuffer.go
paragbhingre, line 42:
```suggestion
// strings returns an empty slice if the buffer is empty.
```

internal/pkg/term/syncbuffer/syncbuffer.go
KollaAdithya: yes, the first line is always going to be build label. In the `docker` package we write the `docker build` label to the buffer and it is always going to be the first line in the buffer.
If we do not write that label to buffer and just print label in the docker package itself then UI will look like this
```
Building and pushing your container image: docker build -t
Building and pushing your container image: docker build -t
5 lines of image1 .
5 lines of image2.
```
But the Desired UI is:
```
Building and pushing your container image: docker build -t
5 lines of image1 .
Building and pushing your container image: docker build -t
5 lines of image2.
 ```
 is it ok to print the label in the docker package itself ?

Comment (Lou1415926) [reply internal/pkg/term/syncbuffer/syncbuffer.go:92]: I agree that we don't want the first UI. What do you think of this:
```go
type SyncBuffer struct {
  label string
  buf bytes.Buffer
}

func (t *TermPrinter) PrintLastNLines(n int) {
  label := t.Buf.label
  logs := t.Buf.string()
}
```

internal/pkg/term/syncbuffer/syncbuffer.go
KollaAdithya: may be i can have `separator` in the `SyncBuffer` as well. Does this address that particular concern.
```
type SyncBuffer struct {
  buf bytes.Buffer
  sep string
}
```

Lou1415926: Oh I was just saying even if `buf` is empty (i.e. `b.Buf.String() == ""`), it seems like we will return a slice of length 1 (`[""]`). 

> If s does not contain sep and sep is not empty, Split returns a slice of length 1 whose only element is s. 
(from https://pkg.go.dev/strings#Split)



We should probably either do
```go
content  := b.Buf.String()
if len(content) == 0 {return nil}
return strings.Split(strings.TrimSpace(b.Buf.String()), "\n")
```

or update this comment so that it is accurate.

internal/pkg/cli/deploy/workload.go
KollaAdithya: If one of the docker build fails, the error messages should be printed to `os.Stderr`.

paragbhingre: But aren't we erasing lines that we already printed? Then why do we need `os.Stderr` here? I don't see where we are printing err message to `os.Stderr`

KollaAdithya: I mean we are writing to [log.DiagonsticWriter](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/term/log/log.go#L16) which it self is a `color supporting writer for os.Stderr.`

PR #4722, opened 2023-03-31
chore: upload static site files to cache

dannyrandall:
This PR makes the static site files upload to the app bucket with the key `/local-assets/<file hash>`. It also uploads a mapping JSON file mapping each hashed file to a destination location in the static site bucket. It then passes that location of the mapping file to the template package. In a future PR, the mapping file will passed to the template package to create a step machine+lambda that will copy the cached files over to a new bucket for the static site.


internal/pkg/cli/deploy/static_site.go
```diff
@@ -87,20 +94,23 @@ func (d *staticSiteDeployer) DeployWorkload(in *DeployWorkloadInput) (ActionReco
...
-				},
-			}); err != nil {
-			return nil, err
+		assets, err := d.uploader.UploadToCache(filepath.Join(f.Context, f.Source), f.Destination, &asset.UploadOpts{
+			Reincludes: f.Reinclude.ToStringSlice(),
+			Excludes:   f.Exclude.ToStringSlice(),
+			Recursive:  f.Recursive,
+		})
+		if err != nil {
+			return err
```
iamhopaul123: Do we need to wrap this error?

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -5,61 +5,84 @@
...
 
-func walkFn(source, dest string, recursive bool, reader afero.Fs, upload UploadFunc, urlsPtr *[]string, matcher filepathMatcher) filepath.WalkFunc {
+func (u *Uploader) walkFn(sourcePath, destPath string, recursive bool, matcher filepathMatcher, assets *[]Cached) filepath.WalkFunc {
 	return func(path string, info fs.FileInfo, err error) error {
 		if err != nil {
 			return err
 		}
 		if info.IsDir() {
-			if !recursive {
+			if !recursive && path != sourcePath {
```
iamhopaul123, line 80: why do we make this change?

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -71,25 +94,61 @@ func walkFn(source, dest string, recursive bool, reader afero.Fs, upload UploadF
...
-			return fmt.Errorf("open file on path %q: %w", path, err)
+			return fmt.Errorf("open %q: %w", path, err)
 		}
 		defer file.Close()
-		fileRel, err := filepath.Rel(source, path)
+
+		_, err = io.Copy(io.MultiWriter(buf, hash), file)
 		if err != nil {
-			return fmt.Errorf("get relative path for %q against %q: %w", path, source, err)
+			return fmt.Errorf("copy %q: %w", path, err)
```
iamhopaul123, line 103: should we use `file`?

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -71,25 +94,61 @@ func walkFn(source, dest string, recursive bool, reader afero.Fs, upload UploadF
...
+
+		if sourcePath == path && destPath == "" {
+			asset.DestinationPath = info.Name()
+		} else if sourcePath != path {
+			rel, err := filepath.Rel(sourcePath, path)
+			if err != nil {
+				return fmt.Errorf("get relative path for %q against %q: %w", path, sourcePath, err)
+			}
+
+			asset.DestinationPath = filepath.Join(destPath, rel)
```
iamhopaul123: Is this because we can get rid of https://github.com/aws/copilot-cli/pull/4722/files#diff-befc0e16cf8c6a7ad74caee13773fd4b095ba8bea5170dd1e39f6e3384e0667fL35-L46 ?

iamhopaul123, line 127: Is this because we can get rid of https://github.com/aws/copilot-cli/pull/4722/files#diff-befc0e16cf8c6a7ad74caee13773fd4b095ba8bea5170dd1e39f6e3384e0667fL35-L46 ?

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -71,25 +94,61 @@ func walkFn(source, dest string, recursive bool, reader afero.Fs, upload UploadF
...
 		}
-		*urlsPtr = append(*urlsPtr, url)
+
+		*assets = append(*assets, asset)
 		return nil
 	}
 }
 
+func (u *Uploader) uploadAssets(assets []Cached) error {
```
iamhopaul123: Awesome! Just a note here we might want to do some validation on the size of the file (we could do it afterwards)

iamhopaul123 (commented):
Nice! LGTM overall

internal/pkg/cli/deploy/static_site.go
dannyrandall: i think the most useful context is in the errors already - current errors are:
- "walk the file tree rooted at <source directory from `f`>"
- "upload \<specific file path\>"

we _could_ prefix those errors with the index of the upload if you think that's nice? something like:
"upload #1/4: upload \<specific file path\>: some error" 

thoughts?

internal/pkg/deploy/upload/asset/asset.go
dannyrandall: since i removed the pre-processing of directories [here](https://github.com/aws/copilot-cli/pull/4722/files#diff-befc0e16cf8c6a7ad74caee13773fd4b095ba8bea5170dd1e39f6e3384e0667fL37-L44), we need to go through all paths in a directory if it is the directory provided.

example: for source `./files`
on first call to the func returned by `walkFn()`, path == sourcePath == "./files". we don't want to skip that directory, because it's the actual directory that was passed in. however, we still want to skip `./files/subdir/` if `recursive` is false.

internal/pkg/cli/deploy/static_site.go
iamhopaul123: yeah that sounds good to me

dannyrandall: sg, i'll make the change!

internal/pkg/deploy/upload/asset/asset.go
iamhopaul123: ah i see. ok then why do we remove the pre-processing of directories 🤔 i feel like it was easier to follow

dannyrandall: it simplifies figuring out what the destination location should be in `walkFn` - previously, files configured were passed in to `walkFn` the exact same way as directory configurations, so it was impossible (difficult?) to figure out if `sourcePath` of `./files/index.html` meant dest should be `index.html` in the case of the configuration being a file, or dest should be `/files/index.html`, in the case of the configuration being `./files`. Does that make any sense?

internal/pkg/deploy/upload/asset/asset.go
dannyrandall: wdym? to get the path? `file` only has `Name()` 😔

iamhopaul123: yeah but the error comes from `_, err = io.Copy(io.MultiWriter(buf, hash), file)` so I would assume `file` should be included in the error msg? Or concatenate them

dannyrandall: oh i see. file is a `afero.File`, so kinda hard to include in the error message 🤔

internal/pkg/deploy/upload/asset/asset.go
dannyrandall: the relative path was the same previously, but yeah, with this we can get rid of the other section

internal/pkg/deploy/upload/asset/asset_test.go
```diff
@@ -4,35 +4,47 @@
...
-			return "", f.err
-		}
-		url, ok := f.objects[key]
-		if !ok {
-			return "", fmt.Errorf("key %q does not exist in fakeS3", key)
+func (f *fakeS3) Upload(bucket, key string, data io.Reader) (string, error) {
+	return "", f.err
+}
+
+func Test_UploadToCache(t *testing.T) {
```
iamhopaul123: shall we make more test cases when 1) source is a dir and dest is not set 2) source is a file and dest is not set etc.

dannyrandall: labelled the existing test cases and added a test for the missing case! (source=file, dest set)

internal/pkg/cli/deploy/static_site.go
```diff
@@ -20,12 +20,15 @@ import (
 	"github.com/spf13/afero"
 )
 
+type cacheUploader interface {
```
efekarakus: nit: should we rename this interface?
```suggestion
type fileUploader interface {
```

internal/pkg/cli/deploy/static_site.go
```diff
@@ -39,12 +42,19 @@ func NewStaticSiteDeployer(in *WorkloadDeployerInput) (*staticSiteDeployer, erro
...
 		svcDeployer:   svcDeployer,
 		staticSiteMft: mft,
-		fs:            afero.NewOsFs(),
-		bucketName:    svcDeployer.resources.S3Bucket,
-		uploadFn:      asset.Upload,
+		fs:            fs,
+		uploader: &asset.CacheUploader{
+			FS:               fs,
+			PathPrefix:       fmt.Sprintf("static-site-cache/%s", svcDeployer.name),
+			AssetMappingPath: fmt.Sprintf("static-site-cache/%s/mapping/%s.json", svcDeployer.name, time.Now().Format(time.RFC3339)),
```
efekarakus: Can you remind me how these assets will be deleted? Is it through a TTL or a different mechanism?

I'm wondering if there is value at all to have a more generic path to re-use assets across workloads, for example:
```go
PathPrefix:       fmt.Sprintf("local-assets", svcDeployer.name),
AssetMappingPath: fmt.Sprintf("local-assets/workloads/%s/mapping/%s.json", svcDeployer.name, time.Now().Format(time.RFC3339)),
```

internal/pkg/cli/deploy/static_site.go
```diff
@@ -87,20 +97,17 @@ func (d *staticSiteDeployer) DeployWorkload(in *DeployWorkloadInput) (ActionReco
...
-			}); err != nil {
-			return nil, err
-		}
+	return d.uploadArtifacts(d.uploadStaticFiles, d.uploadArtifactsToS3, d.uploadCustomResources)
+}
+
+func (d *staticSiteDeployer) uploadStaticFiles(out *UploadArtifactsOutput) error {
+	url, err := d.uploader.UploadFiles(d.staticSiteMft.FileUploads)
+	if err != nil {
+		return err
```
efekarakus: nit: do we need to wrap this err?

internal/pkg/cli/deploy/workload.go
```diff
@@ -110,12 +110,13 @@ type fileReader interface {
 
 // StackRuntimeConfiguration contains runtime configuration for a workload CloudFormation stack.
 type StackRuntimeConfiguration struct {
```
efekarakus, line 108: nit: is there a way of separating these fields between "required" and "optional"? Like which one of these are guaranteed to be populated vs. depending on the circumstance.

For example,
```go
// Required data that is populated on every package.
ImageDigests
CustomResourceURLs

// Optional data populated depending on customer's input.
EnvFileARNs..
```

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -5,61 +5,84 @@
...
 
+	"github.com/aws/copilot-cli/internal/pkg/manifest"
 	"github.com/spf13/afero"
+	"golang.org/x/sync/errgroup"
 )
 
-// UploadFunc is the function signature to upload contents to a destination.
-type UploadFunc func(dest string, contents io.Reader) (url string, err error)
+// CacheUploader uploads local asset files.
+type CacheUploader struct {
```
efekarakus: What do you think of renaming to `ArtifactBucketUploader` to highlight the fact that this uploader is strictly to put assets to our artifact bucket (vs. maybe a customer's existing bucket)

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -5,61 +5,84 @@
...
+	FS afero.Fs
 
-// UploadOpts contains optional configuration for uploading assets.
-type UploadOpts struct {
-	Reincludes []string   // Relative path under source to reinclude files that are excluded in the upload.
-	Excludes   []string   // Relative path under source to exclude in the upload.
-	UploadFn   UploadFunc // Custom implementation on how to upload the contents under a file. Defaults to S3UploadFn.
-	Recursive  bool       // Whether to walk recursively.
+	// Upload is the function called when uploading a file.
+	Upload func(path string, contents io.Reader) (string, error)
```
efekarakus: nit: I get why we do this, it's just a bit strange to see a function as input vs. an interface. Perhaps, not in this PR but what do you think of the following suggestion 👇 

Is there a refactoring opportunity in the `upload/customresource` and `upload/asset` packages. Perhaps, something like:
```go
package upload

type Uploader interface {
   Upload(path string, contents io.Reader) (string, error)
}

type BucketUploader interface {
   Upload(bucket, path, data io.Reader) (string, error)
}

func ConvertS3(s3 BucketUploader) Uploader
```

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -5,61 +5,84 @@
...
-		}
-		paths = make([]string, len(files))
-		for i, f := range files {
-			paths[i] = filepath.Join(source, f.Name())
+type cached struct {
+	localPath string
+	content   io.Reader
+
+	Path            string `json:"path"`
+	DestinationPath string `json:"dest_path"`
```
efekarakus: nit: consider using camelCase (https://stackoverflow.com/a/19287394/1201381)
```suggestion
	DestinationPath string `json:"destPath"`
```

efekarakus: What's the difference between these two fields?

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -5,61 +5,84 @@
...
+}
+
+// UploadFiles hashes each of the files specified in files and uploads
+// them to the path "{CachePathPrefix}/{hash}". After, it uploads a JSON file
+// to CacheMovePath that specifies the uploaded location of every file and it's
```
efekarakus:
```suggestion
// them to the path "PathPrefix/{hash}". After, it uploads a JSON file
// to AssetMappingPath that specifies the uploaded location of every file and it's
```

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -5,61 +5,84 @@
...
-	paths := []string{source}
-	if info.IsDir() {
-		files, err := afero.ReadDir(fs, source)
-		if err != nil {
-			return nil, fmt.Errorf("read directory %q: %w", source, err)
-		}
-		paths = make([]string, len(files))
-		for i, f := range files {
-			paths[i] = filepath.Join(source, f.Name())
+type cached struct {
```
efekarakus:
```suggestion
type asset struct {
```

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -5,61 +5,84 @@
...
 	}
-	return urls, nil
+
+	// upload move file
+	url, err := u.Upload(u.AssetMappingPath, mappingFile)
+	if err != nil {
+		return "", fmt.Errorf("unable to upload cache move: %w", err)
+	}
+
+	return url, nil
```
efekarakus: does the URL need to be returned or can we get it from either of these:
https://github.com/aws/copilot-cli/blob/d85cbbe5e36a7810bbe4342e2a121692719ac5bd/internal/pkg/aws/s3/s3.go#L158-L171

efekarakus (commented):
Looks good! 

A question for @iamhopaul123 and @dannyrandall:
- Is there a particular reason why we don't just point CloudFront to our artifact bucket instead of creating a new static site bucket? I don't recall the reason. Is it recommended to always create a separate S3 bucket

internal/pkg/deploy/upload/asset/asset.go
dannyrandall: i think this makes a lot of sense - i agree the function is less than ideal. works out anyways bc i'm planning on adding an `UploadContext()` so the cancellation works in my `errgroup`! (future PR though 😌)

internal/pkg/deploy/upload/asset/asset.go
dannyrandall: `Path` is the path of the object in the artifact bucket, `DestinationPath` is the eventual path of the object in the maybe-not-created-yet service bucket.

dannyrandall: i'll rename to clarify :)

internal/pkg/cli/deploy/static_site.go
dannyrandall: TTL! And I think that makes sense. Hopefully can avoid some uploads if files are the same.

internal/pkg/cli/deploy/workload.go
dannyrandall: i'm not _opposed_ to doing this, however, i think the only field that's actually required across all workloads is `CustomResourceURLs`...

internal/pkg/deploy/upload/asset/matcher.go
Lou1415926, line 1: are these copy-pasted from the old files or are there any changes that we need to pay attention to?

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -5,61 +5,77 @@
...
 
-func walkFn(source, dest string, recursive bool, reader afero.Fs, upload UploadFunc, urlsPtr *[]string, matcher filepathMatcher) filepath.WalkFunc {
+func (u *ArtifactBucketUploader) walkFn(sourcePath, destPath string, recursive bool, matcher filepathMatcher, assets *[]asset) filepath.WalkFunc {
 	return func(path string, info fs.FileInfo, err error) error {
 		if err != nil {
 			return err
 		}
 		if info.IsDir() {
-			if !recursive {
+			if !recursive && path != sourcePath {
```
Lou1415926, line 80: I was also stopped by this line of change and was confused a little. Although after reading the conversation between you & ph it all made sense. I wonder if there is a better way to communicate what you've said here in the code 🤔 (a comment, maybe?)

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -71,98 +87,73 @@ func walkFn(source, dest string, recursive bool, reader afero.Fs, upload UploadF
...
-}
 
-type filepathMatcher interface {
-	match(path string) (bool, error)
-}
-
-type reincludeMatcher string
+		dest := filepath.Join(destPath, rel)
+		if dest == "." { // happens when sourcePath is a file and destPath is unset
+			dest = info.Name()
```
Lou1415926, line 114: would this logic make sense for all kinds of `assets` (for example if we support lambdas in the future, then the `assets` might be the JS codes), or just the static site assets?

internal/pkg/deploy/upload/asset/matcher.go
dannyrandall: this one is copy pasted!

internal/pkg/deploy/upload/asset/asset.go
dannyrandall: lambdas are mostly different because they require a single zip file - if we were to support a lambda pattern, i would think it would be more as a containerized lambda, not directly uploading assets to s3 🤔 (something to think about in the future!!!)

internal/pkg/deploy/upload/asset/asset.go
dannyrandall: i added a clarification comment, hopefully that helps 🥺

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -5,61 +5,79 @@
...
+	localPath string
+	content   io.Reader
+
+	ArtifactBucketPath string `json:"path"`
+	ServiceBucketPath  string `json:"destPath"`
+}
+
+// UploadFiles hashes each of the files specified in files and uploads
+// them to the path "{PathPrefix}/{hash}". After, it uploads a JSON file
+// to AssetMappingPath that specifies the location of every file in the artifact bucket and it's
```
huanjani: nit:
```suggestion
// to AssetMappingPath that specifies the location of every file in the artifact bucket and its
```

internal/pkg/deploy/upload/asset/asset.go
huanjani: fwiw, PH and I decided that all dirs should be recursive by default. We couldn't think of any use case for a non-recursive source dir (and can add the option later if necessary).

internal/pkg/deploy/upload/asset/matcher.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+		if isMatch {
+			shouldInclude = true
+		}
+	}
+	return shouldInclude, nil
+}
+
+func match(pattern, path string) (bool, error) {
+	isMatch, err := filepath.Match(pattern, path)
```
huanjani, line 81: 🥴 Does [filepath.Match](https://pkg.go.dev/path/filepath#Match) support the "in sequence" and "not in sequence" patterns? https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3/index.html#use-of-exclude-and-include-filters

dannyrandall: it doesn't look like it supports the "not in sequence" pattern!

internal/pkg/cli/deploy/static_site.go
```diff
@@ -39,12 +43,22 @@ func NewStaticSiteDeployer(in *WorkloadDeployerInput) (*staticSiteDeployer, erro
 	if !ok {
 		return nil, fmt.Errorf("manifest is not of type %s", manifestinfo.StaticSiteType)
 	}
+	fs := afero.NewOsFs()
+	assetMappingPath := fmt.Sprintf("local-assets/workloads/%s/mapping/%s.json", svcDeployer.name, time.Now().Format(time.RFC3339))
```
iamhopaul123: Should we include the env name in the full path in case we want to get the last upload info in the future?

internal/pkg/cli/deploy/static_site.go
```diff
@@ -39,12 +43,22 @@ func NewStaticSiteDeployer(in *WorkloadDeployerInput) (*staticSiteDeployer, erro
 	if !ok {
 		return nil, fmt.Errorf("manifest is not of type %s", manifestinfo.StaticSiteType)
 	}
+	fs := afero.NewOsFs()
```
iamhopaul123: should we init this in cli pkg and move it to the `WorkloadDeployerInput` 🤔

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -71,98 +89,73 @@ func walkFn(source, dest string, recursive bool, reader afero.Fs, upload UploadF
...
-		}
-		if isMatch {
-			shouldInclude = false
-		}
-	}
-	for _, matcher := range m.reincludeMatchers {
-		isMatch, err := matcher.match(path)
-		if err != nil {
-			return false, err
+func (u *ArtifactBucketUploader) uploadAssetMappingFile(assets []asset) error {
```
iamhopaul123, line 151: maybe a simple example as a comment about what a mapping file looks like?

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -71,98 +89,73 @@ func walkFn(source, dest string, recursive bool, reader afero.Fs, upload UploadF
...
-}
+		return assets[i].ServiceBucketPath < assets[j].ServiceBucketPath
+	})
 
-func match(pattern, path string) (bool, error) {
-	isMatch, err := filepath.Match(pattern, path)
+	data, err := json.Marshal(assets)
 	if err != nil {
-		return false, fmt.Errorf("match file path %s against pattern %s: %w", path, pattern, err)
+		return fmt.Errorf("encode uploaded assets: %w", err)
```
iamhopaul123, line 162:
```suggestion
		return fmt.Errorf("encode staged assets: %w", err)
```
would this be more accurate? Same for the similar situation otherwise users would think their file has already been "uploaded".

internal/pkg/deploy/upload/asset/matcher.go
```diff
@@ -0,0 +1,86 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
iamhopaul123, line 1: are we just moving all the match-related logic to a new file?

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -71,98 +89,73 @@ func walkFn(source, dest string, recursive bool, reader afero.Fs, upload UploadF
...
+	g, _ := errgroup.WithContext(context.Background())
 
-func buildExcludeMatchers(excludes []string) []filepathMatcher {
-	var matchers []filepathMatcher
-	for _, exclude := range excludes {
-		matchers = append(matchers, excludeMatcher(exclude))
+	for i := range assets {
+		asset := assets[i]
+		g.Go(func() error {
+			if err := u.Upload(asset.ArtifactBucketPath, asset.content); err != nil {
```
efekarakus, line 133: problem for another day: I wonder how we can show progress bars while uploading these assets to s3

efekarakus (approved):
LGTM!

internal/pkg/deploy/upload/asset/matcher.go
dannyrandall: yeah, no changes 😊

Lou1415926: poor danny had to answer the same question twice hahahahaha

dannyrandall: i probably should have mentioned that somewhere lol

internal/pkg/deploy/upload/asset/asset.go
dannyrandall: actually at this point their files have been uploaded to the cache bucket

iamhopaul123 (approved):

PR #2576, opened 2021-07-08
feat: add grace_period and deregistration_delay to http for lb web services.

bvtujo:
Implements two new fields in the manifest under `http`: `grace_period` and `deregistration_delay`. These allow customization of the corresponding fields in the target group health check and service parameters. 

Resolves #2190, resolves #2506 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4734, opened 2023-04-05
fix: separate out the grace_period for ALB and NLB

paragbhingre:
Previously, ALB and NLB's grace period was set from http's `grace_period` manifest field. This PR separate out the field such that ALB and NLB have their own fields. 
For ALB, `grace_period` field was under `http.healthcheck` but grace_period is a service level field, so it should be under `http` and not `http.healthcheck`. So we have fixed that as well. 


internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -669,6 +678,18 @@ func (s *LoadBalancedWebService) convertNetworkLoadBalancer() (networkLoadBalanc
 	return config, nil
 }
 
+func (s *LoadBalancedWebService) convertGracePeriod() *int64 {
+	var gracePeriod *int64 = aws.Int64(60)
+	if s.manifest.HTTPOrBool.GracePeriod != nil && s.manifest.NLBConfig.GracePeriod != nil { // if grace period is mentioned in both ALB and NLB then we prefer the one with max value.
```
KollaAdithya: nit: can we use a switch case ?

internal/pkg/manifest/http.go
```diff
@@ -55,9 +49,32 @@ func (r *HTTPOrBool) UnmarshalYAML(value *yaml.Node) error {
 
 // HTTP holds options for application load balancer.
 type HTTP struct {
-	Main                     RoutingRule   `yaml:",inline"`
-	TargetContainerCamelCase *string       `yaml:"targetContainer"` // Deprecated. Maintained for backwards compatibility, use [RoutingRule.TargetContainer] instead.
-	AdditionalRoutingRules   []RoutingRule `yaml:"additional_rules"`
+	Main                     RoutingRule    `yaml:",inline"`
+	TargetContainerCamelCase *string        `yaml:"targetContainer"` // Deprecated. Maintained for backwards compatibility, use [RoutingRule.TargetContainer] instead.
+	GracePeriod              *time.Duration `yaml:"grace_period"`
```
Lou1415926: > For ALB, grace_period field was under http.healthcheck but grace_period is a service level field, so it should be under http and not http.healthcheck. 

Just in case this wasn't discussed before this PR. I recall there was a discussion around this, when `grace_period` was first introduced in https://github.com/aws/copilot-cli/pull/2576: `grace_period` is a health-check related configuration, therefore we put in under `http.health_check`. Although in ECS it is configured as a service attribute, but we tried to model our manifest "in the way that developers think", instead of "in the way how ECS models it".

paragbhingre: I agree with your point, but it is becoming super weird with our current structure if we keep it inside healthcheck. 
If we keep it inside healthcheck then:
1. Either it has to be under both the main as well as additional routing rules. But it is not a routing rule parameter, so having this parameter in the main or additional rules sounds wrong to me. 
2. Or it has to be under the only Main routing rule, which is also something that is less descriptive of the purpose of that parameter. 

Do let me know your thoughts on this.

Lou1415926: If multiple containers are handling requests routed from an ELB, I'd assume customers would want to set the grace period to accommodate for the container that is the slowest to start 💭 

An alternative that we can consider is to get the max of the grace periods for individual rules.

```yaml
http:
  path: 'api'
  healthcheck:
     grace_period: 1m
  additional_rules:
      - path: 'user'
        healthcheck:
            path: 'user' 
      - path: 'admin'
        healthcheck:
            path: 'admin'
            grace_period: 10m # The grace period is 10m because 10m > 1m.
```

I see a similar logic in this PR where you get the max of `nlb` and `http`'s `grace_periods`.

paragbhingre: Yeah that could be another solution, but then we choose max of the grace periods for individual rules. And at the same time if `nlb`'s `grace_period` is also specified then shall we take max of `nlb` and `http` ?

Edit - that makes me wonder if nlb listeners should also have this `grace_period` parameter?

Lou1415926: I'd imagine we'd want to take the max, if we were to be consistent in the design logic of the "alternative approach".

The pros of the original approach that you purposed by this PR is that it fits nicer with ECS's model. The cons is that it might be too specifically designed for ECS, and misses the abstraction that we've tried to achieve.

For the alternative approach, the pros & cons are the opposite.

I'd also love to hear from the others on this 👍🏼

paragbhingre: As per the discussion with @iamhopaul123 and @efekarakus, here is the summary of what we are going to do:
1. We agree that [healthcheckgraceperiodseconds](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-service.html#cfn-ecs-service-healthcheckgraceperiodseconds) should have been load balancer level setting in ECS. But until they change it, we will keep `grace_period` as is for Main as well as Additional Listener Rules in ALB. That means we will have `grace_period` on `http.Main.HealthCheck` level. So that in the future, whenever ECS changes it, we will open up `grace_period` for additional listener rules as well. For now, we will hide this setting from Additional Rules in our Copilot documents. 
2. In NLB, `grace_period` will be introduced on `nlb.Listener.HealthCheck` level. `grace_period` from additional listeners will be hidden and not shown in our docs. 
3. When ALB and NLB both have `grace_period` set, we will throw a specific error saying they can set either property but not both. Same goes with listener rules and listeners as well. If customers accidentally set `grace_period` on Main as well as Additional Listener Rules, then we error out. 

Do let me know if there are any concerns around this.

internal/pkg/manifest/errors.go
```diff
@@ -98,6 +98,21 @@ func (e *errFieldMutualExclusive) Error() string {
 	return fmt.Sprintf(`must specify one, not both, of "%s" and "%s"`, e.firstField, e.secondField)
 }
 
+type errGracePeriodAlreadyExist struct {
```
Lou1415926:
```suggestion
type errGracePeriodSpecifiedMoreThanOnce struct {
```

internal/pkg/manifest/validate.go
```diff
@@ -800,6 +803,63 @@ func (r HTTPOrBool) validate() error {
...
+			firstField:  "http.healthcheck.grace_period",
+			secondField: "nlb.healthcheck.grace_period",
+		}
+	}
+
+	return nil
+}
+
+// validateGracePeriodForALB validates if ALB or NLB has duplicate grace period mentioned in their additional listeners rules or listeners respectively.
+func (l LoadBalancedWebServiceConfig) validateGracePeriodForALB(alb bool) (bool, error) {
```
Lou1415926: Can we have `validateGracePeriodForALB()` and `validateGracePeriodForNLB()` separately ?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -550,6 +554,14 @@ func (s *BackendService) convertALBListener() (*template.ALBListener, error) {
 	}, nil
 }
 
+func (s *BackendService) convertGracePeriod() *int64 {
+	var gracePeriod *int64 = aws.Int64(60)
```
Lou1415926: We should probably reuse `manifest.DefaultGracePeriod`: https://www.danielmsullivan.com/pages/tutorial_workflow_3bestpractice.html#no-magic-numbers

internal/pkg/manifest/validate.go
Lou1415926: In addition, if I'm understanding https://github.com/aws/copilot-cli/pull/4734#discussion_r1160285656 correctly, we allow `grace_period` in the main rule only. I thought this means that if we see `grace_period` in `additional_rule[x]`, we will error out no matter what 🤔

paragbhingre: That could be another way to error out. But yeah not letting them to specify that param at the first place itself makes more sense. I have changed it in my latest commit. Do let me know how do you feel about new errors.

internal/pkg/deploy/cloudformation/stack/transformers.go
iamhopaul123: unresolve the comment as it has not been resolved yet. Feel free to resolve after pushing the commit.

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -669,6 +681,16 @@ func (s *LoadBalancedWebService) convertNetworkLoadBalancer() (networkLoadBalanc
 	return config, nil
 }
 
+func (s *LoadBalancedWebService) convertGracePeriod() *int64 {
+	var gracePeriod *int64 = aws.Int64(int64(manifest.DefaultHealthCheckGracePeriod))
+	if s.manifest.HTTPOrBool.Main.HealthCheck.Advanced.GracePeriod != nil {
+		gracePeriod = aws.Int64(int64(s.manifest.HTTPOrBool.Main.HealthCheck.Advanced.GracePeriod.Seconds()))
+	} else if s.manifest.NLBConfig.Listener.HealthCheck.GracePeriod != nil {
```
iamhopaul123: i assume we can't set both alb and nlb grace period at the same time. Maybe `if` is enough?

internal/pkg/manifest/errors.go
```diff
@@ -98,6 +98,37 @@ func (e *errFieldMutualExclusive) Error() string {
...
+type errGracePeriodSpecifiedMoreThanOnce struct {
+	firstField  string
+	secondField string
+	index       int
+}
+
+func (e *errGracePeriodSpecifiedMoreThanOnce) Error() string {
+	if strings.Contains(e.firstField, "additional_rules") {
+		return fmt.Sprintf(`cannot define "grace_period" in "http.additional_rules[%d].healthcheck.grace_period"`, e.index)
+	} else if strings.Contains(e.firstField, "additional_listeners") {
```
iamhopaul123: This is very confusing because it has nothing to do with the name of  `errGracePeriodSpecifiedMoreThanOnce`. I feel like we need a different error for this case. can we use `errFieldMutualExclusive` if they specify `grace_period` in both places and a new error type if they specify `grace_period` in additional rules?

Looking forward the key is really about "decoupling" (so that we can make changes in one place; easier to maintain; and less error-prone): we should avoid mixing things together just because they belong to the same "topic" but focusing on what they do (functionalities). For example, in this errors file `errContainersExposingSamePort` is a probably ok because it is specific but we have to do it this way (the error message is specifically about port). Whereas `errSpecifiedBothIngressFields` is a bad one because the error msg has nothing to do with ingress fields and we should've reused `errFieldMutualExclusive` (even if we need new recommendation we could make it based on `errFieldMutualExclusive`)

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: oops I made the change for LBWS and not for BS. I will do it for BS as well.

internal/pkg/deploy/cloudformation/stack/transformers.go
paragbhingre: So we either want to set grace period that is given in ALB or NLB or default. So if we just use `if` then we wouldn't be able to send default value if both ALB and NLB doesn't have `grace_period` specified. Does that make sense?

efekarakus: nit:  consider switching the layout to guard-clauses

```go
if s.manifest.HTTPOrBool.Main.HealthCheck.Advanced.GracePeriod != nil {
   return aws.Int64(int64(s.manifest.HTTPOrBool.Main.HealthCheck.Advanced.GracePeriod.Seconds()))
}
if s.manifest.NLBConfig.Listener.HealthCheck.GracePeriod != nil {
   return aws.Int64(int64(s.manifest.NLBConfig.Listener.HealthCheck.GracePeriod.Seconds()))
}
return aws.Int64(int64(manifest.DefaultHealthCheckGracePeriod))
```

internal/pkg/manifest/validate.go
```diff
@@ -800,6 +803,59 @@ func (r HTTPOrBool) validate() error {
 	return r.HTTP.validate()
 }
 
+func (l LoadBalancedWebServiceConfig) validateGracePeriod() error {
```
efekarakus, line 806: nit: consider switching the receivers to a slightly more explicit name:
```suggestion
func (cfg *LoadBalancedWebServiceConfig) validateGracePeriod() error {
```

efekarakus (commented):
thanks Parag! Looks good, not much else to add besides the existing comments.

internal/pkg/manifest/errors.go
paragbhingre: yeah that sounds good to me. I have changed error message when customer define `grace_period` in both ALB and NLB to errFieldMutualExclusive. And renamed existing `errGracePeriodSpecifiedMoreThanOnce` to `errGracePeriodSpecifiedInAdditionalField` when customer define `grace_period` in additional rules or listeners. Do let me know how does that look to you?

internal/pkg/manifest/errors.go
```diff
@@ -98,6 +98,42 @@ func (e *errFieldMutualExclusive) Error() string {
...
+	return ""
+}
+
+type errGracePeriodSpecifiedInAdditionalField struct {
+	field string
+	index int
+}
+
+func (e *errGracePeriodSpecifiedInAdditionalField) Error() string {
+	if strings.Contains(e.field, "additional_rules") {
```
Lou1415926: This is too coupled with the field name. I suggest we do the ⬇️ :
```go
type errGracePeriodSpecifiedInAdditionalRule struct {}
type errGracePeriodSpecifiedInAdditionalListener struct {}
```

internal/pkg/manifest/errors.go
```diff
@@ -98,6 +98,42 @@ func (e *errFieldMutualExclusive) Error() string {
...
+	return fmt.Sprintf(`cannot define "grace_period" in "nlb.additional_listeners[%d].healthcheck.grace_period"`, e.index)
+}
+
+// RecommendActions returns recommended actions to be taken after the error.
+func (e *errGracePeriodSpecifiedInAdditionalField) RecommendActions() string {
+
+	if strings.Contains(e.field, "additional_rules") {
+		return fmt.Sprintf(`It looks like you have specified "grace_period" under "http.additional_rules[%d].healthcheck". Specify "grace_period" under "http.healthcheck".
+ECS allows "HealthCheckGracePeriodSeconds" as a service level setting and should be specified in the main listener rule only. 
+It is the period of time, in seconds, that the Amazon ECS service scheduler ignores unhealthy Elastic Load Balancing target health checks after a task has first started.`, e.index)
```
Lou1415926: Explaining what `grace_periods` is seems unnecessary 🤔 

We should aim to log only the necessary text on terminal, otherwise the real important message will just get buried in the sea of logs, which negatively affect the user experience.

internal/pkg/manifest/errors.go
```diff
@@ -98,6 +98,42 @@ func (e *errFieldMutualExclusive) Error() string {
...
+	}
+	return fmt.Sprintf(`cannot define "grace_period" in "nlb.additional_listeners[%d].healthcheck.grace_period"`, e.index)
+}
+
+// RecommendActions returns recommended actions to be taken after the error.
+func (e *errGracePeriodSpecifiedInAdditionalField) RecommendActions() string {
+
+	if strings.Contains(e.field, "additional_rules") {
+		return fmt.Sprintf(`It looks like you have specified "grace_period" under "http.additional_rules[%d].healthcheck". Specify "grace_period" under "http.healthcheck".
+ECS allows "HealthCheckGracePeriodSeconds" as a service level setting and should be specified in the main listener rule only.
```
Lou1415926:
```suggestion
		return fmt.Sprintf(`Instead of under "http.additional_rules[%d].healthcheck", specify "grace_period" under the top-level "http.healthcheck".`)
```

internal/pkg/manifest/errors.go
```diff
@@ -98,6 +98,42 @@ func (e *errFieldMutualExclusive) Error() string {
 	return fmt.Sprintf(`must specify one, not both, of "%s" and "%s"`, e.firstField, e.secondField)
 }
 
+// RecommendActions returns recommended actions to be taken after the error.
+func (e *errFieldMutualExclusive) RecommendActions() string {
+	if strings.Contains(e.firstField, "http.healthcheck.grace_period") && strings.Contains(e.secondField, "nlb.healthcheck.grace_period") {
```
Lou1415926: This is too coupled with the field names as well. Consider doing this:
```go
type errGracePeriodsInBothALBAndNLB struct {
  errFieldMutualExclusive
}

func (e *errGracePeriodsInBothALBAndNLB) RecommendedActions() string {}
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -550,6 +554,14 @@ func (s *BackendService) convertALBListener() (*template.ALBListener, error) {
 	}, nil
 }
 
+func (s *BackendService) convertGracePeriod() *int64 {
+	var gracePeriod *int64 = aws.Int64(int64(manifest.DefaultHealthCheckGracePeriod))
+	if s.manifest.HTTP.Main.HealthCheck.Advanced.GracePeriod != nil {
+		gracePeriod = aws.Int64(int64(s.manifest.HTTP.Main.HealthCheck.Advanced.GracePeriod.Seconds()))
+	}
+	return gracePeriod
```
```suggestion
	if s.manifest.HTTP.Main.HealthCheck.Advanced.GracePeriod != nil {
		return aws.Int64(int64(s.manifest.HTTP.Main.HealthCheck.Advanced.GracePeriod.Seconds()))
	}
	return aws.Int64(int64(manifest.DefaultHealthCheckGracePeriod))
```

internal/pkg/manifest/errors.go
```diff
@@ -98,6 +98,42 @@ func (e *errFieldMutualExclusive) Error() string {
...
+	index int
+}
+
+type errGracePeriodsInBothALBAndNLB struct {
+	errFieldMutualExclusive
+}
+
+func (e *errGracePeriodsInBothALBAndNLB) RecommendedActions() string {
+	return fmt.Sprintf(`It looks like you have specified grace_period under both %s and %s. 
+ECS allows "HealthCheckGracePeriodSeconds" as a service level setting and should be specified only once either in http or nlb configuration.`, e.firstField, e.secondField)
```
Lou1415926: nit: what do you think of the following ⬇️ 
```suggestion
	return fmt.Sprintf(`"grace_period" is a configuration shared by "http" and "nlb". Specify it only once under either "http" or "nlb".`, e.firstField, e.secondField)
```

> It looks like you have specified grace_period under both %s and %s.

I feel like this is an error description instead of a recommended action, and it is `Error()`'s duty to describe an error. I know we do this for some of the other errors, but in this case I think we don't need to be super verbose.

> ECS allows "HealthCheckGracePeriodSeconds" as a service level setting

I'm not sure if this necessarily make sense for users. Seeing this, users may think to themselves - "why is GracePeriods being a service-level setting a reason why you can specify it only once?" 


To understand this causation, the users have to understand how ECS services are hooked with load balancers (i.e. via target group). I feel like this is too deep of an understanding that we want to assume.

internal/pkg/manifest/errors.go
```diff
@@ -98,6 +98,42 @@ func (e *errFieldMutualExclusive) Error() string {
 	return fmt.Sprintf(`must specify one, not both, of "%s" and "%s"`, e.firstField, e.secondField)
 }
 
+type errGracePeriodSpecifiedInAdditionalRule struct {
+	index int
+}
+type errGracePeriodSpecifiedInAdditionalListener struct {
+	index int
+}
+
```
Lou1415926: Can we move the struct definition closer to their methods?

internal/pkg/manifest/errors.go
```diff
@@ -98,6 +98,42 @@ func (e *errFieldMutualExclusive) Error() string {
...
+	errFieldMutualExclusive
+}
+
+func (e *errGracePeriodsInBothALBAndNLB) RecommendedActions() string {
+	return fmt.Sprintf(`It looks like you have specified grace_period under both %s and %s. 
+ECS allows "HealthCheckGracePeriodSeconds" as a service level setting and should be specified only once either in http or nlb configuration.`, e.firstField, e.secondField)
+}
+
+func (e *errGracePeriodSpecifiedInAdditionalRule) Error() string {
+	return fmt.Sprintf(`cannot define "grace_period" in "http.additional_rules[%d].healthcheck.grace_period"`, e.index)
```
Lou1415926: nit:
```suggestion
	return fmt.Sprintf(`"grace_period" specified for "http.additional_rules[%d]"`, e.index)
```

internal/pkg/manifest/validate_test.go
```diff
@@ -48,6 +48,99 @@ func TestLoadBalancedWebService_validate(t *testing.T) {
...
+											GracePeriod:        durationp(9 * time.Second),
+										}),
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			wantedErrorMsgPrefix: `validate "grace_period": cannot define "grace_period" in "http.additional_rules[0].healthcheck.grace_period"`,
```
Lou1415926:
```suggestion
			wantedError: &errGracePeriodSpecifiedInAdditionalRule{0},
```

internal/pkg/manifest/validate_test.go
```diff
@@ -48,6 +48,99 @@ func TestLoadBalancedWebService_validate(t *testing.T) {
...
+						AdditionalListeners: []NetworkLoadBalancerListener{
+							{
+								Port:        stringP("80"),
+								HealthCheck: NLBHealthCheckArgs{GracePeriod: durationp(9 * time.Second)},
+							},
+						},
+					},
+				},
+			},
+			wantedErrorMsgPrefix: `validate "grace_period": cannot define "grace_period" in "nlb.additional_listeners[0].healthcheck.grace_period"`,
```
Lou1415926: samesies!

internal/pkg/manifest/validate_test.go
```diff
@@ -48,6 +48,99 @@ func TestLoadBalancedWebService_validate(t *testing.T) {
...
+					},
+					NLBConfig: NetworkLoadBalancerConfiguration{
+						Listener: NetworkLoadBalancerListener{
+							Port:        stringP("80"),
+							HealthCheck: NLBHealthCheckArgs{GracePeriod: durationp(9 * time.Second)},
+						},
+					},
+				},
+			},
+			wantedErrorMsgPrefix: `validate "grace_period": must specify one, not both, of "http.healthcheck.grace_period" and "nlb.healthcheck.grace_period"`,
```
Lou1415926: samesiess

Lou1415926 (approved):

internal/pkg/manifest/errors.go
paragbhingre: Yeah I agree with your thought. I think customer doesn't need to think so deep. We can just tell them what correction they need to do. 👍

internal/pkg/manifest/validate_test.go
paragbhingre: The thing is if we write `wantedError: &errGracePeriodSpecifiedInAdditionalRule{0},` then we only get error as `cannot define "grace_period" in "http.additional_rules[0].healthcheck.grace_period"` without prefix as `validate "grace_period":` and our validate function expects the error with `validate "grace_period":`. So I think what we have right now will work better.

paragbhingre: if we write `wantedError: &errGracePeriodSpecifiedInAdditionalRule{0},` then we get error as `cannot define "grace_period" in "http.additional_rules[0].healthcheck.grace_period"` without prefix as `validate "grace_period":` which is expected from the validate function. So what we have right now works better in this case.

Lou1415926: hmmmmm in that case we can do 
```
wantedError: fmt.Errorf(`validate "grace_period": %w`, &errGracePeriodSpecifiedInAdditionalRule{0})
```

`wantedErrorPrefix` is used to validate error prefixes such as "validate xxx" and is not meant to validate that the errors are equals... and if we change the error string returned by `errGracePeriodSpecifiedInAdditionalRule`'s `Error()`, we will have to modify this test case again.

iamhopaul123 (approved):
:shipit:

PR #4735, opened 2023-04-05
chore: workload stack deletion should use cfn execution role

iamhopaul123:
Right now use env manager role to delete workload stacks which are created by CFN execution role. Also for env stack deletion we use CFN execution role.

### DIFF
--- a/internal/pkg/cli/job_delete.go
+++ b/internal/pkg/cli/job_delete.go
@@ -252,7 +252,7 @@ func (o *deleteJobOpts) deleteJobs(envs []*config.Environment) error {
 			return err
 		}
 		// Delete job stack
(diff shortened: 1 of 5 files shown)


dannyrandall (approved):
wow, good catch!

paragbhingre (approved):
Looks great, just a dumb question. 
Is that the reason why one of the customers are not able to delete svc and jobs due to lack of permissions?

iamhopaul123:
> Looks great, just a dumb question.
Is that the reason why one of the customers are not able to delete svc and jobs due to lack of permissions?

No it's not a dumb question. I think they _might_ be related: they don't have any problem deleting the workload stacks which means their env manager roles are good. But im wondering if they manually modified the CFN execution role, causing for some reasons the env stack deletion just hanged.

PR #4740, opened 2023-04-07
fix: separate out deregistration delay for ALB and NLB

paragbhingre:
This PR creates a field `deregistration_delay` for nlb listeners and also fixes the bug where previously we used to assign `http.Main.deregistration_delay`'s value to all the target groups. After this fix, each target group will have its own deregistration delay.


internal/pkg/manifest/lb_web_svc.go
```diff
@@ -241,12 +241,13 @@ type NetworkLoadBalancerConfiguration struct {
...
-	TargetPort      *int               `yaml:"target_port"`
-	SSLPolicy       *string            `yaml:"ssl_policy"`
-	Stickiness      *bool              `yaml:"stickiness"`
+	Port                *string            `yaml:"port"`
+	HealthCheck         NLBHealthCheckArgs `yaml:"healthcheck"`
+	TargetContainer     *string            `yaml:"target_container"`
+	TargetPort          *int               `yaml:"target_port"`
+	SSLPolicy           *string            `yaml:"ssl_policy"`
+	Stickiness          *bool              `yaml:"stickiness"`
+	DeregistrationDelay *time.Duration     `yaml:"deregistration_delay"`
```
bvtujo, line 251: How discoverable is the deregistration_delay field in the backend service manifest package?  Is it possible that customers set one of these fields and expect it to set both of them, or is that considered a bug? Does that make sense?

I think as our manifest gets more complicated it makes sense for use to be careful to make sure that we have some way of surfacing fields that might get buried. +1 for intellisense schema

paragbhingre: If I am understanding your question correctly, then `deregistration_delay` is a target group level setting, so customers have to set that value for every listener in NLB and every listener rule in ALB. We support NLB only for LBWS for now, so coming back to your question of the backend service, we will only have ALB there, and we can set `deregistration_delay` on every listener rule of internal ALB.

bvtujo (approved):

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -583,19 +583,28 @@ func (conv routingRuleConfigConverter) convert() (*template.ALBListenerRule, err
...
+		Stickiness:          strconv.FormatBool(aws.BoolValue(conv.rule.Stickiness)),
+		HTTPVersion:         aws.StringValue(convertHTTPVersion(conv.rule.ProtocolVersion)),
+		RedirectToHTTPS:     conv.redirectToHTTPS,
+		DeregistrationDelay: convertDeregistrationDelay(conv.rule.DeregistrationDelay),
 	}
 	return config, nil
 }
 
+func convertDeregistrationDelay(dd *time.Duration) *int64 {
+	var deregistrationDelay *int64 = aws.Int64(60)
```
Lou1415926: Can we move `60` to a constant?

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -583,19 +583,28 @@ func (conv routingRuleConfigConverter) convert() (*template.ALBListenerRule, err
...
+		AllowedSourceIps:    convertAllowedSourceIPs(conv.rule.AllowedSourceIps),
+		Stickiness:          strconv.FormatBool(aws.BoolValue(conv.rule.Stickiness)),
+		HTTPVersion:         aws.StringValue(convertHTTPVersion(conv.rule.ProtocolVersion)),
+		RedirectToHTTPS:     conv.redirectToHTTPS,
+		DeregistrationDelay: convertDeregistrationDelay(conv.rule.DeregistrationDelay),
 	}
 	return config, nil
 }
 
+func convertDeregistrationDelay(dd *time.Duration) *int64 {
```
Lou1415926:
```suggestion
func convertDeregistrationDelay(delay *time.Duration) *int64 {
```

internal/pkg/deploy/cloudformation/stack/transformers.go
```suggestion
       if delay == nil {
          return aws.Int64(defaultDeregistrationDeplay)
       }
       return aws.Int64(int64(dd.Seconds()))
```

internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/http-full-config-template.yml
```diff
@@ -409,7 +409,7 @@ Resources:
       Protocol: HTTP
       TargetGroupAttributes:
         - Key: deregistration_delay.timeout_seconds
-          Value: 5 # ECS Default is 300; Copilot default is 60.
+          Value: 60 # ECS Default is 300; Copilot default is 60.
```
Lou1415926, line 412: Can you help me understand why the test result changed from 5 to 60?

efekarakus (approved):
LGTM, approving assuming Wanxian's comments will be addressed.

Feel free to remove the label afterwards

internal/pkg/deploy/cloudformation/stack/testdata/workloads/backend/http-full-config-template.yml
paragbhingre: Previously, we used to accept the deregistration value only from the main listener rule and assign it to all the target groups. But now we are having a deregistration delay for every target group. So in this particular test case, we assigned 5s to the main listener rule, and hence the additional listener rule's target group's deregistration_delay has a default value of 60.

PR #4748, opened 2023-04-10
chore: Implement a new `syncbuffer` package

KollaAdithya:
This PR will create a package `syncbuffer` that  provides safe bytes.Buffer as well printing functionality to the terminal.


internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+	"sync"
+)
+
+// The fileReader interface that can read data from a file.
```
paragbhingre:
```suggestion
// fileReader interface that can read data from a file.
```

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,112 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+	"sync"
+)
+
+// fileReader interface that can read data from a file.
+// It extends the io.Reader interface, which provides a method for reading bytes into a buffer.
+type fileReader interface {
```
efekarakus: why do we need this type instead of using an `io.Reader` directly?

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,112 @@
...
+
+// SyncBuffer is a synchronized buffer that can be used to store output data and coordinate between multiple goroutines.
+type SyncBuffer struct {
+	bufMu sync.Mutex    // bufMu is a mutex that can be used to protect access to the buffer.
+	buf   bytes.Buffer  // buf is the buffer that stores the data.
+	done  chan struct{} // done is a channel that can be used to signal when the operations are complete.
+}
+
+// NewSyncBuffer creates and returns a new SyncBuffer object with an initialized 'done' channel.
+func NewSyncBuffer() *SyncBuffer {
```
efekarakus:
```suggestion
func New() *SyncBuffer {
```
So that it reads `syncbuffer.New()`

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,112 @@
...
+// Write appends the given bytes to the buffer.
+func (b *SyncBuffer) Write(p []byte) (n int, err error) {
+	b.bufMu.Lock()
+	defer b.bufMu.Unlock()
+	return b.buf.Write(p)
+}
+
+// strings returns an empty slice if the buffer is empty.
+// Otherwise, it returns a slice of all the lines stored in the buffer.
+func (b *SyncBuffer) strings() []string {
```
efekarakus: nit: can we move the private methods below the public

nit: 
```suggestion
func (b *SyncBuffer) lines() []string {
```

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+}
+
+// MarkDone closes the Done channel.
+func (b *SyncBuffer) MarkDone() {
+	close(b.done)
+}
+
+// CopyToBuffer reads from a fileReader and writes the data to the buffer.
+func (b *SyncBuffer) CopyToBuffer(r fileReader) error {
```
efekarakus: I wonder if this should be another constructor, besides `New`, like:
```suggestion
// Copy reads all the content of an io.Reader into a SyncBuffer and returns it.
func Copy(r io.Reader) (*SyncBuffer, error) {
```

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,112 @@
...
+
+// New creates and returns a new LabeledSyncBuffer without label.
+func New(syncBuf *SyncBuffer) *LabeledSyncBuffer {
+	return &LabeledSyncBuffer{
+		syncBuf: syncBuf,
+	}
+}
+
+// NewWithLabel creates and returns a new LabeledSyncBuffer with the given label and SyncBuffer.
+func NewWithLabel(label string, syncBuf *SyncBuffer) *LabeledSyncBuffer {
```
efekarakus: I wonder if this should be a method on a  syncbuffer instead so that it reads like this:
```suggestion
func (buf *SyncBuffer) WithLabel(label string) *LabeledSyncBuffer {
```

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,112 @@
...
+	label   string
+	syncBuf *SyncBuffer
+}
+
+// New creates and returns a new LabeledSyncBuffer without label.
+func New(syncBuf *SyncBuffer) *LabeledSyncBuffer {
+	return &LabeledSyncBuffer{
+		syncBuf: syncBuf,
+	}
+}
```
efekarakus: Do we need this constructor? it doesn't seem to bring an additional benefit to a `syncbuffer`

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,112 @@
...
+// NewWithLabel creates and returns a new LabeledSyncBuffer with the given label and SyncBuffer.
+func NewWithLabel(label string, syncBuf *SyncBuffer) *LabeledSyncBuffer {
+	return &LabeledSyncBuffer{
+		label:   label,
+		syncBuf: syncBuf,
+	}
+}
+
+// CopyToLabeledBuffer reads from a fileReader and writes the data to the buffer of LabeledSyncBuffer.
+func (b *LabeledSyncBuffer) CopyToLabeledBuffer(r fileReader) error {
```
efekarakus: I wonder if we need this?
I'm thinking clients can use the following flow:
```go
buf, err := syncbuffer.Copy(someReader /* such as a pipe reader */)
if err != nil { return err }
labeledBuf := buf.WithLabel("title")
```

internal/pkg/term/syncbuffer/syncbuffer_test.go
```diff
@@ -0,0 +1,167 @@
...
+			actual := tc.buffer.IsDone()
+
+			// THEN
+			require.Equal(t, tc.wantedDone, actual)
+
+		})
+	}
+}
+
+func TestSyncBuffer_strings(t *testing.T) {
```
efekarakus: no need for unit tests for private methods, we should be able to test everything from exposed methods

internal/pkg/term/syncbuffer/syncbuffer_test.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+}
+
+func TestSyncBuffer_IsDone(t *testing.T) {
+	testCases := map[string]struct {
+		buffer     *SyncBuffer
+		wantedDone bool
+	}{
+		"Buffer is done": {
+			buffer:     &SyncBuffer{done: make(chan struct{}), buf: bytes.Buffer{}},
```
efekarakus: consider always using _only_ the public API of a package (i.e. don't set any private fields or check against private fields in unit tests).

This is because these fields are internal details, we should be able to change the guts of a struct without having to update the unit tests. In this situation, you'd do:
```suggestion
			buffer:     New(),
```

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/aws/copilot-cli/internal/pkg/term/cursor"
+	"golang.org/x/term"
+)
+
+// pollIntervalForSyncBuffer is the time Interval to wait between checking whether the output buffers are done.
+const pollIntervalForSyncBuffer = 60 * time.Millisecond
+
+// FileWriter is the interface to write to a file.
+type FileWriter interface {
+	io.Writer
```
efekarakus, line 24: Don't we need `Fd() uintptr` as well for this interface? that's what makes it a _File_Writer (the fact that there is a method for a file descriptor)

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+		if doneCount >= len(ltp.buffers) {
+			break
+		}
+	}
+}
+
+// terminalWidth returns the width of the terminal or an error if failed to get the width of terminal.
+func terminalWidth() (int, error) {
+	width, _, err := term.GetSize(int(os.Stderr.Fd()))
```
efekarakus: this should be:
```suggestion
func terminalWidth(fw FileWriter) (int, error) {
	width, _, err := term.GetSize(int(fw.Fd()))
```

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,184 @@
...
+		return nil, fmt.Errorf("get terminal width: %w", err)
+	}
+
+	ltp := &LabeledTermPrinter{
+		term:      fw,
+		buffers:   bufs,
+		termWidth: width,
+		numLines:  -1, // By default set numlines to -1 to print all from buffers.
+	}
+	for _, opt := range opts {
```
efekarakus: nice!

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	return true
+}
+
+// Print prints the label and the last N lines of logs from each buffer to the termPrinter fileWriter.
+// If numLines is -1 then print all the values from buffers.
+// It polls each buffer until all the buffers are marked done,
+// and erases the previous output after sleeping for a short duration.
+func (ltp *LabeledTermPrinter) Print() {
+	if ltp.numLines == -1 {
```
efekarakus: nit: consider moving [magic numbers](https://refactoring.com/catalog/replaceMagicLiteral.html) to consts:

```go
const (
   printAllLinesInBuf = -1
)

```

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,184 @@
...
+		ltp.PrintAll()
+		return
+	}
+	for {
```
efekarakus: I wonder if the whole `pollIntervalForSyncBuffer` and forever-loop should be handled instead by the client of this package.

For example, this method would be:
```go
if ltp.numLines == printAllLinesInBuf {
   ltp.printAll()
   return
}
if ltp.prevWrittenLines > 0 { cursor.EraseLinesAbove(ltp.FileWriter, ltp.prevWrittenLines) } 
ltp.prevWrittenLines = 0
for _, buf := range ltp.buffers {
 ....
}
```

and then the client of this package would use:
```go
ltp := syncbuffer.NewLabeledTermPrinter(...)
for {
   if ltp.IsDone() { break }
   ltp.Print()
   time.Sleep(60 * time.Millisecond)
}
```

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,184 @@
...
+			}
+			outputLogs := ltp.lastNLines(logs)
+			ltp.writeLines(buf.label, outputLogs)
+			ltp.prevWrittenLines += ltp.calculateLinesCount(append(outputLogs, buf.label))
+		}
+		if ltp.isDone() {
+			break
+		}
+		time.Sleep(pollIntervalForSyncBuffer)
+		cursor.EraseLinesAbove(os.Stderr, ltp.prevWrittenLines)
```
efekarakus: we should be using the `FileWriter` not `os.Stderr` directly :)

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	return logLines
+}
+
+// writeLines writes a label and output logs to the terminal associated with the TermPrinter.
+func (ltp *LabeledTermPrinter) writeLines(label string, outputLogs []string) {
+	padding := ""
+	if ltp.padding != 0 {
+		padding = strings.Repeat(" ", ltp.padding)
+	}
```
efekarakus: this should be safe even when 0
```suggestion
	padding := strings.Repeat(" ", ltp.padding)
```

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+		numLines += math.Ceil(float64(len(line)) / float64(ltp.termWidth))
+	}
+	return int(numLines)
+}
+
+// PrintAll writes the entire contents of all the buffers to the file writer.
+// If one of the buffer gets done then print entire content of the buffer.
+// Until all the buffers are written to file writer.
+func (ltp *LabeledTermPrinter) PrintAll() {
```
efekarakus: does this need to be public?

efekarakus (commented):
yayay!

internal/pkg/term/syncbuffer/termprinter.go
KollaAdithya: But the `log.DiagnosticWriter` does not implement the below interface if we have `Fd() uintptr`. What can we do in that case 🤔 

```yaml
type FileWriter interface {
   io.Writer
   Fd() uintptr
}
```

internal/pkg/term/syncbuffer/termprinter.go
KollaAdithya: I will make this private

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,112 @@
...
+// strings returns an empty slice if the buffer is empty.
+// Otherwise, it returns a slice of all the lines stored in the buffer.
+func (b *SyncBuffer) strings() []string {
+	b.bufMu.Lock()
+	defer b.bufMu.Unlock()
+	lines := b.buf.String()
+	if len(lines) == 0 {
+		return nil
+	}
+	return strings.Split(strings.TrimSpace(lines), "\n")
```
Lou1415926: The doc for [`TrimSpace`](https://pkg.go.dev/strings#TrimSpace) says
> with all leading and trailing white space removed

And new line is considered a white space. This makes me wonder if this works nicely with `Split` with `\n` 🤔

internal/pkg/term/syncbuffer/termprinter.go
efekarakus: :cool-crying: we can pass `os.Stderr` then instead of `log.DiagnosticWriter` (they're the same)

internal/pkg/term/syncbuffer/syncbuffer_test.go
```diff
@@ -0,0 +1,150 @@
...
+func (r *mockReader) Read(p []byte) (int, error) {
+	if len(r.data) == 0 {
+		return 0, r.err
+	}
+	n := copy(p, r.data)
+	r.data = r.data[n:]
+	return n, io.EOF
+}
+
+func TestSyncBuffer_CopyToBuffer(t *testing.T) {
```
efekarakus: nit:
```suggestion
func TestCopy(t *testing.T) {
```

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,99 @@
...
+// MarkDone closes the Done channel.
+func (b *SyncBuffer) MarkDone() {
+	close(b.done)
+}
+
+// LabeledSyncBuffer is a struct that combines a SyncBuffer with a string label.
+type LabeledSyncBuffer struct {
+	label   string
+	syncBuf *SyncBuffer
+}
```
efekarakus: nit: consider embedding `*SyncBuffer` instead since a `LabeledSyncBuffer` **_is a_** `SyncBuffer`:
```suggestion
type LabeledSyncBuffer struct {
	label   string
	*SyncBuffer
}
```
this way for example in the `Print` method of the `ltp` you can write:
```go
logs := buf.lines() // instead of logs := buf.syncBuf.lines()
```

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,176 @@
...
+		ltp.prevWrittenLines += ltp.calculateLinesCount(append(outputLogs, buf.label))
+	}
+}
+
+// printAll writes the entire contents of all the buffers to the file writer.
+// If one of the buffer gets done then print entire content of the buffer.
+// Until all the buffers are written to file writer.
+func (ltp *LabeledTermPrinter) printAll() {
+	doneCount := 0
+	for {
```
efekarakus: I'm a bit confused on `printAll`. Isn't this _blocking_ until all the buffers are done?

Instead the behavior that we'd want is just the inner for-loop for this method:
```go
for idx, buf := range ltp.buffers {
	if !buf.syncBuf.IsDone() {
		continue
	}
	outputLogs := ltp.buffers[idx].syncBuf.lines()
	ltp.writeLines(buf.label, outputLogs)
}
```

and the client of `Print` should do:
```go
ltp := syncbuffer.NewLabeledTermPrinter(...)
for {
   if ltp.IsDone() { break }
   ltp.Print()
   time.Sleep(60 * time.Millisecond)
}
```

internal/pkg/term/syncbuffer/termprinter_test.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func TestLabeledTermPrinter_Print(t *testing.T) {
+	testCases := map[string]struct {
+		inLabeledTermPrinter LabeledTermPrinter
+		wanted               string
+	}{
+		"display label with given numLines": {
+			inLabeledTermPrinter: LabeledTermPrinter{
+				buffers: []*LabeledSyncBuffer{
```
efekarakus: it's the same for these test cases, can we use the public API of this package?

efekarakus (approved):
Let's goo! 🚀

internal/pkg/term/syncbuffer/syncbuffer.go
KollaAdithya: Good catch! I created a new function to handle this case,`splitLinesAndTrimSpaces` that first split all lines and then trim the spaces from each line.
Does this sound good?

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,99 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+	"sync"
+)
+
+// SyncBuffer is a synchronized buffer that can be used to store output data and coordinate between multiple goroutines.
+type SyncBuffer struct {
+	bufMu sync.Mutex    // bufMu is a mutex that can be used to protect access to the buffer.
```
dannyrandall:
```suggestion
	bufMu sync.Mutex    // bufMu is a mutex that protects buf.
```

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,99 @@
...
+	buf   bytes.Buffer  // buf is the buffer that stores the data.
+	done  chan struct{} // done is a channel that can be used to signal when the operations are complete.
```
dannyrandall:
```suggestion
	done  chan struct{} // done is a channel that is closed after MarkDone() is called.
```

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,99 @@
...
+		SyncBuffer: buf,
+	}
+}
+
+// Copy reads all the content of an io.Reader into a SyncBuffer and returns it.
+func Copy(r io.Reader) (*SyncBuffer, error) {
+	syncBuf := New()
+	defer func() {
+		syncBuf.MarkDone()
+	}()
```
dannyrandall:
```suggestion
	defer syncBuf.MarkDone()
```

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// WithPadding sets the padding of LabeledTermPrinter.
+func WithPadding(n int) LabeledTermPrinterOption {
+	return func(ltp *LabeledTermPrinter) {
+		ltp.padding = n
+	}
+}
+
+// IsDone returns true if all the buffers are done else returns false.
```
dannyrandall:
```suggestion
// IsDone returns true if all the buffers are done.
```

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,99 @@
...
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+	"sync"
+)
+
+// SyncBuffer is a synchronized buffer that can be used to store output data and coordinate between multiple goroutines.
+type SyncBuffer struct {
```
dannyrandall, line 16: i wonder if we need to export `SyncBuffer` _and_ `LabeledSyncBuffer`? Seems like we'll only ever use the `LabeledSyncBuffer` (correct me if i'm wrong!)

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,169 @@
...
+
+// LabeledTermPrinterOption is a type alias to configure LabeledTermPrinter.
+type LabeledTermPrinterOption func(ltp *LabeledTermPrinter)
+
+// NewLabeledTermPrinter returns a LabeledTermPrinter that can print to the terminal filewriter from buffers.
+func NewLabeledTermPrinter(fw FileWriter, bufs []*LabeledSyncBuffer, opts ...LabeledTermPrinterOption) (*LabeledTermPrinter, error) {
+	width, err := terminalWidth(fw)
+	if err != nil {
+		return nil, fmt.Errorf("get terminal width: %w", err)
+	}
```
dannyrandall: shouldn't we calculate this every time `Print()` is called? in case someone changes the size of their terminal after the `LabeledTermPrinter` is created?

internal/pkg/term/syncbuffer/syncbuffer.go
KollaAdithya: As of Now for `Build And Push` we only use the `Labeled SyncBuffer`, may be in future for some other use cases we can use the `SyncBuffer` with out any label.

dannyrandall: i'd say let's not export it until we find out it's needed! not a huge deal either way since it's an internal package, though.

internal/pkg/term/syncbuffer/termprinter.go
KollaAdithya: That is a valid case. I will call the terminal width whenever we are printing the logs to terminal instead of initializing in the constructor.

internal/pkg/term/syncbuffer/syncbuffer.go
KollaAdithya: 🤔 My major concern with Copy returning *SyncBuffer `func Copy(r io.Reader) (*SyncBuffer, error)`
 as constructor is that we need to wait until all the data is copied to the buffer before printing from it.

The desired approach is to print logs from the buffer as soon as we get some data in it. I.e Both copying and printing should go in parallel.   
```
lbSyncBuffers := make([]*syncbuffer.LabeledSyncBuffer, len(buildArgsPerContainer))
count := 0 

g.Go(func() error {
  syncBuf, err := syncbuffer.Copy(pr) // Wait until the copy to the buffer is done and return the SyncBuffer, then attach the label
  lbSyncBuf := syncBuf.WithLabel(label)
  mu.Lock()
  lbSyncBuffers[count] = lbSyncBuf
  count++
  defer mu.Unlock()
  return nil
})

 ltp, err := syncbuffer.NewLabeledTermPrinter(os.Stderr, lbSyncBuffers)
 
 g.Go(func() error {
        // Print Logic
})
 ```
👀 However, when creating  the `LabeledTermPrinter` object, the `lbSyncBuffers` will be nil because the copy goroutine has not completed the assigned work yet.
This means that by the time the Print goroutine runs, we do not have any data in the buffers of LabeledTermPrinter.


 To address this issue, What I want is to create a new SyncBuffer directly and then attaching a label to it instead of copy returning the SyncBuffer. This way, we can print logs as soon as we get the data into the buffer:

```
// An array of LabeledSyncBuffers
lbSyncBuffers := make([]*syncbuffer.LabeledSyncBuffer, len(buildArgsPerContainer))
count := 0 

// Create a new SyncBuffer and attach a label to it
syncBuf := syncbuffer.New()
lbSyncBuf := syncBuf.WithLabel(label)
lbSyncBuffers[count] = lbSyncBuf
count++

g.Go(func() error {
if err := syncbuffer.Copy(syncBuf,pr)
})

 ltp, err := syncbuffer.NewLabeledTermPrinter(os.Stderr, lbSyncBuffers)

g.Go(func() error {
// Print Logic
})
```
🤔 I am going to modify the Copy function as below if the reasoning sounds good:
`func( buf *SyncBuffer) Copy(r io.Reader) (error)`

File: internal/pkg/term/syncbuffer/termprinter.go (hunk 29 above)
Comment (KollaAdithya) [reply internal/pkg/term/syncbuffer/termprinter.go]: Done!

File: internal/pkg/term/syncbuffer/syncbuffer.go [hunk 30]
```diff
@@ -0,0 +1,96 @@
...
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+	"sync"
+)
+
+// syncBuffer is a synchronized buffer that can be used to store output data and coordinate between multiple goroutines.
+type syncBuffer struct {
```
dannyrandall: i'm realizing this package will be used like 
```
buf := syncbuffer.New()
labeledBuf = buf.WithLabel("asdf")
```

maybe we should only export a `New(label string) *LabeledSyncBuffer` function? it's kinda weird have an exported function return an unexported struct. alternatively, we could just re-export `SyncBuffer` and keep the usage the same (sorry for my prev comment!) - i'll leave it up to you!

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,96 @@
...
+		syncBuffer: buf,
+	}
+}
+
+// Copy reads all the content of an io.Reader into a SyncBuffer and an error if copy is failed.
+func Copy(syncBuf *syncBuffer, r io.Reader) error {
+	defer syncBuf.MarkDone()
+	_, err := io.Copy(&syncBuf.buf, r)
+	if err != nil && !errors.Is(err, io.EOF) {
+		return fmt.Errorf("failed to copy to buffer: %w", err)
```
dannyrandall:
```suggestion
		return err
```
probably don't to add any context to this error, since this function would logically be wrapped with pretty much the same message like:
```
buf := syncbuffer.New()
if err := syncbuffer.Copy(); err != nil {
  return fmt.Errorf("copy buffer: %w", err)
}
```

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,176 @@
...
+// to the LabeledTermPrinter fileWriter and erases the previous output.
+// If numLines is -1 then print all the values from buffers.
+func (ltp *LabeledTermPrinter) Print() error {
+	if ltp.numLines == printAllLinesInBuf {
+		ltp.printAll()
+		return nil
+	}
+	if ltp.prevWrittenLines > 0 {
+		cursor.EraseLinesAbove(ltp.term, ltp.prevWrittenLines)
+	}
```
dannyrandall, line 93: does it cause errors to remove the `if` and just call `cursor.EraseLinesAbove(ltp.term, 0)` if `prevWrittenLines` is 0?

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,176 @@
...
+func (ltp *LabeledTermPrinter) writeLines(label string, outputLogs []string) {
+	fmt.Fprintln(ltp.term, label)
+	for _, logLine := range outputLogs {
+		fmt.Fprintf(ltp.term, "%s%s\n", strings.Repeat(" ", ltp.padding), logLine)
+	}
+}
+
+// calculateLinesCount returns the number of lines needed to print the given string slice based on the terminal width
+// or an error when failed to fetch terminal width.
+func (ltp *LabeledTermPrinter) calculateLinesCount(lines []string) (int, error) {
```
dannyrandall: this function doesn't take into account any width added by padding, right? i wonder if changing `writeLines` to return an `int` of how many lines it wrote would help keep the two functionalities (printing & calculating width) in sync 🤔

internal/pkg/term/syncbuffer/termprinter.go
KollaAdithya: It will give an error. If i remove if block then `EraseLinesAbove` will try to erase the [current line](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/term/cursor/cursor.go#L98) as well.

internal/pkg/term/syncbuffer/syncbuffer.go
KollaAdithya: I reverted back to public

internal/pkg/term/syncbuffer/termprinter.go
KollaAdithya: Instead I changed the signature of the function  to`calculateLinesCount(label string, lines []string)` and added the `padding` for each line except label as we do not have any padding on that

internal/pkg/term/syncbuffer/syncbuffer.go
```diff
@@ -0,0 +1,95 @@
...
+// WithLabel creates and returns a new LabeledSyncBuffer with the given label and SyncBuffer.
+func (buf *SyncBuffer) WithLabel(label string) *LabeledSyncBuffer {
+	return &LabeledSyncBuffer{
+		label:      label,
+		SyncBuffer: buf,
+	}
+}
+
+// Copy reads all the content of an io.Reader into a SyncBuffer and an error if copy is failed.
+func Copy(syncBuf *SyncBuffer, r io.Reader) error {
```
dannyrandall:
```suggestion
func (buf *SyncBuffer) Copy(r io.Reader) error {
```

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("get terminal width: %w", err)
+	}
+	numLines := float64(len(label))
+	for _, line := range lines {
+		// Empty line should be considered as a new line
+		if line == "" {
+			numLines += 1
+		}
+		numLines += math.Ceil(float64(len(line)) + float64(ltp.padding)/float64(width))
```
dannyrandall: i think the order of operations here is wrong - it looks like `len(line) + ltp.padding/width`, but should be `(len(line) + ltp.padding))/width)`. also, we aren't factoring in the width on the label line 🤔 

i think we can make this more consistent - and only loop through lines once - if we change `writeLines` to output the number of lines written. something like:
```go
writeLines() (int, error) {  
  width, err := termWidth()
  // handle err

  numLines := 0
  writeLine := func(line string) {
    fmt.Fprintln(ltp.term, line)
    if len(line) == 0 {
      numLines++
      return
    }
    numLines += math.Ceil(float64(len(line))/float64(width))
  }

  writeLine(ltp.label)
  for _, line := range lines {
    writeLine(fmt.Sprintf("%s%s", strings.Repeat("", ltp.padding), line))
  }
}
```

dannyrandall (commented):
one small thing!

internal/pkg/term/syncbuffer/termprinter.go
KollaAdithya: Done!

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	if ltp.prevWrittenLines > 0 {
+		cursor.EraseLinesAbove(ltp.term, ltp.prevWrittenLines)
+	}
+	ltp.prevWrittenLines = 0
+	for _, buf := range ltp.buffers {
+		logs := buf.lines()
+		if len(logs) == 0 {
+			continue
+		}
```
dannyrandall: i wonder if we should still print the label? i'll approve, you decide what makes more sense!

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,178 @@
...
+// If one of the buffer gets done then print entire content of the buffer.
+// Until all the buffers are written to file writer.
+func (ltp *LabeledTermPrinter) printAll() error {
+	for idx, buf := range ltp.buffers {
+		if !buf.IsDone() {
+			continue
+		}
+		outputLogs := ltp.buffers[idx].lines()
+		if _, err := ltp.writeLines(buf.label, outputLogs); err != nil {
+			return fmt.Errorf("get terminal width: %w", err)
```
dannyrandall:
```suggestion
			return fmt.Errorf("write logs: %w", err)
```

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -0,0 +1,178 @@
...
+	ltp.prevWrittenLines = 0
+	for _, buf := range ltp.buffers {
+		logs := buf.lines()
+		if len(logs) == 0 {
+			continue
+		}
+		outputLogs := ltp.lastNLines(logs)
+		writtenLines, err := ltp.writeLines(buf.label, outputLogs)
+		if err != nil {
+			return fmt.Errorf("get terminal width: %w", err)
```
dannyrandall:
```suggestion
			return fmt.Errorf("write logs: %w", err)
```

dannyrandall (approved):

internal/pkg/term/syncbuffer/termprinter.go
KollaAdithya: removed that if block to print the label even if we do not have logs in the buffer

PR #4750, opened 2023-04-10
refactor: fetch docker build label in `workload` and  ` taskrun`

KollaAdithya:
This PR will fetch the Docker Build label and print before before and push happens. In the next PR we will pass the label to the `syncbuffer` package.

### DIFF
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -265,7 +265,7 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 		resources:                resources,
 		workspacePath:            ws.Path(),
 		uri:                      uri,
-		fs:                       &afero.Afero{Fs: afero.NewOsFs()},
+		fs:                       afero.NewOsFs(),
 		s3Client:                 s3.New(envSession),
 		addons:                   addons,
 		repository:               repository,
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -380,7 +380,7 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 		if err != nil {
 			return fmt.Errorf("generate docker build args: %w", err)
 		}
-		log.Infof(dockerengine.DockerBuildLabel(buildArgs.Platform, buildArgsList))
+		log.Infof(fmt.Sprintf("Building your container image: docker %s\n", strings.Join(buildArgsList, " ")))
 		digest, err := d.repository.BuildAndPush(buildArgs)
 		if err != nil {
 			return fmt.Errorf("build and push image: %w", err)
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -374,6 +374,7 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 		if err != nil {
 			return fmt.Errorf("generate docker build args: %w", err)
 		}
+		// TODO(adi) handle this log in syncBuffer's Print function
 		log.Infof("Building your container image: docker %s\n", strings.Join(buildArgsList, " "))
 		digest, err := d.repository.BuildAndPush(buildArgs)
 		if err != nil {
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -28,6 +28,7 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/deploy/upload/customresource"
 	"github.com/aws/copilot-cli/internal/pkg/describe"
 	"github.com/aws/copilot-cli/internal/pkg/docker/dockerengine"
+	"github.com/aws/copilot-cli/internal/pkg/exec"
 	"github.com/aws/copilot-cli/internal/pkg/manifest"
 	"github.com/aws/copilot-cli/internal/pkg/repository"
 	"github.com/aws/copilot-cli/internal/pkg/template"
@@ -68,6 +69,7 @@ func (noopActionRecommender) RecommendedActions() []string {
 }
 
 type repositoryService interface {
+	URI() (string, error)
 	Login() error
 	BuildAndPush(args *dockerengine.BuildArguments) (string, error)
 }
@@ -147,6 +149,7 @@ type workloadDeployer struct {
 	mft           interface{}
 	rawMft        []byte
 	workspacePath string
+	uri           string
 
 	// Dependencies.
 	fs               afero.Fs
@@ -229,6 +232,10 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 	repoName := fmt.Sprintf("%s/%s", in.App.Name, in.Name)
 	repository := repository.NewWithURI(
 		ecr.New(defaultSessEnvRegion), repoName, resources.RepositoryURLs[in.Name])
+	uri, err := repository.URI()
+	if err != nil {
+		return nil, fmt.Errorf("get ECR repository URI: %w", err)
+	}
 	store := config.NewSSMStore(identity.New(defaultSession), ssm.New(defaultSession), aws.StringValue(defaultSession.Config.Region))
 	envDescriber, err := describe.NewEnvDescriber(describe.NewEnvDescriberConfig{
 		App:         in.App.Name,
@@ -249,6 +256,7 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 	}
 
 	cfn := cloudformation.New(envSession, cloudformation.WithProgressTracker(os.Stderr))
+
 	return &workloadDeployer{
 		name:                     in.Name,
 		app:                      in.App,
@@ -256,7 +264,8 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 		image:                    in.Image,
 		resources:                resources,
 		workspacePath:            ws.Path(),
-		fs:                       afero.NewOsFs(),
+		uri:                      uri,
+		fs:                       &afero.Afero{Fs: afero.NewOsFs()},
 		s3Client:                 s3.New(envSession),
 		addons:                   addons,
 		repository:               repository,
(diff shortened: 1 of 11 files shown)


internal/pkg/cli/deploy/workload.go
```diff
@@ -249,14 +256,16 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
...
 	return &workloadDeployer{
 		name:                     in.Name,
 		app:                      in.App,
 		env:                      in.Env,
 		image:                    in.Image,
 		resources:                resources,
 		workspacePath:            ws.Path(),
-		fs:                       afero.NewOsFs(),
+		uri:                      uri,
+		fs:                       &afero.Afero{Fs: afero.NewOsFs()},
```
paragbhingre: Why do we do this?

internal/pkg/cli/deploy/workload.go
```diff
@@ -380,7 +380,7 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 		if err != nil {
 			return fmt.Errorf("generate docker build args: %w", err)
 		}
-		log.Infof(dockerengine.DockerBuildLabel(buildArgs.Platform, buildArgsList))
+		log.Infof(fmt.Sprintf("Building your container image: docker %s\n", strings.Join(buildArgsList, " ")))
```
paragbhingre: Is this because we want to show this log every single time when docker is building image?

internal/pkg/docker/dockerengine/dockerengine.go
bvtujo, line 1: Excellent refactor 👍

internal/pkg/cli/task_run.go
```diff
@@ -965,12 +967,18 @@ func (o *runTaskOpts) buildAndPushImage() error {
 	if o.dockerfileContextPath != "" {
 		ctx = o.dockerfileContextPath
 	}
-
-	if _, err := o.repository.BuildAndPush(&dockerengine.BuildArguments{
+	buildArgs := &dockerengine.BuildArguments{
+		URI:        uri,
```
bvtujo, line 966: So this struct is missing some of its fields because they aren't supported in task run, right?

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -15,7 +15,6 @@ import (
 	"strings"
 
 	"github.com/aws/copilot-cli/internal/pkg/exec"
-	"github.com/aws/copilot-cli/internal/pkg/term/log"
```
bvtujo, line 18: 🎉  reducing dependencies

paragbhingre (commented):
Looks good. just few questions

internal/pkg/cli/deploy/workload.go
bvtujo: +1 to parag's question. Is it to make sure that the fs dependency is a pointer and not an object?

internal/pkg/cli/deploy/workload.go
bvtujo: Looks like it removes a log dependency from dockerEngine 👍

internal/pkg/cli/deploy/workload.go
KollaAdithya: I wonder this is done by editor automatically lol. I will revert this `fs`field.

internal/pkg/cli/deploy/workload.go
KollaAdithya: > Is this because we want to show this log every single time when docker is building image?
 
 Yeah that's true. We need to print label every time for each image.  In the next PR , this `build label` will passed to [`LabeledSyncBuffer`](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/term/syncbuffer/syncbuffer.go#L53) struct and then `Print` function will handle the logging.

internal/pkg/cli/task_run.go
KollaAdithya: yeah we need to show `uri` in the `build label` as well. Previously we do not set the `uri` in the buildArgs. Instead we set the `uri` over here. https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/repository/repository.go#L58

internal/pkg/cli/deploy/workload.go
KollaAdithya: reverted the `fs` field

internal/pkg/cli/deploy/workload.go
```diff
@@ -147,6 +149,7 @@ type workloadDeployer struct {
 	mft           interface{}
 	rawMft        []byte
 	workspacePath string
+	uri           string
```
dannyrandall: does this need to be on the `workloadDeployer`? it looks like it's only used in the `uploadContainerImages()` function 🤔

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,6 +376,11 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	}
 	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
 	for name, buildArgs := range buildArgsPerContainer {
+		buildArgsList, err := buildArgs.GenerateDockerBuildArgs(dockerengine.New(exec.NewCmd()))
+		if err != nil {
+			return fmt.Errorf("generate docker build args: %w", err)
+		}
+		log.Infof(fmt.Sprintf("Building your container image: docker %s\n", strings.Join(buildArgsList, " ")))
```
dannyrandall: i know we are getting rid of this in the next PR anyways, but we don't need the `fmt.Sprintf` as `log.Infof` already does formatting

internal/pkg/cli/deploy/workload.go
KollaAdithya: I can modify function `Login`  to return the uri as well 🤔

KollaAdithya: removed `uri` from `workloadDeployer`. Instead modified the `login` to return uri as well.

bvtujo (approved):
LGTM!

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -138,14 +133,23 @@ func (c CmdClient) Build(in *BuildArguments) error {
...
+	return args, nil
+}
+
+type dockerConfig struct {
+	CredsStore  string            `json:"credsStore,omitempty"`
+	CredHelpers map[string]string `json:"credHelpers,omitempty"`
+}
+
+// Build will run a `docker build` command for the given ecr repo URI and build arguments.
+func (c CmdClient) Build(in *BuildArguments) error {
```
Lou1415926: If `CmdClient` is specifically designed for running `docker` commands, can we rename it to `DockerCmdClient`?

internal/pkg/repository/repository.go
```diff
@@ -89,21 +89,21 @@ func (r *Repository) URI() (string, error) {
 // Login authenticates with a ECR registry by performing a Docker login,
 // but only if the `credStore` attribute value is not set to `ecr-login`.
 // If the `credStore` value is `ecr-login`, no login is performed.
-// Returns an error, if any occurs during the login process.
-func (r *Repository) Login() error {
+// Returns uri of the repository or an error, if any occurs during the login process.
+func (r *Repository) Login() (string, error) {
 	uri, err := r.URI()
```
Lou1415926: Do you know if `URI()` still needs to be exported?

Lou1415926 (commented):
Looks good! I just have two nits!

internal/pkg/cli/deploy/workload.go
```diff
@@ -361,12 +363,18 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
 		return fmt.Errorf("login to image repository: %w", err)
 	}
 	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
 	for name, buildArgs := range buildArgsPerContainer {
+		buildArgs.URI = uri
+		buildArgsList, err := buildArgs.GenerateDockerBuildArgs(dockerengine.New(exec.NewCmd()))
+		if err != nil {
+			return fmt.Errorf("generate docker build args: %w", err)
+		}
+		log.Infof("Building your container image: docker %s\n", strings.Join(buildArgsList, " "))
```
Lou1415926, line 378: Should this use `spinner`?

internal/pkg/cli/task_run.go
```diff
@@ -965,12 +962,18 @@ func (o *runTaskOpts) buildAndPushImage() error {
...
 		Dockerfile: o.dockerfilePath,
 		Context:    ctx,
 		Tags:       append([]string{imageTagLatest}, additionalTags...),
-	}); err != nil {
+	}
+	buildArgsList, err := buildArgs.GenerateDockerBuildArgs(dockerengine.New(exec.NewCmd()))
+	if err != nil {
+		return fmt.Errorf("generate docker build args: %w", err)
+	}
+	log.Infof("Building your container image: docker %s\n", strings.Join(buildArgsList, " "))
```
Lou1415926, line 975: I think this is probably fine *for now*. We are basically pretending that `cli` knows `repository` builds & pushes via `docker`; however, as a matter of fact this information is hidden from `cli`.

This PR made me realize that we might need a larger refactor around `repository` and `dockerengine` package.

Currently, our `cli` package mostly interacts with `dockerenging` indirectly through `repository` (`cli` -> `repository` -> `dockerengine`). That is, today `cli` is agnostic of whether the login/build/push is done through `docker`. 

However, this does not reflect the real interaction between an image repo, an engine, and a client that wants to build/push images. An image repository is not tied to a particular engine. Instead of `cli` -> `repository` -> `dockerengine`, it prolly should have been something like `dockerengine` <- `cli` -> `repository`.

internal/pkg/repository/repository.go
KollaAdithya: I think we can make the `URI()` to private as we are not using this anywhere else.

internal/pkg/cli/deploy/workload.go
KollaAdithya: Even though we are currently printing the build label over here. In the next [PR](https://github.com/aws/copilot-cli/pull/4760/files#diff-916cf5cf0534ab83ab7ea03b7cb954b59ef10e85c31505cc577bad83b7d35701R422), this build label will passed to [LabeledSyncBuffer](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/term/syncbuffer/syncbuffer.go#L53) struct and then Print function will handle the logging.

Lou1415926: I see. Can you add `// TODO(adi) handle this log in syncBuffer's Print function` here? Personally, I found this useful for two reasons:
1. Remind us that we still need to do this
2. In the next PR, the reviewer will see this "TODO" being removed, along with "handle this log in syncBuffer's Print function". This gives them some background information of the relation between that PR and this PR.

Lou1415926 (approved):
Thank you!

Issue #2885, opened 2021-09-30
enable import of default VPC


huanjani:
When customers sign up for an AWS account, they get a [default VPC](https://docs.aws.amazon.com/vpc/latest/userguide/default-vpc.html#default-vpc-components) for quick start-up. 

Currently, Copilot requires two private subnets VPCs. Without two public subnets, load balancing won't work, but we allow users to proceed without public subnets.

To reduce barriers to entry and adoption, Copilot should not require two private subnets, so folks may use their default VPC and try a demo app or simple service that is public-facing. 

Ideas for increasing subnet flexibility:
- [ ] reverse the current UI by selecting all subnets by default 
- [ ] change the assignment of public/private for simplicity? how?
- [ ] error if user doesn't have at least two subnets
- [ ] warn if user doesn't have two private subnets (no backend services)
- [ ] warn if user doesn't have two public subnets (no load-balanced web services)
- [ ] still require at least two of either kind, for AZ placement best practices
- [ ] with `svc package`, error if env doesn't have what the service needs based on workload type/network placement

mreferre:
> warn if user doesn't have two public subnets (no load-balanced web services)

~~But users could setup a LB web service on backend subnets can't they? It's not the default but if they set the flag to deploy LB web services in private subnets it shouldn't fail should it?~~

huanjani:
I think it will fail, because "If your load balancer is an internet-facing load balancer, you must select public subnets in order for your back-end instances to receive traffic from the load balancer (even if the back-end instances are in private subnets)." https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-manage-subnets.html

huanjani:
Copilot will proceed, but the Service won't build with the error: `Load Balancer failed to create: At least two subnets in two different Availability Zones must be specified`

mreferre:
You are right. [ insert GIF of Bart Simpson disappearing into a hedge ]

efekarakus:
This is all done!! The feature will go out next release https://github.com/aws/copilot-cli/pull/3285

efekarakus:
Hi folks! The fix is now released https://github.com/aws/copilot-cli/releases/tag/v1.16.0 🥳 

See announcement: https://aws.github.io/copilot-cli/blogs/release-v116/

gabrieljoelc:
[...]

PR #4758, opened 2023-04-13
chore(docs): make explicit option to import 2 public and 2 private subnets

huanjani:
Addresses https://github.com/aws/copilot-cli/issues/2885#issuecomment-1506466552.


dannyrandall (approved):

bvtujo (approved):
What happens if someone specifies a VPC with more than two subnets?

huanjani:
> What happens if someone specifies a VPC with more than two subnets?

They select just two of each type even if there are more available.

PR #4760, opened 2023-04-13
chore: concurrent image build with context cancellation

KollaAdithya:
Finally with this PR we can build container images in parallel.


internal/pkg/cli/deploy/workload.go
```diff
@@ -374,22 +388,85 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	if err != nil {
 		return fmt.Errorf("login to image repository: %w", err)
 	}
+
+	var mu sync.Mutex
 	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
+
+	// counter for indexing the labeled output buffers.
+	count := 0
```
paragbhingre, line 396: Isn't `count` too generic variable here? I am just saying because it is being used later down in the code and not around where it has been initialized.

paragbhingre: Isn't `count` too generic variable here? I am just saying because it is being used later down in the code and not around where it has been initialized.

internal/pkg/cli/deploy/workload.go
```diff
@@ -374,22 +388,85 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+	// An array of buffers, one for each container image build and push ouput.
```
paragbhingre:
```suggestion
	// An array of buffers, one for each container image build and push output.
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -374,22 +388,85 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
-		digest, err := d.repository.BuildAndPush(buildArgs)
-		if err != nil {
-			return fmt.Errorf("build and push image: %w", err)
-		}
-		out.ImageDigests[name] = ContainerImageIdentifier{
-			Digest:            digest,
-			CustomTag:         d.image.CustomTag,
-			GitShortCommitTag: d.image.GitShortCommitTag,
+		buf := syncbuffer.New()
+		labeledBuffer := buf.WithLabel(fmt.Sprintf("Building your container image: docker %s\n", strings.Join(buildArgsList, " ")))
```
paragbhingre: Just a thought 💭
How about adding a container name here so that customer can see which container we are talking about here. I mean I am sure this will disappear within few seconds as soon as we have new logs. You can decide it better

internal/pkg/cli/deploy/workload.go
```diff
@@ -374,22 +388,85 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+		labeledBuffers[count] = labeledBuffer
+		count++
```
paragbhingre:
```suggestion
		labeledBuffers[count++] = labeledBuffer
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -374,22 +388,85 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+				Digest:            digest,
+				CustomTag:         d.image.CustomTag,
+				GitShortCommitTag: d.image.GitShortCommitTag,
+			}
+			mu.Unlock()
+			return nil
+		})
+		g.Go(func() error {
+			if err := buf.Copy(pr); err != nil {
+				return fmt.Errorf("copying build and push output for container %s: %w", name, err)
```
paragbhingre: I like that we show which container we are talking about ❤️ . We should maximize this practice in this case so that in every step customers know what container logs they are looking at.

internal/pkg/cli/deploy/workload_test.go
```diff
@@ -606,6 +609,9 @@ func TestWorkloadDeployer_UploadArtifacts(t *testing.T) {
 					return nil, nil
 				}
 			}
+			// mockTerminalWidth := func(fw syncbuffer.FileWriter) (int, error) {
+			// 	return 80, nil
+			// }
```
paragbhingre: nit:
```suggestion
```

paragbhingre (commented):
This is really cool 😎

internal/pkg/cli/task_run.go
```diff
@@ -965,12 +968,18 @@ func (o *runTaskOpts) buildAndPushImage() error {
...
 		Dockerfile: o.dockerfilePath,
 		Context:    ctx,
 		Tags:       append([]string{imageTagLatest}, additionalTags...),
-	}); err != nil {
+	}
+	buildArgsList, err := buildArgs.GenerateDockerBuildArgs(dockerengine.New(exec.NewCmd()))
+	if err != nil {
+		return fmt.Errorf("generate docker build args : %w", err)
+	}
+	log.Infof(fmt.Sprintf("Building your container image: docker %s\n", strings.Join(buildArgsList, " ")))
```
paragbhingre: Do we need to remove `\n` from here as well?

internal/pkg/cli/deploy/workload.go
KollaAdithya: The label will be there. It will not disappear. But i will add the name of the container 👍

internal/pkg/cli/task_run.go
KollaAdithya: We do need the new line in `task run`. In case of `workload`  if we have `\n`, `\n` will be counted as new line then `EraseLinesAbove` will not erase the new line in the label. So `\n` is not needed in workload

internal/pkg/cli/deploy/workload.go
KollaAdithya: is postfix increment possible 🤔

dannyrandall: i'll be honest i always forget what the order of operations is when it looks like this - can we leave it on two lines 🥺

internal/pkg/cli/deploy/workload.go
```diff
@@ -365,22 +388,90 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+			}
+			return nil
+		})
+	}
+	if isCIEnvironment() {
+		numLinesForBuildAndPush = -1
+	}
+	// create a LabeledTermPrinter for rendering build and push output.
+	ltp := syncbuffer.NewLabeledTermPrinter(os.Stderr, labeledBuffers, syncbuffer.WithNumLines(numLinesForBuildAndPush), syncbuffer.WithPadding(paddingForBuildAndPush))
```
Lou1415926: What do you think of this
```go
// In cli/deploy/workload.go
type termPrinter interface { // And use make gen-mocks to generate a mock for unit testing
  IsDone() bool
  Print() error
}
type workloadDeployer struct {
  newTermPrinter (fw FileWriter, bufs []*LabeledSyncBuffer, opts ...LabeledTermPrinterOption) termPrinter
}
func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
    return &workloadDeployer{
        newTermPrinter: syncbuffer.NewLabeledTermPrinter,
        }
    }
}

   //...
   ltp := d.newTermPrinter(os.Stderr, labeledBuffers, syncbuffer.WithNumLines(numLinesForBuildAndPush), syncbuffer.WithPadding(paddingForBuildAndPush))
   //...
}

// In cli/deploy/workload_test.go
type deployMocks struct {
   //...
  mockTermPrinter *mocks.mockTermPrinter
}

func TestWorkloadDeployer_UploadArtifacts(t *testing.T) {
     m := &deployMocks{
		mockTermPrinter: mocks.NewMockTermPrinter(ctril)
      }
      m.mockTermPrinter.Expect().Print().Return(nil)
      wkldDeployer := &workloadDeployer{
           newTermPrinter: func(fw FileWriter, bufs []*LabeledSyncBuffer, opts ...LabeledTermPrinterOption) termPrinter {
            return m.mockTermPrinter
          }
       }
}
```

The reason why it is so hard to write unit test for `UploadArtifacts` is that the constructor for term printer is currently called within `uploadContainerImages()`. However, in order to properly test `UploadArtifacts`, we need to sub the labeled term printer with a mock. The solution is to mock the constructor function call - in our case  `newTermPrinter`.

internal/pkg/cli/deploy/workload.go
```diff
@@ -56,6 +62,14 @@ const (
 	labelForVersion       = "com.aws.copilot.image.version"
 	labelForContainerName = "com.aws.copilot.image.container.name"
 )
+const (
+	paddingForBuildAndPush      = 5
```
efekarakus: nit: I'm trying to think how we can hint that this is number of spaces, maybe this?
```suggestion
	paddingInSpacesForBuildAndPush      = 5
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +392,86 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	if err != nil {
 		return fmt.Errorf("login to image repository: %w", err)
 	}
+
+	var mu sync.Mutex
 	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
+
+	// counter for indexing the labeled output buffers.
```
efekarakus: can we delete all of these comments? 🥺  they don't add additional clarity to the code.

See this: https://web.stanford.edu/~ouster/cgi-bin/cs190-winter18/lecture.php?topic=comments

In general his advice on _implementation_ documentation is to add _what_ and _why_ comments only if it's not obvious what the code is doing. Comments should be written at a different _precision_ level than the code itself, for example the comment can be more abstract or explain way more in detail what's going on. In this situation, the comments are the same level of precision as the code.

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +392,86 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+	cursor := cursor.New()
+	cursor.Hide()
+
 	for name, buildArgs := range buildArgsPerContainer {
+		// create a copy of loop variables to avoid data race.
+		name := name
+		buildArgs := buildArgs
+
+		// create a pipe for streaming the build and push output.
+		pr, pw := io.Pipe()
```
efekarakus: nit: can we declare this right before the `g.Go` calls? (moving the variable closer to where it's used)

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +392,86 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+
+	// create a context and an error group.
+	g, ctx := errgroup.WithContext(context.Background())
+
+	// create a new cursor and hide it.
+	cursor := cursor.New()
+	cursor.Hide()
+
 	for name, buildArgs := range buildArgsPerContainer {
+		// create a copy of loop variables to avoid data race.
```
efekarakus, line 399: only this comment is worthwhile retaining here as it's an unusual situation.

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +392,86 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	if err != nil {
 		return fmt.Errorf("login to image repository: %w", err)
 	}
+
+	var mu sync.Mutex
```
```suggestion
	var digestsMu sync.Mutex
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +392,86 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+			out.ImageDigests[name] = ContainerImageIdentifier{
+				Digest:            digest,
+				CustomTag:         d.image.CustomTag,
+				GitShortCommitTag: d.image.GitShortCommitTag,
+			}
+			mu.Unlock()
+			return nil
+		})
+		g.Go(func() error {
+			if err := buf.Copy(pr); err != nil {
```
efekarakus, line 427: interesting, if I'm understanding correctly `buf.Copy` will block (because of `io.Copy`) until `pw.Close()` is invoked right?

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +392,86 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+				return fmt.Errorf("copying build and push output for container %s: %w", name, err)
+			}
+			return nil
+		})
+	}
+	if isCIEnvironment() {
+		numLinesForBuildAndPush = -1
+	}
```
efekarakus: this is a magic number, where the client pkg `deploy` needs to know internal specification of the `syncbuffer` pkg (ideally, they should be decoupled),  consider a different layout like this for setting up options:
```go
opts := []syncbuffer.LabeledTermPrinterOption{syncbuffer.WithPadding(paddingForBuildAndPush)}
if !isCIEnvironment() {
   opts = append(opts, syncbuffer.WithNumLines(numLinesForBuildAndPush))
}
```
so if there is no explicit numLines passed then we default to -1 internally in the constructor of `syncbuffer.NewLabeledTermPrinter`

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +392,86 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+	// create a LabeledTermPrinter for rendering build and push output.
+	ltp := d.labeledTermPrinter(os.Stderr, labeledBuffers, syncbuffer.WithNumLines(numLinesForBuildAndPush), syncbuffer.WithPadding(paddingForBuildAndPush))
+
+	g.Go(func() error {
+		for {
+			if ltp.IsDone() {
+				break
```
efekarakus: once it's done, you might need one last `ltp.Print()` otherwise it's possible that it's over but we haven't printed our last view.

internal/pkg/cli/deploy/workload.go
```diff
@@ -434,6 +521,15 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
 	return dArgs, nil
 }
 
+// isCIEnvironment checks if the current environment is a continuous integration (CI) system.
+// Returns true by looking for the "CI" environment variable  if it's set to "true", otherwise false.
+func isCIEnvironment() bool {
+	if ci, _ := os.LookupEnv("CI"); ci == "true" {
+		return true
+	}
+	return false
```
efekarakus:
```suggestion
	ci, _ := os.LookupEnv("CI"); 
	return ci == "true" 
```

internal/pkg/cli/deploy/workload.go
KollaAdithya: yeah that's true. When `buf.Copy ` is called, it will block until `pw.Close()` is invoked, at which point it will return the final error (if any) encountered during the copy. So `pw.Close()` is used to signal the end of the build process and allow the copy process to finish.

efekarakus: gotcha! ok so the internal bytes.Buffer in the syncbuffer does get populated while `Copy` is still running that's why when we call `buf.lines()` it's able to return to us partial results.

This makes me think:
- Is there a data race when `buf.Copy` is happening (because of `io.Copy` that's writing to the bytes.Buffer) in one Go routine and from another Go routine we are calling `lines()` to read from the same underlying bytes.Buffer?

It might be interesting to chat w @dannyrandall on it

efekarakus: I think we want something like this to be concurrency-safe:
https://github.com/aws/copilot-cli/blob/0ea2c4ba4b01dd72c74117803e5f9ea7c7c930d2/internal/pkg/term/syncbuffer/syncbuffer.go#L22-L27

```go
 // New creates and returns a new SyncBuffer object with an initialized 'done' channel. 
 func New() *SyncBuffer { 
        pr, pw := io.Pipe()
 	return &SyncBuffer{ 
 		done: make(chan struct{}), 
                r: pr,
                w: pw
 	} 
 } 
 
func (buf *SyncBuffer) Copy(r io.Reader) error {
	defer buf.MarkDone()
	_, err := io.Copy(&buf.w, r)
	if err != nil && !errors.Is(err, io.EOF) {
		return err
	}
	return nil
}

func (buf *SyncBuffer) lines() []string {
	lines := strings.NewReader(buf.r).Strings()
	if len(lines) == 0 {
		return nil
	}
	return splitLinesAndTrimSpaces(lines)
}
```

internal/pkg/repository/repository_test.go
```diff
@@ -52,24 +55,23 @@ func TestRepository_BuildAndPush(t *testing.T) {
 				m.EXPECT().Auth().Return("", "", nil).AnyTimes()
 			},
 			inMockDocker: func(m *mocks.MockContainerLoginBuildPusher) {
-				m.EXPECT().Build(&defaultDockerArguments).Return(errors.New("error building image"))
-				m.EXPECT().Push(gomock.Any(), gomock.Any(), gomock.Any()).Times(0)
```
Lou1415926, line 56: nit: I think we can keep this statement to be clear that we explicitly do not want to invoke `Push` here when the `Build` already failed.

internal/pkg/cli/deploy/workload.go
dannyrandall: @efekarakus is right - I didn't realize we added our own implementation of `Copy()` in the `syncbuffer` package. We should remove that function and just call `io.Copy(buf, pr)` here! Reasoning:

`io.Copy` [just loops calls to `Read()` and `Write()`](https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/io/io.go;l=405-454;drc=dc8e2a6a8ec94f2c98ba20edd57932eba284efb1). since we have our [own implementation of `Write()`](https://github.com/aws/copilot-cli/blob/ad57838bc4a93d0bd5057294db95a59a911b7a12/internal/pkg/term/syncbuffer/syncbuffer.go#L30-L34) that is thread safe, we should just let `io.Copy` use it!

dannyrandall: if we want to keep the `Copy()` function so we can have the `defer buf.MarkDone()`, my recommendation is slightly different than efe's - just pass `buf` directly into the `io.Copy()` call instead of `buf.buf`:

```
func (buf *SyncBuffer) Copy(r io.Reader) error {
  defer buf.MarkDone()
  _, err := io.Copy(buf, r)
  // etc
}
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -56,6 +62,14 @@ const (
 	labelForVersion       = "com.aws.copilot.image.version"
 	labelForContainerName = "com.aws.copilot.image.container.name"
 )
+const (
+	paddingInSpacesForBuildAndPush = 5
+	pollIntervalForBuildAndPush    = 60 * time.Millisecond
+)
+
+var (
+	numLinesForBuildAndPush = 5
```
Lou1415926: is this `var` because it will get modified? If so, can we do the following to avoid editable global vars:
```go
const defaultNumLinesForBuildAndPush = 5

func () { // In the function where `numLinesForBuildAndPush` will potentially be modified
  numLinesForBuildAndPush := defaultNumLinesForBuildAndPush
} 
```

internal/pkg/cli/deploy/workload.go
KollaAdithya: I already made that change  over here.
https://github.com/aws/copilot-cli/pull/4760/files#diff-b12c92294403172e636dd4c441963a6ff7d684f316c919f6a491f76d37a65a37R68

If I am directly writing to  internal bytes.Buffer(which is not thread safe) something like this  `_, err := io.Copy(&buf.buf, r)`. I got warning of Datarace. So I changed it to `_, err := io.Copy(buf, r)`

internal/pkg/cli/deploy/workload.go
```diff
@@ -275,6 +297,9 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 		envSess:                  envSession,
 		store:                    store,
 		envConfig:                envConfig,
+		labeledTermPrinter: func(fw syncbuffer.FileWriter, bufs []*syncbuffer.LabeledSyncBuffer, opts ...syncbuffer.LabeledTermPrinterOption) labeledTermPrinter {
```
Lou1415926: nit:
```suggestion
		labeledTermPrinter: syncbuffer.NewLabeledTermPrinter,
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -139,6 +158,8 @@ type GenerateCloudFormationTemplateOutput struct {
 	Parameters string
 }
 
+type labeledTermPrinterFunc func(fw syncbuffer.FileWriter, bufs []*syncbuffer.LabeledSyncBuffer, opts ...syncbuffer.LabeledTermPrinterOption) labeledTermPrinter
```
Lou1415926: Is there a specific reason why we want to make the function a `Named` type?

internal/pkg/cli/deploy/workload.go
efekarakus: Very nice!

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +392,74 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
 		buildArgsList, err := buildArgs.GenerateDockerBuildArgs(dockerengine.New(exec.NewCmd()))
 		if err != nil {
-			return fmt.Errorf("generate docker build args: %w", err)
-		}
-		// TODO(adi) handle this log in syncBuffer's Print function
-		log.Infof("Building your container image: docker %s\n", strings.Join(buildArgsList, " "))
-		digest, err := d.repository.BuildAndPush(buildArgs)
-		if err != nil {
-			return fmt.Errorf("build and push image: %w", err)
+			return fmt.Errorf("generate docker build args of image %s: %w", name, err)
```
Lou1415926: nit:
```suggestion
			return fmt.Errorf("generate docker build args for %q's image: %w", name, err)
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +392,74 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+		buf := syncbuffer.New()
+		labeledBuffer := buf.WithLabel(fmt.Sprintf("Building your container image %s: docker %s", name, strings.Join(buildArgsList, " ")))
+		labeledBuffers[bufferIdx] = labeledBuffer
+		bufferIdx++
+		pr, pw := io.Pipe()
+		g.Go(func() error {
+			defer pw.Close()
+			digest, err := d.repository.BuildAndPush(ctx, buildArgs, pw)
+			if err != nil {
+				return fmt.Errorf("build and push image %s: %w", name, err)
```
Lou1415926: nit:
```suggestion
				return fmt.Errorf("build and push the image for %q container: %w", name, err)
```

internal/pkg/cli/deploy/workload.go
KollaAdithya: @efekarakus I have a concern with the current logic of `printall`. 
let say we have 3 container images. In case of `CI=true` We are expecting to print the complete output of the images `only 3 times` one  time for each image. 

let us say if one buffer is done and the remaining buffers takes some additional amount of time to be done. 

In this case, With the current logic of `printall` We are continuously printing the output of the first image buffer more than once until  all buffers are done.

So this is the reason i am thinking of having the `infinite for loop` with in the `syncbuffer` package  for `printall` and change the `printall to Public `  something like below
```
func (ltp *LabeledTermPrinter) PrintAll() {
printedCount := 0
	for {
		for idx, buf := range ltp.buffers {
			if !buf.syncBuf.IsDone() {
				continue
			}
			outputLogs := ltp.buffers[idx].syncBuf.strings()
			ltp.writeLines(buf.label, outputLogs)
			printedCount++
		}
		if printedCount >= len(ltp.buffers) {.  ----------> this way ensure we print only once from each buffer.
			break
		}
	}	
}
```	
With this approach I think I need to remove the logic of magic number `-1` for printing the entire content of the buffer. 
🤔  please let me know if this sounds good.

internal/pkg/cli/deploy/workload.go
KollaAdithya: removed the `named` type and declared in the `workload deployer`

internal/pkg/term/syncbuffer/termprinter.go
```diff
@@ -108,14 +108,16 @@ func (ltp *LabeledTermPrinter) Print() error {
...
+		if !ltp.buffers[idx].IsDone() {
 			continue
 		}
 		outputLogs := ltp.buffers[idx].lines()
-		if _, err := ltp.writeLines(buf.label, outputLogs); err != nil {
+		if _, err := ltp.writeLines(ltp.buffers[idx].label, outputLogs); err != nil {
 			return fmt.Errorf("write logs: %w", err)
 		}
+		ltp.buffers = append(ltp.buffers[:idx], ltp.buffers[idx+1:]...)
+		idx--
```
efekarakus, line 120: huh I've never seen this way of doing deleting from the slice (keeping the index in its place), interesting!

- Can we add a testcase in `TestLabeledTermPrinter_Print` that makes sure that multiple calls to `Print` results in only once printing a buffer to `ltp.term` when it enters `printAll`?

internal/pkg/cli/deploy/workload.go
KollaAdithya: This will be a`IncompatibleAssign` 🤔 . if I directly use `syncbuffer.NewLabeledTermPrinter`  because the the function signature of the `labeledTermPrinter` field in workloadDeployer struct does not match the signature of `syncbuffer.NewLabeledTermPrinter` function.

internal/pkg/term/syncbuffer/termprinter_test.go
```diff
@@ -90,6 +90,11 @@ line4 from image2`)
 				buf.MarkDone()
 			}
 			ltp.Print()
+			if ltp.numLines == -1 {
```
efekarakus: nit: can you add a comment here explaining why you're doing this :)

Instead of `ltp.numLines == -1` can we add a bool into the testcase that's like `t.isPrintAll` and then do this for-loop. I'm mentioning this because we can always change the "magic number" from -1 to something else and then the test would stop working.

efekarakus (approved):
:shipit: with a small tiny test adjustment request

internal/pkg/cli/deploy/workload.go
KollaAdithya: To address this concern
Modified `printall` function such that if the entire contents of a single buffer are printed, then remove that buffer from the buffers slice.

internal/pkg/cli/deploy/workload.go
```diff
@@ -275,6 +295,7 @@ func newWorkloadDeployer(in *WorkloadDeployerInput) (*workloadDeployer, error) {
 		envSess:                  envSession,
 		store:                    store,
 		envConfig:                envConfig,
+		labeledTermPrinter:       labeledTermPrinter,
```
dannyrandall, line 298:
```suggestion
		labeledTermPrinter:       syncbuffer.NewLabeledTermPrinter,
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -434,6 +502,13 @@ func buildArgsPerContainer(name, workspacePath string, img ContainerImageIdentif
 	return dArgs, nil
 }
 
+// isCIEnvironment checks if the current environment is a continuous integration (CI) system.
+// Returns true by looking for the "CI" environment variable  if it's set to "true", otherwise false.
+func isCIEnvironment() bool {
+	ci, _ := os.LookupEnv("CI")
+	return ci == "true"
```
```suggestion
	return os.Getenv("CI") == "true"
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +388,71 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
-		if err != nil {
-			return fmt.Errorf("build and push image: %w", err)
+			return fmt.Errorf("generate docker build args for %q's image: %w", name, err)
 		}
-		out.ImageDigests[name] = ContainerImageIdentifier{
-			Digest:            digest,
-			CustomTag:         d.image.CustomTag,
-			GitShortCommitTag: d.image.GitShortCommitTag,
+		buf := syncbuffer.New()
+		labeledBuffer := buf.WithLabel(fmt.Sprintf("Building your container image %s: docker %s", name, strings.Join(buildArgsList, " ")))
```
```suggestion
		labeledBuffer := buf.WithLabel(fmt.Sprintf("Building your container image %q: docker %s", name, strings.Join(buildArgsList, " ")))
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +388,71 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
 		buildArgsList, err := buildArgs.GenerateDockerBuildArgs(dockerengine.New(exec.NewCmd()))
 		if err != nil {
-			return fmt.Errorf("generate docker build args: %w", err)
-		}
-		// TODO(adi) handle this log in syncBuffer's Print function
-		log.Infof("Building your container image: docker %s\n", strings.Join(buildArgsList, " "))
-		digest, err := d.repository.BuildAndPush(buildArgs)
-		if err != nil {
-			return fmt.Errorf("build and push image: %w", err)
+			return fmt.Errorf("generate docker build args for %q's image: %w", name, err)
```
```suggestion
			return fmt.Errorf("generate docker build args for %q: %w", name, err)
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +388,71 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	if err != nil {
 		return fmt.Errorf("login to image repository: %w", err)
 	}
+
+	var digestsMu sync.Mutex
 	out.ImageDigests = make(map[string]ContainerImageIdentifier, len(buildArgsPerContainer))
+	bufferIdx := 0
+	labeledBuffers := make([]*syncbuffer.LabeledSyncBuffer, len(buildArgsPerContainer))
```
dannyrandall: nit: i'd rather get rid of the index and just use `append` - not worried about performance as this is a one time thing and it's a little more readable as there's already a lot of local variables here

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +388,71 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+				Digest:            digest,
+				CustomTag:         d.image.CustomTag,
+				GitShortCommitTag: d.image.GitShortCommitTag,
+			}
+			digestsMu.Unlock()
+			return nil
+		})
+		g.Go(func() error {
+			if err := buf.Copy(pr); err != nil {
+				return fmt.Errorf("copying build and push output for %q container: %w", name, err)
```
```suggestion
				return fmt.Errorf("copy build and push output for %q: %w", name, err)
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +388,71 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+		buf := syncbuffer.New()
+		labeledBuffer := buf.WithLabel(fmt.Sprintf("Building your container image %s: docker %s", name, strings.Join(buildArgsList, " ")))
+		labeledBuffers[bufferIdx] = labeledBuffer
+		bufferIdx++
+		pr, pw := io.Pipe()
+		g.Go(func() error {
+			defer pw.Close()
+			digest, err := d.repository.BuildAndPush(ctx, buildArgs, pw)
+			if err != nil {
+				return fmt.Errorf("build and push the image for %q container: %w", name, err)
```
```suggestion
				return fmt.Errorf("build and push the image %q: %w", name, err)
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +388,71 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+			}
+			digestsMu.Lock()
```
dannyrandall, line 417:
```suggestion
			digestsMu.Lock()
			defer digestsMu.Unlock()
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +388,71 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+	}
+	opts := []syncbuffer.LabeledTermPrinterOption{syncbuffer.WithPadding(paddingInSpacesForBuildAndPush)}
+	if !isCIEnvironment() {
+		opts = append(opts, syncbuffer.WithNumLines(defaultNumLinesForBuildAndPush))
+	}
+	ltp := d.labeledTermPrinter(os.Stderr, labeledBuffers, opts...)
+	g.Go(func() error {
+		for {
+			if err := ltp.Print(); err != nil {
+				return fmt.Errorf("printing logs of docker build and push: %w", err)
```
```suggestion
				return fmt.Errorf("print logs: %w", err)
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +388,71 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+			}
+			if ltp.IsDone() {
+				break
```
```suggestion
				return nil
```

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -20,6 +22,7 @@ import (
 // Cmd is the interface implemented by external commands.
 type Cmd interface {
 	Run(name string, args []string, options ...exec.CmdOption) error
+	RunWithContext(ctx context.Context, name string, args []string, opts ...exec.CmdOption) error
```
dannyrandall, line 25: !!! 🎉

internal/pkg/cli/deploy/workload.go
KollaAdithya: is it good to unlock the map before `return nil`? I mean if I use `defer` then i will unlock after `return ` is executed.

internal/pkg/cli/deploy/workload.go
KollaAdithya: If I directly use `syncbuffer.NewLabeledTermPrinter`  this is giving me an error of `IncompatibleAssign` 🤔 . because the the function signature of the `labeledTermPrinter field` in workloadDeployer struct does not match the signature of `syncbuffer.NewLabeledTermPrinter` function.

dannyrandall: ah gotcha, it's bc the signature expects the _interface_ to be returned 👍

internal/pkg/cli/deploy/workload.go
```diff
@@ -367,24 +388,66 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
+	ltp := d.labeledTermPrinter(os.Stderr, labeledBuffers, opts...)
+	g.Go(func() error {
+		for {
+			if err := ltp.Print(); err != nil {
+				return fmt.Errorf("print logs: %w", err)
+			}
+			if ltp.IsDone() {
+				return nil
+			}
+			time.Sleep(pollIntervalForBuildAndPush)
```
dannyrandall, line 446:
```suggestion
    select {
      case <-ctx.Done():
        return nil
      case <-time.After(pollIntervalForBuildAndPush):
        // loop again
   }
```

so this function stops if ctx is cancelled!

dannyrandall (approved):
looks great! good work ☺️🍾

Issue #4621, opened 2023-03-13
[Feature Design] Copilot Static Site pattern


iamhopaul123:
# New service type “Static Site”
We’ll create a new service type called “Static Site”. And then we’ll have the following commands for it.

* svc init
* svc package
* svc deploy
* svc ls/show/status/logs (operational commands)

```
$ copilot svc init

  Which service type best represents your service's architecture?  [Use arrows to move, type to filter, ? for more help]
  > Request-Driven Web Service  (App Runner)
    Load Balanced Web Service   (Internet to ECS on Fargate)
    Backend Service             (ECS on Fargate)
    Worker Service              (Events to SQS to ECS on Fargate)
    Static Site.                (S3 bucket)
```

## Upload experience
Users can specify what files to be uploaded in their service manifests. For example:

```yaml
name: blob
type: Static Site

files:
  - source: ./foo # Required
    destination: /code # Optional default is "/"
    # Optional, whether the sub dir will be visited. By default it is false.
    recursive: true 
    # by default, *all files are included*. 
    # This means that providing *only* an --include filter will not change what files are transferred.
    # --include will only re-include files that have been excluded from an --exclude filter. 
    # If you only want to upload files with a particular extension, you need to first exclude all files,
    # then re-include the files with the particular extension.
    reinclude: *.jpg
    exclude: *
  - context: ./frontend # Optional. Context for this upload.
    source: dist # will be resolved as ./frontend/dist 
    destination: /dist
```

## Integrate with CloudFront
The experience will be mostly the same as [Copilot CloudFront Integration](https://github.com/aws/copilot-cli/issues/3701). In env manifest, users could specify the “cdn” field to connect CF with the S3 bucket.

```yaml
cdn:
  static:
    alias: example.com # Required
    path: static # Default is *
    location: blob # Required
```

Note that, behind the scene we’ll add proper [bucket policy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-policy.html) so that it can be accessed by the [CloudFront access identity](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudfront-cloudfrontoriginaccessidentity.html).

### svc ls
Users can use copilot svc ls to list all the created static pattern in a certain application

```
$ copilot svc ls

Name                Type
----                ----
blob                Static Site
frontend            Load Balanced Web Service
```

### svc show
svc show shows the configuration of the static pattern:

```
[...]

PR #4761, opened 2023-04-13
chore(cli): add static site source selector

huanjani:
Related: #4621.

Based on feedback, I opted to use a multi selector that includes an option to filter with typing. Tab-completion is a possible enhancement for the future. Also per feedback, only the required field (`Source`) is prompted for; the remaining fields can be input via the workload manifest.

This selector allows for (multiple) typed-in paths, which are validated against the local filesystem for existence. 

```
% copilot svc init
Note: It's best to run this command in the root of your workspace.
Service type: Static Site
Service name: demo
Source(s): ./index.html, ./main.go, Enter custom path to your static site dir/file
Custom Path to Source: demo-app
Another: Yes
X Sorry, your reply was invalid: value must be a valid path
Custom Path to Source: copilot
Another: No
...
```


internal/pkg/cli/svc_init.go
```diff
@@ -91,6 +91,11 @@ These messages can be consumed by the Worker Service.`
 "Internet" will configure your service as public.`
 
 	wkldInitImagePrompt = fmt.Sprintf("What's the %s ([registry/]repository[:tag|@digest]) of the image to use?", color.Emphasize("location"))
+
+	fmtStaticSiteInitDirFilePrompt      = "Which " + color.Emphasize("files or (recursive) directories") + " would you like to upload for %s?\nIt is unnecessary to select files or subdirectories within selected directories."
```
dannyrandall: nit
```suggestion
	fmtStaticSiteInitDirFilePrompt      = "Which " + color.Emphasize("files or directories") + " would you like to upload for %s?\nDirectories are recursively uploaded."
```

internal/pkg/cli/svc_init.go
```diff
@@ -430,7 +438,26 @@ If you'd prefer a new default manifest, please manually delete the existing one.
...
+		if err != nil {
+			return err
+		}
+		if isDir {
+			recursive = true
+		}
+		assets = append(assets, manifest.FileUpload{
+			Source:    source,
+			Recursive: recursive,
+		})
```
```suggestion
		assets = append(assets, manifest.FileUpload{
			Source:    source,
			Recursive: isDir,
		})
```

internal/pkg/cli/svc_init.go
```diff
@@ -474,7 +501,7 @@ func (o *initSvcOpts) validateIngressType() error {
 	if strings.EqualFold(o.ingressType, "internet") || strings.EqualFold(o.ingressType, "environment") {
 		return nil
 	}
-	return fmt.Errorf("invalid ingress type %q: must be one of %s.", o.ingressType, english.OxfordWordSeries(rdwsIngressOptions, "or"))
+	return fmt.Errorf("invalid ingress type %q: must be one of %s", o.ingressType, english.OxfordWordSeries(rdwsIngressOptions, "or"))
```
dannyrandall, line 517: `OxfordWordSeries` ty

internal/pkg/cli/cli.go
```diff
@@ -180,3 +180,11 @@ func displayPath(target string) string {
 	}
 	return rel
 }
+
+func isDir(fs afero.Fs, path string) (bool, error) {
```
dannyrandall: does this actually shorten any logic?🥺

internal/pkg/term/selector/local_selector.go
```diff
@@ -145,6 +145,61 @@ func NewLocalFileSelector(prompt Prompter, fs afero.Fs) (*localFileSelector, err
 	}, nil
 }
 
+// StaticSources asks the user to select from a list of directories and files in the current directory and two levels down.
```
dannyrandall, line 151: 💭 i wonder if we should prompt for directories based on the workspace dir? (git root)? just bc that's how it goes into the manifest file? i'm not sure which is better tbh

internal/pkg/term/selector/local_selector.go
```diff
@@ -219,6 +274,52 @@ func (s *localFileSelector) listDockerfiles() ([]string, error) {
...
+	}
+	var filesAndDirs []string
+	for _, firstLevel := range wdDirsAndFiles {
+		name := firstLevel.Name()
+		if strings.HasPrefix(name, ".") || name == "copilot" {
+			continue
+		}
+		filesAndDirs = append(filesAndDirs, filepath.Dir(name)+"/"+name)
+		if firstLevel.IsDir() {
+			subFiles, err := s.fs.ReadDir(name)
```
dannyrandall: i wonder if we could refactor `listDirsAndFiles()` so that we could recursively call the same function twice instead of duplicating logic? something like:
```
listDirsAndFiles(depth int) ([]string, error) {
  var out []string
  for wdDirsAndFiles {
    if depth < 3 && isDir {
       out = append(listDirsAndFiles(depth+1))
    }
  }
}
```
(very pseudocode, hopefully the idea makes sense!)

internal/pkg/term/selector/local_selector.go
```diff
@@ -145,6 +145,61 @@ func NewLocalFileSelector(prompt Prompter, fs afero.Fs) (*localFileSelector, err
...
+	if len(dirsAndFiles) == 0 {
+		log.Warningln("No directories or files were found in the current working directory. Enter a relative path with the 'custom path' option, or run this command from higher up in your local file directory.")
+	}
+	dirsAndFiles = append(dirsAndFiles, []string{staticSourceUseCustomPrompt}...)
+	var selections []string
+	selections, err = s.prompt.MultiSelect(
+		selPrompt,
+		selHelp,
+		dirsAndFiles,
+		func(v interface{}) error {
```
Lou1415926: Do you know if we can pass in a `null` validator when no validation is needed? iirc it's okay!

internal/pkg/term/selector/local_selector.go
```diff
@@ -145,6 +145,61 @@ func NewLocalFileSelector(prompt Prompter, fs afero.Fs) (*localFileSelector, err
...
+			return nil
+		},
+		prompt.WithFinalMessage(staticAssetsFinalMsg),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("select directories and/or files: %w", err)
+	}
+	for i, selection := range selections {
+		anotherCustomPath := true
```
Lou1415926: Instead of initializing `anotherCustomPath` with `true`, 
```go
for i, selection := range selections {
    if selection == staticSourceUseCustomPrompt {
        for {
            // get custom path.
            anotherCustomPath, err = s.prompt.Confirm()
	    if err != nil {}
            if !anotherCustomPath {
                break
            }
        }
    }
}
```

internal/pkg/term/selector/local_selector.go
```diff
@@ -145,6 +145,61 @@ func NewLocalFileSelector(prompt Prompter, fs afero.Fs) (*localFileSelector, err
...
+					prompt.WithFinalMessage(customPathFinalMsg))
+				if err != nil {
+					return nil, fmt.Errorf("get custom directory or file path: %w", err)
+				}
+				if selection == staticSourceUseCustomPrompt {
+					selections[i] = customPath // The first custom path replaces the prompt string.
+					selection = customPath
+				} else {
+					selections = append(selections, customPath) // Subsequent custom paths are appended.
+				}
```
Lou1415926: The way we are manipulating & examining `selection`  is not super straightforward I think 🤔  I have an alternative proposal to the codes here 💭  not a huge fan of ⬇️ either, but sharing anyway!
```go
addCustomIndex := -1
for i, selection := range selections {
  if selection == statisSourceUseCustomPrompt {
    addCustomIndex = i
    break
  }
}

 if addCustomIndex == -1 {
    return selections
 }

 selections = append(selections[:addCustomIndex], selections[addCustomIndex+1:]...)
 for {
    customPath, err := s.prompt.Get()
    paths = append(selections, customPath)
    another, err := s.prompt.Confirm()
    if !another {
      break
    }
  }
```

internal/pkg/cli/init_test.go
```diff
@@ -168,6 +168,10 @@ func TestInitOpts_Run(t *testing.T) {
 						Value: manifestinfo.WorkerServiceType,
 						Hint:  "Events to SQS to ECS on Fargate",
 					},
+					{
+						Value: manifestinfo.StaticSiteType,
+						Hint: "Internet to CDN to S3 bucket",
```
Lou1415926, line 173: I might have missed something but I just want to make sure: is there always always a CDN in front?

internal/pkg/cli/interfaces.go
```diff
@@ -532,6 +532,10 @@ type wsSelector interface {
 	Workload(msg, help string) (string, error)
 }
 
+type dirFileSelector interface {
```
Lou1415926: I feel like we can name it `staticSouorceSelector` because of how much `StaticSources` is specifically designed for static site. `StaticSources` doesn't look like a generic selector that any clients that need to select dirs/files would just be able to use it as it.

internal/pkg/cli/svc_init.go
```diff
@@ -91,6 +91,11 @@ These messages can be consumed by the Worker Service.`
 "Internet" will configure your service as public.`
 
 	wkldInitImagePrompt = fmt.Sprintf("What's the %s ([registry/]repository[:tag|@digest]) of the image to use?", color.Emphasize("location"))
+
+	fmtStaticSiteInitDirFilePrompt      = "Which " + color.Emphasize("files or directories") + " would you like to upload for %s?\nDirectories are recursively uploaded."
+	staticSiteInitDirFileHelpPrompt     = "Files or (recursive) directories to use for building your static site."
```
Lou1415926: I can be wrong, but I feel like "recursive" is already implied by "directories", unless it's valid/common to build a static site with an empty dir 🤔

huanjani: My main concern is not wanting users to think they have to select every subdir/file within the target dir. 🤷🏼‍♀️

internal/pkg/cli/init_test.go
iamhopaul123: For now yes 🥺

internal/pkg/cli/svc_init.go
iamhopaul123: i think it's probably ok to remove because when you select a folder in a file selector, usually all files in that folder will be selected by default 🤔 We could adjust accordingly if it is confusing to some customers.

internal/pkg/cli/svc_init.go
```diff
@@ -91,6 +91,11 @@ These messages can be consumed by the Worker Service.`
...
+	fmtStaticSiteInitDirFilePathPrompt  = "What is the custom path to the " + color.Emphasize("directory or file") + " for %s?"
+	staticSiteInitDirFilePathHelpPrompt = "Path to directory or file to use for building your static site."
```
iamhopaul123: is this "notFoundPrompt"? according to `StaticSources(selPrompt, selHelp, notFoundPrompt, notFoundHelp string, pathValidator prompt.ValidatorFunc) ([]string, error)`. Maybe we could make them more consistent.

internal/pkg/cli/svc_init.go
huanjani: But we don't do that in our multi select.... I started writing some code to make it do it, but then realized the right arrow selects all, which works great if the user filters by the dir name. But I'm not sure how to make this functionality more discoverable. 🤔

internal/pkg/term/selector/local_selector.go
huanjani: If they input a custom path, should we require it to be based on the workspace dir, too, or accept either and conditionally translate it after the fact? 🤔

iamhopaul123: I feel like maybe it's ok to limit the input dir to be ws only for now? We could have some logic in manifest validation too and then adjust ad hoc.

internal/pkg/term/selector/local_selector.go
huanjani: done

internal/pkg/cli/svc_init.go
huanjani: I guess you're both right, that it's kind of common sense that it's recursive.

internal/pkg/term/selector/local_selector.go
huanjani: What do you all think of this?
```
 % copilot svc init -t "Static Site" -n djf
Note: It's best to run this command in the root of your workspace.

  Which files or directories would you like to upload for djf?
Directories are recursively uploaded.  [Use arrows to move, space to select, type to filter, ? for more help]
  [ ]  /Users/jwh/ECS-DevX/sample-apps/aws-copilot-sample-service/appspec.yaml
  [ ]  /Users/jwh/ECS-DevX/sample-apps/aws-copilot-sample-service/create-deployment.json
  [ ]  /Users/jwh/ECS-DevX/sample-apps/aws-copilot-sample-service/demo-app
  > [ ]  /Users/jwh/ECS-DevX/sample-apps/aws-copilot-sample-service/demo-app/thisfile.go
  [ ]  /Users/jwh/ECS-DevX/sample-apps/aws-copilot-sample-service/demo-app/thisfolder
  [ ]  /Users/jwh/ECS-DevX/sample-apps/aws-copilot-sample-service/demo-app/thisfolder/anotherfile.go
  [ ]  /Users/jwh/ECS-DevX/sample-apps/aws-copilot-sample-service/index.html
```
I feel like it makes more sense to the user because they don't think in terms of the workspace dir. Then we'll translate it to be relative to the workspace dir before populating it in the manifest.

internal/pkg/cli/svc_init.go
```diff
@@ -430,7 +438,22 @@ If you'd prefer a new default manifest, please manually delete the existing one.
...
-	// TODO: add file selection for generating svc manifest.
+	var assets []manifest.FileUpload
+	sources, err := o.askSource()
+	if err != nil {
+		return err
+	}
+	for _, source := range sources {
+		info, err := o.fs.Stat(source)
+		if err != nil {
+			return fmt.Errorf("get Fileinfo describing %s: %w", source, err)
```
```suggestion
			return fmt.Errorf("get info for %q: %w", source, err)
```

internal/pkg/cli/svc_init.go
```diff
@@ -91,6 +91,11 @@ These messages can be consumed by the Worker Service.`
 "Internet" will configure your service as public.`
 
 	wkldInitImagePrompt = fmt.Sprintf("What's the %s ([registry/]repository[:tag|@digest]) of the image to use?", color.Emphasize("location"))
+
+	fmtStaticSiteInitDirFilePrompt      = "Which " + color.Emphasize("directories or files") + " would you like to upload for %s?"
+	staticSiteInitDirFileHelpPrompt     = "Directories or files to use for building your static site."
+	fmtStaticSiteInitDirFilePathPrompt  = "What is the custom path to the " + color.Emphasize("directory or file") + " for %s?"
```
```suggestion
	fmtStaticSiteInitDirFilePathPrompt  = "What is the path to the " + color.Emphasize("directory or file") + " for %s?"
```

internal/pkg/cli/svc_init.go
```diff
@@ -502,6 +525,22 @@ func (o *initSvcOpts) askImage() error {
...
 }
 
+func (o *initSvcOpts) askSource() ([]string, error) {
+	sources, err := o.sourceSel.StaticSources(
+		fmt.Sprintf(fmtStaticSiteInitDirFilePrompt, color.HighlightUserInput(o.name)),
+		staticSiteInitDirFileHelpPrompt,
+		fmt.Sprintf(fmtStaticSiteInitDirFilePathPrompt, color.HighlightUserInput(o.name)),
+		staticSiteInitDirFilePathHelpPrompt,
+		func(v interface{}) error {
+			return validatePath(afero.NewOsFs(), v)
```
```suggestion
			return validatePath(o.fs), v)
```

internal/pkg/cli/svc_init_test.go
```diff
@@ -115,10 +116,13 @@ func TestSvcInitOpts_Validate(t *testing.T) {
 				m.mockStore.EXPECT().GetApplication("phonetool").Return(&config.Application{}, nil)
 			},
 			mockFileSystem: func(mockFS afero.Fs) {
-				mockFS.MkdirAll("hello", 0755)
+				err := mockFS.MkdirAll("hello", 0755)
+				if err != nil {
+					return 
+				}
```
dannyrandall: prob fine to ignore this error

internal/pkg/term/selector/local_selector.go
```diff
@@ -145,6 +146,61 @@ func NewLocalFileSelector(prompt Prompter, fs afero.Fs) (*localFileSelector, err
...
 }
 
+// StaticSources asks the user to select from a list of directories and files in the current directory and two levels down.
+func (s *localFileSelector) StaticSources(selPrompt, selHelp, anotherPathPrompt, anotherPathHelp string, pathValidator prompt.ValidatorFunc) ([]string, error) {
+	dirsAndFiles, err := s.listDirsAndFiles(s.workingDirAbs)
+	if err != nil {
+		return nil, err
+	}
+	if len(dirsAndFiles) == 0 {
+		log.Warningln("No directories or files were found in the current working directory. Enter a relative path with the 'custom path' option, or run this command from higher up in your local file directory.")
```
dannyrandall:
```suggestion
		log.Warningln("No directories or files were found in your workspace. Enter a relative path with the 'custom path' option if you'd like to use a hidden file.")
```

internal/pkg/term/selector/local_selector.go
```diff
@@ -219,6 +275,42 @@ func (s *localFileSelector) listDockerfiles() ([]string, error) {
...
+		if *depth < 3 && file.IsDir() {
+			*depth += 1
+			_, err := s.getDirAndFileNames(relPathName, depth, existingNames)
+			if err != nil {
+				return nil, fmt.Errorf("get dir and file names: %w", err)
+			}
+		}
+	}
+	return *existingNames, nil
+}
```
dannyrandall:
```suggestion
func (s *localFileSelector) getDirAndFileNames(dir string, depth int) ([]string, error) {
	wdDirsAndFiles, err := s.fs.ReadDir(dir)
	if err != nil {
		return nil, fmt.Errorf("read directory: %w", err)
	}
	var names []string
	for _, file := range wdDirsAndFiles {
		name := file.Name()
		if strings.HasPrefix(name, ".") || name == "copilot" {
			continue
		}
		relPathName := dir + "/" + name
		names = append(names, relPathName)
		if depth < 3 && file.IsDir() {
			subNames, err := s.getDirAndFileNames(relPathName, depth+1)
			if err != nil {
				return nil, fmt.Errorf("get dir and file names: %w", err)
			}
			names = append(names, subNames...)
		}
	}
	return names, nil
}
```
hopefully this makes sense :) i don't think we need the `if` at L293 as dir _is_ `s.workingDirAbs` from L281, but maybe i'm mistaken!

internal/pkg/term/selector/local_selector.go
Lou1415926: > they don't think in terms of the workspace dir.

This is true, but I feel like they do think in terms of a project root. For example, if their folder structure is like:
```
+ aws-copilot-sample-service/
    +- copilot/
          - .workspace
    +- demo-app/
          +- thisfolder
                - anotherfile.go
    - appsec.yml
    - index.html
```
Then I think it would probably make sense to prompt for 
```
[ ] demo-app
[ ] demo-app/thisfolder
[ ] demo-app/thisfolder/anotherfile.go
[ ] appsec.yml
[ ] index.html
```

Because `aws-copilot-sample-service/` is the project root?

huanjani: Can we assume the project root is the `cwd`?

Lou1415926: I think today users are able to run commands that require a workspace, such as `copilot svc deploy` and `copilot svc init`, *not* from the project root, but from dirs like `aws-copilot-sample-service/copilot` or `aws-copilot-sample-service/demo-app`, which are within the project, but not necessarily the root.

Therefore, I think it'd make sense for `copilot svc init` with static site to make the same level of assumptions. What do yall think 🤔 ?

huanjani: That makes sense to me. 👍🏼

internal/pkg/term/selector/local_selector.go
huanjani: @Lou1415926: Are you okay with leaving the code as is for now until we come up with something we're excited about?

internal/pkg/cli/init.go
```diff
@@ -97,6 +97,10 @@ type initOpts struct {
 
 func newInitOpts(vars initVars) (*initOpts, error) {
 	fs := afero.NewOsFs()
+	ws, err := workspace.Use(fs)
```
Lou1415926: oo is there any reason we wanted to move this block?

`workspace.Use()` assumes that there *is* already a local workspace, and errors out if there isn't any. I think it's entirely possible though that `copilot init` is run in a completely new place where a worksplace doesn't exist yet.

internal/pkg/cli/svc_init.go
```diff
@@ -430,7 +438,22 @@ If you'd prefer a new default manifest, please manually delete the existing one.
 }
 
 func (o *initSvcOpts) askStaticSite() error {
-	// TODO: add file selection for generating svc manifest.
+	var assets []manifest.FileUpload
+	sources, err := o.askSource()
```
Lou1415926: ohhh we should have a way to skip the prompt!

internal/pkg/workspace/workspace.go
```diff
@@ -104,7 +105,7 @@ func Use(fs afero.Fs) (*Workspace, error) {
 }
 
 // Create creates a new Workspace in the current working directory for appName with summary if it doesn't already exist.
-func Create(appName string, fs afero.Fs) (*Workspace, error) {
+func Create(appName string, fs afero.Fs) (*Workspace, error) {
```
Lou1415926: nit: so that `git` or something does not print out annoying warnings against trailing whitespace 😂 
```suggestion
func Create(appName string, fs afero.Fs) (*Workspace, error) {
```

internal/pkg/workspace/workspace.go
```diff
@@ -60,13 +61,13 @@ const (
...
 	Application string `yaml:"application"` // Name of the application.
-	Path        string `yaml:"-"`           // absolute path to the summary file.
+	Path        string `yaml:"-"`           // Absolute path to the summary file.
 }
 
 // Workspace typically represents a Git repository where the user has its infrastructure-as-code files as well as source files.
 type Workspace struct {
 	workingDirAbs string
-	copilotDirAbs string
+	CopilotDirAbs string
```
Lou1415926: why do we need this to be pbulic?

internal/pkg/cli/svc_init.go
huanjani: we do! it's in the next PR that's waiting in the wings. :)

internal/pkg/term/selector/local_selector.go
```diff
@@ -128,23 +129,80 @@ func (s *staticSelector) askCron(scheduleValidator prompt.ValidatorFunc) (string
...
+		selHelp,
+		dirsAndFiles,
+		nil,
+		prompt.WithFinalMessage(staticAssetsFinalMsg),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("select directories and/or files: %w", err)
+	}
+	for i, selection := range selections {
+		if selection == staticSourceUseCustomPrompt {
```
dannyrandall, line 175: nit: invert for less indentation: 
```suggestion
		if selection != staticSourceUseCustomPrompt {
		    continue
		}
```

internal/pkg/term/selector/local_selector.go
```diff
@@ -128,23 +129,80 @@ func (s *staticSelector) askCron(scheduleValidator prompt.ValidatorFunc) (string
...
+			for {
+				customPath, err := s.prompt.Get(
+					anotherPathPrompt,
+					anotherPathHelp,
+					pathValidator,
+					prompt.WithFinalMessage(customPathFinalMsg))
+				if err != nil {
+					return nil, fmt.Errorf("get custom directory or file path: %w", err)
+				}
+				if selection == staticSourceUseCustomPrompt {
```
dannyrandall: isn't this _always_ true at this point? based on the condition above?

internal/pkg/cli/init.go
huanjani: fixed

internal/pkg/term/selector/local_selector.go
Lou1415926: I'm okay with that, but before merging I have one last proposal 😂 :
```go
var results []string
var askCustom bool
for i, selection := range selections {
  if selection == statisSourceUseCustomPrompt {
    askCustom = true
    continue
  }
  results = append(results, selection)
}

 if !askCustom {
    return results
 }

 for {
    customPath, err := s.prompt.Get()
    paths = append(results, customPath)
    another, err := s.prompt.Confirm()
    if !another {
      break
    }
  }
```

I can see that the original code is trying to resize / copy the list for efficiency. However, I think the length of `selections` can't be too long, which makes the improvement in speed/memory usage fairly marginal. In this case, I feel like if we have to pick between readability & efficiency, I'd prolly vote for readability!

internal/pkg/workspace/workspace.go
```diff
@@ -157,10 +158,15 @@ func Create(appName string, fs afero.Fs) (*Workspace, error) {
 	return ws, nil
 }
 
+// ProjectRoot returns a path to the presumed root of the project, the directory that contains the copilot dir and .workspace file.
+func (ws *Workspace) ProjectRoot() (string) {
+	return strings.TrimSuffix(ws.CopilotDirAbs, "/copilot")
```
Lou1415926: We can use [`filepath.Dir`](https://pkg.go.dev/path/filepath#Dir) to handle forward/backward slash here!

internal/pkg/workspace/workspace.go
huanjani: for unit testing in the selector package :(

Lou1415926: ic!  I think that indicates that we need to `mock` the `ProjectRoot` call in `local_selector_test.go`, instead of using a real  `workspace.Workspace`. I am not sure if 👇🏼 is doable, so I'm fine if we prefer keeping it exported as for now (but we probably need to refactor it later so that we don't need to export a var for another package's testing).
```go
type workspaceRootReader interface {
  ProjectRoot() string
}
type LocalFilesSeletor struct {
  ws workspaceRootReader
}

func TestLocalFileSelector_listDirsAndFiles(t *testing.T) {
		mockFileSystem func(mockFS afero.Fs)
                 mockWS func(m *mocks.workspaceRootReader)
		wantedErr      error
		dirsAndFiles   []string
	}{
		"drill down two (and only two) levels": {
                         mockWS: func(m *mocks.workspaceRootReader) {
                               m.Expect().ProjectRoot().Return("mockRott")
                         }
                 }
         }
```

internal/pkg/term/selector/local_selector.go
huanjani: If the user inputs a custom path, `selection` now == the custom path. So if they opt to enter another and the for-loop repeats, the path will be added to the list differently than the first custom path.

But this all proves WX's point that the code is hard to read, and I'm changing it to her suggestion!:)

internal/pkg/term/selector/local_selector.go
```diff
@@ -128,23 +129,80 @@ func (s *staticSelector) askCron(scheduleValidator prompt.ValidatorFunc) (string
...
+		selPrompt,
+		selHelp,
+		dirsAndFiles,
+		nil,
+		prompt.WithFinalMessage(staticAssetsFinalMsg),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("select directories and/or files: %w", err)
+	}
+	for i, selection := range selections {
```
iamhopaul123: The logic doesn't quite make sense to me. it seems like we keep adding elements to the list (`selections`) we loop, which is an anti-pattern I think. Instead can we record the multi-select results and then ask them separately if `staticSourceUseCustomPrompt` is in `selections`. After that we can keep asking until they select `No` for the confirmation.

iamhopaul123: WX's proposal makes sense to me

internal/pkg/term/selector/local_selector.go
huanjani: Re: WX's latest suggested refactor (below), I implemented it but without `paths` because only the last custom path was being added to `results`.

internal/pkg/term/selector/local_selector.go
huanjani: I think the latest version solves this problem.

internal/pkg/workspace/workspace.go
```diff
@@ -60,13 +60,13 @@ const (
...
 	Application string `yaml:"application"` // Name of the application.
-	Path        string `yaml:"-"`           // absolute path to the summary file.
+	Path        string `yaml:"-"`           // Absolute path to the summary file.
 }
 
 // Workspace typically represents a Git repository where the user has its infrastructure-as-code files as well as source files.
 type Workspace struct {
 	workingDirAbs string
-	copilotDirAbs string
+	CopilotDirAbs string
```
dannyrandall: we can make this private now! 😊

internal/pkg/term/selector/local_selector.go
```diff
@@ -128,23 +129,84 @@ func (s *staticSelector) askCron(scheduleValidator prompt.ValidatorFunc) (string
...
+// StaticSources asks the user to select from a list of directories and files in the current directory and two levels down.
+func (s *localFileSelector) StaticSources(selPrompt, selHelp, anotherPathPrompt, anotherPathHelp string, pathValidator prompt.ValidatorFunc) ([]string, error) {
+	dirsAndFiles, err := s.listDirsAndFiles()
+	if err != nil {
+		return nil, err
+	}
+	if len(dirsAndFiles) == 0 {
+		log.Warningln("No directories or files were found in your workspace. Enter a relative path with the 'custom path' option if you'd like to use a hidden file.")
+	}
+	dirsAndFiles = append(dirsAndFiles, []string{staticSourceUseCustomPrompt}...)
```
dannyrandall:
```suggestion
	dirsAndFiles = append(dirsAndFiles, staticSourceUseCustomPrompt)
```

internal/pkg/term/selector/local_selector.go
```diff
@@ -177,7 +239,7 @@ func (s *localFileSelector) Dockerfile(selPrompt, notFoundPrompt, selHelp, notFo
 	return sel, nil
 }
 
-// ListDockerfiles returns the list of Dockerfiles within the current
+// listDockerfiles returns the list of Dockerfiles within the current
```
dannyrandall, line 245: totally unrelated and not for this pr but this func could probably use your `getDirAndFileNames` func now!

internal/pkg/term/selector/local_selector.go
```diff
@@ -219,6 +281,44 @@ func (s *localFileSelector) listDockerfiles() ([]string, error) {
 	return dockerfiles, nil
 }
 
+// listDirsAndFiles returns the list of directories and files within the current
+// working directory and two subdirectory levels below.
+func (s *localFileSelector) listDirsAndFiles() ([]string, error) {
+	names, err := s.getDirAndFileNames(s.ws.ProjectRoot(), 0)
```
dannyrandall: i wonder if the `depth` passed here should be 3 and we instead subtract in `getDirAndFileNames` until depth == 0 - that way the _user_ of the `getDirAndFileNames` function decides how deep to search!

dannyrandall (approved):
great work 👏

internal/pkg/term/selector/local_selector.go
huanjani: Oh, good call!

internal/pkg/term/selector/local_selector.go
huanjani: mm, that's just adding the one option to the choices.

internal/pkg/workspace/workspace.go
huanjani: I couldn't get WX's suggestion to work, so leaving it for a future refactor.

internal/pkg/term/selector/local_selector.go
huanjani: Ooooh, I love that!

huanjani: Done.

internal/pkg/term/selector/local_selector.go
dannyrandall: yeah that's fine - you just don't need the `[]string` wrapper 😊

huanjani: Oh, ha-- I read the suggestion backwards. Thought you were adding it! 🤦🏼

internal/pkg/cli/init.go
```diff
@@ -241,7 +241,11 @@ func newInitOpts(vars initVars) (*initOpts, error) {
 				dockerfilePath: vars.dockerfilePath,
 				image:          vars.image,
 			}
-			sel, err := selector.NewLocalFileSelector(prompt, fs)
+			ws, err := workspace.Use(fs)
+			if err != nil {
+				return err
+			}
+			sel, err := selector.NewLocalFileSelector(prompt, fs, ws)
```
Lou1415926: I think we probably need to move this block to inside of `useExistingWorkspaceClient`!

internal/pkg/term/selector/local_selector.go
```diff
@@ -219,6 +281,45 @@ func (s *localFileSelector) listDockerfiles() ([]string, error) {
...
 }
 
+// listDirsAndFiles returns the list of directories and files within the current
+// working directory and two subdirectory levels below.
+func (s *localFileSelector) listDirsAndFiles() ([]string, error) {
+	names, err := s.getDirAndFileNames(s.ws.ProjectRoot(), 3)
+	if err != nil {
+		return nil, err
+	}
+	return names, nil
```
Lou1415926: nit:
```suggestion
	return s.getDirAndFileNames(s.ws.ProjectRoot(), 3)
```

internal/pkg/term/selector/local_selector.go
```diff
@@ -219,6 +281,45 @@ func (s *localFileSelector) listDockerfiles() ([]string, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("read directory: %w", err)
+	}
+	var names []string
+	for _, file := range wdDirsAndFiles {
+		name := file.Name()
+		if strings.HasPrefix(name, ".") || name == "copilot" {
+			continue
+		}
+		relPathName := dir + "/" + name
```
Lou1415926: Would `filePath.Join` be more system-agnostic here!

internal/pkg/term/selector/local_selector.go
```diff
@@ -219,6 +281,45 @@ func (s *localFileSelector) listDockerfiles() ([]string, error) {
...
+		wsRelPathName, err := s.ws.Rel(relPathName)
+		if err != nil {
+			return nil, fmt.Errorf("get path relative to workspace: %w", err)
```
Lou1415926: nit:
```suggestion
			return nil, fmt.Errorf("get path relative to workspace for %s: %w", relPathName, err)
```

internal/pkg/term/selector/local_selector.go
```diff
@@ -219,6 +281,45 @@ func (s *localFileSelector) listDockerfiles() ([]string, error) {
...
+		}
+		names = append(names, wsRelPathName)
+		if depth > 0 && file.IsDir() {
+			subNames, err := s.getDirAndFileNames(relPathName, depth-1)
+			if err != nil {
+				return nil, fmt.Errorf("get dir and file names: %w", err)
```
Lou1415926: nit: I feel like we don't need to wrap in this case! Because we'd just get `get dir and file names: get dir and file names: get dir and file names: real err`
```suggestion
				return nil, err
```

internal/pkg/workspace/workspace.go
Lou1415926: Can we add a `TODO` here so that we (maybe an oncall?) remember to revisit this place and turn it private?

Lou1415926 (approved):
Looks great! Approving with labels for some nits 🎊

internal/pkg/cli/init.go
huanjani: I'm considering creating a different selector altogether, so `localFileSelector` doesn't need a ws but the new one does.

Lou1415926: I see! Ja as long as the selector does not require the presence of a workspace (by calling `workspace.Use()`), it can be initialized outside of `useExistingWorkspaceClient`

internal/pkg/cli/init.go
```diff
@@ -282,6 +282,66 @@ func newInitOpts(vars initVars) (*initOpts, error) {
 				o.initWlCmd = &opts
 				o.schedule = &opts.schedule // Surfaced via pointer for logging
 				o.initWkldVars = &opts.initWkldVars
+			case t == manifestinfo.StaticSiteType:
+				ws, err := workspace.Use(fs)
+				if err != nil {
+					var (
+						errWorkspaceNotFound *workspace.ErrWorkspaceNotFound
+					)
```
Lou1415926: tiniest nit:
```suggestion
					var errWorkspaceNotFound *workspace.ErrWorkspaceNotFound
```

internal/pkg/term/selector/local_selector.go
```diff
@@ -125,29 +126,93 @@ func (s *staticSelector) askCron(scheduleValidator prompt.ValidatorFunc) (string
...
+	if err != nil {
+		return nil, fmt.Errorf("get working directory: %w", err)
+	}
+	return &dockerfileSelector{
+		prompt:        prompt,
+		fs:            &afero.Afero{Fs: fs},
+		workingDirAbs: workingDirAbs,
+	}, nil
+}
+
```
Lou1415926: Can we move these definition to just before `func (s *dockerfileSelector) Dockerfile(selPrompt, notFoundPrompt, selHelp, notFoundHelp string, pathValidator prompt.ValidatorFunc) (string, error)` ?

internal/pkg/term/selector/local_selector.go
```diff
@@ -303,23 +302,51 @@ func (s *localFileSelector) getDirAndFileNames(dir string, depth int) ([]string,
 		if strings.HasPrefix(name, ".") || name == "copilot" {
 			continue
 		}
-		relPathName := dir + "/" + name
+		relPathName := filepath.Join(dir, name)
 		wsRelPathName, err := s.ws.Rel(relPathName)
 		if err != nil {
-			return nil, fmt.Errorf("get path relative to workspace: %w", err)
+			return nil, fmt.Errorf("get path relative to workspacefor %q: %w", relPathName, err)
```
Lou1415926:
```suggestion
			return nil, fmt.Errorf("get path relative to workspace for %q: %w", relPathName, err)
```

internal/pkg/cli/init.go
```diff
@@ -286,6 +282,66 @@ func newInitOpts(vars initVars) (*initOpts, error) {
...
+						return err
+					}
+					sourceFiles, err := selector.AskCustomPaths(o.prompt, staticSiteInitDirFilePathPrompt,
+						staticSiteInitDirFilePathHelpPrompt, func(v interface{}) error {
+							return validatePath(fs, v)
+						})
+					if err != nil {
+						return fmt.Errorf("ask for paths to source files: %w", err)
+					}
+					wkldVars.sourcePaths = sourceFiles
```
Lou1415926: Do we intend to `return` here?

internal/pkg/cli/init.go
```diff
@@ -286,6 +282,66 @@ func newInitOpts(vars initVars) (*initOpts, error) {
...
+					initSvcVars: svcVars,
+
+					fs:                fs,
+					sel:               dfSel,
+					sourceSel:         sel,
+					store:             configStore,
+					topicSel:          snsSel,
+					prompt:            prompt,
+					dockerEngine:      dockerengine.New(cmd),
+					wsPendingCreation: true,
```
Lou1415926: Ahh! Actually, this reminds me that "static assets"` is actually not the only service that faces this chicken-egg dilemma! Lookie here: https://github.com/aws/copilot-cli/blob/222ea9cf4fab097abd7229380d661ba1a5f66fae/internal/pkg/cli/job_init.go#L146

I wonder if we can do something similar for static asset:
```go
// In cli/init.go
setupWorkloadInit: func(o *initOpts, wkldType string) error {
   case manifestinfo.IsTypeAService(t):
      var wsPendingCreation bool
      ws, err := workspace.Use(fs)
      var errWorkspaceNotFound *workspace.ErrWorkspaceNotFound
      if err != nil {
        var errWorkspaceNotFound *workspace.ErrWorkspaceNotFound
        if !errors.As(err, &errWorkspaceNotFound) {
            return err
         }
        wsPendingCreation = true
      } 
      opts := initSvcOpts{		
        // ....other stuff is the same
        wsPendingCreation: true,
      }
}


// In cli/svc_init.go
func (o *initSvcOpts) askStaticSite() error {
        if o.wsPendingCreation {
            return AskCustomPath(...)
        }
	var assets []manifest.FileUpload
	sources, err := o.askSource()
	if err != nil {
		return err
	}
	for _, source := range sources {
		info, err := o.fs.Stat(source)
		if err != nil {
			return fmt.Errorf("get info for %q: %w", source, err)
		}
		assets = append(assets, manifest.FileUpload{
			Source:    source,
			Recursive: info.IsDir(),
		})
	}
	o.staticAssets = assets
	return nil
}
```

My main concern with the current approach is that we are prompting questions in `setupWorkloadInit`, which was meant to set up workload init commands. We were supposed to ask questions only in `Ask`, and if we break this diagram, it might be come difficult to debug/maintain the code.

huanjani: Yeah, that totally makes sense! This is much more streamlined and organized.

huanjani: We never actually check for a workspace before setting `wsPendingCreation` to true, so if someone is just adding a workload using `copilot init`, we act like it's not there. I'll add checks.

PR #4764, opened 2023-04-14
chore: remove lambda principle from cfn execution role

dannyrandall:
We don't need the lambda principle on the CFN execution role, it's just an artifact of the past.

### DIFF
--- a/internal/pkg/template/templates/environment/partials/cfn-execution-role.yml
+++ b/internal/pkg/template/templates/environment/partials/cfn-execution-role.yml
@@ -14,7 +14,6 @@ CloudformationExecutionRole:
         Principal:
           Service:
           - 'cloudformation.amazonaws.com'
-          - 'lambda.amazonaws.com'
         Action: sts:AssumeRole
   {{- if .PermissionsBoundary}}
     PermissionsBoundary: !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:policy/{{.PermissionsBoundary}}'
--- a/internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-basic-manifest.yml
+++ b/internal/pkg/deploy/cloudformation/stack/testdata/environments/template-with-basic-manifest.yml
@@ -75,7 +75,6 @@ Resources:
             Principal:
               Service:
                 - 'cloudformation.amazonaws.com'
-                - 'lambda.amazonaws.com'
             Action: sts:AssumeRole
       Path: /
       Policies:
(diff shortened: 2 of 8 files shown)


efekarakus (approved):
awesomeee 🚀

internal/pkg/template/templates/environment/partials/cfn-execution-role.yml
```diff
@@ -14,7 +14,6 @@ CloudformationExecutionRole:
         Principal:
           Service:
           - 'cloudformation.amazonaws.com'
-          - 'lambda.amazonaws.com'
```
efekarakus, line 17: oh derp do we need to bump the environment template version? I can never remember that decision.

Or do we only care about that if we add a new an envcontroller property

dannyrandall: @Lou1415926 can you remind us 😅

Lou1415926: iirc initially there was an env template version so that the `svc deploy` knows to upgrade the env template automatically. This was the `env upgrade` era.

Now, there is no `env upgrade`, and we substituted it with:
1. For an env-controller feature used in service, check if the environment that it deploys to support that feature. We do this by comparing the least env template version that supports that feature and the env's actual version.
2. Maybe for bug fixes that impact `svc deploy`
3. Encourage users to upgrade their env by running `copilot env deploy`

If this is a `chore`, not a `fix`, and it doesn't involve env controller feature, then maybe we don't need to bump the version

I also noticed that https://github.com/aws/copilot-cli/blob/bf4f19872dce711b1be277ae25407799b1d3efde/internal/pkg/template/env.go#L47 probably needed to be env template version instead of the CLI version 🤔

Lou1415926: Maybe we should make `envTemplate` version the same as the CLI version so that it's easier to keep track 😅 . I recall some discussion around this. Maybe we should pick this up soon.

dannyrandall: gotcha - we'll leave the version for this change then? and then this change will get deployed when someone runs `copilot env deploy`.

i think i have the App Runner feature version sitting around in another PR https://github.com/aws/copilot-cli/pull/4150 - i'll work on getting that one merged!

Lou1415926: Ja I think we can leave it!

Lou1415926 (approved):

PR #4772, opened 2023-04-18
fix: do record updates in parallel

### LINKED ISSUE #4711: CustomDomainFunction lambda times out
We have about 15 custom domains spread among different services in the same copilot application. We are observing failures when updating the stack with description "CloudFormation environment template for infrastructure shared among Copilot workloads." due to the `CustomDomainFunction` lambda timing out:

image

I assume a workaround would be to modify the stack template and increase the timeout to 900s (the lambda max timeout). Another option could be to update the lambda function code to reduce the interval in `waitForRecordChange()` to 10s or so, but I'm not so sure if I should just update the zip file containing the lambda code in S3 or there is a nicer way to update it.

Anyway, those are just workarounds since, eventually, it will timeout again if we add even more domains. Ideally, only changed domains would be updated, is that possible? Can you suggest some other solution? 

Thanks

bvtujo:
This PR implements the "small" changes in #4711, running the Custom Domain record updates in parallel. It does not implement the large changes, which would imply modifying the cert rather than deleting and recreating it. 

Fixes #4711 

### DIFF
--- a/cf-custom-resources/lib/custom-domain.js
+++ b/cf-custom-resources/lib/custom-domain.js
@@ -100,43 +100,45 @@ const writeCustomDomainRecord = async function (
   aliasTypes,
   action
 ) {
+  const actions = [];
   for (const alias of aliases) {
     const aliasType = await getAliasType(aliasTypes, alias);
     switch (aliasType) {
       case aliasTypes.EnvDomainZone:
-        await writeARecord(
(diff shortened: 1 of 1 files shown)


Lou1415926 (approved):
❤️ ty!

efekarakus (approved):
🚀

PR #4774, opened 2023-04-18
chore(diff): collapse collapasible diff paths into one line

Lou1415926:


internal/pkg/template/diff/formatter.go
```diff
@@ -123,9 +123,26 @@ func (f *keyedFormatter) formatMod(node diffNode) (string, error) {
 	return processMultiline(content, prefixByFn(prefixMod), indentByFn(f.indent)), nil
 }
 
-func (f *keyedFormatter) formatPath(node diffNode) string {
-	content := process(node.key()+":", prefixByFn(prefixMod), indentByFn(f.indent))
-	return content + "\n"
+// formatPath stringifies a diff path such that all keys with exactly one child are aggregated into one line.
+// For example, `HTTPRuleWithDomainPriorityAction/Properties`.
```
efekarakus: nit:
```suggestion
// joinPath collapses all key names under keyNode into one line as long as there is only modification under the key.
// For example, if only the `DesiredCount` of an ECS service is changed, then the returned path becomes `/Resources/Service/Properties/DesiredCount`. If multiple entries of an ECS service is changed, then the returned path is: `/Resources/Service/Properties`.
```

internal/pkg/template/diff/formatter.go
```diff
@@ -123,9 +123,26 @@ func (f *keyedFormatter) formatMod(node diffNode) (string, error) {
...
+// It returns the stringified path, as well as the node to which the last key on the path belongs.
+func (f *keyedFormatter) formatPath(keyNode diffNode) (string, diffNode) {
```
efekarakus: the return type seems a bit awkward 😂 from the name I'd expected to only return a `string`.

Maybe consider reversing the receiver and parameter?
```go
func (n *diffNode) iterUntilMultipleDiff(fmt formatter) (next *diffNode, path string)
```
I guess the difference is subtle, so I'll leave it to you to see if it's doable.

internal/pkg/template/diff/formatter.go
```diff
@@ -123,9 +123,26 @@ func (f *keyedFormatter) formatMod(node diffNode) (string, error) {
...
+		if len(keyNode.children()) != 1 {
+			break
+		}
+		peek := keyNode.children()[0]
+		if len(peek.children()) == 0 {
+			break
+		}
+		if _, ok := peek.(*node); !ok {
+			break
+		}
```
efekarakus: what does this mean?

efekarakus (commented):
nicesu! mostly some clarification suggestions

internal/pkg/template/diff/formatter.go
Lou1415926: This is saying that if `peek` is for example a seq node, then it shouldn't be joined on the path. 

Joining keys to form a path only makes sense for node that has a key, and `node` is the kind of node with a key. A `seqItemNode` does not have a key (and when printed, is prefixed with "-"). 

I should probably rename `node` to `keyedNode` - it'd probably read better this way.

internal/pkg/template/diff/formatter.go
```diff
@@ -123,9 +123,26 @@ func (f *keyedFormatter) formatMod(node diffNode) (string, error) {
...
+		keyNode = peek
+		content = content + "/" + keyNode.key()
```
dannyrandall: i'm curious about collapsable keys with a sequence in the path? i.e., only the 2nd policy in an iam role is changed, so the diff could show `MyPolicy/Properties/Policies/1/PolicyDocument/Version` or something? or does this only work with all-map paths?

Lou1415926: It only works with all-map paths. Your example would be 
```yaml
MyPolicy/Properties/Policies:
  - PolicyDocument/Version: 1 -> 2
  (3 unchanged items)
```

dannyrandall (approved):

internal/pkg/template/diff/formatter.go
Lou1415926: I ended up doing something similar:
```go
// joinNodes collapses all keyNode on a Tree path into one keyNode, as long as there is only modification under the key.
// For example, if only the `DesiredCount` of an ECS service is changed, then the returned path becomes
// `/Resources/Service/Properties`. If multiple entries of an ECS service is changed, then the returned
// path is `/Resources/Service`.
func joinNodes(curr *keyNode) *keyNode {
	key := curr.key()
	for {
		if len(curr.children()) != 1 {
			break
		}
		peek := curr.children()[0]
		if len(peek.children()) == 0 {
			break
		}
		if _, ok := peek.(*keyNode); !ok {
			break
		}
		key = key + "/" + peek.key()
		curr = peek.(*keyNode)
	}
	return &keyNode{
		keyValue:   key,
		childNodes: curr.children(),
	}
}
```

The "collapsing/joining" job is removed from `keyedFormatter`, and is instead handled by the `TreeWriter`, who is in charge of moving the `node` pointer.

paragbhingre (approved):
Sorry I am late to the party, but it looks really great 👍

PR #4777, opened 2023-04-19
feat: allow configuring multiple ports from CLI

paragbhingre:
1. `copilot init` and `copilot svc init` for LBWS can expose ports through CLI
2. `copilot init` and `copilot svc init` for BS can expose ports through CLI. Exposing single port behavior for BS remains unchanged.  
3. ALL the other workload types won’t be able to expose multiple ports at all. 


internal/pkg/cli/init.go
```diff
@@ -372,8 +373,8 @@ func (o *initOpts) logWorkloadTypeAck() {
 			color.HighlightUserInput(o.initWkldVars.wkldType), color.HighlightUserInput(o.initWkldVars.name), color.HighlightUserInput(o.initWkldVars.appName), color.HighlightUserInput(*o.schedule))
 		return
 	}
-	if aws.Uint16Value(o.port) != 0 {
-		log.Infof("Ok great, we'll set up a %s named %s in application %s listening on port %s.\n", color.HighlightUserInput(o.initWkldVars.wkldType), color.HighlightUserInput(o.initWkldVars.name), color.HighlightUserInput(o.initWkldVars.appName), color.HighlightUserInput(fmt.Sprintf("%d", *o.port)))
+	if len(*o.ports) > 0 {
+		log.Infof("Ok great, we'll set up a %s named %s in application %s listening on port(s) %s.\n", color.HighlightUserInput(o.initWkldVars.wkldType), color.HighlightUserInput(o.initWkldVars.name), color.HighlightUserInput(o.initWkldVars.appName), color.HighlightUserInput(strings.Join(*o.ports, ", ")))
```
huanjani: nit:
```suggestion
		log.Infof("Ok great, we'll set up a %s named %s in application %s listening on %s %s.\n", color.HighlightUserInput(o.initWkldVars.wkldType), color.HighlightUserInput(o.initWkldVars.name), color.HighlightUserInput(o.initWkldVars.appName), 
english.PluralWord(len(*o.ports), "port", "ports"),		color.HighlightUserInput(english.OxfordWordSeries(*o.ports, "and"))
```

internal/pkg/cli/svc_init.go
```diff
@@ -79,8 +79,9 @@ Deployed resources (such as your ECR repository, logs) will contain this %[1]s's
 	wkldInitDockerfilePathHelpPrompt = "Path to Dockerfile to use for building your container image."
 
 	svcInitSvcPortPrompt     = "Which %s do you want customer traffic sent to?"
-	svcInitSvcPortHelpPrompt = `The port will be used by the load balancer to route incoming traffic to this service.
-You should set this to the port which your Dockerfile uses to communicate with the internet.`
+	svcInitSvcPortHelpPrompt = `The port(s) will be used by the load balancer to route incoming traffic to this service.
+You should set this to the port(s) which your Dockerfile uses to communicate with the internet.
+You can also specify multiple container ports in Load Balanced Web Service in a similar pattern to Dockerfile (ports separated by a space), i.e., 3000 3001	`
```
huanjani:
```suggestion
You can specify multiple container ports for Load Balanced Web Services much as you would in a Dockerfile (ports separated by a space), e.g. 3000 3001`
```

internal/pkg/manifest/lb_web_svc_test.go
```diff
@@ -28,8 +28,8 @@ func TestNewHTTPLoadBalancedWebService(t *testing.T) {
 					Name:       "frontend",
 					Dockerfile: "./Dockerfile",
 				},
-				Path: "/",
-				Port: 80,
+				Path:  "/",
+				Ports: []uint16{80},
```
huanjani: Can we add a test with multiple ports? 🥺

internal/pkg/cli/svc_init.go
```diff
@@ -604,24 +621,45 @@ func (o *initSvcOpts) askSvcPort() (err error) {
...
+	copy(o.ports, portList)
+	return nil
+}
 
-	portUint, err := strconv.ParseUint(port, 10, 16)
-	if err != nil {
-		return fmt.Errorf("parse port string: %w", err)
+func (o *initSvcOpts) validateMultiplePorts() error {
+	if (o.wkldType != manifestinfo.LoadBalancedWebServiceType && o.wkldType != manifestinfo.BackendServiceType) && len(o.ports) > 1 {
+		return fmt.Errorf("%s doesn't expose multiple ports", o.wkldType)
```
iamhopaul123: Maybe
```suggestion
		return fmt.Errorf("service type %s does not support multiple ports", o.wkldType)
```

internal/pkg/cli/svc_init.go
```diff
@@ -79,8 +79,9 @@ Deployed resources (such as your ECR repository, logs) will contain this %[1]s's
 	wkldInitDockerfilePathHelpPrompt = "Path to Dockerfile to use for building your container image."
 
 	svcInitSvcPortPrompt     = "Which %s do you want customer traffic sent to?"
-	svcInitSvcPortHelpPrompt = `The port will be used by the load balancer to route incoming traffic to this service.
-You should set this to the port which your Dockerfile uses to communicate with the internet.`
+	svcInitSvcPortHelpPrompt = `The port(s) will be used by the load balancer to route incoming traffic to this service.
+You should set this to the port(s) which your Dockerfile uses to communicate with the internet.
+You can specify multiple container ports for Load Balanced Web Services and Backend Services e.g. 3000,3001`
```
bvtujo: So the syntax is `--port 300,3001` and not `--port 3000 --port 3001`? Or are these both valid with Cobra?

internal/pkg/cli/svc_init.go
```diff
@@ -604,24 +621,45 @@ func (o *initSvcOpts) askSvcPort() (err error) {
...
+			svcInitSvcPortHelpPrompt,
+			validateSvcPort,
+			prompt.WithDefaultInput(defaultPort),
+			prompt.WithFinalMessage("Port:"),
+		)
+		if err != nil {
+			return fmt.Errorf("get port: %w", err)
+		}
+		if len(strings.Split(selectedPorts, ",")) > 1 {
+			return fmt.Errorf("App Runner Service doesn't expose multiple ports")
```
bvtujo: I'm confused about the error case here: is this saying that you can't expose multiple ports with a RDWS?

internal/pkg/cli/svc_init.go
bvtujo: If this error occurs when we attempt to specify multiple ports on services that aren't LBWS or BS, can we change the wording?
```suggestion
		return fmt.Errorf("%s cannot expose multiple ports", o.wkldType)
```

internal/pkg/manifest/backend_svc.go
```diff
@@ -41,7 +41,8 @@ type BackendServiceConfig struct {
 // BackendServiceProps represents the configuration needed to create a backend service.
 type BackendServiceProps struct {
 	WorkloadProps
-	Port        uint16
+	Ports       []uint16
+	Path        string               // Optional path if multiple ports are exposed.
```
bvtujo, line 45: does http field on a backend service lead to internal alb being created? I'm confused how this works.

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -78,15 +82,38 @@ func NewLoadBalancedWebService(props *LoadBalancedWebServiceProps) *LoadBalanced
...
-	svc.LoadBalancedWebServiceConfig.ImageConfig.Port = aws.Uint16(props.Port)
 	svc.LoadBalancedWebServiceConfig.ImageConfig.HealthCheck = props.HealthCheck
 	svc.LoadBalancedWebServiceConfig.Platform = props.Platform
 	if isWindowsPlatform(props.Platform) {
 		svc.LoadBalancedWebServiceConfig.TaskConfig.CPU = aws.Int(MinWindowsTaskCPU)
 		svc.LoadBalancedWebServiceConfig.TaskConfig.Memory = aws.Int(MinWindowsTaskMemory)
 	}
-	if props.HTTPVersion != "" {
-		svc.HTTPOrBool.Main.ProtocolVersion = &props.HTTPVersion
+	if len(props.Ports) > 0 {
```
bvtujo: Do we validate that the same port doesn't appear twice in the dockerfile or manifest? Should this be an error condition?

bvtujo (commented):
LGTM, just a few nitpicky questions!

internal/pkg/manifest/lb_web_svc.go
paragbhingre: Yes, we do validate that at the time of deploying a service but not when we initiate it.

internal/pkg/manifest/backend_svc.go
paragbhingre: Yes you are right. http field in BS creates internal ALB. And hence we are allowing customer to expose multiple ports via internal ALB.

internal/pkg/cli/svc_init.go
paragbhingre: Yes that's right, App Runner service doesn't expose multiple ports.

internal/pkg/cli/svc_init.go
```diff
@@ -240,9 +241,16 @@ func (o *initSvcOpts) Validate() error {
...
+	if len(o.ports) > 0 {
+		if o.wkldType != "" {
+			if err := o.validateMultiplePorts(); err != nil {
+				return err
+			}
+		}
+		for _, port := range o.ports {
+			if err := validateSvcPort(port); err != nil {
+				return err
+			}
```
iamhopaul123: Can we wrap this logic into a method?

internal/pkg/cli/svc_init.go
```diff
@@ -81,7 +81,7 @@ Deployed resources (such as your ECR repository, logs) will contain this %[1]s's
 	svcInitSvcPortPrompt     = "Which %s do you want customer traffic sent to?"
 	svcInitSvcPortHelpPrompt = `The port(s) will be used by the load balancer to route incoming traffic to this service.
 You should set this to the port(s) which your Dockerfile uses to communicate with the internet.
-You can also specify multiple container ports in Load Balanced Web Service in a similar pattern to Dockerfile (ports separated by a space), i.e., 3000 3001	`
+You can specify multiple container ports for Load Balanced Web Services and Backend Services e.g. 3000,3001`
```
huanjani:
```suggestion
You can specify multiple container ports for Load Balanced Web Services and Backend Services, e.g., 3000,3001`
```

internal/pkg/cli/svc_init.go
```diff
@@ -573,8 +589,8 @@ func (o *initSvcOpts) askDockerfile() error {
 
 func (o *initSvcOpts) askSvcPort() (err error) {
 	// If the port flag was set, use that and don't ask.
-	if o.port != 0 {
-		return nil
+	if len(o.ports) > 0 {
+		return o.validateMultiplePorts()
```
iamhopaul123: We have `o.ports` here indicating they used flags and we should've validated this in `Validate()` i guess?

internal/pkg/manifest/backend_svc.go
```diff
@@ -57,6 +58,25 @@ func NewBackendService(props BackendServiceProps) *BackendService {
 	if len(props.Ports) > 0 {
 		svc.BackendServiceConfig.ImageConfig.Port = uint16P(props.Ports[0])
 	}
+	// we set http field in the manifest only when customer has given multiple ports via --port or a prompt.
```
huanjani: nit:
```suggestion
	// We set http field in the manifest only when customer has given multiple ports via --port or a prompt.
```

huanjani (approved):
Just a few small things.

internal/pkg/cli/svc_init.go
```diff
@@ -593,7 +609,8 @@ func (o *initSvcOpts) askSvcPort() (err error) {
 		case 0:
 			// There were no ports detected, keep the default port prompt.
 		case 1:
-			o.port = ports[0].Port
+			o.ports = make([]string, 1)
+			o.ports[0] = strconv.Itoa(int(ports[0].Port))
 			return nil
 		default:
 			defaultPort = strconv.Itoa(int(ports[0].Port))
```
iamhopaul123, line 676: Do we not help them to expose multiple ports?

internal/pkg/cli/svc_init.go
```diff
@@ -604,24 +621,45 @@ func (o *initSvcOpts) askSvcPort() (err error) {
 		return nil
 	}
 
-	port, err := o.prompt.Get(
-		fmt.Sprintf(svcInitSvcPortPrompt, color.Emphasize("port")),
+	if o.wkldType == manifestinfo.RequestDrivenWebServiceType {
```
iamhopaul123: Does a switch statement help here?

internal/pkg/cli/svc_init.go
```diff
@@ -604,24 +621,45 @@ func (o *initSvcOpts) askSvcPort() (err error) {
...
 		svcInitSvcPortHelpPrompt,
 		validateSvcPort,
 		prompt.WithDefaultInput(defaultPort),
-		prompt.WithFinalMessage("Port:"),
+		prompt.WithFinalMessage("Port(s):"),
 	)
 	if err != nil {
 		return fmt.Errorf("get port: %w", err)
 	}
+	portList := strings.Split(selectedPorts, ",")
```
iamhopaul123: How would users know they need to use comma separator for multiple ports? Do we need a different question?

internal/pkg/cli/svc_init.go
```diff
@@ -604,24 +621,45 @@ func (o *initSvcOpts) askSvcPort() (err error) {
...
+	o.ports = make([]string, len(portList))
+	copy(o.ports, portList)
```
iamhopaul123: Why do we need this?

internal/pkg/cli/svc_init.go
paragbhingre: So there are two workflows:
1. `copilot init` that does validation before ask phase. 
2. `copilot svc init` does validation first phase then ask for service type as well as port. So we need to check at this point if the multiple port use case is valid? Because port validation at validation phase doesn't check this condition as `wkldType`is unknown at that point. Does that make sense?

internal/pkg/cli/svc_init.go
paragbhingre: Ummm, do you mean if they have mentioned multiple ports in the Dockerfile, then use them to expose multiple ports? If yes, that could be done as well. But currently, we just let customers configure multiple ports if they want to. The reason I thought of is that we don't want to break customers who have multiple ports mentioned in their docker file today, and suddenly after this feature all those ports will start generating new manifests.

internal/pkg/cli/svc_init.go
paragbhingre: I actually took the refernece of `--import-cert-arns` flag and tried doing the same for consistency purpose. Previously, I thought of doing it as space separated like `3000 3002` so that it is consistent with how we specify ports in dockerfile but then comma separated made more sense in copilot flags world. Happy to get some thoughts here from you.

internal/pkg/cli/svc_init.go
paragbhingre: So if customers doesn't provide any port(s) in `--port` flag then we prompt for `Which port(s) do you want customer traffic sent to?`  and when they specify ports then we set those ports to `o.ports` field. We then use o.ports to build some of our service manifest configurations such as `image.port`  Does that make sense?

internal/pkg/cli/svc_init.go
paragbhingre: I checked and looks like both the syntaxes are valid. That was new to me 🙈

internal/pkg/cli/svc_init.go
bvtujo: Let's also change this wording to "cannot expose multiple ports" if that's okay with you. I think it's just a bit clearer; this way sort of sounds like the error is that the customer _hasn't_ specified multiple ports 🤔

internal/pkg/cli/svc_init.go
bvtujo: I wonder if we could add a prompt cycle, like so:
```
Which port do you want traffic sent to? 
> 8000
Would you like to add another traffic port?
> Yes
  No
Which port do you want traffic sent to? 
> 50051
Would you like to add another traffic port?
  Yes
> No
```
kind of like we do with Storage Init.

That avoids the need to do complicated parsing and if customers want to avoid it they can use the `--port 8000,50051` or `--port 8000 --port 50051` syntax which would skip the prompt entirely.

internal/pkg/cli/svc_init.go
bvtujo: Why can't we just do `o.ports = portList`?

paragbhingre:
I can still complete this as an external contributor 😉

internal/pkg/initialize/workload.go
```diff
@@ -297,7 +294,7 @@ func (w *WorkloadInitializer) newServiceManifest(i *ServiceProps) (encoding.Bina
 	case manifestinfo.RequestDrivenWebServiceType:
 		return w.newRequestDrivenWebServiceManifest(i), nil
```
iamhopaul123: We probably don't need `w.`

internal/pkg/template/templates/workloads/services/backend/manifest.yml
```diff
@@ -34,6 +34,23 @@ image:
     start_period: {{.ImageConfig.HealthCheck.StartPeriod}}
 {{- end}}
 
+{{- if gt (len .HTTP.AdditionalRoutingRules) 0}}
```
iamhopaul123: Do we need this? I thought we won't prompt users to add ports for backend service.

iamhopaul123 (commented):
LGTM overall. Thank you for picking this up!

internal/pkg/template/templates/workloads/services/backend/manifest.yml
paragbhingre: we do not prompt users to add ports for backend service but if customers run `copilot svc init --port 3000,3001` and then choose backend service then we are sure that customer knows about multiple ports concept and they are making a conscious effort to add  internal LB that too with multiple ports. At least that made sense to me unless you have another opinion.

bvtujo: I like this, I just want to make sure that we have a description telling them to check their manifest and customize the paths for each port for backend services. I'll add that in my next commit if something like it doesn't exist.

iamhopaul123 (approved):

internal/pkg/template/templates/workloads/services/lb-web/manifest.yml
```diff
@@ -16,6 +16,18 @@ http:
...
+  {{- if gt (len .HTTPOrBool.AdditionalRoutingRules) 0}}
+  # You can specify additional rules to route customer traffic to different container ports.
+  additional_rules:
+    {{- range $i, $rule := .HTTPOrBool.AdditionalRoutingRules }}
+    - path: '{{$rule.Path}}'
+      target_port: {{$rule.TargetPort}}
+      {{- if $rule.ProtocolVersion }}
+      version: {{$rule.ProtocolVersion}}
+      {{- end }}
+      # healthcheck: '{{$rule.HealthCheck.Basic}}'
```
Lou1415926: Should we delete this line? Or is it meant to render the health check (if there's any)

internal/pkg/template/templates/workloads/services/backend/manifest.yml
```diff
@@ -34,6 +34,23 @@ image:
...
+  # To match all requests you can use the "/" path.
+  path: '{{.HTTP.Main.Path}}'
+  # You can specify a custom health check path. The default is "/".
+  # healthcheck: '{{.HTTP.Main.HealthCheck.Basic}}'
+  # You can specify additional rules to route customer traffic to different container ports.
+  additional_rules:
+    {{- range $i, $rule := .HTTP.AdditionalRoutingRules }}
+    - path: '{{$rule.Path}}'
+      target_port: {{$rule.TargetPort}}
+      # healthcheck: '{{$rule.HealthCheck.Basic}}'
```
Lou1415926: samesies!

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -79,15 +83,38 @@ func NewLoadBalancedWebService(props *LoadBalancedWebServiceProps) *LoadBalanced
...
+		if props.Ports[0] == commonGRPCPort {
+			log.Infof("Detected port %s, setting HTTP protocol version to %s in the manifest.\n",
+				color.HighlightUserInput(strconv.Itoa(int(props.Ports[0]))), color.HighlightCode(GRPCProtocol))
+			svc.HTTPOrBool.Main.ProtocolVersion = aws.String(GRPCProtocol)
+		}
+	}
+	if len(props.Ports) > 1 {
+		svc.LoadBalancedWebServiceConfig.HTTPOrBool.AdditionalRoutingRules = make([]RoutingRule, len(props.Ports)-1)
+		assumedPath := DefaultHealthCheckAdminPath
+		for idx, port := range props.Ports {
```
Lou1415926: Would `for idx, port := range props.Ports[1:]` suffice?

internal/pkg/manifest/lb_web_svc.go
```diff
@@ -79,15 +83,38 @@ func NewLoadBalancedWebService(props *LoadBalancedWebServiceProps) *LoadBalanced
...
+			if idx > 0 {
+				if props.Ports[idx] == commonGRPCPort {
+					log.Infof("Detected port %s, setting HTTP protocol version to %s in the manifest.\n",
+						color.HighlightUserInput(strconv.Itoa(int(props.Ports[idx]))), color.HighlightCode(GRPCProtocol))
+					svc.HTTPOrBool.AdditionalRoutingRules[idx-1].ProtocolVersion = aws.String(GRPCProtocol)
+				}
+				svc.LoadBalancedWebServiceConfig.HTTPOrBool.AdditionalRoutingRules[idx-1].TargetPort = aws.Uint16(port)
+				svc.LoadBalancedWebServiceConfig.HTTPOrBool.AdditionalRoutingRules[idx-1].Path = aws.String(assumedPath)
```
Lou1415926: Honestly not super sure about rendering `admin`,`superadmin` and `supersuperadmin` in the default manifest generated from  `svc init` 😂 !

internal/pkg/initialize/workload.go
```diff
@@ -226,8 +223,8 @@ func (w *WorkloadInitializer) initService(props *ServiceProps) (string, error) {
 	log.Successf(manifestMsgFmt, svcWlType, color.HighlightUserInput(props.Name), color.HighlightResource(path))
 
 	helpText := "Your manifest contains configurations like your container size and port."
-	if props.Port != 0 {
-		helpText = fmt.Sprintf("Your manifest contains configurations like your container size and port (:%d).", props.Port)
+	if len(props.Ports) > 0 {
+		helpText = fmt.Sprintf("Your manifest contains configurations like your container size and %s (:%s).", english.PluralWord(len(props.Ports), "port", "ports"), strings.Trim(strings.Replace(fmt.Sprint(props.Ports), " ", ",", -1), "[]"))
```
Lou1415926: nit: this is fine but I found it relies on how `fmt.Sprint` format a list of uint. An alternative way that. I had in mind would be
```suggestion
		helpText = fmt.Sprintf("Your manifest contains configurations like your container size and %s (:%s).", english.PluralWord(len(props.Ports), "port", "ports"), strings.Join(applyAll(props.Ports, strconv. FormatUint), ","))
```

https://github.com/aws/copilot-cli/blob/50e492715d85affe75c66c9012a4afb4d242d616/internal/pkg/cli/cli.go#L149

But we would prolly need to change the function signature of `applyAll` to return a different type than `T`

internal/pkg/cli/svc_init.go
Lou1415926: I feel like it makes more sense to frame this prompt to "add additional path" rather than "add additional container port".

internal/pkg/manifest/lb_web_svc.go
bvtujo: How about `path1`, `path2`, `path3`?

Lou1415926: yeah that sounds good to me! 

Although I realize this comment is sort of related to https://github.com/aws/copilot-cli/pull/4777#discussion_r1184102260.

With the current implementation, we orient the whole `svc init` experience around container ports by asking "what container ports on your LBWS are going to receive traffic?" I might be biased, but this feels like an awkward way to think in 🤔 

I think we can consider asking "what paths does the service listen to?" and then as what are the container ports that are going to handle the traffic from those paths.

internal/pkg/template/templates/workloads/services/lb-web/manifest.yml
bvtujo: I think it's there to make healthcheck functionality visible since it's not super intuitive that you could turn it on. But we don't want to turn it on by default.

bvtujo: I think since we're not adding multiple ports and rendering this by default i'm going to remove it.

bvtujo:
I've made some changes here which: 
a) remove the multiple ports flag in favor of printing out some suggested actions on `svc init` if we detect multiple ports. 
b) remove the additional_rules logic for injecting the `http` field in the manifest for backend services.

Lou1415926 (approved):

PR #4779, opened 2023-04-20
chore: static site custom resources

dannyrandall:


cf-custom-resources/lib/trigger-state-machine.js
```diff
@@ -0,0 +1,135 @@
...
+      `${err.message} (Log: ${context.logGroupName}/${context.logStreamName})`
+    );
+  }
+};
+
+let deadlineExpired = () => {
+  return new Promise((resolve, reject) => {
+    setTimeout(
+      reject,
+      9 * 60 * 1000 + 30 * 1000 /* 9.5 minutes */,
```
Lou1415926: interestinggg why 9.5?

cf-custom-resources/lib/trigger-state-machine.js
```diff
@@ -0,0 +1,135 @@
...
+"use strict";
+
+const aws = require("aws-sdk");
+
+/**
+ * Main handler, invoked by Lambda
+ */
+exports.handler = async function (event, context) {
+  const responseData = {};
+  const physicalResourceId = event.PhysicalResourceId || event.LogicalResourceId;
```
Lou1415926, line 13: I see - we don't ever ever want to change `physicalResourceId` right? In this case, what about we just do `const physicalResourceId = "trigger-state-machine"`

cf-custom-resources/test/trigger-state-machine-test.js
```diff
@@ -0,0 +1,200 @@
...
+      })
+      .expectResolve(() => {
+        expect(request.isDone()).toBe(true);
+      });
+  });
+
+  test("happy path", () => {
+    const request = nock(responseURL)
+      .put("/", (body) => {
+        return body.Status === "SUCCESS";
```
Lou1415926: we can porlly test that the `PhysicalResourceId` is not changing in these test cases as well!

Lou1415926 (commented):
Looking good!!

cf-custom-resources/lib/copy-assets.js
```diff
@@ -0,0 +1,21 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
iamhopaul123, line 1: lol i like how succinct this lambda is

iamhopaul123 (approved):
:shipit:

cf-custom-resources/lib/trigger-state-machine.js
dannyrandall: tbh just copied from another lambda :joy:....i guess i could up it to like 14 minutes - closer to the real timeout, but still enough time to make sure the custom resource responds to cfn

cf-custom-resources/lib/trigger-state-machine.js
dannyrandall: my thought here was - if we ever reuse this custom resource in the same lambda, we'd have two different logical id's with the same physical id, whereas this gives us that flexibility+easy backwards compatibility (even if we change the logical id, we wouldn't accidentally break existing services)...maybe i'm overthinking it though and we should keep it simpler?

cf-custom-resources/lib/copy-assets.js
dannyrandall: might get more complicated with batching :)

cf-custom-resources/lib/trigger-state-machine.js
Lou1415926: ja! The other custom resources time out at like 14.5 minutes I think, leaving half a minute to respond (which is enough)

Lou1415926 (approved):

cf-custom-resources/lib/copy-assets.js
```diff
@@ -0,0 +1,21 @@
...
+
+"use strict";
+
+const aws = require("aws-sdk");
+
+/**
+ * Main handler, invoked by Lambda
+ */
+exports.handler = async function (event, context, callback) {
+	const s3 = new aws.S3();
```
efekarakus, line 12: nit: consider moving the initialization of the client outside of the handler (https://docs.aws.amazon.com/lambda/latest/operatorguide/static-initialization.html)

I think it might make a performance difference if the SFN state machine is invoking this Lambda fn for each object. 💭 

It would have been cool if we could have used the aws-sdk-v3 and just imported the s3 client :cool-crying:

cf-custom-resources/lib/trigger-state-machine.js
```diff
@@ -0,0 +1,135 @@
...
+          stateMachineArn: event.ResourceProperties.StateMachineARN,
+          input: JSON.stringify({
+            SourceBucket: event.ResourceProperties.SourceBucket,
+            AssetMappingFilePath: event.ResourceProperties.AssetMappingFilePath,
+            DestinationBucket: event.ResourceProperties.DestinationBucket,
+          })
+        };
+
+        const res = await sf.startSyncExecution(params).promise();
+        if (res.status === "FAILED") {
```
efekarakus, line 30: nit: Awesome, maybe we should add a comment here?
```

// Even if the execution starts and does not throw an error it does not mean the execution was successful.
// See https://docs.aws.amazon.com/step-functions/latest/apireference/API_StartSyncExecution.html#StepFunctions-StartSyncExecution-response-status
res.status !== "SUCCEEDED"
```

cf-custom-resources/lib/trigger-state-machine.js
```diff
@@ -0,0 +1,135 @@
...
+          throw new Error(`State machine failed: ${res.cause}`);
```
efekarakus, line 31: Do you have an example of what this looks like?

My main worry is that if the `copy-asset` fn fails for one of the objects, how easy will it be to troubleshoot for the customer what exactly happened? will they have to go to the SFN console or can we display enough info from the progress tracker?

efekarakus (commented):
Awesome job!

cf-custom-resources/lib/copy-assets.js
dannyrandall: i actually had the initialization outside initially, but https://www.npmjs.com/package/aws-sdk-mock 😔 

> NB: The AWS Service needs to be initialised inside the function being tested in order for the SDK method to be mocked

cf-custom-resources/lib/trigger-state-machine.js
dannyrandall: Ok these are the three error classes I can think of, and a screenshot of what the error looks like:
1. Trigger state machine lambda fails (no permission to execute state machine)
Screen Shot 2023-05-12 at 13 08 50

2. State machine fails (no permission to get asset mapping file)
Screen Shot 2023-05-12 at 13 18 09

3. Copy lambda fails (no permission to Put in dest bucket)
Screen Shot 2023-05-12 at 13 32 37

PR #4782, opened 2023-04-21
fix: don't add the ABR struct to the service if the manifest doesn't call for it

### LINKED ISSUE #4778: Deployments fail in GovCloud (copilot 1.27.0 and higher)
It looks like there may be a regression in version 1.27.0 when deploying to GovCloud west.

`copilot svc deploy --name web --env test`

During the update of the CloudFormation stack I see this error:

_Resource handler returned message: "Service was unable to process this request. (Service: AmazonECS; Status Code: 500; Error Code: ServerException; Request ID: 61897105-e675-4939-b35d-04cd80602377; Proxy: null)" (RequestToken: 210743fb-65a7-fe9e-20d4-5e94e4b5cdc4, HandlerErrorCode: GeneralServiceException)_

If it helps to debug, here is the diff... I don't know which part of this may be causing a problem with GovCloud.

```
~ Conditions:
    - IsDefaultRootPath: !Equals [!Ref RulePath, "/"]
~ Parameters:
    - Stickiness:
    -     Type: String
    -     Default: false
~ Resources:
    ~ HTTPListenerRuleWithDomain:
        ~ Metadata:
            ~ aws:copilot:description: 'An HTTP listener rule that redirects HTTP to HTTPS' -> 'An HTTP listener rule for path `/` that redirects HTTP to HTTPS'
        ~ Properties:
            ~ Conditions:
                (1 unchanged item)
                ~ - (changed item)
                  ~ PathPatternConfig:
                      ~ Values:
                          ~ - IsDefaultRootPath -> /*
                          - - - "/*"
                          - - - !Sub "/${RulePath}"
                          -   - !Sub "/${RulePath}/*"
    ~ HTTPRuleWithDomainPriorityAction:
        ~ Properties:
            - RulePath: !Ref RulePath
            + RulePath: ["/"]
    ~ HTTPSListenerRule:
        ~ Metadata:
for forwarding HTTPS traffic to your tasks' -> 'An HTTPS listener rule for path `/` that forwards HTTPS traffic to your tasks'
        ~ Properties:
            ~ Conditions:
                (1 unchanged item)
                ~ - (changed item)
                  ~ PathPatternConfig:
                      ~ Values:
                          ~ - IsDefaultRootPath -> /*
                          - - - "/*"
                          - - - !Sub "/${RulePath}"
                          -   - !Sub "/${RulePath}/*"
    ~ HTTPSRulePriorityAction:
        ~ Properties:
            - RulePath: !Ref RulePath
            + RulePath: ["/"]
    ~ RulePriorityFunction:
        ~ Properties:
            ~ Code:
                ~ S3Key: manual/scripts/custom-resources/rulepriorityfunction/584fba94a6e8a3ba8893428c31648e101fb8bf840647ce0b4505ffc86088ab50.zip -> manual/scripts/custom-resources/rulepriorityfunction/ac6830d3d4de8167bed1ce48eaf073ccbffe41076a1f88ea5c09b7b0ad71cb14.zip
    ~ Service:
        ~
  Continue with the deployment? (y/N)                                                                                                                                                                                            Properties:
            ~ DeploymentConfiguration:
                + Alarms:
                +     AlarmNames: []
                +     Enable: false
                +     Rollback: true
            ~ LoadBalancers:
                ~ - (changed item)
                  ~ ContainerName: !Ref TargetContainer -> web
                  ~ ContainerPort: !Ref TargetPort -> 443
    ~ TargetGroup:
        ~ Metadata:
            ~ aws:copilot:description: "A target group to connect the load balancer to your service" -> "A target group to connect the load balancer to your service on port 443"
        ~ Properties:
            ~ Port: !Ref TargetPort -> 443
            ~ TargetGroupAttributes:
                (1 unchanged item)
                ~ - (changed item)
                  ~ Value: !Ref Stickiness -> false
```

bvtujo:
This PR fixes #4778. It seems ABR is not available in GovCloud, so all service deployments with 1.27.0 into GovCloud regions fail. 

This PR changes the alarm injection logic to only add the ABR config if it's specified in the manifest, and exercises it one more time
in an integ test. 


internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
```diff
@@ -33,11 +33,8 @@ DeploymentConfiguration:
       {{- end }}
     {{- end }}
     Enable: true
-  {{- else }}
-    AlarmNames: []
-    Enable: false
-  {{- end}}
```
iamhopaul123, line 39: We still want to keep this because otherwise removing ABR setting from the manifest won't result in CFN updating on ECS service to disable ARB. In order to turn it off, we have to explicitly specify this in CFN template.

bvtujo: I don't think I understand. In the case where customers specify 
```
deployment:
  rollback_alarms:
    cpu_utilization: 70
```
then the ECS DeploymentConfiguration gets rendered like this: 
```
DeploymentConfiguration:
  MinHealthyPercent: 50
  MaxHealthyPercent: 200
  Alarms:
    AlarmNames:
      - "AppEnvWkldCopilotCPUAlarm"
    Enable: true
    Rollback: true
```

Removing the ABR config from the manifest should just remove all the Alarms keys from the CFN. Shouldn't that be sufficient to disable the feature?

iamhopaul123: > Removing the ABR config from the manifest should just remove all the Alarms keys from the CFN

Yes this is correct.

> Shouldn't that be sufficient to disable the feature?

I don't think so...ECS update service api is stateful so you have to explicitly disable the alarm otherwise it won't pick up any changes. This is due to the desired count issue ECS service used to have. The benefit is for updateService API when you don't specify desiredCount, ECS just assumes you don't want to update fields that are not specified. Same logic applies to the alarm field.

bvtujo: How irritating. i did some testing to confirm and it looks like you're right. I wrote a service with the following config:
```yaml
deployment:
  rollback_alarms:
    cpu_utilization: 80
```
then deployed it. Then I applied the following YAML patch:
```yaml
- op: remove
  path: /Resources/Service/Properties/DeploymentConfiguration/Alarms

- op: remove
  path: /Resources/CPURollbackAlarm
```
and while the alarm is deleted correctly, the service is not updated. I guess I'm going to close this and point future Govcloud users to the yamlpatch workaround.

bvtujo: What a thorny, annoying trade-off.

bvtujo:
Reopened because I added the NoValue logic we used for Service Connect in GovCloud

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
```diff
@@ -33,11 +33,13 @@ DeploymentConfiguration:
...
     Rollback: true
+  {{- else }}
+    !If
+    - IsGovCloud
+    - !Ref AWS::NoValue
+    - Enable: false
```
Lou1415926, line 41: Would the non-govcloud users need the `AlarmNames: []` line?

Lou1415926 (approved):
LGTM just a small question!

internal/pkg/template/templates/workloads/partials/cf/service-base-properties.yml
bvtujo: yes, definitely. I think it shouldn't matter for functionality if Enable is false but this is cleaner.

iamhopaul123 (approved):

PR #4784, opened 2023-04-21
chore: more actionable rdws deployment failed error message

siddvoh:
Updates the error shown when App Runner services fail to deploy.


internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -576,7 +576,9 @@ func (cf CloudFormation) errOnFailedStack(stackName string) error {
 	}
 	status := aws.StringValue(stack.StackStatus)
 	if cloudformation.StackStatus(status).IsFailure() {
-		return fmt.Errorf("stack %s did not complete successfully and exited with status %s", stackName, status)
+		return fmt.Errorf("stack %s did not complete successfully and exited with status %s.\n"+
```
dannyrandall: i think this part is already printed (the status/stack name) - maybe i'm wrong? but would be redundant if true

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -576,7 +576,9 @@ func (cf CloudFormation) errOnFailedStack(stackName string) error {
...
+			"You can retry building your service by updating the manifest (if needed) and running "+
```
dannyrandall: is there a way we could print this _only_ if the app runner resource fails? for example, it probably doesn't make sense to print a message telling someone to rebuild their service if the failure was due to a misconfigured RDS addon or something

internal/pkg/deploy/cloudformation/cloudformation.go
siddvoh: So what we print right now (without any changes) this:
Screen Shot 2023-04-01 at 8 15 21 AM
I'm just adding a small line after this.

internal/pkg/deploy/cloudformation/cloudformation.go
siddvoh: I'm not sure if we can see what is the exact error here (at least for now). We could maybe update the statement to be something like "Please see the error above or update the manifest(if needed)". By doing this, if we have an error which is "null" (current scenario with AR services) the user would want to look at their manifest. In all other cases, we should see a meaningful actionable error so we should be good for them in any case.

Lou1415926: +1 to @dannyrandall ! A sort of related & minor concern that I have is that, wordings like "(if needed)" is quite ambiguous. Like we've discussed before, this message is mostly useful for people that has very little experience with Copilot or AWS. I'm worried that for this group of audience, seeing "if needed" would just add to their frustration - like "how do I know if it's needed?"

siddvoh: That does make sense. Do we have any suggestions on what the wording should be? Is this better: "You can retry building your service by fixing the error shown above or by double-checking your service's manifest and running copilot svc deploy again."?

Lou1415926: What about something like this:
> "You may fix the error by updating the service code or the configuration in the manifest."

siddvoh: I like it. So then is this okay:

>You may fix the error by updating the service code or the configuration in the manifest. You can then retry building your service by running copilot svc deploy.

internal/pkg/deploy/cloudformation/cloudformation.go
dannyrandall: instead of a multi-line error - could we return an error struct with `RecommendActions()` for the follow up steps? an example: https://github.com/aws/copilot-cli/blob/ea86849093b9ffb03a8efec4cb595ad4fd10fc4a/internal/pkg/cli/svc_deploy.go#L555-L589

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -576,7 +576,9 @@ func (cf CloudFormation) errOnFailedStack(stackName string) error {
 	}
 	status := aws.StringValue(stack.StackStatus)
 	if cloudformation.StackStatus(status).IsFailure() {
-		return fmt.Errorf("stack %s did not complete successfully and exited with status %s", stackName, status)
+		return fmt.Errorf("stack %s did not complete successfully and exited with status %s.\n"+
+			"You may fix the error by updating the service code or the configuration in the manifest." +
+			"You can then retry building your service by running copilot svc deploy.", stackName, status)
```
dannyrandall:
```suggestion
			"You may fix the error by updating the service code or the manifest configuration." +
			"You can then retry deploying your service by running %s.", color.HighlightCode("copilot svc deploy"), stackName, status)
```

internal/pkg/deploy/cloudformation/cloudformation.go
siddvoh: I think we're better off keeping it as a multi-line error for now. When we move to stage 3 (and have error codes/error messages), we would then likely need to make the recommended actions struct. For now, it seems like a straightforward static error message would be sufficient and clearer.

Let me know if you'd still like me to make the RecommendedActions struct for this change

dannyrandall (approved):

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -576,7 +577,9 @@ func (cf CloudFormation) errOnFailedStack(stackName string) error {
 	}
 	status := aws.StringValue(stack.StackStatus)
 	if cloudformation.StackStatus(status).IsFailure() {
-		return fmt.Errorf("stack %s did not complete successfully and exited with status %s", stackName, status)
+		return fmt.Errorf("stack %s did not complete successfully and exited with status %s.\n"+
+			"You may fix the error by updating the service code or the manifest configuration." +
```
dannyrandall:
```suggestion
			"You may fix the error by updating the service code or the manifest configuration.\n" +
```

internal/pkg/deploy/cloudformation/cloudformation.go
siddvoh: Included struct.

dannyrandall: awesome! :heart: just wanted to clarify the implementation - your struct should have two functions on it: `Error() string` and `RecommendActions() string` 😊. The `Error()` function returns a short, one line error, and the `RecommendActions()` function returns to more detailed instructions.

KollaAdithya: +1 to Danny comment. Something like this ⬇️ 
```
type errFailedService struct {
    stackName string
    status    string
}
func (e *errFailedService) RecommendActions() string {
    return fmt.Sprintf("You may fix the error by updating the service code or the manifest configuration.\n"+
        "You can then retry deploying your service by running %s.",color.HighlightCode("copilot svc deploy"))
}
func (e *errFailedService) Error() string {
    return fmt.Sprintf("stack %s did not complete successfully and exited with status %s.", e.stackName, e.status)
}
if cloudformation.StackStatus(status).IsFailure() {
    return &errFailedService{
        stackName: stackName,
        status: status
    }
}
```

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -569,14 +570,30 @@ func (cf CloudFormation) deleteAndRenderStack(name, description string, deleteFn
...
 	if err != nil {
 		return err
 	}
 	status := aws.StringValue(stack.StackStatus)
 	if cloudformation.StackStatus(status).IsFailure() {
-		return fmt.Errorf("stack %s did not complete successfully and exited with status %s", stackName, status)
+		return &errFailedService{
+			stackName: stackName,
+			status: status,
+		}
```
Lou1415926: I feel like we should return this error only if the AppRunner resource is the one that's failing - considering how this is part of the effort to return actionable error message for AR only. 

We can probably rely on either [DescribeStackResource](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DescribeStackResource.html) or [DescribeStackEvents](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DescribeStackResource.html) to isolate failures that are due to AppRunner.  Even better, perhaps we could considering reusing [ErrorEvents](https://github.com/aws/copilot-cli/blob/7dc2153907df656b5b29615c7e30f424a5f9f136/internal/pkg/aws/cloudformation/cloudformation.go#L328), and iterate through the error events, decide whether there is one whose `ResourceType` is `AWS::AppRunner::Service`, and return this customized error only if there is one.

siddvoh: I think I would agree with this if we were currently showing errors/Recommended actions specific to App Runner. However, currently we are showing a generic error which lets the users know that they can make manifest changes. Since the Stack Failing could happen either because of 1) IAM create error and 2) App Runner create error, it makes sense to have the generic Recommended actions for both the cases as both the cases are handled with manifest updates and the recommended action is not specific to App Runner.

Lou1415926: TBH I'm still not super duper convinced that users do not already know that they should update manifest/application code 😂 Especially when we have already mentioned at the end of `env init`  and  `svc init` that "you can configure your manifest". 

Some other times, the stack update fails because of transient error, user's manual modification outside of Copilot, and, unfortunately, our bugs. I'm not sure if it's soothing that we still ask users to "update your configuration / application code" in these cases 🥲 

These were the main reasons why I try to avoid returning this error on _all_ stack failures. I understand that we are adding this right now because this was considered the first step towards more actionable error messages for AR. However, because this error message as it is right now is not super helpful (for the above reasons), I'd like to at least limit its occurrence.

siddvoh: Understood. Everything included now.

internal/pkg/aws/cloudformation/stack.go
```diff
@@ -99,6 +101,11 @@ func WithDisableRollback() StackOption {
 // StackEvent is an alias the SDK's StackEvent type.
 type StackEvent cloudformation.StackEvent
 
+func (e StackEvent) String() string {
```
Lou1415926: We can remove this if we don't actually use it! It helped us visualize what information are included in a `StackEvent`, but if it's not used anywhere, we can just keep it on our local work environment.

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -572,12 +572,16 @@ func (cf CloudFormation) deleteAndRenderStack(name, description string, deleteFn
 
 type errFailedService struct {
 	stackName string
+	service   string
```
Lou1415926:
```suggestion
	resourceType   string
```

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -590,9 +594,12 @@ func (cf CloudFormation) errOnFailedStack(stackName string) error {
 	}
 	status := aws.StringValue(stack.StackStatus)
 	if cloudformation.StackStatus(status).IsFailure() {
+		events, _ := cf.cfnClient.ErrorEvents(stackName)
+
 		return &errFailedService{
 			stackName: stackName,
-			status: status,
+			service:   aws.StringValue(events[0].ResourceType),
```
Lou1415926: I'd like us to be safe here -
```go
events, _ := cf.cfnClient.ErrorEvents(stackName)
var failedResourceType string
if events != nil {
   failedResourceType = events[0].ResourceType
}
return &errFailedService{
			stackName: stackName,
			service:   failedResourceType,
			status: status,
}
```

Just to avoid the potential index out of bound error!

Lou1415926 (approved):

internal/pkg/deploy/cloudformation/cloudformation.go
dannyrandall: maybe this instead?
```
if len(events) > 0 {
  failedResourceType = events[0].ResourceType
}
```

https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices
> avoid making a distinction between a nil slice and a non-nil, zero-length slice, as this can lead to subtle programming errors.

internal/pkg/aws/cloudformation/stack.go
siddvoh: Let me try and let update if needed.

internal/pkg/deploy/cloudformation/cloudformation.go
siddvoh: Implementing both these suggestions! :)

internal/pkg/aws/cloudformation/stack.go
siddvoh: Works fine! Removing.

PR #4762, opened 2023-04-13
chore: static site template

dannyrandall:
This PR is mostly about the static site CFN template. I'll pull the inline lambdas out into their own custom resources in a future PR, so don't worry about reviewing them right now. They should give you a general idea of how the whole flow works though!

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the Apache 2.0 License.

PR #4785, opened 2023-04-21
chore(static-site): hash uploaded content for asset mapping file

dannyrandall:
This change makes it so that the **asset mapping file name** is now a **hash of all the uploaded content**. In the template package (see https://github.com/aws/copilot-cli/pull/4762), the path to that file is passed to the custom resource that triggers the copy state machine. By making the file name based on a hash of the content, the state machine will only be triggered if any of the uploaded content changes.


internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -29,60 +30,62 @@ type ArtifactBucketUploader struct {
...
 	localPath string
-	content   io.Reader
+	content   []byte
 
 	ArtifactBucketPath string `json:"path"`
 	ServiceBucketPath  string `json:"destPath"`
 }
 
 // UploadFiles hashes each of the files specified in files and uploads
 // them to the path "{PathPrefix}/{hash}". After, it uploads a JSON file
```
Lou1415926:
```suggestion
// them to the path "{AssetDir}/{hash}". After, it uploads a JSON file
```

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"path"
```
Lou1415926, line 16: ooohhh good catch!

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -140,30 +143,40 @@ func (u *ArtifactBucketUploader) uploadAssets(assets []asset) error {
...
 //	  "path": "local-assets/12345asdf",
 //	  "destPath": "index.html"
-//	}
-func (u *ArtifactBucketUploader) uploadAssetMappingFile(assets []asset) error {
-	// stable output
+//	}]
+//
+// The uploaded path of the file is returned along with an error, if any.
+func (u *ArtifactBucketUploader) uploadAssetMappingFile(assets []asset) (string, error) {
 	sort.Slice(assets, func(i, j int) bool {
```
Lou1415926, line 160: Will this ever result in a equal comparison (I assume no - it probably would never make sense for two files to be uploaded to the same bucket and the same path, no matter how this package evolves 🤔 💭  ..)? If so, just in case, maybe we want to use https://pkg.go.dev/sort#SliceStable

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -140,30 +143,40 @@ func (u *ArtifactBucketUploader) uploadAssets(assets []asset) error {
...
 		}
 		return assets[i].ServiceBucketPath < assets[j].ServiceBucketPath
 	})
 
+	// hash using the sorted list so order-only changes in assets (caused by
+	// manifest or walkFn reordering) doesn't result in a mapping name change.
+	hash := sha256.New()
+	for _, asset := range assets {
+		// hash.Write is documented to never return an error
+		hash.Write(asset.content)
```
Lou1415926: If `asset` has changed its path that doesn't result in an reorder, do we want to produce a different hash?

dannyrandall: good catch!! i've included the mapping file content in the hash which should cover this case.

internal/pkg/cli/deploy/static_site.go
```diff
@@ -43,16 +41,14 @@ func NewStaticSiteDeployer(in *WorkloadDeployerInput) (*staticSiteDeployer, erro
...
-		assetMappingPath: assetMappingPath,
+		svcDeployer:   svcDeployer,
+		staticSiteMft: mft,
+		fs:            svcDeployer.fs,
 		uploader: &asset.ArtifactBucketUploader{
-			FS:               svcDeployer.fs,
-			PathPrefix:       "local-assets",
-			AssetMappingPath: assetMappingPath,
+			FS:                  svcDeployer.fs,
+			AssetDir:            "local-assets",
```
iamhopaul123: Should we make "local-assets" a constant so that it can be reused for `AssetMappingFileDir`?

internal/pkg/deploy/upload/asset/asset.go
dannyrandall: it _is_ possible that two mappings are the exact same - though likely not what the user is intending to configure. e.g., two uploads: one of folder `./static` and one of file `./static/index.html` with `dest: ./static` would result in _two_ `index.html hash` -> `/static/index.html` entries in assets

i think the right approach here is probably just to remove any duplicate elements instead of using slicestable - thoughts?

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -140,30 +143,40 @@ func (u *ArtifactBucketUploader) uploadAssets(assets []asset) error {
...
 		if assets[i].ArtifactBucketPath != assets[j].ArtifactBucketPath {
 			return assets[i].ArtifactBucketPath < assets[j].ArtifactBucketPath
 		}
 		return assets[i].ServiceBucketPath < assets[j].ServiceBucketPath
 	})
 
+	// hash using the sorted list so order-only changes in assets (caused by
+	// manifest or walkFn reordering) doesn't result in a mapping name change.
+	hash := sha256.New()
+	for _, asset := range assets {
```
iamhopaul123: Very nit just a note that i don't think `asset.localPath` matters

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -140,30 +143,40 @@ func (u *ArtifactBucketUploader) uploadAssets(assets []asset) error {
...
+
 	data, err := json.Marshal(assets)
 	if err != nil {
-		return fmt.Errorf("encode uploaded assets: %w", err)
+		return "", fmt.Errorf("encode uploaded assets: %w", err)
 	}
 
-	if err := u.Upload(u.AssetMappingPath, bytes.NewBuffer(data)); err != nil {
-		return fmt.Errorf("upload to %q: %w", u.AssetMappingPath, err)
+	path := path.Join(u.AssetMappingFileDir, hex.EncodeToString(hash.Sum(nil)))
```
iamhopaul123: can we avoid `path` as a var name? as it overrides the pkg name

iamhopaul123 (commented):
LGTM :shipit: only nits

internal/pkg/deploy/upload/asset/asset.go
dannyrandall: chatted offline, we are only hashing content

internal/pkg/deploy/upload/asset/asset.go
Lou1415926: > e.g., two uploads: one of folder ./static and one of file ./static/index.html with dest: ./static would result in two index.html hash -> /static/index.html entries in assets

Makes sense! In this case, the user could have just not specified `./static/index.html` right? If we don't remove the duplicated entries, it wouldn't have caused issues because we basically just upload the same thing twice. If this is the case, then

> i think the right approach here is probably just to remove any duplicate elements instead of using slicestable


Agreed!

dannyrandall: > the user could have just not specified ./static/index.html right?

yeah, exactly! 👍

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -174,9 +175,31 @@ func (u *ArtifactBucketUploader) uploadAssetMappingFile(assets []asset) (string,
 		return "", fmt.Errorf("encode uploaded assets: %w", err)
 	}
 
-	path := path.Join(u.AssetMappingFileDir, hex.EncodeToString(hash.Sum(nil)))
-	if err := u.Upload(path, bytes.NewBuffer(data)); err != nil {
+	// include mapping file in hash so any src->dst
+	// mapping changes result in a new hash
+	hash.Write(data)
```
Lou1415926: I'm trying to enumerate what would happen ⬇️ 💭 which looks good to me!!

Change content of a file ➡️  `ArtifactBucketPath` changes, i.e. the path to which it is uploaded is changed (because the path contains the hash) ➡️ `data` changes ➡️  This `hash` changes, because both `asset.content` and `data` change,

Change `dest` of an upload ➡️ `ServiceBucketPath` changes ➡️  This `hash` changes, because of `data`.

Move a file to a new local path (i.e. change `localPath` without changing `content`) ➡️ `localPath` ➡️  This `hash` does not change

Lou1415926 (approved):

internal/pkg/deploy/upload/asset/asset.go
dannyrandall: based on this ^ i realized we can just hash the asset mapping file and not bother hashing the actual asset file content, since the `ArtifactBucketPath` will only change if the file content changes. updated the pr!

iamhopaul123 (approved):

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -140,16 +143,20 @@ func (u *ArtifactBucketUploader) uploadAssets(assets []asset) error {
 	return g.Wait()
 }
 
-// uploadAssetMappingFile uploads a JSON file to u.AssetMappingPath containing
-// the current location of each file in the artifact bucket and the desired location
+// uploadAssetMappingFile uploads a JSON file containing the current
```
efekarakus, line 146: nit: why the "current" distinction, can the objects move in the artifact bucket?

efekarakus (commented):
Looks great!

The PR description says:
>By making the file name based on a hash of the content, **the state machine will only be triggered if any of the uploaded content changes.**

How do you achieve that in the state machine?

internal/pkg/deploy/upload/asset/asset.go
dannyrandall: ah i see that might be a little confusing, i could probably just drop the word. i meant "location of the file _right now_" as opposed to the service bucket location

dannyrandall:
> How do you achieve that in the state machine?

good question! the mapping file path [is an input](https://github.com/aws/copilot-cli/blob/222ea9cf4fab097abd7229380d661ba1a5f66fae/internal/pkg/template/templates/workloads/services/static-site/cf.yml#LL117C32-L117C32) to the custom resource that triggers the state machine - so cloudformation will only call `UPDATE` on the state-machine-trigger-lambda if the file path changes 😊

Issue #4780, opened 2023-04-20
`--diff` should not show a diff between full/short form of CFN intrinsic functions


Lou1415926:
Today, `--diff` shows a diff between these two snippets:
```yaml
# Short form
Subnets:
  - !Ref: PublicSubnet1

# Full form
Subnets:
  - Ref: PublicSubnet1
```

CFN templates generated by CDK uses full form everywhere, whereas Copilot uses the short form. Therefore, users running `--diff` for the first time after they applied CDK overrides will see a large amount of diff, while a majority of them are false alarm because of this.

It'd be great if `--diff` can tell that they are not really different. 

Doc on full/short form: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-ref.html

PR #4789, opened 2023-04-22
chore(diff): ignore difference for intrinsic functions that only changes between full/short form

Lou1415926:
#4780

### DIFF
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -202,6 +202,9 @@ func parseSequence(fromNode, toNode *yaml.Node, overriders ...overrider) ([]diff
 	}
 	cachedDiff := make(map[string]cachedEntry)
 	lcsIndices := longestCommonSubsequence(fromSeq, toSeq, func(idxFrom, idxTo int) bool {
+		if diff, ok := cachedDiff[cacheKey(idxFrom, idxTo)]; ok {
+			return diff.err == nil && diff.node == nil
+		}
 		diff, err := parse(&(fromSeq[idxFrom]), &(toSeq[idxTo]), "", overriders...)
 		if diff != nil { // NOTE: cache the diff only if a modification could have happened at this position.
 			cachedDiff[cacheKey(idxFrom, idxTo)] = cachedEntry{
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -203,6 +203,10 @@ func parseSequence(fromNode, toNode *yaml.Node, overriders ...overrider) ([]diff
 	}
 	cachedDiff := make(map[string]cachedEntry)
 	lcsIndices := longestCommonSubsequence(fromSeq, toSeq, func(idxFrom, idxTo int) bool {
+		// Note: This function passed as `eq` should be a pure function. Therefore, its output is the same
+		// given the same `idxFrom` and `idxTo`. Hence, it is not necessary to parse the nodes again.
+		// In `lcs.go`, `eq` can be called twice on the same indices: once when computing LCS length, and
+		// once when back-tracing to construct the LCS.
 		if diff, ok := cachedDiff[cacheKey(idxFrom, idxTo)]; ok {
 			return diff.err == nil && diff.node == nil
 		}
--- a/internal/pkg/template/diff/diff.go
+++ b/internal/pkg/template/diff/diff.go
@@ -136,9 +136,11 @@ func (from From) parseRoot(to []byte, overrider overrider) (Tree, error) {
 	}, nil
 }
 
-func parse(from, to *yaml.Node, key string, overrider overrider) (diffNode, error) {
-	if overrider.match(from, to, key, overrider) {
-		return overrider.parse(from, to, key, overrider)
+func parse(from, to *yaml.Node, key string, overriders ...overrider) (diffNode, error) {
+	for _, overrider := range overriders {
+		if overrider.match(from, to, key, overrider) {
+			return overrider.parse(from, to, key, overrider)
+		}
 	}
 	// Handle base cases.
 	if to == nil || from == nil || to.Kind != from.Kind {
@@ -163,11 +165,11 @@ func parse(from, to *yaml.Node, key string, overrider overrider) (diffNode, erro
 	var err error
 	switch {
 	case to.Kind == yaml.SequenceNode && from.Kind == yaml.SequenceNode:
-		children, err = parseSequence(from, to, overrider)
+		children, err = parseSequence(from, to, overriders...)
 	case to.Kind == yaml.DocumentNode && from.Kind == yaml.DocumentNode:
 		fallthrough
 	case to.Kind == yaml.MappingNode && from.Kind == yaml.MappingNode:
-		children, err = parseMap(from, to, overrider)
+		children, err = parseMap(from, to, overriders...)
 	default:
 		return nil, fmt.Errorf("unknown combination of node kinds: %v, %v", to.Kind, from.Kind)
 	}
@@ -187,7 +189,7 @@ func isYAMLLeaf(node *yaml.Node) bool {
 	return len(node.Content) == 0
 }
 
-func parseSequence(fromNode, toNode *yaml.Node, overrider overrider) ([]diffNode, error) {
+func parseSequence(fromNode, toNode *yaml.Node, overriders ...overrider) ([]diffNode, error) {
 	fromSeq, toSeq := make([]yaml.Node, len(fromNode.Content)), make([]yaml.Node, len(toNode.Content)) // NOTE: should be the same as calling `Decode`.
 	for idx, v := range fromNode.Content {
 		fromSeq[idx] = *v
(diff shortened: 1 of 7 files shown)


internal/pkg/template/diff/override_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+package diff
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIntrinsicFuncFullShortFormConverter(t *testing.T) {
```
iamhopaul123, line 13: Can we get a test case with diff and remove commented cases?

iamhopaul123 (commented):
LGTM overall!

internal/pkg/template/diff/diff.go
```diff
@@ -96,15 +98,12 @@ func (from From) ParseWithCFNIgnorer(to []byte) (Tree, error) {
 			},
 		},
 	}
-	return from.parseRoot(to, ignorer)
+	intrinsic := &intrinsicFuncFullShortFormConverter{}
+	return []overrider{ignorer, intrinsic}
```
dannyrandall: nit nit nit:
```suggestion
	return []overrider{
	    &ignorer{
	        key: "Metadata"
	        // etc
	    },
	    &intrinsicFuncFullShortFormConverter{}
	}
```

internal/pkg/template/diff/override.go
```diff
@@ -34,18 +36,65 @@ func (m *ignorer) match(_, _ *yaml.Node, key string) bool {
...
+type intrinsicFuncFullShortFormConverter struct{}
 
-// Match always returns false for a noopOverrider.
-func (*noopOverrider) match(_, _ *yaml.Node, _ string) bool {
-	return false
+// match returns true if from and to node represent the same intrinsic function written in different (full/short) form.
+// Example1: "!Ref abc" and "Ref: abc" will return true.
+// Example2: "!Ref abc" and "!Ref abc" will return false because they are written in the same form (i.e. short).
+// Example3: "!Ref abc" and "Fn::GetAtt: abc" will return false because they are different intrinsic functions.
+// For more on intrinsic functions and full/short forms, read https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-ToJsonString.html.
```
dannyrandall: great doc! 😍

internal/pkg/template/diff/override.go
```diff
@@ -34,18 +36,65 @@ func (m *ignorer) match(_, _ *yaml.Node, key string) bool {
...
+func (_ *intrinsicFuncFullShortFormConverter) match(from, to *yaml.Node, _ string, _ overrider) bool {
+	if from == nil || to == nil {
+		return false
+	}
+	if from.Kind == to.Kind || from.Kind != yaml.MappingNode && to.Kind != yaml.MappingNode {
+		// A full/short form conversion always involve at least one mapping node.
```
dannyrandall, line 56: unless they are the same kind of node?

dannyrandall: maybe i'm just misunderstanding the order of operations with the `this || that && the other` - could you add some `()` to clarify?

internal/pkg/template/diff/override.go
```diff
@@ -34,18 +36,65 @@ func (m *ignorer) match(_, _ *yaml.Node, key string) bool {
...
 
 // Parse is a no-op for an ignorer.
-func (m *ignorer) parse(_, _ *yaml.Node, _ string) (diffNode, error) {
+func (m *ignorer) parse(_, _ *yaml.Node, _ string, _ overrider) (diffNode, error) {
 	return nil, nil
 }
 
-type noopOverrider struct{}
+// intrinsicFuncFullShortFormConverter handles comparison between full/short form of an intrinsic function.
+type intrinsicFuncFullShortFormConverter struct{}
```
dannyrandall, line 44: will this ignore value changes? for example:
```
# from
!Ref abc
# to
Ref: xyz
```
this doesn't get ignored, right?

internal/pkg/template/diff/override.go
iamhopaul123: i think the first condition is to make sure we ignore the case when both of them are of the same type (both short or both long), but I agree more comments will be better.

internal/pkg/template/diff/override_test.go
```diff
@@ -0,0 +1,130 @@
...
+Name: !Sub
+  - 'www.${Domain}'
+  - Domain: !Ref RootDomainName`,
+			curr: `
+Name:
+  Fn::Sub:
+    - 'www.${Domain}'
+    - Domain: !Ref RootDomainName`,
+		},
+		// 		"no diff in Fn::Transform vs !Transform": {// TODO(lou1415926)
```
paragbhingre, line 387: Are we going to add these commented cases in later PRs?

internal/pkg/template/diff/diff.go
```diff
@@ -96,15 +98,12 @@ func (from From) ParseWithCFNIgnorer(to []byte) (Tree, error) {
...
 }
 
 // Parse constructs a diff tree that represent the differences of a YAML document against the From document.
-func (from From) Parse(to []byte) (Tree, error) {
-	return from.parseRoot(to, &noopOverrider{})
-}
-
-func (from From) parseRoot(to []byte, overrider overrider) (Tree, error) {
+func (from From) Parse(to []byte, overriders ...overrider) (Tree, error) {
```
efekarakus, line 107: This is so interesting, I'm just thinking out loud but we are accepting a _private_ type as input to a public method.

The downside is that clients of this package can't provide their own `overrider` but can use ones that are defined in this package. I think that's okay, it's just unusual to see.

internal/pkg/template/diff/override.go
```diff
@@ -34,18 +36,65 @@ func (m *ignorer) match(_, _ *yaml.Node, key string) bool {
...
 
-// Match always returns false for a noopOverrider.
-func (*noopOverrider) match(_, _ *yaml.Node, _ string) bool {
-	return false
+// match returns true if from and to node represent the same intrinsic function written in different (full/short) form.
+// Example1: "!Ref abc" and "Ref: abc" will return true.
```
efekarakus: huh this made me think that it's strange CFN doesn't do `Fn::Ref` instead of `Ref`.

Like what if a service API's property name was called `Ref` 🤯 what would happen lol

internal/pkg/template/diff/override.go
```diff
@@ -34,18 +36,65 @@ func (m *ignorer) match(_, _ *yaml.Node, key string) bool {
...
+	}
+	var fullFormNode, shortFormNode *yaml.Node
+	if from.Kind == yaml.MappingNode {
+		fullFormNode, shortFormNode = from, to
+	} else {
+		fullFormNode, shortFormNode = to, from
+	}
+	if len(fullFormNode.Content) != 2 {
+		// The full form mapping node always contain only one child node.
+		// Read https://www.efekarakus.com/2020/05/30/deep-dive-go-yaml-cfn.html.
```
efekarakus, line 67: 🤣

internal/pkg/template/diff/override.go
efekarakus: nit: do you think we should update the comment to not include any values like `!Ref` and `Ref:` as examples rather than `!Ref abc`

efekarakus (approved):
LGTM! +1 to Danny's comment to add parenthesis around the conditionals, and then feel free to remove the label

internal/pkg/template/diff/diff.go
Lou1415926: It's totally okay to make `overrider` a public interface - if `template/diff` was a public package consumed by arbitrary clients, I would have made `override` public.

It is private in this case because we know who the clients are, and we know that they would not intend to implement their own `overrider`.

In terms of package interface design though, I think it indeed makes more sense to make it public than private.

Lou1415926: In fact, since 
> we know who the clients are, and we know that they would not intend to implement their own overrider.

In fact, I think perhaps I should revert to the previous interface design
```
Parse()
ParseWithCFNOverriders()
```

internal/pkg/template/diff/override.go
Lou1415926: Nope! I'm pushing tests to show this

internal/pkg/template/diff/override_test.go
Lou1415926: Commented case will be uncommented in the next PR. Keeping them their because of the "ToDo" (and I also don't want to maintain another file that reminds me what the test case should look like...)

internal/pkg/template/diff/diff.go
Lou1415926: done!

internal/pkg/template/diff/diff.go
```diff
@@ -201,7 +203,10 @@ func parseSequence(fromNode, toNode *yaml.Node, overrider overrider) ([]diffNode
 	}
 	cachedDiff := make(map[string]cachedEntry)
 	lcsIndices := longestCommonSubsequence(fromSeq, toSeq, func(idxFrom, idxTo int) bool {
-		diff, err := parse(&(fromSeq[idxFrom]), &(toSeq[idxTo]), "", overrider)
+		if diff, ok := cachedDiff[cacheKey(idxFrom, idxTo)]; ok {
+			return diff.err == nil && diff.node == nil
```
efekarakus, line 211: nit: can you add a comment here on what scenario this represents? 🥺

efekarakus (approved):

internal/pkg/template/diff/diff.go
Lou1415926: done!

iamhopaul123 (approved):

PR #4795, opened 2023-04-24
chore(e2e): add multiple ports use case in app with domain e2e test

paragbhingre:


e2e/app-with-domain/app_with_domain_test.go
```diff
@@ -288,6 +290,7 @@ var _ = Describe("App With Domain", func() {
 			var resp *http.Response
 			var fetchErr error
 			urls := strings.Split(route.URL, " or ")
+			expectedResponseBody := "This is Copilot express app for admin"
 			for _, url := range urls {
```
dannyrandall, line 294: this is not from you!!! - but this is not correct if there were multiple urls. all of the `Eventually()` functions will only be looking at the _last_ URL in the list! can you add a line below the for loop like this?
```suggestion
			for _, url := range urls {
			    url := url
```

e2e/app-with-domain/app_with_domain_test.go
```diff
@@ -298,6 +301,19 @@ var _ = Describe("App With Domain", func() {
 					resp, fetchErr = http.Get(strings.Replace(url, "https", "http", 1))
 					return resp.StatusCode, fetchErr
 				}, "60s", "1s").Should(Equal(200))
+				Eventually(func() (int, error) {
+					resp, fetchErr = http.Get(fmt.Sprintf("%s/admin", strings.Replace(url, "https", "http", 1)))
+					body, err := io.ReadAll(resp.Body)
+					if err != nil {
+						return 0, err
+					}
+					fmt.Printf("response: %s\n", string(body))
```
dannyrandall: remove prints for cleaner logs? it will be in the error message if it's unexpected

e2e/app-with-domain/app_with_domain_test.go
```diff
@@ -298,6 +301,19 @@ var _ = Describe("App With Domain", func() {
...
+					// The LBWS sets its state to "consumed" when the worker service processes a message
+					if string(body) != expectedResponseBody {
+						return 0, fmt.Errorf("the message content is '%s', but expected '%s'", string(body), expectedResponseBody)
```
dannyrandall:
```suggestion
						return 0, fmt.Errorf("the message content is %q, but expected %q", string(body), expectedResponseBody)
```

e2e/app-with-domain/src/index.js
```diff
@@ -0,0 +1,31 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+let express = require('express');
+
+let app1 = express();
```
dannyrandall, line 6: does `app.listen()` block in js? if not - can we group the app1/app2 stuff logically? like:
```
const app1 = express();
app1.get("/", etc...)
app1.listen(port, etc...)

const app2 = express();
// etc
```

e2e/app-with-domain/app_with_domain_test.go
iamhopaul123: Does running them in parallel make more sense I wonder 🤔 I guess the same logic can be applied to `for _, route := range svc.Routes {`

dannyrandall: oh good point - it's probably not an issue if `Eventually` doesn't run in parallel...which it probably doesn't?

dannyrandall: ok it blocks until the condition is true - nvm @paragbhingre ! all good here

e2e/app-with-domain/app_with_domain_test.go
```diff
@@ -298,6 +301,19 @@ var _ = Describe("App With Domain", func() {
 					resp, fetchErr = http.Get(strings.Replace(url, "https", "http", 1))
 					return resp.StatusCode, fetchErr
 				}, "60s", "1s").Should(Equal(200))
+				Eventually(func() (int, error) {
+					resp, fetchErr = http.Get(fmt.Sprintf("%s/admin", strings.Replace(url, "https", "http", 1)))
```
dannyrandall, line 305:
```suggestion
					resp, fetchErr = http.Get(fmt.Sprintf("%s/admin", strings.Replace(url, "https", "http", 1)))
					if fetchErr != nil {
					   return 0, err
					}
					defer resp.Body.Close()
```

huanjani (approved):

iamhopaul123:
Close in favor of https://github.com/aws/copilot-cli/pull/4932

PR #4803, opened 2023-04-25
chore(s3): add s3 pkg to get real bucket name

iamhopaul123:


internal/pkg/s3/s3.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// New inits a new Client.
+func New(sess *session.Session) *Client {
+	return &Client{
+		rgGetter: resourcegroups.New(sess),
+	}
+}
+
+// BucketName returns the bucket name given the Copilot app, env, and Static Site service name.
```
bvtujo, line 36: Will this return buckets created in addons stacks? They should be tagged with the same app, env, svc tags as the main services. 

Can we add some more detail to the documentation describing this, if true, or add some error checking logic on svcType or bucketName? Or perhaps this validation is better done a level higher in the abstraction.

iamhopaul123: The same logic would apply to the ecs service where they define another ecs service in the addons stack 🤔 I don't think it's a valid use case to put an S3 bucket as addons for static site (just like it doesn't make sense to add an ECS service to any LBWS/BS/worker)?

iamhopaul123: Maybe we keep the logic for querying the real name for ECS service and S3 bucket as is for now? We can always replace the logic with checking CFN stack to find out instead of using tags in the future (until they have a valid use case)

bvtujo: Makes sense to me. I bring it up because I've just had some recent oncall experience with customers whose stacks get into baffling configurations that I worry about guardrails more than necessary.

bvtujo (approved):

internal/pkg/s3/s3.go
```diff
@@ -0,0 +1,57 @@
...
+	buckets, err := c.rgGetter.GetResourcesByTags(bucketType, map[string]string{
+		deploy.AppTagKey:     app,
+		deploy.EnvTagKey:     env,
+		deploy.ServiceTagKey: svc,
+	})
+	if err != nil {
+		return "", fmt.Errorf("get S3 bucket with tags (%s, %s, %s): %w", app, env, svc, err)
+	}
+	if len(buckets) == 0 {
+		return "", fmt.Errorf("no S3 bucket found for %s in environment %s", svc, env)
```
dannyrandall: nit
```suggestion
		return "", fmt.Errorf("no S3 bucket found with tags %s, %s, %s", svc, env)
```

internal/pkg/s3/s3.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	if len(buckets) > 1 {
+		return "", fmt.Errorf("more than one S3 bucket with the name %s found in environment %s", svc, env)
+	}
+	bucketName, _, err := s3.ParseARN(buckets[0].ARN)
+	if err != nil {
+		return "", fmt.Errorf("parse ARN for %s: %w", svc, err)
```
dannyrandall: nit:
```suggestion
		return "", fmt.Errorf("parse ARN %s: %w", buckets[0].ARN, err)
```
the function calling `BucketName()` already has the app/env/service context for the error, but won't have the ARN info so maybe including the ARN instead would be nice

dannyrandall (approved):
lgtm!

internal/pkg/s3/s3.go
```diff
@@ -0,0 +1,57 @@
...
+	GetResourcesByTags(resourceType string, tags map[string]string) ([]*resourcegroups.Resource, error)
+}
+
+// Client retrieves Copilot S3 service information from AWS.
+type Client struct {
+	rgGetter resourceGetter
+}
+
+// New inits a new Client.
+func New(sess *session.Session) *Client {
```
efekarakus, line 30: Rather than the session, should we aim for all packages except `cli` to pass down dependencies?

```
type ResourcesByTagsGetter interface {
   GetResourcesByTags(resourceType string, tags map[string]string) ([]*resourcegroups.Resource, error)
}

func New(rgapi ResourcesByTagsGetter) *Client
```

internal/pkg/s3/s3.go
```diff
@@ -0,0 +1,57 @@
...
+	buckets, err := c.rgGetter.GetResourcesByTags(bucketType, map[string]string{
+		deploy.AppTagKey:     app,
+		deploy.EnvTagKey:     env,
+		deploy.ServiceTagKey: svc,
+	})
+	if err != nil {
+		return "", fmt.Errorf("get S3 bucket with tags (%s, %s, %s): %w", app, env, svc, err)
+	}
+	if len(buckets) == 0 {
+		return "", fmt.Errorf("no S3 bucket found with tags %s, %s, %s", svc, env, svc)
```
efekarakus, line 47: nit: the tags in the error message are inconsistent, consider:

```go
type tags map[string]string
func (tags tags) String() string {
   serialized := make([]string, len(tags))
   var i = 0
   for k, v := range tags {
     serialized[i] = fmt.Sprintf("%s=%s", k, v)
     i += 1
   }
   return strings.Join(serialized, ", ")
}
```

and invoking tags.String() in the error messages

efekarakus: The second error message also has `svc` twice instead of `app` 😅

iamhopaul123: Lol I'll fix in the current PR.

> and invoking tags.String() in the error messages

We can probably add it if it happens more in this lib, since it also takes extra logic to construct a map that's only used for err msg.

efekarakus: sounds good! but it'd be helpful either way to also include the keys of the tags and not just the values so that customers can debug

internal/pkg/s3/s3.go
iamhopaul123: I'm trying to be consistent with `internal/ecs` pkg. Maybe we can do a campaign to make this behavior consistent across the codebase (add it to the sprint planning task)?

internal/pkg/s3/s3.go
iamhopaul123: Actually I'll update with `tags`. It's a great idea!

PR #4807, opened 2023-04-26
fix: access override templates using forward slashes

### LINKED ISSUE #4791: Allow Memory Hard Limit setting.
Per the article:

https://aws.amazon.com/blogs/developer/configuring-net-garbage-collection-for-amazon-ecs-and-aws-lambda/

.NET GC sees the underlying hosts' total memory instead of the containers memory which can lead to OOM errors. The article states that you can specify the hard limit of the container which solves the issue, however; I haven't seen an easy to implement mechanism via manifest file. I'm not sure if this is implemented, but I could see it being very useful.

Lou1415926:
Fix #4791 


efekarakus (approved):
Ugh :( my bad

efekarakus:
Thanks for fixing this!

huanjani (approved):

PR #4808, opened 2023-04-27
fix: use the more portable CDK executable for cdk overrides

Lou1415926:
The old one points to a script with shebang line, which is ignored by Windows. As a result, when I used cdk overrides on Windows, it ends up not being able to interprete the cdk.js file


bvtujo (approved):
Nice find on this fix!

huanjani (approved):

internal/pkg/override/cdk.go
```diff
@@ -126,7 +126,7 @@ func (cdk *CDK) transform(body []byte) ([]byte, error) {
 
 	// We assume that a node_modules/ dir is present with the CDK downloaded after running "npm install".
 	// This way clients don't need to install the CDK toolkit separately.
-	cmd := cdk.exec.Command(filepath.Join("node_modules", "aws-cdk", "bin", "cdk"), "synth", "--no-version-reporting")
+	cmd := cdk.exec.Command(filepath.Join("node_modules", ".bin", "cdk"), "synth", "--no-version-reporting")
```
efekarakus, line 129: What does the `.bin/cdk` file's content look like out of curiosity?

Lou1415926: I think it looks the same as `aws-cdk/bin/cdk` on Unix systems, but on Windows it looked different! After I got home I can check on my little Windows machine.

Putting the DNM label on so that i can post the snippet before it's merged!

dannyrandall: (on windows) `node_modules/.bin/cdk`:
```
#!/bin/sh
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")

case `uname` in
    *CYGWIN*|*MINGW*|*MSYS*) basedir=`cygpath -w "$basedir"`;;
esac

if [ -x "$basedir/node" ]; then
  exec "$basedir/node"  "$basedir/../aws-cdk/bin/cdk" "$@"
else 
  exec node  "$basedir/../aws-cdk/bin/cdk" "$@"
fi
```

(on mac) `node_modules/.bin/cdk`
```
#!/usr/bin/env node
require('./cdk.js');
```

efekarakus: ty!

PR #4811, opened 2023-04-27
chore(manifest): make cdn location to accept static site

iamhopaul123:

### DIFF
--- a/internal/pkg/aws/cloudfront/cloudfront.go
+++ b/internal/pkg/aws/cloudfront/cloudfront.go
@@ -6,7 +6,7 @@ const (
 	// CertRegion is the only AWS region accepted by CloudFront while attaching certificates to a distribution.
 	CertRegion = "us-east-1"
 
-	// Regex validation format for S3 bucket as CloudFront origin domain
+	// S3BucketOriginDomainFormat is the Regex validation format for S3 bucket as CloudFront origin domain
 	// See https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html#DownloadDistValuesDomainName
 	S3BucketOriginDomainFormat = `.+\.s3.*\.\w+-\w+-\d+\.amazonaws\.com`
 )
--- a/internal/pkg/aws/cloudfront/cloudfront.go
+++ b/internal/pkg/aws/cloudfront/cloudfront.go
@@ -5,4 +5,8 @@ package cloudfront
 const (
 	// CertRegion is the only AWS region accepted by CloudFront while attaching certificates to a distribution.
 	CertRegion = "us-east-1"
+
+	// Regex validation format for S3 bucket as CloudFront origin domain
+	// See https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html#DownloadDistValuesDomainName
+	S3BucketOriginDomainFormat = `.+\.s3.*\.\w+-\w+-\d+\.amazonaws\.com`
 )
--- a/internal/pkg/cli/deploy/env.go
+++ b/internal/pkg/cli/deploy/env.go
@@ -20,7 +20,7 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/aws/ec2"
 	"github.com/aws/copilot-cli/internal/pkg/aws/elbv2"
 	"github.com/aws/copilot-cli/internal/pkg/aws/partitions"
-	"github.com/aws/copilot-cli/internal/pkg/aws/s3"
+	awss3 "github.com/aws/copilot-cli/internal/pkg/aws/s3"
 	"github.com/aws/copilot-cli/internal/pkg/aws/sessions"
 	"github.com/aws/copilot-cli/internal/pkg/cli/deploy/patch"
 	"github.com/aws/copilot-cli/internal/pkg/config"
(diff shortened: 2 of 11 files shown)


internal/pkg/aws/cloudfront/cloudfront.go
```diff
@@ -5,4 +5,8 @@ package cloudfront
 const (
 	// CertRegion is the only AWS region accepted by CloudFront while attaching certificates to a distribution.
 	CertRegion = "us-east-1"
+
+	// S3BucketOriginDomainFormat is the Regex validation format for S3 bucket as CloudFront origin domain
+	// See https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html#DownloadDistValuesDomainName
+	S3BucketOriginDomainFormat = `.+\.s3.*\.\w+-\w+-\d+\.amazonaws\.com`
```
efekarakus, line 11: What's the difference with `.s3-website`? Looking at https://docs.aws.amazon.com/AmazonS3/latest/userguide/WebsiteEndpoints.html I'd have guessed that would be the relevant endpoint for `static site`? can you explain the difference to me 🥺

internal/pkg/cli/deploy/env.go
```diff
@@ -154,8 +160,9 @@ func NewEnvDeployer(in *NewEnvDeployerInput) (*envDeployer, error) {
 		env: in.Env,
 
 		templateFS:       template.New(),
-		s3:               s3.New(envManagerSession),
+		s3:               awss3.New(envManagerSession),
 		prefixListGetter: ec2.New(envRegionSession),
+		bucketNameGetter: s3.New(envManagerSession),
```
efekarakus, line 165: (just brainstorming 🧠 ⚡ ):

Seeing these two field names, it feels odd that we would need to initialize one client for s3 and another for getting a bucket name 💭 

I wonder if instead our "higher"-level packages like `internal/pkg/s3` should _embed_ the `internal/pkg/aws/s3`  clients so that they strictly augment the AWS SDK's method and we don't have to import two pkgs?

internal/pkg/manifest/env.go
```diff
@@ -273,14 +275,43 @@ func (cfg *EnvironmentCDNConfig) UnmarshalYAML(value *yaml.Node) error {
...
+	Location StaticSiteOrLocation `yaml:"location,omitempty"`
+	Alias    string               `yaml:"alias,omitempty"`
+	Path     string               `yaml:"path,omitempty"`
+}
+
+// StaticSiteOrLocation is a custom type which supports unmarshaling yaml which
+// can either be of Static Site service name or a S3 bucket URL.
+type StaticSiteOrLocation struct {
+	StaticSite string
+	URL        string
```
```suggestion
	BucketURL        string
```

efekarakus (commented):
Looks good boss, mainly a question around the regex

internal/pkg/aws/cloudfront/cloudfront.go
iamhopaul123: That's a good question. s3-website I think is just so that you can have website endpoint vs rest API endpoint: https://docs.aws.amazon.com/AmazonS3/latest/userguide/WebsiteEndpoints.html#WebsiteRestEndpointDiff

internal/pkg/cli/deploy/env.go
iamhopaul123: Do you mean all the public things in `aws/s3` should be embedded in `s3`? How about public function/variables?

huanjani (approved):
LGTM.

internal/pkg/cli/deploy/env.go
```diff
@@ -403,6 +410,9 @@ func (d *envDeployer) buildStackInput(in *DeployEnvironmentInput) (*cfnstack.Env
 	if err != nil {
 		return nil, err
 	}
+	if err = d.renderStaticSite(in.Manifest); err != nil {
```
dannyrandall:
```suggestion
	if err := d.renderStaticSite(in.Manifest); err != nil {
```

internal/pkg/manifest/env.go
```diff
@@ -273,14 +275,43 @@ func (cfg *EnvironmentCDNConfig) UnmarshalYAML(value *yaml.Node) error {
...
+		return err
+	}
+	reg := regexp.MustCompile(cloudfront.S3BucketOriginDomainFormat)
+	if reg.MatchString(s.ImportedBucket) {
+		return nil
+	}
+	if err := value.Decode(&s.StaticSite); err != nil {
+		return err
+	}
+	s.ImportedBucket = ""
```
dannyrandall:
```suggestion
	s.StaticSite, s.ImportedBucket = s.ImportedBucket, ""
```

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -96,7 +96,7 @@ CloudFrontDistribution:
             {{- end}}
       {{- if .CDNConfig.Static}}
         - Id: !Sub 'copilot-${AppName}-${EnvironmentName}-s3'
-          DomainName: {{.CDNConfig.Static.Location}}
+          DomainName: {{.CDNConfig.Static.ImportedBucket}}
```
dannyrandall, line 99: what about static site bucket?

iamhopaul123: next pr sir

internal/pkg/cli/deploy/env.go
iamhopaul123: why should we make this change?

dannyrandall: eh, i guess i don't have a strong preference. `:=` just limits the scope of this specific `err` to the `if` block, and i'm just used to seeing that 😊 but it's perfectly fine to leave it!!!

dannyrandall (approved):
adding a do-not-merge label for efe's feedback!

PR #4814, opened 2023-04-28
chore(ci): add unit test into `local-test` make target

Lou1415926:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,10 +111,10 @@ run-integ-test:
 	go test -race -count=1 -timeout 120m -tags=integration ${PACKAGES}
 
 .PHONY: local-test
-local-test: package-custom-resources custom-resource-tests run-local-integ-test package-custom-resources-clean
+local-test: package-custom-resources custom-resource-tests run-local-test package-custom-resources-clean
 
 .PHONY: run-local-test
-run-local-integ-test:
+run-local-test:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -111,10 +111,10 @@ run-integ-test:
 	go test -race -count=1 -timeout 120m -tags=integration ${PACKAGES}
 
 .PHONY: local-test
-local-test: package-custom-resources custom-resource-tests run-local-test package-custom-resources-clean
+local-test: package-custom-resources custom-resource-tests run-local-integ-test run-unit-test package-custom-resources-clean
 
 .PHONY: run-local-test
-run-local-test:
+run-local-integ-test:
```
dannyrandall: don't the unit tests already get run here? so now this'll just run them twice?

Lou1415926: Didn't realize `--tag` was inclusive and was not excluding tests that are not tagged.

Anyway, I ended up with this PR because I found out that `make test` was not passing at `StaticSite_Parameters` test, and the fact that `make local-test` was passing made me think that it wasn't running unit test.

The real reason turns out to be that `StaticSite_Parameters` didn't stub the `fs`, therefore it required `package-custom-resources` to pass.

I've fixed it accordingly.

dannyrandall (approved):

huanjani (approved):

PR #4816, opened 2023-04-29
docs: fix PolicyName indent

interu:
The following errors occurred when implemented as per the sample.
"Property PolicyName cannot be empty."

The indentation of PolicyName has been changed for correct operation.


bvtujo (approved):
Thank you for this fix!

efekarakus (approved):
Thank you!!

PR #4823, opened 2023-05-02
chore: empty static site s3 bucket on `svc delete`

dannyrandall:


internal/pkg/aws/s3/s3.go
```diff
@@ -111,14 +112,18 @@ func (s *S3) EmptyBucket(bucket string) error {
...
 				Objects: objectsToDelete,
+				Quiet:   aws.Bool(true), // we don't care about success values
 			},
 		})
-		if err != nil {
+		switch {
+		case err != nil:
 			return fmt.Errorf("delete objects from bucket %s: %w", bucket, err)
+		case len(delResp.Errors) > 0:
+			return fmt.Errorf("%d/%d objects failed to delete. first failure on key %q: %s", len(delResp.Errors), len(objectsToDelete), aws.StringValue(delResp.Errors[0].Key), aws.StringValue(delResp.Errors[0].Message))
```
huanjani: Love the fraction. 
I can't seem to find guidance on capitalization for multiple-sentence error messages. But I think this slight tweak in wording prevents it from reading like the first of multiple failures of that one key.
```suggestion
			return fmt.Errorf("%d/%d objects failed to delete. First failed on key %q: %s", len(delResp.Errors), len(objectsToDelete), aws.StringValue(delResp.Errors[0].Key), aws.StringValue(delResp.Errors[0].Message))
```

internal/pkg/cli/delete/static_site_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"errors"
+	"testing"
+
+	"github.com/aws/copilot-cli/internal/pkg/cli/delete/mocks"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestStaticSite_CleanResources(t *testing.T) {
+	type moks struct {
```
huanjani: on purpose???
```suggestion
	type mocks struct {
```

internal/pkg/cli/delete/static_site_test.go
```diff
@@ -0,0 +1,70 @@
...
+			},
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+
+			m := &moks{
```
huanjani:
```suggestion
			m := &mocks{
```

huanjani (approved):
Looks good to me!

internal/pkg/cli/delete/static_site_test.go
dannyrandall: yeah lol :sob: so it doesn't override the _package_ named `mocks`

internal/pkg/aws/s3/s3.go
```diff
@@ -85,7 +85,8 @@ func (s *S3) EmptyBucket(bucket string) error {
 	}
 
 	listParams := &s3.ListObjectVersionsInput{
-		Bucket: aws.String(bucket),
+		Bucket:  aws.String(bucket),
+		MaxKeys: aws.Int64(1000), // max that can be deleted in a single DeleteObjects call
```
efekarakus: what's the default? maybe we can capture that in the comment

internal/pkg/aws/s3/s3.go
```diff
@@ -111,14 +112,18 @@ func (s *S3) EmptyBucket(bucket string) error {
...
 				Objects: objectsToDelete,
+				Quiet:   aws.Bool(true), // we don't care about success values
 			},
 		})
-		if err != nil {
+		switch {
+		case err != nil:
 			return fmt.Errorf("delete objects from bucket %s: %w", bucket, err)
+		case len(delResp.Errors) > 0:
+			return fmt.Errorf("%d/%d objects failed to delete. First failed on key %q: %s", len(delResp.Errors), len(objectsToDelete), aws.StringValue(delResp.Errors[0].Key), aws.StringValue(delResp.Errors[0].Message))
```
efekarakus: Is this common having a period and starting with a capital letter in the error? what do you think of instead:
```suggestion
			return errors.Join(
				fmt.Errorf("%d/%d objects failed to delete", len(delResp.Errors), len(objectsToDelete)),
				fmt.Errorf("first failed on key %q: %s",  aws.StringValue(delResp.Errors[0].Key), aws.StringValue(delResp.Errors[0].Message)
			)
```

internal/pkg/cli/delete/delete.go
```diff
@@ -0,0 +1,12 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package delete
```
efekarakus: package comments please 🥺

internal/pkg/cli/delete/delete.go
```diff
@@ -0,0 +1,12 @@
...
+
+// NoOpDeleter does nothing.
+type NoOpDeleter struct{}
+
+// CleanResources returns nil.
+func (n *NoOpDeleter) CleanResources(app, env, wkld string) error {
```
efekarakus: 1. What do you think of renaming the package to `cleanup` or `clean`
2. What do you think of having the method signature to be just `Clean() error` and instead have the constructor take the `app, env, workload string` parameters?
   The advantage is that we could implement the interface even for Environments or Application abstractions if that makes sense 💭

internal/pkg/cli/delete/static_site.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+type bucketEmptier interface {
+	EmptyBucket(string) error
+}
+
+// StaticSiteDeleter is used to clean up resources created for a static site.
+type StaticSiteDeleter struct {
+	BucketResourceGetter bucketResourceGetter
+	BucketEmptier        bucketEmptier
```
efekarakus: nit: 
Consider making the fields private and instead having a constructor take these dependencies as inputs.

I feel like there are few advantages of the constructor route with encapsulation:
1. Constructor signatures can make it clear which dependencies are required vs. optional
2. Clients of this package cannot accidentally mess up the internal state of the struct

internal/pkg/aws/s3/s3_test.go
```diff
@@ -271,6 +282,36 @@ func TestS3_EmptyBucket(t *testing.T) {
...
+				}).Return(&s3.DeleteObjectsOutput{
+					Errors: []*s3.Error{
+						{
+							Key:     aws.String("mock/key"),
+							Message: aws.String("some error"),
+						},
+					},
+				}, nil)
+			},
+			wantErr: fmt.Errorf(`1/10 objects failed to delete. First failed on key "mock/key": some error`),
```
Lou1415926: noice

internal/pkg/cli/delete/delete.go
```diff
@@ -0,0 +1,12 @@
...
+	return nil
+}
```
Lou1415926: nit:
```suggestion
func (_ *NoOpDeleter) CleanResources(_, _, _string) error {
	return nil
}
```

internal/pkg/cli/delete/static_site_test.go
```diff
@@ -0,0 +1,70 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package delete
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/aws/copilot-cli/internal/pkg/cli/delete/mocks"
```
efekarakus: I'm totally OK with just keeping what you have 👍 

My personal preference nowadays is just to implement the test double myself and make it clear what [type of](https://samuel.karp.dev/blog/2023/02/flexible-test-doubles-in-go/) double it is.

examples:
1. https://github.com/aws/copilot-cli/blob/836be87d5ff9b2912bbf5ba97626252dac62b21b/internal/pkg/template/overrides_test.go#L62-L76
2. https://github.com/aws/copilot-cli/blob/836be87d5ff9b2912bbf5ba97626252dac62b21b/internal/pkg/template/templatetest/templatetest.go#L14

internal/pkg/cli/svc_delete.go
```diff
@@ -80,11 +89,20 @@ func newDeleteSvcOpts(vars deleteSvcVars) (*deleteSvcOpts, error) {
...
-			return cloudformation.New(session, cloudformation.WithProgressTracker(os.Stderr))
+		getSvcCFN: func(sess *awssession.Session) wlDeleter {
+			return cloudformation.New(sess, cloudformation.WithProgressTracker(os.Stderr))
 		},
-		getECR: func(session *awssession.Session) imageRemover {
-			return ecr.New(session)
+		getECR: func(sess *awssession.Session) imageRemover {
+			return ecr.New(sess)
+		},
+		newSvcDeleter: func(sess *awssession.Session, typ string) wkldDeleter {
```
efekarakus: nit:
```suggestion
		newSvcDeleter: func(sess *awssession.Session, manifestType string) wkldDeleter {
```

internal/pkg/cli/svc_delete.go
```diff
@@ -80,11 +89,20 @@ func newDeleteSvcOpts(vars deleteSvcVars) (*deleteSvcOpts, error) {
...
-		getSvcCFN: func(session *awssession.Session) wlDeleter {
-			return cloudformation.New(session, cloudformation.WithProgressTracker(os.Stderr))
+		getSvcCFN: func(sess *awssession.Session) wlDeleter {
+			return cloudformation.New(sess, cloudformation.WithProgressTracker(os.Stderr))
 		},
-		getECR: func(session *awssession.Session) imageRemover {
-			return ecr.New(session)
+		getECR: func(sess *awssession.Session) imageRemover {
+			return ecr.New(sess)
+		},
```
Lou1415926, line 97: lol why is this even in the diff

internal/pkg/cli/svc_delete.go
```diff
@@ -52,14 +60,15 @@ type deleteSvcOpts struct {
...
-	getECR    func(session *awssession.Session) imageRemover
+	store         store
+	sess          sessionProvider
+	spinner       progress
+	prompt        prompter
+	sel           configSelector
+	appCFN        svcRemoverFromApp
+	getSvcCFN     func(sess *awssession.Session) wlDeleter
+	getECR        func(sess *awssession.Session) imageRemover
+	newSvcDeleter func(sess *awssession.Session, typ string) wkldDeleter
```
efekarakus: I wonder if after the `clean` rename, if this distinction will be easier to follow:
```suggestion
	newResourceCleaner func(sess *awssession.Session, typ string) cleaner
```

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -33,8 +33,8 @@ Resources:
       AccessControl: Private
       BucketEncryption:
         ServerSideEncryptionConfiguration:
-        - ServerSideEncryptionByDefault:
-            SSEAlgorithm: AES256
+          - ServerSideEncryptionByDefault:
```
efekarakus, line 36: I feel attacked

internal/pkg/deploy/cloudformation/stack/testdata/workloads/static-site.stack.yml
```diff
@@ -211,3 +211,24 @@ Resources:
...
         - !Sub arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+
+  EnvManagerS3Access:
+    Metadata:
+      aws:copilot:description: A policy that gives the Env Manager role access to this site's S3 Bucket
+    Type: AWS::IAM::Policy
+    Properties:
+      Roles:
+        - !Sub "${AppName}-${EnvName}-EnvManagerRole"
+      PolicyName: !Sub "${WorkloadName}-S3Access"
```
Lou1415926, line 222: Does the policy name have to be unique in a region / globally?

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -236,3 +236,24 @@ Resources:
...
         - !Sub arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+  
+  EnvManagerS3Access:
+    Metadata:
+      aws:copilot:description: A policy that gives the Env Manager role access to this site's S3 Bucket
+    Type: AWS::IAM::Policy
+    Properties:
+      Roles:
+        - !Sub "${AppName}-${EnvName}-EnvManagerRole"
+      PolicyName: !Sub "${WorkloadName}-S3Access"
```
efekarakus, line 247: does this need to be unique in the account? if two environments are in the same AWS account then this policy name might clash, if so we need to namespace with !Sub "${AppName}-${EnvName}-${WorkloadName}-S3Access" (and we should limit the number of chars too)

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -236,3 +236,24 @@ Resources:
                 Resource: !Sub arn:aws:s3:::${Bucket}/*
       ManagedPolicyArns:
         - !Sub arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+  
+  EnvManagerS3Access:
+    Metadata:
+      aws:copilot:description: A policy that gives the Env Manager role access to this site's S3 Bucket
+    Type: AWS::IAM::Policy
```
efekarakus, line 243: What on earth! So this is adding an inline policy to the env manager role? I didn't know we could do this 🤯

internal/pkg/deploy/cloudformation/stack/testdata/workloads/static-site.stack.yml
dannyrandall: it needs to be unique in a region since it's attached to the `EnvManagerRole`! Separate regions have their own EnvManagerRole, so I think this should be fine.

internal/pkg/template/templates/workloads/services/static-site/cf.yml
dannyrandall: discussion here: https://github.com/aws/copilot-cli/pull/4823#discussion_r1183907396

internal/pkg/deploy/cloudformation/stack/testdata/workloads/static-site.stack.yml
Lou1415926: `EnvMangerRole` is per-env though! It's still possible for a region to have multiple envs.

internal/pkg/cli/svc_delete.go
dannyrandall: that's a great point....i have no idea :joy:

internal/pkg/aws/s3/s3.go
dannyrandall: :homer_disappear: apparently it's 1000 so i'll just remove this! https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectVersions.html#API_ListObjectVersions_RequestSyntax

internal/pkg/deploy/cloudformation/stack/testdata/workloads/static-site.stack.yml
dannyrandall: sorry, it's unique per environment - the name just has to be unique among other inline policies in `EnvManagerRole`

internal/pkg/cli/delete/static_site.go
dannyrandall: I agree with your reasons listed above for this case, but want to probe a little further: what about structs that have, say, 7 **required** fields? Beyond 5 or 6 function parameters, it's pretty normal to pass a struct for function parameters. So is this preferred to just saying all fields on `Foo` are required?

```
type Foo struct {
  // 7 required fields
}

type FooRequired struct {
  // 7 required fields
}

func New(req FooRequired) Foo {
  return Foo{
    // apply req
  }
}
```

and how about if there were additional optional fields? something like this?
```
type Foo struct {
  // 7 required fields
  // 3 optional fields
}

type FooRequired struct {
  // 7 required fields
}

type FooOption func(*Foo)

func WithOptionA(a string) FooOption {
  return func(f *Foo) {
    f.a = a
  }
}

func NewFoo(req FooRequired, opts ...FooOption) Foo {
  f := Foo{
    //apply req
  }
  
  // apply options
  return f
}
```

In both of these cases, the alternative approach in my mind is just labelling the fields, like:
```
type Foo struct {
  // A is required
  A string
  // 6 more required

  // X is optional
  X string
  // 2 more optional
}
```

I see benefits/drawbacks both ways, so I'm just curious on your thoughts?

internal/pkg/aws/s3/s3.go
huanjani: That was my fault!

internal/pkg/cli/delete/static_site_test.go
dannyrandall: i like it! i incorporated this feedback and stuff found https://google.github.io/styleguide/go/best-practices#test-double-packages-and-types - let me know what you think!

internal/pkg/cli/cleantest/cleantest.go
efekarakus, line 1: nit: should we move this package under `internal/pkg/cli/clean/cleantest`

internal/pkg/cli/svc_delete.go
```diff
@@ -41,6 +45,10 @@ var (
 	errSvcDeleteCancelled = errors.New("svc delete cancelled - no changes made")
 )
 
+type cleaner interface {
+	Clean() error
```
efekarakus, line 49: 🧹

internal/pkg/cli/clean/static_site.go
```diff
@@ -0,0 +1,56 @@
...
+
+// StaticSite is used to clean up resources created for a static site.
+type StaticSite struct {
+	app, env, svc        string
+	bucketResourceGetter bucketResourceGetter
+	bucketEmptier        bucketEmptier
+}
+
+// NewStaticSite returns an initialized static site.
+func NewStaticSite(app, env, svc string, rg bucketResourceGetter, emptier bucketEmptier) *StaticSite {
```
efekarakus, line 29: nit: I wonder if the names should be like this:
```suggestion
type StaticSiteCleaner struct {
	app, env, svc        string
	bucketResourceGetter bucketResourceGetter
	bucketEmptier        bucketEmptier
}

// StaticSite returns an initialized static site cleaner.
func StaticSite(app, env, svc string, rg bucketResourceGetter, emptier bucketEmptier) *StaticSite {
```

so that it reads: 

```go
cleaner := clean.StaticSite()
cleaner.Clean()
```

efekarakus (approved):
Looks great, just some tiny suggestions

internal/pkg/cli/delete/static_site.go
efekarakus: ugh honestly I think the last option: 
```
type Foo struct {
  // A is required
  A string
  // 6 more required

  // X is optional
  X string
  // 2 more optional
}
```
Is only OK if actually all the fields are _optional_, i.e. the struct internally can call an [`init` method](https://github.com/robpike/ivy/blob/f066ab6d0b3dd95e6cc5fce39ab5f2a3ffd6f9a0/config/config.go#L76) and set the values. Otherwise, I'd avoid this option.

My preferred option is actually like this when there are many parameters:
```go
type Foo struct {
  // 7 required fields
  // 3 optional fields
}

type FooConfig struct {
  // 7 required fields
}

type FooOpts struct {
  // 3 optional fields.
}

func NewFoo(cfg FooConfig) Foo
func NewFooWithOpts(cfg FooConfig, opts FooOpts) Foo
```

PR #4824, opened 2023-05-02
chore(cli): add `--sources` flag for static sites

huanjani:


internal/pkg/cli/svc_init.go
```diff
@@ -263,6 +264,19 @@ func (o *initSvcOpts) Validate() error {
 			return err
 		}
 	}
+	if len(o.sourcePaths) != 0 {
+		if o.wkldType != manifestinfo.StaticSiteType {
+			return fmt.Errorf("'--%s' must be specified with '--%s %s'", sourcesFlag, typeFlag, manifestinfo.StaticSiteType)
```
```suggestion
			return fmt.Errorf("'--%s' must be specified with '--%s %q'", sourcesFlag, typeFlag, manifestinfo.StaticSiteType)
```
to match the help suggestion

internal/pkg/cli/svc_init.go
```diff
@@ -346,7 +360,22 @@ func (o *initSvcOpts) Execute() error {
 	if err != nil {
 		return err
 	}
-	manifestPath, err := o.init.Service(&initialize.ServiceProps{
+	var manifestPath string
```
dannyrandall: i think we could just get rid of this and instead do
```
o.manifestPath, err = o.init.Service()
```

internal/pkg/cli/svc_init.go
```diff
@@ -346,7 +360,22 @@ func (o *initSvcOpts) Execute() error {
...
+	if o.wkldType == manifestinfo.StaticSiteType {
```
dannyrandall: why can't we just pass the zero value fields to `o.init.Service()` in the case of static site?

internal/pkg/cli/svc_init.go
```diff
@@ -263,6 +264,19 @@ func (o *initSvcOpts) Validate() error {
 			return err
 		}
 	}
+	if len(o.sourcePaths) != 0 {
+		if o.wkldType != manifestinfo.StaticSiteType {
+			return fmt.Errorf("'--%s' must be specified with '--%s %s'", sourcesFlag, typeFlag, manifestinfo.StaticSiteType)
+		}
+		if err := validateStaticSiteSources(o.fs, o.sourcePaths); err != nil {
+			return err
+		}
```
dannyrandall: is validating a bit redundant? like the error coming from `convertStringsToAssets()` will be pretty similar if a file doesn't exist, right?

dannyrandall (approved):

internal/pkg/cli/flag.go
```diff
@@ -181,6 +182,8 @@ Cannot be specified with --%s or --%s.`, dockerFileFlag, dockerFileContextFlag)
 Cannot be specified with --%s.`, imageFlag)
 	dockerFileContextFlagDescription = fmt.Sprintf(`Path to the Docker build context.
 Cannot be specified with --%s.`, imageFlag)
+	sourcesFlagDescription = fmt.Sprintf(`List of paths to source directories or files for a Static Site.
```
efekarakus: nit: can we say 
```suggestion
	sourcesFlagDescription = fmt.Sprintf(`List of relative paths to source directories or files for a Static Site.
```
Perhaps we can even do:
```suggestion
	sourcesFlagDescription = fmt.Sprintf(`List of relative paths to source directories or files.
```
Since the "must be specified" already mentions Static Site

efekarakus (approved):
:shipit: with the do-not-merge label for a small suggestion

internal/pkg/cli/svc_init.go
```diff
@@ -383,7 +394,9 @@ func (o *initSvcOpts) RecommendActions() error {
 
 func (o *initSvcOpts) askSvcDetails() error {
 	if o.wkldType == manifestinfo.StaticSiteType {
-		return o.askStaticSite()
+		if len(o.staticAssets) == 0 {
+			return o.askStaticSite()
+		}
 	}
```
Lou1415926: nit: what about about
```go
if o.wkldType == manifestinfo.StaticSiteType {
   return o.askStaticSite()
}
func (o *initSvcOpts) askStaticSite() error {
   if len(o.staticAssets) == 0 { return nil }
}
```

Lou1415926 (approved):

PR #4831, opened 2023-05-03
chore: add cfn template for cf static site

iamhopaul123:


internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -42,27 +42,6 @@ Resources:
         Rules:
           - ObjectOwnership: BucketOwnerEnforced
 
-  BucketPolicy:
```
iamhopaul123, line 45: Remove for now as the bucket is managed by us and will always be private.

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -153,3 +158,32 @@ CertificateReplicator:
     CertificateArn: !Ref HTTPSCert
 {{- end}}
 {{- end}}
+
+{{- if .CDNConfig.Static}}
+{{- if .CDNConfig.Static.StaticSiteBucket}}
```
Lou1415926: nit: 
```suggestion
{{- and .CDNConfig.Static (.CDNConfig.Static.StaticSiteBucket)}}
```

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -153,3 +158,32 @@ CertificateReplicator:
...
+{{logicalIDSafe (bucketNameFromURL .CDNConfig.Static.StaticSiteBucket)}}BucketPolicyForCloudFront:
+  Metadata:
+    'aws:copilot:description': 'A bucket policy to grant CloudFront read access to the Static Site bucket "{{bucketNameFromURL .CDNConfig.Static.StaticSiteBucket}}"'
+  Type: AWS::S3::BucketPolicy
+  DeletionPolicy: Retain
```
Lou1415926: oo why do we need to retain this policy?

internal/pkg/cli/task_run_test.go
```diff
@@ -1062,7 +1062,7 @@ func TestTaskRunOpts_Execute(t *testing.T) {
 				m.store.EXPECT().GetApplication("my-app").Return(&config.Application{Name: "my-app"}, nil).Times(2)
 				m.deployer.EXPECT().GetTaskStack(inGroupName).Return(&info, nil)
 				key := "manual/env-files/magic.env/4963d64294508aa3fa103ccac5ad1537944c577d469608ddccad09b6f79b6406.env"
-				arn := "arn:aws:s3:::bigbucket/" + key
+				arn := "https://bigbucket.s3-us-west-2.amazonaws.com/" + key
```
Lou1415926: Is `Upload` expected to return an URL, instead of ARN now? If so, can we also rename the variable to `bucketURL`?

internal/pkg/cli/deploy/env.go
```diff
@@ -436,15 +437,18 @@ func (d *envDeployer) buildStackInput(in *DeployEnvironmentInput) (*cfnstack.Env
 }
 
 func (d *envDeployer) renderStaticSite(mft *manifest.Environment) error {
-	if mft == nil || mft.CDNConfig.Config.Static.Location.StaticSite == "" {
+	if mft == nil || mft.CDNConfig.Config.Static.Location.StaticBucket != "" ||
```
Lou1415926, line 444: plz excuse my ignorance - can `StaticBucket != ""` and `StaticSite != ""` be `true` at the same time, for the same location?

internal/pkg/aws/s3/s3.go
```diff
@@ -214,21 +216,22 @@ func (s *S3) upload(bucket, key string, buf io.Reader) (string, error) {
...
 func parseS3URI(uri string) (bucket, key string, err error) {
 	parsed := strings.SplitN(strings.TrimPrefix(uri, s3URIPrefix), "/", 2)
-	if len(parsed) != 2 {
+	bucket = parsed[0]
+	if bucket == "" {
 		return "", "", fmt.Errorf("cannot parse S3 URI %s into bucket name and key", uri)
 	}
-	return parsed[0], parsed[1], nil
+	if len(parsed) == 2 {
+		key = parsed[1]
```
dannyrandall, line 232: what does this change? was it panicking with a uri that didn't have a key?

internal/pkg/aws/s3/s3.go
```diff
@@ -214,21 +216,22 @@ func (s *S3) upload(bucket, key string, buf io.Reader) (string, error) {
...
+	return
 }
 
-// parseObjectURL parses the bucket name and object key from a [virtual-hosted-style access URL].
+// parseHTTPURL parses the bucket name and optional key from a [virtual-hosted-style access URL].
 // For example: https://DOC-EXAMPLE-BUCKET1.s3.us-west-2.amazonaws.com/puppy.png
 //
 // [virtual-hosted-style access URL]: https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-bucket-intro.html#virtual-host-style-url-ex
-func parseObjectURL(url string) (bucket, key string, err error) {
+func parseHTTPURL(url string) (bucket, key string, err error) {
```
dannyrandall:
```suggestion
func parseHTTPURI(url string) (bucket, key string, err error) {
```
nit: to match the `parseS3URI` naming?

internal/pkg/aws/s3/s3_test.go
```diff
@@ -322,6 +322,11 @@ func TestS3_ParseURL(t *testing.T) {
 			wantedBucketName: "stackset-myapp-infrastru-pipelinebuiltartifactbuc-1nk5t9zkymh8r",
 			wantedKey:        "scripts/dns-cert-validator/dd2278811c3",
 		},
+		"parses bucket URL": {
```
dannyrandall, line 361:
```suggestion
		"parses bucket URL with no key": {
```

internal/pkg/deploy/cloudformation/stack/backend_svc_test.go
```diff
@@ -41,28 +41,6 @@ func TestBackendService_Template(t *testing.T) {
 		fs = realEmbedFS
 	})
 	fs = templatetest.Stub{}
-
-	t.Run("returns a wrapped error when addons template parsing fails", func(t *testing.T) {
```
dannyrandall: why get rid of this test?

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -96,7 +96,12 @@ CloudFrontDistribution:
             {{- end}}
       {{- if .CDNConfig.Static}}
         - Id: !Sub 'copilot-${AppName}-${EnvironmentName}-s3'
+        {{- if .CDNConfig.Static.ImportedBucket}}
           DomainName: {{.CDNConfig.Static.ImportedBucket}}
+        {{- end}}
+        {{- if .CDNConfig.Static.StaticSiteBucket}}
```
dannyrandall, line 102: can they only have a single static site service exposed via cloudfront? or can we support multiple origins?

internal/pkg/template/templates/environment/partials/cdn-resources.yml
```diff
@@ -153,3 +158,32 @@ CertificateReplicator:
...
+{{- if .CDNConfig.Static}}
+{{- if .CDNConfig.Static.StaticSiteBucket}}
+{{logicalIDSafe (bucketNameFromURL .CDNConfig.Static.StaticSiteBucket)}}BucketPolicyForCloudFront:
+  Metadata:
+    'aws:copilot:description': 'A bucket policy to grant CloudFront read access to the Static Site bucket "{{bucketNameFromURL .CDNConfig.Static.StaticSiteBucket}}"'
+  Type: AWS::S3::BucketPolicy
+  DeletionPolicy: Retain
+  Condition: CreateALB
+  Properties:
+    Bucket: {{bucketNameFromURL .CDNConfig.Static.StaticSiteBucket}}
```
dannyrandall, line 169: will this cause the static site to fail to delete? because the policy, owned by this stack, hasn't been deleted?

or, will the env stack fail to delete/update after the static site is deleted, because this resource has a drift (if it got deleted along with the bucket)?

internal/pkg/template/templates/environment/partials/cdn-resources.yml
iamhopaul123: you are right. i don't think we need to

internal/pkg/cli/task_run_test.go
iamhopaul123: it always expects a url...it is a bad test in the past. i'll fix the var name too

internal/pkg/cli/deploy/env.go
iamhopaul123: i don't think so because they unmarshal from the same field https://github.com/aws/copilot-cli/blob/50989ce59f74fab653b6864c87dbb0141d37a0bd/internal/pkg/manifest/env.go#L298

iamhopaul123: And `StaticBucket` is the rendered bucket URL value for `StaticSite`. I just don't want to overwrite the value for `StaticSite` as it's not an idempotent behavior

internal/pkg/aws/s3/s3.go
iamhopaul123: yeah and it didn't panic for what it should panic like `https://.s3.us-west-2.....` when the bucket name is empty

internal/pkg/aws/s3/s3_test.go
iamhopaul123: bucket URL won't have key lol. it's how it is different from object url

internal/pkg/deploy/cloudformation/stack/backend_svc_test.go
iamhopaul123: an accident :homer-disappearing:

internal/pkg/template/templates/environment/partials/cdn-resources.yml
iamhopaul123: not for this release. we can make it a union type for the env manifest ad hoc

internal/pkg/template/templates/environment/partials/cdn-resources.yml
iamhopaul123: I've tested out and it's not tracked by drift detection and won't cause the workload/env stack deletion failure

internal/pkg/aws/s3/s3_test.go
dannyrandall: 🤦

dannyrandall (approved):

huanjani (approved):
:shipit:

PR #4833, opened 2023-05-03
chore: add lifecycle rules to assets in artifact bucket

dannyrandall:
This also addresses some nit feedback from https://github.com/aws/copilot-cli/pull/4823.


internal/pkg/cli/clean/static_site.go
```diff
@@ -18,16 +18,16 @@ type bucketEmptier interface {
...
 	app, env, svc        string
 	bucketResourceGetter bucketResourceGetter
 	bucketEmptier        bucketEmptier
 }
 
-// NewStaticSite returns an initialized static site.
-func NewStaticSite(app, env, svc string, rg bucketResourceGetter, emptier bucketEmptier) *StaticSite {
-	return &StaticSite{
+// StaticSite returns an initialized static site cleaner.
+func StaticSite(app, env, svc string, rg bucketResourceGetter, emptier bucketEmptier) *StaticSiteCleaner {
```
iamhopaul123, line 29: Would `NewStaticSiteCleaner` make more sense for the function name🤔

iamhopaul123 (approved):
awesome！

internal/pkg/cli/clean/static_site.go
dannyrandall: :sob: https://github.com/aws/copilot-cli/pull/4823#discussion_r1184413903

iamhopaul123: :homer-disappearing:

cf-custom-resources/lib/copy-assets.js
```diff
@@ -11,11 +11,27 @@ const aws = require("aws-sdk");
...
 
-	const params = {
+	try {
+		await s3.headObject({
+			Bucket: event.destBucket,
+			Key: event.mapping.destPath
+		}).promise();
+
+		// if headObject is successful, the file already exists
+		// in destBucket. no need to copy, so we're done!
```
huanjani: just checking: what about in an 'update' situation? the file exists but a newer version needs to be copied....

dannyrandall: good catch😅 let me see what we can do about that!

dannyrandall: ok i think i'm actually going to revert these changes.

we _could_ add a `headObject` call to the source object, and only do the upload if the destination file does not exist (what we have now) _**or**_ if the [files etag's](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html) are mismatched. however, since [copying data between two s3 buckets the same region is free](https://aws.amazon.com/s3/pricing/), i think it's simpler & faster to just remove that logic and let s3 do the copy even if it's unnecessary.

huanjani (approved):
:shipit:

internal/pkg/cli/deploy/svc.go
```diff
@@ -98,20 +97,37 @@ type svcStackConfigurationOutput struct {
...
 
-func validateAppVersionForAlias(appName string, appVersionGetter versionGetter) error {
+type errAppOutOfDate struct {
+	svc           string
+	curVersion    string
+	neededVersion string
+}
+
+func (e *errAppOutOfDate) Error() string {
+	return fmt.Sprintf("app version must be > %s", e.neededVersion)
```
Lou1415926, line 107: Should this be >= 🤔

Lou1415926 (commented):
me late again!

internal/pkg/cli/deploy/svc.go
dannyrandall: fixed in https://github.com/aws/copilot-cli/pull/4850!

PR #4839, opened 2023-05-04
chore(diff): handle `GetAtt` comparison

Lou1415926:


internal/pkg/template/diff/override.go
```diff
@@ -40,58 +43,185 @@ func (m *ignorer) parse(_, _ *yaml.Node, _ string, _ overrider) (diffNode, error
...
-		return nil, nil
+		return nil, err
 	}
 	return &keyNode{
 		keyValue:   key,
 		childNodes: []diffNode{diff},
 	}, nil
 }
 
+func intrinsicFuncName(node *yaml.Node) string {
```
iamhopaul123, line 211: I feel like it would be safer for us to model thing like "!Join" and "Fn::Join" as different types (2 types initiated by calling regex matching and error out earlier if one doesn't match) and call methods to get func name or convert into the other type 💭

internal/pkg/template/diff/override.go
```diff
@@ -40,58 +43,185 @@ func (m *ignorer) parse(_, _ *yaml.Node, _ string, _ overrider) (diffNode, error
...
+	intrinsicFuncMapTagConverter
+}
+
+// match returns true if both from node and to node are calling the "GetAtt" intrinsic function.
+// "GetAtt" only accepts either sequence or scalar, therefore match returns false if either of from and to has invalid 
+// input node to "GetAtt".
+// Example1: "!GetAtt" and "!GetAtt" returns true.
+// Example2: "!GetAtt" and "Fn::GetAtt" returns true.
+// Example3: "!Ref" and "!GetAtt" returns false.
+// Example4: "!GetAtt [a,b]" and "Fn::GetAtt:a:b" returns false because the input type is wrong.
```
iamhopaul123: What would happen if we don't deal with this separately 🤔 To me at least for "GetAtt" part, `"!GetAtt [a,b]" and "Fn::GetAtt:a:b" `seem to be identical and we only need to show them how their parameters are different from each other.

internal/pkg/template/diff/override.go
Lou1415926: That is an option but that'd be like 20+ overriders to process for each YAML nodes :(, so I didn't opt for this route.

> it would be safer

I agree that the the logics that handle each intrinsic functions will be more decoupled this way, but why is it necessarily safer? 

> error out earlier if one doesn't match

Can you elaborate on this?

dannyrandall: i think i'm partial to just the `TrimPrefix`, but i wonder if instead of a function that trims _both_ "Fn::" and "!" we should explicitly trim the correct value - if it's in the tag form, won't it always mean we should strip "!"? And "Fn::" in the long form?

internal/pkg/template/diff/override.go
Lou1415926: `Fn::GetAtt: [a,b]` and `Fn::GetAtt: a.b` is identical and valid.

But  ⬇️ is invalid. `Fn::GetAtt` the intrinsic function does not accept a map like this.

```yaml
Fn::GetAtt:
   a: b
```

Therefore, `match` shouldn't return true in this case, because either the user wrote something invalid, or `Fn::GetAtt` is not meant to be an intrinsic function, and is instead a key name that happens to have the same name as the intrinsic func.

internal/pkg/template/diff/override.go
```diff
@@ -40,58 +43,185 @@ func (m *ignorer) parse(_, _ *yaml.Node, _ string, _ overrider) (diffNode, error
...
+// Example2: "!Ref" and "!Ref" will return false because they are written in the same form (i.e. short).
+// Example3: "!Ref" and "Fn::GetAtt:" will return false because they are different intrinsic functions.
+// For more on intrinsic functions and full/short forms, read https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-ToJsonString.html.
+func (converter *intrinsicFuncMapTagConverter) match(from, to *yaml.Node, key string, overrider overrider) bool {
+	if !converter.intrinsicFunc.match(from, to, key, overrider) {
 		return false
 	}
-	return eqIntrinsicFunc(intrinsicFuncFullFormName(fullFormNode), shortFormNode.Tag)
+	// Exactly one of from and to is full form.
+	return (from.Kind == yaml.MappingNode || to.Kind == yaml.MappingNode) && (from.Kind != to.Kind)
```
dannyrandall, line 111: is there a case you know of where one of these would _not_ be a `MappingNode` but also `from.Kind != to.Kind`?

internal/pkg/template/diff/override.go
```diff
@@ -40,58 +43,185 @@ func (m *ignorer) parse(_, _ *yaml.Node, _ string, _ overrider) (diffNode, error
...
+// Note that this does not hold for "GetAtt" function: "!GetAtt: [1,2]" and "!GetAtt: 1.2" should be considered the same.
+// parse assumes that from and to are matched by intrinsicFuncMapTagConverter.
+func (*intrinsicFuncMapTagConverter) parse(from, to *yaml.Node, key string, overrider overrider) (diffNode, error) {
 	var diff diffNode
 	var err error
 	if from.Kind == yaml.MappingNode {
 		// The full form mapping node always contain only one child node. The second element in `Content` is the 
 		// value of the child node. Read https://www.efekarakus.com/2020/05/30/deep-dive-go-yaml-cfn.html.
-		diff, err = parse(from.Content[1], stripTag(to), intrinsicFuncFullFormName(from), overrider)
+		diff, err = parse(from.Content[1], stripTag(to), intrinsicFunctions[intrinsicFuncName(from)].fullName, overrider)
```
dannyrandall: would this panic in the case
```
SomeValue
  Fn::Sub:
```

internal/pkg/template/diff/override.go
Lou1415926: I think the gist here is that I try to make the radius of these overriders as small as possible. 

I think of this in terms of false positive vs. false negative. A false positive is "when there is no diff but we say there is", and false negative is "when there is a diff but we fail to detect it".

It's annoying for users to see false positive, but it's dangerous if there are a lot of false negative (because they may have missed some important diff)

These overriders are written to reduce false positive (improve UX), but they must not increase false negative.  This is why i try to make `match` as strict as possible.

dannyrandall:
```suggestion
// Example4: "!GetAtt [a,b]" and "Fn::GetAtt: a:b" returns false because the input type is wrong.
```

internal/pkg/template/diff/override.go
```diff
@@ -40,58 +43,185 @@ func (m *ignorer) parse(_, _ *yaml.Node, _ string, _ overrider) (diffNode, error
...
+		return false
+	}
+	fromValue, toValue := from, to
+	if from.Kind == yaml.MappingNode {
+		fromValue = from.Content[1] // A valid full-form intrinsic function always contain a child node.
+	}
+	if to.Kind == yaml.MappingNode {
+		toValue = to.Content[1]
+	}
+	switch {
```
dannyrandall: is this the same? i.e., `fromValue` and `toValue` both need to be either a scalar or sequence node
```suggestion
	if (fromValue.Kind == yaml.ScalerNode || fromValue.Kind == yaml.SequenceNode) && (toValue.Kind == yaml.ScalerNode || toValue.Kind == yaml.SequenceNode)
```

internal/pkg/template/diff/override.go
```diff
@@ -101,28 +231,15 @@ func stripTag(node *yaml.Node) *yaml.Node {
...
-	"Fn::Split":       "!Split",
-	"Fn::Sub":         "!Sub",
-	"Fn::Transform":   "Transform",
-}
-
-func eqIntrinsicFunc(fullFormName, shortFormName string) bool {
-	expectedShort, ok := intrinsicFuncFull2Short[fullFormName]
-	return ok && shortFormName == expectedShort
+// Transform scalar node "LogicalID.Attr" to sequence node [LogicalID, Attr].
+func getAttScalarToSeq(scalarNode *yaml.Node) (*yaml.Node, error) {
```
dannyrandall, line 239: i wonder if it's easier to compare the other direction? (seq -> scalar)? bc then it's just 
```
strings.Join(seqNode.Value, ".")
```
right? or am i missing something 🤔

Lou1415926: > For the Fn::GetAtt attribute name, you can use the Ref function. [doc](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-getatt.html#getatt-supported-functions)

It would have been OK if `GetAtt` only accepts strings for the "Attributes" part : (


```yaml
# from
Fn::GetAtt: a.b
# to
Fn::GetAtt:
  - a
  - !Ref BAttr
```

I'm not sure how to convert `[a, !Ref BAttr]` to a scalar 🥲

internal/pkg/template/diff/override.go
Lou1415926: Currently, I don't think there is! If a function is to be full-form, it must have been written as `Fn::FuncName:<something>`, which is innately a map, no matter what type of data `<something>` is.

internal/pkg/template/diff/override.go
Lou1415926: Tested - not panicking!

Are you concerned of `from.Content[1]`? When the yaml looks like your example, `from`'s `Content` would still have a length of two, and its second element (i.e. the value to the key `Fn::Sub`) would be `&{8 0 !!null   <nil> []    <line> <col>}`

internal/pkg/template/diff/override.go
Lou1415926: yeah! updated!

internal/pkg/template/diff/override.go
iamhopaul123: > I agree that the the logics that handle each intrinsic functions will be more decoupled this way, but why is it necessarily safer?

Because we make less assumptions than trimming prefix(assuming we can't get anything similar to CFN intrinsic function)? Regex almost guarantees that we only accepts things in certain format and reject earlier if one candidate doesn't match any of the rules. But overall i'm ok with this implementation too!

internal/pkg/template/diff/override.go
dannyrandall: ok, awesome! yeah, that was the case i was thinking about. sounds good 👍

internal/pkg/template/diff/override.go
dannyrandall: ah, rip. of course, the nested functions :(

internal/pkg/template/diff/override.go
Lou1415926: ah I think I get it now. I understood this comment wrong!

With the current implementation, the `override` would  unexpectedly `match` a snippet like this:
```yaml
Sub: 123
```

In this example, `Sub` is either written incorrectly, or is just the name of a field. We should only match `Fn::Sub` for a full form. 

Good catch!

I will combine @iamhopaul123 and @dannyrandall 's suggestion 👍🏼

Lou1415926: done!

internal/pkg/template/diff/override.go
```diff
@@ -40,58 +47,173 @@ func (m *ignorer) parse(_, _ *yaml.Node, _ string, _ overrider) (diffNode, error
...
+func intrinsicFuncName(node *yaml.Node) string {
+	if node.Kind != yaml.MappingNode {
+		return strings.TrimPrefix(intrinsicFuncShortFormRegex.FindString(node.Tag), "!")
+	}
+	if len(node.Content) != 2 {
+		// The full form mapping node always contain only one child node, whose key is the func name in full form.
+		// Read https://www.efekarakus.com/2020/05/30/deep-dive-go-yaml-cfn.html.
+		return ""
+	}
+	return strings.TrimPrefix(intrinsicFuncFullFormRegex.FindString(node.Content[0].Value), "Fn::")
```
dannyrandall: i think this would end up returning the same thing for these two values of `node.Content[0].Value`: `Fn::Foo` and `Fn::Foo-bar`. i'm probably just being pedantic, but i wonder if we should use the regex to see if the value _matches_ our expected form, and then just trim `Fn::` off of that?

Lou1415926: > returning the same thing 

There is a `$` anchor at the end of regex though, so `Fn::Foo-bar` would give `Fn::Foo-bar` (it can't partially match `"Fn::Foo"` because of the `$`). Does this solve your concern or could you elaborate 👁️ ?

dannyrandall:
```
reg := regexp.MustCompile(`^Fn::[A-Z][A-Za-z\d]*|Ref$`)
fmt.Printf("str: %q\n", reg.FindString("Fn::Foo-bar"))
fmt.Printf("str: %q\n", reg.FindString("Fn::Foo"))
```

tbh i don't _super_ know how it works, but this prints:
```
str: "Fn::Foo"
str: "Fn::Foo"
```
but you're right, you have the `$`, so you would think that `Fn::Foo-bar` should return `""`...

Lou1415926: I see lol perhaps it's parsing as "either start with Fn::xxx"  or "Ref and end it"

dannyrandall: i think it's something weird with the `|` and different capture groups. I tried:
```
reg := regexp.MustCompile(`^((Fn::[A-Z][A-Za-z\d]*)|(Ref))$`)
fmt.Printf("str: %q\n", reg.FindString("Fn::Foo-bar"))
fmt.Printf("str: %q\n", reg.FindString("Fn::Foo"))
fmt.Printf("match: %v\n", reg.MatchString("Fn::Foo-bar"))
fmt.Printf("match: %v\n", reg.MatchString("Fn::Foo"))
```

output:
```
str: ""
str: "Fn::Foo"
match: false
match: true
```

Lou1415926: yeah I think it wouldn't match `-bar` because there is a dash there, and dash isn't matched by `[A-Za-z\d]`

Edit: oh but that doesn't explain why `reg.MatchString("Fn::Foo-bar"))` is `false`...

dannyrandall: `Fn::Foo-bar` shouldn't match! like you said, `-` is not matched by `[A-Za-z\d]`, and the regex requires `[A-Za-z\d]` until the _end of the line_.

internal/pkg/template/diff/override.go
```diff
@@ -40,58 +43,169 @@ func (m *ignorer) parse(_, _ *yaml.Node, _ string, _ overrider) (diffNode, error
 	return nil, nil
 }
 
-// intrinsicFuncFullShortFormConverter handles comparison between full/short form of an intrinsic function.
-type intrinsicFuncFullShortFormConverter struct{}
+// Check https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html for
+// a complete list of intrinsic functions. Some are not included here as they do not need an overrider.
+var (
+	exists                     = struct{}{}
+	intrinsicFunctionFullNames = map[string]struct{}{"Ref": exists, "Fn::Base64": exists, "Fn::Cidr": exists,
```
dannyrandall: gofmt? maybe on separate lines?

dannyrandall (approved):
great work!!!

iamhopaul123 (approved):

PR #4840, opened 2023-05-04
chore: enable s3-website url for static site bucket

dannyrandall:


internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -41,6 +41,8 @@ Resources:
       OwnershipControls:
         Rules:
           - ObjectOwnership: BucketOwnerEnforced
+      WebsiteConfiguration:
+        IndexDocument: index.html
```
efekarakus, line 45: Is this going to be configurable?

PR #4847, opened 2023-05-08
chore(s3): upload should guess content type as possible

iamhopaul123:


internal/pkg/aws/s3/s3.go
```diff
@@ -218,6 +221,15 @@ func (s *S3) upload(bucket, key string, buf io.Reader) (string, error) {
 	return resp.Location, nil
 }
 
+func defaultContentTypeFromExt(key string) *string {
+	contentType := mime.TypeByExtension(filepath.Ext(key))
+	if contentType == "" {
+		return nil
```
Lou1415926, line 227: aye just curious - do you know if s3 tries to determine the content type when it's ""? Or does it just default to something.

internal/pkg/aws/s3/s3.go
```diff
@@ -218,6 +221,15 @@ func (s *S3) upload(bucket, key string, buf io.Reader) (string, error) {
...
+	}
+	fmt.Println(contentType)
```
Lou1415926: :doge Just to be annoying and point out trivial things

Lou1415926 (approved):

internal/pkg/aws/s3/s3.go
iamhopaul123: i don't know about an empty string but if it is nil s3 does have a default value (which mostly doesn't make any sense and the link will download the object)

internal/pkg/aws/s3/s3.go
iamhopaul123: :homer-disappearing:

huanjani (approved):
This is cool!

efekarakus (commented):
noice

PR #4848, opened 2023-05-08
chore(manifest): validate source(s) in Static Site manifest

huanjani:


internal/pkg/manifest/validate.go
```diff
@@ -588,6 +589,24 @@ func (s StaticSiteConfig) validate() error {
 }
 
 func (f FileUpload) validate() error {
+	if err := f.validateSource(f.Source); err != nil {
+		return err
+	}
+	return nil
```
```suggestion
        return f.validateSource(f.Source)
```

internal/pkg/deploy/cloudformation/stack/testdata/workloads/static-site-manifest.yml
```diff
@@ -1,16 +1,16 @@
...
+  - source: ./frontend/dist
     destination: /dist
   - source: ./foo 
+    recursive: true
     destination: /code 
-    include: "*.jpg"
-    exclude: "*"
\ No newline at end of file
+    exclude: "*"
+    reinclude: "*.jpg"
```
Lou1415926: qq: which part of the code change in this PR lead to this change in `testdata` ?

internal/pkg/manifest/validate.go
```diff
@@ -588,6 +589,24 @@ func (s StaticSiteConfig) validate() error {
...
+}
+
+// validateSource returns nil if Source is configured correctly.
+func (f FileUpload) validateSource(source string) error {
```
Lou1415926: do you think we can read `f.Source` directly instead of passing `source` as a param?
```go
func (f FileUpload) validateSource() error {
  if f.Source == "" {}
}
```

internal/pkg/deploy/cloudformation/stack/static_site_integration_test.go
```diff
@@ -64,6 +66,15 @@ func TestStaticSiteService_TemplateAndParamsGeneration(t *testing.T) {
 					Name: &envName,
 				},
 			}
+
+			// Create in-memory mock file system.
```
Lou1415926: dumb q! where is the mocked `fs` used in this test case?

internal/pkg/manifest/validate.go
```diff
@@ -588,6 +589,24 @@ func (s StaticSiteConfig) validate() error {
...
+	if source == "" {
+		return &errFieldMustBeSpecified{
+			missingField: "source",
+		}
+	}
+	fs := afero.NewOsFs()
+	_, err := fs.Stat(f.Source)
+	if err != nil {
+		return fmt.Errorf("source '%s' must be a valid path", f.Source)
+	}
```
Lou1415926: Can we guarantee that the `err` can always be attributed to the path being invalid? 

In addition, small nit:
```suggestion
	if err != nil {
		return fmt.Errorf("source '%s' must be a valid path: %w", f.Source, err)
	}
```

Lou1415926 (commented):
LGTM! Some small questions and nits!

internal/pkg/manifest/validate.go
Lou1415926: > Stat returns the FileInfo structure describing file. If there is an error, it will be of type *PathError.

Oh I see from https://github.com/spf13/afero/blob/d92c300a1b6eaa1690b7aecab2ef7fd0d64a94e1/os.go#L86 and https://pkg.go.dev/os#File.Stat that the err will always be a `*PathError` for now 👍🏼

internal/pkg/deploy/cloudformation/stack/testdata/workloads/static-site-manifest.yml
huanjani: The `yaml` field and template always had `reinclude`, actually. It was just wrong here in the `testdata`.

internal/pkg/deploy/cloudformation/stack/static_site_integration_test.go
huanjani: resolved offline

internal/pkg/cli/deploy/static_site.go
```diff
@@ -149,6 +152,16 @@ func (d *staticSiteDeployer) stackConfiguration(in *StackRuntimeConfiguration) (
 	return conf, nil
 }
 
+func (d *staticSiteDeployer) validateSources() error {
+	for _, upload := range d.staticSiteMft.FileUploads {
+		_, err := d.fs.Stat(upload.Source)
+		if err != nil {
+			return fmt.Errorf("source '%s' must be a valid path: %w", upload.Source, err)
```
Lou1415926: nit:
```suggestion
			return fmt.Errorf("source %q must be a valid path: %w", upload.Source, err)
```

internal/pkg/deploy/cloudformation/stack/testdata/workloads/static-site-manifest.yml
Lou1415926: aye then I wonder why the test wasn't failing before :panic

Lou1415926 (approved):
lgtm! perhaps we can add one unit test case for the `cli/deploy` package - but otherwise looks great!!

internal/pkg/deploy/cloudformation/stack/static_site_integration_test.go
```diff
@@ -45,7 +45,7 @@ func TestStaticSiteService_TemplateAndParamsGeneration(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			ctrl := gomock.NewController(t)
 			defer ctrl.Finish()
-
+
```
dannyrandall: wait i don't think this should be here 🤔

dannyrandall (approved):
looks great!

PR #4850, opened 2023-05-08
chore: add static site alias field

dannyrandall:


internal/pkg/cli/deploy/static_site.go
```diff
@@ -143,3 +148,20 @@ func (d *staticSiteDeployer) stackConfiguration(in *StackRuntimeConfiguration) (
...
 	return conf, nil
 }
+
+func (d *staticSiteDeployer) validateAlias() error {
+	if d.staticSiteMft.Alias == "" {
+		return nil
+	}
+
+	hasImportedCerts := len(d.envConfig.HTTPConfig.Public.Certificates) != 0
+	if hasImportedCerts {
```
Lou1415926, line 158: I wonder if people can still technically use an alias under `*.app.domain.com`, that has nothing to do with the env 🤔

Lou1415926 (approved):

internal/pkg/cli/deploy/static_site.go
dannyrandall: that's a great question... maybe for now i'll leave it as is (matching NLB logic) but if we have extra time we can see if it works with the cert lambda.

internal/pkg/cli/deploy/svc.go
```diff
@@ -131,3 +133,68 @@ func validateMinAppVersion(app, svc string, appVersionGetter versionGetter, minV
...
+	alias string
+	app   *config.Application
+	env   string
+}
+
+func (e *errInvalidAlias) Error() string {
+	return fmt.Sprintf("alias %q is not supported in hosted zones managed by Copilot", e.alias)
+}
+
+func (e *errInvalidAlias) RecommmendActions() string {
```
iamhopaul123, line 186: awesome!

iamhopaul123 (approved):

internal/pkg/cli/deploy/static_site.go
```diff
@@ -143,3 +148,20 @@ func (d *staticSiteDeployer) stackConfiguration(in *StackRuntimeConfiguration) (
...
+		return fmt.Errorf("cannot specify alias when env %q imports one or more certificates", d.env.Name)
```
efekarakus, line 159: :wut: so the imported certificates on ELB limits if you can use `alias` for a static site?

Maybe I'm miss understanding, but my expectation would be that users can specify both `alias` and `certs` in a `Static Site` manifest

internal/pkg/cli/deploy/svc.go
```diff
@@ -104,7 +106,7 @@ type errAppOutOfDate struct {
 }
 
 func (e *errAppOutOfDate) Error() string {
-	return fmt.Sprintf("app version must be > %s", e.neededVersion)
+	return fmt.Sprintf("app version must be >= %s", e.neededVersion)
```
efekarakus, line 109: :doge:

internal/pkg/cli/deploy/static_site.go
dannyrandall: you're 100% right that it doesn't make sense - i was just thinking to start off with the same restrictions we have for NLB until we can validate (and fix, if necessary) that the cert lambda will work with imported certificates.

also to clarify - there is no `certs` field in the Static Site manifest right now - these are env level certs.

PR #4853, opened 2023-05-09
chore: make s3 object copy depend on cfn success

dannyrandall:

### DIFF
--- a/internal/pkg/template/templates/workloads/services/static-site/cf.yml
+++ b/internal/pkg/template/templates/workloads/services/static-site/cf.yml
@@ -3,6 +3,14 @@
 AWSTemplateFormatVersion: 2010-09-09
 Description: CloudFormation template that represents a static site backed by S3.
 
+{{- /*
+!¡!¡!¡!¡!¡!¡
+If you are adding resources to this template, make sure to add the resources to
+the TriggerStateMachineAction.DependsOn list. This will ensure that files
+are only copied to Bucket if all other resources are successfully created.
+!¡!¡!¡!¡!¡!¡
+*/ -}}
+
 {{- if .SerializedManifest }}
 Metadata:
   Manifest: |
@@ -149,7 +157,19 @@ Resources:
     Metadata:
       aws:copilot:description: A custom resource to trigger the file-copying state machine
     Type: Custom::TriggerStateMachine
-    DependsOn: CopyAssetsStateMachine
+    DependsOn: 
+      - CopyAssetsStateMachine {{- /* This is a real dependency */}}
+      - Bucket
+      - BucketPolicyForCloudFront
+      - CloudFrontOriginAccessControl
+      - CloudFrontDistribution
+      - TriggerStateMachineFunction
+      - TriggerStateMachineFunctionRole
+      - CopyAssetsStateMachine
+      - CopyAssetsStateMachineRole
+      - CopyAssetsFunction
+      - CopyAssetsFunctionRole
+      - EnvManagerS3Access
     Properties:
       ServiceToken: !GetAtt TriggerStateMachineFunction.Arn
       StateMachineARN: !GetAtt CopyAssetsStateMachine.Arn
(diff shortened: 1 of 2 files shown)


internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -3,6 +3,14 @@
 AWSTemplateFormatVersion: 2010-09-09
 Description: CloudFormation template that represents a static site backed by S3.
 
+{{- /*
+!¡!¡!¡!¡!¡!¡
```
Lou1415926, line 7: For a second I couldn't recognize the letter `i` 😂

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -3,6 +3,14 @@
...
+If you are adding resources to this template, make sure to add the resources to
```
Lou1415926, line 8: Is `TriggerStateMachineAction`  _always_ expected to be the last resource to get CREATE/UPDATEd? If so, I wonder if it's worth it to write a test that errors out if a resource is in the template but not in `TriggerStateMachineAction.DependsOn`  💭

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -88,7 +96,8 @@ Resources:
     Properties:
       DistributionConfig:
         DefaultCacheBehavior:
-          AllowedMethods: ["GET", "HEAD", "OPTIONS", "PUT", "PATCH", "POST", "DELETE"]
+          Compress: true
+          AllowedMethods: ["GET", "HEAD"]
```
Lou1415926, line 100: Just for learning purpose - what are the considerations behind these changes 👀 ?

internal/pkg/template/templates/workloads/services/static-site/cf.yml
iamhopaul123: Agreed. We could probably do it after the release. I wish there's an integrated method doing that instead of always adding to the list...

iamhopaul123 (approved):

Lou1415926 (approved):
Feel free to remove the label if we decide to add that test but want to do it after the release!

internal/pkg/template/templates/workloads/services/static-site/cf.yml
dannyrandall: since the s3 origin is only for static content, the other methods don't make sense - it doesn't make sense to send a `PUT` (for example) to a url that will only ever accept `GET`s. `HEAD` is used by browsers to check if the file has changed (and therefore, should be `GET`'d) without downloading the file.

internal/pkg/template/templates/workloads/services/static-site/cf.yml
efekarakus: I'm not even sure that it needs to be added to the `DependsOn` thought 💭 

Like if we add a completely random resource that can indeed be created in parallel with the `TriggerStateMachine` custom resource, we wouldn't need to add it as a dependsOn *methinks*

dannyrandall: my main thought here is to prevent this flow:
1. Resources start being created
2. Trigger state machine which copies files into s3
3. Some other resource fails, triggering rollback

At this state, some/all files were "deployed" to s3, and we don't have a mechanism to roll that back. By making the "start the copy" action depend on everything else, at least we will only do the copy if all the other resources have successfully been created/updated. It doesn't solve for failures _within_ the state machine/copy lambda, though.

efekarakus: your reasoning makes sense to me 👍 

Maybe we can strike a good balance: the default is that the test ensures that the resource is added as a dependency, _however_ there should be a way to modify the test so that I can explicitly opt-out a dependsOn.

What do you think?

dannyrandall: yeah, i think that's a good balance! we'll have to find a way to generate all possible versions of the template to validate....or something but i think the test route is a good idea.

PR #4858, opened 2023-05-10
chore: static site copy should use correct content-type

iamhopaul123:
An alternative way solving content-type issue instead of https://github.com/aws/copilot-cli/pull/4857. Resolving the possible edge case when two identical files use different extensions.

### DIFF
--- a/cf-custom-resources/lib/copy-assets.js
+++ b/cf-custom-resources/lib/copy-assets.js
@@ -9,10 +9,19 @@ const aws = require("aws-sdk");
  * Main handler, invoked by Lambda
  */
 exports.handler = async function (event, context, callback) {
-	const s3 = new aws.S3();
-	await s3.copyObject({
-		CopySource: event.srcBucket + "/" + event.mapping.path,
-		Bucket: event.destBucket,
-		Key: event.mapping.destPath
-	}).promise();
+  const s3 = new aws.S3();
+  await s3
+    .copyObject({
+      CopySource: event.srcBucket + "/" + event.mapping.path,
+      Bucket: event.destBucket,
+      Key: event.mapping.destPath,
+      ContentType:
+        event.mapping.contentType !== ""
+          ? event.mapping.contentType
+          : undefined,
+      // Required otherwise ContentType won't be applied.
+      // See https://github.com/aws/aws-sdk-js/issues/1092 for more.
+      MetadataDirective: "REPLACE",
+    })
+    .promise();
 };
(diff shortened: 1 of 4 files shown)


cf-custom-resources/lib/copy-assets.js
```diff
@@ -9,10 +9,19 @@ const aws = require("aws-sdk");
...
-	}).promise();
-};
\ No newline at end of file
+  const s3 = new aws.S3();
+  await s3
+    .copyObject({
+      CopySource: event.srcBucket + "/" + event.mapping.path,
+      Bucket: event.destBucket,
+      Key: event.mapping.destPath,
+      ContentType:
```
dannyrandall, line 18: what happens if this is set to `""`?

iamhopaul123: no error but doesn't make any sense either.

dannyrandall (approved):
nice!!

Lou1415926 (approved):

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -43,6 +44,7 @@ type asset struct {
 
 	ArtifactBucketPath string `json:"path"`
 	ServiceBucketPath  string `json:"destPath"`
+	ContentType        string `json:"contentType"`
```
efekarakus, line 47: nit: do you think it's worthwhile doing `,omitempty` for these tags to just have a file that's slightly smaller if the field is empty?

internal/pkg/deploy/upload/asset/asset.go
```diff
@@ -180,12 +183,12 @@ func (u *ArtifactBucketUploader) uploadAssetMappingFile(assets []asset) (string,
 
 // dedupe returns a copy of assets with duplicate entries removed.
 func dedupe(assets []asset) []asset {
-	type key struct{ a, b string }
+	type key struct{ a, b, c string }
```
efekarakus, line 186: nit:
```go
type key struct{ field1, field2, field3 string }
```

internal/pkg/deploy/upload/asset/asset.go
iamhopaul123: I thought about this but the file itself is relatively small and maybe it's good to keep each element in the array consistent to have the same key set?

internal/pkg/deploy/upload/asset/asset.go
iamhopaul123: Added in https://github.com/aws/copilot-cli/pull/4870

PR #4854, opened 2023-05-09
chore: add custom domain action to static site

iamhopaul123:
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the Apache 2.0 License.

PR #4861, opened 2023-05-10
chore: add cert validator for static site

iamhopaul123:
based on https://github.com/aws/copilot-cli/pull/4854


cf-custom-resources/lib/wkld-cert-validator.js
```diff
@@ -0,0 +1,881 @@
...
+
+  // Initialize global variables.
+  envHostedZoneID = props.EnvHostedZoneId;
+  envName = props.EnvName;
+  appName = props.AppName;
+  serviceName = props.ServiceName;
+  domainName = props.DomainName;
+  rootDNSRole = props.RootDNSRole;
+  cloudFrontCert = props.IsCloudFrontCertificate;
+  certificateDomain = cloudFrontCert
```
Lou1415926: Instead of a `bool`, what do you think of receiving an enum that accepts either `CloudFront` or `NLB`.

cf-custom-resources/lib/wkld-cert-validator.js
```diff
@@ -0,0 +1,881 @@
...
+    ? `${serviceName}.${envName}.${appName}.${domainName}`
+    : `${serviceName}-nlb.${envName}.${appName}.${domainName}`;
```
Lou1415926: hm I don't recall why the NLB one has the `-nlb` at the end. Perhaps it was the default NLB alias, and that the default alias can't collapse with the default alias for HTTPS? Would static site have the same issue?

cf-custom-resources/lib/wkld-cert-validator.js
```diff
@@ -0,0 +1,881 @@
...
+        let oldAliasesSorted = [...oldAliases].sort().join(",");
+        if (oldAliasesSorted === aliasesSorted) {
+          break;
+        }
+      // Fallthrough to "Create". When the aliases are different, the same actions are taken for both "Update" and "Create".
+      case "Create":
+        await validateAliases(aliases, loadBalancerDNS);
+        const certificateARN = await requestCertificate({
+          aliases: aliases,
+          idempotencyToken: CRYPTO.createHash("md5")
```
Lou1415926: what linter are you using 😢  why does it sometimes create a ton of new lines for function paraemters / promise chains and sometimes doesn't :homer-disappearing

cf-custom-resources/lib/wkld-cert-validator.js
```diff
@@ -0,0 +1,881 @@
...
+        if (!targetRecordExists(alias, recordSet)) {
+          return;
+        }
+        if (recordSet[0].Type !== "A") {
+          return;
+        }
+        let aliasTarget = recordSet[0].AliasTarget;
+        if (
+          aliasTarget &&
+          loadBalancerDNS &&
```
Lou1415926: Do we not want to make sure that the alias is pointing to the correct CF DNS as well?

cf-custom-resources/lib/wkld-cert-validator.js
```diff
@@ -0,0 +1,881 @@
...
+          aliasTarget.DNSName.toLowerCase() ===
+            `${loadBalancerDNS.toLowerCase()}.`
```
Lou1415926: nit: not a big deal, but I feel like sometimes it is easier to read when things are in the same line 🤔 
```suggestion
          aliasTarget.DNSName.toLowerCase() === `${loadBalancerDNS.toLowerCase()}.`
```

cf-custom-resources/lib/wkld-cert-validator.js
```diff
@@ -0,0 +1,881 @@
...
+      .describeCertificate({
+        CertificateArn: certificateARN,
+      })
+      .promise();
+    const options = Certificate.DomainValidationOptions || [];
+    options.forEach((option) => {
+      if (
+        option.ResourceRecord &&
+        (aliases.has(option.DomainName) ||
+          option.DomainName.toLowerCase() === certificateDomain.toLowerCase())
```
Lou1415926: same nit about it being easier to read on the same line
```suggestion
        (aliases.has(option.DomainName) ||  option.DomainName.toLowerCase() === certificateDomain.toLowerCase())
```

Lou1415926 (approved):
LGTM!

cf-custom-resources/lib/wkld-cert-validator.js
iamhopaul123: I use https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint 🤔

cf-custom-resources/lib/wkld-cert-validator.js
iamhopaul123: it was the default NLB alias. I don't think it'll be an issue because the alias will be dedicated to this static site service only. And unlike LBWS, if they don't use alias we won't have any default alias for them. Since we always create a CF by default we can just use the CF default domain and it'll be unique instead of creating a default alias for them.

cf-custom-resources/lib/wkld-cert-validator.js
iamhopaul123: Maybe let's keep it a boolean for now because it's easier to use and we can refactor to enum later if we need more types (most likely we'll get all lambda code rewritten in go before we need that). The refactor will be backward compatible. What do you think 🤔

cf-custom-resources/lib/wkld-cert-validator.js
Lou1415926: > And unlike LBWS, if they don't use alias we won't have any default alias for them. 

Should we write the CFN condition differently for static site then? Right now the condition triggers on "HasAssociatedDomain", whereas the intention is to trigger it when "there is an associated domain" + "there is a static site alias"

cf-custom-resources/lib/wkld-cert-validator.js
Lou1415926: Would the refactor trigger an unnecessary update to the existing NLB custom resource / existing static site custom resource though?

cf-custom-resources/lib/wkld-cert-validator.js
iamhopaul123: So static site alias is different because we won't be switching the alias destination between CF/ALB. Also from implementation perspective unless we write a Custom Resource to update the CF, we won't be able to do this as the certificate is required by CF creation (we wouldn't know the CF DNS name in this lambda)

cf-custom-resources/lib/wkld-cert-validator.js
Lou1415926: In addition, does it matter if we use `${serviceName}.${envName}.${appName}.${domainName}` as the main domain for static site cert? Perhaps we should use the specific alias specified under `http.alias` for the static site instead?

Thinking about a user that has a domain associated with their app, and specified `http.alias: example.com` for their static site.

This would validate `${serviceName}.${envName}.${appName}.${domainName}` as well although they never asked for it. Not sure if this is an issue.

cf-custom-resources/lib/wkld-cert-validator.js
Lou1415926: What would happen in this case ⬇️ 
```yaml
# A LBWS
http:
  alias: svc.example.com

# A static site
http:
  alias: svc.example.com # Mistakenly wrote the same alias here
```


Originally, the purpose of this check was to make sure that the alias record that we are trying to update/delete is indeed owned by this particular service. We didn't want to update the alias record if it was pointing to something unexpected. I feel like the same concern applies to static site 🤔 


---


Scratch all of the above!


Would we have an issue when we try to delete the `http.alias` of a static site? There will be an `aliasTarget`, and we will through the error  `Alias ${alias} is already in use by ${aliasTarget.DNSName}. This could be another load balancer of a different service.`?

cf-custom-resources/lib/wkld-cert-validator.js
iamhopaul123: let me try to figure it out how to make it nicer

iamhopaul123: Actually i a prettifier extension overrides this. I increased the line length limit and hopefully i'll look nicer!

cf-custom-resources/lib/wkld-cert-validator.js
iamhopaul123: I think a key difference for static site that makes this not a concern is the fact that this custom resource won't update A record as it does in LBWS. I'll do more manual test to make sure this is not an issue.

cf-custom-resources/lib/wkld-cert-validator.js
iamhopaul123: > Right now the condition triggers on "HasAssociatedDomain",

Right now it is also in the go template's if statement for if alias exists. So I think `whereas the intention is to trigger it when "there is an associated domain" + "there is a static site alias"` this is how it works now

iamhopaul123: > This would validate ${serviceName}.${envName}.${appName}.${domainName} as well although they never asked for it. Not sure if this is an issue.

I feel it is probably fine because it'll make it easier for them to switch to the alias we reserve for them? Also it'll make the static site certificate consistent with the other Copilot managed service certificates.

cf-custom-resources/lib/wkld-cert-validator.js
iamhopaul123: Discussed offline will add more comment justifying why we do this.

cf-custom-resources/lib/wkld-cert-validator.js
iamhopaul123: There'll be a super fast update to update the custom resource action input. but it'll quit instantly because there's no substantial update for DNS.

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -399,4 +409,106 @@ Resources:
                 Resource: "*"
       ManagedPolicyArns:
         - !Sub arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+
+  CertificateValidatorAction:
```
dannyrandall, line 406: can we add these to the trigger state machine action `DependsOn` if alias is enabled?

dannyrandall: i guess i'm not sure how that will work with the `Condition` field here... Would it make sense to not even include these resources in the template if alias is not set?

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -399,4 +409,106 @@ Resources:
...
+      Path: /
+      Policies:
+        - PolicyName: "CertValidatorPolicy"
+          PolicyDocument:
+            Version: '2012-10-17'
+            Statement:
+              - Sid: AllowAssumeRole
+                Effect: Allow
+                Action: sts:AssumeRole
+                Resource: "*"
```
dannyrandall: can we scope this one down?

cf-custom-resources/lib/wkld-cert-validator.js
```diff
@@ -0,0 +1,805 @@
...
+const AWS = require("aws-sdk");
+const CRYPTO = require("crypto");
+const ATTEMPTS_VALIDATION_OPTIONS_READY = 10;
+const ATTEMPTS_RECORD_SETS_CHANGE = 10;
+const DELAY_RECORD_SETS_CHANGE_IN_S = 30;
+const ATTEMPTS_CERTIFICATE_VALIDATED = 19;
+const ATTEMPTS_CERTIFICATE_NOT_IN_USE = 12;
+const DELAY_CERTIFICATE_VALIDATED_IN_S = 30;
+
+let envHostedZoneID, appName, envName, serviceName, certificateDomain, domainTypes, rootDNSRole, domainName, cloudFrontCert;
```
Lou1415926: nit: cuz seems like `cloudFrontCert` is a bool
```suggestion
let envHostedZoneID, appName, envName, serviceName, certificateDomain, domainTypes, rootDNSRole, domainName, isCloudFront;
```

cf-custom-resources/lib/wkld-cert-validator.js
```diff
@@ -0,0 +1,805 @@
...
+      .promise()
+      .then(({ ResourceRecordSets: recordSet }) => {
+        if (!targetRecordExists(alias, recordSet)) {
+          return;
+        }
+        if (recordSet[0].Type !== "A") {
+          return;
+        }
+        let aliasTarget = recordSet[0].AliasTarget;
+        // If loadBalancerDNS is empty, it means we are using this lambda for dedicated CloudFront for Static Site.
```
Lou1415926:
```suggestion
        // If loadBalancerDNS is empty, it means we are using this lambda for dedicated CloudFront for Static Site. CloudFront can't perform the same validation, because passing the CF domain would introduce a circular dependency (the CF can't be created/updated before cert is validated)
```

Lou1415926 (approved):
noice! :shipit:

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -363,7 +365,6 @@ Resources:
     Metadata:
       'aws:copilot:description': "An IAM role {{- if .PermissionsBoundary}} with permissions boundary {{.PermissionsBoundary}} {{- end}} to update the Route 53 hosted zone"
     Type: AWS::IAM::Role
-    Condition: HasAssociatedDomain
```
Lou1415926, line 368: oo why is this not needed anymore?

iamhopaul123: it's essentially "if alias" logic for now.

huanjani (approved):
🎉 🙇🏼‍♀️

internal/pkg/template/templates/workloads/services/static-site/cf.yml
Lou1415926: oh I see: if they don't have `alias` and have a associated domain with the app, they will still access the bucket through the default CF domain?

PR #4863, opened 2023-05-11
chore: remove dependency on workspace for copilot app delete

KollaAdithya:
Following are the changes in this PR
  1. Instead of depending on workspace summary to delete copilot app, we fetch the list of copilot applications from SSM and prompt the user to select the application name. 
  
Screenshot 2023-05-10 at 11 56 47 PM

  2.  When deleting `.workspace` file, if we could not find summary file up to 5 levels from working directory then skip deleting the summary file.
  
3. Delete `.workspace` only if we can find the summary file and also app name provided by user matches with application name in the summary.


internal/pkg/cli/app_delete.go
```diff
@@ -383,7 +409,7 @@ func buildAppDeleteCommand() *cobra.Command {
 		}),
 	}
 
-	cmd.Flags().StringVarP(&vars.name, nameFlag, nameFlagShort, tryReadingAppName(), appFlagDescription)
+	cmd.Flags().StringVarP(&vars.name, nameFlag, nameFlagShort, "", appFlagDescription)
```
Lou1415926: I just looked at the code for `tryReadingAppName` - it returns an empty string when there is no workspace. Therefore, it seems to be fine keeping `tryReadingAppName` here? what do you think

internal/pkg/cli/app_delete.go
```diff
@@ -221,6 +231,15 @@ func (o *deleteAppOpts) Execute() error {
 	return nil
 }
 
+func (o *deleteAppOpts) askAppName() error {
```
Lou1415926: Not an action item for you at all, but I just want to share an alternative way of writing this, which is my personal preference for the time being. 


```suggestion
func (o *deleteAppOpts) Ask() error {
   if err := o.validateOrAskAppName(); err != nil { return err }
   if o.skipConfirmation { return nil }
   // ... the rest
}
func (o *deleteAppOpts) validateOrAskAppName() error {
      if o.name != "" {
           _, err := o.store.GetApplication(o.name)
	   return err
      }
      name, err := o.sel.Application(appDeleteNamePrompt, "")
	if err != nil {
		return fmt.Errorf("select application name: %w", err)
	}
	o.name = name
	return nil
}
```

This way, it is easy to visualize the _high-level steps_ taken in `Ask` - that we first do something with the app name, then we do something asking users to confirm the deletion.

internal/pkg/cli/app_delete.go
```diff
@@ -356,12 +375,19 @@ func (o *deleteAppOpts) deleteAppConfigs() error {
...
+		// When there's a local application summary.
+		summary, err := ws.Summary()
+		if err == nil && summary.Application == o.name {
+			o.spinner.Start(fmt.Sprintf(fmtDeleteAppWsStartMsg, workspace.SummaryFileName))
+			if err := ws.DeleteWorkspaceFile(); err != nil {
+				o.spinner.Stop(log.Serrorf("Error deleting %s file.\n", workspace.SummaryFileName))
+				return fmt.Errorf("delete %s file: %w", workspace.SummaryFileName, err)
+			}
+			o.spinner.Stop(log.Ssuccessf(fmt.Sprintf(fmtDeleteAppWsStopMsg, workspace.SummaryFileName)))
+		}
```
Lou1415926: is it possible to reverse the `if` condition and reduce nestings?

internal/pkg/workspace/workspace_test.go
```diff
@@ -1559,8 +1559,8 @@ func TestWorkspace_DeleteWorkspaceFile(t *testing.T) {
 					Fs: fs,
 				},
 			}
-			ws.fs.MkdirAll("copilot", 0755)
-			ws.fs.Create(tc.copilotDir + "/" + ".workspace")
+			ws.fs.MkdirAll(tc.copilotDir, 0755)
+			ws.fs.Create(filepath.Join(tc.copilotDir, ".workspace"))
```
Lou1415926: do you know if we really need these? They seem to be doing the same as as in `tc.fs()` 🤔

internal/pkg/workspace/workspace.go
```diff
@@ -450,7 +450,7 @@ func (ws *Workspace) WriteEnvironmentManifest(marshaler encoding.BinaryMarshaler
 // DeleteWorkspaceFile removes the .workspace file under copilot/ directory.
 // This will be called during app delete, we do not want to delete any other generated files.
 func (ws *Workspace) DeleteWorkspaceFile() error {
-	return ws.fs.Remove(filepath.Join(CopilotDirName, SummaryFileName))
+	return ws.fs.Remove(filepath.Join(ws.CopilotDirAbs, SummaryFileName))
```
Lou1415926, line 453: omg great find!

Lou1415926 (commented):
super noice!

internal/pkg/cli/app_delete.go
KollaAdithya: done. reversed the condition of if statements

iamhopaul123 (approved):
LGTM!

internal/pkg/cli/app_delete.go
```diff
@@ -356,8 +373,17 @@ func (o *deleteAppOpts) deleteAppConfigs() error {
 }
 
 func (o *deleteAppOpts) deleteWs() error {
+	ws, err := o.existingWorkSpace()
+	if err != nil {
+		return nil
+	}
+	// When there's a local application summary.
+	summary, err := ws.Summary()
+	if err != nil || summary.Application != o.name {
```
Lou1415926, line 382: noicee

Lou1415926 (approved):

Lou1415926 (approved):

PR #4870, opened 2023-05-12
chore: show URI after static site deployment

iamhopaul123:
Also fix the error when using `svc package` with static site and unnecessary prompt for using `--force` when there's no change. 


iamhopaul123:
Based on https://github.com/aws/copilot-cli/pull/4861

internal/pkg/cli/svc_deploy.go
```diff
@@ -489,17 +492,6 @@ func validateWorkloadManifestCompatibilityWithEnv(ws wsEnvironmentsLister, env v
 }
 
 func (o *deploySvcOpts) uriRecommendedActions() ([]string, error) {
-	type reachable interface {
```
Lou1415926, line 492: Why do we remove this 👀 ? Will we insist on grabbing the URI for a worker service for example?

internal/pkg/cli/svc_deploy.go
```diff
@@ -216,6 +217,9 @@ func (o *deploySvcOpts) Execute() error {
 	if err != nil {
 		return err
 	}
+	if o.forceNewUpdate && o.svcType == manifestinfo.StaticSiteType {
```
Lou1415926, line 220: I was searching up where is `o.svcType` figured out, and realized it was in `configureClients`, which is quite unobvious 😓 What do you think of just moving the following lines from `configureClients` to here in `Execute()`:
```go
svc, err := o.store.GetService(o.appName, o.name)
if err != nil {
  return fmt.Errorf("get service %s configuration: %w", o.name, err)
}
o.svcType = svc.Type
```

internal/pkg/cli/svc_deploy_test.go
```diff
@@ -174,6 +177,16 @@ func TestSvcDeployOpts_Execute(t *testing.T) {
 
 			wantedError: fmt.Errorf("get available features of the prod-iad environment stack: some error"),
 		},
+		"error out if force deploy for static site service": {
+			inForceFlag: true,
+			inSvcType:   manifestinfo.StaticSiteType,
```
Lou1415926, line 185: isn't `svcType` figured out by getting the service info from SSM? Do you think we can add a `store` mock here and mock return `GetService`?

internal/pkg/cli/svc_deploy.go
Lou1415926: We should also mention this in our flag description!

internal/pkg/deploy/cloudformation/stack/static_site.go
```diff
@@ -100,9 +100,12 @@ func (s *StaticSite) Template() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	bucket, path, err := s3.ParseURL(s.assetMappingURL)
-	if err != nil {
-		return "", err
+	var bucket, path string
+	if s.assetMappingURL != "" {
```
Lou1415926, line 98: When is this empty? when there is no asset to upload?

internal/pkg/describe/static_site.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"fmt"
+
+	"github.com/dustin/go-humanize/english"
+)
+
+const (
+	staticSiteOutputCFDomainName    = "CloudFrontDistributionDomainName"
+	staticSiteOutputCFAltDomainName = "CloudFrontDistributionAlternativeDomainName"
```
Lou1415926, line 14: Question! Are they supposed to be in the CloudFormation `Outputs` section?

Lou1415926 (commented):
sorry I prematurely submitted my review just now - here is the rest

internal/pkg/cli/svc_deploy.go
iamhopaul123: > We should also mention this in our flag description!

Sounds good! I'll mention it in the description

> What do you think of just moving the following lines from configureClients to here in Execute()

I feel like the same logic could be applied to the other logic in `configureClients` too. The purpose of that function is for facilitating the unit test so that the test can focus on the main logic for execute rather than setting up prerequisites. If that is the case, maybe it's better to keep it as is 🤔 Maybe it would be less confusing if we cache the `svc` as a concrete struct like `env` as opposed to `svcType`?

internal/pkg/cli/svc_deploy.go
iamhopaul123: because `NewReachableService` will handle the logic and exit gracefully if `err != nil`

internal/pkg/cli/svc_deploy_test.go
iamhopaul123: i feel `GetService` is not the main thing we want to test. we could assume the service is properly configured in SSM (whose format is probably what we won't change). And if that is the case, most likely for every test case we'll have `GetService` which I feel cumbersome and distracting

internal/pkg/deploy/cloudformation/stack/static_site.go
iamhopaul123: > when there is no asset to upload?

When we don't perform assets uploading for example `svc package` without `--upload-assets`. Also overall I feel like we should check if it is empty before performing `s3.Parse` without making the assumption that `s.assetMappingURL` is always non-empty

internal/pkg/describe/static_site.go
iamhopaul123: Yeah i think so because they are attributes of the dedicated CF just like we output the attributes of the LB in env stack.

internal/pkg/cli/svc_deploy.go
Lou1415926: My main problem with it is that `svcType` is not a client, so it is a pretty hidden secret that `o.svcType` is figured out in `configureClients`.

It is a piece of information returned by `o.store`, which is a client and can be mocked. I know you replied in the other thread that `GetService` is not a main logic to be tested. However, it _is_ a prerequisite for the rest of deployment, and has to be invoked in for the whole `Execute` to complete. My argument is that, if we remove that `o.store.GetService` line from `svc_deploy.go`, our unit test won't catch that, but our `svc deploy` will certainly break.

So I argue that it _is_ a main piece of logic, however, it could be moved to `Ask`.

internal/pkg/describe/static_site.go
Lou1415926: I guess this is not done yet? Adding `CloudFrontDistributionDomainName` and `CloudFrontDistributionAlternativeDomainName` into the `Outputs` section

iamhopaul123: I think i added them https://github.com/aws/copilot-cli/pull/4870/files#diff-15ae3d1d11e7d786765fd0df1a1cdd256c6083500fc8c1e0f3856ffa7c4716f9R507-R517

Lou1415926: :homer-disappearing

internal/pkg/deploy/cloudformation/stack/static_site.go
Lou1415926: qq - is `bucket` and `path` computable just from the hash of all assets, even if they are not actually uploaded?

internal/pkg/cli/svc_deploy.go
iamhopaul123: Ok moved the logic to `Ask`

internal/pkg/deploy/cloudformation/stack/testdata/workloads/static-site.stack.yml
```diff
@@ -447,4 +447,14 @@ Resources:
...
+
+Outputs:
+  CloudFrontDistributionDomainName:
+    Value: !GetAtt CloudFrontDistribution.DomainName
+    Export:
+      Name: !Sub ${AWS::StackName}-CloudFrontDistributionDomainName
+  CloudFrontDistributionAlternativeDomainName:
+    Value: "example.com"
+    Export:
+      Name: !Sub ${AWS::StackName}-CloudFrontDistributionAlternativeDomainName
```
Lou1415926, line 470: do you know if there is a limit in the `Export.Name` length?

internal/pkg/deploy/cloudformation/stack/static_site.go
iamhopaul123: Yeah that's true but it won't be useful anyway to include in the template I think. The same logic could be applied to the other uploadable like custom resources, addons, and env files.

internal/pkg/deploy/cloudformation/stack/testdata/workloads/static-site.stack.yml
iamhopaul123: 255 characters. We can adjust accordingly to truncate "-CloudFrontDistributionAlternativeDomainName" as needed in the future.

internal/pkg/deploy/cloudformation/stack/static_site.go
Lou1415926: The other uploadable's URLs are precomputed and filled into the template, even without `--upload-assets` though! 

I am okay with not writing the URLs because I don't have a good argument against it right now. I feel like there is a difference between "providing an invalid template" vs. "providing a valid template that may not work" - not writing the precomputed URLs is the former, writing them without necessarily uploading them is the later. Though I don't see how realistically it's a huge difference in terms of usability.

Lou1415926: Oh `svc package --diff` _might_ go wrong in this case, we can test it out.

internal/pkg/cli/svc_deploy.go
Lou1415926:
```go
// NewReachableService returns a ReachableService based on the type of the service.
func NewReachableService(app, svc string, store ConfigStoreSvc) (ReachableService, error) {
	cfg, err := store.GetWorkload(app, svc)
	if err != nil {
		return nil, err
	}
	in := NewServiceConfig{
		App:         app,
		Svc:         svc,
		ConfigStore: store,
	}
	switch cfg.Type {
	case manifestinfo.LoadBalancedWebServiceType:
		return NewLBWebServiceDescriber(in)
	case manifestinfo.RequestDrivenWebServiceType:
		return NewRDWebServiceDescriber(in)
	case manifestinfo.BackendServiceType:
		return NewBackendServiceDescriber(in)
	default:
		return nil, fmt.Errorf("service %s is of type %s which cannot be reached over the network", svc, cfg.Type)
	}
}
```

```go
describer, err := describe.NewReachableService(o.appName, o.name, o.store)
if err != nil {
    return nil, err
} 
```

Looks like we will error out in this case! Should we handle detect ` fmt.Errorf("service %s is of type %s which cannot be reached over the network", svc, cfg.Type)` error here and swallow it?

iamhopaul123: Oh that's right we should swallow the error

internal/pkg/deploy/cloudformation/stack/static_site.go
iamhopaul123: > The other uploadable's URLs are precomputed and filled into the template, even without --upload-assets though!

I don't think I follow. I just tested it out and I got

```yaml
  TriggerStateMachineFunction:
    Metadata:
      aws:copilot:description: A lambda that triggers the file-copying state machine
    Type: AWS::Lambda::Function
    Properties:
      Handler: index.handler
      Role: !GetAtt TriggerStateMachineFunctionRole.Arn
      Runtime: nodejs16.x
      Timeout: 900
      MemorySize: 512
      Code:
        S3Bucket: 
        S3Key: 
```

Lou1415926: https://github.com/aws/copilot-cli/pull/4509/files I'm thinking about this PR

Lou1415926 (approved):

internal/pkg/deploy/cloudformation/stack/static_site.go
iamhopaul123: Gotcha I see. We should probably add that for static site too. Maybe for custom resources it's ok because they should be something copilot handles under the hood where as static assets are more similar to env files/addons template URL/ECR image?

internal/pkg/cli/flag.go
```diff
@@ -259,7 +259,8 @@ Allows you to categorize resources.`
 
 	// Deployment.
 	deployTestFlagDescription = `Deploy your service or job to a "test" environment.`
-	forceFlagDescription      = "Optional. Force a new service deployment using the existing image."
+	forceFlagDescription      = `Optional. Force a new service deployment using the existing image.
+Note that it is not available when service type is "Static Site".`
```
huanjani:
```suggestion
Not available with the "Static Site" service type.`
```

internal/pkg/describe/static_site.go
```diff
@@ -0,0 +1,69 @@
...
+	wkldDescr, err := d.initWkldStackDescriber(envName)
+	if err != nil {
+		return URI{}, err
+	}
+	outputs, err := wkldDescr.Outputs()
+	if err != nil {
+		return URI{}, fmt.Errorf("get stack output for service %q: %w", d.svc, err)
+	}
+	uris := []string{outputs[staticSiteOutputCFDomainName]}
+	if outputs[staticSiteOutputCFAltDomainName] != "" {
```
huanjani, line 62: I like that both are listed!

huanjani (approved):
Looks good! Just one wording nit for the flag.

PR #4871, opened 2023-05-12
chore(e2e): ignore error when creating secrets 

Lou1415926:
1. Changed the name of the secrets, because the old secret was deleted and we are getting 
```
An error occurred (InvalidRequestException) when calling the GetSecretValue operation: You can't perform this operation on the secret because it was marked for deletion.
```
2. Fix the spec `strings.Contains(expectedSecrets[secret.Name], expectedSecrets[secret.Name])` that always return true :X
3. Ignore error when creating/deleting secrets.


e2e/apprunner/apprunner_suite_test.go
```diff
@@ -17,6 +17,8 @@ import (
 var cli *client.CLI
 var appName string
 
+const ssmName = "e2e-apprunner-ssm-param-secret"
```
iamhopaul123, line 20: Do we need to include time info to make them unique 🤔

Lou1415926: I went with the route where we "not error out when the secret already exists" instead of "creating unique secrets". Seems unnecessary to use a new secret every time.

iamhopaul123 (approved):

e2e/apprunner/apprunner_suite_test.go
```diff
@@ -35,20 +37,15 @@ var _ = BeforeSuite(func() {
...
-	Expect(err).NotTo(HaveOccurred())
-	err = command.Run("aws", []string{"secretsmanager", "create-secret", "--name", "e2e-apprunner-MyTestSecret", "--secret-string", "\"{\"user\":\"diegor\",\"password\":\"EXAMPLE-PASSWORD\"}\"", "--tags", "[{\"Key\":\"copilot-application\",\"Value\":\"" + appName + "\"},{\"Key\":\"copilot-environment\", \"Value\":\"" + envName + "\"}]"})
-	Expect(err).NotTo(HaveOccurred())
+	_ = command.Run("aws", []string{"ssm", "put-parameter", "--name", ssmName, "--value", "abcd1234", "--type", "String", "--tags", "[{\"Key\":\"copilot-application\",\"Value\":\"" + appName + "\"},{\"Key\":\"copilot-environment\", \"Value\":\"" + envName + "\"}]"})
+	_ = command.Run("aws", []string{"secretsmanager", "create-secret", "--name", secretName, "--secret-string", "\"{\"user\":\"diegor\",\"password\":\"EXAMPLE-PASSWORD\"}\"", "--tags", "[{\"Key\":\"copilot-application\",\"Value\":\"" + appName + "\"},{\"Key\":\"copilot-environment\", \"Value\":\"" + envName + "\"}]"})
```
huanjani, line 42: Oh, whoops-- I didn't realize the secret discussion we had in our team mtg was about this particular e2e test 🤦🏼. In #4877 I added the `--force-overwrite-replica-secret` flag to the `create-secret` call.

huanjani: But your solution catches more cases.

Lou1415926: ooh initially last week it errored out because we were trying to create duplicated secrets, and then we deleted the secret manually, and it takes 7 days to delete, during which time we can't create a secret with the same name nor use the secret that's pending deletion. So at this point we are doomed and we have to change the name of the secret :rip

I don't know why it errored out initially even after the fix! Perhaps the failure happened before May 15 🤔 ?

Lou1415926:
Closing in favor of #4877

PR #4873, opened 2023-05-15
chore(templates): remove unnecessary !Sub in Aurora Serverless v2 template

andreas-bergstrom:
!Sub not needed here as the string does not use CF-variables?


huanjani (approved):
Thanks for catching this and for making the change!

Lou1415926 (approved):
yay thank you!!

PR #4877, opened 2023-05-16
chore(e2e): add overwrite flag for App Runner ssm parameter and secret

huanjani:
Currently, the parameter name created by the App Runner e2e test isn't unique. If something unexpected happens and the parameter isn't deleted, we'll get `"An error occurred (ParameterAlreadyExists) when calling the PutParameter operation: The parameter already exists."`. Rather than adding a time stamp to make the parameter name unique, I opted to leverage the `--overwrite` flag so we don't end up with orphaned resources. 

Ref: https://docs.aws.amazon.com/cli/latest/reference/ssm/put-parameter.html


iamhopaul123 (approved):

Lou1415926:
It seems like we should merge either https://github.com/aws/copilot-cli/pull/4871 or this one! I am fine with closing my own, but we should probably change the secret name from `e2e-apprunner-ssm-param` to something else, because of the reason stated in https://github.com/aws/copilot-cli/pull/4871#discussion_r1195674854.

Also, I think we need to fix a statement in `e2e/apprunner/apprunner_test.go` that _always_ evaluates to `true` 😂  see my change here! https://github.com/aws/copilot-cli/blob/4aafbabce0047011189aa00419b83e719bedb95a/e2e/apprunner/apprunner_test.go

Lou1415926 (approved):

PR #4885, opened 2023-05-17
chore(e2e): separate overwrite flag from tags

huanjani:
For `put-parameter`, the overwrite flag cannot be used with tags. I've deleted the dangling duplicate parameter from Systems Manager. Hopefully the reaper can solve this problem going forward.


KollaAdithya (approved):
should we reopen this #4871 🤔

Lou1415926:
what about we keep the `--overwrite` and use https://docs.aws.amazon.com/systems-manager/latest/userguide/tagging-parameters.html#tagging-parameters-update-command-line to add the tag?

Lou1415926 (approved):

PR #4888, opened 2023-05-17
chore: replace copy assets lambda with state machine

dannyrandall:
This PR primarily is replacing the copy assets lambda with equivalent functionality built in to the copy assets state machine.

Also, some nit feedback from previous PRs has been addressed by adding/fixing comments.


internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -179,25 +179,30 @@ Resources:
...
+    DependsOn:
       - Bucket
       - BucketPolicyForCloudFront
       - CloudFrontOriginAccessControl
       - CloudFrontDistribution
       - TriggerStateMachineFunction
       - TriggerStateMachineFunctionRole
-      - CopyAssetsStateMachine
+      - CopyAssetsStateMachine {{- /* This is a real dependency */}}
       - CopyAssetsStateMachineRole
```
iamhopaul123, line 191: Do we still need this?

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -222,19 +227,42 @@ Resources:
...
-              mapping.$: $$.Map.Item.Value
-              destBucket.$: $.DestinationBucket
-              srcBucket.$: $.SourceBucket
             ItemProcessor:
               ProcessorConfig:
                 Mode: INLINE
-              StartAt: CopyFile
+              StartAt: ContentTypeChoice
               States:
+                ContentTypeChoice:
```
iamhopaul123, line 236: Why not copy with whatever contenttype we have in the mapping file 🤔

internal/pkg/template/templates/workloads/services/static-site/cf.yml
dannyrandall: yeah, still only want to trigger the state machine _after_ everything else finishes

internal/pkg/template/templates/workloads/services/static-site/cf.yml
dannyrandall: if it's not present using `$.contentType` throws an error🥺 and if it's `contentType: ""`, then it uploads with an empty string `Content-Type` instead of the default value (`binary/octet-stream`)

internal/pkg/template/templates/workloads/services/static-site/cf.yml
iamhopaul123: no i meant do we need this resource?

internal/pkg/template/templates/workloads/services/static-site/cf.yml
iamhopaul123: Interesting seems like we don't have a good way of specifying "REPLACE" only if `contentType != ""`

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -180,25 +180,30 @@ Resources:
...
       - TriggerStateMachineFunction
       - TriggerStateMachineFunctionRole
-      - CopyAssetsStateMachine
+      - CopyAssetsStateMachine {{- /* This is a real dependency */}}
       - CopyAssetsStateMachineRole
-      - CopyAssetsFunction
-      - CopyAssetsFunctionRole
       - EnvManagerS3Access
+      {{- if .StaticSiteAlias}}
+      - CustomDomainAction
```
Lou1415926, line 194: we should prolly add a test maybe after the release, that look through all resources in the template, and check if every one of them is in the `DependsOn` list.

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -223,19 +228,42 @@ Resources:
...
                 CopyFile:
                   Type: Task
                   End: true
-                  Resource: !GetAtt CopyAssetsFunction.Arn
+                  Resource: arn:aws:states:::aws-sdk:s3:copyObject
+                  Parameters:
+                    CopySource.$: States.Format('{{.AssetMappingFileBucket}}/{}', $.path)
+                    Bucket: !Ref Bucket
+                    Key.$: $.destPath
+                    MetadataDirective: "REPLACE"
```
iamhopaul123, line 254: wait if without content type why do we need this?

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -223,19 +228,42 @@ Resources:
...
+                        IsPresent: false
+                      - Variable: $.contentType
+                        StringMatches: ""
+                      Next: CopyFile
+                  Default: CopyFileWithContentType
                 CopyFile:
                   Type: Task
                   End: true
-                  Resource: !GetAtt CopyAssetsFunction.Arn
+                  Resource: arn:aws:states:::aws-sdk:s3:copyObject
```
Lou1415926, line 249: ohh interesting 👁️

internal/pkg/template/templates/workloads/services/static-site/cf.yml
dannyrandall: oh `CopyAssetsStateMachineRole`? yeah, it's the role used by the state machine: https://github.com/aws/copilot-cli/pull/4888/files#diff-15ae3d1d11e7d786765fd0df1a1cdd256c6083500fc8c1e0f3856ffa7c4716f9R211

dannyrandall: i did remove these two, if that's what you're thinking https://github.com/aws/copilot-cli/pull/4888/files#diff-15ae3d1d11e7d786765fd0df1a1cdd256c6083500fc8c1e0f3856ffa7c4716f9L193-L194. They were similarly named :)

internal/pkg/template/templates/workloads/services/static-site/cf.yml
dannyrandall: theoretically, if `contentType` was empty, then we should reset the `Content-Type` of the file to the default. example:
1. upload `index.html`, content type in s3 is `text/html`
2. rename `index.html` -> `index.xyz` on local machine without changing the file's content
3. upload `index.xyz`, detected content type is `""`, s3 content type should change to `binary/octet-stream` instead of staying `text/html`

Lou1415926 (approved):
other logics LGTM, I will let you & boss discuss the REPLACE bit

internal/pkg/template/templates/workloads/services/static-site/cf.yml
iamhopaul123: oh that's what i thought 🤦

internal/pkg/template/templates/workloads/services/static-site/cf.yml
dannyrandall: we could add a new state to do that - my thinking was that we should treat `contentType: ""` and missing content type the same. what do you think?

iamhopaul123 (approved):
LGTM!

PR #4896, opened 2023-05-18
chore(template): addr small cf config nits from bugbash

iamhopaul123:

### DIFF
--- a/internal/pkg/template/templates/workloads/services/static-site/cf.yml
+++ b/internal/pkg/template/templates/workloads/services/static-site/cf.yml
@@ -96,7 +96,9 @@ Resources:
     Properties:
       OriginAccessControlConfig:
           Description: !Sub 'Access control for static s3 origin for ${AppName}-${EnvName}-${WorkloadName}'
-          Name: copilot-origin-access-control
+          # Truncate the name to allow at most 64 characters.
+          {{$stackName := printf "%s-%s-%s" .AppName .EnvName .WorkloadName}}
+          Name: {{printf "%.64s" $stackName}}
           OriginAccessControlOriginType: s3
           SigningBehavior: always
           SigningProtocol: sigv4
--- a/internal/pkg/template/templates/workloads/services/static-site/cf.yml
+++ b/internal/pkg/template/templates/workloads/services/static-site/cf.yml
@@ -97,8 +97,7 @@ Resources:
       OriginAccessControlConfig:
           Description: !Sub 'Access control for static s3 origin for ${AppName}-${EnvName}-${WorkloadName}'
           # Truncate the name to allow at most 64 characters.
-          {{$stackName := printf "%s-%s-%s" .AppName .EnvName .WorkloadName}}
-          Name: {{printf "%.64s" $stackName}}
+          Name: {{printf "%.64s" (printf "%s-%s-%s" .AppName .EnvName .WorkloadName)}}
           OriginAccessControlOriginType: s3
           SigningBehavior: always
           SigningProtocol: sigv4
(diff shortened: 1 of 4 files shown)


internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -95,8 +95,8 @@ Resources:
     Type: AWS::CloudFront::OriginAccessControl
     Properties:
       OriginAccessControlConfig:
-          Description: Access control for static s3 origin
-          Name: !Sub 'copilot-${AppName}-${EnvName}-${WorkloadName}-origin-access-control'
+          Description: !Sub 'Access control for static s3 origin for ${AppName}-${EnvName}-${WorkloadName}'
+          Name: copilot-origin-access-control
```
Lou1415926: Is the origin's name supposed to be unique in an account?

iamhopaul123: name is like a "tag" thing. the real id is the resource ID

iamhopaul123: And i moved the previous content for "Name" to "Description"

Lou1415926: Screenshot 2023-05-18 at 3 52 31 PM

I got this from the console though!

iamhopaul123: Oh that's right. Ok i think the only thing we could do here is to truncate the previous name

iamhopaul123: OK I've tested out and it seems like they allow names like `adf-asdf-` which has a "-" at the end.

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -95,8 +95,10 @@ Resources:
     Type: AWS::CloudFront::OriginAccessControl
     Properties:
       OriginAccessControlConfig:
-          Description: Access control for static s3 origin
-          Name: !Sub 'copilot-${AppName}-${EnvName}-${WorkloadName}-origin-access-control'
+          Description: !Sub 'Access control for static s3 origin for ${AppName}-${EnvName}-${WorkloadName}'
+          # Truncate the name to allow at most 64 characters.
+          {{$stackName := printf "%s-%s-%s" .AppName .EnvName .WorkloadName}}
+          Name: {{printf "%.64s" $stackName}}
```
Lou1415926: wait - i'm worried if this could happen
```
apppppppppp-envvvvvvvvvvvvvvvvvvvv-service1
apppppppppp-envvvvvvvvvvvvvvvvvvvv-service1
```
both would be truncated to `apppppppppp-envvvvvvvvvvv` 😓

iamhopaul123: Well it is to help to _some extent_ but not all cases. I guess `apppppppppp-envvvvvvvvvvv` is relatively uncommon. We can deal with this problem completely by hashing but i don't want to sacrifice the readability either. In the future we could change if many ppl complain because they have really long app and env name.

Lou1415926: I think once truncate happens, it's likely end with only "app-env" left, because the "-service" part is always the last to truncate :( 

I think somewhere in the codebase that we equally distribute the quota to truncate among app, env and service, so it's more likely that the truncated name is unique.

Something like `subStringlength = (64-3) / 3`, and then return `app[len(app)-subStringlength:]-env[len(app)-subStringlength:]-svc[len(app)-subStringlength:]`

However I agree that people have workaround - worse case they could use override to hardcode a name. But this YAML patch file can never go away even after we update the code.

You make the call! I will approve.

iamhopaul123: Or we truncate by 58 characters and append with 6 char hashing

Lou1415926 (approved):

dannyrandall (approved):
awesome!!!

PR #4900, opened 2023-05-19
chore: compute source full path from workspace dir

Lou1415926:
The author is @huanjani! I've tried to create a PR under her name but I got 
>  "fork_collab Fork collab can't be granted by someone without permission"

But the author is @huanjani!


internal/pkg/cli/deploy/static_site.go
```diff
@@ -166,14 +179,26 @@ func (d *staticSiteDeployer) stackConfiguration(in *StackRuntimeConfiguration) (
...
+		fullPath := filepath.Join(root, source.Source)
+		fileInfo, err := d.fs.Stat(fullPath)
 		if err != nil {
-			return fmt.Errorf("source %q must be a valid path: %w", upload.Source, err)
+			return nil, fmt.Errorf("source %q must be a valid path relative to the workspace root %q: %w", source.Source, root, err)
 		}
+		convertedFileUploads[i] = manifest.FileUpload{
+			Source:      fullPath,
+			Destination: source.Destination,
+			Recursive:   fileInfo.IsDir(),
```
dannyrandall: shouldn't this just match the manifest configuration? so:
```suggestion
			Recursive:   source.Recursive,
```

internal/pkg/cli/deploy/static_site.go
```diff
@@ -166,14 +179,26 @@ func (d *staticSiteDeployer) stackConfiguration(in *StackRuntimeConfiguration) (
 	return conf, nil
 }
 
-func (d *staticSiteDeployer) validateSources() error {
-	for _, upload := range d.staticSiteMft.FileUploads {
-		_, err := d.fs.Stat(upload.Source)
+// convertSources transforms the source's path relative to the project root into the full path.
+func (d *staticSiteDeployer) convertSources() ([]manifest.FileUpload, error) {
+	var convertedFileUploads = make([]manifest.FileUpload, len(d.staticSiteMft.FileUploads))
```
dannyrandall:
```suggestion
	convertedFileUploads := make([]manifest.FileUpload, len(d.staticSiteMft.FileUploads))
```

internal/pkg/cli/svc_init.go
```diff
@@ -805,10 +807,24 @@ func validateWorkspaceApp(wsApp, inputApp string, store store) error {
 
 func (o initSvcOpts) convertStringsToAssets(sources []string) ([]manifest.FileUpload, error) {
 	assets := make([]manifest.FileUpload, len(sources))
+	if o.wsPendingCreation {
```
dannyrandall: since we don't _know_ that our current location is the eventual project root, should we just skip validation in this case? since it'll get validated on `svc deploy` anyways

Lou1415926: Good question. I re-thought the process, and realized a few things:
1. When there is no workspace, like you said, we should leave the validation to `svc deploy`. Realistically, `Validate` and `Ask` without a workspace is a scenario that only happens during `copilot init`. 
2. When there is a workspace, we should validate the path _during_ the prompts, not after, when we convert the path to struct. If we validate during the prompts, folks can fix their path immediately without having to restart the whole command again.

The code has been updated accordingly!

internal/pkg/cli/deploy/static_site.go
```diff
@@ -168,14 +181,30 @@ func (d *staticSiteDeployer) stackConfiguration(in *StackRuntimeConfiguration) (
...
 		if err != nil {
-			return fmt.Errorf("source %q must be a valid path: %w", upload.Source, err)
+			return fmt.Errorf("source %q must be a valid path relative to the workspace root %q: %w", upload.Source, d.wsRoot, err)
 		}
 	}
 	return nil
 }
 
+// convertSources transforms the source's path relative to the project root into the full path.
+// and validate that they exist.
```
dannyrandall:
```suggestion
// convertSources transforms the source's path relative to the project root into the absolute path.
```

internal/pkg/cli/svc_init.go
```diff
@@ -155,13 +156,15 @@ type initSvcOpts struct {
 	staticAssets []manifest.FileUpload
 
 	// For workspace validation.
+	wsRootGetter      wsRootGetter
```
dannyrandall: maybe i'm not seeing it - but since the value is cached, it seems like we don't need this on `initSvcOpts`?

internal/pkg/cli/svc_init.go
```diff
@@ -286,6 +292,21 @@ func (o *initSvcOpts) Validate() error {
 	return nil
 }
 
+func (o *initSvcOpts) validateSourcePaths(sources []string) error {
+	if o.wsPendingCreation {
+		// This can happen during `copilot init`, that we have to `Validate` before there is a workspace.
+		// In this case, we skip the validation for path, and let `svc deploy` handles the validation.
```
```suggestion
		// In this case, we skip the validation for path, and let `svc deploy` handle the validation.
```

internal/pkg/cli/validate.go
```diff
@@ -332,6 +332,17 @@ func validatePath(fs afero.Fs, val interface{}) error {
 	return nil
 }
 
+func validateNonEmpty(val interface{}) error {
```
dannyrandall: nit
```suggestion
func validateNonEmptyString(val interface{}) error {
```

dannyrandall (approved):
looks great! just a couple of small things 😊

huanjani (approved):
😻!!!

PR #4903, opened 2023-05-19
chore: fix `app upgrade` issues

dannyrandall:
This fixes a couple of issues with `copilot app upgrade`.

1. The info spinner printed before the stack progress tracker had an issue because the spinner doesn't support not being the last line on the terminal. This was causing that line to constantly jump between its original location and a copy of it showing up at the bottom of the terminal. To fix this, the spinner was replaced with plain log statements.
2. Running `copilot app upgrade` in a random directory would print an error saying the stack "-infrastructure-roles" didn't exist. This is because 1) `Ask()` and `Execute()` were not being run, and 2) the version getter was created using the app name from the flag or workspace, but not the name captured within `Ask()`.
3. If a **stack set** upgrade resulted in no changes in any of the **stack set instances**, the stack renderer used for the stack set instances would _never_ complete because the logic waits until there is a completion event timestamped _after_ the stack set upgrade started. However, if there are no changes, no change set is executed on the stack set instances, so there will never be any events. This was resolved by adding the stack set operation ID as an output to the stack set template. This way, there's always a change set on upgrade, even if there's not actually anything to do.


internal/pkg/cli/app_upgrade.go
```diff
@@ -60,19 +61,21 @@ func newAppUpgradeOpts(vars appUpgradeVars) (*appUpgradeOpts, error) {
...
 		appUpgradeVars: vars,
 		store:          store,
 		identity:       identity.New(sess),
 		prog:           termprogress.NewSpinner(log.DiagnosticWriter),
 		route53:        route53.New(sess),
 		sel:            selector.NewAppEnvSelector(prompt.New(), store),
-		versionGetter:  d,
 		upgrader:       cloudformation.New(sess, cloudformation.WithProgressTracker(os.Stderr)),
+		newVersionGetter: func(s string) (versionGetter, error) {
+			d, err := describe.NewAppDescriber(s)
```
Lou1415926: nitt
```suggestion
		newVersionGetter: func(appName string) (versionGetter, error) {
			d, err := describe.NewAppDescriber(appName)
```

internal/pkg/cli/app_upgrade.go
```diff
@@ -109,13 +117,13 @@ func (o *appUpgradeOpts) Execute() error {
 	if err != nil {
 		return fmt.Errorf("get application %s: %w", o.name, err)
 	}
-	o.prog.Start(fmt.Sprintf(fmtAppUpgradeStart, color.HighlightUserInput(o.name), color.Emphasize(version), color.Emphasize(deploy.LatestAppTemplateVersion)))
+	log.Infof(fmtAppUpgradeStart, color.HighlightUserInput(o.name), color.Emphasize(deploy.LatestAppTemplateVersion), color.Emphasize(deploy.LatestAppTemplateVersion))
```
Lou1415926: Should it be:
```suggestion
	log.Infof(fmtAppUpgradeStart, color.HighlightUserInput(o.name), color.Emphasize(version), color.Emphasize(deploy.LatestAppTemplateVersion))
```

internal/pkg/cli/app_upgrade.go
```diff
@@ -205,7 +210,7 @@ func buildAppUpgradeCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			return opts.Execute()
```
iamhopaul123, line 208: 🥹

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -71,12 +70,12 @@ func (cf CloudFormation) UpgradeApplication(in *deploy.CreateAppInput) error {
...
-		return err
-	}
-	if err := cf.upgradeAppStack(s); err != nil {
-		return err
+	if err := cf.upgradeAppStack(appConfig); err != nil {
+		var empty *cloudformation.ErrChangeSetEmpty
+		if !errors.As(err, &empty) {
+			return err
+		}
+		log.Infof("No changes for %s.\n", appConfig.StackName())
```
iamhopaul123: We don't usually log outside of the `cli` pkg. Can we make it an error and catch it in the `cli` pkg to log it?

iamhopaul123 (commented):
LGTM overall! Just one nit

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -104,36 +103,40 @@ func (cf CloudFormation) upgradeAppStackSet(config *stack.AppStackConfig) error
 	}
 }
 
-func (cf CloudFormation) upgradeAppStack(s *cloudformation.Stack) error {
-	for {
-		// Upgrade app stack.
+func newRenderAppStackInput(cfnStack *cloudformation.Stack) *executeAndRenderChangeSetInput {
```
Lou1415926: why do we prefer this over just embedding the code in `upgradeAppStack` 👀 ?

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -104,36 +103,40 @@ func (cf CloudFormation) upgradeAppStackSet(config *stack.AppStackConfig) error
...
+		label := fmt.Sprintf("Proposing infrastructure changes for %s.", s.Name)
+		spinner.Start(label)
+		defer stopSpinner(spinner, err, label)
+
+		// get tags from existing stack
 		descr, err := cf.cfnClient.Describe(s.Name)
 		if err != nil {
-			return fmt.Errorf("describe stack %s: %w", s.Name, err)
-		}
-		if cloudformation.StackStatus(aws.StringValue(descr.StackStatus)).InProgress() {
```
Lou1415926, line 114: I wonder why we needed the retry - is this because we did automatic `app upgrade` from `svc deploy` before @iamhopaul123 do you remember?

I just wanted to make sure that we are removing codes that are actually removable 👀

Lou1415926 (approved):
LGTM! Applying a label for my question!

Lou1415926 (approved):
LGTM! Applying a label for my question!

internal/pkg/deploy/cloudformation/app.go
iamhopaul123: Yeah i guess that's the main reason. Previously `app upgrade` was called much more frequently then it is now, and we expected there would be very likely two upgrade happening at the same time will cause this issue without retrying. I seems ok to me to remove it now.

internal/pkg/deploy/cloudformation/app.go
dannyrandall: haha this was more of a "copy the way it was done somewhere else" than me actually thinking about it. i removed it!!

internal/pkg/deploy/cloudformation/app.go
dannyrandall: hmmm...i don't think there's a good way for us to break that up without inserting some knowledge about the two stacks for the app into `cli` - essentially moving the two steps (`UpgradeAppStack` and `UpgradeAppStackSet`) over to `cli`. we could:
a) remove the log
b) break `UpgradeApplication` into `UpgradeAppStack()` and `UpgradeAppStackSet()`, and call those directly from `cli`
c) leave as is.

i'm happy with any of these options - thoughts?

efekarakus: Why not a), does it look awkward?

dannyrandall: discussed offline - removing the log message!

internal/pkg/template/templates/app/cf.yml
```diff
@@ -148,3 +148,6 @@ Outputs:
   TemplateVersion:
     Description: Required output to force the stackset to update if mutating version.
     Value: {{.TemplateVersion}}
+  StackSetOpId:
+    Description: Required output to forces the stackset to update on every operation
```
efekarakus: This explains why we need the output (to force stackset updates), but why do we need to force the stackset update in the first place? Should we add that as a comment?

efekarakus (commented):
Looks great!

iamhopaul123 (approved):

internal/pkg/template/templates/app/cf.yml
dannyrandall: good idea - reasoning is in the PR description (3) but i'll document it in the template as well.

PR #4906, opened 2023-05-19
chore: do not create ecr for static site

iamhopaul123:
Fixing this by adding one boolean parameter to `AddWkldToApp` to indicate if we need to create ECR or not. Sounds like a  non-scalable solution but I don't see a good reason why we want to continue adding more resources per service in stackset instance, as it is not compatible if we want to support `svc deploy` from manifest in the future. Also this way (how we manage ECR) strengthen the coupling between app and services.


internal/pkg/deploy/cloudformation/app.go
```diff
@@ -231,8 +231,8 @@ func (cf CloudFormation) getResourcesForStackInstances(app *config.Application,
 // AddServiceToApp attempts to add new service specific resources to the application resource stack.
 // Currently, this means that we'll set up an ECR repo with a policy for all envs to be able
 // to pull from it.
-func (cf CloudFormation) AddServiceToApp(app *config.Application, svcName string) error {
-	if err := cf.addWorkloadToApp(app, svcName); err != nil {
+func (cf CloudFormation) AddServiceToApp(app *config.Application, svcName string, createECR bool) error {
+	if err := cf.addWorkloadToApp(app, svcName, createECR); err != nil {
```
Lou1415926: What about instead of `createECR` we do this:
```go
type AddServiceToAppOpt func(*stack.AppResourcesService)

func WithoutECR(config *stack.AppResourcesService) { config.WithECR = false }

func (cf CloudFormation) AddServiceToApp(app *config.Application, svcName string, opts ...AddServiceToAppOpt) error
```

I know there are no other `opts` so far, but I feel like it isn't too early to establish a more scalable pattern now either. In the future, if we need to add more `opts` func to `addServiceToApp`, it'd be easy for the (potentially new) developers to just copy the pattern instead of keep adding boolean.

internal/pkg/initialize/workload_test.go
```diff
@@ -488,7 +488,40 @@ func TestWorkloadInitializer_Service(t *testing.T) {
 				m.EXPECT().AddServiceToApp(&config.Application{
 					Name:      "app",
 					AccountID: "1234",
-				}, "frontend")
+				}, "frontend", true)
+			},
+		},
+		"writes Static Site manifest, and creates repositories successfully": {
```
Lou1415926: if we are not creating repo for static site - 
```suggestion
		"writes Static Site manifest": {
```

Lou1415926 (approved):
I will approve and leave the decision (on the `opts`  comment) to you!

internal/pkg/deploy/cloudformation/app.go
iamhopaul123: Addressed!

bvtujo (approved):
LGTM! Nice and clean.

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -220,11 +220,17 @@ func (cf CloudFormation) getResourcesForStackInstances(app *config.Application,
 	return regionalResources, nil
 }
 
+// AddWorkloadToAppOpt allows passing optional parameters to AddServiceToApp.
+type AddWorkloadToAppOpt func(*stack.AppResourcesService)
+
+// AddWorkloadToAppOptWithoutECR adds a workload to app without creating an ECR repo.
+func AddWorkloadToAppOptWithoutECR(config *stack.AppResourcesService) { config.WithECR = false }
```
dannyrandall, line 227: nit: most functional options i've seen return an option function:
```suggestion
func AddWorkloadToAppWithoutECR(config *stack.AppResourcesService) AddWorkloadToAppOpt {
  return func(config *stack.AppResourcesService) {
    config.WithECR = false
  } 
}
```
i think it's mostly so that a user of the option could provide parameters if they were necessary. in this case, they're not, so it doesn't really matter either way!

internal/pkg/deploy/cloudformation/stack/app.go
```diff
@@ -27,10 +27,23 @@ type DeployedAppMetadata struct {
...
+type AppResourcesService struct {
+	Name    string `yaml:"Name,flow"`
+	WithECR bool   `yaml:"WithECR,flow"`
+}
+
+// UnmarshalYAML overrides the default YAML unmarshaling logic for the Image
+// struct, allowing it to perform more complex unmarshaling behavior.
+// This method implements the yaml.Unmarshaler (v3) interface.
+func (s *AppResourcesService) UnmarshalYAML(value *yaml.Node) error {
+	return value.Decode(&s.Name)
```
dannyrandall, line 46: i'm confused how this will work when a _second_ service is created. i see how the flow works if you disable ecr for the first service, but what about when a new app template is generated for a _future_ `svc init`? won't the `WithECR` field always be false? so all of the ECR repos will be deleted?

PR #4914, opened 2023-05-22
fix: don't prepend a leading / when it's not intended in healthchecks and path patterns.

bvtujo:
This PR attempts to standardize how our manifests accept path patterns for primary paths, additional routing rules, and HTTP health checks. 

Previously, specifying the following manifest:
```yaml
http:
  path: /app
  healthcheck: /app
```
would result in requests being routed to `//app` and healthchecks against `/app`, meaning that the service would come up healthy but no traffic would reach it if the expected path was `loadbalancer.com/app`. 

The opposite manifest, 
```yaml
http:
  path: app
  healthcheck: app
```
would fail to deploy, as a healthcheck path of `app` is invalid. 

This PR adds a new transformer, convertPath, which attempts to fix this behavior. 

If a customer provides `app` or `/app` in the `path` or `healthcheck` field, the path will be rendered as `/app`. There is no longer a hardcoded leading `/` in the manifest configuration for routing rules. 

Customers who depend on the paths being rendered as `loadbalancer.com//app` can still access this behavior by specifying that path explicitly in the manifest:
```yaml
http:
  path: //app
```

I also investigated the behavior of the rule priority function and don't believe that any of these changes impact it. 

I didn't change any of the `RulePath` parameters in our integ tests as this parameter is no longer used. I'd love to remove it in a future release but CFN deployments will fail for stacks which use it, and it's some trouble to detect whether the parameter exists and optionally include it in the template param configuraiton. 


Lou1415926 (approved):

internal/pkg/deploy/cloudformation/stack/transformers.go
```diff
@@ -573,6 +573,18 @@ type routingRuleConfigConverter struct {
 	redirectToHTTPS bool
 }
 
+// convertPath attempts to standardize manifest paths on '/path' or '/' patterns.
+//   - If the path starts with a / (including "/"), return it unmodified.
+//   - Otherwise, prepend a leading '/' character.
+//
+// CFN health check and path patterns expect a leading '/', so we do that here instead of in the template.
+func convertPath(path string) string {
+	if path[0] == '/' {
```
iamhopaul123, line 587: Maybe we should deal with the empty input case in this function?

iamhopaul123 (approved):
LGTM!

PR #4922, opened 2023-05-23
chore(diff): fix bug in GetAtt overrider

Lou1415926:

### DIFF
--- a/internal/pkg/template/diff/override.go
+++ b/internal/pkg/template/diff/override.go
@@ -196,21 +196,20 @@ func (converter *getAttConverter) parse(from, to *yaml.Node, key string, overrid
 		toValue = to.Content[1]
 	}
 	// If the input node are of the same type (i.e. both seq or both scalar), parse them normally.
-	if fromValue.Kind == toValue.Kind {
-		return converter.intrinsicFuncMapTagConverter.parse(from, to, key, overrider)
-	}
 	// Otherwise, first convert the scalar input to seq input, then parse.
-	var err error
-	switch {
-	case fromValue.Kind == yaml.ScalarNode:
-		fromValue, err = getAttScalarToSeq(fromValue)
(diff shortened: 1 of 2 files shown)


internal/pkg/template/diff/override.go
```diff
@@ -196,21 +196,20 @@ func (converter *getAttConverter) parse(from, to *yaml.Node, key string, overrid
 		toValue = to.Content[1]
 	}
 	// If the input node are of the same type (i.e. both seq or both scalar), parse them normally.
-	if fromValue.Kind == toValue.Kind {
-		return converter.intrinsicFuncMapTagConverter.parse(from, to, key, overrider)
-	}
 	// Otherwise, first convert the scalar input to seq input, then parse.
```
iamhopaul123, line 199: nit: should we update the comment's "if...otherwise" accordingly?

internal/pkg/template/diff/override_test.go
```diff
@@ -234,6 +234,12 @@ ImageId: !FindInMap
 				}
 			},
 		},
+		"diff in Fn::GetAtt vs !GetAtt when comparing scalar to scalar both with tags": {
```
iamhopaul123: aren't old and curr the same thing 🤔 why is it useful?

Lou1415926: I should change the test case name to "no diff...." because I want to make sure no diff is being produced.

internal/pkg/template/diff/override.go
Lou1415926: The logic is still the same, I just switched the `if` clause. 
```go
// If the input node are of the same type (i.e. both seq or both scalar), parse them normally.
// Otherwise, first convert the scalar input to seq input, then parse.
```
is still an accurate description of the code here.

iamhopaul123 (approved):

dannyrandall (approved):

PR #4923, opened 2023-05-23
chore: fix broken app stack set for ecr management

iamhopaul123:
Fix broken experience introduced by https://github.com/aws/copilot-cli/pull/4906/files#r1202547020

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/app.go
+++ b/internal/pkg/deploy/cloudformation/stack/app.go
@@ -33,17 +33,51 @@ type AppResourcesConfig struct {
 	Version  int                   `yaml:"Version"`
 }
 
-// AppResourcesService is a configuration for service for a deployed Application StackSet
+// AppResourcesService is a union type of either legacy or new
+// service configuration for a deployed Application StackSet
 type AppResourcesService struct {
+	Name   *LegacyAppResourcesServiceConfig `yaml:"Name,flow"`
+	Config AppResourcesServiceConfig
+}
+
+// LegacyAppResourcesServiceConfig is the legacy service configuration for a deployed Application StackSet
+type LegacyAppResourcesServiceConfig string
+
+// AppResourcesServiceConfig is a service configuration for a deployed Application StackSet
+type AppResourcesServiceConfig struct {
 	Name    string `yaml:"Name,flow"`
 	WithECR bool   `yaml:"WithECR,flow"`
+}
+
+func (s *AppResourcesServiceConfig) isEmpty() bool {
+	return s.Name == "" && !s.WithECR
 }
 
 // UnmarshalYAML overrides the default YAML unmarshaling logic for the Image
 // struct, allowing it to perform more complex unmarshaling behavior.
 // This method implements the yaml.Unmarshaler (v3) interface.
 func (s *AppResourcesService) UnmarshalYAML(value *yaml.Node) error {
-	return value.Decode(&s.Name)
+	if err := value.Decode(&s.Config); err != nil {
+		switch err.(type) {
+		case *yaml.TypeError:
+			break
+		default:
+			return err
+		}
+	}
+	if !s.Config.isEmpty() {
+		return nil
+	}
+	if err := value.Decode(&s.Name); err != nil {
+		return err
+	}
+	// Switch to new app resource service configuration and unset legacy configuration.
+	s.Config = AppResourcesServiceConfig{
+		Name:    string(*s.Name),
+		WithECR: true,
+	}
+	s.Name = nil
+	return nil
 }
 
 // AppStackConfig is for providing all the values to set up an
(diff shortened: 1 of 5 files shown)


internal/pkg/deploy/cloudformation/stack/app.go
```diff
@@ -33,17 +33,51 @@ type AppResourcesConfig struct {
...
+type AppResourcesServiceConfig struct {
 	Name    string `yaml:"Name,flow"`
 	WithECR bool   `yaml:"WithECR,flow"`
 }
 
+func (s *AppResourcesServiceConfig) isEmpty() bool {
+	return s.Name == "" && !s.WithECR
+}
+
 // UnmarshalYAML overrides the default YAML unmarshaling logic for the Image
```
Lou1415926, line 56:
```suggestion
// UnmarshalYAML overrides the default YAML unmarshaling logic for the AppResourcesService
```

internal/pkg/deploy/cloudformation/stack/app.go
```diff
@@ -33,17 +33,51 @@ type AppResourcesConfig struct {
 	Version  int                   `yaml:"Version"`
 }
 
-// AppResourcesService is a configuration for service for a deployed Application StackSet
+// AppResourcesService is a union type of either legacy or new
+// service configuration for a deployed Application StackSet
 type AppResourcesService struct {
+	Name   *LegacyAppResourcesServiceConfig `yaml:"Name,flow"`
```
Lou1415926, line 39: What about this:
```suggestion
	NameFromLegacyConfig   *string  // Deprecated
```
I am not sure if we need the YAML tag, since the legacy metadata would not have `Name` in it.

In fact, I'm not even sure if the legacy config needs to be in `AppResourcesService`. Could we just:

```go
func (s *AppResourcesService) UnmarshalYAML(value *yaml.Node) error {
	if err := value.Decode(&s.Config); err != nil {
		switch err.(type) {
		case *yaml.TypeError:
			break
		default:
			return err
		}
	}
	if !s.Config.isEmpty() {
		return nil
	}
        var legacyName string
	if err := value.Decode(&legacyName); err != nil {
		return err
	}
	// Switch to new app resource service configuration and unset legacy configuration.
	s.Config = AppResourcesServiceConfig{
		Name:    legacyName,
		WithECR: true,
	}
	return nil
}
```

internal/pkg/deploy/cloudformation/stack/app.go
```diff
@@ -33,17 +33,51 @@ type AppResourcesConfig struct {
...
+	Config AppResourcesServiceConfig
+}
+
+// LegacyAppResourcesServiceConfig is the legacy service configuration for a deployed Application StackSet
+type LegacyAppResourcesServiceConfig string
+
+// AppResourcesServiceConfig is a service configuration for a deployed Application StackSet
+type AppResourcesServiceConfig struct {
```
dannyrandall, line 47: nit: lowercase the yaml tags (`"name", "withECR"`)

internal/pkg/deploy/cloudformation/stack/app.go
dannyrandall: it seems like, other than `Unmarshal()`, we _never_ want to use `Name` - they should _always_ use `Config.Name`. maybe unexport it, since it's not needed to unmarshal+marshal anyways?

internal/pkg/deploy/cloudformation/stack/app.go
```diff
@@ -127,7 +161,12 @@ func (c *AppStackConfig) Template() (string, error) {
 func (c *AppStackConfig) ResourceTemplate(config *AppResourcesConfig) (string, error) {
 	// Sort the account IDs and Services so that the template we generate is deterministic
 	sort.Strings(config.Accounts)
-	sort.SliceStable(config.Services, func(i, j int) bool { return config.Services[i].Name < config.Services[j].Name })
+	sort.SliceStable(config.Services, func(i, j int) bool {
+		if config.Services[i].Name != nil {
+			return *config.Services[i].Name < *config.Services[j].Name
+		}
```
dannyrandall, line 167: similar to my above comment - won't `Services[i].Name` never be set?

internal/pkg/template/templates/app/cf.yml
```diff
@@ -8,7 +8,8 @@ Metadata:
   TemplateVersion: '{{.TemplateVersion}}'
   Version: {{.Version}}
   Services:{{if not $services}} []{{else}}{{range $service := $services}}
-  - {{$service.Name}}{{end}}{{end}}
+  - Name: {{$service.Config.Name}}
+    WithECR: {{$service.Config.WithECR}}{{end}}{{end}}
```
dannyrandall, line 12:
```suggestion
    - Name: {{$service.Config.Name}}
      WithECR: {{$service.Config.WithECR}}{{end}}{{end}}
```
nit (indent within services)

internal/pkg/template/templates/app/cf.yml
```diff
@@ -103,17 +104,17 @@ Resources:
...
+  ECRRepo{{logicalIDSafe $service.Config.Name}}:
     Metadata:
-      'aws:copilot:description': 'ECR container image repository for "{{$service.Name}}"'
+      'aws:copilot:description': 'ECR container image repository for "{{$service.Config.Name}}"'
     Type: AWS::ECR::Repository
     Properties:
-      RepositoryName: {{$app}}/{{$service.Name}}
+      RepositoryName: {{$app}}/{{$service.Config.Name}}
       Tags:
         -
```
dannyrandall, line 115: not you and nit but can we move `Key` to this line

internal/pkg/deploy/cloudformation/stack/app.go
iamhopaul123: We might want to make them consistent? It's capitalizing all the other fields

dannyrandall: ah ok, makes sense!

iamhopaul123:
Close in favor of #4924

PR #4924, opened 2023-05-23
chore: another fix broken app stack set for ecr management 

iamhopaul123:
Another way of dealing the problem as https://github.com/aws/copilot-cli/pull/4923 does

### DIFF
--- a/internal/pkg/deploy/cloudformation/stack/app.go
+++ b/internal/pkg/deploy/cloudformation/stack/app.go
@@ -21,62 +21,47 @@ import (
 // DeployedAppMetadata wraps the Metadata field of a deployed
 // application StackSet.
 type DeployedAppMetadata struct {
-	Metadata AppResourcesConfig `yaml:"Metadata"`
+	Metadata AppResources `yaml:"Metadata"`
 }
 
-// AppResourcesConfig is a configuration for a deployed Application
-// StackSet.
+// AppResources is a configuration for a deployed Application StackSet.
+type AppResources struct {
+	AppResourcesConfig `yaml:",inline"`
+}
+
+// AppResourcesConfig is a configuration for a deployed Application StackSet.
 type AppResourcesConfig struct {
-	Accounts []string              `yaml:"Accounts,flow"`
-	Services []AppResourcesService `yaml:"Services,flow"`
-	App      string                `yaml:"App"`
-	Version  int                   `yaml:"Version"`
+	Accounts  []string               `yaml:"Accounts,flow"`
+	Services  []string               `yaml:"Services,flow"`
+	Workloads []AppResourcesWorkload `yaml:"Workloads,flow"`
+	App       string                 `yaml:"App"`
+	Version   int                    `yaml:"Version"`
 }
 
-// AppResourcesService is a union type of either legacy or new
-// service configuration for a deployed Application StackSet
-type AppResourcesService struct {
-	Name   *LegacyAppResourcesServiceConfig `yaml:"Name,flow"`
-	Config AppResourcesServiceConfig
-}
-
-// LegacyAppResourcesServiceConfig is the legacy service configuration for a deployed Application StackSet
-type LegacyAppResourcesServiceConfig string
-
-// AppResourcesServiceConfig is a service configuration for a deployed Application StackSet
-type AppResourcesServiceConfig struct {
+// AppResourcesWorkload is a workload configuration for a deployed Application StackSet
+type AppResourcesWorkload struct {
 	Name    string `yaml:"Name,flow"`
 	WithECR bool   `yaml:"WithECR,flow"`
 }
 
-func (s *AppResourcesServiceConfig) isEmpty() bool {
+func (s *AppResourcesWorkload) isEmpty() bool {
 	return s.Name == "" && !s.WithECR
 }
 
 // UnmarshalYAML overrides the default YAML unmarshaling logic for the Image
 // struct, allowing it to perform more complex unmarshaling behavior.
 // This method implements the yaml.Unmarshaler (v3) interface.
-func (s *AppResourcesService) UnmarshalYAML(value *yaml.Node) error {
-	if err := value.Decode(&s.Config); err != nil {
-		switch err.(type) {
-		case *yaml.TypeError:
-			break
-		default:
-			return err
-		}
-	}
-	if !s.Config.isEmpty() {
-		return nil
-	}
-	if err := value.Decode(&s.Name); err != nil {
+func (s *AppResources) UnmarshalYAML(value *yaml.Node) error {
+	if err := value.Decode(&s.AppResourcesConfig); err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


internal/pkg/deploy/cloudformation/stack/app.go
```diff
@@ -21,29 +21,48 @@ import (
...
 // UnmarshalYAML overrides the default YAML unmarshaling logic for the Image
 // struct, allowing it to perform more complex unmarshaling behavior.
 // This method implements the yaml.Unmarshaler (v3) interface.
-func (s *AppResourcesService) UnmarshalYAML(value *yaml.Node) error {
-	return value.Decode(&s.Name)
+func (s *AppResources) UnmarshalYAML(value *yaml.Node) error {
+	if err := value.Decode(&s.AppResourcesConfig); err != nil {
+		return err
+	}
+	for _, svc := range s.AppResourcesConfig.Services {
```
Lou1415926, line 54: I think we have to do this _after_ `Unmarshal` finishes for the whole YAML

Edit: o I thought the receiver was `AppResourcesWorkload`

internal/pkg/deploy/cloudformation/stack/app.go
```diff
@@ -21,29 +21,48 @@ import (
...
+}
+
+// AppResourcesConfig is a configuration for a deployed Application StackSet.
 type AppResourcesConfig struct {
-	Accounts []string              `yaml:"Accounts,flow"`
-	Services []AppResourcesService `yaml:"Services,flow"`
-	App      string                `yaml:"App"`
-	Version  int                   `yaml:"Version"`
+	Accounts  []string               `yaml:"Accounts,flow"`
+	Services  []string               `yaml:"Services,flow"`
```
Lou1415926:
```suggestion
	Services  []string               `yaml:"Services,flow"` // Deprecated.
```

internal/pkg/deploy/cloudformation/stack/app_test.go
```diff
@@ -404,6 +411,54 @@ Metadata:
...
 		Version:  7,
-		Services: []AppResourcesService{{Name: "testsvc1"}, {Name: "testsvc2"}},
+		Workloads: []AppResourcesWorkload{
+			{Name: "testsvc1", WithECR: true},
+			{Name: "testsvc2", WithECR: true},
+		},
 	}, *config)
 }
+
+func TestAppResourcesService_UnmarshalYAML(t *testing.T) {
```
Lou1415926, line 421: Can we write these unit tests targeting the whole `AppResources`  instead of just the `Services` / `Workloads` fields?

internal/pkg/deploy/cloudformation/stack/app.go
dannyrandall: maybe a note how this is there for app templates <= v1.1.0?

internal/pkg/deploy/cloudformation/stack/app.go
```diff
@@ -21,29 +21,48 @@ import (
...
+	Workloads []AppResourcesWorkload `yaml:"Workloads,flow"`
```
dannyrandall: do we want to use flow for []maps? might get kinda messy?

internal/pkg/deploy/cloudformation/stack/app.go
dannyrandall: and some deprecated note, use `Workloads` instead

internal/pkg/template/templates/app/cf.yml
```diff
@@ -1,14 +1,15 @@
...
 # to support the CodePipeline for a workspace
 Description: Cross-regional resources to support the CodePipeline for a workspace
 Metadata:
   TemplateVersion: '{{.TemplateVersion}}'
   Version: {{.Version}}
-  Services:{{if not $services}} []{{else}}{{range $service := $services}}
-  - {{$service.Name}}{{end}}{{end}}
+  Workloads:{{if not $workloads}} []{{else}}{{range $workload := $workloads}}
+  - Name: {{$workload.Name}}
+    WithECR: {{$workload.WithECR}}{{end}}{{end}}
```
dannyrandall:
```suggestion
    - Name: {{$workload.Name}}
      WithECR: {{$workload.WithECR}}{{end}}{{end}}
```

Lou1415926 (commented):
LGTM! Just a small ask on the unit tests

internal/pkg/deploy/cloudformation/stack/app.go
```diff
@@ -21,29 +21,44 @@ import (
...
 // application StackSet.
 type DeployedAppMetadata struct {
-	Metadata AppResourcesConfig `yaml:"Metadata"`
+	Metadata AppResources `yaml:"Metadata"`
 }
 
-// AppResourcesConfig is a configuration for a deployed Application
-// StackSet.
+// AppResources is a configuration for a deployed Application StackSet.
+type AppResources struct {
```
dannyrandall, line 28: do we need this? why not only have `AppResourcesConfig`

internal/pkg/template/templates/app/cf.yml
```diff
@@ -1,16 +1,17 @@
 # Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: MIT-0
-AWSTemplateFormatVersion: '2010-09-09'{{$accounts := .Accounts}}{{$app := .App}}{{$services := .Services}}{{$svcTag := .ServiceTagKey}}
+AWSTemplateFormatVersion: '2010-09-09'{{$accounts := .Accounts}}{{$app := .App}}{{$workloads := .Workloads}}{{$svcTag := .ServiceTagKey}}
```
dannyrandall, line 3: wow i can't find it - where is .ServiceTagKey coming from?

dannyrandall (approved):
looks great! just some questions but otherwise good

Lou1415926 (approved):

internal/pkg/deploy/cloudformation/stack/app.go
iamhopaul123: because otherwise we won't be able to do the custom Unmarshal. 
```golang
func (s *AppResourcesConfig) UnmarshalYAML(value *yaml.Node) error {
	if err := value.Decode(&s); err != nil {
		return err
	}
```
just won't work.

dannyrandall: ah gotcha 👍

internal/pkg/template/templates/app/cf.yml
iamhopaul123: https://github.com/aws/copilot-cli/blob/5879fc0db12d0c13ce4d560b88ad03ca6db6f14d/internal/pkg/deploy/cloudformation/stack/app.go#L134

iamhopaul123: And the reason why i keep the name is because the tag is literally "copilot-service" :coolguy-crying

PR #4930, opened 2023-05-24
chore(e2e): change the alias used by `app-with-domain` 

Lou1415926:
Because it clashes with the alias used by `imported-cert` causing transient error sometimes.

### DIFF
--- a/e2e/app-with-domain/app_with_domain_test.go
+++ b/e2e/app-with-domain/app_with_domain_test.go
@@ -9,10 +9,8 @@ import (
 	"sync"
 	"time"
 
-	. "github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
-
 	"github.com/aws/copilot-cli/e2e/internal/client"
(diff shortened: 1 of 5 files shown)


dannyrandall (approved):

iamhopaul123 (approved):

iamhopaul123 (approved):

efekarakus (approved):
Maybe we should use a subdomain that hints at the test name "test-appdomain", "prod-appdomain", "test-importcert" so that in the future if people refer to these tests that can see that its namespaced by test name

Lou1415926:
> Maybe we should use a subdomain that hints at the test name "test-appdomain", "prod-appdomain", "test-importcert" so that in the future if people refer to these tests that can see that its namespaced by test name

Addressed, but left `copilot-e2e-tests.ecs.aws.dev` as one of the alias for the `frontend` in `app-with-domain`'s `test` environment, because I guess the intention was to make sure that the root domain can be used as an alias.

PR #4932, opened 2023-05-24
test(e2e): add multiple ports use case in app with domain e2e test

iamhopaul123:
Original PR https://github.com/aws/copilot-cli/pull/4795


e2e/app-with-domain/app_with_domain_suite_test.go
```diff
@@ -17,6 +18,7 @@ import (
 
 const (
 	waitingInterval = 60 * time.Second
+	domainName      = "copilot-e2e-tests.ecs.aws.dev"
```
Lou1415926: This is likely having conflicts now :doge

e2e/app-with-domain/copilot/frontend/manifest.yml
```diff
@@ -5,11 +5,16 @@ image:
   build: 
     dockerfile: src/Dockerfile
     context: src
-  port: 80
+  port: 3000
```
Lou1415926, line 8: Just curious - did we have specific reasons why we wanted to change the port from 80 to 3000? Or is it just a change with no extra considerations

e2e/app-with-domain/copilot/frontend/manifest.yml
```diff
@@ -5,11 +5,16 @@ image:
...
 
 http:
   path: '/'
-  alias: ["frontend-app-domain.copilot-e2e-tests.ecs.aws.dev", "copilot-e2e-tests.ecs.aws.dev"]
+  alias: ["frontend-app-domain.${DOMAINNAME}", "${DOMAINNAME}"]
+  additional_rules:
+    - path: 'admin'
+      target_port: 3002
+      healthcheck: '/admin'
+      alias: ["frontend-app-domain.${DOMAINNAME}", "${DOMAINNAME}"]
```
Lou1415926: I don't remember - what happens if there is no `http.additional_rules[x].alias` only `http.alias`? and what happens if they differ?

e2e/app-with-domain/app_with_domain_test.go
```diff
@@ -288,6 +288,7 @@ var _ = Describe("App With Domain", func() {
 			var resp *http.Response
 			var fetchErr error
 			urls := strings.Split(route.URL, " or ")
+			expectedResponseBody := "This is Copilot express app for admin"
```
Lou1415926: Do we want to hit the root path `/` or `/me` as well? I feel like this only proves that port `3002` is accepting the traffic.

e2e/app-with-domain/copilot/frontend/manifest.yml
iamhopaul123: they just have nothing to do with each other i think. And we'll use the default alias if it is not specified.

e2e/app-with-domain/copilot/frontend/manifest.yml
iamhopaul123: I think it's just to be more consistent with the new 3002? But then why 3000/3002 instead of 80/90 i think it's just personal preference.

e2e/app-with-domain/app_with_domain_test.go
iamhopaul123: Yeah we do it's just we don't check the response body. I'll update to be more consistent

e2e/app-with-domain/app_with_domain_test.go
```diff
@@ -288,9 +288,42 @@ var _ = Describe("App With Domain", func() {
...
+					if fetchErr != nil {
+						return 0, fetchErr
+					}
+					defer resp.Body.Close()
+					body, err := io.ReadAll(resp.Body)
+					if err != nil {
+						return 0, err
+					}
+					if string(body) != expectedRootResponseBody {
+						return 0, fmt.Errorf("the message content is '%q', but expected '%q'", string(body), expectedRootResponseBody)
```
Lou1415926: nit:
```suggestion
						return 0, fmt.Errorf("the message content is %q, but expected %q", string(body), expectedRootResponseBody)
```

Lou1415926 (approved):

huanjani (approved):
LGTM!

PR #4938, opened 2023-05-27
chore(cicd): fix several mistakes in e2e test buildspec

Lou1415926:

### DIFF
--- a/e2e/app-with-domain/app_with_domain_suite_test.go
+++ b/e2e/app-with-domain/app_with_domain_suite_test.go
@@ -33,7 +33,7 @@ var _ = BeforeSuite(func() {
 	copilotCLI, err := client.NewCLI()
 	cli = copilotCLI
 	Expect(err).NotTo(HaveOccurred())
-	appName = fmt.Sprintf("e2e-domain-%d", time.Now().Unix())
+	appName = fmt.Sprint(time.Now().Unix())
 })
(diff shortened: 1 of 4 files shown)


e2e/app-with-domain/app_with_domain_suite_test.go
```diff
@@ -33,7 +33,7 @@ var _ = BeforeSuite(func() {
 	copilotCLI, err := client.NewCLI()
 	cli = copilotCLI
 	Expect(err).NotTo(HaveOccurred())
-	appName = fmt.Sprintf("e2e-domain-%d", time.Now().Unix())
+	appName = fmt.Sprint(time.Now().Unix())
```
efekarakus, line 36: I guess no longer needed because of the separate accounts/test?

efekarakus (approved):

e2e/app-with-domain/app_with_domain_suite_test.go
Lou1415926: that is right. Also another important motivation was that, with `e2e-domain-xxx`, the certificate that we request per env would be: `test.e2e-domain-xxxxxx.app-with-domain.copilot.e2e.dev.test` which exceeds 60 characters limit from ACM

iamhopaul123 (approved):

PR #4946, opened 2023-05-30
chore(cicd): environments are created via. the profile of the same name

Lou1415926:
Simplify the pointer from environment name to profile name, so that we wouldn't have to look at the e2e test code to know the specific profile used.

### DIFF
--- a/.release/buildspec_e2e.yml
+++ b/.release/buildspec_e2e.yml
@@ -73,6 +73,8 @@ batch:
           TEST_SUITE: multi-svc-app
           APP_REGION: us-west-2
           APP_ACCOUNT: e2e/account/multi_svc_app
+          TEST_REGION: us-west-2
+          TEST_ACCOUNT: e2e/account/multi_svc_app
     - identifier: root
       env:
         privileged-mode: true
--- a/.release/buildspec_e2e.yml
+++ b/.release/buildspec_e2e.yml
@@ -188,6 +188,8 @@ batch:
           TEST_SUITE: cloudfront
           APP_REGION: eu-central-1
           APP_ACCOUNT: e2e/account/cloudfront
+          TEST_REGION: eu-central-1
+          TEST_ACCOUNT: e2e/account/cloudfront
 
 phases:
   install:
--- a/.release/buildspec_e2e.yml
+++ b/.release/buildspec_e2e.yml
@@ -136,6 +136,8 @@ batch:
           TEST_SUITE: exec
           APP_REGION: ap-northeast-1
           APP_ACCOUNT: e2e/account/exec
+          TESTENV_REGION: ap-northeast-1
+          TESTENV_ACCOUNT: e2e/account/exec
     - identifier: apprunner
       env:
         privileged-mode: true
--- a/.release/buildspec_e2e.yml
+++ b/.release/buildspec_e2e.yml
@@ -128,6 +128,18 @@ batch:
           APP_ACCOUNT: e2e/account/import_certs
           TESTENV_REGION: us-west-2
           TESTENV_ACCOUNT: e2e/account/import_certs
+    - identifier: isolated
+      env:
+        privileged-mode: true
+        type: LINUX_CONTAINER
+        compute-type: BUILD_GENERAL1_LARGE
+        image: aws/codebuild/standard:6.0
+        variables:
+          TEST_SUITE: isolated
+          APP_REGION: us-west-2
+          APP_ACCOUNT: e2e/account/isolated
+          TESTENV_REGION: us-west-2
+          TESTENV_ACCOUNT: e2e/account/isolated
     - identifier: exec
       env:
         privileged-mode: true
--- a/.release/buildspec_e2e.yml
+++ b/.release/buildspec_e2e.yml
@@ -94,6 +94,8 @@ batch:
           TEST_SUITE: sidecars
           APP_REGION: us-east-1
           APP_ACCOUNT: e2e/account/sidecars
+          TEST_REGION: us-east-1
+          TEST_ACCOUNT: e2e/account/sidecars
     - identifier: task
       env:
         privileged-mode: true
--- a/.release/buildspec_e2e.yml
+++ b/.release/buildspec_e2e.yml
@@ -106,6 +106,8 @@ batch:
           TEST_SUITE: app-with-domain
           APP_REGION: us-west-2
           APP_ACCOUNT: e2e/account/app_with_domains
+          TEST_REGION: us-west-2
+          TEST_ACCOUNT: e2e/account/app_with_domains
           PRODENV_REGION: us-east-1
           PRODENV_ACCOUNT: e2e/account/app_with_domains/prod
     - identifier: import_certs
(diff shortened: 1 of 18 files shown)


.release/buildspec_e2e.yml
```diff
@@ -21,6 +21,8 @@ batch:
           TEST_SUITE: addons
           APP_REGION: us-west-2
           APP_ACCOUNT: e2e/account/addons
+          TEST_REGION: us-west-2
+          TEST_ACCOUNT: e2e/account/addons
```
iamhopaul123: Can we make it optional and by default to be the same as app? Same for the region.

.release/buildspec_e2e.yml
```diff
@@ -31,6 +33,12 @@ batch:
           TEST_SUITE: customized-env
           APP_REGION: us-west-1
           APP_ACCOUNT: e2e/account/customized_env
+          TESTENV_REGION: us-west-1
+          TESTENV_ACCOUNT: e2e/account/customized_env
+          PRODENV_REGION: us-west-1
+          PRODENV_ACCOUNT: e2e/account/customized_env
+          SHAREDENV_REGION: us-west-1
```
iamhopaul123, line 40: why do we need this shared env 🤔

.release/buildspec_e2e.yml
```diff
@@ -31,6 +33,12 @@ batch:
           TEST_SUITE: customized-env
           APP_REGION: us-west-1
           APP_ACCOUNT: e2e/account/customized_env
+          TESTENV_REGION: us-west-1
```
iamhopaul123, line 36: Does it make more sense to combine `TEST_REGION` and `TESTENV_REGION`?

.release/buildspec_e2e.yml
```diff
@@ -210,16 +246,24 @@ phases:
...
+          printf "[profile test]\nregion=$TESTENV_REGION\n" >> /tmp/.aws/config
+          printf "[test]\n" >> /tmp/.aws/credentials
           printf "aws_access_key_id=$(echo $account | jq '.access_key' -r)\n" >> /tmp/.aws/credentials
           printf "aws_secret_access_key=$(echo $account | jq '.secret_access_key' -r)\n\n" >> /tmp/.aws/credentials
         fi
       - |
         if [ ! -z "$PRODENV_REGION" ]; then
           account=$(aws secretsmanager get-secret-value --secret-id ${PRODENV_ACCOUNT} | jq '.SecretString' -r)
-          printf "[profile e2eprodenv]\nregion=$PRODENV_REGION\n" >> /tmp/.aws/config
-          printf "[e2eprodenv]\n" >> /tmp/.aws/credentials
```
iamhopaul123, line 222: We probably need to update our contribute.md for this change. https://github.com/aws/copilot-cli/blob/mainline/CONTRIBUTING.md#building-and-testing

.release/buildspec_e2e.yml
Lou1415926: Because multi env has a "shared" environment 👀

.release/buildspec_e2e.yml
Lou1415926: good catch, those were typos. `TEST_REGION` was supposed to be  `TESTENV_REGION`

.release/buildspec_e2e.yml
iamhopaul123: Do we need an env var for that? Or it is supposed to be using the same profile as whatever environment it is shared with?

Lou1415926: The environment's name is "shared": https://github.com/aws/copilot-cli/blob/83f09c7decb89560667f171e278aa57d72a05853/e2e/customized-env/customized_env_test.go#L102

> Or it is supposed to be using the same profile as whatever environment it is shared with?

Did you mean create "shared" environment with the profile "test"?

.release/buildspec_e2e.yml
Lou1415926: I sort of wanted these information to be as obvious and plain-text-ed as possible. Like I don't want people to have to look at `phases.build` in order to understand what account / region is being used. 

Are you concerned about the number of lines in the buildspec?

.release/buildspec_e2e.yml
iamhopaul123: Yeah env name is "shared" doesn't necessarily mean we need a separate profile for that. Especially if two environments are sharing the same vpc, it is almost guaranteed that they should use the same profile.

iamhopaul123: Discussed offline we'll keep as is so that the oncall don't need to look up which profile we use for an env by checking the e2e test file.

CONTRIBUTING.md
```diff
@@ -47,16 +47,15 @@ Our integration tests ensure that we can call these remote services and get the
 **End to End tests** run the CLI in a container and test the actual commands - including spinning and tearing down remote resources (like ECS clusters and VPCs).
 These tests are the most comprehensive and run on both Windows and Linux build fleets.
 Feel free to run these tests - but they require two AWS accounts to run in, so be mindful that resources will be created and destroyed.
-You'll need three [named profiles](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html): `default`, `e2etestenv` and `e2eprodenv`.
-Each e2e profile needs to be configured for a different AWS account and a different region than the other e2e profiles.
+You'll need three [named profiles](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html): `default`, `test` and `prod`.
```
dannyrandall: what about `shared`? also, is the `default` one required now?

e2e/multi-env-app/multi_env_suite_test.go
```diff
@@ -15,17 +15,13 @@ import (
 
 var cli *client.CLI
 var appName string
-var testEnvironmentProfile string
-var prodEnvironmentProfile string
```
dannyrandall, line 19: were these unused?

dannyrandall (approved):
just a couple questions, but looks good!

e2e/multi-env-app/multi_env_suite_test.go
Lou1415926: they were used as in
```go
// In multi_env_suite_test.go
var testEnvironmentProfile string
var _ = BeforeSuite(func() {
	testEnvironmentProfile = "e2etestenv"
}

// In multi_env_test.go
_, err := copilot.EnvInit(&client.EnvInitRequest{
    AppName: appName,
    EnvName: "test",
    Profile: testEnvironmentProfile,
})
```

But since now the profile's name is always the same as the env name, we can just write 
```go
// In multi_env_test.go
_, err := copilot.EnvInit(&client.EnvInitRequest{
    AppName: appName,
    EnvName: "test",
    Profile:"test",
})
```
replacing `testEnvironmentProfile` with a plain sting `"test"`! So the var is no longer needed.

.release/buildspec_e2e.yml
huanjani: good catch!

iamhopaul123 (approved):

PR #4961, opened 2023-06-05
chore: add validation and recommended action for Listener rule if condition values exceed 5

### LINKED ISSUE #4601: Allow more than 5 Domains per Service
## Context
We have a multi tenant app with > 20 Domains. And cannot add more than 5 Domains to the ALB, cause the Matching Rules are not split.

## ToDo
If there are more than 5 Domains added to the ALB they should be split in multiple Matching Rules cause the maximum condition Count is 5.

More info here: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-limits.html

Gitter discussion: https://matrix.to/#/!QdxoBcgpJveoAoIPCc:gitter.im/$mjaMtBb694mAzgGB2p92KApi0CDxe-ad4j39yfWpLcA?via=gitter.im&via=matrix.org&via=matrix.martinides.de

KollaAdithya:
resolves #4601

Validate and add a Recommended Actions if they hit the quota of 5 conditions per rule.
Also logs the URI of all the host headers for `additional listener rules` as well.


cf-custom-resources/lib/alb-rule-priority-generator.js
```diff
@@ -165,16 +165,17 @@ exports.nextAvailableRulePriorityHandler = async function (event, context) {
       case "Create":
       case "Update":
           for (let i=0; i < event.ResourceProperties.RulePath.length; i++){
+            for(let j=0;j< parseInt(event.ResourceProperties.ConditionGroups[i]);j++){
```
iamhopaul123: Just wanted to double check: does this work for backward-compatibility? Like if previously they had `allowed_ip` to be `[“205.251.233.52/24", “192.0.2.0/25”]` and `alias` to be `example.com`, and got it deployed. After making this change, we'll split the rule into two. Can they successfully update the workload stack?

internal/pkg/template/workload.go
```diff
@@ -483,6 +489,114 @@ func (cfg *ALBListener) RulePaths() []string {
...
+
+func calculateConditionsHTTP(path string, rcPerRule int, wkldType string) int {
+	if wkldType == backendWkld {
+		rcPerRule = rcPerRule - 2
+	}
+	if path != rootPath {
+		return rcPerRule - 2
+	}
+	return rcPerRule - 1
+}
```
iamhopaul123: It doesn't make a lot of sense to me for these "1" and "2" without looking at the CFN template. I think this requires a lot of assumptions for how we write our CFN templates which is fragile. But I also fully understand the reason why we want to do this: to fulfill the requirement that we fully utilize the "5" max condition values per listener rule, so that we will be less likely to exceed the at most "100" listener rules per LB.

This makes me rethink if we should sacrifice this "fully utilizing CFN limit" to make the code robust and less likely to break for better maintainability 💭  In reality it is less likely they'll need that many aliases for __non-prod__ environment. I'm wondering if we could just assume the condition group num to be always "1" so that to greatly simplify the code. And we'll have "1" alias per listener rule. And the total number of listener rules we'll have is
```
totalNum = 0
for rule := range ${listener rules they define in the manifest)}
    totalNum += len(rule.allowed_ip} * len(rule.aliases)
```
Granted, we create more listener rules so it's easier to exceed the `100` listener rules per LB, but we can simplify this `template/workload.go` file as well as getting rid of the additional `ConditionGroups` param for priority function which is difficult to understand without looking through the codebase.

I'm also open for any suggestion regarding how we should solve this problem.

iamhopaul123 (commented):
This is great work! Thank you so much for making this change. I think the implementation overall makes sense to me for requirement we specify in the description. I also put some thought in the comments regarding the requirement itself, given the fact that it'll potentially be difficult to maintain and introducing more assumptions.

internal/pkg/template/workload.go
Lou1415926: I haven't read the code and the discussion closely (will do), but I'd just like to raise a point for early discussion. I think we should in general value a solution that does _not_ involve updating the custom resource (such as rule priority calculator) if that's possible at all. Because it's easy for customers to override the CFN template now (via CDK override), but it's still a huge trouble for them to update the custom resource to match what they've overriden.

internal/pkg/template/workload.go
```diff
@@ -483,6 +489,114 @@ func (cfg *ALBListener) RulePaths() []string {
...
+	Aliases          []string
+}
+
+// GenerateConditionGroups will split allowed source ips and aliases into separate condition groups per listener rule.
+// This is needed because Listener rule has quota of 5 conditions per rule.
+// https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-limits.html
+func (lr ALBListenerRule) GenerateConditionGroups(wkldType string, isHTTPS bool) []conditionGroup {
+	var groups []conditionGroup
+	remaining := lr.calculateRemainingConditions(wkldType, isHTTPS)
+	if len(lr.Aliases)+len(lr.AllowedSourceIps) <= remaining {
```
Lou1415926: I _think_ there is no relation between the number of aliases and the number of allowed source IPs. i.e. you can have:
```yaml
Conditions:
      - Field: 'source-ip'
        SourceIpConfig:
          Values: [ip1, ip2, ip3, ip4, ip5]
      - Field: 'host-header'
        HostHeaderConfig:
          Values: [alias1, alias2, alias3, alias4, alias5]
```

internal/pkg/template/workload.go
```diff
@@ -483,6 +489,114 @@ func (cfg *ALBListener) RulePaths() []string {
 	return rulePaths
 }
 
+// ConditionGroups returns a slice of all the Splits on the Conditions across multiple listener rules.
+func (cfg *ALBListener) ConditionGroups(wkldType string, isHTTPS bool) []string {
```
Lou1415926: Based on what it does, I would probably name this function something like `GroupCountPerRule`. `ConditionGroups` implies that the return value is a list of condition groups, not a list of ints.

internal/pkg/template/workload.go
```diff
@@ -74,6 +74,12 @@ const (
 	NoExposedContainerPort = "-1"
 )
 
+const (
+	totalConditionsPerRule = 5
+	rootPath               = "/"
+	backendWkld            = "Backend Service"
```
Lou1415926: nit nit nit:
```suggestion
	backendWkldType            = "Backend Service"
```

Lou1415926 (commented):
Noice code - very easy to follow for such complex logic! 

I left a few feedbacks, mostly nits, on the codes. But don't worry about addressing them for now, because I think @iamhopaul123 raises a valid point that we should address first!

internal/pkg/template/workload.go
Lou1415926: I think I will plus 1 to @iamhopaul123. A bonus for what boss suggests is ⬆️  easier for customers to do CDK override in case they really do hit the 100 rules limits

internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -39,34 +40,36 @@ func New(sess *session.Session) *ELBV2 {
...
-		if aws.StringValue(condition.Field) == "host-header" {
-			// Values is a legacy field that allowed specifying only a single host name.
-			// The alternative is to use HostHeaderConfig for multiple values.
-			// Only one of these fields should be set, but we collect from both to be safe.
-			for _, value := range condition.Values {
-				hostHeaderSet[aws.StringValue(value)] = true
-			}
-			if condition.HostHeaderConfig == nil {
+	for idx := range ruleARNs {
+		rule := resp.Rules[idx]
```
Lou1415926: Is it always true that `len(resp.Rules) >= len(ruleARNs)`? If one of the `ruleARNs` is not found, what does the API call return?

internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -39,34 +40,36 @@ func New(sess *session.Session) *ELBV2 {
...
 	if err != nil {
-		return nil, fmt.Errorf("get listener rule for %s: %w", ruleARN, err)
+		return nil, fmt.Errorf("get listener rule for %s: %w", english.WordSeries(ruleARNs, "and"), err)
 	}
 	if len(resp.Rules) == 0 {
-		return nil, fmt.Errorf("cannot find listener rule %s", ruleARN)
+		return nil, fmt.Errorf("cannot find listener rule %s", english.WordSeries(ruleARNs, "and"))
 	}
-	rule := resp.Rules[0]
 	hostHeaderSet := make(map[string]bool)
```
Lou1415926: nit: what do you think about
```suggestion
       exists := struct{}{}
	hostHeaderSet := make(map[string]struct{})
	...
	for _, value := range condition.Values {
					hostHeaderSet[aws.StringValue(value)] = exists
				}
```
I can think of two benefit of this - 
1. If we use `bool`, then it's ambiguous what does `hostHeaderSet["this"] = false` means.
2. `struct{}` takes 0 byte

internal/pkg/aws/elbv2/elbv2_test.go
```diff
@@ -144,33 +144,37 @@ func TestELBV2_TargetsHealth(t *testing.T) {
...
 
 func TestELBV2_ListenerRuleHostHeaders(t *testing.T) {
-	mockARN := "mockListenerRuleARN"
+	mockARN1 := "mockListenerRuleARN1"
+	mockARN2 := "mockListenerRuleARN2"
 	testCases := map[string]struct {
-		setUpMock func(m *mocks.Mockapi)
-
+		setUpMock   func(m *mocks.Mockapi)
+		inmockARNs  []string
```
Lou1415926: nit:
```suggestion
		inARNs  []string
```
I think the `mock` keyword should be added to dependencies that are assumed to work, not to the input of the functions. I'd think of a function as a black box that maps an input to the output; the dependencies are inside of the black box and need to be mocked 🤔 💭

internal/pkg/cli/deploy/svc.go
```diff
@@ -198,3 +203,37 @@ func (e *errInvalidAlias) RecommmendActions() string {
...
+	return nil
+}
+
+type errMaxConditionValuesPerRule struct {
+	aliases          []string
+	allowedSourceIps []string
+}
+
+func (e *errMaxConditionValuesPerRule) Error() string {
+	return fmt.Sprintf("Listener Rule can not have more than five Condition Values %s %s", english.WordSeries(e.aliases, "and"),
```
Lou1415926:
```suggestion
	return fmt.Sprintf("listener rule has more than five conditions %s %s", english.WordSeries(e.aliases, "and"),
```

internal/pkg/cli/deploy/svc.go
```diff
@@ -198,3 +203,37 @@ func (e *errInvalidAlias) RecommmendActions() string {
...
+		english.WordSeries(e.allowedSourceIps, "and"))
+}
+
+func (e *errMaxConditionValuesPerRule) RecommendedActions() string {
+	return fmt.Sprintf(`You can split the aliases in the manifest with "http.additional_rules" with same path and container port %s`, color.Emphasize(`
```
Lou1415926: since we already know the `aliases` that they have, do you think we could output a customized snippets using the `aliases` 🤔 ?

internal/pkg/cli/deploy/lbws.go
```diff
@@ -245,6 +245,13 @@ func (d *lbWebSvcDeployer) validateRuntimeRoutingRule(rule manifest.RoutingRule)
 				return fmt.Errorf("validate aliases against the imported CDN certificate for env %s: %w", d.env.Name, err)
 			}
 		}
+		allowedSourceIPs := make([]string, len(rule.AllowedSourceIps))
+		for idx, IP := range rule.AllowedSourceIps {
+			allowedSourceIPs[idx] = string(IP)
+		}
+		if err := validateConditionValuesPerRule(aliases, allowedSourceIPs); err != nil {
+			return fmt.Errorf("validate condition values per listener rule:%w", err)
```
Lou1415926:
```suggestion
			return fmt.Errorf("validate condition values per listener rule: %w", err)
```

internal/pkg/cli/deploy/backend.go
```diff
@@ -163,5 +163,12 @@ func (d *backendSvcDeployer) validateRuntimeRoutingRule(rule manifest.RoutingRul
 		return fmt.Errorf("validate aliases against the imported certificate for env %s: %w", d.env.Name, err)
 	}
 
+	allowedSourceIPs := make([]string, len(rule.AllowedSourceIps))
+	for idx, IP := range rule.AllowedSourceIps {
+		allowedSourceIPs[idx] = string(IP)
+	}
+	if err := validateConditionValuesPerRule(aliases, allowedSourceIPs); err != nil {
+		return fmt.Errorf("validate condition values per listener rule:%w", err)
```
iamhopaul123:
```suggestion
		return fmt.Errorf("validate condition values per listener rule: %w", err)
```
same for lbws

internal/pkg/cli/deploy/backend.go
```diff
@@ -163,5 +163,12 @@ func (d *backendSvcDeployer) validateRuntimeRoutingRule(rule manifest.RoutingRul
 		return fmt.Errorf("validate aliases against the imported certificate for env %s: %w", d.env.Name, err)
 	}
 
+	allowedSourceIPs := make([]string, len(rule.AllowedSourceIps))
+	for idx, IP := range rule.AllowedSourceIps {
+		allowedSourceIPs[idx] = string(IP)
```
iamhopaul123:
```suggestion
	for idx, ip := range rule.AllowedSourceIps {
		allowedSourceIPs[idx] = string(ip)
```

internal/pkg/cli/deploy/svc.go
```diff
@@ -21,6 +22,10 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/term/log"
 )
 
+const (
+	maxConditionsPerRule = 5
```
iamhopaul123: maybe add a comment with proper reference.

internal/pkg/cli/deploy/svc.go
```diff
@@ -198,3 +203,37 @@ func (e *errInvalidAlias) RecommmendActions() string {
...
+
+type errMaxConditionValuesPerRule struct {
+	aliases          []string
+	allowedSourceIps []string
+}
+
+func (e *errMaxConditionValuesPerRule) Error() string {
+	return fmt.Sprintf("Listener Rule can not have more than five Condition Values %s %s", english.WordSeries(e.aliases, "and"),
+		english.WordSeries(e.allowedSourceIps, "and"))
+}
```
iamhopaul123: Why not move this validation to manifest validation 🤔

internal/pkg/describe/uri.go
```diff
@@ -282,23 +283,22 @@ func (d *uriDescriber) uri() (accessURI, error) {
...
+	var ruleARNs []string
 	for _, resource := range svcResources {
 		if resource.Type == svcStackResourceListenerRuleResourceType &&
-			((httpsEnabled && resource.LogicalID == svcStackResourceHTTPSListenerRuleLogicalID) ||
-				(!httpsEnabled && resource.LogicalID == svcStackResourceHTTPListenerRuleLogicalID)) {
-			ruleARN = resource.PhysicalID
-			break
+			(httpsEnabled && strings.HasPrefix(resource.LogicalID, svcStackResourceHTTPSListenerRuleLogicalID)) ||
+			(!httpsEnabled && strings.HasPrefix(resource.LogicalID, svcStackResourceHTTPListenerRuleLogicalID)) {
+			ruleARNs = append(ruleARNs, resource.PhysicalID)
```
iamhopaul123: Was this a bug previously? The fact that we only use the first matched listener rule arn.

KollaAdithya: yeah that's true. we log only the URI of only `HTTPSLISTENERRULE` but not URI in `HTTPSLISTENERRULE1`

internal/pkg/aws/elbv2/elbv2.go
KollaAdithya: I tested this out. On purpose I provided wrong ARN to the `aws elbv2 describe-rules --rule-arns RuleARN1 RuleARN2` It fetched an error `An error occurred (RuleNotFound) when calling the DescribeRules operation: One or more rules not found`.

Even if the API call  has `aws elbv2 describe-rules --rule-arns RuleARN1 RuleARN1` then response will have dupilcated and len(resp.Rules)=2.

So it always be `len(resp.Rules)=len(ruleARNS)`

internal/pkg/cli/deploy/svc.go
KollaAdithya: done!

internal/pkg/manifest/validate.go
```diff
@@ -45,6 +46,12 @@ const (
 	awsXRAY = "awsxray"
 )
 
+const (
+	// listener rule has a quota of five condition values per rule.
+	// please refer to https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-limits.html
```
huanjani: nit:
```suggestion
	// Listener rules have a quota of five condition values per rule.
	// Please refer to https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-limits.html.
```

iamhopaul123 (approved):

internal/pkg/aws/elbv2/elbv2.go
```diff
@@ -39,34 +40,37 @@ func New(sess *session.Session) *ELBV2 {
...
-			// The alternative is to use HostHeaderConfig for multiple values.
-			// Only one of these fields should be set, but we collect from both to be safe.
-			for _, value := range condition.Values {
-				hostHeaderSet[aws.StringValue(value)] = true
-			}
-			if condition.HostHeaderConfig == nil {
+	exists := struct{}{}
+	hostHeaderSet := make(map[string]struct{})
+	for idx := range ruleARNs {
+		rule := resp.Rules[idx]
```
Lou1415926: I wonder if we could just:
```suggestion
	for _, rule := range resp.Rules{
```

internal/pkg/describe/uri.go
```diff
@@ -282,23 +283,22 @@ func (d *uriDescriber) uri() (accessURI, error) {
...
-	var ruleARN string
+	var ruleARNs []string
 	for _, resource := range svcResources {
 		if resource.Type == svcStackResourceListenerRuleResourceType &&
-			((httpsEnabled && resource.LogicalID == svcStackResourceHTTPSListenerRuleLogicalID) ||
-				(!httpsEnabled && resource.LogicalID == svcStackResourceHTTPListenerRuleLogicalID)) {
-			ruleARN = resource.PhysicalID
-			break
+			(httpsEnabled && strings.HasPrefix(resource.LogicalID, svcStackResourceHTTPSListenerRuleLogicalID)) ||
+			(!httpsEnabled && strings.HasPrefix(resource.LogicalID, svcStackResourceHTTPListenerRuleLogicalID)) {
```
Lou1415926: Correct me if I'm wrong - originally this was `x && (a||b)`, now it is `x && a || b`. I think `&&` has higher evaluation priority than `||`, therefore it seems like removing the parenthesis changes the condition.

internal/pkg/describe/uri_test.go
```diff
@@ -115,16 +115,21 @@ func TestLBWebServiceDescriber_URI(t *testing.T) {
 					}, nil),
 					m.ecsDescriber.EXPECT().StackResources().Return([]*describeStack.Resource{
 						{
-							LogicalID:  svcStackResourceHTTPSListenerRuleLogicalID,
+							LogicalID:  "HTTPSListenerRuleGroup0",
```
Lou1415926: nit: can we change the logical ID to the real logical ID that we are using today - `HTTPSListenerRule0`,  `HTTPSListenerRule1`

internal/pkg/manifest/validate.go
```diff
@@ -45,6 +46,12 @@ const (
 	awsXRAY = "awsxray"
 )
 
+const (
+	// Listener rules have a quota of five condition values per rule.
```
Lou1415926, line 50: nice comment!

internal/pkg/manifest/validate.go
```diff
@@ -895,6 +902,17 @@ func (r RoutingRule) validate() error {
...
 	}
+	aliases, err := r.Alias.ToStringSlice()
+	if err != nil {
+		return fmt.Errorf("convert aliases to string slice: %w", err)
+	}
+	allowedSourceIps := make([]string, len(r.AllowedSourceIps))
+	for idx, ip := range r.AllowedSourceIps {
+		allowedSourceIps[idx] = string(ip)
+	}
+	if err := validateConditionValuesPerRule(aliases, allowedSourceIps); err != nil {
```
Lou1415926: Could we just pass `r.AllowedSourceIps` into this function, and change the function signature to accept `[]IPNet`?

internal/pkg/manifest/validate.go
```diff
@@ -2230,3 +2248,52 @@ func (i ImageLocationOrBuild) validate() error {
 	}
 	return nil
 }
+
+func validateConditionValuesPerRule(aliases []string, allowedSourceIPs []string) error {
+	if (len(aliases) >= maxConditionsPerRule) || (len(allowedSourceIPs) >= maxConditionsPerRule) ||
+		(len(aliases)+len(allowedSourceIPs) >= maxConditionsPerRule) {
```
Lou1415926: Could this be simplified to ⬇️ ?
```suggestion
	if len(aliases)+len(allowedSourceIPs) >= maxConditionsPerRule {
```

internal/pkg/manifest/validate.go
```diff
@@ -2230,3 +2248,52 @@ func (i ImageLocationOrBuild) validate() error {
...
+	return nil
+}
+
+type errMaxConditionValuesPerRule struct {
+	aliases          []string
+	allowedSourceIps []string
+}
+
+func (e *errMaxConditionValuesPerRule) Error() string {
+	return fmt.Sprintf("listener rule has more than five conditions %s %s", english.WordSeries(e.aliases, "and"),
```
Lou1415926, line 2271: This made me realize how exposed the infrastructure is 😂 ... I think ideally people use `additional_rules` only when they want an additional `path` or `alias`.

This is not a one way door and I'm perfectly fine with our current approach as an immediate remedy. I just think we as a team should rethink about this approach some day 😂

internal/pkg/manifest/validate.go
```diff
@@ -2230,3 +2248,52 @@ func (i ImageLocationOrBuild) validate() error {
...
+		english.WordSeries(e.allowedSourceIps, "and"))
+}
+
+func (e *errMaxConditionValuesPerRule) RecommendActions() string {
+	return fmt.Sprintf(`You can split the aliases in the manifest with "http.additional_rules" with the same path and container port:
```
```suggestion
	return fmt.Sprintf(`You can split the "alias" and "allowed_source_ips" field into separate rules, so that each rule contains up to 5 values:
```

internal/pkg/manifest/validate.go
```diff
@@ -2230,3 +2248,52 @@ func (i ImageLocationOrBuild) validate() error {
...
+%s`, color.HighlightCodeBlock(fmt.Sprintf(`http:
+  path: "/"
+  alias: ["%s", "%s"]
+  allowed_source_ips: ["192.0.2.0/24", "198.51.100.10/32"]
+  additional_rules: 
+%s`, e.aliases[0], e.aliases[1], fmtAdditionalRules(e.aliases))))
+}
+
+func fmtAdditionalRules(aliases []string) string {
```
Lou1415926: What do we think of a function that takes in `aliases` and `sourceIPs` ? We can use the same splitting logic that you had in the previous version when we did `RuleGroupX`.

internal/pkg/describe/uri.go
KollaAdithya: yes you are right :home disapppearing:

internal/pkg/manifest/validate.go
KollaAdithya: I kept this `typecast to string` as is, because I want to use `english.WordSeries` and also considering this [logic](https://github.com/aws/copilot-cli/pull/4961/files#diff-5fc13795122fe030799a09ef956a8b66cec3d40bf1a81266323edbf3626b5685R2343-R2349) where I need to have both aliases and allowedsourceIps to be `[]string`

internal/pkg/manifest/validate.go
```diff
@@ -2271,29 +2273,92 @@ func (e *errMaxConditionValuesPerRule) Error() string {
...
+	cgList := e.generateConditionGroups()
+	var fmtListenerRules strings.Builder
+	fmtListenerRules.WriteString(fmt.Sprintf(`http:
+  path: %s
+  alias: %s
+  allowed_source_ips: %s
+  additional_rules:`, e.path, fmtStringArray(cgList[0].aliases), fmtStringArray(cgList[0].allowedSourceIps)))
+	for i := 1; i < len(cgList); i++ {
+		fmtListenerRules.WriteString(fmt.Sprintf(`
+    path: %s
```
```suggestion
  - path: %s
```

Lou1415926 (approved):

PR #4966, opened 2023-06-07
chore(cli): add S3 bucket object tree to `svc show`

huanjani:
sample output:
```
% copilot svc show 
Service name: static-site
About

  Application  bugbash-static
  Name         static-site
  Type         Static Site

Routes

  Environment  URL
  -----------  ---
  test         d1vwytbnh1k1gy.cloudfront.net
  prod         d1yb44q5409rdj.cloudfront.net

S3 Bucket Objects

  Environment  test
.
├── error.html
├── index.html
├── css
│   ├── Style.css
│   ├── all.min.css
│   └── bootstrap.min.css
└── images
    ├── bg-masthead.jpg
    └── new
        └── hi.html

  Environment  prod
.
├── ReadMe.md
├── error.html
├── index.html
├── Images
│   ├── AWS-free.PNG
│   ├── S3WebsiteHosting-Architecture.PNG
│   ├── SampleWebsite.PNG
│   ├── Step1-CreateBucket-B.PNG
│   ├── Step1-CreateBucket-C.PNG
│   ├── Step1-CreateBucket-D.PNG
│   ├── Step1-CreateBucket-E.PNG
│   ├── Step1-CreateBucket.PNG
│   ├── Step2-UploadContents.PNG
│   ├── Step3-BucketPolicy-A.PNG
│   ├── Step3-BucketPolicy-B.PNG
│   ├── Step3-BucketPolicy.PNG
│   ├── Step4-A.PNG
│   └── Step4-B.PNG
├── css
│   ├── Style.css
│   ├── all.min.css
│   └── bootstrap.min.css
└── images
    ├── bg-masthead.jpg
    └── new
        └── hi.html
```

### DIFF
--- a/internal/pkg/aws/s3/s3.go
+++ b/internal/pkg/aws/s3/s3.go
@@ -219,6 +219,8 @@ func (s *S3) GetBucketTree(bucket string) (string, error) {
 		return "", nil
 	}
 
+	var contents []*s3.Object
+	var prefixes []*s3.CommonPrefix
 	listResp := &s3.ListObjectsV2Output{}
 	for {
 		listParams := &s3.ListObjectsV2Input{
@@ -230,6 +232,8 @@ func (s *S3) GetBucketTree(bucket string) (string, error) {
 		if err != nil {
 			return "", fmt.Errorf("list objects for bucket %s: %w", bucket, err)
 		}
+		contents = append(contents, listResp.Contents...)
+		prefixes = append(prefixes, listResp.CommonPrefixes...)
 		if listResp.NextContinuationToken == nil {
 			break
 		}
@@ -237,11 +241,11 @@ func (s *S3) GetBucketTree(bucket string) (string, error) {
 
 	tree := treeprint.New()
 	// Add top-level files.
-	for _, object := range listResp.Contents {
+	for _, object := range contents {
 		tree.AddNode(aws.StringValue(object.Key))
 	}
 	// Recursively add folders and their children.
-	if err := s.addNodes(tree, listResp.CommonPrefixes, bucket); err != nil {
+	if err := s.addNodes(tree, prefixes, bucket); err != nil {
 		return "", err
 	}
 	return tree.String(), nil
@@ -251,9 +255,12 @@ func (s *S3) addNodes(tree treeprint.Tree, prefixes []*s3.CommonPrefix, bucket s
 	if len(prefixes) == 0 {
 		return nil
 	}
+
 	listResp := &s3.ListObjectsV2Output{}
 	var err error
 	for _, prefix := range prefixes {
+		var respContents []*s3.Object
+		var respPrefixes []*s3.CommonPrefix
 		branch := tree.AddBranch(filepath.Base(aws.StringValue(prefix.Prefix)))
 		for {
 			listParams := &s3.ListObjectsV2Input{
@@ -266,15 +273,17 @@ func (s *S3) addNodes(tree treeprint.Tree, prefixes []*s3.CommonPrefix, bucket s
 			if err != nil {
 				return fmt.Errorf("list objects for bucket %s: %w", bucket, err)
 			}
+			respContents = append(respContents, listResp.Contents...)
+			respPrefixes = append(respPrefixes, listResp.CommonPrefixes...)
 			if listResp.NextContinuationToken == nil {
 				break
 			}
 		}
-		for _, file := range listResp.Contents {
+		for _, file := range respContents {
 			fileName := filepath.Base(aws.StringValue(file.Key))
 			branch.AddNode(fileName)
 		}
-		if err := s.addNodes(branch, listResp.CommonPrefixes, bucket); err != nil {
+		if err := s.addNodes(branch, respPrefixes, bucket); err != nil {
 			return err
 		}
 	}
--- a/internal/pkg/aws/s3/s3.go
+++ b/internal/pkg/aws/s3/s3.go
@@ -221,6 +221,7 @@ func (s *S3) GetBucketTree(bucket string) (string, error) {
 	if !exists {
 		return "", nil
 	}
+
 	for {
 		listParams := &s3.ListObjectsV2Input{
 			Bucket:            aws.String(bucket),
@@ -243,7 +244,7 @@ func (s *S3) GetBucketTree(bucket string) (string, error) {
 	}
 	// Recursively add folders and their children.
 	if err := s.addNodes(tree, listResp.CommonPrefixes, bucket); err != nil {
-		return tree.String(), err
+		return "", err
 	}
 	return tree.String(), nil
 }
@@ -255,7 +256,7 @@ func (s *S3) addNodes(tree treeprint.Tree, prefixes []*s3.CommonPrefix, bucket s
 	listResp := &s3.ListObjectsV2Output{}
 	var err error
 	for _, prefix := range prefixes {
-		branch := tree.AddBranch(s.extractFileName(aws.StringValue(prefix.Prefix)))
+		branch := tree.AddBranch(s.extractNameFromPath(aws.StringValue(prefix.Prefix)))
 		for {
 			listParams := &s3.ListObjectsV2Input{
 				Bucket:            aws.String(bucket),
@@ -272,7 +273,7 @@ func (s *S3) addNodes(tree treeprint.Tree, prefixes []*s3.CommonPrefix, bucket s
 			}
 		}
 		for _, file := range listResp.Contents {
-			fileName := s.extractFileName(aws.StringValue(file.Key))
+			fileName := s.extractNameFromPath(aws.StringValue(file.Key))
 			branch.AddNode(fileName)
 		}
 		if err := s.addNodes(branch, listResp.CommonPrefixes, bucket); err != nil {
@@ -282,9 +283,9 @@ func (s *S3) addNodes(tree treeprint.Tree, prefixes []*s3.CommonPrefix, bucket s
 	return nil
 }
 
-// extractFileName returns the last element of a path, whether the name of
-// an individual file or the end of an S3 prefix.
-func (s *S3) extractFileName(path string) string {
+// extractNameFromPath returns the last element of a path, whether the name of
+// an individual file or the end of an S3 prefix (folder name).
+func (s *S3) extractNameFromPath(path string) string {
 	path, _ = strings.CutSuffix(path, `/`)
 	if strings.Contains(path, `/`) {
 		split := strings.Split(path, `/`)
--- a/internal/pkg/aws/s3/s3.go
+++ b/internal/pkg/aws/s3/s3.go
@@ -209,17 +209,17 @@ func (s *S3) isBucketExists(bucket string) (bool, error) {
 }
 
 // GetBucketTree retrieves the objects in an S3 bucket and creates an ASCII tree representing their folder structure.
-func (s *S3) GetBucketTree(bucket string) error {
+func (s *S3) GetBucketTree(bucket string) (treeprint.Tree, error) {
 	listResp := &s3.ListObjectsV2Output{}
 	var err error
 
 	// isBucketExists check to make sure the bucket exists before proceeding.
 	isExists, err := s.isBucketExists(bucket)
 	if err != nil {
-		return fmt.Errorf("unable to determine the existence of bucket %s: %w", bucket, err)
+		return nil, fmt.Errorf("unable to determine the existence of bucket %s: %w", bucket, err)
 	}
 	if !isExists {
-		return nil
+		return nil, nil
 	}
 	for {
 		listParams := &s3.ListObjectsV2Input{
@@ -229,7 +229,7 @@ func (s *S3) GetBucketTree(bucket string) error {
 		}
 		listResp, err = s.s3Client.ListObjectsV2(listParams)
 		if err != nil {
-			return fmt.Errorf("list objects for bucket %s: %w", bucket, err)
+			return nil, fmt.Errorf("list objects for bucket %s: %w", bucket, err)
 		}
 		if listResp.NextContinuationToken == nil {
 			break
@@ -238,14 +238,13 @@ func (s *S3) GetBucketTree(bucket string) error {
 
 	tree := treeprint.New()
 	if err := s.addNodes(tree, listResp.CommonPrefixes, bucket); err != nil {
-		return err
+		return tree, err
 	}
 
 	for _, object := range listResp.Contents {
 		tree.AddNode(aws.StringValue(object.Key))
 	}
-	fmt.Println(tree.String())
-	return nil
+	return tree, nil
 }
 
 func (s *S3) addNodes(tree treeprint.Tree, prefixes []*s3.CommonPrefix, bucket string) error {
--- a/internal/pkg/aws/s3/s3.go
+++ b/internal/pkg/aws/s3/s3.go
@@ -82,12 +82,12 @@ func (s *S3) EmptyBucket(bucket string) error {
 	var listResp *s3.ListObjectVersionsOutput
 	var err error
 
-	isBucket, err := s.isBucket(bucket)
+	bucketExists, err := s.bucketExists(bucket)
 	if err != nil {
 		return fmt.Errorf("unable to determine the existence of bucket %s: %w", bucket, err)
 	}
 
-	if !isBucket {
+	if !bucketExists {
 		return nil
 	}
 
@@ -194,7 +194,7 @@ func FormatARN(partition, location string) string {
 	return fmt.Sprintf("arn:%s:s3:::%s", partition, location)
 }
 
-func (s *S3) isBucket(bucket string) (bool, error) {
+func (s *S3) bucketExists(bucket string) (bool, error) {
 	input := &s3.HeadBucketInput{
 		Bucket: aws.String(bucket),
 	}
@@ -211,7 +211,7 @@ func (s *S3) isBucket(bucket string) (bool, error) {
 
 // GetBucketTree retrieves the objects in an S3 bucket and creates an ASCII tree representing their folder structure.
 func (s *S3) GetBucketTree(bucket string) (string, error) {
-	exists, err := s.isBucket(bucket)
+	exists, err := s.bucketExists(bucket)
 	if err != nil {
 		return "", fmt.Errorf("unable to determine the existence of bucket %s: %w", bucket, err)
 	}
--- a/internal/pkg/aws/s3/s3.go
+++ b/internal/pkg/aws/s3/s3.go
@@ -79,13 +79,13 @@ func (s *S3) EmptyBucket(bucket string) error {
 	var listResp *s3.ListObjectVersionsOutput
 	var err error
 
-	// isBucketExists checks to make sure the bucket exists before proceeding to empty it.
-	isExists, err := s.isBucketExists(bucket)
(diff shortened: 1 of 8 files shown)


internal/pkg/aws/s3/s3.go
```diff
@@ -77,13 +79,13 @@ func (s *S3) EmptyBucket(bucket string) error {
 	var listResp *s3.ListObjectVersionsOutput
 	var err error
 
-	// Bucket is exists check to make sure the bucket exists before proceeding in emptying it
-	isExists, err := s.isBucketExists(bucket)
+	// isBucket checks to make sure the bucket exists before proceeding to empty it.
```
iamhopaul123: we probably don't need this comment as the code pretty much 
 explains itself.

internal/pkg/aws/s3/s3.go
```diff
@@ -190,8 +192,8 @@ func FormatARN(partition, location string) string {
 	return fmt.Sprintf("arn:%s:s3:::%s", partition, location)
 }
 
-// Check whether the bucket exists before proceeding with empty the bucket
-func (s *S3) isBucketExists(bucket string) (bool, error) {
+// isBucket checks whether the bucket exists before emptying the bucket
```
iamhopaul123: same and this is probably inappropriate too as it is not only describing what it does.

internal/pkg/aws/s3/s3.go
```diff
@@ -206,6 +208,92 @@ func (s *S3) isBucketExists(bucket string) (bool, error) {
 	return true, nil
 }
 
+// GetBucketTree retrieves the objects in an S3 bucket and creates an ASCII tree representing their folder structure.
+func (s *S3) GetBucketTree(bucket string) (string, error) {
+	listResp := &s3.ListObjectsV2Output{}
+	var err error
+
+	// isBucket check to make sure the bucket exists before proceeding.
```
iamhopaul123: same for this

internal/pkg/aws/s3/s3.go
```diff
@@ -206,6 +208,92 @@ func (s *S3) isBucketExists(bucket string) (bool, error) {
 	return true, nil
 }
 
+// GetBucketTree retrieves the objects in an S3 bucket and creates an ASCII tree representing their folder structure.
+func (s *S3) GetBucketTree(bucket string) (string, error) {
+	listResp := &s3.ListObjectsV2Output{}
+	var err error
```
iamhopaul123: nit: we can declare these later

internal/pkg/aws/s3/s3.go
```diff
@@ -206,6 +208,92 @@ func (s *S3) isBucketExists(bucket string) (bool, error) {
...
+		return "", nil
+	}
+
+	for {
+		listParams := &s3.ListObjectsV2Input{
+			Bucket:            aws.String(bucket),
+			Delimiter:         aws.String("/"),
+			ContinuationToken: listResp.NextContinuationToken,
+		}
+		listResp, err = s.s3Client.ListObjectsV2(listParams)
```
iamhopaul123, line 231: Do we not yield the result for each `listResp`? Seems like we only use `listResp` from the last round of loop. Also we might want to add a unit test to catch this case.

internal/pkg/aws/s3/s3.go
```diff
@@ -206,6 +208,92 @@ func (s *S3) isBucketExists(bucket string) (bool, error) {
...
+	for _, prefix := range prefixes {
+		branch := tree.AddBranch(s.extractNameFromPath(aws.StringValue(prefix.Prefix)))
+		for {
+			listParams := &s3.ListObjectsV2Input{
+				Bucket:            aws.String(bucket),
+				Delimiter:         aws.String("/"),
+				ContinuationToken: listResp.ContinuationToken,
+				Prefix:            prefix.Prefix,
+			}
+			listResp, err = s.s3Client.ListObjectsV2(listParams)
```
iamhopaul123, line 272: same here

internal/pkg/aws/s3/s3.go
```diff
@@ -206,6 +208,92 @@ func (s *S3) isBucketExists(bucket string) (bool, error) {
...
+		return "", fmt.Errorf("unable to determine the existence of bucket %s: %w", bucket, err)
+	}
+	if !exists {
+		return "", nil
+	}
+
+	for {
+		listParams := &s3.ListObjectsV2Input{
+			Bucket:            aws.String(bucket),
+			Delimiter:         aws.String("/"),
```
iamhopaul123: can we make delimiter "/" a constant so that we can use it all the occurrences in this file

internal/pkg/aws/s3/s3.go
```diff
@@ -206,6 +208,92 @@ func (s *S3) isBucketExists(bucket string) (bool, error) {
...
+		if err := s.addNodes(branch, listResp.CommonPrefixes, bucket); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// extractNameFromPath returns the last element of a path, whether the name of
+// an individual file or the end of an S3 prefix (folder name).
+func (s *S3) extractNameFromPath(path string) string {
```
iamhopaul123: Is it equivalent to https://pkg.go.dev/path/filepath#Base

iamhopaul123 (commented):
LGTM overall especially the example included in the description!

iamhopaul123 (commented):
LGTM overall especially the example included in the description!

internal/pkg/aws/s3/s3.go
```diff
@@ -190,8 +194,7 @@ func FormatARN(partition, location string) string {
 	return fmt.Sprintf("arn:%s:s3:::%s", partition, location)
 }
 
-// Check whether the bucket exists before proceeding with empty the bucket
-func (s *S3) isBucketExists(bucket string) (bool, error) {
+func (s *S3) isBucket(bucket string) (bool, error) {
```
Lou1415926: nit: what about ⬇️ 

```suggestion
func (s *S3) bucketExists(bucket string) (bool, error) {
```

internal/pkg/aws/s3/s3.go
```diff
@@ -206,6 +209,78 @@ func (s *S3) isBucketExists(bucket string) (bool, error) {
 	return true, nil
 }
 
+// GetBucketTree retrieves the objects in an S3 bucket and creates an ASCII tree representing their folder structure.
+func (s *S3) GetBucketTree(bucket string) (string, error) {
+	exists, err := s.isBucket(bucket)
+	if err != nil {
+		return "", fmt.Errorf("unable to determine the existence of bucket %s: %w", bucket, err)
```
Lou1415926: nit: probably don't need to wrap since `isBucket` already wraps it!
```suggestion
		return "", err
```

internal/pkg/aws/s3/s3.go
```diff
@@ -206,6 +209,78 @@ func (s *S3) isBucketExists(bucket string) (bool, error) {
...
+		tree.AddNode(aws.StringValue(object.Key))
+	}
+	// Recursively add folders and their children.
+	if err := s.addNodes(tree, listResp.CommonPrefixes, bucket); err != nil {
+		return "", err
+	}
+	return tree.String(), nil
+}
+
+func (s *S3) addNodes(tree treeprint.Tree, prefixes []*s3.CommonPrefix, bucket string) error {
```
Lou1415926, line 254: nit: what do you think of passing a pointer? Probably could make the function calls slightly less heavy
```suggestion
func (s *S3) addNodes(tree *treeprint.Tree, prefixes []*s3.CommonPrefix, bucket string) error {
```

Edit: scratch that, `Tree` is an interface already!

internal/pkg/aws/s3/s3.go
```diff
@@ -206,6 +209,78 @@ func (s *S3) isBucketExists(bucket string) (bool, error) {
...
+		}
+	}
+
+	tree := treeprint.New()
+	// Add top-level files.
+	for _, object := range listResp.Contents {
+		tree.AddNode(aws.StringValue(object.Key))
+	}
+	// Recursively add folders and their children.
+	if err := s.addNodes(tree, listResp.CommonPrefixes, bucket); err != nil {
```
Lou1415926: Just out of curiosity, could we end up adding nodes repeatedly like this? And if we could, would it be a concern?

For example, can this happen:
```go
listResp1 := &s3.ListObjectsV2Output{
  Contents: []Object{
      {Key: "/static/a/stuff1.go"},
      {Key: "/static/a/stuff2.go"},
  },
  CommonPrefixes: []CommonPrefix{{Prefix: "/static/a"}},
}

// tree.AddNode("/static/a/stuff1.go")
// tree.AddNode("/static/a/stuff2.go")

listResp1 := &s3.ListObjectsV2Output{
  Prefix: "static/a"
  Contents: []Object{
      {Key: "/static/a/stuff1.go"},
      {Key: "/static/a/stuff2.go"},
  },
}

// tree.AddNode("/static/a/stuff1.go")
// tree.AddNode("/static/a/stuff2.go")
```

huanjani: No, luckily it can't happen-- in a given response, the contents (and new nodes) are only the "files" at that level, so 
```
Contents: []Object{
 {Key: "/static/a/stuff1.go"},
 {Key: "/static/a/stuff2.go"},
}
```
would never be returned in the same call as   
```
CommonPrefixes: []CommonPrefix{{Prefix: "/static/a"}},
```
I also tested the double-folder scenario (see the bottom branch of the tree:
```
.
├── ReadMe.md
├── error.html
├── index.html
├── Images
│   ├── AWS-free.PNG
│   ├── S3WebsiteHosting-Architecture.PNG
│   ├── SampleWebsite.PNG
│   ├── Step1-CreateBucket-B.PNG
│   ├── Step1-CreateBucket-C.PNG
│   ├── Step1-CreateBucket-D.PNG
│   ├── Step1-CreateBucket-E.PNG
│   ├── Step1-CreateBucket.PNG
│   ├── Step2-UploadContents.PNG
│   ├── Step3-BucketPolicy-A.PNG
│   ├── Step3-BucketPolicy-B.PNG
│   ├── Step3-BucketPolicy.PNG
│   ├── Step4-A.PNG
│   └── Step4-B.PNG
├── css
│   ├── Style.css
│   ├── all.min.css
│   └── bootstrap.min.css
├── images
│   ├── bg-masthead.jpg
│   └── new
│       └── hi.html
└── superDir
    └── subDir
        ├── first
        └── second
```
For that, the first response was:
```
{
  CommonPrefixes: [
    {
      Prefix: "Images/"
    },
    {
      Prefix: "css/"
    },
    {
      Prefix: "images/"
    },
    {
      Prefix: "superDir/"
    }
  ],
  Contents: [{
      ETag: "\"a79797e25657d20790c1378583563505\"",
      Key: "ReadMe.md",
      LastModified: 2023-06-08 16:51:06 +0000 UTC,
      Size: 5615,
      StorageClass: "STANDARD"
    },{
      ETag: "\"5c468e1e8557522fe179f10abee49baf\"",
      Key: "error.html",
      LastModified: 2023-06-08 16:51:06 +0000 UTC,
      Size: 1506,
      StorageClass: "STANDARD"
    },{
      ETag: "\"9b5b13eb525455876d1979662b0b3fb4\"",
      Key: "index.html",
      LastModified: 2023-06-08 16:51:06 +0000 UTC,
      Size: 3776,
      StorageClass: "STANDARD"
    }],
  Delimiter: "/",
  IsTruncated: false,
  KeyCount: 7,
  MaxKeys: 1000,
  Name: "bugbash-static-test-testing-tree-bucket-xkntyqr45d5g",
  Prefix: ""
}
```
then for the `superDir` prefix call:
```
{
  CommonPrefixes: [{
      Prefix: "superDir/subDir/"
    }],
  Delimiter: "/",
  IsTruncated: false,
  KeyCount: 1,
  MaxKeys: 1000,
  Name: "bugbash-static-test-testing-tree-bucket-xkntyqr45d5g",
  Prefix: "superDir/"
}
```
then for the `superDir/subDir` prefix call:
```
{
  Contents: [{
      ETag: "\"d41d8cd98f00b204e9800998ecf8427e\"",
      Key: "superDir/subDir/first",
      LastModified: 2023-06-08 16:51:06 +0000 UTC,
      Size: 0,
      StorageClass: "STANDARD"
    },{
      ETag: "\"d41d8cd98f00b204e9800998ecf8427e\"",
      Key: "superDir/subDir/second",
      LastModified: 2023-06-08 16:51:06 +0000 UTC,
      Size: 0,
      StorageClass: "STANDARD"
    }],
  Delimiter: "/",
  IsTruncated: false,
  KeyCount: 2,
  MaxKeys: 1000,
  Name: "bugbash-static-test-testing-tree-bucket-xkntyqr45d5g",
  Prefix: "superDir/subDir/"
}
```

internal/pkg/aws/s3/s3.go
huanjani: discussed offline

iamhopaul123 (approved):
Awesome!

internal/pkg/aws/s3/s3.go
```diff
@@ -206,6 +209,87 @@ func (s *S3) isBucketExists(bucket string) (bool, error) {
...
+
+func (s *S3) addNodes(tree treeprint.Tree, prefixes []*s3.CommonPrefix, bucket string) error {
+	if len(prefixes) == 0 {
+		return nil
+	}
+
+	listResp := &s3.ListObjectsV2Output{}
+	var err error
+	for _, prefix := range prefixes {
+		var respContents []*s3.Object
```
Lou1415926, line 262: looks like we can look to parallelize these calls in the future!

internal/pkg/describe/static_site.go
```diff
@@ -30,16 +33,24 @@ type StaticSiteDescriber struct {
 	store                  DeployedEnvServicesLister
 	initWkldStackDescriber func(string) (workloadDescriber, error)
 	wkldDescribers         map[string]workloadDescriber
+	awsS3Client            bucketDescriber
+	s3Client               bucketNameGetter
 }
 
 // NewStaticSiteDescriber instantiates a static site service describer.
 func NewStaticSiteDescriber(opt NewServiceConfig) (*StaticSiteDescriber, error) {
+	defaultSess, err := sessions.ImmutableProvider().Default()
```
Lou1415926: ooo does env manager session not have enough permission to access the bucket?

internal/pkg/describe/static_site.go
```diff
@@ -172,6 +207,15 @@ func (w *staticSiteDesc) HumanString() string {
 			fmt.Fprintf(writer, "  %s\t%s\n", route.Environment, route.URL)
 		}
 	}
+	if len(w.Objects) != 0 {
+		fmt.Fprint(writer, color.Bold.Sprint("\nS3 Bucket Objects\n"))
+		writer.Flush()
+		for _, object := range w.Objects {
+			fmt.Fprintf(writer, "\n  %s\t%s\n", "Environment", object.Environment)
+			fmt.Fprintf(writer, object.Tree)
+			writer.Flush()
```
Lou1415926: It looks like there is no tab to align, so probably we can flush once after the `for _, object := range w.Objects` loop?

internal/pkg/describe/static_site.go
```diff
@@ -97,7 +109,22 @@ func (d *StaticSiteDescriber) Describe() (HumanJSONStringer, error) {
...
 		}
+		bucketName, err := d.s3Client.BucketName(d.app, env, d.svc)
+		if err != nil {
+			return nil, fmt.Errorf("get bucket name for %q env: %w", env, err)
+		}
+		tree, err := d.awsS3Client.GetBucketTree(bucketName)
+		if err != nil {
+			return nil, fmt.Errorf("get tree representation of bucket contents: %w", err)
+		}
+		if tree != "" {
```
Lou1415926, line 128: 💭  I wonder if we should show 
```
Environment test
   <the tree>
Environment prod
  There is no file in this environment.
```
instead of 
```
Environment test
   <the tree>
```

Lou1415926 (commented):
Looks great!! I just have two more questions 🥺 mostly the env manager role session and whether we should let the users know that there is no file in an env instead of not showing the env at all!

internal/pkg/describe/static_site.go
huanjani: We have a manifest validation preventing users from deploying to an env if there are no files to upload. So if we message anything, it should be that the svc isn't deployed to the env, not that there are no files in the env. But maybe mentioning the env at all is kind of misleading.

Lou1415926: i see - that makes sense to me!

huanjani: Yeah, I think especially bc the env wouldn't be mentioned in the command output at all if it's not deployed to that env, we should just leave this as is. Let me know if you disagree!

Lou1415926: yeah I think I agree that we should leave it as is in this case 👀  I thought user could deploy a static site with 0 files, hence the proposal. Since this is not true, the current design is better!

internal/pkg/describe/static_site.go
huanjani: Hmm. How would this work, with the describer not being env-specific?

Lou1415926: How do we think about initialize one s3 client per env, instead of using one client for all envs?
```go
type StaticSiteDescriber struct {
   initS3Client func(string) (bucketDescriber, bucketNameGetter, err)
}

func NewStaticSiteDescriber(opt NewServiceConfig) (*StaticSiteDescriber, error) {
        describer := &StaticSiteDescriber{/*...*/}
	describer.initS3Client = func(env string) (bucketDescriber, bucketNameGetter, err) {
		environment, err := opt.configStore.GetEnvironment(opt.app, env)
	        if err != nil {
		        return nil, fmt.Errorf("get environment %s: %w", env, err)
	        }
	        sess, err := sessions.ImmutableProvider().FromRole(environment.ManagerRoleARN, environment.Region)
	        if err != nil {
		        return nil, err
	        }
                return awsS3.New(defaultSess), s3.New(sess), nil
	}
	return describer, nil
}

func (d *StaticSiteDescriber) Describe() (HumanJSONStringer, error) {
  for _, env := range environments {
     bucketDescriber, bucketNameDescriber, err := d.initS3Client(env)
     //...
  }
}
```

This is similar to how we initialize `wkldDescribers` today!

This way, the identities that are reading the s3 buckets are the corresponding env manager roles - which we know have the correct permissions. The "default" profile just has to be able to assume the env manager role.

Whereas if we use the default session, this requires that the "default" profile has to be able to have permissions like `GetBucketObjects` in the app account, and Copilot doesn't control whether "default" has this permission.

I feel like ideally Copilot should use the `envManagerRole` for any operations that involve reading / deploying to an env! The "default" profile should be able to just assume the role and leave everything to `envManagerRole` for that point.

internal/pkg/describe/service.go
```diff
@@ -127,9 +128,10 @@ type appRunnerServiceDescriber struct {
 
 // NewServiceConfig contains fields that initiates service describer struct.
 type NewServiceConfig struct {
-	App         string
-	Svc         string
-	ConfigStore ConfigStoreSvc
+	App             string
+	Svc             string
+	ConfigStore     ConfigStoreSvc
+	SessionProvider *sessions.Provider
```
Lou1415926: nit: just wanted to make sure if this field was used 👁️ !

Lou1415926 (approved):
Looks great! Approving with a label but feel free to remove it whenever you're raedy!

Issue #4767, opened 2023-04-17
Support exposing UDP ports when using Network Load Balancer


dannyrandall:
From https://github.com/aws/copilot-cli/issues/1783#issuecomment-1504086954.

Copilot currently supports exposing TCP ports through an NLB, and it would be nice to support UDP ports. Additionally, it would be nice to specify a range of UDP ports to expose through the NLB instead of listing a range one-by-one.

h5aaimtron:
My use cases is specific to a voice chat application (WebRTC) that requires a range of UDP ports to be opened to facilitate the SFU (Selective Forwarding Unit) running. I could see this being true beyond my specific use-case as other chat systems, games, etc. could utilize port ranges.

tjhorner:
I am also interested in this functionality, so I went ahead and added `udp` to the list of valid NLB protocols in [this file](https://github.com/aws/copilot-cli/blob/a1c8d330ec7518a019ad92be079b63f02474833b/internal/pkg/manifest/validate.go#L58) and passed `port: 80/udp` to `nlb` in my service manifest.

The deployment appears to have worked:

<img width="273" alt="image" src="https://github.com/aws/copilot-cli/assets/2646487/886b89cf-6a85-4d5f-b01d-601f5997c1f8">

However, I haven't written any tests to determine if there are side effects I should be aware of (i.e., health check fail?). But it does work in this limited scenario.

dannyrandall:
@tjhorner thanks for that!☺️  I think you're pretty much right that the health check is the only thing missing - We'd love a PR if you're interested! I think the only other things you'll have to update are:
- Manifest configuration changes:
  - Change `NLBHealthCheckArgs.Port` to a string https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/manifest/svc.go#L378-L389
  - Update the validation logic for `NLBHealthCheckArgs.Port` to call `validateNLBPort()` for the `Port` field https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/manifest/validate.go#L907-L912
- CFN Template data changes: 
  - Add a `Protocol` field to `NLBHealthCheck` https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/workload.go#L418-L425
[...]

PR #4980, opened 2023-06-13
feat: Support UDP listeners in Network Load Balancer

tjhorner:
This PR adds basic UDP support for Network Load Balancers. For example, you can now do this in a `Load Balanced Web Service`:

```yaml
nlb:
  port: 1812/udp
  # Run the healthcheck against a different TCP port
  healthcheck:
    port: 8080
```

Additional listeners are also supported:

```yaml
nlb:
  # [...]
  additional_listeners:
    - port: 1912/udp
    # healthcheck omitted
```

I also updated the English documentation to reflect this change, but not the Japanese documentation.

This addresses issue #4767.


site/content/docs/manifest/lb-web-service.en.md
```diff
@@ -241,7 +241,7 @@ List of all available properties for a `'Load Balanced Web Service'` manifest. T
 
         nlb:
           port: 8080/tcp              # Traffic on port 8080/tcp is forwarded to the main container, on port 8080.
-          additional_rules:  
+          additional_listeners:
```
dannyrandall: good catch!

internal/pkg/manifest/svc.go
```diff
@@ -513,19 +513,28 @@ func (cfg NetworkLoadBalancerListener) exposedPorts(exposedPorts []ExposedPort,
 	if cfg.TargetPort != nil {
 		targetPort = uint16(aws.IntValue(cfg.TargetPort))
 	}
+	targetProtocol := strings.ToLower(TCP)
+	if nlbProtocol != nil {
+		protocol := strings.ToLower(aws.StringValue(nlbProtocol))
+		// Expose TCP port for TLS listeners.
+		if protocol != strings.ToLower(TLS) {
+			targetProtocol = protocol
+		}
```
Lou1415926: Yay thanks for taking care of this case!!


Small nit - `strings.EqualFold` is case-insensitive, so we can use it for the comparison here
```suggestion
		// Expose TCP port for TLS listeners.
		if !strings.EqualFold(aws.StringValue(nlbProtocol), TLS)
			targetProtocol = protocol
		}
```

internal/pkg/manifest/svc.go
```diff
@@ -513,19 +513,28 @@ func (cfg NetworkLoadBalancerListener) exposedPorts(exposedPorts []ExposedPort,
...
+	}
 	for _, exposedPort := range exposedPorts {
-		if targetPort == exposedPort.Port {
+		if targetPort == exposedPort.Port && targetProtocol == exposedPort.Protocol {
```
Lou1415926, line 525: nit:
```suggestion  
                 // NOTE: ECS doesn't support exposing the same port twice.
                 // This is a feature request to support multiple protocols on the same port: https://github.com/aws/containers-roadmap/issues/850.
		if targetPort == exposedPort.Port  {
```
ECS doesn't let you expose the same port twice, even if they are different protocols. Although as a client side tool we don't have to do the same validations as the service, I think I'd prefer erroring out early here.

NLB can listen on a port for `TPC_UDP` traffic - judging by this design pattern, it is possible that when ECS does support this feature, they will support an "additional protocol" called `TCP_UDP`, instead of letting people expose the same port twice. This is my own personal intuition though. Here is a related feature request to ECS: https://github.com/aws/containers-roadmap/issues/850. 

Anyway, based on this observation, it's likely that we won't need to change this if statement even after they'e supported the feature. So maintenance-wise, this probably won't cause burden in the future anyway.

Lou1415926 (approved):
Thank you very much for the contribution ❤️ ! This looks great, I just have some small nits!

internal/pkg/manifest/svc.go
tjhorner: TIL! Will update to use this

internal/pkg/manifest/svc.go
tjhorner: Sure, that makes sense. I'll add that note and revert the change there.

Wouldn't this just silently fail? No `err` is being returned, but I'm not sure if consumers of this method would interpret a `nil` `[]ExposedPort` as a failure.

Lou1415926: My bad!! I somehow was thinking this code was written in `manifest/validate.go`, and was trying to point out that the "uniqueness" of a port is determined by the port only, not port + protocol.

Now that I've had my coffee, I realize this comment doesn't make sense for the code here. This function is called  after we've done all the manifest validation. At this point, it should be assumed that the manifest is totally valid - that is, no port is exposed twice.

Therefore, I think we can keep it as you have (port + protocol). Since the validation already happened, there is effectively no difference between `if targetPort == exposedPort.Port` and `if targetPort == exposedPort.Port && targetProtocol == exposedPort.Protocol`; however, the latter is clearly and makes more sense 😂 

The validation code already checks that "no port is exposed twice" - it doesn't take protocol into consideration. So we are good there as well.

Sorry for the confusion!!

Lou1415926: hey there @tjhorner ! Do you want me to take over the code and address the nit, so that this PR can be merged before our next release?

tjhorner: Sure, that's fine. Thanks!

dannyrandall (approved):
awesome! thank you @tjhorner!!!

PR #4985, opened 2023-06-13
chore(cli): add `svc status` output for Static Site service type

huanjani:
Very basic output:
```
% copilot svc status          
Service: static-site
Bucket Summary

  Bucket Name     bugbash-static-test-static-site-bucket-1qdzh62y4l3gj
  Total Objects   22
  Total Size      3.7 MB
```


internal/pkg/aws/s3/s3.go
```diff
@@ -251,6 +251,40 @@ func (s *S3) GetBucketTree(bucket string) (string, error) {
 	return tree.String(), nil
 }
 
+// GetBucketSizeAndCount returns the total size and number of objects in an S3 bucket.
+func (s *S3) GetBucketSizeAndCount(bucket string) (string, int, error) {
```
iamhopaul123: The fact that this public function is specifically designed for status use case makes me think maybe we could rename this function. Otherwise it'll be too coupled with each other and not generic enough.

Also by looking at `GetBucketTree` I realized they share many very similar logics with each other (aka listing objects within the bucket). How about we just have 

```golang
func (*S3) ListObjects(bucket string) (Objects, error)

type Objects []*s3.Object

func (Objects) TotalObjectCount() int

func (Objects) TotalSize() string

func (Objects) Tree() (string, error)
```

huanjani: Hm, yeah. I hadn't reused `GetBucketTree` because of the delimiter difference, but I could pass that in. Will refactor!

huanjani: I kept total count and total size together (for now) to avoid making another API call.
I have `ListObjects` returning the output because Objects are just one element of the response that we need (we also need prefixes for tree-making).

internal/pkg/aws/s3/s3.go
```diff
@@ -194,68 +195,99 @@ func FormatARN(partition, location string) string {
 	return fmt.Sprintf("arn:%s:s3:::%s", partition, location)
 }
 
-func (s *S3) bucketExists(bucket string) (bool, error) {
-	input := &s3.HeadBucketInput{
-		Bucket: aws.String(bucket),
+// BucketTree creates an ASCII tree representing the folder structure of a bucket's objects.
+func (s *S3) BucketTree(bucket string) (string, error) {
+	outputs, err := s.listObjects(bucket, "/")
+	if err != nil {
```
dannyrandall: nit
```suggestion
	if err != nil || outputs == nil {
```

internal/pkg/aws/s3/s3.go
```diff
@@ -194,68 +195,99 @@ func FormatARN(partition, location string) string {
...
 	if err != nil {
-		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == errCodeNotFound {
-			return false, nil
+		return "", 0, err
+	}
+	if outputs == nil {
+		return "", 0, nil
+	}
+	var size int64
+	var number int
```
dannyrandall: nit
```suggestion
	var count int
```

internal/pkg/aws/s3/s3.go
```diff
@@ -194,68 +195,99 @@ func FormatARN(partition, location string) string {
...
 			ContinuationToken: listResp.NextContinuationToken,
 		}
 		listResp, err = s.s3Client.ListObjectsV2(listParams)
 		if err != nil {
-			return "", fmt.Errorf("list objects for bucket %s: %w", bucket, err)
+			return nil, fmt.Errorf("list objects for bucket %s: %w", bucket, err)
 		}
-		contents = append(contents, listResp.Contents...)
-		prefixes = append(prefixes, listResp.CommonPrefixes...)
+		outputs = append(outputs, *listResp)
```
dannyrandall, line 256: i wonder if it makes sense to merge the output into a single `s3.ListObjectsV2Output` instead? then the other functions only have to loop once?

internal/pkg/describe/status_describe.go
```diff
@@ -211,6 +242,48 @@ func (s *ecsStatusDescriber) Describe() (HumanJSONStringer, error) {
 	}, nil
 }
 
+// Describe returns the status of an AppRunner service.
+func (a *appRunnerStatusDescriber) Describe() (HumanJSONStringer, error) {
+	svc, err := a.svcDescriber.Service()
+	if err != nil {
+		return nil, fmt.Errorf("get AppRunner service description for App Runner service %s in environment %s: %w", a.svc, a.env, err)
```
dannyrandall:
```suggestion
		return nil, fmt.Errorf("get App Runner service description for App Runner service %s in environment %s: %w", a.svc, a.env, err)
```

internal/pkg/describe/status_describe.go
```diff
@@ -211,6 +242,48 @@ func (s *ecsStatusDescriber) Describe() (HumanJSONStringer, error) {
...
+	}
+	return &appRunnerServiceStatus{
+		Service:   *svc,
+		LogEvents: logEventsOutput.Events,
+	}, nil
+}
+
+// Describe returns the status of a Static Site service.
+func (d *staticSiteStatusDescriber) Describe() (HumanJSONStringer, error) {
+	bucketDataGetter, bucketNameGetter, err := d.initS3Client(d.env)
```
dannyrandall: nit
```suggestion
	dataGetter, nameGetter, err := d.initS3Client(d.env)
```

dannyrandall (approved):
just a few small nits and a question!

internal/pkg/aws/s3/s3.go
huanjani: That would be nice, but we have to keep the prefixes and contents separate because in the recursive call we use each prefix as a delimiter.

internal/pkg/cli/svc_status.go
```diff
@@ -74,7 +74,18 @@ func newSvcStatusOpts(vars svcStatusVars) (*svcStatusOpts, error) {
 					ConfigStore: configStore,
 				})
 				if err != nil {
-					return fmt.Errorf("creating status describer for apprunner service %s in application %s: %w", o.svcName, o.appName, err)
+					return fmt.Errorf("create status describer for App Runner service %s in application %s: %w", o.svcName, o.appName, err)
+				}
+				o.statusDescriber = d
+			} else if wkld.Type == manifestinfo.StaticSiteType {
```
iamhopaul123: Maybe consider using a switch here?

iamhopaul123 (approved):
:shipit:

internal/pkg/aws/s3/s3.go
```diff
@@ -194,68 +195,96 @@ func FormatARN(partition, location string) string {
...
+	}
+	if outputs == nil {
+		return "", 0, nil
+	}
+	var size int64
+	var count int
+	for _, output := range outputs {
+		for _, object := range output.Contents {
+			size = size + aws.Int64Value(object.Size)
+			count = count + 1
```
dannyrandall: nits
```suggestion
			size += aws.Int64Value(object.Size)
			count++
```

internal/pkg/aws/s3/s3.go
```diff
@@ -194,68 +195,96 @@ func FormatARN(partition, location string) string {
...
+	if err := s.addNodes(tree, prefixes, bucket); err != nil {
+		return "", err
+	}
+	return tree.String(), nil
+}
+
+// BucketSizeAndCount returns the total size and number of objects in an S3 bucket.
+func (s *S3) BucketSizeAndCount(bucket string) (string, int, error) {
+	outputs, err := s.listObjects(bucket, "")
 	if err != nil {
```
dannyrandall:
```suggestion
	if err != nil || outputs == nil {
```

internal/pkg/aws/s3/s3.go
```diff
@@ -194,68 +195,96 @@ func FormatARN(partition, location string) string {
...
-
-	return true, nil
+	return humanize.Bytes(uint64(size)), count, nil
 }
 
-// GetBucketTree retrieves the objects in an S3 bucket and creates an ASCII tree representing their folder structure.
-func (s *S3) GetBucketTree(bucket string) (string, error) {
+func (s *S3) listObjects(bucket, delimiter string) ([]s3.ListObjectsV2Output, error) {
 	exists, err := s.bucketExists(bucket)
 	if err != nil {
```
dannyrandall:
```suggestion
	if err != nil || !exists {
```

internal/pkg/aws/s3/s3.go
```diff
@@ -194,68 +195,96 @@ func FormatARN(partition, location string) string {
...
+		Bucket: aws.String(bucket),
 	}
-	// Recursively add folders and their children.
-	if err := s.addNodes(tree, prefixes, bucket); err != nil {
-		return "", err
+	_, err := s.s3Client.HeadBucket(input)
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == errCodeNotFound {
+			return false, nil
+		}
```
dannyrandall:
```suggestion
		var aerr *awserr.Error
		if errors.As(err, &aerr) && aerr.Code() == errCodeNotFound {
			return false, nil
		}
```
using `errors.As()` makes it so that any wrapped errors are caught as well!

dannyrandall (approved):
looks great!! just a few more nits!

PR #4991, opened 2023-06-14
chore: add --diff flag to pipeline deploy

Varun359:
Currently there is no diff flag for copilot pipeline deploy command. This PR adds that flag.


internal/pkg/cli/pipeline_deploy.go
```diff
@@ -258,13 +270,71 @@ func (o *deployPipelineOpts) Execute() error {
 		PermissionsBoundary: o.app.PermissionsBoundary,
 	}
 
+	if o.showDiff {
+		tpl, err := o.pipelineStackConfig(deployPipelineInput).Template()
+		if err != nil {
+			return fmt.Errorf("stack template local: %w", err)
```
huanjani: nit: can we make this error msg more descriptive/specific? 🤔

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -258,13 +270,71 @@ func (o *deployPipelineOpts) Execute() error {
...
+		}
+	}
 	if err := o.deployPipeline(deployPipelineInput); err != nil {
 		return err
 	}
-
 	return nil
 }
 
+// PipelineDeployDiff returns the stringified diff of the template against the deployed template of the pipeline.
```
huanjani:
```suggestion
// pipelineDeployDiff returns the stringified diff of the template against the deployed template of the pipeline.
```

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -258,13 +270,71 @@ func (o *deployPipelineOpts) Execute() error {
...
+		return "", err
+	}
+	return buf.String(), nil
+}
+
+func (o *deployPipelineOpts) diff(in *deploy.CreatePipelineInput, tpl string) error {
+
+	if out, err := o.pipelineDeployDiff(in, tpl); err != nil {
+		var errHasDiff *errHasDiff
+		if !errors.As(err, &errHasDiff) {
```
huanjani: I wonder if we should do a switch here instead of the `else if`s.

internal/pkg/cli/pipeline_deploy_test.go
```diff
@@ -718,6 +722,201 @@ func TestDeployPipelineOpts_Execute(t *testing.T) {
...
+				m.deployer.EXPECT().Template(gomock.Any()).Return("name: mockEnv\ntype: Environment", nil)
+
+				m.mockDiffWriter = &strings.Builder{}
+
+				m.prompt.EXPECT().Confirm(continueDeploymentPrompt, "").Return(false, errors.New("some error"))
+
+			},
+			expectedError: fmt.Errorf("ask whether to continue with the deployment: some error"),
+		},
+		"Successfully show diff and create a new pipeline": {
```
huanjani: tiny nit:
```suggestion
		"successfully show diff and create a new pipeline": {
```

internal/pkg/cli/pipeline_deploy_test.go
```diff
@@ -718,6 +722,201 @@ func TestDeployPipelineOpts_Execute(t *testing.T) {
...
+					if in.IsLegacy {
+						return errors.New("should not be a legacy pipeline")
+					}
+					return nil
+				})
+				m.prog.EXPECT().Stop(log.Ssuccessf(fmtPipelineDeployComplete, pipelineName)).Times(1)
+
+			},
+		},
+		"Successfully show diff and redeploy an existing pipeline": {
```
huanjani, line 869: same tiny nit:
```suggestion
		"successfully show diff and redeploy an existing pipeline": {
```

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -446,13 +516,17 @@ func (o *deployPipelineOpts) deployPipeline(in *deploy.CreatePipelineInput) erro
...
 
 	// If the stack already exists - we update it
-	shouldUpdate, err := o.shouldUpdate()
-	if err != nil {
-		return err
-	}
-	if !shouldUpdate {
-		return nil
+
+	if !o.showDiff {
```
huanjani:
```suggestion
	if !o.showDiff {
```

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -258,13 +270,71 @@ func (o *deployPipelineOpts) Execute() error {
...
+	buf := strings.Builder{}
+	if err := diffTree.Write(&buf); err != nil {
+		return "", err
+	}
+	return buf.String(), nil
+}
+
+func (o *deployPipelineOpts) diff(in *deploy.CreatePipelineInput, tpl string) error {
+
+	if out, err := o.pipelineDeployDiff(in, tpl); err != nil {
```
huanjani:
```suggestion
	if out, err := o.pipelineDeployDiff(in, tpl); err != nil {
```

huanjani (commented):
Wow! 👏🏼 Great job! Nice unit tests.
Just some small nits.

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -446,13 +516,17 @@ func (o *deployPipelineOpts) deployPipeline(in *deploy.CreatePipelineInput) erro
...
-		return nil
+
+	if !o.showDiff {
+		shouldUpdate, err := o.shouldUpdate()
+		if err != nil {
+			return err
+		}
+		if !shouldUpdate {
+			return nil
+		}
```
dannyrandall, line 514: chatted with @huanjani about this offline, but i wonder if we should just remove this logic entirely (the `shouldUpdate()` stuff, which you didn't add! just adding to this pr since it came up here 😛)? it feels kinda odd to ask "do you want to deploy" after someone typed `pipeline deploy` 🤔

internal/pkg/cli/pipeline_deploy.go
Varun359: We can do that too but since there are no large number of comparisons and I am not just doing the simple value comparisons instead performing operations which are involving function calls and error handling. Wouldn't it be better if we use if/else if in this case than using switch 🤔?

internal/pkg/cli/pipeline_deploy.go
dannyrandall: would be fine to do as a separate PR if you'd prefer :)

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -62,21 +66,25 @@ type deployPipelineVars struct {
...
-	newJobListCmd                   func(io.Writer, string) cmd
+	pipelineDeployer    pipelineDeployer
+	sel                 wsPipelineSelector
+	prog                progress
+	prompt              prompter
+	region              string
+	store               store
+	ws                  wsPipelineReader
+	codestar            codestar
+	writer              io.Writer
```
dannyrandall: nit, to keep it consistent with `svc deploy --diff`:
```suggestion
	diffWriter              io.Writer
```

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -258,13 +270,70 @@ func (o *deployPipelineOpts) Execute() error {
...
+	}
 	if err := o.deployPipeline(deployPipelineInput); err != nil {
 		return err
 	}
-
 	return nil
 }
 
+// PipelineDeployDiff returns the stringified diff of the template against the deployed template of the pipeline.
+func (o *deployPipelineOpts) pipelineDeployDiff(in *deploy.CreatePipelineInput, template string) (string, error) {
```
iamhopaul123: Can we do something like env to move this to the `cli/deploy` pkg?

internal/pkg/cli/pipeline_deploy.go
iamhopaul123: Can we reuse https://github.com/aws/copilot-cli/blob/b408b6d0a833d5e9be7f560e2296e24014e63fb5/internal/pkg/cli/svc_deploy.go#L604

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -258,13 +270,70 @@ func (o *deployPipelineOpts) Execute() error {
 		PermissionsBoundary: o.app.PermissionsBoundary,
 	}
 
+	if o.showDiff {
+		tpl, err := o.pipelineStackConfig(deployPipelineInput).Template()
+		if err != nil {
+			return fmt.Errorf("generate the new template for deploy diff: %w", err)
```
dannyrandall:
```suggestion
			return fmt.Errorf("generate the new template for diff: %w", err)
```

internal/pkg/cli/pipeline_deploy.go
```diff
@@ -258,13 +270,70 @@ func (o *deployPipelineOpts) Execute() error {
...
+		return "", fmt.Errorf("parse the diff against the deployed pipeline stack %q: %w", in.Name, err)
+	}
+	buf := strings.Builder{}
+	if err := diffTree.Write(&buf); err != nil {
+		return "", err
+	}
+	return buf.String(), nil
+}
+
+func (o *deployPipelineOpts) diff(in *deploy.CreatePipelineInput, tpl string) error {
```
dannyrandall: i wonder if we could implement the `templateDiffer` interface (`DeployDiff()`) with `o` and avoid writing our own `diff()` function?

internal/pkg/cli/pipeline_deploy.go
Varun359: I can make a separate PR for this so that it makes clear that I am just making this change

internal/pkg/cli/pipeline_deploy.go
Varun359: This would need a big refactor, Probably we will have a separate refactor for this.

internal/pkg/cli/pipeline_deploy_test.go
```diff
@@ -718,6 +722,206 @@ func TestDeployPipelineOpts_Execute(t *testing.T) {
...
+					m.deployedPipelineLister.EXPECT().ListDeployedPipelines(appName).Return([]deploy.Pipeline{}, nil),
+
+					m.pipelineStackConfig.EXPECT().Template().Return("template one", nil),
+					m.deployedPipelineLister.EXPECT().ListDeployedPipelines(appName).Return([]deploy.Pipeline{}, nil),
+
+					m.deployer.EXPECT().Template(gomock.Any()).Return("", fmt.Errorf("some error")))
+			},
+			expectedError: fmt.Errorf("retrieve the deployed template for %q: some error", pipelineName),
+		},
+		"failed to get the continueDeploymentPrompt": {
```
dannyrandall:
```suggestion
		"failed prompt to accept diff": {
```

internal/pkg/cli/pipeline_deploy_test.go
```diff
@@ -718,6 +722,206 @@ func TestDeployPipelineOpts_Execute(t *testing.T) {
...
+				m.deployer.EXPECT().GetRegionalAppResources(gomock.Any()).Return(mockResources, nil)
+
+				// check if the pipeline has been deployed using a legacy naming.
+				m.deployedPipelineLister.EXPECT().ListDeployedPipelines(appName).Return([]deploy.Pipeline{}, nil)
+
+				m.pipelineStackConfig.EXPECT().Template().Return("name: mockEnv\ntype: Environment", nil)
+				m.deployedPipelineLister.EXPECT().ListDeployedPipelines(appName).Return([]deploy.Pipeline{}, nil)
+				m.deployer.EXPECT().Template(gomock.Any()).Return("name: mockEnv\ntype: Environment", nil)
+
+				m.mockDiffWriter = &strings.Builder{}
```
dannyrandall: i think you can remove this from all the tests - looks like it's already set on 949?

dannyrandall (approved):
great job!🎉🎉🎉 just a couple small things 😊

iamhopaul123 (approved):

PR #4992, opened 2023-06-14
chore: add function to delete single stackset instance

bvtujo:
- chore: add methods for removing a single stackset instance
- test: add unit tests for stackset deletion instance

Modular PR 1 of 2 for conditionally cleaning up stackset instance on env delete. 
This contains "stack" functions for working with the stackset and removing only one instance.

The next one will contain the changes to the CLI logic. 

### DIFF
--- a/internal/pkg/aws/cloudformation/stackset/stackset.go
+++ b/internal/pkg/aws/cloudformation/stackset/stackset.go
@@ -112,7 +112,7 @@ func (ss *StackSet) UpdateAndWait(name, template string, opts ...CreateOrUpdateO
 	return ss.WaitForOperation(name, id)
 }
 
-func (ss *StackSet) getStackSetUniqueAccountsAndRegions(name string) ([]InstanceSummary, error) {
+func (ss *StackSet) getInstanceSummaries(name string) ([]InstanceSummary, error) {
 	summaries, err := ss.InstanceSummaries(name)
 	if err != nil {
 		// If the stack set doesn't exist - just move on.
@@ -132,24 +132,11 @@ func (ss *StackSet) getStackSetUniqueAccountsAndRegions(name string) ([]Instance
 	return summaries, nil
 }
 
+// DeleteInstance deletes the stackset instance for the stackset with the given name in the given account
+// and region and returns the operation ID.
+// If there is no instance in the given account and region, this function will return an operation ID
+// but the API call will take no action.
 func (ss *StackSet) DeleteInstance(name, account, region string) (string, error) {
-	summaries, err := ss.getStackSetUniqueAccountsAndRegions(name)
-	if err != nil {
-		return "", err
-	}
-	accountAndRegionPresentInInstances := false
-	for _, summary := range summaries {
-		if summary.Region == region && summary.Account == account {
-			accountAndRegionPresentInInstances = true
-			break
-		}
-	}
-
-	// No-op for idempotency if the account and region is not found in the deployed instances.
-	if !accountAndRegionPresentInInstances {
-		return "", nil
-	}
-
 	out, err := ss.client.DeleteStackInstances(&cloudformation.DeleteStackInstancesInput{
 		StackSetName: aws.String(name),
 		Accounts:     aws.StringSlice([]string{account}),
@@ -168,7 +155,7 @@ func (ss *StackSet) DeleteInstance(name, account, region string) (string, error)
 // If the stack set does not have any instances, then return [ErrStackSetInstancesNotFound].
 // Both errors should satisfy [IsEmptyStackSetErr], otherwise it's an unexpected error.
 func (ss *StackSet) DeleteAllInstances(name string) (string, error) {
-	summaries, err := ss.getStackSetUniqueAccountsAndRegions(name)
+	summaries, err := ss.getInstanceSummaries(name)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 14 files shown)


internal/pkg/deploy/cloudformation/app.go
```diff
@@ -312,6 +312,51 @@ func (cf CloudFormation) RemoveJobFromApp(app *config.Application, jobName strin
...
+// RemoveEnvFromAppOpts contains the parameters to call RemoveEnvFromApp.
+type RemoveEnvFromAppOpts struct {
+	App                 *config.Application
+	EnvName             string
+	EnvAccountID        string
+	EnvRegion           string
+	DeleteStackInstance bool // If this is the last environment in a region, we can delete it.
+}
+
+func (cf CloudFormation) RemoveEnvFromApp(opts *RemoveEnvFromAppOpts) error {
```
iamhopaul123, line 388: Comment pls 🙏 If i understand correctly the only thing we need to do for `RemoveEnvFromApp` should be optionally delete the stackset instance? Also sorry this might be a stupid question but I'm not sure how we figure out this is the last environment we try to delete from this account/region...

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -312,6 +312,51 @@ func (cf CloudFormation) RemoveJobFromApp(app *config.Application, jobName strin
...
+
+	newDeploymentConfig := stack.AppResourcesConfig{
+		Version:   previouslyDeployedConfig.Version + 1,
+		Workloads: previouslyDeployedConfig.Workloads,
+		Accounts:  accountList,
+		App:       appConfig.Name,
+	}
+	if err := cf.deployAppConfig(appConfig, &newDeploymentConfig, true); err != nil {
+		return err
+	}
```
iamhopaul123: Why do we need this 🤔 I thought we wanted to remove env from app...

internal/pkg/aws/cloudformation/stackset/stackset.go
```diff
@@ -112,27 +112,66 @@ func (ss *StackSet) UpdateAndWait(name, template string, opts ...CreateOrUpdateO
 	return ss.WaitForOperation(name, id)
 }
 
-// DeleteAllInstances removes all stack instances from a stack set and returns the operation ID.
-// If the stack set does not exist, then return [ErrStackSetNotFound].
-// If the stack set does not have any instances, then return [ErrStackSetInstancesNotFound].
-// Both errors should satisfy [IsEmptyStackSetErr], otherwise it's an unexpected error.
-func (ss *StackSet) DeleteAllInstances(name string) (string, error) {
+func (ss *StackSet) getStackSetUniqueAccountsAndRegions(name string) ([]InstanceSummary, error) {
```
Lou1415926: nit: what do you think of just ⬇️ 
```suggestion
func (ss *StackSet) getInstanceSummary(name string) ([]InstanceSummary, error) {
```

internal/pkg/aws/cloudformation/stackset/stackset.go
```diff
@@ -112,27 +112,66 @@ func (ss *StackSet) UpdateAndWait(name, template string, opts ...CreateOrUpdateO
...
 	}
+	return summaries, nil
+}
+
+func (ss *StackSet) DeleteInstance(name, account, region string) (string, error) {
+	summaries, err := ss.getStackSetUniqueAccountsAndRegions(name)
+	if err != nil {
+		return "", err
+	}
+	accountAndRegionPresentInInstances := false
```
Lou1415926: nit nit nit:
```suggestion
	var accountAndRegionPresentInInstances bool
```

internal/pkg/aws/cloudformation/stackset/stackset.go
```diff
@@ -112,27 +112,66 @@ func (ss *StackSet) UpdateAndWait(name, template string, opts ...CreateOrUpdateO
...
+			break
+		}
+	}
+
+	// No-op for idempotency if the account and region is not found in the deployed instances.
+	if !accountAndRegionPresentInInstances {
+		return "", nil
+	}
+
+	out, err := ss.client.DeleteStackInstances(&cloudformation.DeleteStackInstancesInput{
```
Lou1415926, line 140: What would happen if we just call `DeleteStackInstances` without first checking the summaries?

internal/pkg/deploy/cloudformation/app.go
Lou1415926, line 388: What do you think of renaming this function to `RemoveLastEnvInAccountRegionFromApp` (crappy name but...) then we can remove the `DeleteStackInstance` from `opts` 💭

internal/pkg/aws/cloudformation/stackset/stackset.go
bvtujo: hmm it's actually idempotent 🤔 Going to just remove the check and put the stackset summary logic in the CLI package. This is an unfortunately leaky abstraction here, since we don't have copilot concepts down at the level of stackset operations.

internal/pkg/deploy/cloudformation/app.go
bvtujo: I think it's possible to have the situation where we don't need to delete a stackset instance but we do need to redeploy the app and other stackset instances to remove an account from IAM roles and DNS delegation. I don't love having a function call which might be a no-op but everything should be idempotent.

bvtujo: We can continue this discussion on the comments for #5060 if that's okay with you

internal/pkg/deploy/cloudformation/app.go
bvtujo: If this is the last environment in a given account, we need to redeploy the app roles stack and all the other stackset instances to update all the IAM roles so they no longer trust that account. We can continue this discussion in the comments of #5060 if that's okay with you.

bvtujo:
closing in favor of #5060 and #5061

PR #4993, opened 2023-06-14
chore: Interpolate `slice of strings` in the manifest

### LINKED ISSUE #4928: Support non-string shell env vars in manifest interpolation
Related: https://github.com/aws/copilot-cli/discussions/4907

Currently, only string-type shell env vars can be rendered into manifest correctly. This is because Copilot substitutes `${ENV_VAR_NAME}`s with their actual values always as `string`s https://github.com/aws/copilot-cli/blob/83f09c7decb89560667f171e278aa57d72a05853/internal/pkg/manifest/interpolate.go#L75
without switching the node `Kind` to the actual kind (e.g. `!!seq`). Because `${ENV_VAR_NAME}` is always recognized as a `string` by the default YAML unmarshaler, the unmarshal fails for fields that are not expecting a string.

KollaAdithya:
closes #4928


iamhopaul123 (commented):
We might need to think about more scenarios: right now it works for both string and list. However, it won't work for something like
```yaml
image:
  build: api/Dockerfile
  port: 8080
  healthcheck: ${HEALTHCHECK}
```
where
```
HEALTHCHECK={command: ["CMD-SHELL", "curl -f http://localhost:8080/_healthcheck || exit 1"]}
```

Maybe instead of doing the unmarshaling and operate in the yaml node level, we could just do string substitution?

KollaAdithya:
> Maybe instead of doing the unmarshaling and operate in the yaml node level, we could just do string substitution?

This is good option as well. we can just replace all the occurences of the `${ }` in the manifest.
But only the downside is they can now interpolate the manifest fields as well.
```${security_group_field_name}: [sg1,sg2,sg3]```.  But there is no reason why they would do manifest field interpolation.

**Comment by iamhopaul123** [comment]
> ${security_group_field_name}: [sg1,2,3]. But there is no reason why they would do manifest field interpolation.

Yeah I guess that's true...Open to other thoughts too.

**Comment by dannyrandall** [comment]
maybe i'm misunderstanding something....why don't we just interoplate at the `ReadWorkloadManifest()` level? https://github.com/aws/copilot-cli/blob/b408b6d0a833d5e9be7f560e2296e24014e63fb5/internal/pkg/workspace/workspace.go#L350-L360

So essentially the flow would be `read manifest bytes -> interpolate on manifest string -> yaml.Unmarshal`

**Comment by KollaAdithya** [comment]
After a offline discussion agreed on that we support only `list` and `string` interpolation for now. If any more use cases come up for the `whole manifest substitution` or `map` interpolation. we will implement that 

```yaml
# "svc" manifest
${my_manifest}
```

internal/pkg/manifest/interpolate.go
```diff
@@ -72,7 +72,21 @@ func (i *Interpolator) applyInterpolation(node *yaml.Node) error {
 		if err != nil {
 			return err
 		}
-		node.Value = interpolated
+		if strings.HasPrefix(interpolated, "[") && strings.HasSuffix(interpolated, "]") {
```
Lou1415926: I wonder if this is more robust:
```suggestion
                var s []string
                err := json.Unmarshal([]byte(interpolated), &s)
                if err != nil && len(s) != 0 {
                     // `interpolated` is a slice, do the node type conversion
                }
```

iamhopaul123 (commented):
Looks good to me overall. Can we update the doc for this as well? Just a callout this seems to only work for "slice of string" instead of general "slice of anything", which seems good to me.

bvtujo (approved):
slick! Looks good to me. :shipit:

Lou1415926 (approved):

PR #5021, opened 2023-06-21
chore: use universal template version for app and add allow-downgrade flag to env init

iamhopaul123:

### DIFF
--- a/internal/pkg/cli/app_upgrade.go
+++ b/internal/pkg/cli/app_upgrade.go
@@ -21,6 +21,7 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/term/log"
 	"github.com/aws/copilot-cli/internal/pkg/term/prompt"
 	"github.com/aws/copilot-cli/internal/pkg/term/selector"
+	"github.com/aws/copilot-cli/internal/pkg/version"
 	"github.com/spf13/cobra"
 	"golang.org/x/mod/semver"
 )
@@ -103,26 +104,26 @@ func (o *appUpgradeOpts) Execute() error {
 		return err
 	}
 
-	version, err := vg.Version()
+	appVersion, err := vg.Version()
 	if err != nil {
 		return fmt.Errorf("get template version of application %s: %v", o.name, err)
 	}
-	if !shouldUpgradeApp(o.name, version) {
+	if !shouldUpgradeApp(o.name, appVersion) {
 		return nil
 	}
 	app, err := o.store.GetApplication(o.name)
 	if err != nil {
 		return fmt.Errorf("get application %s: %w", o.name, err)
 	}
-	log.Infof(fmtAppUpgradeStart, color.HighlightUserInput(o.name), color.Emphasize(version), color.Emphasize(deploy.LatestAppTemplateVersion))
+	log.Infof(fmtAppUpgradeStart, color.HighlightUserInput(o.name), color.Emphasize(appVersion), color.Emphasize(version.LatestTemplateVersion()))
 	defer func() {
 		if err != nil {
-			log.Errorf(fmtAppUpgradeFailed, color.HighlightUserInput(o.name), color.Emphasize(deploy.LatestAppTemplateVersion))
+			log.Errorf(fmtAppUpgradeFailed, color.HighlightUserInput(o.name), color.Emphasize(version.LatestTemplateVersion()))
 			return
 		}
-		log.Successf(fmtAppUpgradeComplete, color.HighlightUserInput(o.name), color.Emphasize(deploy.LatestAppTemplateVersion))
+		log.Successf(fmtAppUpgradeComplete, color.HighlightUserInput(o.name), color.Emphasize(version.LatestTemplateVersion()))
 	}()
-	err = o.upgradeApplication(app, version, deploy.LatestAppTemplateVersion)
+	err = o.upgradeApplication(app, appVersion, version.LatestTemplateVersion())
 	if err != nil {
 		return err
 	}
@@ -141,20 +142,20 @@ func (o *appUpgradeOpts) askName() error {
 	return nil
 }
 
-func shouldUpgradeApp(appName string, version string) bool {
-	diff := semver.Compare(version, deploy.LatestAppTemplateVersion)
+func shouldUpgradeApp(appName string, appVersion string) bool {
+	diff := semver.Compare(appVersion, version.LatestTemplateVersion())
 	if diff < 0 {
 		// Newer version available.
 		return true
 	}
 
-	msg := fmt.Sprintf("Application %s is already on the latest version %s, skip upgrade.", appName, deploy.LatestAppTemplateVersion)
+	msg := fmt.Sprintf("Application %s is already on the latest version %s, skip upgrade.", appName, version.LatestTemplateVersion())
 	if diff > 0 {
 		// It's possible that a teammate used a different version of the CLI to upgrade the application
 		// to a newer version. And the current user is on an older version of the CLI.
 		// In this situation we notify them they should update the CLI.
 		msg = fmt.Sprintf(`Skip upgrading application %s to version %s since it's on version %s. 
-Are you using the latest version of AWS Copilot?`, appName, deploy.LatestAppTemplateVersion, version)
+Are you using the latest version of AWS Copilot?`, appName, version.LatestTemplateVersion(), appVersion)
 	}
 	log.Debugln(msg)
 	return false
--- a/internal/pkg/cli/app_upgrade.go
+++ b/internal/pkg/cli/app_upgrade.go
@@ -52,6 +52,8 @@ type appUpgradeOpts struct {
 	upgrader appUpgrader
 
 	newVersionGetter func(string) (versionGetter, error)
+
+	templateVersion string // Overridden in tests.
 }
 
 func newAppUpgradeOpts(vars appUpgradeVars) (*appUpgradeOpts, error) {
@@ -74,6 +76,7 @@ func newAppUpgradeOpts(vars appUpgradeVars) (*appUpgradeOpts, error) {
 			}
 			return d, nil
 		},
+		templateVersion: version.LatestTemplateVersion(),
 	}, nil
 }
 
@@ -108,22 +111,22 @@ func (o *appUpgradeOpts) Execute() error {
 	if err != nil {
 		return fmt.Errorf("get template version of application %s: %v", o.name, err)
 	}
-	if !shouldUpgradeApp(o.name, appVersion) {
+	if !o.shouldUpgradeApp(appVersion) {
 		return nil
 	}
 	app, err := o.store.GetApplication(o.name)
 	if err != nil {
 		return fmt.Errorf("get application %s: %w", o.name, err)
 	}
-	log.Infof(fmtAppUpgradeStart, color.HighlightUserInput(o.name), color.Emphasize(appVersion), color.Emphasize(version.LatestTemplateVersion()))
+	log.Infof(fmtAppUpgradeStart, color.HighlightUserInput(o.name), color.Emphasize(appVersion), color.Emphasize(o.templateVersion))
 	defer func() {
 		if err != nil {
-			log.Errorf(fmtAppUpgradeFailed, color.HighlightUserInput(o.name), color.Emphasize(version.LatestTemplateVersion()))
+			log.Errorf(fmtAppUpgradeFailed, color.HighlightUserInput(o.name), color.Emphasize(o.templateVersion))
 			return
 		}
-		log.Successf(fmtAppUpgradeComplete, color.HighlightUserInput(o.name), color.Emphasize(version.LatestTemplateVersion()))
+		log.Successf(fmtAppUpgradeComplete, color.HighlightUserInput(o.name), color.Emphasize(o.templateVersion))
 	}()
-	err = o.upgradeApplication(app, appVersion, version.LatestTemplateVersion())
+	err = o.upgradeApplication(app, appVersion, o.templateVersion)
 	if err != nil {
 		return err
 	}
@@ -142,20 +145,20 @@ func (o *appUpgradeOpts) askName() error {
 	return nil
 }
 
-func shouldUpgradeApp(appName string, appVersion string) bool {
-	diff := semver.Compare(appVersion, version.LatestTemplateVersion())
+func (o *appUpgradeOpts) shouldUpgradeApp(appVersion string) bool {
+	diff := semver.Compare(appVersion, o.templateVersion)
 	if diff < 0 {
 		// Newer version available.
 		return true
 	}
 
-	msg := fmt.Sprintf("Application %s is already on the latest version %s, skip upgrade.", appName, version.LatestTemplateVersion())
+	msg := fmt.Sprintf("Application %s is already on the latest version %s, skip upgrade.", o.name, o.templateVersion)
 	if diff > 0 {
 		// It's possible that a teammate used a different version of the CLI to upgrade the application
 		// to a newer version. And the current user is on an older version of the CLI.
 		// In this situation we notify them they should update the CLI.
 		msg = fmt.Sprintf(`Skip upgrading application %s to version %s since it's on version %s. 
-Are you using the latest version of AWS Copilot?`, appName, version.LatestTemplateVersion(), appVersion)
+Are you using the latest version of AWS Copilot?`, o.name, o.templateVersion, appVersion)
 	}
 	log.Debugln(msg)
 	return false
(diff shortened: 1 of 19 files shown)


internal/pkg/version/version.go
```diff
@@ -4,5 +4,25 @@
 // Package version holds variables for generating version information
 package version
 
+const (
+	// LegacyAppTemplateVersion is the version associated with the application template before we started versioning.
+	LegacyAppTemplateVersion = "v0.0.0"
+	// AliasLeastAppTemplateVersion is the least version number available for HTTPS alias.
+	AliasLeastAppTemplateVersion = "v1.0.0"
+	// StaticSiteMinAppTemplateVersion is the minimum app version required to deploy a static site.
+	StaticSiteMinAppTemplateVersion = "v1.2.0"
```
Lou1415926: nit: to have constant names... maybe. Just trying to imagine what would I type when I need to use this variable while coding 💭 I might just type "version.App" or "version.MinV" and hope the IDE can give me the correct suggestions

```suggestion
         AppTemplateMinVersionAlias = "v1.0.0"
	// StaticSiteMinAppTemplateVersion is the minimum app version required to deploy a static site.
	AppTemplateMinVersionStaticSite = "v1.2.0"
```

internal/pkg/version/version.go
```diff
@@ -4,5 +4,25 @@
...
+
+	defaultTemplateVersion = "v1.29.0" // v1.29.0 is when we use Copilot version to version templates.
```
Lou1415926: Do we need to update this value after every release to pass the tests?

internal/pkg/describe/app.go
```diff
@@ -51,8 +51,8 @@ func (a *App) HumanString() string {
 	writer.Flush()
 	fmt.Fprintf(writer, "  %s\t%s\n", "Name", a.Name)
 	availableVersion := ""
-	if deploy.LatestAppTemplateVersion != a.Version {
-		availableVersion = color.Yellow.Sprintf("(latest available: %s)", deploy.LatestAppTemplateVersion)
+	if version.LatestTemplateVersion() != a.Version {
+		availableVersion = color.Yellow.Sprintf("(latest available: %s)", version.LatestTemplateVersion())
```
Lou1415926: I could see the value of this information - we were letting the users know that "your app template is now outdated, which may be a reason why your deployment is failing. Try app upgrade".

However, it no longer has the same meaning, now that the app template version is the same as the released binary version. That is, we bump the version even if the app template has no change.

How should we know that "the app template is outdated" now? Should we have a version compatibility check (just like for env) for app as well?

internal/pkg/version/version.go
iamhopaul123: Nope just for this time. We used to be required to update some tests every time we bump the template version

internal/pkg/cli/app_upgrade.go
```diff
@@ -141,20 +142,20 @@ func (o *appUpgradeOpts) askName() error {
 	return nil
 }
 
-func shouldUpgradeApp(appName string, version string) bool {
-	diff := semver.Compare(version, deploy.LatestAppTemplateVersion)
+func shouldUpgradeApp(appName string, appVersion string) bool {
```
Lou1415926: Related to the question I had here https://github.com/aws/copilot-cli/pull/5021/files#r1239025625.

I think we should recommend running `app upgrade` only if the current app template is missing some latest features. I would expect that we return `true` here only in this case.

Basically, I think there are two types of "versioning":
1. The binary version used to deploy the app.
2. The version number on which the app template is updated.

I agree that we can use the binary version to version the app template, but we should:
1. Keep an internal record of 2
2. Let users be comfortable with having a "false-outdated" app template verion. That is, they should be comfortable when they see that the app template version is on v1.27.0, while the binary is v1.29.0. How to convey this message is the question.

internal/pkg/cli/env_init.go
```diff
@@ -257,10 +261,21 @@ func (o *initEnvOpts) Ask() error {
...
+	}
+	if diff := semver.Compare(appVersion, version.LatestTemplateVersion()); diff > 0 && !o.allowAppDowngrade {
+		return &errCannotDowngradeAppVersion{
+			appName:    o.name,
+			appVersion: appVersion,
+		}
+	}
 	app, err := o.store.GetApplication(o.appName)
 	if err != nil {
-		// Ensure the app actually exists before we do a deployment.
```
Lou1415926, line 263: nit: I think we can keep this comment but modify it to "...before we write the manifest". I feel like the value of this comment is to prevent people from moving this code block to just before L279.

internal/pkg/cli/env_init.go
```diff
@@ -257,10 +261,21 @@ func (o *initEnvOpts) Ask() error {
...
 func (o *initEnvOpts) Execute() error {
-	o.initRuntimeClients()
+	if err := o.initRuntimeClients(); err != nil {
+		return err
+	}
+	appVersion, err := o.appVersionGetter.Version()
+	if err != nil {
+		return fmt.Errorf("get template version of application %s: %w", o.appName, err)
+	}
+	if diff := semver.Compare(appVersion, version.LatestTemplateVersion()); diff > 0 && !o.allowAppDowngrade {
```
Lou1415926: Can you help me understand why we need to check the app version while running `env init`? Is it because of the stack set?

iamhopaul123: Yeah we'll update the stackset, which could cause version downgrade issue.

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -136,7 +137,7 @@ func (cf CloudFormation) DelegateDNSPermissions(app *config.Application, account
 		AccountID:          app.AccountID,
 		DomainName:         app.Domain,
 		DomainHostedZoneID: app.DomainHostedZoneID,
-		Version:            deploy.LatestAppTemplateVersion,
+		Version:            version.LatestTemplateVersion(),
```
Lou1415926, line 140: What do you think of the following structure as an alternative:
```go
// In version/version.go
var Version string
var LatestAppTemplateVersion = Version

// In internal/pkg/deploy/cloudformation/app.go
var latestAppTemplateVersion = version.LatestAppTemplateVersion
func (cf CloudFormation) DelegateDNSPermissions(app *config.Application, accountID string) error {
   deployApp := deploy.CreateAppInput{
		Name:               app.Name,
		AccountID:          app.AccountID,
		DomainName:         app.Domain,
		DomainHostedZoneID: app.DomainHostedZoneID,
		Version:            latestAppTemplateVersion,
}

// In internal/pkg/deploy/cloudformation/app.go
func TestCloudFormation_DelegateDNSPermissions(t *testing.T) {
  mockLatestAppTemplateVersion := v1.99.99
       latestAppTemplateVersion = mockLatestAppTemplateVersion
  }
}
```

I can see how this introduces more lines of code changes, and requires individual packages to implement their own way of mocking the latest app template version - so more churning for developers.

However, the benefit is that it keeps the "mocking" logic within individual packages, instead of providing a general mock from the "source" package (in this case, the "source" package is `version`)

I am fine either way, just proposing a different idea!

internal/pkg/cli/flag.go
```diff
@@ -16,21 +16,22 @@ import (
...
+	workloadFlag       = "workload"
+	svcTypeFlag        = "svc-type"
+	jobTypeFlag        = "job-type"
+	typeFlag           = "type"
+	profileFlag        = "profile"
+	yesFlag            = "yes"
+	jsonFlag           = "json"
+	allFlag            = "all"
+	forceFlag          = "force"
+	allowDowngradeFlag = "allow-downgrade"
```
Lou1415926, line 31: We have the `--yes` and `--diff-yes` flags. Perhaps we can consider `--downgrade-yes` just to be consistent!

internal/pkg/deploy/cloudformation/app.go
iamhopaul123: Good question. The reasons for having `LatestTemplateVersion()` are:

1. Previously it was a const so that it is immutable, and we might want to keep this benefit so that we don't accidentally modify the version number
2. To avoid the fact that each test function has to assign value to `LatestAppTemplateVersion` otherwise it'll be empty since the env var is not set for each test environment.

> However, the benefit is that it keeps the "mocking" logic within individual packages, instead of providing a general mock from the "source" package (in this case, the "source" package is version)

If a package needs its own mock, it can override this behavior I think. For example in CLI package instead of using `version.LatestTemplateVersion()`, it uses a cache variable in the struct and override it in the unit test.

internal/pkg/cli/flag.go
iamhopaul123: Both `--yes` and `--diff-yes` are for skipping conformation question to unblock CICD or script users. I don't think it is the same case here 🤔

internal/pkg/describe/app.go
iamhopaul123: What if we just remove the information from human string 🤔 We'll prompt them to run `app upgrade` if they really need to (for example: static site requires at least v1.2.0). I agreed having this information will lead to a misconception that their apps are outdated but might not be the case.

internal/pkg/deploy/cloudformation/app.go
iamhopaul123: Alternatively we can keep the function and don't set general default value for testing (it'll be empty for tests and we enforce each test to write its own mocking)

internal/pkg/describe/app.go
Lou1415926: Or something like
```
$ copilot@v1.30.0 copilot app show

# When v1.29.0 has different features than v1.27.0
 Version: v1.27.0 (new features available in v1.29.0)

# When v1.27.0 ~ v1.30.0 all have the same app template
 Version: v1.27.0 (up-to-date)
```

internal/pkg/deploy/cloudformation/app.go
Lou1415926: > Previously it was a const so that it is immutable, and we might want to keep this benefit so that we don't accidentally modify the version number

Makes sense to me! 

> Alternatively we can keep the function and don't set general default value for testing (it'll be empty for tests and we enforce each test to write its own mocking)

I like this too! I kinda like the idea of keeping the mocks within the client testing. (Maybe I'm just not aware of the pattern where you provide a general mock from the source package - is this a thing?)

internal/pkg/cli/flag.go
Lou1415926: Just thinking out loud here to be clear on the usage of this flag 💭 

```
# The app template is different in v1.30.0 vs. v1.28.0.
$ copiot@v1.30.0 app init
$ copilot@v1.28.0 env init  # Error out!

# Or
$ copiot@v1.26.0 app init
$ copilot@v1.30.0 app upgrade
$ copilot@v1.28.0 env init  # Error out!
```

People could just run `$ copilot@v1.30.0 env init` instead, so that it wouldn't trigger a downgrade.

Then why would a user want to force an app downgrade? There are several possibilities:
1. They _have to_ run `env init` with `v1.28.0` (due to bugs, breaking changes, etc.).
2. They want to downgrade the app template, but there is no `app deploy` nor `app downgrade`. They have to use `env init` to do so.

Lou1415926: What do you think of `--force-app-downgrade` then?

```
# The app template is different in v1.30.0 vs. v1.28.0.
$ copiot@v1.30.0 app init
$ copilot@v1.28.0 env init  --force-app-downgrade

# Or
$ copiot@v1.26.0 app init
$ copilot@v1.30.0 app upgrade
$ copilot@v1.28.0 env init --force-app-downgrade
```

iamhopaul123: > They want to downgrade the app template, but there is no app deploy nor app downgrade. They have to use env init to do so.

I was on the fence about this (actually i reverted my changes for `app upgrade` for adding this flag). The main reason is because using `app upgrade` sounds weird. Maybe it's not that strange if they only need to downgrade the app template?

> What do you think of --force-app-downgrade then?

`--force-app-downgrade` sounds to me that we'll `force` anyway. Like `git push -f` creates a new commit id no matter if there's any change or not. For our downgrade flag it's like only downgrade if needed and the following example

```
$ copiot@v1.30.0 app init
$ copilot@v1.30.0 env init  --force-app-downgrade
```
won't downgrade anyway...

Lou1415926: > I was on the fence about this (actually i reverted my changes for app upgrade for adding this flag). The main reason is because using app upgrade sounds weird. Maybe it's not that strange if they only need to downgrade the app template?

Ah it's fine - I'm not proposing that we support `app downgrade` command, or support downgrading from `app upgrade` (also don't think we should, at least not now). I'm just trying to think why users will want to run `env init --downgrade`.

> --force-app-downgrade sounds to me that we'll force anyway.

That makes sense to me. This reminds me of `--force` in service. Both flags perform the action optionally even if the flag is there. 

I guess `--allow` makes the most sense after all!

internal/pkg/describe/app.go
iamhopaul123: I'd prefer to not add this for now, as I would assume most of them discover new features by looking at our release notes/WNP 🤔

Lou1415926 (approved):

internal/pkg/version/version.go
```diff
@@ -4,5 +4,19 @@
...
+	AppTemplateMinVersionAlias = "v1.0.0"
+	// AppTemplateMinVersionStaticSite is the minimum app version required to deploy a static site.
+	AppTemplateMinVersionStaticSite = "v1.2.0"
+)
+
 // Version is this binary's version. Set with linker flags when building Copilot.
 var Version string
+
+// LatestTemplateVersion is the latest version number available for Copilot templates.
+func LatestTemplateVersion() string {
```
dannyrandall, line 20: why do we need this? isn't it the same as `version.Version`?

internal/pkg/version/version.go
```diff
@@ -4,5 +4,19 @@
 // Package version holds variables for generating version information
 package version
 
+const (
+	// LegacyAppTemplateVersion is the version associated with the application template before we started versioning.
+	LegacyAppTemplateVersion = "v0.0.0"
+	// AppTemplateMinVersionAlias is the least version number available for HTTPS alias.
+	AppTemplateMinVersionAlias = "v1.0.0"
+	// AppTemplateMinVersionStaticSite is the minimum app version required to deploy a static site.
+	AppTemplateMinVersionStaticSite = "v1.2.0"
```
dannyrandall: since these are in the `version` package now, i wonder if we could drop the "version" from them? `version.LegacyAppTemplate`, `version.MinStaticSite`, etc

internal/pkg/cli/flag.go
```diff
@@ -258,8 +259,10 @@ Allows you to categorize resources.`
 	diffAutoApproveFlagDescription = "Skip interactive approval of diff before deploying."
 
 	// Deployment.
-	deployTestFlagDescription = `Deploy your service or job to a "test" environment.`
-	forceFlagDescription      = `Optional. Force a new service deployment using the existing image.
+	deployTestFlagDescription     = `Deploy your service or job to a "test" environment.`
+	allowDowngradeFlagDescription = `Optional. Allow using a earlier version of Copilot to update Copilot components lastly
+updated by a newer version of Copilot.`
```
dannyrandall:
```suggestion
	allowDowngradeFlagDescription = `Optional. Allow using an older version of Copilot to update Copilot components
updated by a newer version of Copilot.`
```

internal/pkg/cli/env_init.go
```diff
@@ -257,10 +265,23 @@ func (o *initEnvOpts) Ask() error {
...
 func (o *initEnvOpts) Execute() error {
-	o.initRuntimeClients()
+	if err := o.initRuntimeClients(); err != nil {
+		return err
+	}
+	appVersion, err := o.appVersionGetter.Version()
+	if err != nil {
+		return fmt.Errorf("get template version of application %s: %w", o.appName, err)
+	}
+	if diff := semver.Compare(appVersion, o.templateVersion); diff > 0 && !o.allowAppDowngrade {
```
dannyrandall: i wonder if `o.allowAppDowngrade` should skip this logic entirely? in case `Version()` is returning an error due to a messed up template, it would allow you to get around that issue

dannyrandall (commented):
just a few small things!

internal/pkg/version/version.go
iamhopaul123: `version.Version` is a variable that can be altered by mistake. Previously we used const so it wasn't a problem before.

dannyrandall (approved):
!

internal/pkg/cli/flag.go
```diff
@@ -260,7 +260,7 @@ Allows you to categorize resources.`
 
 	// Deployment.
 	deployTestFlagDescription     = `Deploy your service or job to a "test" environment.`
-	allowDowngradeFlagDescription = `Optional. Allow using a earlier version of Copilot to update Copilot components lastly
+	allowDowngradeFlagDescription = `Optional. Allow using a older version of Copilot to update Copilot components
```
Lou1415926, line 263: nit: "Allow using _an_ older..." 🥺

PR #5029, opened 2023-06-27
chore: add pipeline package file with execute function

Varun359:
This pipeline package file just has an execute function useful for pipeline override


internal/pkg/cli/pipeline_package.go
```diff
@@ -0,0 +1,220 @@
...
+	jobBuffer   *bytes.Buffer
+}
+
+func (o *packagePipelineOpts) Execute() error {
+	pipelines, err := o.ws.ListPipelines()
+	if err != nil {
+		return err
+	}
+
+	pipeline_path := ""
```
dannyrandall:
```suggestion
	pipelinePath := ""
```

internal/pkg/cli/pipeline_package.go
```diff
@@ -0,0 +1,220 @@
...
+	newJobListCmd                   func(io.Writer, string) cmd
+
+	//catched variables
+	pipelineMft *manifest.Pipeline
+	app         *config.Application
+	svcBuffer   *bytes.Buffer
+	jobBuffer   *bytes.Buffer
+}
+
+func (o *packagePipelineOpts) Execute() error {
```
dannyrandall, line 44: i wonder if we should do https://github.com/aws/copilot-cli/pull/4991#discussion_r1231599761 as part of this PR? seems like most of this code is duplicated from `pipeline_deploy.go` and we should combine in `cli/deploy`

dannyrandall: chatted with @iamhopaul123 offline - let's save the refactor until after the `override` feature is merged!

dannyrandall (approved):

internal/pkg/cli/pipeline_package.go
```diff
@@ -0,0 +1,220 @@
...
+	pipelines, err := o.ws.ListPipelines()
+	if err != nil {
+		return fmt.Errorf("list all pipelines in the workspace: %w", err)
+	}
+
+	pipelinePath := ""
+	for _, pipeline := range pipelines {
+		if pipeline.Name == o.name {
+			pipelinePath = pipeline.Path
```
iamhopaul123, line 53: What if at the end of the loop the pipeline is not found and `pipelinePath` is still empty?

internal/pkg/cli/pipeline_package_test.go
```diff
@@ -0,0 +1,326 @@
...
+		Region:    region,
+		AccountID: accountID,
+	}
+
+	testCases := map[string]struct {
+		inApp          *config.Application
+		inAppName      string
+		inPipelineName string
+		inRegion       string
+		inPipelineFile string
```
iamhopaul123: I think we can simplify the unit tests by removing _some_ (I didn't check every one of them) of these inputs because we don't test extra things by giving different inputs.

For example, for each test case, no matter what `inAppName` we give, the logic will be exactly the same. As a result, we can probably just set it [here](https://github.com/aws/copilot-cli/pull/5029/files#diff-1acf3bb0c37dfcaf7dcf95ea9f1c9d7388f2dfe917673c7ba5c9145a05025c88R291)

internal/pkg/cli/pipeline_package_test.go
```diff
@@ -0,0 +1,326 @@
...
+		callMocks      func(m packagePipelineMocks)
+		expectedError  error
+	}{
+
+		"returns an error if ie fails to get the list of pipelines": {
```
iamhopaul123:
```suggestion
		"returns an error if fail to get the list of pipelines": {
```

internal/pkg/cli/pipeline_package_test.go
```diff
@@ -0,0 +1,326 @@
...
+			callMocks: func(m packagePipelineMocks) {
+				gomock.InOrder(
+					m.ws.EXPECT().ListPipelines().Return(nil, errors.New("some error")),
```
iamhopaul123: We can probably make `errors.New("some error")` a variable and reuse it.

internal/pkg/cli/pipeline_package_test.go
Varun359: Makes sense, I have updated the test cases.

internal/pkg/cli/pipeline_package.go
Varun359: yeah I errored out if it is empty

iamhopaul123: > yeah I errored out if it is empty

I don't see how this is explicitly handled 🤔 Can you show me where the error is?

internal/pkg/cli/pipeline_package.go
```diff
@@ -0,0 +1,224 @@
...
+	jobBuffer   *bytes.Buffer
+}
+
+func (o *packagePipelineOpts) Execute() error {
+	pipelines, err := o.ws.ListPipelines()
+	if err != nil {
+		return fmt.Errorf("list all pipelines in the workspace: %w", err)
+	}
+
+	pipelinePath := ""
```
Lou1415926: nit: `pipelinePath` gets initialized as `""` when declared. I feel like unless we want to  initialize `pipelinePath` to non-zero values such as `"defaultPath"`, writing it as `var pipelinePath` is less noisy, and is super clear on the data type.

```suggestion
	var pipelinePath string
```

internal/pkg/cli/pipeline_package.go
```diff
@@ -0,0 +1,224 @@
...
+	for _, pipeline := range pipelines {
+		if pipeline.Name == o.name {
+			pipelinePath = pipeline.Path
+			break
+		}
+	}
+	if pipelinePath == "" {
+		return fmt.Errorf("pipeline '%s' not found", o.name)
```
Lou1415926: nit: `%q` automatically applies the quotations!
```suggestion
		return fmt.Errorf("pipeline %q not found", o.name)
```

internal/pkg/cli/pipeline_package.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	if err := pipelineMft.Validate(); err != nil {
+		return nil, fmt.Errorf("validate pipeline manifest: %w", err)
+	}
+	o.pipelineMft = pipelineMft
+	return pipelineMft, nil
+}
+
+func (o *packagePipelineOpts) isLegacy(inputName string) (bool, error) {
```
Lou1415926, line 151: Just want to double check - after the refactor that @dannyrandall and @iamhopaul123 suggested, these duplicated codes will be gone, is that correct?

iamhopaul123: Yeah I think that's the plan.

internal/pkg/cli/pipeline_package.go
```diff
@@ -0,0 +1,224 @@
...
+
+	var pipelinePath string
+	for _, pipeline := range pipelines {
+		if pipeline.Name == o.name {
+			pipelinePath = pipeline.Path
+			break
+		}
+	}
+	if pipelinePath == "" {
+		return fmt.Errorf("pipeline %q not found", o.name)
```
Varun359, line 58: I have added it after the loop @iamhopaul123 if pipelinePath is empty (I have added that after your comment and updated the code😅)

iamhopaul123 (approved):

iamhopaul123:
Adding the DNM label. Feel free to remove it after the integ test fixing PR https://github.com/aws/copilot-cli/pull/5030 is merged.

PR #5030, opened 2023-06-27
chore(cli): apply allow-downgrade to wkld init

iamhopaul123:
...and fix integ tests

### DIFF
--- a/internal/pkg/cli/svc_init.go
+++ b/internal/pkg/cli/svc_init.go
@@ -228,7 +228,7 @@ func newInitSvcOpts(vars initSvcVars) (*initSvcOpts, error) {
 				ConfigStore: store,
 			})
 			if err != nil {
-				return nil, fmt.Errorf("initiate env describer: %w", err)
+				return nil, err
 			}
 			return envDescriber, nil
 		},
--- a/internal/pkg/cli/svc_init.go
+++ b/internal/pkg/cli/svc_init.go
@@ -15,6 +15,7 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/aws/identity"
 	"github.com/aws/copilot-cli/internal/pkg/describe"
 	"github.com/aws/copilot-cli/internal/pkg/manifest/manifestinfo"
+	"github.com/aws/copilot-cli/internal/pkg/version"
 	"github.com/dustin/go-humanize/english"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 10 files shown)


internal/pkg/cli/svc_init.go
```diff
@@ -204,20 +208,34 @@ func newInitSvcOpts(vars initSvcVars) (*initSvcOpts, error) {
...
+		},
+		initEnvDescriber: func(appName string, envName string) (envDescriber, error) {
+			envDescriber, err := describe.NewEnvDescriber(describe.NewEnvDescriberConfig{
+				App:         appName,
+				Env:         envName,
+				ConfigStore: store,
+			})
+			if err != nil {
+				return nil, fmt.Errorf("initiate env describer: %w", err)
+			}
```
dannyrandall: is this context actually useful? or could we just do
```
return describe.NewEnvDescriber()
```

internal/pkg/cli/svc_init.go
```diff
@@ -355,6 +362,15 @@ func (o *initSvcOpts) Ask() error {
 
 // Execute writes the service's manifest file and stores the service in SSM.
 func (o *initSvcOpts) Execute() error {
+	if !o.allowAppDowngrade {
```
dannyrandall, line 365: ok just thinking out loud 💭 what if we created a separate struct to embed in `initSvcOpts`, `initJobOpts`, and `initEnvOpts` like:
```
type downgradeValidator struct {
  allowDowngrade   bool
  newVersionGetter func() (versionGetter, error)
}

func (d *downgradeValidator) validate() error {}
```

then in `Execute()`
```
if err := o.appDowngradeValidator.validate(); err != nil {
}
```

and in `newOpts()`
```
o.downgradeValidator.newVersionGetter = func() (versionGetter, error) {
  return describe.NewAppDescriber(o.appName)
}
```

dannyrandall: trying to think of something we can reuse when we get to `svc deploy` and `env deploy`

dannyrandall (commented):
looks great, just a refactor thought!

internal/pkg/cli/svc_init.go
iamhopaul123: Hmm i thought about it but would prefer leaving `allowAppDowngrade` out because it's controlled by a flag. Always calling `o.appDowngradeValidator.validate()` makes less sense to me.

> trying to think of something we can reuse when we get to svc deploy and env deploy

Maybe it's ok? Within the if statement we just init the struct and call validate method.

dannyrandall: we could use the embedded value as a flag?
```
cmd.Flags().BoolVar(&vars.appDowngradeValidator.allowDowngrade, allowDowngradeFlag, false, allowDowngradeFlagDescription)
```

but yeah, you're right, it's not necessarily a ton of logic. i'll leave it up to you!

Lou1415926 (approved):
Looks good! Adding the label for you to decide on the refactor that danny said!

dannyrandall (approved):
adding dnm for https://github.com/aws/copilot-cli/pull/5030#discussion_r1245772379!

internal/pkg/cli/svc_init.go
iamhopaul123: oh mb forgot to push the commit

PR #5033, opened 2023-06-29
chore(cli): apply allow-downgrade to env deploy/package

iamhopaul123:
Also fixes integ tests and init e2e test suite

### DIFF
--- a/internal/pkg/cli/env_deploy.go
+++ b/internal/pkg/cli/env_deploy.go
@@ -154,6 +154,10 @@ func validateEnvVersion(vg versionGetter, name, templateVersion string) error {
 	if err != nil {
 		return fmt.Errorf("get template version of environment %s: %w", name, err)
 	}
+	if envVersion == version.EnvTemplateBootstrap {
+		// Allow update to bootstrap env stack anyway.
+		return nil
+	}
 	if diff := semver.Compare(envVersion, templateVersion); diff > 0 {
 		return &errCannotDowngradeEnvVersion{
 			envName:         name,
--- a/internal/pkg/cli/env_deploy.go
+++ b/internal/pkg/cli/env_deploy.go
@@ -17,25 +17,29 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/cli/deploy"
 	"github.com/aws/copilot-cli/internal/pkg/config"
 	"github.com/aws/copilot-cli/internal/pkg/deploy/cloudformation/stack"
+	"github.com/aws/copilot-cli/internal/pkg/describe"
 	"github.com/aws/copilot-cli/internal/pkg/manifest"
 	"github.com/aws/copilot-cli/internal/pkg/term/color"
 	"github.com/aws/copilot-cli/internal/pkg/term/log"
 	"github.com/aws/copilot-cli/internal/pkg/term/prompt"
 	"github.com/aws/copilot-cli/internal/pkg/term/selector"
+	"github.com/aws/copilot-cli/internal/pkg/version"
 	"github.com/aws/copilot-cli/internal/pkg/workspace"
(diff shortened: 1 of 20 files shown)


internal/pkg/cli/env_package.go
```diff
@@ -148,16 +163,25 @@ func (o *packageEnvOpts) Ask() error {
 	if _, err := o.getAppCfg(); err != nil {
 		return err
 	}
-	return o.validateOrAskEnvName()
+	return o.validateOrAskname()
```
Lou1415926: nit:
```suggestion
	return o.validateOrAskName()
```

internal/pkg/deploy/cloudformation/cloudformation_integration_test.go
```diff
@@ -55,6 +55,7 @@ func Test_App_Infrastructure(t *testing.T) {
 	deployer := cloudformation.New(sess, cloudformation.WithProgressTracker(os.Stderr))
 	cfClient := awsCF.New(sess)
 	require.NoError(t, err)
+	version.Version = "v1.28.0"
```
Lou1415926, line 58: what about we just nuke this file 🤔

internal/pkg/cli/env_deploy.go
```diff
@@ -133,8 +149,32 @@ func (o *deployEnvOpts) Ask() error {
 	return o.validateOrAskEnvName()
 }
 
+func validateEnvVersion(vg versionGetter, name, templateVersion string) error {
+	envVersion, err := vg.Version()
```
Lou1415926, line 153: what about if the version is "bootstrap"?

internal/pkg/version/version.go
```diff
@@ -11,6 +11,10 @@ const (
 	AppTemplateMinAlias = "v1.0.0"
 	// AppTemplateMinStaticSite is the minimum app version required to deploy a static site.
 	AppTemplateMinStaticSite = "v1.2.0"
+	// LegacyEnvTemplate is the version associated with the environment template before we started versioning.
+	LegacyEnvTemplate = "v0.0.0"
+	// EnvTemplateBootstrap is the version of an environment template that contains only bootstrap resources.
+	EnvTemplateBootstrap = "bootstrap"
```
Lou1415926, line 17: A small call out which I think is probably fine - 
When people run `env init`, we deploy an "environment template" that contains only the CFN execution role and the env manager role. These two roles are used when people run `env deploy`.

An "environment template" at this point contains only the two roles and doesn't not contain any actual environment resources. This template is on "bootstrap" at this point.

With the current implementation, we (probably) won't be able to detect a "downgrade" in the following case:
```
$ copilot-v1.30.0 env init # Deploy an template on "bootstrap".
$ copilot-v1.29.0 env deploy 
```

Which I think is probably fine in most of the cases - unless we've patched some important permissions two either CFNExecutionRole or EnvManagerRole on v1.30.0.

internal/pkg/deploy/cloudformation/cloudformation_integration_test.go
iamhopaul123: Yeah I am inclined too but I feel like we need at least a separate PR to remove the integration test.

internal/pkg/cli/env_deploy.go
iamhopaul123: Good call we need to deal with this case too.

internal/pkg/version/version.go
iamhopaul123: Yeah I think it should be fine because `env init` won't function anyway.

huanjani (approved):
lgtm

Lou1415926 (approved):

PR #5035, opened 2023-06-29
chore: add CloudFront cache invalidation for Static Site services

### LINKED ISSUE #5024: Invalidate CF distribution on deploy of static site
The Static Site Service does currently not invalidate existing assets on the cloudfront distribution when redeploying the service, so e.g. when index.html is updated to add a new menu item, it's not immediately visible. 

I think there should be an invalidations option for the static site service, e.g.

```
name: static-site
type: Static Site
files:
  - source: source
    recursive: true
    invalidations:
        - index.html

### PR DESCRIPTION
**Pull request by huanjani**
One of the ways in which CloudFront reduces latency is by caching objects at edge locations, reducing the number of requests that users’ origin servers must respond to directly. For Copilot-managed Static Site services, this caching behavior has the unfortunate side effect of redeployments using cached— and out-of-date— versions of source files. By default, files expire after 24 hours.

We want users to benefit from edge caches, so we don’t want to [manage cache expiration](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html) by changing TTL values or the [cache policy](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-cache-policies.html). Instead, we can allow the caching but [invalidate the cache](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Invalidation.html) each time the Static Site service is redeployed, after (updated) files are uploaded. This way, the updated files are served from the origin server rather than the outdated cached files.

We opted to invalidate all files (`/*`) for simplicity, efficiency, and [cost-savings](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Invalidation.html#PayingForInvalidation). If, in the future, customers request the ability to exclude files from invalidation, we can consider adding that config to the workload manifest.

Resolves: #5024.

### CODE REVIEW

File: internal/pkg/template/templates/workloads/services/static-site/cf.yml [hunk 1]
```diff
@@ -318,6 +330,32 @@ Resources:
                 Action:
                   - s3:PutObject
                 Resource: !Sub arn:aws:s3:::${Bucket}/*
+        - PolicyName: CacheInvalidation
```
iamhopaul123, line 333: Do we have any reference for this? How do we make sure this is the minimum required policy? especially for `Resource: "*"` for the first statement

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -283,6 +283,18 @@ Resources:
...
                     # See https://github.com/aws/aws-sdk-js/issues/1092 for more.
                     MetadataDirective: "REPLACE"
+          InvalidateCache:
+            Type: Task
+            End: true
+            Resource: arn:aws:states:::aws-sdk:cloudfront:createInvalidation
+            Parameters:
+              DistributionId: !Ref CloudFrontDistribution
+              InvalidationBatch:
+                CallerReference.$: States.UUID()
```
iamhopaul123, line 293: Should the param be `CallerReference` instead of `CallerReference.$`? Also how does `States.UUID()` work?

internal/pkg/template/templates/workloads/services/static-site/cf.yml
```diff
@@ -283,6 +283,18 @@ Resources:
...
+                Paths:
+                  Quantity: 1
```
iamhopaul123, line 295: Is the quantity here essentially `len(Items)`?

internal/pkg/template/templates/workloads/services/static-site/cf.yml
huanjani: The third example in https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/security_iam_id-based-policy-examples.html#security_iam_id-based-policy-examples-aws-managed

internal/pkg/template/templates/workloads/services/static-site/cf.yml
huanjani: `States.UUID()` is an intrinsic function within the Amazon States Language: https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-intrinsic-functions.html#asl-intrsc-func-uuid-generate

"To use intrinsic functions you must specify .$ in the key value in your state machine definitions"

internal/pkg/template/templates/workloads/services/static-site/cf.yml
huanjani: Yes. Do you think I should use `Fn::Length:` instead? I think in this case it may be more straightforward to just hard code the number bc we're always using the wildcard??

internal/pkg/template/templates/workloads/services/static-site/cf.yml
iamhopaul123: Maybe we add the shortened source URL as a comment?

internal/pkg/template/templates/workloads/services/static-site/cf.yml
iamhopaul123: Yeah if it works i guess i'll make more sense and explains itself

internal/pkg/template/templates/workloads/services/static-site/cf.yml
huanjani: If we're concerned about the permissiveness, should we add the copilot tag condition?

iamhopaul123: Im not sure if tag works for this. Maybe test it out and we should def. do if possible.

internal/pkg/template/templates/workloads/services/static-site/cf.yml
huanjani: Ah, it's an unsupported function. :(

internal/pkg/template/templates/workloads/services/static-site/cf.yml
huanjani: added the tags condition

iamhopaul123 (approved):

dannyrandall (approved):
looks great!!

PR #5043, opened 2023-07-05
chore: add wkld template versioning and svc deploy allowdowngrade

iamhopaul123:


huanjani (approved):
lgtm!

internal/pkg/deploy/cloudformation/stack/static_site_integration_test.go
```diff
@@ -83,6 +85,7 @@ func TestStaticSiteService_TemplateAndParamsGeneration(t *testing.T) {
 			// validate generated template
 			tmpl, err := serializer.Template()
 			require.NoError(t, err)
+			fmt.Println(tmpl)
```
KollaAdithya: do we need this print statement ?

iamhopaul123: good call :homer-disappearing:

internal/pkg/deploy/cloudformation/stack/scheduled_job.go
```diff
@@ -196,6 +196,7 @@ func (j *ScheduledJob) Template() (string, error) {
 		Publish:                  publishers,
 		Platform:                 convertPlatform(j.manifest.Platform),
 		EnvVersion:               j.rc.EnvVersion,
+		Version:                  j.rc.Version,
```
dannyrandall, line 199: do we need to add the flags to `job deploy`? or is that for another PR

internal/pkg/describe/env.go
```diff
@@ -201,7 +201,6 @@ func (d *EnvDescriber) Version() (string, error) {
 	if err != nil {
 		return "", err
 	}
-
```
dannyrandall, line 204: :joy:

internal/pkg/describe/workload.go
```diff
@@ -29,34 +30,57 @@ type workloadStackDescriber struct {
...
 }
 
-type workloadConfig struct {
-	app         string
-	name        string
-	configStore ConfigStoreSvc
+// NewWorkloadConfig contains fields that initiates workload describer struct.
+type NewWorkloadConfig struct {
+	App         string
+	Env         string
```
dannyrandall, line 36: good refactor 👍

internal/pkg/describe/workload.go
```diff
@@ -29,34 +30,57 @@ type workloadStackDescriber struct {
...
 		sess: sess,
 	}, nil
 }
 
+// Version returns the CloudFormation template version associated with
+// the workload by reading the Metadata.Version field from the template.
+//
+// If the Version field does not exist, then it's a legacy template and it returns an version.LegacyWorkloadTemplate and nil error.
+func (d *WorkloadStackDescriber) Version() (string, error) {
+	raw, err := d.cfn.StackMetadata()
```
dannyrandall, line 66: what happens if the service hasn't been deployed yet? doesn't this return an error, when we actually want them to be able to deploy the template?

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
```diff
@@ -2,8 +2,9 @@
 # SPDX-License-Identifier: MIT-0
 AWSTemplateFormatVersion: 2010-09-09
 Description: CloudFormation template that represents a load balanced web service on Amazon ECS.
-{{- if .SerializedManifest }}
 Metadata:
+  Version: {{ .Version }}
```
dannyrandall, line 6: i'm trying to think about what happens in the case when someone upgrades the CLI, runs `svc deploy`, and the only thing that changes about the template is this field. I assume that will return an empty change set error and not actually update the template?

assume this ⬆️ is true, then you can do something like this, assuming no changes template changes other than `Metadata.Version` between v1.40 and v1.42:
```
copilot@v1.40 svc deploy
copilot@v1.42 svc deploy # empty change set, template not updated
copilot@v1.41 svc deploy # no version error - just empty change set again, since the template didn't actually upgraded to v1.42
copilot@v1.39 svc deploy # not allowed due to template downgrade
```

i think this is actually pretty great - i'm just validating that's how this will work?

dannyrandall (commented):
looks great! just a couple questions

internal/pkg/deploy/cloudformation/stack/scheduled_job.go
iamhopaul123: yeah another PR this one is big enough

internal/pkg/template/templates/workloads/services/lb-web/cf.yml
iamhopaul123: Yeah i think wx thought about the same scenario for env template. I think should be fine?

internal/pkg/describe/workload.go
iamhopaul123: Good point we need to address this issue

iamhopaul123: Done

KollaAdithya (approved):
:shipit:

PR #5046, opened 2023-07-06
chore: add  cancel update stack api call

KollaAdithya:

### DIFF
--- a/internal/pkg/aws/cloudformation/cloudformation.go
+++ b/internal/pkg/aws/cloudformation/cloudformation.go
@@ -366,6 +366,16 @@ func (c *CloudFormation) ListStacksWithTags(tags map[string]string) ([]StackDesc
 	return summaries, nil
 }
 
+func (c *CloudFormation) CancelUpdateStack(stackName string) error {
+	_, err := c.client.CancelUpdateStack(&cloudformation.CancelUpdateStackInput{
+		StackName: aws.String(stackName),
+	})
+	if err != nil {
+		return fmt.Errorf("cancel update stack: %w", err)
+	}
+	return nil
+}
+
 func (c *CloudFormation) create(stack *Stack) (string, error) {
 	cs, err := newCreateChangeSet(c.client, stack.Name)
 	if err != nil {
--- a/internal/pkg/aws/cloudformation/cloudformation.go
+++ b/internal/pkg/aws/cloudformation/cloudformation.go
@@ -369,10 +369,9 @@ func (c *CloudFormation) ListStacksWithTags(tags map[string]string) ([]StackDesc
 // CancelUpdateStack attempts to cancel the update for a CloudFormation stack specified by the stackName.
 // Returns an error if failed to cancel CloudFormation stack update.
 func (c *CloudFormation) CancelUpdateStack(stackName string) error {
-	_, err := c.client.CancelUpdateStack(&cloudformation.CancelUpdateStackInput{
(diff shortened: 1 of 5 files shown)


internal/pkg/aws/cloudformation/cloudformation.go
```diff
@@ -366,6 +366,18 @@ func (c *CloudFormation) ListStacksWithTags(tags map[string]string) ([]StackDesc
 	return summaries, nil
 }
 
+// CancelUpdateStack attempts to cancel the update for a CloudFormation stack specified by the stackName.
+// Returns an error if failed top cancel CloudFormation stack update.
+func (c *CloudFormation) CancelUpdateStack(stackName string) error {
+	_, err := c.client.CancelUpdateStack(&cloudformation.CancelUpdateStackInput{
```
iamhopaul123: > You can cancel only stacks that are in the UPDATE_IN_PROGRESS state.

Should we ignore `ErrStackNotFound` and `ErrStackUpdateNotInProgress` (not an existing custom error type) by default in this function?

KollaAdithya: yeah we can ignore that error. Since api call only cancels stack which is in UPDATE_IN_PROGRESS state

internal/pkg/aws/cloudformation/errors.go
```diff
@@ -80,3 +80,16 @@ func stackDoesNotExist(err error) bool {
 	}
 	return false
 }
+
+// stackNotInUpdateProgress returns true if error is cannot callCancelUpdateStack which is not in UPDATE_IN_PROGRESS state.
+func stackNotInUpdateProgress(err error) bool {
```
iamhopaul123:
```suggestion
func cancelUpdateStackNotInUpdateProgress(err error) bool {
```

internal/pkg/aws/cloudformation/errors.go
```diff
@@ -80,3 +80,16 @@ func stackDoesNotExist(err error) bool {
 	}
 	return false
 }
+
+// stackNotInUpdateProgress returns true if error is cannot callCancelUpdateStack which is not in UPDATE_IN_PROGRESS state.
```
iamhopaul123:
```suggestion
// stackNotInUpdateProgress returns true if the underlying error is CancelUpdateStack cannot be called for a stack that is not in UPDATE_IN_PROGRESS state.
```

internal/pkg/aws/cloudformation/cloudformation.go
```diff
@@ -366,6 +366,20 @@ func (c *CloudFormation) ListStacksWithTags(tags map[string]string) ([]StackDesc
 	return summaries, nil
 }
 
+// CancelUpdateStack attempts to cancel the update for a CloudFormation stack specified by the stackName.
+// Returns an error if failed to cancel CloudFormation stack update.
+func (c *CloudFormation) CancelUpdateStack(stackName string) error {
+	_, err := c.client.CancelUpdateStack(&cloudformation.CancelUpdateStackInput{
+		StackName: aws.String(stackName),
+	})
+	if err != nil {
```
iamhopaul123:
```suggestion
	if _, err := c.client.CancelUpdateStack(&cloudformation.CancelUpdateStackInput{
		StackName: aws.String(stackName),
	}) err != nil {
```

iamhopaul123 (approved):
LGTM with just a nit

huanjani (approved):
Looks great. I love the PR size.

PR #5049, opened 2023-07-06
chore: boilerplate for `local run` command

Varun359:


internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/aws/copilot-cli/internal/pkg/config"
+	"github.com/aws/copilot-cli/internal/pkg/term/prompt"
+	"github.com/aws/copilot-cli/internal/pkg/term/selector"
+	"github.com/aws/copilot-cli/internal/pkg/workspace"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+)
+
+type localRunVars struct {
+	name    string
```
iamhopaul123: From the context it is difficult to indicate what the `name` is.
```suggestion
	wkldName    string
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,127 @@
...
+	appName string
+	envName string
+}
+
+type localRunOpts struct {
+	localRunVars
+
+	store store
```
iamhopaul123, line 26: It doesn't seem this is used 🤔

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,127 @@
...
+		return fmt.Errorf("select environment: %w", err)
+	}
+	o.envName = name
+	return nil
+}
+
+func (o *localRunOpts) askWorkloadName() error {
+	if o.name != "" {
+		return nil
+	}
```
iamhopaul123: We should validate the value of `o.name` passed by flag as well, to make sure it is a valid existed wkld name in the current workspace.

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,127 @@
...
+
+func (o *localRunOpts) Execute() error {
+	//TODO: Get build information from the manifest and task definition for services
+	//By username:varun359
+
+	return nil
+}
+
+func (o *localRunOpts) askEnvName() error {
+	if o.envName != "" {
```
iamhopaul123: Same for the env: we need to ensure it is an existed env in SSM config store.

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,127 @@
...
+		ws:    ws,
+		sel:   selector.NewLocalWorkloadSelector(prompt.New(), store, ws),
+	}
+	return opts, nil
+}
+
+func (o *localRunOpts) Validate() error {
+	if o.appName == "" {
+		return errNoAppInWorkspace
+	}
```
iamhopaul123, line 49: Same for the app. we need to query the ssm store to make sure it is a valid app name existed in SSM parameter store.

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,127 @@
...
+	return nil
+}
+
+// BuildLocalRunCmd builds the command for running a workload locally
+func BuildLocalRunCmd() *cobra.Command {
+	vars := localRunVars{}
+	cmd := &cobra.Command{
+		Use:   "local run",
+		Short: "Run the workload locally",
+		Long:  "Run the workload locally while replicating the ECS environment",
```
iamhopaul123: Maybe
```suggestion
		Short: "Run the workload locally",
		Long:  "Run the workload locally for debugging in a simulated AWS environment",
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	return opts, nil
+}
+
+func (o *localRunOpts) Validate() error {
+	if o.appName == "" {
+		return errNoAppInWorkspace
+	}
+	_, err := o.store.GetApplication(o.appName)
+	if err != nil {
```
iamhopaul123:
```suggestion
	if _, err := o.store.GetApplication(o.appName); err != nil {
```

internal/pkg/cli/local_run.go
Varun359: We are sending it to the selector as a parameter right? And also now we are using for Validation.

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,145 @@
...
+	name, err := o.sel.Workload("Select a workload from your workspace that you want to run locally", "")
+	if err != nil {
+		return fmt.Errorf("select Workload: %w", err)
+	}
+	o.wkldName = name
+	return nil
+}
+
+func (o *localRunOpts) validateWkldName() error {
+	if _, err := o.store.GetWorkload(o.appName, o.wkldName); err != nil {
```
iamhopaul123: > // NewLocalWorkloadSelector returns a new selector that chooses applications and environments from the config store, but
// services from the local workspace.

We should validate if the local workload exists (it could be possible that the workload is not in SSM and only has manifest locally)

internal/pkg/cli/local_run.go
iamhopaul123: > We are sending it to the selector as a parameter right

We use `store` the variable not `o.store`.

> And also now we are using for Validation.

This is true.

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,148 @@
...
+	if !contains(o.wkldName, names) {
+		return fmt.Errorf("service %q does not exist in the workspace", o.wkldName)
+	}
+	return nil
+}
+
+// BuildLocalRunCmd builds the command for running a workload locally
+func BuildLocalRunCmd() *cobra.Command {
+	vars := localRunVars{}
+	cmd := &cobra.Command{
```
dannyrandall, line 74: while we're working on this feature, can we set [`Hidden: true`](https://pkg.go.dev/github.com/spf13/cobra#Command)?

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	if err := o.validateOrAskEnvName(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (o *localRunOpts) Execute() error {
+	//TODO: Get build information from the manifest and task definition for services
+	//By username:varun359
```
dannyrandall: nit
```suggestion
	//TODO(varun359): Get build information from the manifest and task definition for services
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (o *localRunOpts) validateEnvName() error {
+	if _, err := o.store.GetEnvironment(o.appName, o.envName); err != nil {
+		return fmt.Errorf("get environment %s : %w", o.envName, err)
+	}
+	return nil
+}
+
+func (o *localRunOpts) validateOrAskWorkloadName() error {
```
dannyrandall: since `local run` only works with _deployed_ workloads, shouldn't we validate that the workload is deployed as well?

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (o *localRunOpts) Execute() error {
+	//TODO: Get build information from the manifest and task definition for services
+	//By username:varun359
+
+	return nil
+}
+
+func (o *localRunOpts) validateOrAskEnvName() error {
```
dannyrandall: similar to my other comment - shouldn't this only prompt/work for envs that the workload is actually deployed in?

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,261 @@
...
+	ws, err := workspace.Use(afero.NewOsFs())
+	if err != nil {
+		return nil, err
+	}
+
+	prompter := prompt.New()
+	opts := &localRunOpts{
+		localRunVars: vars,
+
+		prompt:      prompter,
```
dannyrandall: nit
```suggestion
		prompt:      prompt.New(),
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,261 @@
...
+		return err
+	}
+	if len(envs) == 0 {
+		return fmt.Errorf("no deployed environments found in the app %s", o.appName)
+	}
+	if len(envs) == 1 {
+		log.Infof("Only one environment found, defaulting to: %s\n", color.HighlightUserInput(envs[0]))
+		o.envName = envs[0]
+		return nil
+	}
```
dannyrandall: nit: totally personal preference, but when i have multiple `if`s related to the same-ish condition, i like to group them in a `switch` to visually relate them!
```
switch {
case err != nil:
case len(envs) == 0:
case len(envs) == 1:
}
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,261 @@
...
+		return fmt.Errorf("select environment: %w", err)
+	}
+	o.envName = selectedEnvName
+	return nil
+}
+
+func (o *localRunOpts) getDeployedEnvironments(appName string) ([]string, error) {
+	envConfig, err := o.store.ListEnvironments(o.appName)
+	if err != nil {
+		return nil, fmt.Errorf("get environments for the app %s: %w", appName, err)
```
dannyrandall:
```suggestion
		return nil, fmt.Errorf("get environments for app %s: %w", appName, err)
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	selectedEnvName, err := o.prompt.SelectOne("Select an environment in which you want to test", "", envs, prompt.WithFinalMessage("Environment:"))
+	if err != nil {
+		return fmt.Errorf("select environment: %w", err)
+	}
+	o.envName = selectedEnvName
+	return nil
+}
+
+func (o *localRunOpts) getDeployedEnvironments(appName string) ([]string, error) {
```
dannyrandall:
```suggestion
func (o *localRunOpts) getDeployedEnvironments() ([]string, error) {
```
we can just use `o.appName` everywhere!

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	var envs []string
+	for _, env := range envConfig {
+		envs = append(envs, env.Name)
+	}
+
+	var deployedEnvs []string
+	for _, env := range envs {
+		isDeployed, err := o.isEnvironmentDeployed(env)
```
dannyrandall: could we combine these two loops into a single loop?
```
for _, env := range envConfig {
  isDeployed, err := o.isEnvironmentDeployed(env.Name)
}
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+type localRunOpts struct {
+	localRunVars
+
+	store                store
+	ws                   wsWlDirReader
+	prompt               prompter
+	deployStore          deployedEnvironmentLister
+	envFeaturesDescriber func(string) (versionCompatibilityChecker, error)
```
dannyrandall: we should name this function `envVersionChecker` or something - i think it'll make more sense in the `isEnvironmentDeployed` function!

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func (o *localRunOpts) isEnvironmentDeployed(envName string) (bool, error) {
+	var checker versionCompatibilityChecker
+
+	envDescriber, err := o.envFeaturesDescriber(envName)
+	if err != nil {
+		return false, err
+	}
+	checker = envDescriber
```
dannyrandall: this can be simplified to 
```suggestion
	checker, err := o.envFeaturesDescriber(envName)
	if err != nil {
		return false, err
	}
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,261 @@
...
+
+	currVersion, err := checker.Version()
```
dannyrandall: i think this will return an error if the env has been `init`'d, but not `deploy`ed - but we only want to filter it out in that case. i wonder if we should just `return false, nil` if this returns an error?

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	if currVersion == version.EnvTemplateBootstrap {
+		return false, nil
+	}
+	return true, nil
+}
+func (o *localRunOpts) validateEnvName() error {
+	if _, err := o.store.GetEnvironment(o.appName, o.envName); err != nil {
+		return fmt.Errorf("get environment %s: %w", o.envName, err)
+	}
```
dannyrandall: i think we can probably skip this since it happens at the beginning of `NewEnvDescriber()`!

cmd/copilot/main.go
```diff
@@ -81,6 +81,7 @@ func buildRootCmd() *cobra.Command {
 	cmd.AddCommand(cli.BuildSvcCmd())
 	cmd.AddCommand(cli.BuildJobCmd())
 	cmd.AddCommand(cli.BuildTaskCmd())
+	cmd.AddCommand(cli.BuildLocalRunCmd())
```
Lou1415926, line 84: Should we add the command only after the implementation is finished, or after the next release?

Edit: nvm I saw the `Hidden: true`!

internal/pkg/cli/local_run.go
Varun359: apology for the oversight🤦‍♂️

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,257 @@
...
+func (o *localRunOpts) validateOrAskEnvName() error {
+	if o.envName != "" {
+		return o.validateEnvName()
+	}
+	envs, err := o.getDeployedEnvironments()
+	switch {
+	case err != nil:
+	case len(envs) == 0:
+	case len(envs) == 1:
+	}
```
Lou1415926: 👀 are these meant to be deleted

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,257 @@
...
+	switch {
+	case err != nil:
+		return err
+	case len(envs) == 0:
+		return fmt.Errorf("no deployed environments found in the app %s", o.appName)
```
```suggestion
		return fmt.Errorf("no environment in the app %q is deployed", o.appName)
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	selectedEnvName, err := o.prompt.SelectOne("Select an environment in which you want to test", "", envs, prompt.WithFinalMessage("Environment:"))
+	if err != nil {
+		return fmt.Errorf("select environment: %w", err)
+	}
+	o.envName = selectedEnvName
+	return nil
+}
+
+func (o *localRunOpts) getDeployedEnvironments() ([]string, error) {
```
Lou1415926: what do you think of moving this to the deploy package? 
```go
// In deploy/deploy.go
type EnvironmentStore struct {
	configStore         ConfigStoreClient
         versionGetter      versionGetter    
}

func (s *EnvironmentStore) ListDeployedEnvironments() ([]string, error) {}

// here
envs, err := o.environmentStore.ListDeployedEnvironments()
	switch {
	case err != nil:
	case len(envs) == 0:
	case len(envs) == 1:
	}
```

internal/pkg/cli/local_run.go
Varun359: Thanks for mentioning, it is very evident that only one loop is needed. I think some of the things I just wrote to see what output I am getting . In this case I just printed out envs and deployedEnvs separately to differentiate the output and then forgot to remove one extra loop😑.

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,257 @@
...
+	return nil
+}
+
+func (o *localRunOpts) validateOrAskWorkloadName() error {
+	if o.wkldName != "" {
+		return o.validateWkldName()
+	}
+	services, err := o.deployStore.ListDeployedServices(o.appName, o.envName)
+	if err != nil {
+		return fmt.Errorf("Get services: %w", err)
```
```suggestion
		return fmt.Errorf("get deployed services: %w", err)
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	jobs, err := o.deployStore.ListDeployedJobs(o.appName, o.envName)
+	if err != nil {
+		return fmt.Errorf("Get jobs: %w", err)
```
```suggestion
		return fmt.Errorf("get deployed jobs: %w", err)
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	workloads := append(services, jobs...)
+	if len(workloads) == 0 {
+		return fmt.Errorf("no workloads found in this environment %s", o.envName)
```
```suggestion
		return fmt.Errorf("no workload is deployed to this environment %s", o.envName)
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return err
+	}
+	if !isDeployed {
+		return fmt.Errorf(`cannot use an environment which is not deployed Please run "copilot env deploy, --name %s" to deploy the environment first`, o.envName)
+	}
+	return nil
+}
+
+func (o *localRunOpts) validateOrAskWorkloadName() error {
```
Lou1415926: qq: I see we require the workload to be deployed AND to be in the current workspace? I just want to double check -  that environment does not have the second requirement, is that right?

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return fmt.Errorf("Get jobs: %w", err)
+	}
+
+	workloads := append(services, jobs...)
+	if len(workloads) == 0 {
+		return fmt.Errorf("no workloads found in this environment %s", o.envName)
+	}
+	if len(workloads) == 1 {
+		log.Infof("Only one workload found in this environment, defaulting to: %s\n", color.HighlightUserInput(workloads[0]))
```
Lou1415926: I'm starting to wonder if the prompt flow makes sense 🤔 . An alternative that I can think of is to reverse the env-wkld order. 
```
$ copilot local run
What workload do you want to test? web
Select an environment in which you want to test "web":
   > test
   > prod
```
What are our thoughts on this?

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+		o.wkldName = selectedWorloadName
+	}
+	localWorkloads, err := o.ws.ListWorkloads()
+	if err != nil {
+		return fmt.Errorf("list workloads in the workspace : %w", err)
+	}
+	if !contains(o.wkldName, localWorkloads) {
+		return fmt.Errorf("selected service %q does not exist in the workspace", o.wkldName)
+	}
```
Lou1415926: In this case, we shouldn't have included these workloads in the options that we prompted with.

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,257 @@
...
+	return nil
+}
+
+// BuildLocalRunCmd builds the command for running a workload locally
+func BuildLocalRunCmd() *cobra.Command {
+	vars := localRunVars{}
+	cmd := &cobra.Command{
+		Use:    "local run",
+		Short:  "Run the workload locally",
+		Long:   "Run the workload locally for debugging in a simulated AWS environment",
```
Lou1415926:
```suggestion
		Long:   "Run the workload locally for debugging in a simulated AWS environment",
```
I am not sure about the "simulated". It sounds like we are simulating the VPC, db, etc. but we are not right?

internal/pkg/cli/local_run.go
Varun359: If the env is just  `init`'d, but not `deploy`ed, In that case only we get `bootstrap` right from the metadata?

internal/pkg/cli/local_run.go
Varun359: yes😅

internal/pkg/cli/local_run.go
Varun359: This is not even going to happen I guess but what if the workload got deleted locally?

Lou1415926: > what if the workload got deleted locally

Then it won't appear in the value returned by `o.ws.ListWorkloads`, and we won't think that the workload is "in the workspace"

internal/pkg/cli/local_run.go
dannyrandall: oh good point! i forgot about that 😊

internal/pkg/cli/local_run.go
Varun359: Yes, we had a discussion on this earlier. To get the deployed workloads. We should iterate through all the envs to get them and then when the user selects the workload, we need to get all the environments in which the workload is deployed in and prompt to the user.

````
envs := getAllDeployedEnvs() #-step1
for env in envs:
  wklds.add(env.listServices())
  wklds.add(env.listJobs())

#prompt to select a deployed service
if service in multiple envs:
  #prompt to select env for service
````

Even this approach works but In this case doesn't it increase the latency(delay) even more after doing this? Since we are doing the step 1 in the above code already. Isn't it better to prompt the user the deployedEnvs first (instead of iterating through it and get the deployed workloads) and then get deployedWorkloads(there is no difference between both I guess)?

Comment (Lou1415926) [reply internal/pkg/cli/local_run.go]: 👍🏼  I agree that performance-wise, prompting env and then workload reduces the number of API calls / for loops that have to happen.

> there is no difference between both I guess

I would somewhat disagree with this though. In terms of the set of workloads that we end up getting, they are indeed the same. But in terms of user experience, different prompting orders are not the same.

I think when people run `local run`, their mindset is likely workload-oriented (i.e. they think "I want to test my 'web' workload"), instead of environment-oriented (i.e. "I want to test my "test" environment). This was why I liked the workload-first prompting better.

Again, I agree that the current approach is better at performance. However, there are a ton of for-loops / API calls that we can parallelize. If we have done these parallelization, I think performance might not have been that big of a difference (maybe it'd be 0.7s if we prompt workload first vs. 0.5s if we prompt environment first).

This is not a huge issue, I can commit to the current prompting order! We should probably have a sprint task that parallelizes these things (not in your scope!), so that performance will no longer be a major concern when we think about designs like this.

File: internal/pkg/cli/local_run.go (hunk 31 above)
Comment (Varun359) [reply internal/pkg/cli/local_run.go]: We need this since  this is a validation when we give svc name as a flag right?

Comment (Lou1415926) [reply internal/pkg/cli/local_run.go]: When we give svc name as a flag, I thought we'd have hit L180:
```
if o.wkldName != "" {
	return o.validateWkldName()
}
```
and then go straight to `validateWkldName`.

This `if !contains(o.wkldName, localWorkloads)` line will only execute if workload is prompted for 🤔 ?

Comment (Varun359) [reply internal/pkg/cli/local_run.go]: Oh sorry I thought you commented I that functions check!

File: internal/pkg/cli/local_run.go [hunk 33]
```diff
@@ -0,0 +1,243 @@
...
+	return nil
+}
+
+func (o *localRunOpts) validateOrAskEnvName() error {
+	if o.envName != "" {
+		return o.validateEnvName()
+	}
+
+	if len(o.wkldDeployedToEnvs[o.wkldName]) == 1 {
+		log.Infof("Only one environment found, defaulting to: %s\n", color.HighlightUserInput(o.wkldDeployedToEnvs[o.wkldName][0]))
```
```suggestion
		log.Infof("The workload %s is deployed to only one environment, defaulting to: %s\n", color.HighlightUserInput(o.wkldName), color.HighlightUserInput(o.wkldDeployedToEnvs[o.wkldName][0]))
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,243 @@
...
+	if err != nil {
+		return fmt.Errorf("list deployed environments for application %s: %w", o.appName, err)
+	}
+	isDeployed, err := o.isEnvironmentDeployed(o.envName)
+	if err != nil {
+		return err
+	}
+	if !isDeployed {
+		return fmt.Errorf(`cannot use an environment which is not deployed Please run "copilot env deploy, --name %s" to deploy the environment first`, o.envName)
+	}
```
Lou1415926:
```suggestion
```

We might not need to check this anymore, because if it is "bootstrap", then the workload wouldn't have been able to deployed to the environment. Is there any other concern that you've thought of that I've missed?

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func (o *localRunOpts) validateOrAskWorkloadName() error {
+	if o.wkldName != "" {
+		return o.validateWkldName()
+	}
+
+	localWorkloads, err := o.ws.ListWorkloads()
+	if err != nil {
+		return fmt.Errorf("list workloads in the workspace %s : %w", o.appName, err)
```
Lou1415926:
```suggestion
		return fmt.Errorf("list workloads in the workspace %s: %w", o.appName, err)
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,243 @@
...
+
+	if len(o.deployedWkld) == 0 {
+		return fmt.Errorf("no workload is deployed in app %s", o.appName)
+	}
+	if len(o.deployedWkld) == 1 {
+		log.Infof("Only one deployed workload found, defaulting to: %s\n", color.HighlightUserInput(o.deployedWkld[0]))
+		o.wkldName = o.deployedWkld[0]
+		return nil
+	}
+	selectedWorloadName, err := o.prompt.SelectOne("Select a workload that you want to run locally", "", o.deployedWkld, prompt.WithFinalMessage("workload name"))
```
Lou1415926:
```suggestion
	selectedWorkloadName, err := o.prompt.SelectOne("Select a workload that you want to run locally", "", o.deployedWkld, prompt.WithFinalMessage("Workload:"))
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,243 @@
...
+	if err != nil {
+		return fmt.Errorf("select Workload: %w", err)
```
Lou1415926:
```suggestion
		return fmt.Errorf("select a workload: %w", err)
```

internal/pkg/cli/local_run_test.go
```diff
@@ -0,0 +1,260 @@
...
+	ws          *mocks.MockwsWlDirReader
+	prompt      *mocks.Mockprompter
+	deployStore *mocks.MockdeployedEnvironmentLister
+}
+
+func TestLocalRunOpts_Validate(t *testing.T) {
+	testCases := map[string]struct {
+		wantedError error
+	}{
+		"no app in workspace": {
```
Lou1415926, line 30: Can we add two more test cases:
1. "fail to read the application from SSM store" with a `m.store.Expect().GetApplication().Return(nil, testError)`
2. "success"

Lou1415926 (approved):
LGTM :shipit: all are nits

internal/pkg/cli/local_run.go
Varun359: Good point! I wanted to show the error to deploy the env first, but I think I can do it inside the isEnvironmentDeployed fn. Updated😊

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,243 @@
...
+		})
+		if err != nil {
+			return nil, fmt.Errorf("new environment compatibility checker: %v", err)
+		}
+		return envDescriber, nil
+	}
+	return opts, nil
+}
+
+func (o *localRunOpts) Validate() error {
```
iamhopaul123, line 46: Can we add comments to all the public methods/functions?

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,243 @@
...
+		deployStore:        deployStore,
+	}
+	opts.envVersionGetter = func(envName string) (versionGetter, error) {
+		envDescriber, err := describe.NewEnvDescriber(describe.NewEnvDescriberConfig{
+			App:         opts.appName,
+			Env:         envName,
+			ConfigStore: opts.store,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("new environment compatibility checker: %v", err)
```
iamhopaul123: nit we don't usually wrap the error returned by initiators
```suggestion
			return nil, err
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,243 @@
...
+	if err != nil {
+		return fmt.Errorf("list deployed environments for application %s: %w", o.appName, err)
+	}
+	_, err = o.isEnvironmentDeployed(o.envName)
+	if err != nil {
+		return err
+	}
+	// if !isDeployed {
+	// 	return fmt.Errorf(`cannot use an environment which is not deployed Please run "copilot env deploy, --name %s" to deploy the environment first`, o.envName)
+	// }
```
iamhopaul123: Why do we even call `isEnvironmentDeployed` if we are not using its result anyway? I'd suggest add in the next PR if we are using it in the future.

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,91 @@
...
+	opts := &localRunOpts{
+		localRunVars: vars,
+
+		store: store,
+	}
+	return opts, nil
+}
+
+// Validate validates the application name for running a workload locally.
+// It ensures that the application name is provided and exists in the workspace.
```
iamhopaul123: It seems very specific about the implementation for app part, I'd suggest move this into the function and get a more general description (maybe look at the other commands)

iamhopaul123 (approved):
:shipit:

PR #5060, opened 2023-07-10
chore: add methods to conditionally remove stackset instance from app

bvtujo:
PR 1 of 2 replacing #4992.

This adds support methods and a whole boatload of mocks and unit tests to the stackset client and Cloudformation packages. These are necessary to conditionally remove the stackset instance, or redeploy the stackset instances & app to remove an account from dns delegation and IAM roles.


internal/pkg/aws/cloudformation/stackset/stackset.go
```diff
@@ -112,27 +112,53 @@ func (ss *StackSet) UpdateAndWait(name, template string, opts ...CreateOrUpdateO
...
+// but the API call will take no action.
+func (ss *StackSet) DeleteInstance(name, account, region string) (string, error) {
+	out, err := ss.client.DeleteStackInstances(&cloudformation.DeleteStackInstancesInput{
+		StackSetName: aws.String(name),
+		Accounts:     aws.StringSlice([]string{account}),
+		Regions:      aws.StringSlice([]string{region}),
+		RetainStacks: aws.Bool(false),
+	})
+	if err != nil {
+		return "", fmt.Errorf("delete stack instances in region %v for account %v for stackset %s: %w",
```
iamhopaul123:
```suggestion
		return "", fmt.Errorf("delete stack instance in region %v for account %v for stackset %s: %w",
```
I guess we only have one stack instance?

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -313,12 +373,50 @@ func (cf CloudFormation) RemoveJobFromApp(app *config.Application, jobName strin
...
+	EnvAccountID         string
+	EnvRegion            string
+	DeleteStackInstance  bool // If this is the last environment in a region, we can delete it.
+	RemoveAccountFromApp bool // If this is the last environment in an account, we need to redeploy the app stack.
+}
+
+// RemoveEnvFromApp optionally redeploys the app stack to remove the account and deletes the stackset
+// instance for that region. This method cannot check that deleting the stackset or removing the app
+// won't break copilot. Be careful.
+func (cf CloudFormation) RemoveEnvFromApp(opts *RemoveEnvFromAppOpts) error {
```
iamhopaul123, line 386: Why don't we move the logic for determining the value for "DeleteStackInstance" and "RemoveAccountFromApp" here in this function? So as to make sure calling `RemoveEnvFromApp` is always safe.

iamhopaul123 (commented):
This is awesome!

internal/pkg/deploy/cloudformation/app.go
bvtujo: We'll have to duplicate this logic somewhere, though, because we have to check in env_delete.go whether or not to empty the s3 buckets and ecr repos. I don't love that, but I see the advantage of having logic in the RemoveEnvFromApp call.

iamhopaul123: > because we have to check in env_delete.go whether or not to empty the s3 buckets and ecr repos.

Maybe we can let `RemoveEnvFromApp` decide if we need to empty the s3 buckets and ecr repo before deleting the stackset instance?

internal/pkg/cli/job_delete.go
```diff
@@ -303,8 +304,7 @@ func (o *deleteJobOpts) emptyECRRepos(envs []*config.Environment) error {
 		}
 	}
 
-	// TODO: centralized ECR repo name
```
Lou1415926, line 306: ty for doing the todo!

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -313,12 +373,77 @@ func (cf CloudFormation) RemoveJobFromApp(app *config.Application, jobName strin
 	return nil
 }
 
+// RemoveEnvFromAppOpts contains the parameters to call RemoveEnvFromApp.
+type RemoveEnvFromAppOpts struct {
+	App          *config.Application
+	EnvToDelete  *config.Environment
+	Environments []*config.Environment
```
iamhopaul123, line 380: I feel like we should probably leave `Environments` out of the input params? Like I would expect  `RemoveEnvFromApp` to take only `envToRemove` and `App`.

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -313,12 +373,77 @@ func (cf CloudFormation) RemoveJobFromApp(app *config.Application, jobName strin
...
+}
+
+// RemoveEnvFromApp optionally redeploys the app stack to remove the account and, if necessary, empties
```
iamhopaul123, line 383: Shall we update this comment accordingly?

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -129,6 +129,66 @@ func (cf CloudFormation) upgradeAppStack(conf *stack.AppStackConfig) error {
...
+		return fmt.Errorf("update application to remove DNS delegation from account %s: %w", accountID, err)
+	}
+	// Update stackset instances to remove account.
+	appResourcesConfig, err := cf.getLastDeployedAppConfig(appStack)
+	if err != nil {
+		return err
+	}
+	appResourcesConfig.Accounts = newAccountList
+	if err := cf.deployAppConfig(newCfg, appResourcesConfig, true); err != nil {
+		return fmt.Errorf("update application regional resources to remove account %s: %w", accountID, err)
```
iamhopaul123: nit: do we need to wrap this error?

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -129,6 +129,66 @@ func (cf CloudFormation) upgradeAppStack(conf *stack.AppStackConfig) error {
...
+		}
+		return fmt.Errorf("update application to remove DNS delegation from account %s: %w", accountID, err)
+	}
+	// Update stackset instances to remove account.
+	appResourcesConfig, err := cf.getLastDeployedAppConfig(appStack)
+	if err != nil {
+		return err
+	}
+	appResourcesConfig.Accounts = newAccountList
+	if err := cf.deployAppConfig(newCfg, appResourcesConfig, true); err != nil {
```
iamhopaul123, line 186: Callout not for this PR but we probably need to refactor `deployAppConfig` to avoid using any boolean value (maybe split into two)

iamhopaul123 (approved):

internal/pkg/cli/pipeline_deploy_test.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/config"
 	"github.com/aws/copilot-cli/internal/pkg/deploy"
 	deploycfn "github.com/aws/copilot-cli/internal/pkg/deploy/cloudformation"
+	deploymock "github.com/aws/copilot-cli/internal/pkg/deploy/cloudformation/mocks"
```
Lou1415926: Do we have to use the mocks from the other packages 👀  Could we declare the interface as well as the mock in `cli` package?

internal/pkg/deploy/cloudformation/app.go
bvtujo: Can you think of an easier way for the Cloudformation package to get access to the environments than passing them in here? I don't want to pass a `store` into the `CloudFormation` struct if I can avoid it; that feels like too much abstraction blending to me and it would be a first for this package.

bvtujo: The CFN stacks we're dealing with don't have any metadata about where the environments are deployed. Typically we fetch information about the application with a config store or deploy store object, but again, passing a store to CFN feels like inverting the level of abstraction (and may lead to a circular dependency if we use a deploystore)

bvtujo:
Applying a label to stop merging while we work through discussions

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -313,12 +373,77 @@ func (cf CloudFormation) RemoveJobFromApp(app *config.Application, jobName strin
...
+	}
+
+	deployApp := &deploy.CreateAppInput{
+		Name:               opts.App.Name,
+		AccountID:          opts.App.AccountID,
+		DomainName:         opts.App.Domain,
+		DomainHostedZoneID: opts.App.DomainHostedZoneID,
+		Version:            opts.App.Version,
+	}
+	appConfig := stack.NewAppStackConfig(deployApp)
```
Lou1415926: nit:
```suggestion
	appConfig := stack.NewAppStackConfig(&deploy.CreateAppInput{
		Name:               opts.App.Name,
		AccountID:          opts.App.AccountID,
		DomainName:         opts.App.Domain,
		DomainHostedZoneID: opts.App.DomainHostedZoneID,
		Version:            opts.App.Version,
	})
```

internal/pkg/deploy/cloudformation/app.go
iamhopaul123: Yeah totally agreed. I'm good for leaving it as is for now and probably refactor in the following PR if we can find a good way.

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -313,12 +373,77 @@ func (cf CloudFormation) RemoveJobFromApp(app *config.Application, jobName strin
...
+
+	if !regionHasOtherEnvs {
```
Lou1415926, line 412: I like that these are handled in the same function! It feels weird though that all these logics non-CFN are in the `deploy/cloudformation` package. Maybe this signals that we will need a refactor soon.

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -569,6 +694,17 @@ func (cf CloudFormation) deleteStackSetInstances(name string) error {
...
 
+func (cf CloudFormation) deleteStackSetInstance(name, account, region string) error {
+	opId, err := cf.appStackSet.DeleteInstance(name, account, region)
+	if err != nil {
+		if IsEmptyErr(err) {
+			return nil
+		}
+		return err
+	}
+	return cf.appStackSet.WaitForOperation(name, opId)
```
Lou1415926, line 704: Just making sure - if there is no instance in the said account + region, i take that we will still get an operation ID. What happens when we wait for that empty operation? Is that an instant return?

internal/pkg/deploy/cloudformation/app_test.go
```diff
@@ -1204,3 +1205,525 @@ func TestCloudFormation_RenderStackSet(t *testing.T) {
...
+						StackID: "some-stack",
+					},
+				}, nil)
+				regionalCfn.EXPECT().Describe("some-stack").Return(mockValidAppResourceStack(), nil)
+				s3.EXPECT().EmptyBucket("tests3-bucket-us-west-2").Return(nil)
+				ecr.EXPECT().ClearRepository("phonetool-svc").Return(nil)
+				// Delete stackset instance
+				appStackSet.EXPECT().DeleteInstance("phonetool-infrastructure", "1234", "us-west-2").Return("123", nil)
+				appStackSet.EXPECT().WaitForOperation("phonetool-infrastructure", "123").Return(nil)
+
```
Lou1415926, line 1409: nit: we can add ⬇️  to be super clear that we want no such call. The other maintainers would be able to see the intention clearly.
```go
appStackSet.Expect().Describe("phonetool-infrastructure").Times(0)
appStackSet.Expect(). Update("phonetool-infrastructure").Times(0)

```

internal/pkg/deploy/cloudformation/app_test.go
```diff
@@ -1204,3 +1205,525 @@ func TestCloudFormation_RenderStackSet(t *testing.T) {
...
+  Version: 17
+  Workloads:
+    - Name: ar
+      WithECR: true
+  Accounts:
+    - 1234
+    - 5678`,
+				}, nil)
+				appStackSet.EXPECT().Update("phonetool-infrastructure", gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return("123", nil)
+
```
Lou1415926, line 1491: Samesies :ty

Lou1415926 (commented):
lgtm! just a couple of nits & questions

internal/pkg/cli/pipeline_deploy_test.go
bvtujo: I don't know what's going on with this package; for whatever reason my local mainline wouldn't build because of some interface issues. I'd prefer to remove this commit (and these pipeline changes) entirely. Probably will happen when I resolve merge conflicts.

internal/pkg/deploy/cloudformation/app.go
bvtujo: Agreed, I think that perhaps this stuff needs to get pulled up one level into the deploy package. Or moved over into clideploy. It's probably worth putting all our deletion logic into that package as well as the deployment logic, now that I think about it.

internal/pkg/deploy/cloudformation/app.go
bvtujo: Yeah, it instantly returns "succeeded". CFN views it as a successful delete even though nothing was deleted.

internal/pkg/cli/pipeline_deploy_test.go
bvtujo: fixed by #5077; i'd like to get that merged to mainline before this PR

Lou1415926 (approved):
Applying a DNM so that this can be merged after the mock PR

internal/pkg/deploy/cloudformation/app.go
Lou1415926: Quick question! Why do we need to call `deployAppConfig` which looks like is updating the stackset? 

My understanding so far is that stackset is only touched when we add/delete new regions & add/delete new workloads. Why do we need to update stackset when we remove a account?

Lou1415926: oh I think I see now :homer-disappearing

We need to update the app stack so that we remove the account from delegation role; we need to update the stackset so that we remove the account from the metadata.

PR #5061, opened 2023-07-10
chore: update env delete to optionally remove stackset and dns delegation

bvtujo:
PR 2 of 2 replacing #4992.


internal/pkg/cli/env_delete.go
```diff
@@ -169,9 +172,23 @@ func (o *deleteEnvOpts) Execute() error {
 		return err
 	}
 	o.prog.Stop(log.Ssuccessf(fmtRetainEnvRolesComplete, o.name))
+
 	if err := o.deleteStack(); err != nil {
+		o.prog.Stop(log.Serrorf("Failed to delete resources for the %q environment\n", o.name))
+		return err
+	}
+
```
Lou1415926: did we have to `Start` in order to `Stop`? I'm not sure about what happens if we `Stop` a `prog` displayer that hasn't been `Start`ed yet.

```suggestion
        o.prog.Start(fmt.Sprintf("Deleting resources for the %q environment\n", o.name))
	if err := o.deleteStack(); err != nil {
		o.prog.Stop(log.Serrorf("Failed to delete resources for the %q environment\n", o.name))
		return err
         }
         o.prog.Stop(log.Ssuccessf("Deleted resources for the %q environment\n", o.name))
```

internal/pkg/cli/env_delete.go
```diff
@@ -169,9 +172,23 @@ func (o *deleteEnvOpts) Execute() error {
...
+	envInfo, err := o.getEnvConfig()
+	if err != nil {
+		return err
+	}
+	// Un-delegate DNS and optionally delete stackset instance.
+	o.prog.Start("Cleaning up app-level resources and permissions\n")
+	if err := o.cleanUpAppResources(); err != nil {
+		o.prog.Stop(log.Serrorf("Failed to remove ECR Repos and S3 bucket from region %s\n", envInfo.Region))
```
Lou1415926:
```suggestion
		o.prog.Stop(log.Serrorf("Failed to remove environment resource from app"))
```

If I understand it correctly, we don't remove ECR repos and S3 buckets from region, if the environment is _not_ the last environment in that region. Therefore, I feel like we should just print a more general log here.

Lou1415926 (approved):

internal/pkg/deploy/cloudformation/app.go
```diff
@@ -415,7 +415,8 @@ func (cf CloudFormation) RemoveEnvFromApp(opts *RemoveEnvFromAppOpts) error {
 		if err != nil {
 			return err
 		}
-		if err := cf.s3Client.EmptyBucket(resources.S3Bucket); err != nil {
+		s3 := cf.regionalS3Client(opts.EnvToDelete.Region)
```
iamhopaul123:
```suggestion
		s3Client := cf.regionalS3Client(opts.EnvToDelete.Region)
```
in case we conflate with the pkg name

iamhopaul123 (approved):

PR #5062, opened 2023-07-10
fix: filter out non-active ECS clusters

### LINKED ISSUE #5052: Bug: "more than one ECS cluster are found with tags" ..
My `copilot task run` invocation is failing with the error message: `run task <task name>: get cluster for environment dev1: more than one ECS cluster are found with tags "copilot-application"="<app name>","copilot-environment"="dev1"`

I didn't see any other such clusters in the web console. I put some debugging prints in the copilot source code and found which ARN the second cluster was about. With aws cli I could call describe-clusters on it and saw that it was in INACTIVE state. In https://docs.aws.amazon.com/cli/latest/reference/ecs/delete-cluster.html it is said that "Clusters with an INACTIVE status might remain discoverable in your account for a period of time".

The deleted cluster is from (at least) last week, this is not a transient state.

It would seem to me that when searching for ECS clusters by tag, the copilot code is either not filtering out INACTIVE status clusters or there is some other problem in in the cluster finding logic.

Update: also happens with other commands, eg `copilot job delete` - discovered as I attempted to delete the environment contents (trying to recreate the environment with a new name).

dannyrandall:
Fixes #5052

### DIFF
--- a/internal/pkg/aws/ecs/ecs.go
+++ b/internal/pkg/aws/ecs/ecs.go
@@ -319,6 +319,29 @@ func (e *ECS) HasDefaultCluster() (bool, error) {
 	return true, nil
 }
 
+// ActiveClusters returns the subset of arns that have an ACTIVE status.
+func (e *ECS) ActiveClusters(arns ...string) ([]string, error) {
+	resp, err := e.client.DescribeClusters(&ecs.DescribeClustersInput{
+		Clusters: aws.StringSlice(arns),
+	})
+	switch {
+	case err != nil:
+		return nil, fmt.Errorf("describe clusters: %w", err)
+	case len(resp.Failures) > 0:
+		return nil, fmt.Errorf("describe clusters: %s", resp.Failures[0].GoString())
+	}
+
+	var active []string
+	for _, cluster := range resp.Clusters {
+		if aws.StringValue(cluster.Status) == clusterStatusActive {
+			active = append(active, aws.StringValue(cluster.ClusterArn))
+		}
+	}
+
+	return active, nil
+
+}
+
 // RunTask runs a number of tasks with the task definition and network configurations in a cluster, and returns after
 // the task(s) is running or fails to run, along with task ARNs if possible.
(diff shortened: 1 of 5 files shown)


internal/pkg/aws/ecs/ecs.go
```diff
@@ -319,6 +319,29 @@ func (e *ECS) HasDefaultCluster() (bool, error) {
...
+// ActiveClusters returns the subset of arns that have an ACTIVE status.
+func (e *ECS) ActiveClusters(arns ...string) ([]string, error) {
+	resp, err := e.client.DescribeClusters(&ecs.DescribeClustersInput{
+		Clusters: aws.StringSlice(arns),
+	})
+	switch {
+	case err != nil:
+		return nil, fmt.Errorf("describe clusters: %w", err)
+	case len(resp.Failures) > 0:
+		return nil, fmt.Errorf("describe clusters: %s", resp.Failures[0].GoString())
```
iamhopaul123, line 331: Why do we need this 🤔

iamhopaul123 (approved):
Great fix! Do we only have this problem with cluster?s

huanjani (approved):
lgtm! thx for the fix.

erno:
Reporter of https://github.com/aws/copilot-cli/issues/5052 here - I had a change to test this as I could still trigger the bug with released copilot version. This fixes it for me.

dannyrandall:
@erno thanks for the confirmation!!☺️

internal/pkg/aws/ecs/ecs.go
dannyrandall: just in case the describe fails for any of the ARNs - seems reasonable to error out in that case? do you think we should just ignore it? 

the only case i can realistically see something like this occurring is in this scenario:
1. we get the service ARNs by tag
2. ECS _really_ deletes the service
3. we call describe services here with the ARN that is now deleted

iamhopaul123: Ohh I guess I didn't realize in this case `DescribeClusters` would actually return `nil` err and put the err in the response.

PR #5069, opened 2023-07-11
chore: signal package for handling interrupts

KollaAdithya:
This PR will handle signal interrupts received and notify to channel

### DIFF
--- a/internal/pkg/term/signal/signal.go
+++ b/internal/pkg/term/signal/signal.go
@@ -1,5 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+
+// Package signal provides a way to capture and handle operating system signals.
 package signal
 
 import (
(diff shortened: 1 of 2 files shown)


internal/pkg/term/signal/signal.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// NotifySignals starts capturing the specified signals and returns a channel for receiving them.
+func (s *Signal) NotifySignals() <-chan os.Signal {
+	signal.Notify(s.signalCh, s.sigs...)
+	return s.signalCh
+}
+
+// StopCatchSignals stops capturing signals and closes the signalChannel.
+func (s *Signal) StopCatchSignals() {
```
iamhopaul123, line 33: How do you plan to use these methods 🤔 My worry is these methods won't be useful enough as they are pretty simple and we might need to make these wrap up more logic when handling signals.

KollaAdithya: Agee with you. I will add more logic to this PR that will handle the whole `ctrl-c` for all the workloads. But the only concern may be PR will be big.

KollaAdithya:
Closing this in favor of #5097

PR #5074, opened 2023-07-12
chore(test): use mock instead of real stack struct for static site unit test

Lou1415926:
Currently, the static site unit test is using the real `stack.StaticSite` struct in the `cli/deploy` unit test. This requires the `custom-resource` to be built from real FS, i.e. you have to build first in order to run the unit test.

This PR uses mocks instead of the real `stack.StaticSite` struct 


iamhopaul123 (approved):
🙏

bvtujo (approved):

PR #5075, opened 2023-07-12
chore(workspace): search for X levels up from the start directory

Lou1415926:

### DIFF
--- a/internal/pkg/workspace/errors.go
+++ b/internal/pkg/workspace/errors.go
@@ -32,6 +32,20 @@ func (e *ErrFileNotExists) Error() string {
 	return fmt.Sprintf("file %s does not exists", e.FileName)
 }
 
+// ErrTargetNotFound means that we couldn't locate the target file or the target directory.
+type ErrTargetNotFound struct {
+	startDir              string
+	target                string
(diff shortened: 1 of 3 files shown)


internal/pkg/workspace/errors.go
```diff
@@ -32,18 +32,30 @@ func (e *ErrFileNotExists) Error() string {
 	return fmt.Sprintf("file %s does not exists", e.FileName)
 }
 
+// ErrTargetNotFound means that we couldn't locate the target file or the target directory.
+type ErrTargetNotFound struct {
```
iamhopaul123, line 36: Why do we need an additional error type 🤔 Is it because `ErrWorkspaceNotFound` is specifically for directory?

Lou1415926: > Is it because ErrWorkspaceNotFound is specifically for directory?

This is partially the reason. It is mostly because `ErrWorkspaceNotFound` has `RecommendedActions` that's very specific.

iamhopaul123 (approved):

KollaAdithya (approved):
lgtm!

Issue #5050, opened 2023-07-07
Support `yarn install` for Copilot CDK overrides


houdinisparks:
My project uses
- yarn as a package manager
- yarn workspaces to handle our monorepo packages

I will like to keep the CDK dependencies living in the same dependency ecosystem the monorepo is living in, for maintainability purposes.

Currently the when using CDK Overrides, it assumes that the project uses `npm` and does a `npm install`:

In [_copilot-cli/internal/pkg/override/cdk.go:_](https://github.com/aws/copilot-cli/blob/28f0de09f17e2340011dd63c3cb631f2b204eaf2/internal/pkg/override/cdk.go#L102)
```go 
func (cdk *CDK) install() error {
	if _, err := cdk.exec.LookPath("npm"); err != nil {
		return &errNPMUnavailable{parent: err}
	}

	cmd := cdk.exec.Command("npm", "install")  # <-- this line here
	cmd.Stdout = cdk.execWriter
	cmd.Stderr = cdk.execWriter

	if err := cmd.Run(); err != nil {
		return fmt.Errorf(`run %q: %w`, cmd.String(), err)
	}
	return nil
}
```

Resulting in the error below

```
$ copilot svc package --name app --env dev                                                                                                                           
npm ERR! code EUNSUPPORTEDPROTOCOL
npm ERR! Unsupported URL Type "workspace:": workspace:^

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/<REDACTED>/.npm/_logs/2023-07-07T06_10_40_933Z-debug-0.log
generate workload app template against environment dev: override template: run "/<REDACTED>/.nvm/versions/node/v18.13.0/bin/npm install": exit status 1
arm64 (base) FAIL
```

It will be great if we could pass in a flag or something to `copilot package` to use `yarn` instead of `npm` when interacting with the `cdk` for overrides.

Lou1415926:
Hey @houdinisparks Good callout, thanks for letting us know! For now, a workaround that I can think of is to create a symlink: `ln -s /usr/local/bin/yarn /usr/local/bin/npm`. After this, you should be able to run `copilot svc package` and `copilot svc deploy`.

[...]

PR #5076, opened 2023-07-12
chore: infer package manager when overriding with CDK

Lou1415926:
Will have to rebase on #5075

Resolve parts of #5050 on using yarn instead of npm for cdk override, with a small change from the [the first part of the comment I gave](https://github.com/aws/copilot-cli/issues/5050#issuecomment-1625851258).

I decided not include the drafted proposal of "reading package manager selection from env var"; instead, this PR just defaults to `npm` if after all the checks, there is still a tie. This is because I am not clear on whether it is a common design practice to use env vars to decide which package manager to use, and I expect that the current solution (check install + check project)  is already enough to resolve most of the case.

We can always add the enhancement of "reading package manager selection from env var" afterwards, if it's deemed needed.

This does not handle the second ask from the user, i.e. not assuming the location of `node_modules/`.

### DIFF
--- a/internal/pkg/override/cdk.go
+++ b/internal/pkg/override/cdk.go
@@ -235,7 +235,7 @@ func (cdk *CDK) closestProjectManager() (string, error) {
 			return afero.Exists(cdk.fs, path)
 		}, candidate.lockFile)
 		if err == nil {
-			distance := strings.Count(wd, "/") - strings.Count(path, "/")
+			distance := strings.Count(wd, string(filepath.Separator)) - strings.Count(path, string(filepath.Separator))
 			if distance < closestDistance {
 				closestCandidate = candidate.name
 				closestDistance = distance
--- a/internal/pkg/override/cdk.go
+++ b/internal/pkg/override/cdk.go
@@ -35,6 +35,7 @@ type CDK struct {
 		LookPath func(file string) (string, error)
 		Command  func(name string, args ...string) *exec.Cmd
 		Find     func(startDir string, maxLevels int, matchFn workspace.TraverseUpProcessFn) (string, error)
+		Getwd    func() (dir string, err error)
 	} // For testing os/exec calls.
 }
 
@@ -86,10 +87,12 @@ func WithCDK(root string, opts CDKOpts) *CDK {
 			LookPath func(file string) (string, error)
 			Command  func(name string, args ...string) *exec.Cmd
 			Find     func(startDir string, maxLevels int, matchFn workspace.TraverseUpProcessFn) (string, error)
+			Getwd    func() (dir string, err error)
 		}{
 			LookPath: lookPathFn,
 			Command:  cmdFn,
 			Find:     workspace.TraverseUp,
+			Getwd:    os.Getwd,
 		},
 	}
 }
@@ -209,8 +212,6 @@ func (cdk *CDK) installedPackageManagers() ([]string, error) {
 	return installed, nil
 }
 
-var getwd = os.Getwd
-
 // closestProjectManager returns the package manager of the project.
 // It searches five levels up from the working directory and look for the lock file of each package managers.
 // If no lock file is found, it returns an empty string.
@@ -218,7 +219,7 @@ var getwd = os.Getwd
 // If multiple are found, it returns the package manager whose lock file is closer to the working dir.
 // If multiple are equally close, then it returns the one whose name is the alphabetically smallest.
 func (cdk *CDK) closestProjectManager() (string, error) {
-	wd, err := getwd()
+	wd, err := cdk.exec.Getwd()
 	if err != nil {
 		return "", fmt.Errorf("get working directory: %w", err)
 	}
--- a/internal/pkg/override/cdk.go
+++ b/internal/pkg/override/cdk.go
@@ -45,7 +45,6 @@ type CDKOpts struct {
 	EnvVars    map[string]string                           // Environment variables key value pairs to pass to the "cdk synth" command.
 	LookPathFn func(executable string) (string, error)     // Search for the executable under $PATH. Defaults to exec.LookPath.
 	CommandFn  func(name string, args ...string) *exec.Cmd // Create a new executable command. Defaults to exec.Command rooted at the overrides/ dir.
-	FindFn     func(startDir string, maxLevels int, matchFn workspace.TraverseUpProcessFn) (string, error)
 }
 
 // WithCDK instantiates a new CDK Overrider with root being the path to the overrides/ directory.
@@ -79,10 +78,6 @@ func WithCDK(root string, opts CDKOpts) *CDK {
 	if opts.CommandFn != nil {
 		cmdFn = opts.CommandFn
 	}
-	findFn := workspace.TraverseUp
-	if opts.FindFn != nil {
-		findFn = opts.FindFn
-	}
 	return &CDK{
 		rootAbsPath: root,
 		execWriter:  writer,
@@ -94,7 +89,7 @@ func WithCDK(root string, opts CDKOpts) *CDK {
 		}{
 			LookPath: lookPathFn,
 			Command:  cmdFn,
-			Find:     findFn,
+			Find:     workspace.TraverseUp,
 		},
 	}
 }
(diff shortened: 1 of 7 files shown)


huanjani:
Related: #5050

internal/pkg/override/cdk.go
```diff
@@ -172,6 +182,92 @@ func (cdk *CDK) cleanUp(in []byte) ([]byte, error) {
...
 }
 
+type packageManager struct {
+	name     string
+	lockFile string
+}
+
+var packageManagers = []packageManager{ // Alphabetically sorted based on name.
+	{
+		name:     "npm",
```
KollaAdithya, line 194: can we use constants for `npm` and `yarn`
```
const (
    npm   = "npm"
    yarn  = "yarn"
)
```

internal/pkg/override/cdk.go
```diff
@@ -172,6 +182,92 @@ func (cdk *CDK) cleanUp(in []byte) ([]byte, error) {
...
+		lockFile: "package-lock.json",
+	},
+	{
+		name:     "yarn",
+		lockFile: "yarn.lock",
+	},
+}
+
+func (cdk *CDK) installedPackageManager() ([]string, []error) {
```
KollaAdithya: I am not completely sure if this works.

can we use `multierror.Append()` instead of returning `[]error`. If this does not work pls ignore this comment.

internal/pkg/override/cdk.go
```diff
@@ -172,6 +182,92 @@ func (cdk *CDK) cleanUp(in []byte) ([]byte, error) {
...
+	}
+	var closestCandidate string
+	closestDistance := math.MaxInt
+	var errTargetNotFound *workspace.ErrTargetNotFound
+	for _, candidate := range packageManagers {
+		path, err := cdk.exec.Find(wd, 5, func(path string) (bool, error) {
+			return afero.Exists(cdk.fs, path)
+		}, candidate.lockFile)
+		if err == nil {
+			distance := strings.Count(wd, "/") - strings.Count(path, "/")
```
KollaAdithya: Is it possible to use `filepath.Separator`
```
distance := strings.Count(wd, string(filepath.Separator)) - strings.Count(path, string(filepath.Separator))
```

internal/pkg/override/cdk.go
Lou1415926: good point, I addressed the feedback using `errors.Join`https://pkg.go.dev/errors#example-Join

internal/pkg/override/cdk.go
Lou1415926: I prefer keeping it as is if that's fine with you! Here is my reasoning -

I feel like we want to use `constant`s (instead of a "magic value") mostly in code logic:
```go
// Example: "npm" is a magic value
if _, err := cdk.exec.LookPath("npm"); err == nil {}

// Example: no magic value
npm := "npm"
if _, err := cdk.exec.LookPath(npm); err == nil {}
```

Using constant instead magic value makes it easier to refactor the code - you just need to change the value of `npm` to take care of all references to it. 

In this case, the `packageManagers` is a package-scoped variable instead of a part of "the logic". Consider this - if we use constants here, the code will looks like:
```go
const (
    npm   = "npm"
    yarn  = "yarn"
)

var packageManagers = []packageManager{
   {  name: npm, }
   {   name: yarn, }
}

if _, err := cdk.exec.LookPath(candidate.name); err == nil {}

```

The number of magic value did not decrease. All the logic code is still using `candidate.name` to refer to the package manager name. The only place that refers to `npm` is `packageManagers`.

Simply put, I feel like the problems with using "magic values" is already solved by `candidate.name` - and adding a layer of constant on top of `packageManagers` may not be a huge help.

Lou1415926: But I did add this https://github.com/aws/copilot-cli/pull/5076/commits/7ad6de58ca07f4aa43640b4ea5f711e750e7093e that's to address your comment!

KollaAdithya: This is really helpful. I got to learn when to use `const` and when not to

internal/pkg/override/cdk.go
```diff
@@ -172,6 +181,95 @@ func (cdk *CDK) cleanUp(in []byte) ([]byte, error) {
...
+	for _, candidate := range packageManagers {
+		if _, err := cdk.exec.LookPath(candidate.name); err == nil {
+			installed = append(installed, candidate.name)
+		} else {
+			lookUpExecutableErrs = append(lookUpExecutableErrs, &errorExecutableNotFound{
+				executable: candidate.name,
+				error:      err,
+			})
+		}
+	}
```
dannyrandall: seems like the only error case we care about is `len(installed) == 0` - should we remove this case and just return an error if we don't find _any_?

internal/pkg/override/cdk.go
```diff
@@ -172,6 +181,95 @@ func (cdk *CDK) cleanUp(in []byte) ([]byte, error) {
...
+		name:     "npm",
+		lockFile: "package-lock.json",
+	},
+	{
+		name:     "yarn",
+		lockFile: "yarn.lock",
+	},
+}
+
+func (cdk *CDK) installedPackageManager() ([]string, error) {
```
dannyrandall: nit
```suggestion
func (cdk *CDK) installedPackageManagers() ([]string, error) {
```

internal/pkg/override/cdk.go
```diff
@@ -172,6 +181,95 @@ func (cdk *CDK) cleanUp(in []byte) ([]byte, error) {
...
+			lookUpExecutableErrs = append(lookUpExecutableErrs, &errorExecutableNotFound{
+				executable: candidate.name,
+				error:      err,
+			})
+		}
+	}
+	return installed, errors.Join(lookUpExecutableErrs...)
+}
+
+var getwd = os.Getwd
```
dannyrandall: should this be a func on `CDK`?

internal/pkg/override/cdk.go
```diff
@@ -172,6 +181,95 @@ func (cdk *CDK) cleanUp(in []byte) ([]byte, error) {
...
+	}
+	if len(installed) == 1 {
+		return installed[0], nil
+	}
+	manager, err := cdk.closestProjectManager()
+	if err != nil {
+		return "", err
+	}
+	if manager == "" {
+		manager = defaultPackageManager
```
dannyrandall:
```suggestion
		return defaultPackageManager, nil
```

internal/pkg/override/errors.go
```diff
@@ -7,17 +7,30 @@ import (
...
+}
+
+type errPackageManagerUnavailable struct {
+	parentError error
 }
 
-func (err *errNPMUnavailable) Error() string {
-	return fmt.Sprintf(`"npm" cannot be found: "npm" is required to override with the Cloud Development Kit: %v`, err.parent)
+func (err *errPackageManagerUnavailable) Error() string {
+	return fmt.Sprintf("cannot find a package manager to override with the Cloud Development Kit:\n%s", err.parentError.Error())
```
dannyrandall: nit - i think this is the case? does it only work with javascript package managers?
```suggestion
	return fmt.Sprintf("cannot find a javascript package manager to override with the Cloud Development Kit:\n%s", err.parentError.Error())
```

dannyrandall (commented):
looks mostly good, just some small feedback!

internal/pkg/override/cdk.go
Lou1415926: Good point, I addressed this by doing the following instead:
```go
func (cdk *CDK) installedPackageManagers() ([]string, error) {
	var installed []string
	for _, candidate := range packageManagers {
		if _, err := cdk.exec.LookPath(candidate.name); err == nil {
			installed = append(installed, candidate.name)
		} else if !errors.Is(err, exec.ErrNotFound) { // Return unknown error! Swallow ErrNotFound.
			return nil, err
		}
	}
	return installed, nil
}


func (cdk *CDK) packageManager() (string, error) {
	installed, err := cdk.installedPackageManagers()
	if err != nil { // Handle unknown error.
		return "", err
	}
	if len(installed) == 0 {
		return "", &errPackageManagerUnavailable{}
	}
```

dannyrandall (approved):

KollaAdithya (approved):
LGTM!

PR #5077, opened 2023-07-13
chore: bring mocks back into alignment with interfaces

bvtujo:
Our generated mocks got out of alignment with our tests. Some unit test files relied on mocks of an interface that no longer existed, for example, leading to merge conflicts in PRs which had to re-generate mocks for other reasons. 

This PR fixes pipeline unit tests using a new `stackConfiguration` interface which is identical to the deploy.StackConfiguration interface.

I think this signals a need for two action items
1. Add a CI check to generate mocks and check for an empty git diff
2. Create a new `PipelineDeployer` object in the `clideploy` package at `internal/pkg/cli/deploy/pipeline.go` to capture cfn/deployment abstractions for `pipeline deploy` and `pipeline package`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,8 +57,9 @@ jobs:
 
       - name: Check for nonzero git diff
         run: |
-            git status --porcelain=v1 1>gitstatus.txt
-            [ -z "$(cat gitstatus.txt)" ] && echo "Generated mocks all match"
+             git status --porcelain=v1 1>gitstatus.txt
+             cat gitstatus.txt
+             [ -z "$(cat gitstatus.txt)" ] && echo "Generated mocks all match"
 
   test:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,11 +55,13 @@ jobs:
       - name: Generate mocks
         run: make gen-mocks
 
-      - name: Check for nonzero git diff
+      - name: Print out local diff after mockgen
         run: |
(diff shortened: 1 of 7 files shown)


internal/pkg/cli/interfaces.go
```diff
@@ -6,6 +6,7 @@ package cli
 import (
 	"context"
 	"encoding"
+	sdkcloudformation "github.com/aws/aws-sdk-go/service/cloudformation"
```
Lou1415926, line 9: We will be able to remove this dependency from `cli` package after we have `cli/deploy/pipeline.go`, is that right?

Lou1415926 (approved):

iamhopaul123 (approved):

internal/pkg/cli/interfaces.go
bvtujo: Yeah, the dependency will be moved down into the `clideploy` package and disappear from `cli`.

.github/workflows/ci.yml
```diff
@@ -36,6 +36,33 @@ jobs:
...
+
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Generate mocks
+        run: make gen-mocks
+
+      - name: Print out local diff after mockgen
+        run: |
+             git status --porcelain=v1 1>gitstatus.txt
```
dannyrandall, line 60: nit: could also just https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---exit-code

dannyrandall (commented):
awesome, thanks @bvtujo! i love the new action workflow👏

PR #5078, opened 2023-07-13
chore: make RemoveEnvFromApp truly idempotent

bvtujo:
Makes RemoveEnvFromApp really idempotent; previously we were getting errors when no resources existed before trying to empty the s3 buckets and ecr repos. 

Also fixes a bug where the s3 bucket emptier doesn't get created in the right region and therefore can't empty the bucket.


Lou1415926 (approved):
lgtm! If the code with the problems aren't released yet, I think we can keep the title as `chore`

Edit: I reviewed again after the idempotent commit, and the approval still hold 👍🏼

huanjani (approved):
LGTM

PR #5086, opened 2023-07-14
chore: decrypt secrets from SSM and Secrets Manager

Varun359:
This is PR is linked to #5104

### DIFF
--- a/internal/pkg/cli/interfaces.go
+++ b/internal/pkg/cli/interfaces.go
@@ -6,8 +6,9 @@ package cli
 import (
 	"context"
 	"encoding"
-	sdkcloudformation "github.com/aws/aws-sdk-go/service/cloudformation"
 	"io"
+
+	sdkcloudformation "github.com/aws/aws-sdk-go/service/cloudformation"
 
 	"github.com/aws/aws-sdk-go/aws/session"
 	awscloudformation "github.com/aws/copilot-cli/internal/pkg/aws/cloudformation"
@@ -518,6 +519,7 @@ type deploySelector interface {
 	appSelector
 	DeployedService(prompt, help string, app string, opts ...selector.GetDeployedWorkloadOpts) (*selector.DeployedService, error)
 	DeployedJob(prompt, help string, app string, opts ...selector.GetDeployedWorkloadOpts) (*selector.DeployedJob, error)
+	DeployedWorkload(prompt, help string, app string, opts ...selector.GetDeployedWorkloadOpts) (*selector.DeployedWorkload, error)
 }
 
 type pipelineEnvSelector interface {
--- a/internal/pkg/cli/local_run.go
+++ b/internal/pkg/cli/local_run.go
@@ -11,11 +11,17 @@ import (
 	"github.com/aws/copilot-cli/internal/pkg/aws/identity"
 	"github.com/aws/copilot-cli/internal/pkg/aws/sessions"
 	"github.com/aws/copilot-cli/internal/pkg/config"
+	"github.com/aws/copilot-cli/internal/pkg/deploy"
+	"github.com/aws/copilot-cli/internal/pkg/term/prompt"
+	"github.com/aws/copilot-cli/internal/pkg/term/selector"
 	"github.com/spf13/cobra"
 )
 
+const workloadAskPrompt = "Which workload would you like to run locally?"
+
 type localRunVars struct {
 	wkldName string
+	wkldType string
 	appName  string
 	envName  string
 }
@@ -23,6 +29,7 @@ type localRunVars struct {
 type localRunOpts struct {
 	localRunVars
 
+	sel   deploySelector
 	store store
 }
 
@@ -34,10 +41,14 @@ func newLocalRunOpts(vars localRunVars) (*localRunOpts, error) {
 	}
 
 	store := config.NewSSMStore(identity.New(defaultSess), ssm.New(defaultSess), aws.StringValue(defaultSess.Config.Region))
+	deployStore, err := deploy.NewStore(sessProvider, store)
+	if err != nil {
+		return nil, err
+	}
 	opts := &localRunOpts{
 		localRunVars: vars,
-
-		store: store,
+		sel:          selector.NewDeploySelect(prompt.New(), store, deployStore),
+		store:        store,
 	}
 	return opts, nil
 }
@@ -56,8 +67,28 @@ func (o *localRunOpts) Validate() error {
 
 // Ask prompts the user for any unprovided required fields and validates them.
 func (o *localRunOpts) Ask() error {
-	//TODO(varun359): Validate and Ask SvcEnvName
+	return o.validateAndAskWkldEnvName()
+}
 
+func (o *localRunOpts) validateAndAskWkldEnvName() error {
+	if o.envName != "" {
+		if _, err := o.store.GetEnvironment(o.appName, o.envName); err != nil {
+			return err
+		}
+	}
+	if o.wkldName != "" {
+		if _, err := o.store.GetWorkload(o.appName, o.wkldName); err != nil {
+			return err
+		}
+	}
+
+	deployedWorkload, err := o.sel.DeployedWorkload(workloadAskPrompt, "", o.appName, selector.WithEnv(o.envName), selector.WithName(o.wkldName))
+	if err != nil {
+		return fmt.Errorf("select a deployed workload from application %s: %w", o.appName, err)
+	}
+	o.wkldName = deployedWorkload.Name
+	o.envName = deployedWorkload.Env
+	o.wkldType = deployedWorkload.Type
 	return nil
 }
 
--- a/internal/pkg/cli/local_run_test.go
+++ b/internal/pkg/cli/local_run_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/aws/copilot-cli/internal/pkg/cli/mocks"
 	"github.com/aws/copilot-cli/internal/pkg/config"
+	"github.com/aws/copilot-cli/internal/pkg/term/selector"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
 )
@@ -18,6 +19,7 @@ var testError = errors.New("some error")
 
 type localRunAskMocks struct {
 	store *mocks.Mockstore
+	sel   *mocks.MockdeploySelector
 }
 
 func TestLocalRunOpts_Validate(t *testing.T) {
@@ -70,6 +72,117 @@ func TestLocalRunOpts_Validate(t *testing.T) {
 				require.EqualError(t, err, tc.wantedError.Error())
 			} else {
 				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestLocalRunOpts_Ask(t *testing.T) {
+	const (
+		testAppName  = "testApp"
+		testEnvName  = "testEnv"
+		testWkldName = "testWkld"
+		testWkldType = "testWkldType"
+	)
+	testCases := map[string]struct {
+		inputAppName  string
+		inputEnvName  string
+		inputWkldName string
+
+		setupMocks     func(m *localRunAskMocks)
+		wantedWkldName string
+		wantedEnvName  string
+		wantedWkldType string
+		wantedError    error
+	}{
+		"error if provided environment is not present in the workspace": {
+			inputAppName: testAppName,
+			inputEnvName: testEnvName,
+			setupMocks: func(m *localRunAskMocks) {
+				m.store.EXPECT().GetEnvironment(testAppName, testEnvName).Return(nil, testError)
+			},
+			wantedError: testError,
+		},
+		"error if provided workload is not present in the workspace": {
+			inputAppName:  testAppName,
+			inputWkldName: testWkldName,
+			setupMocks: func(m *localRunAskMocks) {
+				m.store.EXPECT().GetEnvironment(gomock.Any(), gomock.Any()).Times(0)
+				m.store.EXPECT().GetWorkload(testAppName, testWkldName).Return(nil, testError)
+			},
+			wantedError: testError,
+		},
+		"successfully validate env and svc with flags passed in": {
+			inputAppName:  testAppName,
+			inputWkldName: testWkldName,
+			inputEnvName:  testEnvName,
+			setupMocks: func(m *localRunAskMocks) {
+				m.store.EXPECT().GetEnvironment(testAppName, testEnvName).Return(&config.Environment{Name: "testEnv"}, nil)
+				m.store.EXPECT().GetWorkload(testAppName, testWkldName).Return(&config.Workload{Name: "testWkld"}, nil)
+				m.sel.EXPECT().DeployedWorkload(workloadAskPrompt, "", testAppName, gomock.Any()).Return(&selector.DeployedWorkload{
+					Env:  "testEnv",
+					Name: "testWkld",
+					Type: "testWkldType",
+				}, nil)
+			},
+			wantedEnvName:  testEnvName,
+			wantedWkldName: testWkldName,
+			wantedWkldType: testWkldType,
+		},
+		"prompt for workload and environment": {
+			inputAppName: testAppName,
+			setupMocks: func(m *localRunAskMocks) {
+				m.store.EXPECT().GetEnvironment(gomock.Any(), gomock.Any()).Times(0)
+				m.store.EXPECT().GetWorkload(gomock.Any(), gomock.Any()).Times(0)
+				m.sel.EXPECT().DeployedWorkload(workloadAskPrompt, "", testAppName, gomock.Any()).Return(&selector.DeployedWorkload{
+					Env:  "testEnv",
+					Name: "testWkld",
+					Type: "testWkldType",
+				}, nil)
+			},
+			wantedEnvName:  testEnvName,
+			wantedWkldName: testWkldName,
+			wantedWkldType: testWkldType,
+		},
+		"return error while failed to select workload": {
+			inputAppName: testAppName,
+			setupMocks: func(m *localRunAskMocks) {
+				m.sel.EXPECT().DeployedWorkload(workloadAskPrompt, "", testAppName, gomock.Any()).
+					Return(nil, testError)
+			},
+			wantedError: fmt.Errorf("select a deployed workload from application %s: %w", testAppName, testError),
+		},
+	}
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			// GIVEN
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+			m := &localRunAskMocks{
+				store: mocks.NewMockstore(ctrl),
+				sel:   mocks.NewMockdeploySelector(ctrl),
+			}
+			tc.setupMocks(m)
+			opts := localRunOpts{
+				localRunVars: localRunVars{
+					appName:  tc.inputAppName,
+					wkldName: tc.inputWkldName,
+					envName:  tc.inputEnvName,
+				},
+				store: m.store,
+				sel:   m.sel,
+			}
+
+			// WHEN
+			err := opts.Ask()
+
+			// THEN
+			if tc.wantedError == nil {
+				require.NoError(t, err)
+				require.Equal(t, tc.wantedWkldName, opts.wkldName)
+				require.Equal(t, tc.wantedEnvName, opts.envName)
+			} else {
+				require.EqualError(t, err, tc.wantedError.Error())
 			}
 		})
 	}
--- a/internal/pkg/cli/mocks/mock_interfaces.go
+++ b/internal/pkg/cli/mocks/mock_interfaces.go
@@ -5774,6 +5774,26 @@ func (mr *MockdeploySelectorMockRecorder) DeployedService(prompt, help, app inte
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "DeployedService", reflect.TypeOf((*MockdeploySelector)(nil).DeployedService), varargs...)
 }
 
+// DeployedWorkload mocks base method.
+func (m *MockdeploySelector) DeployedWorkload(prompt, help, app string, opts ...selector.GetDeployedWorkloadOpts) (*selector.DeployedWorkload, error) {
+	m.ctrl.T.Helper()
+	varargs := []interface{}{prompt, help, app}
+	for _, a := range opts {
+		varargs = append(varargs, a)
+	}
+	ret := m.ctrl.Call(m, "DeployedWorkload", varargs...)
+	ret0, _ := ret[0].(*selector.DeployedWorkload)
+	ret1, _ := ret[1].(error)
+	return ret0, ret1
+}
+
+// DeployedWorkload indicates an expected call of DeployedWorkload.
+func (mr *MockdeploySelectorMockRecorder) DeployedWorkload(prompt, help, app interface{}, opts ...interface{}) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	varargs := append([]interface{}{prompt, help, app}, opts...)
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "DeployedWorkload", reflect.TypeOf((*MockdeploySelector)(nil).DeployedWorkload), varargs...)
+}
+
 // MockpipelineEnvSelector is a mock of pipelineEnvSelector interface.
 type MockpipelineEnvSelector struct {
 	ctrl     *gomock.Controller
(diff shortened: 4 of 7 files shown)


internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+func (o *localRunOpts) Execute() error {
+	taskDef, err := o.ecsLocalClient.TaskDefinition(o.appName, o.envName, o.wkldName)
+	if err != nil {
+		return fmt.Errorf("get task definition: %w", err)
+	}
+
+	awsSession := session.Must(session.NewSessionWithOptions(
+		session.Options{
+			SharedConfigState: session.SharedConfigEnable,
+		}))
```
Lou1415926: Eventually we might need to take into consideration the actual permissions of `ExecutionRole`. However I think it is okay _for now_ that we are using user's current credential (instead of the actual `ExecutionRole`) in `local run`. (I'd like to confirm with @iamhopaul123 and dannyrandall@ though that this is the option that we are going with today 👀 )

My question is what do you think of using the `defaultSession` (see L69) instead of creating a new session?

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+
+	_, err = o.getDecryptedSSMSecrets(taskDef, awsSession)
+	if err != nil {
+		fmt.Println("unable to get secrets from SSM store")
```
Lou1415926: We typically wrap an error only if it is returned from an external package:
```go
if err := otherPackage.AFunction(); err != nil { return fmt.Errorf("error from AFunction: %w", err) }
```

The wrapping gives a context of the error and gives the users/us an idea of where the error happens.

The "package-level" context is in most of the time enough to identify the place at which the error is returned. Wrap an intra-package error, in this case, usually just adds noise to the error message. Consider if we wrap every single error on every calls, then the error message would be super long and is basically a function call stack 😂 


```suggestion
		return err
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	_, err = o.getDecryptedSecretManagerSecrets(awsSession)
+	if err != nil {
+		fmt.Println("unable to get secrets from Secret Manager")
```
Lou1415926:
```suggestion
		return err
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+		session.Options{
+			SharedConfigState: session.SharedConfigEnable,
+		}))
+
+	_, err = o.getDecryptedSSMSecrets(taskDef, awsSession)
+	if err != nil {
+		fmt.Println("unable to get secrets from SSM store")
+	}
+
+	_, err = o.getDecryptedSecretManagerSecrets(awsSession)
```
Lou1415926: We are ignoring the first output (the actual secrets) because they will be used in the next PR, is that right?

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+	return nil
+}
+
+func (o *localRunOpts) getDecryptedSSMSecrets(taskDef *awsecs.TaskDefinition, awsSession *session.Session) ([]ssmSecret, error) {
+	var SSMSecretsList []ssmSecret
+	secrets := taskDef.Secrets()
+	ssmClient := ssm.New(awsSession)
+	for _, secret := range secrets {
+		resp, err :=
+			ssmClient.GetParameter(&ssm.GetParameterInput{
```
Lou1415926:
```suggestion
		resp, err := ssmClient.GetParameter(&ssm.GetParameterInput{
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+				Name:           aws.String(secret.ValueFrom),
+				WithDecryption: aws.Bool(true),
+			})
+		if err != nil {
+			return nil, err
```
Lou1415926: This is an example where we _should_ wrap the error.
```suggestion
			return nil, fmt.Errorf("get SSM parameter %q: %w", aws.String(secret.ValueFrom), err)
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+		}
+		secretInfo := ssmSecret{
+			Name:  secret.Name,
+			Value: *resp.Parameter.Value,
```
Lou1415926: Is `resp.Parameter.Value` guaranteed to be non-nil? Dereferencing a `nil` pointer will result in `panic`.

In addition, I'd suggest using `aws.StringValue(resp.Parameter.Value)`, which will return an empty string if the pointer is nil.

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+	_, err = o.getDecryptedSecretManagerSecrets(awsSession)
+	if err != nil {
+		fmt.Println("unable to get secrets from Secret Manager")
+	}
+
+	return nil
+}
+
+func (o *localRunOpts) getDecryptedSSMSecrets(taskDef *awsecs.TaskDefinition, awsSession *session.Session) ([]ssmSecret, error) {
+	var SSMSecretsList []ssmSecret
```
Lou1415926: nit: a suggestion on the variable name + we don't need to capitalize a variable that's not exported 👀 
```suggestion
	var ssmSecrets []ssmSecret
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+		}
+		SSMSecretsList = append(SSMSecretsList, secretInfo)
+	}
+	return SSMSecretsList, nil
+}
+
+func (o *localRunOpts) getDecryptedSecretManagerSecrets(awsSession *session.Session) ([]secretManager, error) {
+	secretsManagerClient := secretsmanager.New(awsSession)
+	var secretStruct secretManager
+	var SecretManagerSecrets []secretManager
```
Lou1415926: same!

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+	input := &secretsmanager.ListSecretsInput{}
```
Lou1415926: Where is `input` used 🤔

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+package cli
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/secretsmanager"
+	"github.com/aws/aws-sdk-go/service/ssm"
```
Lou1415926: Can we keep the dependency on `aws-sdk-go/service` within `internal/aws`, so that `internal/cli` depends on `internal/aws`, and `internal/aws` depends on `aws-sdk-go/service`?


```go
// In internal/aws/ssm/ssm.go
func (s *SSM) GetSecretValue(name string) (string, error)

// In internal/aws/secretsmanager/secretsmanager.go
func (s *SecretsManager) GetSecretValue(name string) (string, error)

// Here in `getDecrpyted...`
value, err := o.secretsmanager.GetSecretValue(name)
```

A few benefits that I can think of:
1. This keeps most, if not all, of external dependencies within fewer packages. Consider if `aws-sdk-go/service` changes their API, we'd only need to touch our codes in `internal/aws`, instead of all over the place in the whole codebase.
2. Keeps logic that's specific to interacting with `aws-sdk-go/service` within `internal/aws`. For example, pagination is a common thing in HTTP requests. Consider this - if each of our packages directly interacts with `aws-sdk-go` instead of through `internal/aws`, we'd have paginations for-loops all over the place, distracting us from the main logic, making the code harder to read. 
3. Reuse the code. Still the pagination example - if we had handled the pagination in a `internal/aws` package, we wouldn't have to have the same pagination for-loop every where. 
4. Our internal packages can benefit from our own abstraction.  Consider this example - suppose we have
```go
// In internal/aws/ecs/ecs.go

import "github.com/aws/aws-sdk-go/service/ecs"

type ecsClient interface {
   DeployService(*ecs.DeployServiceInput) error
}

type ECS struct {
   client ecsClient
}

func New(sess session) *ECS { 
    return &ECS{ 
        client: ecs.New(sess) // initialized with a client from aws-sdk-go/ecs
    }
}

func (ecs *ECS) UpdateService() error {
     if err := ecs.client.UpdateService(&ecs.DeployServiceInput{}); err != nil {...}
}
```
In a third package (e.g. `cli`), we can have ⬇️ :
```go
import "github.com/aws/copilot-cli/internal/pkg/aws/ecs"

type  svcUpdator interface {
   UpdateService() error
}

type lbws struct {
   updator svcUpdator
}

func NewLBWS() *lbws {
  return &lbws{
     deployer: ecs.New(sess) // This is a client from internal ecs package, not aws-sdk-go.
  }
}

func (lbws *lbws) Execute() error {
   if err := lbws.deployer.UpdateService(); err != nil { ... }
}
```

If some day, we decide to switch from ECS to AppRunner as the service provider, we can just add a new package:
```go
// In internal/aws/apprunner/apprunner.go
import "github.com/aws/aws-sdk-go/service/apprunner"

type apprunnerClient interface {
   DeployService(*apprunner.DeployServiceInput) error // Unlike aws-sdk-go/service/ecs, the apprunner API is `DeployService`, not `UpdateService`.
}

type AppRunner struct {
   client apprunnerClient // This will be initialized with apprunner.New(sess), that is, this is a client from aws-sdk-go
}

func (ar *AppRunner) UpdateService() error { // We abstract away `DeployService`, and keep the same `UpdateService` interface as our `internal/aws/ecs` package to minimize the amount of code changes
     if err := ar.client.DeployService(&apprunner.DeployServiceInput{}); err != nil {...}
}
```
and then just change one line of code in `cli` 
```go
func NewLBWS() *lbws {
  return &lbws{
     deployer: apprunner.New(sess) // Previously ecs.New(sess). 
  }
}
```


This is easy and clean. However, consider if we didn't have `internal/aws` as a layer between `ask-sdk-go` and `cli`.......  we'd have a lot of code changes all over the place:
```go
// In cli package
import "github.com/aws/aws-sdk-go/service/apprunner" // changed from import "github.com/aws/aws-sdk-go/service/ecs" 

type  svcUpdator interface {
   DeployService(*apprunner.DeployServiceInput) error // changed from UpdateService(*ecs.UpdateServiceInput)
}

type lbws struct {
   updator svcUpdator
}

func NewLBWS() *lbws {
  return &lbws{
     deployer: apprunner.New(sess) // changed from ecs.New(sess)
  }
}

func (lbws *lbws) Execute() error {
   if err := lbws.deployer.DeployService(&apprunner.DeployServiceInput); err != nil {  // changed from  lbws.deployer.UpdateService(&ecs.UpdateServiceInput)
     // A bunch of error handling specific to apprunner.DeployService // changed from "a bunch of error handling specific to ecs.UpdateService
   }
}
```


Hope this helps!

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func (o *localRunOpts) getDecryptedSecretManagerSecrets(awsSession *session.Session) ([]secretManager, error) {
+	secretsManagerClient := secretsmanager.New(awsSession)
+	var secretStruct secretManager
+	var SecretManagerSecrets []secretManager
+	input := &secretsmanager.ListSecretsInput{}
+	err := secretsManagerClient.ListSecretsPages(input, func(page *secretsmanager.ListSecretsOutput, lastPage bool) bool {
+		for _, secret := range page.SecretList {
+			secretName := *secret.Name
```
Lou1415926: Same comment - can we use `aws.StringValue()`?

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+			in := &secretsmanager.GetSecretValueInput{
+				SecretId:     aws.String(secretName),
+				VersionStage: aws.String("AWSCURRENT"),
+			}
+			result, _ := secretsManagerClient.GetSecretValue(in)
```
Lou1415926: why do we swallow the second return value?

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+
+			secretValue := aws.StringValue(result.SecretString)
+
+			_ = json.Unmarshal([]byte(secretValue), &secretStruct)
```
Lou1415926: why do we swallow the error here? If we _know for sure_ that there won't be an unmarshal error, can we have a comment here that says so 🥺

Lou1415926: In addition, I think we should initialize a new `secretStruct` for individual secrets instead of reusing the same one from all secrets. 

We don't want this to happen:
```go
// say secret1's JSON string looks like  `{"name": "firstsecret", "value": "1"}`
// and secret2's JSON string looks like   `{"name": "secondsecret"}`
var secretStruct secretManager
_ = json.Unmarshal([]byte(secret1), &secretStruct)
// secretStruct is {Name: "firstsecret", Value: "1"}

_ = json.Unmarshal([]byte(secret2), &secretStruct)
// secretStruct is {Name: "secondsecret", Value: "1"} but it should be {Name: "secondsecret", Value: ""}
```

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+	"github.com/aws/copilot-cli/internal/pkg/workspace"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+)
+
+type ecsLocalClient interface {
+	TaskDefinition(app, env, svc string) (*awsecs.TaskDefinition, error)
+}
+
+type secretManager struct {
```
Lou1415926: This looks very specific to RDS database secret JSON string. I don't think _all_ secretsmanager secret look like this 👀

internal/pkg/cli/local_run.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	return SSMSecretsList, nil
+}
+
+func (o *localRunOpts) getDecryptedSecretManagerSecrets(awsSession *session.Session) ([]secretManager, error) {
+	secretsManagerClient := secretsmanager.New(awsSession)
+	var secretStruct secretManager
+	var SecretManagerSecrets []secretManager
+	input := &secretsmanager.ListSecretsInput{}
+	err := secretsManagerClient.ListSecretsPages(input, func(page *secretsmanager.ListSecretsOutput, lastPage bool) bool {
```
Lou1415926: Just a question, not a request! Is there a specific reason why we prefer `ListSecretsPages` over `ListSecrets` with pagination token?

internal/pkg/cli/local_run.go
Varun359: yes, you are right

internal/pkg/cli/local_run.go
Varun359: The answer to your first question is yes we need to get the ExecutionRole later but for now we are using the user's current credentials.
the answer to your next question is yes, we can use the `defaultSession`. I have created on extra, looks like there is no need for new `awsSession`.

internal/pkg/cli/local_run.go
Varun359: Correct me if I am wrong. I just thought if there are many secrets using `ListSecretsPages` would just simplify the code which handles pagination internally, reducing the complexity of managing pagination tokens.

internal/pkg/cli/local_run.go
Varun359: Makes sense. This is some explanation👀.

internal/pkg/cli/local_run.go
Lou1415926: That SGTM 👍🏼 . Also I just noticed that we are listing all the secrets in the current account + region. Should we just get the value of the secret that found from task definition (like what you did in `getDecryptedSSMSecrets`)

internal/pkg/aws/ssm/ssm.go
```diff
@@ -132,3 +133,16 @@ func convertTags(inTags map[string]string) []*ssm.Tag {
...
+
+// GetSecretValue retrieves the value of a secret parameter from AWS Systems Manager Parameter Store.
+// It takes the name of the parameter as input and returns the corresponding value as a string.
+func (s *SSM) GetSecretValue(name string) (string, error) {
+	resp, err := s.client.GetParameter(&ssm.GetParameterInput{
+		Name:           aws.String(name),
+		WithDecryption: aws.Bool(true),
+	})
+	if err != nil {
+		return "", fmt.Errorf("get secret from SSM parameter %s: %w", name, err)
```
iamhopaul123:
```suggestion
		return "", fmt.Errorf("get parameter %s from SSM with decryption: %w", name, err)
```

internal/pkg/aws/secretsmanager/secretsmanager.go
```diff
@@ -114,6 +115,19 @@ func (s *SecretsManager) DescribeSecret(secretName string) (*DescribeSecretOutpu
 	}, nil
 }
 
+// GetSecretValue retrieves the value of a secret from AWS Secrets Manager.
+// It takes the name of the secret as input and returns the corresponding value as a string.
+func (s *SecretsManager) GetSecretValue(name string) (string, error) {
+	resp, err := s.secretsManager.GetSecretValue(&secretsmanager.GetSecretValueInput{
+		SecretId:     aws.String(name),
+		VersionStage: aws.String("AWSCURRENT"),
```
iamhopaul123: We probably don't need this.
```suggestion
```
> If you don't specify either a VersionStage or VersionId, Secrets Manager returns the AWSCURRENT version.

internal/pkg/aws/secretsmanager/secretsmanager.go
```diff
@@ -114,6 +115,19 @@ func (s *SecretsManager) DescribeSecret(secretName string) (*DescribeSecretOutpu
...
+	})
+	if err != nil {
+		return "", fmt.Errorf("get secret from secrets manager %v: %w", aws.String(name), err)
```
iamhopaul123:
```suggestion
		return "", fmt.Errorf("get secret %s from secrets manager: %w", name, err)
```

internal/pkg/cli/local_run.go
```diff
@@ -5,15 +5,38 @@ package cli
...
 )
 
+type ecsLocalClient interface {
+	TaskDefinition(app, env, svc string) (*awsecs.TaskDefinition, error)
+}
+
+type decryptedSecret struct {
+	Name  string
+	Value string
+}
```
Lou1415926: What do you think just naming it `envVar` ? A "decrpytedSecret" is basically a env var 👀  Later when we read the `EnvironmentVariables` from the task def, we'd be able to reuse the same struct.
```suggestion
type envVar struct {
	Name  string
	Value string
}
```

internal/pkg/aws/ssm/ssm.go
```diff
@@ -132,3 +133,16 @@ func convertTags(inTags map[string]string) []*ssm.Tag {
 	}
 	return tags
 }
+
+// GetSecretValue retrieves the value of a secret parameter from AWS Systems Manager Parameter Store.
```
Lou1415926: nit:

Because an SSM parameter is not always a "secret" (in SSM's terminology). An SSM param could be a `PlainString`, which is not encrypted; it could be a `SecretString`, which _is_ encrypted (and can be thought of as a "secret" in SSM's terminology).



```suggestion
// GetSecretValue retrieves the value of a parameter from AWS Systems Manager Parameter Store.
```

internal/pkg/cli/interfaces.go
```diff
@@ -664,6 +671,10 @@ type secretPutter interface {
 	PutSecret(in ssm.PutSecretInput) (*ssm.PutSecretOutput, error)
 }
 
+type secretDecrypter interface {
+	GetSecretValue(s string) (string, error)
+}
+
```
Lou1415926: I think we need only one of the `secretDecrypter` and `secretGetter` interfaces
```suggestion
type secretDecrypter interface {
	GetSecretValue(s string) (string, error)
}

```

internal/pkg/cli/local_run.go
```diff
@@ -23,7 +46,17 @@ type localRunVars struct {
...
-	store store
+	deployedWkld       []string
+	wkldDeployedToEnvs map[string][]string
+	store              store
+	ws                 wsWlDirReader
+	prompt             prompter
+	ecsLocalClient     ecsLocalClient
+	sess               *session.Session
+	secretsManager     secretsManager
+	ssmDecrypter       secretDecrypter
```
Lou1415926: These two can be instantiated differently (one sm, the other ssm), but they could be the same interface (either `secretGetter` or `secretDecrypter`)

internal/pkg/cli/local_run.go
```diff
@@ -56,15 +119,182 @@ func (o *localRunOpts) Validate() error {
...
+	}
+
+	return nil
+}
+
+func (o *localRunOpts) getDecryptedSSMSecrets(taskDef *awsecs.TaskDefinition, awsSession *session.Session) ([]decryptedSecret, error) {
+	var ssmSecrets []decryptedSecret
+	secrets := taskDef.Secrets()
+	for _, secret := range secrets {
+		if !strings.HasPrefix(secret.ValueFrom, "arn:aws:secretsmanager:") {
```
Lou1415926: did we mean `ssm` here 👀 ?

internal/pkg/cli/local_run.go
```diff
@@ -56,15 +119,182 @@ func (o *localRunOpts) Validate() error {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (o *localRunOpts) getDecryptedSSMSecrets(taskDef *awsecs.TaskDefinition, awsSession *session.Session) ([]decryptedSecret, error) {
+	var ssmSecrets []decryptedSecret
+	secrets := taskDef.Secrets()
```
Lou1415926: What do you think of these signatures:
```go
func (o *localRunOpts) decryptedSSM([]*ecs.ContainerSecret) ([]envVar, error) { }
func (o *localRunOpts) decryptedSecretsManager([]*ecs.ContainerSecret) ([]envVar, error) { }
func (o *localRunOpts) Execute() error {
    secrets := taskDef.Secrets()
    _, err := o.decryptedSSM(secrets)
   _, err := o.decryptedSecretsManager(secrets)
}
```

internal/pkg/cli/local_run.go
Lou1415926: Let me know what you think of this!
```go
// In internal/aws/secretsmanager/secretsmanager.go
const namespace = "secretsmanager"

// IsService returns true if the given ARN is a secretsmanager ARN.
func IsService(arn string) bool  // Use arn.Parse Ref: https://docs.aws.amazon.com/sdk-for-go/api/aws/arn/#Parse to check whether the arn is of secretsmanager. 


// In internal/aws/ssm/ssm.go
const namespace = "ssm"

// IsService returns true if the given ARN is an ssm ARN.
func IsService(arn string) bool
```

This lets individual `aws/x` package manages the service namespace that they should be looking for; using `arn.Parse` makes the code more robust - checking prefix only works for the "aws" partition, but there are also china partition (looks like "arn:aws-cn:ssm") and govcloud partition ("arn:aws-gov:ssm")

internal/pkg/cli/local_run.go
```diff
@@ -56,15 +119,182 @@ func (o *localRunOpts) Validate() error {
...
+		if !strings.HasPrefix(secret.ValueFrom, "arn:aws:secretsmanager:") {
+			secretValue, err := o.ssmDecrypter.GetSecretValue(secret.ValueFrom)
+			if err != nil {
+				return nil, err
+			}
+			secretInfo := decryptedSecret{
+				Name:  secret.Name,
+				Value: secretValue,
+			}
+			ssmSecrets = append(ssmSecrets, secretInfo)
```
```suggestion
			ssmSecrets = append(ssmSecrets, envVar{
				Name:  secret.Name,
				Value: secretValue,
			})
```

Lou1415926 (commented):
This is so much more nicer!! Thank you!

Can we rebase the PR so that this PR only contains the relevant logic?

internal/pkg/cli/local_run.go
```diff
@@ -56,15 +119,182 @@ func (o *localRunOpts) Validate() error {
...
+
+	_, err = o.getDecryptedSecretManagerSecrets(taskDef, o.sess)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (o *localRunOpts) getDecryptedSSMSecrets(taskDef *awsecs.TaskDefinition, awsSession *session.Session) ([]decryptedSecret, error) {
```
iamhopaul123: I feel like this method shouldn't be in this file because it's not specific to local run. Can we move it to the `ecs` pkg (not the aws ecs one)? 

```go
func DecryptedSSMSecrets(secrets, secretGetter) ([]decryptedSecret, error)
```

PR #5097, opened 2023-07-18
chore: Enable ctrl-c for all workloads

KollaAdithya:


internal/pkg/cli/deploy/lbws.go
```diff
@@ -128,15 +128,20 @@ func (d *lbWebSvcDeployer) GenerateCloudFormationTemplate(in *GenerateCloudForma
...
 	}
-	if err := d.deploy(in.Options, *stackConfigOutput); err != nil {
+	deployOut, err := d.deploy(in.Options, *stackConfigOutput)
+	if err != nil {
 		return nil, err
 	}
-	return noopActionRecommender{}, nil
+	return &DeployWorkloadOutput{
+		IsWkldDeleted:        deployOut.IsWkldDeleted,
+		IsWkldUpdateCanceled: deployOut.IsWkldUpdateCanceled,
```
iamhopaul123: It looks quite strange to me that `DeployWorkloadOutput` has `IsWkldDeleted` and `IsWkldUpdateCanceled` because it's supposed to deploy the workload. Instead of modifying the signature of `DeployWorkload`, what do you think using custom error type (returned by `WaitForSignalAndHandleInterrupt` instead of multiple return values) to indicate if wkld is interrupted to be deleted or update canceled?

Potentially we can implement recommendedAction for that custom error too.

internal/pkg/term/signal/signal_test.go
```diff
@@ -34,11 +33,11 @@ func TestSignal(t *testing.T) {
 
 			// THEN
 			for _, wantedSignal := range tc.inSignals {
-				if err := syscall.Kill(syscall.Getpid(), wantedSignal.(syscall.Signal)); err != nil {
-					require.Error(t, err)
+				err := syscall.Kill(syscall.Getpid(), wantedSignal.(syscall.Signal))
+				if err == nil {
```
iamhopaul123: Do we not check if the real `err` is what we want?

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -241,6 +252,52 @@ func (cf CloudFormation) Template(stackName string) (string, error) {
...
+		select {
+		case <-cf.signalClient.NotifySignals():
+			cancel()
+			cf.signalClient.StopCatchSignals()
+			stackDescr, err := cf.cfnClient.Describe(stackName)
+			if err != nil {
+				return isDeleted, isUpdateCanceled, err
+			}
+			if aws.StringValue(stackDescr.StackStatus) == sdkcloudformation.StackStatusCreateInProgress {
+				log.Infof(color.Red.Sprintf("Received Interrupt for Ctrl-C.\nPressing Ctrl-C again will exit immediately but the deletion of stack %s still happens\n", stackName))
```
iamhopaul123: Can we use "\`" and "\`" for better readability? Also maybe `Info` in enough?

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -241,6 +252,52 @@ func (cf CloudFormation) Template(stackName string) (string, error) {
...
+// Returns stack deletion status, update cancellation status, and an error, if any.
+func (cf CloudFormation) WaitForSignalAndHandleInterrupt(ctx context.Context, cancel context.CancelFunc, stackName string) (isDeleted bool, isUpdateCanceled bool, err error) {
+	for {
+		select {
+		case <-cf.signalClient.NotifySignals():
+			cancel()
+			cf.signalClient.StopCatchSignals()
+			stackDescr, err := cf.cfnClient.Describe(stackName)
+			if err != nil {
+				return isDeleted, isUpdateCanceled, err
```
iamhopaul123: Do we need to wrap this error?

internal/pkg/term/signal/signal.go
```diff
@@ -0,0 +1,36 @@
...
+// Package signal provides a way to capture and handle operating system signals.
+package signal
+
+import (
+	"os"
+	"os/signal"
+)
+
+// Signal represents a signal handler for capturing and handling operating system signals.
+type Signal struct {
```
iamhopaul123: I am still not convinced that we need this package. It looks pretty straightforward to me if we just use `os.Signal` directly.

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -241,6 +252,52 @@ func (cf CloudFormation) Template(stackName string) (string, error) {
...
+			}
+			if aws.StringValue(stackDescr.StackStatus) == sdkcloudformation.StackStatusCreateInProgress {
```
iamhopaul123: Would it be a better idea to use switch case for `aws.StringValue(stackDescr.StackStatus)`?

iamhopaul123 (commented):
This is really amazing! One thing that can be improved is to make this PR smaller by splitting into `cancelUpdateAndRender` impl, `WaitForSignalAndHandleInterrupt` impl, and `deployAndHandleInterrupt` (including everything else).

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -191,6 +200,7 @@ type CloudFormation struct {
 	regionalECRClient func(region string) imageRemover
 	region            string
 	console           progress.FileWriter
+	signalClient      signalClient
```
Lou1415926: nit: we can probably rename this client because we only cares about sigInt right now.
```suggestion
	sigInt      signalClient
```
The other codes will read ⬇️  and it'd be more obvious why we `cancel()` upon receiving a signal
```go
case <-cf.sigInt.NotifySignals():
cancel()
cf.sigInt.StopCatchSignals()
```

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -241,6 +252,52 @@ func (cf CloudFormation) Template(stackName string) (string, error) {
...
+						return cf.cfnClient.CancelUpdateStack(stackName)
+					},
+				}); err != nil {
+					return isDeleted, isUpdateCanceled, err
+				}
+				isUpdateCanceled = true
+				return isDeleted, isUpdateCanceled, nil
+			}
+			return isDeleted, isUpdateCanceled, nil
+		case <-ctx.Done():
```
Lou1415926: Do we need to stop listening for the signals in this case?

internal/pkg/cli/deploy/workload.go
```diff
@@ -360,6 +368,38 @@ func (d *workloadDeployer) generateCloudFormationTemplate(conf stackSerializer)
...
+		return nil
+	}
+	waitForSignalAndHandleInterrupt := func() error {
+		isDeleted, isUpdateCanceled, err := d.deployer.WaitForSignalAndHandleInterrupt(ctx, cancel, conf.StackName())
+		out.IsWkldDeleted = isDeleted
+		out.IsWkldUpdateCanceled = isUpdateCanceled
+		if err != nil {
+			return err
+		}
+		return nil
```
Lou1415926: I have a question 🤚🏼 , not a request! Why do you prefer putting `WaitForSignal` in `cloudformation` package? An alternative I can think of is having the "wait" logic here:

```go
// Here
waitForSignal := func() error {
    sigInt  ch os.Signal
    signal.Notify(sigInt, os.SIGINT)
    select {
		case <-sigInt:
                    log.Infof("Received SIGINT.....")
                    cancel()
                    err := d.deployer.CancelUpdate()
                 case <- ctx.Done():
                     signal.Stop(sigInt)
                     close(sigInt)
     }
}
```

KollaAdithya: If I want to enable to `ctrl-c` for `env deploy`  then I think 
1. I need to replicate the same logic for the env deployer.
2. And also in order to know the status of the cloudformation stack, I need to have public method `DescribeStack` that i can call `d.deployer.DescribeStack`.
3. And also in order to mock the various api calls such as `DescribeEvents` and `TemplateBodyFromChangeSet` and so on ..
All these methods are not  available in the `deploy` pkg.

internal/pkg/term/signal/signal.go
KollaAdithya: I will just remove this package and just use `os.Signal`.

internal/pkg/cli/deploy/workload.go
Lou1415926: > I need to replicate the same logic for the env deployer.

I think arguably, this can be mitigated by ⬇️ 
```go
func waitForSignal(handler func() error ) error {
     case <-sigInt:
            log.Infof("Received SIGINT.....")
            cancel()
            if err := handler(); err != nil {...}
   case <- ctx.Done():
       signal.Stop(sigInt)
       close(sigInt)
}

// In workload deployer 
g.Go(deployService)
g.Go(waitForSignal(func() error {
    err := d.deployer.CancelUpdate()
}))

// In env deployer 
g.Go(deployEnv)
g.Go(waitForSignal(func() error {
    err := d.deployer.CancelEnvironmentUpdate()
}))
```

> And also in order to know the status of the cloudformation stack, I need to have public method DescribeStack that i can call d.deployer.DescribeStack.'

This can & probably should be inside of `CancelUpdate` right?
```go
// In deploy/cloudformation/cloudformation.go
func (cf CloudFormation) CancelUpdate() error {
   stackDescr, err := cf.cfnClient.DescribeStack()
   switch  aws.StringValue(stackDescr.StackStatus) {
   case sdkcloudformation.StackStatusCreateInProgress:
   case sdkcloudformation.StackStatusUpdateInProgress 
   }
}
```

KollaAdithya: This is really good. Yeah I think this is a better way of having the abstractions than what i have before.

PR #5099, opened 2023-07-18
refactor: separate pipeline CFN files and standardize its parsing

Lou1415926:


internal/pkg/deploy/cloudformation/stack/embed.go
```diff
@@ -41,6 +41,10 @@ type envReadParser interface {
 	ParseEnvBootstrap(data *template.EnvOpts, options ...template.ParseOption) (*template.Content, error)
 }
 
+type pipelineParser interface {
```
KollaAdithya, line 44: Just a question
Do we need to embed this `pipelineParser` interface in [`embedFS`](https://github.com/aws/copilot-cli/pull/5099/files#diff-ea5b99d39bfb17792d9011cfc84e35b1a0203546608fdff1a5e16065f85ac971R49) interface ?

iamhopaul123: It seems like `embedFS` is only useful for unit tests. Maybe thinking about moving the interface to the test files?

internal/pkg/deploy/cloudformation/stack/pipeline.go
```diff
@@ -35,23 +31,15 @@ func (p *pipelineStackConfig) StackName() string {
...
-	content, err := p.parser.Parse(pipelineCfnTemplatePath, p, template.WithFuncs(cfTemplateFunctions), template.WithFuncs(map[string]interface{}{
-		"isCodeStarConnection": func(source interface{}) bool {
-			type connectionName interface {
-				ConnectionName() (string, error)
-			}
-			_, ok := source.(connectionName)
-			return ok
-		},
-	}))
+	content, err := p.parser.ParsePipeline(p)
```
iamhopaul123, line 34: Awesome!

iamhopaul123 (approved):

internal/pkg/deploy/cloudformation/stack/embed.go
Lou1415926: > Do we need to embed this pipelineParser interface in [embedFS](https://github.com/aws/copilot-cli/pull/5099/files#diff-ea5b99d39bfb17792d9011cfc84e35b1a0203546608fdff1a5e16065f85ac971R49) interface ?

We _could_ because theoretically pipelineParser is a part of the embedded FS that reads & parses the templates.... However, on the other hand, in reality we need `pipelineParser` in `embedFS` only if pipeline has custom resources.

I'm fine either way. If y'all don't have strong opinion on this, I'll leave `pipelineParser` outside of `embedFS` until we have custom resources for pipeline.

> Maybe thinking about moving the interface to the test files?

The interface _is_ used in the real code though:
https://github.com/aws/copilot-cli/blob/c7f476bbe7d7ff6c3396206db4647a6d7a545511/internal/pkg/deploy/cloudformation/stack/embed.go#L56

This keeps both `realEmbedFs` and `fs`as type of `embedFS` (the interface that is mock-able) instead of `*template.Template` (the real struct that can't be mocked).

I don't think the code will build if we move the interface definition to _test file : (

KollaAdithya (approved):
lgtm

PR #5103, opened 2023-07-19
chore: render events of cancel update stack

KollaAdithya:
Based on the feedback I am splitting PR #5097 into three parts
- [x]  render events for cancel update stack.
- [ ]  Wait for SIGINT signal and handle that interrupt accordingly.
- [ ]  Apply the `ctrl-c` for all the workloads.


KollaAdithya:
I give an thought about splitting PR #5097 but since the static checks are failing for this unused functions.

PR #5106, opened 2023-07-19
chore: use + for prerelease number

iamhopaul123:
Problem is the new CLI is not technically "new" in terms of the version number. Adding `allow-downgrade` flag to override.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ GOBIN=${ROOT_SRC_DIR}/bin/tools
 COVERAGE=coverage.out
 
 DESTINATION=./bin/local/${BINARY_NAME}
-VERSION=$(shell git describe --always --tags)
+VERSION=$(shell git describe --always --tags | sed 's/-/+/')
 
 BINARY_S3_BUCKET_PATH=https://ecs-cli-v2-release.s3.amazonaws.com
 


huanjani (approved):

Lou1415926 (approved):
An alternative is to use this function https://pkg.go.dev/golang.org/x/mod/semver#Prerelease to clean up the version numbers before calling `semver.Compare`. Or just deem `pre-release` as larger than version 🤔 

I feel like the root of the issue is on our versioning practice... `v1.29.0-1-g482e3147` indicates that this is a pre-v1.29.0 release (therefore it is smaller than v1.29.0). However, in reality, this is a pre-v1.30.0 release, because v1.29.0 is already released 😂 If this had been tagged with `v1.30.0-g482e3147`, I bet this won't trigger the "downgrade" error.

I do wonder though, why was this error not triggered earlier? Do you know if this is flaky? 

Anyway, I think it's fine to add these flags just to get around this issue, but I think this issue probably indicates a slightly larger problem to solve, and I worry that adding this flag here (which doesn't make sense at the first glance - people have to take a look at git blame and find this PR to understand the background story of adding this flag 😂 ) eventually accumulates into tech debt.

iamhopaul123:
> An alternative is to use this function https://pkg.go.dev/golang.org/x/mod/semver#Prerelease to clean up the version numbers before calling `semver.Compare`. Or just deem `pre-release` as larger than version 🤔
> 
> I feel like the root of the issue is on our versioning practice... `v1.29.0-1-g482e3147` indicates that this is a pre-v1.29.0 release (therefore it is smaller than v1.29.0). However, in reality, this is a pre-v1.30.0 release, because v1.29.0 is already released 😂 If this had been tagged with `v1.30.0-g482e3147`, I bet this won't trigger the "downgrade" error.
> 
> I do wonder though, why was this error not triggered earlier? Do you know if this is flaky?
> 
> Anyway, I think it's fine to add these flags just to get around this issue, but I think this issue probably indicates a slightly larger problem to solve, and I worry that adding this flag here (which doesn't make sense at the first glance - people have to take a look at git blame and find this PR to understand the background story of adding this flag 😂 ) eventually accumulates into tech debt.

I think I agree with most of it. The only problem is I don't want to update our codebase because of the test. The culprit I think it's `git describe` returns "v1.30.0-1-g482e3147" instead of "v1.30.0+1-g482e3147" and the latter one makes more sense. I think we should maybe investigate how to generate a local binary version number that makes more sense.

PR #5109, opened 2023-07-20
chore: add cfn snippets for bare-boned pre/post deployment actions

Lou1415926:


internal/pkg/deploy/pipeline.go
```diff
@@ -729,3 +741,15 @@ func (a *TestCommandsAction) Name() string {
 func (a *TestCommandsAction) Commands() []string {
 	return a.commands
 }
+
+// Action represents a CodePipeline action.
+type Action struct {
+	action
```
iamhopaul123: Why do we need both `Action` and `action`? Can we consolidate the logic to one struct?

internal/pkg/template/templates/cicd/partials/role-policy-document.yml
```diff
@@ -0,0 +1,77 @@
...
+      - codebuild:BatchPutCodeCoverages
+    Resource: !Sub arn:aws:codebuild:${AWS::Region}:${AWS::AccountId}:report-group/pipeline-{{$.AppName}}-*
+  - Effect: Allow
+    Action:
+      - s3:PutObject
+      - s3:GetObject
+      - s3:GetObjectVersion
+    # TODO: This might not be necessary. We may only need the bucket
+    # that is in the same region as the pipeline.
+    # Loop through all the artifact buckets created in the stackset
```
iamhopaul123, line 18: We might need to prioritize TODOs in this file for appsec review.

internal/pkg/template/templates/cicd/partials/actions.yml
```diff
@@ -0,0 +1,67 @@
...
+{{- range $action := $stage.PostDeployments}}
+Post{{logicalIDSafe $stage.Name}}DeploymentAction{{logicalIDSafe $action.Name}}BuildProjectRole:
+  Type: AWS::IAM::Role
+  Properties:
+    Path: /
+  {{ include "role-config" $ | indent 4}}
+    {{- if $.PermissionsBoundary }}
+    PermissionsBoundary: !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:policy/{{$.PermissionsBoundary}}'
+    {{- end }}
+Policies:
```
iamhopaul123: Indentation is not right?

internal/pkg/template/templates/cicd/partials/actions.yml
```diff
@@ -0,0 +1,67 @@
...
+    ServiceRole: !GetAtt Pre{{logicalIDSafe $stage.Name}}DeploymentAction{{logicalIDSafe $action.Name}}BuildProjectRole.Arn
+    Artifacts:
+      Type: CODEPIPELINE
+{{ include "action-config" $action | indent 4}}
+{{- end}}
+{{- end}}
+
+{{- range $stage := .Stages}}
+{{- range $action := $stage.PostDeployments}}
+Post{{logicalIDSafe $stage.Name}}DeploymentAction{{logicalIDSafe $action.Name}}BuildProjectRole:
```
iamhopaul123, line 37: It seems like the pre and post action build project role are pretty much the same. Can we consolidate them?

internal/pkg/template/templates/cicd/partials/action-config.yml
```diff
@@ -0,0 +1,23 @@
+{{- if or (eq .Build.EnvironmentType "LINUX_CONTAINER") (eq .Build.EnvironmentType "ARM_CONTAINER")}}
```
iamhopaul123, line 1: Can we also reuse for the current pipeline CFN build?

internal/pkg/template/templates/cicd/partials/role-policy-document.yml
```diff
@@ -0,0 +1,77 @@
+Version: '2012-10-17'
```
iamhopaul123, line 1: Is this policy document the same (is there any new thing we added here)?

internal/pkg/template/templates/cicd/pipeline_cfn.yml
```diff
@@ -236,6 +238,21 @@ Resources:
                 Provider: Manual
               RunOrder: {{$stage.Approval.RunOrder}}
             {{- end}}
+            {{- range $action := $stage.PreDeployments }}
```
iamhopaul123, line 241: Why do we only add preDeployments here? What about postDeployments?

internal/pkg/template/templates/cicd/partials/role-policy-document.yml
Lou1415926: It's the same as before.

internal/pkg/template/templates/cicd/partials/actions.yml
Lou1415926: Do you mean have another partial "action-role.yml"? And in here we'd have ⬇️ ? 🤔 
```yaml
Ppre{{logicalIDSafe $stage.Name}}DeploymentAction{{logicalIDSafe $action.Name}}BuildProjectRole:
  {{- include "action-role" $stage | indent 2}}
Post{{logicalIDSafe $stage.Name}}DeploymentAction{{logicalIDSafe $action.Name}}BuildProjectRole:
  {{- include "action-role" $stage | indent 2}} 
```

iamhopaul123: No i meant "do we need pre and post to be separate role?" or can we create one role per action for both pre and post.

internal/pkg/template/templates/cicd/partials/action-config.yml
Lou1415926: 👍🏼  Just to confirm, do you mean having the current pipeline CFN build project (i.e. build-action.yml) to include this snippet?

```yaml
BuildProject:
  Properties:
     {{- include "action-config" . | indent 6}}
```

I will make ⬆️  change, but i want to make sure that we are on the same page!

internal/pkg/deploy/pipeline.go
Lou1415926: Does this make more sense to you?
```go
// Step represents a CodePipeline action backed by a CodeBuild project.
type Step struct {
	action
	Build
	name string
}
```

I think that was because of a bad naming. The struct named `Action` was actually a specific "type" of action, it is not equivalent to the generic `action` struct.

internal/pkg/template/templates/cicd/partials/actions.yml
Lou1415926: Ah I see! Umm I feel like different action may require different permissions. For example, a db-migration action might need certain RDS permissions, or to assume a role that has the RDS permissions  (this is not always the EnvManagerRole, as they _could_ have an RDS that's outside of Copilot)...

> can we create one role per action for both pre and post.

Pre/post actions do not come in pair, i.e. I could have one pre-action that does db migration, and one post-action that does integration test - these two actions are independent of each other. 

There are several options here:
1. Share a role among all pre/post actions for now, since all of them require the same permissions - for now. Later, when people can customize the permissions, we split them up. 
2. Do not share the same role.

I ended up choosing the second, because I think even though there isn't a great manifest-based way to customize the role yet, they could easily modify  the role via. YAML patch/CDK override. With the first approach, they'd have create a new role, customize it, and change the role reference from the project.

What do you think!

internal/pkg/template/templates/cicd/pipeline_cfn.yml
huanjani: In line 257 👍🏼 .

huanjani (approved):
LGTM!

internal/pkg/deploy/pipeline.go
```diff
@@ -729,3 +739,15 @@ func (a *TestCommandsAction) Name() string {
...
+// Step represents a CodePipeline action backed by a CodeBuild project.
+type Step struct {
+	action
+	Build
+	name string
+}
+
+// Name returns the name of the action.
+func (a *Step) Name() string {
+	return a.name
```
huanjani: nit:
```suggestion
func (s *Step) Name() string {
	return s.name
```

KollaAdithya (approved):

PR #5111, opened 2023-07-21
fix: make sure copilot URL in buildspec is downloadable

bvtujo:
This resolves a problem where the pipeline buildspecs do not generate a correct download URL for the latest copilot version.

This is because '+' characters are URL-encoded to "%2B". 
When the pipeline tries to download an in-between version (v1.29.0+5-g74ef584b for example), it fails because of a wget error that 
the specified version is not found in the bucket. There's no object at the url https://ecs-cli-v2-release.s3.us-west-2.amazonaws.com/copilot-linux-v1.29.0+5-g74ef584b. 


iamhopaul123 (approved):
🙏

KollaAdithya (approved):

PR #5112, opened 2023-07-21
chore: enable ctrl-c only after changeset is created and executed

KollaAdithya:
Even though this PR looks big, half of the changes are just unit tests.

Better solution than #5097 

**Old Approach:**
I closed PR 5097. In the previous PR #5097, there is a drawback with the logic when users initiated a ctrl-c interruption during the process of `proposing infrastructure changes`. The issue was that the interruption signal (SIGINT) was being captured even before the changeset is created and executed. As a result, when the user pressed ctrl-c during `proposing infrastructure changes`, we display  a message like `Received interrupt for ctrl-c`; However, despite the interruption, the code continued to create and execute the changeset, leading to the stack creation process continuing even after the user intended to stop it.

**New Approach:**
To address the  issue, this PR had a better solution for handling ctrl-c interruptions during the process of proposing infrastructure changes. Listening to the SIGINT signal is performed only after the changeset is created and executed. The necessary changes to achieve this have been made by relocating the complete logic for waiting and handling the signal to the `deploy/cloudformation` package.


iamhopaul123 (approved):
Awesome! Just to double check: do we need to update the deploy and jobdeploy with the same thing?

KollaAdithya:
With this PR we are already enabling `ctrl-c` for both `svc deploy` and `job deploy` under the hood `ctrl-c` will automatically apply to  `copilot deploy`

internal/pkg/cli/deploy/rdws.go
```diff
@@ -112,12 +112,17 @@ func (d *rdwsDeployer) DeployWorkload(in *DeployWorkloadInput) (ActionRecommende
 	if err != nil {
 		return nil, err
 	}
+	recommendations := &rdwsDeployOutput{
+		rdwsAlias: stackConfigOutput.rdSvcAlias,
+	}
 	if err := d.deploy(in.Options, stackConfigOutput.svcStackConfigurationOutput); err != nil {
+		var errStackUpdateCanceledOnInterrupt *cloudformation.ErrStackUpdateCanceledOnInterrupt
+		if errors.As(err, &errStackUpdateCanceledOnInterrupt) {
+			return recommendations, err
```
dannyrandall: do we want to return the same recommendations if the stack update was cancelled? 🤔

internal/pkg/cli/deploy/worker.go
```diff
@@ -129,12 +130,17 @@ func (d *workerSvcDeployer) DeployWorkload(in *DeployWorkloadInput) (ActionRecom
 	if err != nil {
 		return nil, err
 	}
+	recommendations := &workerSvcDeployOutput{
+		subs: stackConfigOutput.subscriptions,
+	}
 	if err := d.deploy(in.Options, stackConfigOutput.svcStackConfigurationOutput); err != nil {
+		var errStackUpdateCanceledOnInterrupt *cloudformation.ErrStackUpdateCanceledOnInterrupt
+		if errors.As(err, &errStackUpdateCanceledOnInterrupt) {
+			return recommendations, err
```
dannyrandall: same question as rdws

internal/pkg/cli/env_delete_test.go
```diff
@@ -353,7 +354,7 @@ Resources:
 				store := mocks.NewMockenvironmentStore(ctrl)
 				store.EXPECT().ListEnvironments("phonetool").Return([]*config.Environment{
 					&mockEnv,
-					&config.Environment{
+					{
```
dannyrandall, line 357: nice catch😇

internal/pkg/cli/deploy/rdws.go
dannyrandall: / maybe the recommendations depend on if the stack was deleted (first deployment) or cancelled (second+ deployment)

dannyrandall: oh i see that the recommendations aren't returned in `cli/svc_deploy.go` if the stack was deleted. i still wonder if the cancel case should have different recommendations/none at all 🤔. for this case, couldn't the recommendation end up recommending an invalid URL?

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -363,6 +408,128 @@ func (cf CloudFormation) executeAndRenderChangeSet(in *executeAndRenderChangeSet
...
+			cancelFn()
+			stopCatchSignals(sigCh)
+			stackDescr, err := cf.cfnClient.Describe(stackName)
+			if err != nil {
+				return fmt.Errorf("describe stack %s: %w", stackName, err)
+			}
+			switch aws.StringValue(stackDescr.StackStatus) {
+			case sdkcloudformation.StackStatusCreateInProgress:
+				log.Infof(`Received Interrupt for Ctrl-C.
+Pressing Ctrl-C again will exit immediately but the deletion of stack %s still happens
```
dannyrandall:
```suggestion
Pressing Ctrl-C again will exit immediately but the deletion of stack %s will continue
```

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -363,6 +408,128 @@ func (cf CloudFormation) executeAndRenderChangeSet(in *executeAndRenderChangeSet
...
+
+	}
+	waitCtx, cancelWait := context.WithTimeout(context.Background(), waitForStackTimeout)
+	defer cancelWait()
+	g, ctx := errgroup.WithContext(waitCtx)
+	g.Go(in.cancelUpdateFn)
+	renderer, err := cf.createChangeSetRenderer(g, ctx, aws.StringValue(stackDescr.ChangeSetId), in.stackName, in.description, progress.RenderOptions{})
+	if err != nil {
+		return err
+	}
```
dannyrandall, line 476: could we move this below 480? just to keep the format similar
```
g.Go()
g.Go()

something else

g.Wait()
```

internal/pkg/deploy/cloudformation/cloudformation_test.go
```diff
@@ -817,6 +866,393 @@ Resources:
 	require.Contains(t, buf.String(), "A DynamoDB table to store data")
 }
 
+func testDeployWorkload_OnCancelUpdateSuccess(t *testing.T, stackName string, when func(cf CloudFormation) error) {
+	// GIVEN
+	wantedErr := &ErrStackUpdateCanceledOnInterrupt{stackName: stackName}
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	deploymentTime := time.Date(2020, time.November, 23, 18, 0, 0, 0, time.UTC)
```
dannyrandall: why not use `time.Time{}`?

internal/pkg/deploy/cloudformation/cloudformation_test.go
```diff
@@ -817,6 +866,393 @@ Resources:
 	require.Contains(t, buf.String(), "A DynamoDB table to store data")
 }
 
+func testDeployWorkload_OnCancelUpdateSuccess(t *testing.T, stackName string, when func(cf CloudFormation) error) {
```
dannyrandall: seems like a these tests have a lot of duplicated code - is it possible to move that to a helper function?

internal/pkg/deploy/cloudformation/cloudformation.go
KollaAdithya: `renderer` is getting passed as input parameter to one of the go routine though 🤔 
```_, err := progress.Render(ctx, progress.NewTabbedFileWriter(cf.console), renderer)```

So I will move this above both the go routines. It looks something like this.
```
renderer, err := cf.createChangeSetRenderer(g, ctx, aws.StringValue(stackDescr.ChangeSetId), in.stackName, in.description, progress.RenderOptions{})
g.Go()
g.Go(
_, err := progress.Render(ctx, progress.NewTabbedFileWriter(cf.console), renderer)
)

internal/pkg/cli/deploy/rdws.go
KollaAdithya: Yes you are right. if I change alias in the manifest. This will display the wrong url in the recommended actions.
I will go with no recommendations in case of `cancel update stack` as well.

internal/pkg/deploy/cloudformation/cloudformation_test.go
KollaAdithya: addressed the fb in this [commit](https://github.com/aws/copilot-cli/pull/5112/commits/b89e9b2281c85e56ff4e6297cd5be04a9414d503)

internal/pkg/cli/svc_deploy.go
```diff
@@ -308,6 +311,15 @@ func (o *deploySvcOpts) Execute() error {
 		},
 	})
 	if err != nil {
+		if errors.As(err, &errStackDeletedOnInterrupt) {
+			o.noDeploy = true
```
dannyrandall, line 315: i'm probably missing something 😅 - what does this field do?

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -339,10 +357,41 @@ func (cf CloudFormation) executeAndRenderChangeSet(in *executeAndRenderChangeSet
...
+	defer cancel()
+	g.Go(func() error {
+		defer cancel()
+		if err := cf.renderChangeSet(ctx, changeSetID, in); err != nil {
+			if !in.enableInterrupt {
+				return err
+			}
+			if !errors.Is(err, context.Canceled) {
+				return err
+			}
```
dannyrandall: i think this can be simplified to just
```suggestion
			if !errors.Is(err, context.Canceled) {
				return err
			}
```
since the error will never be `context.Canceled` if `in.enableInterrupt` is false

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -339,10 +357,41 @@ func (cf CloudFormation) executeAndRenderChangeSet(in *executeAndRenderChangeSet
...
+	})
+	if in.enableInterrupt {
+		g.Go(func() error {
+			return cf.waitForSignalAndHandleInterrupt(ctx, cancel, sigChannel, in.stackName)
+		})
+	}
+	if err := g.Wait(); err != nil {
+		return err
+	}
+	return nil
```
dannyrandall:
```suggestion
	return g.Wait()
```

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -363,6 +412,128 @@ func (cf CloudFormation) executeAndRenderChangeSet(in *executeAndRenderChangeSet
...
+	if err != nil {
+		return fmt.Errorf("describe stack %s: %w", in.stackName, err)
+	}
+	if stackDescr.ChangeSetId == nil {
+		return fmt.Errorf("ChangeSetID not found for stack %s", in.stackName)
+
+	}
+	waitCtx, cancelWait := context.WithTimeout(context.Background(), waitForStackTimeout)
+	defer cancelWait()
+	g, ctx := errgroup.WithContext(waitCtx)
```
dannyrandall: since we don't use the wait ctx any differently, let's name them the same to avoid unnecessary variables
```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), waitForStackTimeout)
	defer cancel()
	g, ctx := errgroup.WithContext(ctx)
```

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -363,6 +412,128 @@ func (cf CloudFormation) executeAndRenderChangeSet(in *executeAndRenderChangeSet
...
+		_, err := progress.Render(ctx, progress.NewTabbedFileWriter(cf.console), renderer)
+		return err
+	})
+	if err := g.Wait(); err != nil {
+		return err
+	}
+	if err := cf.errOnFailedCancelUpdate(in.stackName); err != nil {
+		return err
+	}
+	return nil
```
dannyrandall:
```suggestion
	return cf.errOnFailedCancelUpdate(in.stackName)
```

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -363,6 +412,128 @@ func (cf CloudFormation) executeAndRenderChangeSet(in *executeAndRenderChangeSet
...
+	close(sigCh)
+}
+
+// ErrStackDeletedOnInterrupt means stack is deleted on interrupt.
+type ErrStackDeletedOnInterrupt struct {
+	stackName string
+}
+
+func (e *ErrStackDeletedOnInterrupt) Error() string {
+	return fmt.Sprintf("stack %s is deleted on interrupt signal", e.stackName)
```
dannyrandall:
```suggestion
	return fmt.Sprintf("stack %s was deleted on interrupt signal", e.stackName)
```

internal/pkg/deploy/cloudformation/cloudformation.go
```diff
@@ -363,6 +412,128 @@ func (cf CloudFormation) executeAndRenderChangeSet(in *executeAndRenderChangeSet
...
+}
+
+// ErrStackUpdateCanceledOnInterrupt means stack update is canceled on interrupt.
+type ErrStackUpdateCanceledOnInterrupt struct {
+	stackName string
+}
+
+func (e *ErrStackUpdateCanceledOnInterrupt) Error() string {
+	return fmt.Sprintf("update for stack %s is canceled on interrupt signal", e.stackName)
```
dannyrandall:
```suggestion
	return fmt.Sprintf("update for stack %s was canceled on interrupt signal", e.stackName)
```

dannyrandall (dismissed):
i love this PR!🚀  just some small nits and one more question 😊

internal/pkg/cli/svc_deploy.go
KollaAdithya: This is mainly used to skip displaying Recommended Actions. 
https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/cli/svc_deploy.go#L331:L333

And also `noDeploy` is also set when we run `copilot svc deploy --diff` to just show the diff and selects `no` for `ContinueWithDeployment` prompt.

https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/cli/svc_deploy.go#L289:L292

internal/pkg/deploy/cloudformation/cloudformation.go
KollaAdithya: sorry 😞 I missed this

internal/pkg/cli/svc_deploy.go
dannyrandall: ah gotcha, thanks 👍

dannyrandall (approved):
great work!

PR #5113, opened 2023-07-21
chore: decrypt secrets from `SSM` and `secrets manager`

Varun359:
This is only for ECS services


internal/pkg/ecs/ecs.go
```diff
@@ -232,6 +244,24 @@ func (c Client) StopDefaultClusterTasks(familyName string) error {
 	return c.ecsClient.StopTasks(taskIDs, ecs.WithStopTaskReason(taskStopReason))
 }
 
+// DecryptedSSMSecrets returns the decrypted parameters from the SSM store.
+func (c Client) DecryptedSSMSecrets(secrets []*awsecs.ContainerSecret, secretGetter *session.Session) ([]EnvVar, error) {
```
iamhopaul123:
```suggestion
func (c Client) DecryptedSSMSecrets(secrets []*awsecs.ContainerSecret) ([]EnvVar, error) {
```

internal/pkg/cli/local_run.go
```diff
@@ -94,11 +110,42 @@ func (o *localRunOpts) validateAndAskWkldEnvName() error {
...
+
+	_, err = o.decryptedSecretManagerSecrets(secrets, o.sess)
+	if err != nil {
+		return err
+	}
 
 	return nil
 }
 
+func (o *localRunOpts) decryptedSecretManagerSecrets(secrets []*awsecs.ContainerSecret, awsSession *session.Session) ([]ecs.EnvVar, error) {
```
iamhopaul123: We can do the same thing to secret manager secrets.

internal/pkg/aws/secretsmanager/secretsmanager.go
```diff
@@ -114,6 +118,30 @@ func (s *SecretsManager) DescribeSecret(secretName string) (*DescribeSecretOutpu
...
+// IsService returns true if the given ARN is secrets manager.
+func (s *SecretsManager) IsService(name string) (bool, error) {
+	parseArn, err := arn.Parse(name)
+	if err != nil {
+		return false, fmt.Errorf("parse secrets manager arn: %w", err)
+	}
+	if parseArn.Service == namespace {
+		return true, nil
+	}
+	return false, nil
```
Lou1415926: nit:
```suggestion
	return parseArn.Service == namespace, nil
```

internal/pkg/aws/ssm/ssm.go
```diff
@@ -61,6 +65,31 @@ func (s *SSM) PutSecret(in PutSecretInput) (*PutSecretOutput, error) {
...
+// IsService returns true if the given ARN is ssm.
+func (s *SSM) IsService(name string) (bool, error) {
+	parseArn, err := arn.Parse(name)
+	if err != nil {
+		return false, fmt.Errorf("parse ssm arn: %w", err)
+	}
+	if parseArn.Service == namespace {
+		return true, nil
+	}
+	return false, nil
```
Lou1415926: nit:
```suggestion
	return parseArn.Service == namespace, nil
```

Lou1415926: hmm looking at how similar this is from the `secretsmanager` one, i'm starting to think maybe we could just do this 😂 
```go
// aws/ssm/ssm.go
const Namespace = "ssm"

// aws/secretsmanager/secretsmanager.go
const Namespace = "secretsmanager"

// cli
func decrypt(name string) {
     parsed, err := arn.Parse(name)
     if err != nil { /* assume it is ssm*/ }}

     if parsed.Service == secretsmanager.Namespace { }
      if parsed.Service == ssm.Namespace { }
}
```

internal/pkg/cli/local_run.go
```diff
@@ -94,7 +107,21 @@ func (o *localRunOpts) validateAndAskWkldEnvName() error {
...
-	//TODO(varun359): Get build information from the manifest and task definition for workloads
+	taskDef, err := o.ecsLocalClient.TaskDefinition(o.appName, o.envName, o.wkldName)
+	if err != nil {
+		return fmt.Errorf("get task definition: %w", err)
+	}
+
+	secrets := taskDef.Secrets()
+	_, err = o.ecsLocalClient.DecryptedSSMSecrets(secrets)
+	if err != nil {
+		return err
```
```suggestion
		return fmt.Errorf("get secret values from ssm: %w: err")
```

internal/pkg/aws/secretsmanager/secretsmanager.go
```diff
@@ -114,6 +118,30 @@ func (s *SecretsManager) DescribeSecret(secretName string) (*DescribeSecretOutpu
...
+		SecretId: aws.String(name),
+	})
+	if err != nil {
+		return "", fmt.Errorf("get secret %s from secrets manager: %w", name, err)
+	}
+	return aws.StringValue(resp.SecretString), nil
+}
+
+// IsService returns true if the given ARN is secrets manager.
+func (s *SecretsManager) IsService(name string) (bool, error) {
```
iamhopaul123: Maybe
```suggestion
func (s *SecretsManager) IsServiceARN(name string) (bool, error) {
```

internal/pkg/cli/local_run.go
```diff
@@ -94,7 +107,21 @@ func (o *localRunOpts) validateAndAskWkldEnvName() error {
 
 // Execute builds and runs the workload images locally.
 func (o *localRunOpts) Execute() error {
-	//TODO(varun359): Get build information from the manifest and task definition for workloads
+	taskDef, err := o.ecsLocalClient.TaskDefinition(o.appName, o.envName, o.wkldName)
```
Lou1415926, line 103: Can we add some unit tests for these logics?

We will need to:
1. Put the `ecsLocalClient` interface to `interface.go`
2. Run `make gen-mocks`
3. Write the unit tests with the generated mocks.

internal/pkg/ecs/ecs.go
```diff
@@ -64,6 +75,8 @@ type ServiceDesc struct {
 // Client retrieves Copilot information from ECS endpoint.
 type Client struct {
 	rgGetter       resourceGetter
+	secretGetter   secretGetter
+	secretManger   secretGetter
```
Lou1415926: nit:
```suggestion
	ssm   secretGetter
	secretsmanager   secretGetter
```

internal/pkg/ecs/ecs.go
```diff
@@ -232,6 +247,50 @@ func (c Client) StopDefaultClusterTasks(familyName string) error {
 	return c.ecsClient.StopTasks(taskIDs, ecs.WithStopTaskReason(taskStopReason))
 }
 
+// DecryptedSSMSecrets returns the decrypted parameters from the SSM store.
+func (c Client) DecryptedSSMSecrets(secrets []*ecs.ContainerSecret) ([]EnvVar, error) {
```
iamhopaul123: Can we combine this one with `DecryptedSecretManagerSecrets` since

- these two take the same input and output type
- we almost always call `DecryptedSSMSecrets` and `DecryptedSecretManagerSecrets` together

so that we don't need to iterate twice. The output can be a struct with decrypted SSM secrets and secret manager secrets.

Lou1415926: I think we are fine for now (no action to take other than what boss suggested)!

I think the `ecs/ecs` package needs a refactor some day though. _(I think)_ ideally we will have these functions ⬇️ 
```go
type TaskDefinition struct {
  ecs.TaskDefinition
}

// TaskDefinition returns the task definition of the service.
func (c Client) TaskDefinition(app, env, svc string) (*TaskDefinition, error)

// EnvVars returns the environment variables of the task definition
func (taskDef *TaskDefinition) EnvVars() ([]EnvVar, error)

type EnvVar interface {
  Name() string
  Value() string
}

type SecretEnvVar interface {
  EnvVar
  ValueFrom() string
}


type plainTextEnvVar struct { // Implements EnvVar.
  name string
  value string
}

func (plain *plainTextEnvVar) Name() string 
func (plain *plainTextEnvVar) Value() string

type ssmEnvVar struct {
  name string
  valueFrom string
} // Implements EnvVar and SecretEnvVar.

// Returns the name of the environment variable whose value is from SSM.
func (ssm *ssmEnvVar) Name() string

// Returns the value of the environment variable.
func (ssm *ssmEnvVar) Value() string

// Returns the name of the SSM parameter where the value is stored.
func (ssm *ssmEnvVar) ValueFrom() string

type secretsManagerEnvVar struct {
    name string
    valueFrom string
} // Implements EnvVar and SecretEnvVar.

func (sm *secretsManagerEnvVar) Name() string
func (sm *secretsManagerEnvVar) Value() string
func (sm *secretsManagerEnvVar) ValueFrom() string



```

internal/pkg/aws/secretsmanager/secretsmanager.go
```diff
@@ -114,6 +118,18 @@ func (s *SecretsManager) DescribeSecret(secretName string) (*DescribeSecretOutpu
...
 }
 
+// GetSecretValue retrieves the value of a secret from AWS Secrets Manager.
+// It takes the name of the secret as input and returns the corresponding value as a string.
+func (s *SecretsManager) GetSecretValue(name string) (string, error) {
+	resp, err := s.secretsManager.GetSecretValue(&secretsmanager.GetSecretValueInput{
+		SecretId: aws.String(name),
+	})
+	if err != nil {
+		return "", fmt.Errorf("get secret %s from secrets manager: %w", name, err)
```
dannyrandall:
```suggestion
		return "", fmt.Errorf("get secret %q from secrets manager: %w", name, err)
```

internal/pkg/aws/ssm/ssm.go
```diff
@@ -61,6 +65,19 @@ func (s *SSM) PutSecret(in PutSecretInput) (*PutSecretOutput, error) {
...
 
+// GetSecretValue retrieves the value of a parameter from AWS Systems Manager Parameter Store.
+// It takes the name of the parameter as input and returns the corresponding value as a string.
+func (s *SSM) GetSecretValue(name string) (string, error) {
+	resp, err := s.client.GetParameter(&ssm.GetParameterInput{
+		Name:           aws.String(name),
+		WithDecryption: aws.Bool(true),
+	})
+	if err != nil {
+		return "", fmt.Errorf("get parameter %s from SSM with decryption: %w", name, err)
```
dannyrandall:
```suggestion
		return "", fmt.Errorf("get parameter %q from SSM: %w", name, err)
```

internal/pkg/ecs/ecs.go
```diff
@@ -52,6 +53,15 @@ type ecsClient interface {
 type stepFunctionsClient interface {
 	StateMachineDefinition(stateMachineARN string) (string, error)
 }
+type secretGetter interface {
+	GetSecretValue(secretName string) (string, error)
+}
+
+// EnvVar contains values of the environmental variables
```
dannyrandall:
```suggestion
// EnvVar contains the value of an environment variable
```

internal/pkg/ecs/ecs.go
```diff
@@ -52,6 +53,15 @@ type ecsClient interface {
 type stepFunctionsClient interface {
 	StateMachineDefinition(stateMachineARN string) (string, error)
 }
+type secretGetter interface {
```
dannyrandall, line 57:
```suggestion

type secretGetter interface {
```

internal/pkg/ecs/ecs.go
```diff
@@ -232,6 +246,37 @@ func (c Client) StopDefaultClusterTasks(familyName string) error {
 	return c.ecsClient.StopTasks(taskIDs, ecs.WithStopTaskReason(taskStopReason))
 }
 
+// DecryptedSecrets returns the decrypted parameters from either the SSM store or Secrets Manager.
```
dannyrandall:
```suggestion
// DecryptedSecrets returns the decrypted parameters from either SSM parameter store or Secrets Manager.
```

internal/pkg/ecs/ecs.go
```diff
@@ -232,6 +246,37 @@ func (c Client) StopDefaultClusterTasks(familyName string) error {
...
+func (c Client) DecryptedSecrets(secrets []*ecs.ContainerSecret) ([]EnvVar, error) {
+	var ssmSecrets []EnvVar
+	var secretManagerSecrets []EnvVar
+	for _, secret := range secrets {
+		parsed, err := arn.Parse(secret.ValueFrom)
+		if err != nil || parsed.Service == ssm.Namespace {
```
dannyrandall: why are we ignoring the error from parsing the arn?

internal/pkg/ecs/ecs.go
```diff
@@ -232,6 +246,37 @@ func (c Client) StopDefaultClusterTasks(familyName string) error {
 	return c.ecsClient.StopTasks(taskIDs, ecs.WithStopTaskReason(taskStopReason))
 }
 
+// DecryptedSecrets returns the decrypted parameters from either the SSM store or Secrets Manager.
+func (c Client) DecryptedSecrets(secrets []*ecs.ContainerSecret) ([]EnvVar, error) {
+	var ssmSecrets []EnvVar
+	var secretManagerSecrets []EnvVar
```
dannyrandall: could we just have a single list? seems like they just get combined anyways...
```suggestion
	var vars []EnvVar
```

dannyrandall (commented):
looks good! just some small nits and one question!

internal/pkg/ecs/ecs.go
Varun359: If in the Task Definition a secret doesn't have arn as its value and have something like `/copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/my-secret` as the value we get an error. So in that case we are assuming it to be a value from ssm parameter store.

dannyrandall: i think secrets in a task definition _have_ to be ARNs, right? https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Secret.html

> **valueFrom**
The secret to expose to the container. The supported values are either the full ARN of the AWS Secrets Manager secret or the full ARN of the parameter in the SSM Parameter Store.

dannyrandall: discussed offline, there's another line in the doc i missed:
> **Note**
If the SSM Parameter Store parameter exists in the same Region as the task you're launching, then you can use either the full ARN or name of the parameter. If the parameter exists in a different Region, then the full ARN must be specified.

internal/pkg/ecs/ecs.go
```diff
@@ -232,6 +247,35 @@ func (c Client) StopDefaultClusterTasks(familyName string) error {
...
+			secretValue, err := c.ssm.GetSecretValue(secret.ValueFrom)
+			if err != nil {
+				return nil, err
+			}
+			vars = append(vars, EnvVar{
+				Name:  secret.Name,
+				Value: secretValue,
+			})
+		}
+		if parsed.Service == secretsmanager.Namespace {
```
dannyrandall:
```suggestion
		} else if parsed.Service == secretsmanager.Namespace {
```

dannyrandall (approved):
nice work! 🚀

internal/pkg/ecs/ecs.go
```diff
@@ -232,6 +247,34 @@ func (c Client) StopDefaultClusterTasks(familyName string) error {
 	return c.ecsClient.StopTasks(taskIDs, ecs.WithStopTaskReason(taskStopReason))
 }
 
+// DecryptedSecrets returns the decrypted parameters from either SSM parameter store or Secrets Manager.
+func (c Client) DecryptedSecrets(secrets []*ecs.ContainerSecret) ([]EnvVar, error) {
+	var vars []EnvVar
+	for _, secret := range secrets {
+		parsed, err := arn.Parse(secret.ValueFrom)
+		if err != nil || parsed.Service == ssm.Namespace {
```
iamhopaul123: The logic is confusing to me. We have two conditions here for `err` and `parsed.Service` 6 possible cases: 2 for `err` (nil or not nil) and 3 for `parsedService` (ssm.Namespace, secretsmanager.Namespace, or others).

I think we can combine some cases. For example, if `err != nil` that means we didn't parse successfully and we should probably error out in this situation. Otherwise we can do a switch for the `parsed.Service` and make API calls accordingly (for the default case which is other namespace value we should probably also error out).

iamhopaul123 (approved):
:shipit:

PR #5122, opened 2023-07-24
fix: validation of sidecar image fields in the manifest

### LINKED ISSUE #5108: service & sidecar image/image.location is confusing
According to the documentation, to point a main service to a docker registry we use `image.location` whilst sidecars expect just `image`. 

This difference in syntax is slightly confusing, but it's made worse because if you incorrectly use `image.location` in the sidecar config, it then tries to build a Dockerfile in the root of the project.

I questioned my sanity for awhile until I went back to look at the docs. 

It would be helpful if the configuration keys are consistent, and/or if sidecars fail with a loud error if missing both `image` and `image.build` instead of defaulting to looking for a Dockerfile in the root of the repo.

KollaAdithya:
resolves #5108
I would like to apologize 🙇 for missing this validation when I am working on the sidecar build. The below specification leads to building a container image from the root of the repository.
i.e

```
sidecars:
 nginx:
   port:8080
```
**Fix**

We have three scenarios, First two scenarios requires validation where as third scenario does not require validation.

1. If none of `image` and `image.build` and `image.location` are specified. We need a validation on the manifest because we need atleast one of the fields to be specified in the manifest.
2. If both `image.build` and `image.location` with out `image` are specified then we validate the manifest.
3. If both `image` and `image.build` is specified this will not require a validation, this will lead to unmarshal error for the manifest.[no need of validation]
```
sidecars:
 ngnix:
  image: 535307839156.dkr.ecr.us-west-2.amazonaws.com/jerry
    build: ../../demo/web/Dockerfile
```


internal/pkg/manifest/validate.go
```diff
@@ -1498,6 +1498,21 @@ func (s SidecarConfig) validate() error {
 	}
 	return s.ImageOverride.validate()
 }
+func (s SidecarConfig) validateSideCarImageFields() error {
```
iamhopaul123: Can we simplify the logic by

1. first check if `image` is set. If not error out saying it is a required field;
2. if `image` is a struct, check if either `location` or `build` is set.

This way we'll have clearer error msg for customers and from impl perspective we don't need to define a new error type specifically for this.

iamhopaul123 (commented):
Also it's not your fault. We own the codebase collectively :homer-disappearing:

internal/pkg/manifest/validate.go
```diff
@@ -1498,6 +1498,15 @@ func (s SidecarConfig) validate() error {
...
```
dannyrandall: nit, for consistency with struct
```suggestion
func (s SidecarConfig) validateSidecarImageFields() error {
```

internal/pkg/manifest/validate.go
```diff
@@ -1498,6 +1498,15 @@ func (s SidecarConfig) validate() error {
...
+	if s.Image.Basic == nil && s.Image.Advanced.Location == nil && s.Image.Advanced.Build.isEmpty() {
+		return fmt.Errorf(`must specify one of image or image.build or image.location`)
```
```suggestion
		return fmt.Errorf(`must specify one of "image", "image.build", or "image.location"`)
```

internal/pkg/manifest/validate.go
```diff
@@ -1498,6 +1498,15 @@ func (s SidecarConfig) validate() error {
...
+	}
+	if err := s.Image.Advanced.validate(); err != nil {
+		return fmt.Errorf(`validate "build": %w`, err)
```
```suggestion
		return fmt.Errorf(`validate "image.build": %w`, err)
```

internal/pkg/manifest/validate_test.go
```diff
@@ -332,10 +332,16 @@ func TestLoadBalancedWebService_validate(t *testing.T) {
 					ImageConfig: testImageConfig,
 					Sidecars: map[string]*SidecarConfig{
 						"foo": {
+							Image: Union[*string, ImageLocationOrBuild]{
+								Basic: aws.String("mockImage"),
+							},
```
dannyrandall: these should use `BasicToUnion()` instead of directly creating the struct so that `isBasic` gets set on the Union struct

internal/pkg/manifest/validate.go
```diff
@@ -1498,6 +1498,15 @@ func (s SidecarConfig) validate() error {
 	}
 	return s.ImageOverride.validate()
 }
+func (s SidecarConfig) validateSideCarImageFields() error {
+	if s.Image.Basic == nil && s.Image.Advanced.Location == nil && s.Image.Advanced.Build.isEmpty() {
```
dannyrandall: this can be simplified:
```suggestion
	if s.IsZero() {
```

internal/pkg/manifest/validate.go
```diff
@@ -1498,6 +1498,15 @@ func (s SidecarConfig) validate() error {
...
+		return fmt.Errorf(`must specify one of image or image.build or image.location`)
+	}
+	if err := s.Image.Advanced.validate(); err != nil {
```
dannyrandall: this will call validate on the correct underlying type (Advanced, in this case, if it's set)
```suggestion
	if err := s.Image.validate(); err != nil {
```

dannyrandall (commented):
just some nits about the `Union` type!

internal/pkg/manifest/validate.go
```diff
@@ -2236,7 +2245,6 @@ func (i ImageLocationOrBuild) validate() error {
 		return &errFieldMutualExclusive{
 			firstField:  "build",
 			secondField: "location",
-			mustExist:   true,
```
dannyrandall: wait aren't the mutually exclusive?

KollaAdithya: I think we set `mustExist: true` if both `image.build` and `image.location` are empty. we are doing that validation over here. https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/manifest/validate.go#L714:L719 

In `errors.go` we have error message based on `mustExist` field. https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/manifest/errors.go#L94:L99

internal/pkg/manifest/validate.go
```diff
@@ -1498,6 +1498,15 @@ func (s SidecarConfig) validate() error {
 	}
 	return s.ImageOverride.validate()
 }
+func (s SidecarConfig) validateSidecarImageFields() error {
```
Lou1415926: nit: the "sidecar" keyword in the function name is redundant with the struct name
```suggestion
func (s SidecarConfig) validateImages() error {
```

internal/pkg/manifest/validate.go
dannyrandall: interesting - should we just remove the check here then? is it just a duplicate validation?

internal/pkg/manifest/validate.go
```diff
@@ -1498,6 +1498,15 @@ func (s SidecarConfig) validate() error {
...
+	if s.Image.IsZero() {
+		return fmt.Errorf(`must specify one of "image", "image.build, or "image.location"`)
+	}
+	if err := s.Image.validate(); err != nil {
+		return fmt.Errorf(`validate "image.build": %w`, err)
```
Lou1415926: I thought `s.Image.validate()` would validate both `image.location` and `image.build` right?
```suggestion
		return fmt.Errorf(`validate "image": %w`, err)
```

KollaAdithya: That's true 👍

internal/pkg/manifest/validate.go
KollaAdithya: It is not a redundant validation. we are checking two cases.
1. if both `image.build` and `image.location` are empty
2. If both `image.build` and `image.location` are not empty.
I just moved both validation to one place in this commit [9fd2a4e](https://github.com/aws/copilot-cli/pull/5122/commits/9fd2a4e431c3ee2242411aeb97677f6b4012e3c4)

internal/pkg/manifest/validate.go
```diff
@@ -1498,6 +1491,15 @@ func (s SidecarConfig) validate() error {
 	}
 	return s.ImageOverride.validate()
 }
+func (s SidecarConfig) validateImages() error {
```
```suggestion
func (s SidecarConfig) validateImage() error {
```

internal/pkg/manifest/validate.go
```diff
@@ -1459,6 +1452,9 @@ func (l Logging) validate() error {
 
 // validate returns nil if SidecarConfig is configured correctly.
 func (s SidecarConfig) validate() error {
+	if err := s.validateImages(); err != nil {
```
```suggestion
	if err := s.validateImage(); err != nil {
```

internal/pkg/manifest/validate.go
```diff
@@ -2232,7 +2234,7 @@ func (i ImageLocationOrBuild) validate() error {
 	if err := i.Build.validate(); err != nil {
 		return fmt.Errorf(`validate "build": %w`, err)
 	}
-	if !i.Build.isEmpty() && i.Location != nil {
+	if i.Build.isEmpty() == (i.Location == nil) {
```
Lou1415926, line 2237: Won't this error out if we have ⬇️  ?
```yaml
sidecars:
  nginx:
     image: <mylocation>
```

Do you know what happens if we had ⬇️ 
```yaml
name: mysvc

image:
   port: 80
```
That is, neither `location` or `build` is set for the _main container_. Do we error out on validation, or do we error out when we try to build the image?

KollaAdithya: In case of main container with the below config we error out on validation.

```
image:
  port: 80
```

`✘ validate manifest against environment "test": validate "image": must specify one of "build" and "location"`

In case of below  config ⬇️ for sidecars
```
sidecars:
  nginx:
     image: <mylocation>
```  
we do not error out in this case because we are doing validation on `image.build` and `image.location` only

The cases we error out  are ⬇️ 
```
sidecars:
  nginx:
   image:
    build: ../../demo/web/Dockerfile
    location: "location"
   port: 81
   ```
   `validate manifest against environment "test": validate "sidecars[nginx]": validate "image": must specify one of "build" and "location"`
  ``` 
   sidecars:
  nginx:
   port: 81
   ```
   `validate manifest against environment "test": validate "sidecars[nginx]": must specify one of "image", "image.build, or "image.location"`

Comment (Lou1415926) [reply internal/pkg/manifest/validate.go:2237]: > In case of below config ⬇️ for sidecars
> ```
> sidecars:
>  nginx:
>     image: <mylocation>
> ```
> we do not error out in this case.

I agree that we shouldn't! 


Correct me if I'm wrong though, with current logic (the revision that i'm looking at rn), we will:
1. Go to `if s.Image.IsZero() {` (L1495); because Image is not zero (it has the string) we don't error out
2. Go to `if err := s.Image.validate(); err != nil {`, where `s.image` is an `ImageLocationOrBuild`), this goes to the code that this comment is placed under. It will error out with `errFieldMutualExclusive` because both `i.Build.isEmpty`  and `i.Location == nil` are true.

Comment (KollaAdithya) [reply internal/pkg/manifest/validate.go:2237]: exactly this is true!

> 1. Go to if s.Image.IsZero() { (L1495); because Image is not zero (it has the string) we don't error out

`IsZero()` will return true both `Basic` and `Advanced` are zero. 
In this case 
```
sidecars:
 nginx:
    image: <mylocation>
```
In this case `Basic` value is set so we do not error out.

`if err := s.Image.validate(); err != nil ` will call the validation of `Union` type over here which will validate both Basic and Advanced Type.
https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/manifest/union.go#L145

In this case we have `Image  Union[*string, ImageLocationOrBuild]`
as there is no validation on the `Basic `type which is just a string, but we have validation on `ImageLocationOrBuild`

Comment (dannyrandall) [reply internal/pkg/manifest/validate.go:2237]: @Lou1415926 I _think_ that `SidecarConfig.Image` has a couple layers of nesting that is making this confusing:
- `SidecarConfig.Image` is the union type between `*string` and `ImageLocationOrBuild`  
- `SidecarConfig.Image.Advanced` is how to access the `ImageLocationOrBuild` field
- `SidecarConfig.Image.validate()` calls `ImageLocationOrBuild.validate()` only if the config set `build` or `location`

Therefore, in the situation you proposed:
```
sidecars:
  nginx:
    image: <mylocation>
```

My understanding is that `s.Image.IsZero()` returns false, as you said. Then, `s.Image.Validate()` will _**not call any validation function**_, because the underlying value of the union is `*string`, which doesn't have a `validate()` function on it!

Lou1415926: makes sense!!

Lou1415926 (approved):

internal/pkg/manifest/validate.go
dannyrandall: ah @KollaAdithya you beat me too it! I agree with your synopsis 👍

dannyrandall (approved):

Issue #4892, opened 2023-05-18
Improve UX


pax-k:
Hey! Killing the command with CTRL+C while it provisions resources gives me a hard time when I try to work with copilot. Trying to redeploy / do stuff around yields errors. I have to manually delete apps / service / envs, or to kill Cloudformation stacks manually. Would it be an idea for the CLI to listen for CTLR+C and trigger a rollback?

Also it's not yet clear for me how to handle multiple apps from a monorepo. Should the copilot directory be at the root of the monorepo or at the root of each app?

[Also this https://github.com/aws/copilot-cli/issues/3676](https://github.com/aws/copilot-cli/issues/3676#issuecomment-1553372063)](https://github.com/aws/copilot-cli/issues/3676#issuecomment-1553372063)

Also please add `remove` alongside `delete`. Remove should just soft-delete / remove aws resources, without deleting any local files - leave that to `delete`.

Also the official documentation uses the words "app" / "service" interchangeably like they mean the same thing: https://aws.github.io/copilot-cli/docs/commands/svc-init/

> After running this command, the CLI creates sub-directory with your app name in your local copilot directory where you'll find a manifest file.

No, "the CLI creates sub-directory with your **service** name."

The `copilot svc init` command should also ask for build context, besides Dockerfile path. 

And even if I previously created a dev environment, the `copilot svc init` command after it finishes still recommends test environment:
```- Run copilot svc deploy --name gateway --env test` to deploy your service to a test environment.```

So this is the full chain of commands I have to run if I use a custom domain:
```
copilot app init some-app --domain domain.xyz
copilot init
copilot env init
copilot env deploy --name dev
copilot svc init --name gateway --svc-type "Load Balanced Web Service"
copilot svc deploy --name gateway --env dev
```
Notice the first 2 commands, the order feels bad, but this is what the docs say: https://aws.github.io/copilot-cli/docs/commands/app-init/

If any of these steps fail or are killed while provisioning, I have a hard time rolling back.
I would love to be able to add multiple apps to the same copilot/ directory and compress everything into 2 commands:

- `copilot init`: should ask for app name, domain if necessary, then for environment name, then for service name
- `copilot deploy`: should ask what app context to use, what service, what environment, done

If it happens that I trigger CTRL+C, the script should trigger a rollback.
If some CloudFormation tasks fails, I want to see reason and the logs.
[...]

PR #5123, opened 2023-07-25
chore: ctrl-c for env deploy and detach flag for svc/job/env deploy and copilot deploy

KollaAdithya:
This PR will

- Enable ctrl-c for env deploy
- detach flag for `copilot svc/job/env deploy` and `copilot deploy`.

The Ux will look like this below
```
Building your container image frontend:
   build output

✔ Proposing infrastructure changes for stack ninja-test-frontend

kolladit$:
```


internal/pkg/deploy/cloudformation/env.go
```diff
@@ -64,6 +64,8 @@ func (cf CloudFormation) UpdateAndRenderEnvironment(conf StackConfiguration, buc
 		}
 		return changeSetID, nil
 	}
+	in.enableInterrupt = true
+	in.detach = true
```
iamhopaul123: Why do we always set them to true for env deploy, which is not the same as https://github.com/aws/copilot-cli/pull/5123/files#diff-082bb2a58e8378e094bbe9302a375183abd32e5cce67e2e4f90bc46fd5302c41R30?

internal/pkg/cli/flag.go
```diff
@@ -269,6 +270,7 @@ rollback in case of deployment failure.
 We do not recommend using this flag for a
 production environment.`
 	forceEnvDeployFlagDescription = "Optional. Force update the environment stack template."
+	detachFlagDescription         = "Optional. Skip displaying CloudFormation stack events."
```
iamhopaul123: Maybe
```suggestion
	detachFlagDescription         = "Optional. Skip displaying CloudFormation deployment progress."
```

internal/pkg/deploy/cloudformation/env.go
KollaAdithya: This is because function [`func (cf CloudFormation) newUpsertChangeSetInput`](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/deploy/cloudformation/cloudformation.go#L311) is getting called in both 
[`func (cf CloudFormation) DeployService`](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/deploy/cloudformation/workload.go#L18) and [`func (cf CloudFormation) DeployTask`](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/deploy/cloudformation/task.go#L20)

But we want to enable interrupt and detach only for `workloads` not for `task run`.

iamhopaul123: but isn't `detach` the param defined but not used at all?

KollaAdithya: oops sorry 🤦  i changed it to `in.detach=detach`

internal/pkg/cli/env_deploy.go
```diff
@@ -223,6 +226,15 @@ func (o *deployEnvOpts) Execute() error {
 		}
 	}
 	if err := deployer.DeployEnvironment(deployInput); err != nil {
+		var errStackDeletedOnInterrupt *deploycfn.ErrStackDeletedOnInterrupt
+		var errStackUpdateCanceledOnInterrupt *deploycfn.ErrStackUpdateCanceledOnInterrupt
+		if errors.As(err, &errStackDeletedOnInterrupt) {
+			return nil
+		}
+		if errors.As(err, &errStackUpdateCanceledOnInterrupt) {
+			log.Successf("Successfully rolled back service %s to the previous configuration.\n", color.HighlightUserInput(o.name))
```
huanjani: tiny nit:
```suggestion
			log.Successf("Successfully rolled back service %q to the previous configuration.\n", color.HighlightUserInput(o.name))
```

huanjani (approved):

huanjani:
Related: #4892

internal/pkg/cli/env_deploy.go
```diff
@@ -223,6 +226,15 @@ func (o *deployEnvOpts) Execute() error {
 		}
 	}
 	if err := deployer.DeployEnvironment(deployInput); err != nil {
+		var errStackDeletedOnInterrupt *deploycfn.ErrStackDeletedOnInterrupt
```
iamhopaul123: Can we flip the logic to make less code indented?

```go
err := deployer.DeployEnvironment(deployInput)
if err == nil {
  return nil
}
```
and a switch statement to check the `err` type.

internal/pkg/cli/env_deploy.go
```diff
@@ -223,6 +226,15 @@ func (o *deployEnvOpts) Execute() error {
...
+		var errStackUpdateCanceledOnInterrupt *deploycfn.ErrStackUpdateCanceledOnInterrupt
+		if errors.As(err, &errStackDeletedOnInterrupt) {
+			return nil
+		}
+		if errors.As(err, &errStackUpdateCanceledOnInterrupt) {
+			log.Successf("Successfully rolled back service %q to the previous configuration.\n", color.HighlightUserInput(o.name))
+			return nil
+		}
```
iamhopaul123: Should we add "successfully deployed env ..." for `env deploy` 🤔

internal/pkg/cli/env_deploy.go
```diff
@@ -256,6 +258,7 @@ After fixing the deployment, you can:
 		}
 		return fmt.Errorf("deploy environment %s: %w", o.name, err)
 	}
+	log.Ssuccessf("Deployed environment %s.\n", color.HighlightUserInput(o.name))
```
iamhopaul123: we probably need detached flag checking there to exit earlier.

iamhopaul123 (approved):

internal/pkg/cli/env_deploy.go
KollaAdithya: I think we do not need that because we have an earlier return stmt here 
https://github.com/aws/copilot-cli/pull/5123/files#diff-ed267edf51bd253fe120a5d1b9ed09b357b1c61eafac3501a95642aced180ee1R237
and here
https://github.com/aws/copilot-cli/pull/5123/files#diff-ed267edf51bd253fe120a5d1b9ed09b357b1c61eafac3501a95642aced180ee1R241

KollaAdithya: I just moved the log statement and added detach condition in this [commit](https://github.com/aws/copilot-cli/pull/5123/commits/057e19dc2d738c90ac46320f2903d6773a738c31)
```
if err == nil {
if detach{
return nil
}
  log.Successf("succesfully deployed Environment")
  return nil
}

PR #5125, opened 2023-07-25
fix(ecs): filter out non-active ECS services

iamhopaul123:
Applies similar fix as https://github.com/aws/copilot-cli/pull/5062 to ecs service.


internal/pkg/aws/ecs/service.go
```diff
@@ -123,32 +123,66 @@ type ServiceArn string
...
+
+func (s *ServiceArn) parse() (*parsedServiceArn, error) {
 	serviceArn := string(*s)
 	parsedArn, err := arn.Parse(serviceArn)
 	if err != nil {
-		return "", err
+		return nil, err
+	}
+	if parsedArn.Service != EndpointsID {
+		return nil, fmt.Errorf("expect an ECS arn: got %q", serviceArn)
```
dannyrandall:
```suggestion
		return nil, fmt.Errorf("expected an ECS arn, but got %q", serviceArn)
```

internal/pkg/aws/ecs/service.go
```diff
@@ -123,32 +123,66 @@ type ServiceArn string
...
 // will return my-project-test-Cluster-9F7Y0RLP60R7
 func (s *ServiceArn) ClusterName() (string, error) {
+	parsedArn, err := s.parse()
+	if err != nil {
+		return "", err
+	}
+	return parsedArn.clusterName, nil
+}
+
+type parsedServiceArn struct {
```
dannyrandall: i wonder if it makes sense to do something like:
```
type ServiceARN struct {
  // AccountID, Partition, Region, etc
}

func ParseServiceARN() (ServiceARN, error) {
  // contents of parse()
}
```

then users of the struct can pull fields without `parse()`ing multiple times

internal/pkg/ecs/ecs.go
```diff
@@ -449,10 +450,23 @@ func (c Client) serviceARN(app, env, svc string) (*ecs.ServiceArn, error) {
 	if len(services) == 0 {
 		return nil, fmt.Errorf("no ECS service found with tags %s", tags.String())
 	}
-	if len(services) > 1 {
+	arns := make([]string, len(services))
+	for i := range services {
+		arns[i] = services[i].ARN
+	}
+	serviceARN := ecs.ServiceArn(arns[0])
```
dannyrandall: is this making the assumption that all of tagged the services are in the same cluster? isn't it possible that's not true?

iamhopaul123: They are in the same cluster i think for this use case. We can make another function or modify this one if we need a different behavior in the other place

internal/pkg/aws/ecs/ecs.go
```diff
@@ -333,13 +333,35 @@ func (e *ECS) ActiveClusters(arns ...string) ([]string, error) {
...
 	}
 
 	return active, nil
+}
 
+// ActiveServices returns the subset of service arns that have an ACTIVE status.
+func (e *ECS) ActiveServices(clusterARN string, serviceARNs ...string) ([]string, error) {
+	resp, err := e.client.DescribeServices(&ecs.DescribeServicesInput{
+		Cluster:  aws.String(clusterARN),
+		Services: aws.StringSlice(serviceARNs),
```
bvtujo, line 360: What happens if there are no arns specified? Is this an error, or does the DescribeServices call describe all services on the cluster?

Can we add a unit test looking for this behavior?

internal/pkg/ecs/ecs.go
bvtujo: I think having to specify an app+env provides enough security, right? I don't think Copilot will let you make two envs of the same name in the same app. Though what happens if someone creates an app in tools account A, then an env in account B, then an app in tools account C, and an env in account B? Maybe this is out of scope 😂

internal/pkg/ecs/ecs.go
```diff
@@ -449,10 +450,23 @@ func (c Client) serviceARN(app, env, svc string) (*ecs.ServiceArn, error) {
...
+	clusterArn, err := serviceARN.ClusterArn()
+	if err != nil {
+		return nil, fmt.Errorf("get cluster arn: %w", err)
+	}
+	active, err := c.ecsClient.ActiveServices(clusterArn, arns...)
+	if err != nil {
+		return nil, fmt.Errorf("check if clusters are active: %w", err)
```
bvtujo:
```suggestion
		return nil, fmt.Errorf("check if services are active: %w", err)
```

bvtujo (changes requested):

internal/pkg/ecs/ecs.go
iamhopaul123: Actually i'll change the logic

iamhopaul123:
> I think having to specify an app+env provides enough security, right? I don't think Copilot will let you make two envs of the same name in the same app. Though what happens if someone creates an app in tools account A, then an env in account B, then an app in tools account C, and an env in account B? Maybe this is out of scope 

Good call i think i added some extra logic to catch this.

internal/pkg/aws/ecs/ecs.go
iamhopaul123: very good call will add

internal/pkg/aws/ecs/ecs.go
```diff
@@ -342,9 +345,23 @@ func (e *ECS) ActiveClusters(arns ...string) ([]string, error) {
 }
 
 // ActiveServices returns the subset of service arns that have an ACTIVE status.
```
bvtujo, line 344: Can we add a comment to suggest that this function should only be called on services in the same cluster?

bvtujo (approved):
lgtm just one tiny nit

internal/pkg/aws/ecs/ecs.go
```diff
@@ -342,9 +345,24 @@ func (e *ECS) ActiveClusters(arns ...string) ([]string, error) {
...
-func (e *ECS) ActiveServices(clusterARN string, serviceARNs ...string) ([]string, error) {
+// Note that all services should be in the same cluster.
+func (e *ECS) ActiveServices(serviceARNs ...string) ([]string, error) {
+	if len(serviceARNs) == 0 {
```
dannyrandall: not sure if the api catches this error or just ignores it, but the max is 10

internal/pkg/aws/ecs/service.go
```diff
@@ -163,26 +151,29 @@ func (s *ServiceArn) parse() (*parsedServiceArn, error) {
...
 // For example: arn:aws:ecs:us-west-2:1234567890:service/my-project-test-Cluster-9F7Y0RLP60R7/my-project-test-myService-JSOH5GYBFAIB
 // will return my-project-test-myService-JSOH5GYBFAIB
-func (s *ServiceArn) ServiceName() (string, error) {
-	parsedArn, err := s.parse()
-	if err != nil {
-		return "", err
-	}
-	return parsedArn.name, nil
+func (s *ServiceArn) ServiceName() string {
+	return s.name
```
dannyrandall, line 169: any reason to not just export this on the struct as `s.ServiceName`? (same for cluster name)

internal/pkg/aws/ecs/ecs.go
dannyrandall: i kinda think - if the ECS api catches these errors (0 ARNs specified, > 10 specified), we shouldn't explicitly handle them here. at least for 0, it seems like we already have logic to catch that error in the caller.

if the ECS api has some weird behavior on them - like returning all services in the cluster or something - then we can handle them explicitly.

internal/pkg/aws/ecs/service.go
iamhopaul123: Becuase everything in that struct should be private and init by the Parse function.

dannyrandall (approved):

internal/pkg/aws/ecs/ecs.go
```diff
@@ -333,13 +333,47 @@ func (e *ECS) ActiveClusters(arns ...string) ([]string, error) {
...
+// Note that all services should be in the same cluster.
+func (e *ECS) ActiveServices(serviceARNs ...string) ([]string, error) {
+	var prevSvcArn *ServiceArn
+	for _, arn := range serviceARNs {
+		svcArn, err := ParseServiceArn(arn)
+		if err != nil {
+			return nil, err
+		}
+		if prevSvcArn != nil && prevSvcArn.clusterName != svcArn.clusterName {
+			return nil, fmt.Errorf("service %q and service %q should be in the same cluster", prevSvcArn.String(), svcArn.String())
```
joanniclaborde, line 354: Hi! I keep getting this error when doing `copilot service show`, and it's referencing old (deleted) clusters and services. Is it possible that this cluster check should be done after inactive services are filtered out?

KollaAdithya:
Hey @joanniclaborde !

Which version of Copilot are you using. Fix for this is released in Copilot version `v1.29.1` 
https://github.com/aws/copilot-cli/releases/tag/v1.29.1

joanniclaborde:
> Hey @joanniclaborde !
> 
> Which version of Copilot are you using. Fix for this is released in Copilot version `v1.29.1` https://github.com/aws/copilot-cli/releases/tag/v1.29.1

I'm using the latest version, v1.29.1.

KollaAdithya:
Can you please paste the error message to help us debug the issue?

joanniclaborde:
```
% copilot svc show                                                                                                                                                            
Service name: my-svc
✘ describe service my-svc: retrieve rollback alarm names: get service my-svc: check if services are active: service "arn:aws:ecs:us-east-1:1234567890:service/my-app-staging-Cluster-s63uAzrXXXXX/my-app-staging-my-svc-Service-c0iGyjmXXXXX" and service "arn:aws:ecs:us-east-1:1234567890:service/my-app-staging-Cluster-7tGl5VSXXXXX/my-app-staging-my-svc-Service-2DLOI9QXXXXX" should be in the same cluster
```

`my-app-staging-Cluster-s63uAzrXXXXX` is the current active cluster, while `my-app-staging-Cluster-7tGl5VSXXXXX` is an old deleted one.

I still get that same error message when I delete the whole app and recreate it (with the same names, of course).

Thanks!

KollaAdithya:
Thanks for reporting the bug. I am able to reproduce it locally. We will work on the fix for this!

dannyrandall:
Hey @joanniclaborde! We just released this fix [v1.30.0](https://github.com/aws/copilot-cli/releases/tag/v1.30.0)!🚀🎉 Thanks for reporting the issue!

PR #5126, opened 2023-07-25
fix(cli): init should set appVersionGetter with correct app name

iamhopaul123:
Right now `init` interactive wizard is broken because we don't set `appVersionGetter` with correct app name. This PR fixes this issue.


internal/pkg/cli/init.go
```diff
@@ -145,15 +141,17 @@ func newInitOpts(vars initVars) (*initOpts, error) {
...
-		store:            configStore,
-		appDeployer:      deployer,
-		prog:             spin,
-		prompt:           prompt,
-		identity:         id,
-		appVersionGetter: appVersionGetter,
-		appCFN:           cloudformation.New(defaultSess, cloudformation.WithProgressTracker(os.Stderr)),
-		sess:             defaultSess,
-		templateVersion:  version.LatestTemplateVersion(),
+		store:       configStore,
```
dannyrandall, line 143: i can't comment on the lines i want😭 but i wonder if we should avoid setting `initEnvVars.appName` in general here, as it _only_ works if flags were passed in?

dannyrandall (commented):
looks great! just a question

internal/pkg/cli/init.go
iamhopaul123: ok done!

dannyrandall (approved):

KollaAdithya (approved):

PR #5134, opened 2023-07-27
chore: fix sc validation for backend and worker

### LINKED ISSUE #5133: docs: Service Connect default value
[The documentation for Backend Service](https://aws.github.io/copilot-cli/docs/manifest/backend-service/#network-connect) says  `network.connect` dafaults to false, but actually it is defined as true in [the manifest template](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/workloads/services/backend/manifest.yml#L47).

[The documentation for Load Balanced Web Service](https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#network-connect) says  `network.connect` dafaults to false, but actually it is defined as true in [the manifest template](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/workloads/services/lb-web/manifest.yml#L41).

[The documentation for Worker Service](https://aws.github.io/copilot-cli/docs/manifest/worker-service/#network-connect) says  `network.connect` dafaults to false, and [the manifest template](https://github.com/aws/copilot-cli/blob/mainline/internal/pkg/template/templates/workloads/services/worker/manifest.yml) has no default value. Maybe the default value is defined somewhere.

Is Service Connect default value different between service types?

iamhopaul123:
fixes #5133 


internal/pkg/deploy/cloudformation/stack/worker_svc.go
```diff
@@ -122,10 +122,6 @@ func (s *WorkerService) Template() (string, error) {
 	if err != nil {
 		return "", fmt.Errorf(`convert "publish" field for service %s: %w`, s.name, err)
 	}
-	var scConfig *template.ServiceConnect
-	if s.manifest.Network.Connect.Enabled() {
```
bvtujo, line 126: What if customers want to enable service connect *from* the worker service? For example, my worker service might need to call a backend service in order to do its work, and I want it to be running the SC sidecar so I can use simple DNS aliases like `POST http://backend/putAnalytics`. Service connect will still be enabled, but the worker service won't be given an address. 
For example, the config in [lines 636-43](https://github.com/aws/copilot-cli/pull/5134/files#diff-e66d8cce2b51a4d4bc88c33fbf62cbbe60388aec3b393cea0de628e52d2d7ae4L636-L643) is actually valid for this use case: there are no services that the worker advertises, but it can still make requests via service connect to other services in the cluster. 

Could we make it explicit with a config setting in the manifest that's valid for all ECS service types? 

```yaml
network:
  connect:
    outbound_only: true # Mutually exclusive with alias
```

For worker service type, we could additionally make 
```yaml
network:
  connect: true
```
default to this behavior (which I think we do now).

WDYT?

bvtujo: Actually this is backward breaking; templates which previously deployed (worker service with `connect: true`) will no longer deploy. Is there a bug that this change fixes?

iamhopaul123: > What if customers want to enable service connect from the worker service?

I see that makes sense to me!

iamhopaul123:
Close this PR because of https://github.com/aws/copilot-cli/pull/5134#discussion_r1278050662

PR #5137, opened 2023-07-27
chore: add `build` images and `run` containers for `local run`

Varun359:


internal/pkg/cli/local_run.go
```diff
@@ -106,11 +196,195 @@ func (o *localRunOpts) Execute() error {
...
+	if err != nil {
+		return fmt.Errorf("get environment %q configuration: %w", o.envName, err)
+	}
+	o.targetEnv = env
+
+	app, err := o.store.GetApplication(o.appName)
+	if err != nil {
+		return fmt.Errorf("get application %q configuration: %w", o.appName, err)
+	}
+	o.targetApp = app
```
dannyrandall: i think we should set these fields in `Validate()` or `Ask()`!

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -201,6 +210,72 @@ func (c DockerCmdClient) Push(ctx context.Context, uri string, w io.Writer, tags
...
+	err := c.runner.Run("docker", []string{"ps", "-q", "-f", "name=" + containerName}, exec.Stdout(buf))
+	if err != nil {
+		return false, fmt.Errorf("run docker ps: %w", err)
+	}
+
+	output := strings.TrimSpace(buf.String())
+	if output == "" {
+		return false, nil // The container is not running.
+	}
+	return true, nil // The container is running.
```
dannyrandall: doesn't this return true if _any_ containers are running? it doesn't seem like we are filtering to ensure `containerName` is running

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -201,6 +210,72 @@ func (c DockerCmdClient) Push(ctx context.Context, uri string, w io.Writer, tags
...
+	args := []string{"run"}
+
+	//Add container name option.
+	if in.ContainerName != "" {
+		args = append(args, "--name", in.ContainerName)
+	}
+
+	// Add port mappings.
+	for hostPort, containerPort := range in.ContainerPorts {
+		args = append(args, "-p", fmt.Sprintf("%s:%s", hostPort, containerPort))
```
dannyrandall: can we expand these to the full version? `-p` and `-e`? just makes it clearer what is happening

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -201,6 +210,72 @@ func (c DockerCmdClient) Push(ctx context.Context, uri string, w io.Writer, tags
...
+	}
+
+	//Add the image name.
+	if in.ImageURI != "" {
+		args = append(args, in.ImageURI)
+	}
+
+	//Add "sleep infinity" command for the "pause" container
+	if in.ContainerName == "pause" {
+		args = append(args, "sleep", "infinity")
```
dannyrandall: it feels like we should add a `Command` option to `RunOptions` that let's the caller specify what command to run, instead of choosing it based on the container name

internal/pkg/cli/deploy/workload.go
```diff
@@ -430,17 +468,20 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
...
 			defer digestsMu.Unlock()
-			out.ImageDigests[name] = ContainerImageIdentifier{
+			in.Out.ImageDigests[name] = ContainerImageIdentifier{
 				Digest:            digest,
-				CustomTag:         d.image.CustomTag,
-				GitShortCommitTag: d.image.GitShortCommitTag,
+				CustomTag:         in.CustomTag,
+				GitShortCommitTag: in.GitShortCommitTag,
 			}
+			imageName := fmt.Sprintf("%s:%s", uri, buildArgs.Tags[0])
```
dannyrandall: do we validate anywhere that there is at least one tag?

internal/pkg/cli/deploy/workload.go
```diff
@@ -393,24 +409,46 @@ func (img ContainerImageIdentifier) Tag() string {
...
+	}
+	return uploadBuildContainerImages(in)
+}
+
+// BuildContainerImages builds the all the images given the build arguments
+func BuildContainerImages(in *BuildImageArgs) error {
+	return uploadBuildContainerImages(in)
+}
+
+func uploadBuildContainerImages(in *BuildImageArgs) error {
```
dannyrandall: could we just rename this function `BuildContainerImages?

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +196,195 @@ func (o *localRunOpts) Execute() error {
...
+		envVars[envVariable.Name] = envVariable.Value
+	}
+
+	containerPorts := make(map[string]string)
+	containerdef := taskDef.ContainerDefinitions
+	for _, container := range containerdef {
+		for _, portMapping := range container.PortMappings {
+			hostPort := aws.Int64Value(portMapping.HostPort)
+			hostPortStr := strconv.FormatInt(hostPort, 10)
+			containerport := aws.Int64Value(portMapping.ContainerPort)
```
dannyrandall: if the host port is unset, shouldn't we default to containerPort?

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +196,195 @@ func (o *localRunOpts) Execute() error {
...
+		return fmt.Errorf("building container image: %w", err)
+	}
+
+	containerNames := o.out.ContainerNames
+	imageNames := o.out.ImageNames
+
+	secretsList := make(map[string]string)
+	for _, s := range decryptedSecrets {
+		secretsList[s.Name] = s.Value
+	}
```
dannyrandall: can we move this up by the env var parsing?

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +196,195 @@ func (o *localRunOpts) Execute() error {
...
+
+	for i, image := range imageNames {
+		imageInfo := imageInfo{
+			containerName: containerNames[i],
+			imageURI:      image,
+		}
+		o.imageInfoList = append(o.imageInfoList, imageInfo)
+	}
```
dannyrandall:
```suggestion
	for i, image := range imageNames {
		o.imageInfoList = append(o.imageInfoList, imageInfo{
			containerName: containerNames[i],
			imageURI:      image,
		})
	}
```

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +196,195 @@ func (o *localRunOpts) Execute() error {
...
+
+	err = o.runPauseContainer(containerPorts)
+	if err != nil {
+		return err
+	}
+
+	err = o.runContainers(o.imageInfoList, secretsList, envVars)
+	if err != nil {
+		return err
+	}
```
dannyrandall: i'm wondering if we should prefix or suffix the container name with a random identifier to create run "sessions". that would enable someone to use `local run` for multiple services without running into naming conflicts

dannyrandall: or maybe it just makes sense to prefix all the containers (including pause) with the app, env, and service name?

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +196,195 @@ func (o *localRunOpts) Execute() error {
...
+	return nil
+}
+
+func getBuiltSideCarImages(sidecars map[string]*manifest.SidecarConfig) []imageInfo {
+	//get the image URI for the sidecars which are in a registry
+	sideCarBuiltImages := make(map[string]string)
+	for sideCarName, sidecar := range sidecars {
+		if uri, hasLocation := sidecar.ImageURI(); hasLocation {
+			sideCarBuiltImages[sideCarName] = uri
+		}
```
dannyrandall: could we move all the logic into this loop?

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +196,195 @@ func (o *localRunOpts) Execute() error {
...
+		imageInfo := imageInfo{
+			containerName: sidecarName,
+			imageURI:      image,
+		}
+		sideCarBuiltImageInfo = append(sideCarBuiltImageInfo, imageInfo)
+	}
+	return sideCarBuiltImageInfo
+}
+
+func (o *localRunOpts) runPauseContainer(containerPorts map[string]string) error {
```
dannyrandall: i think passing a parent context into both this and `runContainers` will be helpful in stopping the running containers!

internal/pkg/docker/dockerengine/dockerengine.go
Varun359: It does show only if the container with specific name is running the -f flag which is used is to filter right? I will change it to --filter. Is that what you asked?

internal/pkg/cli/deploy/workload.go
Varun359: When we appending latest as tag to every container. So we have at least one tag. https://github.com/aws/copilot-cli/blob/123b8afbae354008a87b786723da002ff84303e1/internal/pkg/cli/deploy/workload.go#L492

internal/pkg/cli/local_run.go
Varun359: Thats a good idea.

internal/pkg/docker/dockerengine/dockerengine.go
dannyrandall: oh, cool! `--filter` will help - i thought `-f` was just for the output format 🤦‍♂️

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -201,6 +212,71 @@ func (c DockerCmdClient) Push(ctx context.Context, uri string, w io.Writer, tags
...
+	err := c.runner.Run("docker", []string{"ps", "-q", "--filter", "name=" + containerName}, exec.Stdout(buf))
+	if err != nil {
+		return false, fmt.Errorf("run docker ps: %w", err)
+	}
+
+	output := strings.TrimSpace(buf.String())
+	if output == "" {
+		return false, nil // The container is not running.
+	}
+	return true, nil // The container is running.
```
Lou1415926: nit:
```suggestion
        return output != "", nil
```

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -201,6 +212,71 @@ func (c DockerCmdClient) Push(ctx context.Context, uri string, w io.Writer, tags
...
+	}
+	return args
+}
+
+// Run runs a Docker container with the sepcified options.
+func (c DockerCmdClient) Run(ctx context.Context, options *RunOptions) error {
+	args := options.generateRunArguments()
+
+	//Execute the Docker run command.
+	if err := c.runner.RunWithContext(ctx, "docker", args); err != nil {
```
Lou1415926: nit nit nit:
```suggestion
	if err := c.runner.RunWithContext(ctx, "docker", options.generateRunArguments()); err != nil {
```
(feel free to address this nit later if you are currently using the `args` variable for debugging)

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -201,6 +212,71 @@ func (c DockerCmdClient) Push(ctx context.Context, uri string, w io.Writer, tags
 	return parts[1], nil
 }
 
+func (in *RunOptions) generateRunArguments() []string {
+	args := []string{"run"}
+
+	//Add container name option.
```
Lou1415926: I feel like we probably don't need these comments because the code is pretty self-explanatory!

internal/pkg/repository/repository.go
```diff
@@ -55,6 +55,14 @@ func NewWithURI(registry Registry, name, uri string) *Repository {
 	}
 }
 
+// Build build the image from Dockerfile
+func (r *Repository) Build(ctx context.Context, args *dockerengine.BuildArguments, w io.Writer) (digest string, err error) {
+	if err := r.docker.Build(ctx, args, w); err != nil {
+		return "", fmt.Errorf("build Dockerfile at %s: %w", args.Dockerfile, err)
```
Lou1415926: nit nit nit!
```suggestion
		return "", fmt.Errorf("build from Dockerfile at %s: %w", args.Dockerfile, err)
```

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -201,6 +212,71 @@ func (c DockerCmdClient) Push(ctx context.Context, uri string, w io.Writer, tags
...
+		args = append(args, "--env", fmt.Sprintf("%s=%s", key, value))
+	}
+
+	//Add environment variables.
+	for key, value := range in.EnvVars {
+		args = append(args, "--env", fmt.Sprintf("%s=%s", key, value))
+	}
+
+	//Add the image name.
+	if in.ImageURI != "" {
```
Lou1415926: Since this is a required field, we probably don't need this `if` statement.

internal/pkg/cli/deploy/workload.go
```diff
@@ -393,24 +409,46 @@ func (img ContainerImageIdentifier) Tag() string {
...
+	}
+	return buildContainerImages(in)
+}
+
+// BuildContainerImages builds the all the images given the build arguments
+func BuildContainerImages(in *BuildImageArgs) error {
+	return buildContainerImages(in)
+}
+
+func buildContainerImages(in *BuildImageArgs) error {
```
Lou1415926: I feel like we could just do ⬇️ - I don't see why calling an exported function from the same package would be a problem. Is there any concern that I'm not thinking of?
```suggestion
	return BuildContainerImages(in)
}

// BuildContainerImages builds the all the images given the build arguments
func BuildContainerImages(in *BuildImageArgs) error {
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -222,6 +223,21 @@ type ContainerImageIdentifier struct {
 	GitShortCommitTag string
 }
 
+// BuildImageArgs represent the input parameters for building and uploading container images.
+type BuildImageArgs struct {
+	Name               string
+	WorkspacePath      string
+	Image              ContainerImageIdentifier
+	Out                *UploadArtifactsOutput
```
Lou1415926: I think `Out` doesn't belong in `BuildImageArgs`, because whatever in `Out` isn't used in image building. We could  do something like:
```go
func BuildContainerImages(in *BuildImageArgs, out *UploadArtifactsOutput)
```

internal/pkg/cli/deploy/workload.go
Lou1415926: In addition, I feel like a better name would be `BuildAndPushContainerImages` 👀

internal/pkg/cli/deploy/workload.go
```diff
@@ -222,6 +223,21 @@ type ContainerImageIdentifier struct {
 	GitShortCommitTag string
 }
 
+// BuildImageArgs represent the input parameters for building and uploading container images.
+type BuildImageArgs struct {
```
Lou1415926: `BuildImageArgs` sounds like the arguments we will pass into `docker build`, which is not what it is. I suggest a naming like `BuildAndPushImagesInput` to note that this is simply a struct that contains on the parameters that are being passed into the specific function named `BuildAndPushImages`.

internal/pkg/cli/deploy/workload.go
```diff
@@ -393,24 +409,46 @@ func (img ContainerImageIdentifier) Tag() string {
...
+		Mft:                d.mft,
+		Out:                out,
+		CustomTag:          d.image.CustomTag,
+		GitShortCommitTag:  d.image.GitShortCommitTag,
+		BuildFunc:          d.repository.BuildAndPush,
+		Login:              d.repository.Login,
+		CheckDockerEngine:  d.docker.CheckDockerEngineRunning,
+		LabeledTermPrinter: d.labeledTermPrinter,
+	}
+	return buildContainerImages(in)
```
Lou1415926: nit nit:
```suggestion
	return buildContainerImages(&BuildImageArgs{
		Name:               d.name,
		WorkspacePath:      d.workspacePath,
		Image:              d.image,
		Mft:                d.mft,
		Out:                out,
		CustomTag:          d.image.CustomTag,
		GitShortCommitTag:  d.image.GitShortCommitTag,
		BuildFunc:          d.repository.BuildAndPush,
		Login:              d.repository.Login,
		CheckDockerEngine:  d.docker.CheckDockerEngineRunning,
		LabeledTermPrinter: d.labeledTermPrinter,
	})
```

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +198,205 @@ func (o *localRunOpts) Execute() error {
...
+	}
+	o.appliedDynamicMft = mft
+	o.repository, err = o.configureClients(o)
+	if err != nil {
+		return err
+	}
+
+	err = o.buildContainerImages(o)
+	if err != nil {
+		return fmt.Errorf("building container image: %w", err)
```
```suggestion
		return err
```

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +198,205 @@ func (o *localRunOpts) Execute() error {
...
+	case *manifest.ScheduledJob:
+		sideCarListInfo = getBuiltSideCarImages(t.Sidecars)
+	case *manifest.LoadBalancedWebService:
+		sideCarListInfo = getBuiltSideCarImages(t.Sidecars)
+	case *manifest.WorkerService:
+		sideCarListInfo = getBuiltSideCarImages(t.Sidecars)
+	case *manifest.BackendService:
+		sideCarListInfo = getBuiltSideCarImages(t.Sidecars)
+	}
+	o.imageInfoList = append(o.imageInfoList, sideCarListInfo...)
```
Lou1415926: After reading the code, I understand now that `sideCarListInfo` contains all the sidecars _whose image are specified via. `image.location`_. The image information of the sidecars that used `image.build` are _already_ in `o.imageInfoList` at this point.

However, this was not obvious. At first, it seemed to me that we were including the sidecar image info twice - once from `buildContainerImages`, and once from `getBuiltSideCarImages`.

I guess the easiest way right now to improve this is just to rename the variables and functions to `sidecarImageLocations`, and maybe add a comment here. I do have the intuition though that we should think about how to refactor these when time permits.

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +198,205 @@ func (o *localRunOpts) Execute() error {
...
+
+	containerNames := o.out.ContainerNames
+	imageNames := o.out.ImageNames
+
+	for i, image := range imageNames {
+		o.imageInfoList = append(o.imageInfoList, imageInfo{
+			containerName: containerNames[i],
+			imageURI:      image,
+		})
+	}
```
Lou1415926: I see we are assuming here that each item in `o.out.ContainerNames` correspond to the item with the same index in `o.out.ImageURI`. This protocol isn't established anywhere, and can only be inferred by looking at the code in `cli/deploy` (which is a different package than where we are now). This is unfortunately error-prone in future development : (. I think we could just have a struct
```go
// in cli/deploy
type ImagePerContainer struct {
 ContainerName string
 ImageURI string
}
```

or is it an option to embed these information into `ContainerImageIdentifier`?

Lou1415926 (commented):
This is quite a huge PR! I haven't finished reviewing all yet, but here are my suggestions so far!

internal/pkg/cli/deploy/workload.go
```diff
@@ -393,18 +416,47 @@ func (img ContainerImageIdentifier) Tag() string {
...
+	return buildImages(in, out)
+}
+
+func buildImages(in *ImageActionInput, out *UploadArtifactsOutput) error {
+	return performImageAction(in, out, in.Builder.Build)
+}
+
+func buildAndPushImages(in *ImageActionInput, out *UploadArtifactsOutput) error {
+	return performImageAction(in, out, in.Builder.BuildAndPush)
+}
```
dannyrandall: seems like we could get rid of these two functions and just call `performImageAction()` directly from `BuildContainerImages()` and `uploadContainerImages()`

internal/pkg/cli/deploy/workload.go
```diff
@@ -544,6 +599,8 @@ type UploadArtifactsOutput struct {
 	AddonsURL                      string
 	CustomResourceURLs             map[string]string
 	StaticSiteAssetMappingLocation string
+	ImageNames                     []string
+	ContainerNames                 []string
```
dannyrandall: would it be easier to just add `imageName` to `ContainerImageIdentifier`? (in the `ImageDigests` map)

internal/pkg/cli/local_run.go
```diff
@@ -31,12 +51,34 @@ type localRunVars struct {
...
+	configureClients     func(o *localRunOpts) (repositoryService, error)
+	labeledTermPrinter   func(fw syncbuffer.FileWriter, bufs []*syncbuffer.LabeledSyncBuffer, opts ...syncbuffer.LabeledTermPrinterOption) clideploy.LabeledTermPrinter
+	unmarshal            func([]byte) (manifest.DynamicWorkload, error)
+	newInterpolator      func(app, env string) interpolator
 }
 
+// type imageInfo struct {
+// 	containerName string
+// 	imageURI      string
+// }
```
dannyrandall:
```suggestion
```

internal/pkg/cli/local_run.go
```diff
@@ -49,12 +91,58 @@ func newLocalRunOpts(vars localRunVars) (*localRunOpts, error) {
...
+		ws:                 ws,
+		ecsLocalClient:     ecs.New(defaultSess),
+		newInterpolator:    newManifestInterpolator,
+		sessProvider:       sessProvider,
+		unmarshal:          manifest.UnmarshalWorkload,
+		sess:               defaultSess,
+		cmd:                exec.NewCmd(),
+		dockerEngine:       dockerengine.New(exec.NewCmd()),
+		labeledTermPrinter: labeledTermPrinter,
+		out:                clideploy.UploadArtifactsOutput{},
```
dannyrandall: you don't need to set this as it's the zero value

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +207,194 @@ func (o *localRunOpts) Execute() error {
...
 	}
 
+	secretsList := make(map[string]string)
+	for _, s := range decryptedSecrets {
+		secretsList[s.Name] = s.Value
+	}
+
+	envVars := make(map[string]string)
+	envVariables := taskDef.EnvironmentVariables()
+	for _, envVariable := range envVariables {
```
dannyrandall:
```suggestion
	for _, envVariable := range taskDef.EnvironmentVariables() {
```

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +207,194 @@ func (o *localRunOpts) Execute() error {
...
+		envVars[envVariable.Name] = envVariable.Value
+	}
+
+	containerPorts := make(map[string]string)
+	containerdef := taskDef.ContainerDefinitions
+	for _, container := range containerdef {
```
dannyrandall:
```suggestion
	for _, container := range taskDef.ContainerDefinitions {
```

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +207,194 @@ func (o *localRunOpts) Execute() error {
...
+			hostPort := aws.Int64Value(portMapping.HostPort)
+			hostPortStr := strconv.FormatInt(hostPort, 10)
+			var containerPort int64
+			if portMapping.ContainerPort == nil {
+				containerPort = hostPort
+			} else {
+				containerPort = aws.Int64Value(portMapping.ContainerPort)
+			}
+			containerportStr := strconv.FormatInt(containerPort, 10)
+			containerPorts[hostPortStr] = containerportStr
```
dannyrandall:
```suggestion
			hostPort := strconv.Itoa(aws.Int64Value(portMapping.HostPort))
			
			containerPort := hostPort
			if portMapping.ContainerPort == nil {
				containerPort = strconv.Itoa(aws.Int64Value(portMapping.ContainerPort))
			}
			containerPorts[hostPortStr] = containerPort
```

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +207,194 @@ func (o *localRunOpts) Execute() error {
...
+	o.appliedDynamicMft = mft
+	o.repository, err = o.configureClients(o)
+	if err != nil {
+		return err
+	}
+
+	err = o.buildContainerImages(o)
+	if err != nil {
+		return err
+	}
```
dannyrandall:
```suggestion
	if err := o.buildContainerImages(); err != nil {
		return err
	}
```

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +207,194 @@ func (o *localRunOpts) Execute() error {
...
+	manifestContent := o.appliedDynamicMft.Manifest()
+	switch t := manifestContent.(type) {
+	case *manifest.ScheduledJob:
+		sidecarImageLocations = getBuiltSidecarImageLocations(t.Sidecars)
+	case *manifest.LoadBalancedWebService:
+		sidecarImageLocations = getBuiltSidecarImageLocations(t.Sidecars)
+	case *manifest.WorkerService:
+		sidecarImageLocations = getBuiltSidecarImageLocations(t.Sidecars)
+	case *manifest.BackendService:
+		sidecarImageLocations = getBuiltSidecarImageLocations(t.Sidecars)
```
dannyrandall, line 275: i think this works?
```suggestion
	case *manifest.ScheduledJob, *manifest.LoadBalancedWebService, *manifest.WorkerService, *manifest.BackendService:
		sidecarImageLocations = getBuiltSidecarImageLocations(t.Sidecars)
```

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +207,194 @@ func (o *localRunOpts) Execute() error {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (o *localRunOpts) getContainerSuffix() string {
+	containerSuffix := fmt.Sprintf("%s-%s-%s", o.appName, o.envName, o.wkldName)
+	return containerSuffix
```
dannyrandall:
```suggestion
	return fmt.Sprintf("%s-%s-%s", o.appName, o.envName, o.wkldName)
```

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +201,187 @@ func (o *localRunOpts) Execute() error {
 	}
 
 	secrets := taskDef.Secrets()
-	_, err = o.ecsLocalClient.DecryptedSecrets(secrets)
+	decryptedSecrets, err := o.ecsLocalClient.DecryptedSecrets(secrets)
 	if err != nil {
 		return fmt.Errorf("get secret values: %w", err)
 	}
 
+	secretsList := make(map[string]string)
```
KollaAdithya: As we know no of secrets before hand. 
can we do `secretsList := make(map[string]string,len(decryptedSecrets))`

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +201,187 @@ func (o *localRunOpts) Execute() error {
...
+	for _, s := range decryptedSecrets {
+		secretsList[s.Name] = s.Value
+	}
+
+	envVars := make(map[string]string)
```
KollaAdithya: similarly for `envars`
```
envVars := make(map[string]string,len(taskDef.EnvironmentVariables()))
```

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +201,187 @@ func (o *localRunOpts) Execute() error {
...
+	case *manifest.WorkerService:
+		sidecarImageLocations = getBuiltSidecarImageLocations(t.Sidecars)
+	case *manifest.BackendService:
+		sidecarImageLocations = getBuiltSidecarImageLocations(t.Sidecars)
+	}
+	o.imageInfoList = append(o.imageInfoList, sidecarImageLocations...)
+
+	// g, ctx := errgroup.WithContext(context.Background())
+	// ctx, cancel := context.WithCancel(ctx)
+	// defer cancel()
```
KollaAdithya:
```suggestion
```

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +201,187 @@ func (o *localRunOpts) Execute() error {
...
+	if err != nil {
+		return fmt.Errorf("run pause container: %w", err)
+	}
+
+	return nil
+
+}
+
+func (o *localRunOpts) runContainers(ctx context.Context, imageInfoList []clideploy.ImagePerContainer, secrets map[string]string, envVars map[string]string) error {
+	var errGroup errgroup.Group
```
KollaAdithya: can we do something like below, because we can combine errorgroup with context
```yaml
g.Go(func() error
err := g.Wait()
```

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +201,187 @@ func (o *localRunOpts) Execute() error {
...
+
+	//channel to receive any error from the goroutine
+	errCh := make(chan error, 1)
+
+	go func() {
+		errCh <- o.dockerEngine.Run(ctx, runOptions)
+	}()
+
+	// go routine to check if pause container is running
+	go func() {
```
KollaAdithya, line 327: does this work?
```
	g.Go(func() error {
		return o.dockerEngine.Run(ctx, runOptions)
	})
	g.Go(func() error {
		for {
			isRunning, err := o.dockerEngine.IsContainerRunning(containerNameWithSuffix)
			if err != nil {
				return fmt.Errorf("check if pause container is running: %w", err)
			}
			if isRunning {
				return nil
			}
			// If the container isn't running yet, sleep for a short duration before checking again.
			time.Sleep(time.Second)
		}
	})
	
	if err := g.Wait(); err != nil {
		return fmt.Errorf("run pause container: %w", err)
	}

Comment (Varun359) [reply internal/pkg/cli/local_run.go:327]: I have tried using this but for some reason it doesn't exit the `runPauseContainers()` at all. It's waiting all the time!!!! So, what's happening is, the pause container is running and I can see the isRunning status true after certain time but it is not exiting the routine. Apparently, the calI is not going to `runContainers()` at all. I have tried doing this. Do you think I am missing something🤔? 

And out of curiosity I have a question as well. Do we have any drawback of the previous approach using error channel?

File: internal/pkg/cli/local_run.go (hunk 33 above)
Comment (Varun359) [reply internal/pkg/cli/local_run.go:275]: This think this will not work we get `t.Sidecars undefined (type any has no field or method Sidecars)`

File: internal/pkg/cli/deploy/workload.go [hunk 40]
```diff
@@ -393,18 +417,38 @@ func (img ContainerImageIdentifier) Tag() string {
...
+		GitShortCommitTag:  d.image.GitShortCommitTag,
+		Login:              d.repository.Login,
+		CheckDockerEngine:  d.docker.CheckDockerEngineRunning,
+		LabeledTermPrinter: d.labeledTermPrinter,
+	}, out, d.repository.BuildAndPush)
+
+}
+
+// BuildContainerImages builds the all the images given the build arguments
+func BuildContainerImages(in *ImageActionInput, out *UploadArtifactsOutput) error {
```
iamhopaul123, line 436: Shall we have similar function signature as `uploadContainerImages` for example

```go
func (d *workloadDeployer) BuildContainerImages(out *UploadArtifactsOutput) error {
```

so that `BuildContainerImages` is mockable in the unit test (and more consistent)?

internal/pkg/cli/deploy/workload.go
```diff
@@ -393,18 +417,38 @@ func (img ContainerImageIdentifier) Tag() string {
 }
 
 func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) error {
```
iamhopaul123:
```suggestion
func (d *workloadDeployer) buildAndPushContainerImages(out *UploadArtifactsOutput) error {
```

internal/pkg/cli/deploy/workload.go
```diff
@@ -393,18 +417,38 @@ func (img ContainerImageIdentifier) Tag() string {
...
+	}, out, d.repository.BuildAndPush)
+
+}
+
+// BuildContainerImages builds the all the images given the build arguments
+func BuildContainerImages(in *ImageActionInput, out *UploadArtifactsOutput) error {
+	return performImageAction(in, out, in.Builder.Build)
+}
+
+func performImageAction(in *ImageActionInput, out *UploadArtifactsOutput, buildFunc func(ctx context.Context, args *dockerengine.BuildArguments, w io.Writer) (string, error)) error {
```
iamhopaul123:
```suggestion
func processContainerImages(in *ImageActionInput, out *UploadArtifactsOutput, buildFunc func(ctx context.Context, args *dockerengine.BuildArguments, w io.Writer) (string, error)) error {
```

internal/pkg/cli/local_run.go
KollaAdithya: Sorry for the late response. I was just thinking of better Readability. If it does not work, you can ignore it.

iamhopaul123 (commented):
Do we need to add any unit test for `dockerengine.go` and `repository.go`?

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +201,182 @@ func (o *localRunOpts) Execute() error {
...
 		return fmt.Errorf("get secret values: %w", err)
 	}
 
+	secretsList := make(map[string]string, len(decryptedSecrets))
+	for _, s := range decryptedSecrets {
+		secretsList[s.Name] = s.Value
+	}
+
+	envVars := make(map[string]string, len(taskDef.EnvironmentVariables()))
+	for _, envVariable := range taskDef.EnvironmentVariables() {
```
dannyrandall: nit
```suggestion
	for _, e := range taskDef.EnvironmentVariables() {
```

internal/pkg/cli/local_run.go
```diff
@@ -106,11 +201,182 @@ func (o *localRunOpts) Execute() error {
...
+				ContainerName: sideCarName,
+				ImageURI:      uri,
+			})
+		}
+	}
+	return sideCarBuiltImageLocations
+}
+
+func (o *localRunOpts) runPauseContainer(ctx context.Context, containerPorts map[string]string) error {
+	containerNameWithSuffix := fmt.Sprintf("%s-%s", pauseContainerName, o.containerSuffix)
```
dannyrandall, line 311: i think we actually want this to be a prefix (same for `runContainers` below) - it should be something like "appName-envName-svcName-pause"

dannyrandall (approved):
just some small things!

iamhopaul123 (approved):

PR #5139, opened 2023-07-28
chore: add option for only initialized workloads in selector

bvtujo:
This PR adds an options pattern to LocalWorkloadSelector so that it can grab all workloads defined locally, not just those represented in the config store. 

This is a prerequisite for initialization during deployment.

When you initialize a LocalWorkloadSelector without `selector.OnlyInitializedWorkloads`, you see something like this when you run `sel.Workload()`:
Screen Shot 2023-07-28 at 12 39 44 PM

This behavior will be reused for the `deploy` environment selector. 


internal/pkg/term/selector/selector.go
```diff
@@ -748,24 +757,24 @@ func (s *LocalWorkloadSelector) Service(msg, help string) (string, error) {
...
-		log.Infof("Only found one service, defaulting to: %s\n", color.HighlightUserInput(serviceNames[0]))
-		return serviceNames[0], nil
+	options, err := s.generateOptions(storeServices, wsServiceNames, svcWorkloadType)
+	if err != nil {
+		return "", err
 	}
 
-	selectedServiceName, err := s.prompt.SelectOne(msg, help, serviceNames, prompt.WithFinalMessage(svcNameFinalMsg))
+	if len(options) == 1 {
+		log.Infof("Only found one service, defaulting to: %s\n", color.HighlightUserInput(options[0].Value))
```
huanjani: Nit: this is totally not on you, but I wonder if we can take this opportunity to change all of these from "Only found" to "Found only"...? 😬

internal/pkg/term/selector/selector.go
```diff
@@ -778,27 +787,83 @@ func (s *LocalWorkloadSelector) Job(msg, help string) (string, error) {
...
+	// If there are no local workload names, error out.
+	if len(wsWlNames) == 0 {
+		return nil, fmt.Errorf("no %s found in workspace", pluralNounString)
+	}
+	// Get the list of un-initialized workloads that are present in the workspace and add them to options.
+	unInitializedLocalWorkloads := filterOutStrings(wsWlNames, initializedLocalWorkloads)
+	for _, wl := range unInitializedLocalWorkloads {
+		options = append(options, prompt.Option{
+			Value: wl,
+			Hint:  "un-initialized",
```
huanjani: nit
```suggestion
			Hint:  "uninitialized",
```

internal/pkg/term/selector/selector.go
```diff
@@ -778,27 +787,83 @@ func (s *LocalWorkloadSelector) Job(msg, help string) (string, error) {
...
+
+// LocalWorkloadsOnly modifies LocalWorkloadSelector to show ALL workloads in the workspace,
+// regardless of whether they exist in the config store.
+var LocalWorkloadsOnly SelectOption = func(s *LocalWorkloadSelector) {
+	s.localWorkloadsOnly = true
+}
+
+// Workload fetches all jobs and services in a workspace and prompts the user to select one.
+// It can optionally select only initialized workloads which exist in the app (default behavior)
+// or
```
huanjani: ??

bvtujo: 🙊

huanjani (approved):
LGTM!

internal/pkg/term/selector/selector.go
```diff
@@ -295,11 +300,15 @@ func NewConfigSelector(prompt Prompter, store configLister) *ConfigSelector {
 
 // NewLocalWorkloadSelector returns a new selector that chooses applications and environments from the config store, but
 // services from the local workspace.
-func NewLocalWorkloadSelector(prompt Prompter, store configLister, ws workspaceRetriever) *LocalWorkloadSelector {
-	return &LocalWorkloadSelector{
+func NewLocalWorkloadSelector(prompt Prompter, store configLister, ws workspaceRetriever, options ...SelectOption) *LocalWorkloadSelector {
```
iamhopaul123, line 303: Maybe we can make `NewLocalWorkloadSelector` to take a struct as input so that we don't need `SelectOption`.

internal/pkg/term/selector/selector.go
```diff
@@ -778,27 +787,84 @@ func (s *LocalWorkloadSelector) Job(msg, help string) (string, error) {
...
+	var options []prompt.Option
+
+	// Get the list of initialized workloads that are present in the workspace
+	initializedLocalWorkloads := filterWlsByName(storeWls, wsWlNames)
+	for _, wl := range initializedLocalWorkloads {
+		options = append(options, prompt.Option{Value: wl})
+	}
+
+	// Return early if we're only looking for config store workloads.
+	if !s.allLocalWorkloads {
```
iamhopaul123: I feel like `allLocalWorkloads` should be `true` by default for `LocalWorkloadSelector`. Maybe consider renaming the var as `WithOnlyInitialized`? And our existing callers (who wish to keep the current behavior to only include initialized workloads) for `LocalWorkloadSelector`'s methods will set `WithOnlyInitialized` to `true`.

bvtujo: Done! I flipped the behavior.

iamhopaul123 (approved):

PR #5141, opened 2023-07-28
fix: stop previous versions of copilot from downgrading stack set

### LINKED ISSUE #5140: cannot unmarshal !!str `See #5140` into []string
## Why am I seeing this error?

In v1.28, Copilot introduced our first non ECS-based workload type, Static Site. As part of that, we added new logic to support workloads that didn't need an ECR repo.

After release, we discovered a bug that occurred in the following scenario:
1. Upgrade the Application Stack Set by creating a new service or upgrading an application
  ```
  ➜ copilot@v1.28.0 svc init
  # or
  ➜ copilot@v1.28.0 app upgrade
  ```
2. Downgrade the Application Stack Set by creating a new service with an older version of Copilot
  ```
  # any Copilot version <= v1.27.0
  ➜ copilot@v1.27.0 svc init
  ```

If step 1 was done with >= v1.29.1, then step 2 should error out with a link to this issue.

If step 1 was done with v1.28.0 or v1.29.0, any services in the Application will have their ECR repos removed from the Application Stack Set. While this **does not result in any service downtime**, it makes it impossible to deploy a _new_ version of a service without [a workaround](#v1.28.0,-v1.29.0). If you have images in your ECR repo, the repo should still exist in ECR, just not in the CloudFormation Stack Set.

## What is Copilot doing to mitigate this issue?
While we can't patch all of older versions of Copilot that are out there and force everyone to upgrade, we introduced a new version downgrade checker feature to prevent this class (upgrade followed by downgrade) of bug in the future. Stack downgrades will now error out, and we added a new flag `--allow-downgrade` to explicitly allow a downgrade to all `init`, `deploy` and `package` commands.

Additionally, for Application Stack Sets deployed by Copilot >= v1.29.1, we added a line to the template that causes Copilot <= v1.27 to error on operations that modify the Application Stack Set with a link to this issue.

### >= v1.29.1
If you got this error message:
```
cannot unmarshal !!str `See #5140` into []string
```
This means that someone has updated your Application Stack Set with Copilot >= v1.29.1. Upgrade your version of Copilot to the latest version, and run your command again.

### v1.28.0, v1.29.0
If you are seeing one of these errors:
```
Container.image repository should not be null or empty
```
```
image is not supported by App Runner: @sha256:<hash>
```
```
failed to normalize image reference ":<hash>". Launch a new task to retry.
```

This likely means your Application Stack Set was deployed with Copilot v1.28.0 or v1.29.0 and subsequently downgraded. The error is a result of your ECR repos being removed from the Application Stack Set. To resolve this issue, first upgrade your version of Copilot to the latest version. Then, do the following **_for each service in your application_**.
1. Delete the service from SSM Parameter Store
  a. Open the System Manager console in the same region/account the `<APP NAME>-infrastructure-roles` CloudFormation stack is deployed in.
  b. Search for the parameter named `/copilot/applications/<APP NAME>/components/<SVC NAME>`
  c. Remember the "type" field from the parameter value
  d. Delete the parameter
2. Check if the ECR repo exists by opening the ECR console, and searching for the repo `<APP NAME>/<SVC NAME>`. If it does, do the following to backup the images in it:
  a. Install [`skopeo`](https://github.com/containers/skopeo).
  b. Click on the "View push commands" button in the ECR repo console.
  c. Login to ECR by running the first step (`aws ecr get-login-password ...`) on your machine, but change the word "`docker`" to "`skopeo`".
  d. Copy the ECR repo URL from the 4th step - the bit after "`docker push`". It should be in the format `<ACCOUNT ID>.dkr.ecr.<REGION>.amazonaws.com/<APP NAME>/<SVC NAME>`. Note that you should not copy the `:latest` tag.
  e. Create a new folder to store all of the images in: `mkdir <SVC NAME>`
  f. Download all of the image digests to your machine from the ECR repo:
    ```
    aws ecr list-images --repository-name <APP NAME>/<SVC NAME> --output json --max-results 1000 --no-cli-pager | jq -r '.imageIds[].imageDigest' | xargs -P1 -n1 -I{} skopeo copy --all --preserve-digests docker://<ECR REPO URL>@{} dir:<SVC NAME>/{}
    ```
   g. Verify the directories created in the `<SVC NAME>` directory match all of the image digests in your ECR repo.
   h. Delete the ECR Repo using the console.
3. Recreate the service. This will add the service back to SSM Parameter Store and recreate the ECR repo.
```
➜ copilot svc init --name <svc name> --svc-type <type from SSM parameter>
```
4. If you had to do step two, do the following to push all of the saved images to the new ECR repo:
  a. Run the command:
    ```
    ls -1 <SVC NAME> | xargs -P1 -n1 -I{} skopeo copy --all --preserve-digests dir:<SVC NAME>/{} docker://<ECR REPO URL>
    ```
    b. Verify the new ECR Repo has the correct set of image digests.

dannyrandall:
Fixes #5140 


Lou1415926 (approved):
noice thanks!

KollaAdithya (approved):
ty!

Issue #5124, opened 2023-07-25
Docker build log should be printed when docker build fails


jie-lyu:
Currently, when CI=true, when docker build fails in `copilot svc package`, only 

```
upload resources required for deployment for frontend: build and push the image "frontend": build Dockerfile at /path/to/Dockerfile: building image: exit status 1
```

is printed. AFAIK, there is no method to investigate what really is happening.

When CI != true, a few more lines is printed, but still not enough information is provided.

iamhopaul123:
Hello @jie17. This is a very good call. We are prioritizing this enhancement now.

geoff-mwi:
Is there any workaround either by addons, overrides, forks or build emulation that can expose the Docker build messages/errors? 

Our pipeline intermittently fails on the Docker build step (even though local Docker build works perfectly) and similar to @jie17 we have no way of understanding what went wrong. 

For our case, rerunning our pipeline shortly after with no changes results in a failure. Waiting until the next day and rerunning the pipeline again results in a success. My assumption for our particular error is there is some rate limiting happening maybe with docker pull.

The docker build messages are also outputted as if being run interactively which is not ideal. If there is an enhancement to output the docker build messages, would it be possible to change the docker build progress output to `plain` instead of `tty`?

KollaAdithya:
Hello @geoff-mwi !

one way I can think of is to see if the docker build is successful before deploying the service. i.e;
you can run `docker build . -f Dockerfile`. Even though it adds an extra burden on your side.

We are currently working on to solve this issue on docker build output. I had a PR([5148](https://github.com/aws/copilot-cli/pull/5148)) that will solve the issue if you just have single main container image in the manifest. This will show the output of docker build if it fails as well.

In case of multiple container images i.e (sidecars + main container) images issue still persists.

Regarding the docker pull rate issue, you can refer to this [workaround ](https://github.com/aws/copilot-cli/issues/1869#issuecomment-765643674) that will help you solve this issue.

> would it be possible to change the docker build progress output to plain instead of tty

If you want to have docker build output to be `plain` you need to set the `CI=true`.

I also had a question which version of copilot are you currently using?

geoff-mwi:
Hi @KollaAdithya, thank you for the response.

[...]

Issue #5067, opened 2023-07-11
`image.build` log sometimes swallows useful debug information


Lou1415926:
![thescreenshot](https://github.com/aws/copilot-cli/assets/79273084/429a19b5-24c3-4f15-9d91-0e220a017c35)

As shown in the screenshot, the `image.build` progress displayer occasionally erases logs that happen to be useful. 

I think we can:
- [ ] Fix this in the case where there are multiple image builds happening in parallel.
- [x] When there is only one image to build, remove the rolling display.

dannyrandall:
We addressed the second item in [v1.30.0](https://github.com/aws/copilot-cli/releases/tag/v1.30.0), by removing the rolling display if there is only a single image to build.

ryanc-scalegrowth:
Coming back to this topic as it's rather frustrating to debug docker build errors when building in the pipeline with multiple images.
Is there any way to display docker build errors on failure in the pipeline? Even a quick and dirty workaround for now would be good.

PR #5148, opened 2023-08-01
chore: default the writer of docker build to `Stderr` incase of single image

KollaAdithya:
related #5124, #5067
This PR will remove the rolling display when there is only one image to build.

### DIFF
--- a/internal/pkg/cli/deploy/workload.go
+++ b/internal/pkg/cli/deploy/workload.go
@@ -408,7 +408,30 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	if err != nil {
 		return fmt.Errorf("login to image repository: %w", err)
 	}
+	isMultipleContainerImages := len(buildArgsPerContainer) > 1
+	if isMultipleContainerImages {
+		return d.buildContainerImagesInParallel(uri, buildArgsPerContainer, out)
+	}
+	return d.buildSingleContainerImage(uri, buildArgsPerContainer, out)
+}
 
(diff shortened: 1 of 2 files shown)


internal/pkg/cli/deploy/workload.go
```diff
@@ -401,6 +401,7 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	if len(buildArgsPerContainer) == 0 {
 		return nil
 	}
+	isMultipleContainerImages := len(buildArgsPerContainer) != 1
```
huanjani: maybe??
```suggestion
	isMultipleContainerImages := len(buildArgsPerContainer) > 1
```

huanjani:
Could we add a unit test for multiple images?

KollaAdithya:
I think we already had a test case for this scenario of multiple images over here
https://github.com/aws/copilot-cli/pull/5148/files#diff-cc7a53c33c7639692deae85af8e5ec1187f3ef5dc845b29f53f9ae698043ef7cR311

huanjani (approved):
lgtm!

internal/pkg/cli/deploy/workload.go
```diff
@@ -401,6 +401,7 @@ func (d *workloadDeployer) uploadContainerImages(out *UploadArtifactsOutput) err
 	if len(buildArgsPerContainer) == 0 {
 		return nil
 	}
+	isMultipleContainerImages := len(buildArgsPerContainer) > 1
```
iamhopaul123: Does it make more sense for us to call a separate function if `isMultipleContainerImages` is true (after `repository.Login()`)? The reason is we probably don't need the parallel part at all for single build case and it'll be easier to read/maintain

iamhopaul123 (approved):
awesome :shipit:

PR #5154, opened 2023-08-03
Add support for podman image digests

### LINKED ISSUE #3170: Support Podman either as alias to Docker or natively

mjrlee:
Adds a new "engine" build argument that can be used to handle podman's different way of handing image digests.

Fixes #3170


internal/pkg/manifest/workload.go
```diff
@@ -397,6 +406,7 @@ type DockerBuildArgs struct {
 	Args       map[string]string `yaml:"args,omitempty"`
 	Target     *string           `yaml:"target,omitempty"`
 	CacheFrom  []string          `yaml:"cache_from,omitempty"`
+	Engine     *string           `yaml:"engine,omitempty"`
```
iamhopaul123: Since essentially `nil` and `""` are the same case (we'll set to default value "docker"), maybe we can use `string` instead of `*string`

internal/pkg/manifest/workload.go
```diff
@@ -237,6 +238,14 @@ func (i *ImageLocationOrBuild) cacheFrom() []string {
 	return i.Build.BuildArgs.CacheFrom
 }
 
+// engine returns the engine to use for building the image if it exists, otherwise "docker"
+func (i *ImageLocationOrBuild) engine() *string {
+	if i.Build.BuildArgs.Engine != nil {
+		return i.Build.BuildArgs.Engine
+	}
+	return aws.String("docker")
```
iamhopaul123: Can we make "docker" a constant?

internal/pkg/docker/dockerengine/dockerengine.go
```diff
@@ -170,12 +171,22 @@ func (c DockerCmdClient) Login(uri, username, password string) error {
 }
 
 // Push pushes the images with the specified tags and ecr repository URI, and returns the image digest on success.
-func (c DockerCmdClient) Push(ctx context.Context, uri string, w io.Writer, tags ...string) (digest string, err error) {
+func (c DockerCmdClient) Push(ctx context.Context, uri string, engine string, w io.Writer, tags ...string) (digest string, err error) {
```
iamhopaul123, line 236: i feel like we should have a different function for podman push (same signature as `Push` it's just with a different name), so that we don't need to change the function signature and also the push function is quite different when the engine type is podman.

iamhopaul123 (commented):
This is great and thank you for contributing! Just wanted to double check: have you tested this feature locally and verified you would be able to build images using podman after this change? Or is there anything else that needs to be done for the full workflow?

ammerzon:
@mjrlee, are there any updates on this, or do you need support? 😁

internal/pkg/docker/dockerengine/dockerengine.go
mjrlee: I did consider this, but it seems like there's still a lot of shared functionality that would be a pain to update in multiple places.

The alternative could be to build a new type PodmanCmdClient that implements ContainerLoginBuildPusher by just calling the existing methods - but again that seemed to add more complexity than it removed.

mjrlee:
Sorry, I haven't had much time to look at this recently.

I can confirm I'm currently using it with podman and it's working well.

I've resolved a couple of comments on the PR, and replied to another.

